PR #3, opened 2020-11-10
Block the use of `latest` OCI images by default

autodidaddict:
Firstly, there's now a pattern established for getting parameters from the host builder to the control plane actor, and there are some "unused" variables like `max_actors` and `max_providers` that are there as placeholders to make it fairly obvious how to add more tuning / configuration parameters to the control plane.

Secondly, the control plane itself isn't doing any enforcement of the `allow_latest` flag because .. .well, the CP isn't implemented yet ;) If you want to know where it _will_ be, check out the placeholder functions on the `ControlInterface` struct, which is an object handed to control plane implementations (e.g. our NATS / lattice CP).

The limited enforcement of the "allow latest" flag that _is_ in this host is from the host functions that add/load actors and providers. If  the image reference ends with `:latest` and this flag is _false_, then an attempt to download the bytes will return an error, which then will propogate out to whatever consumer triggered that download (which should ultimately include the soon-to-be-implemented control interface).

### DIFF
--- a/crates/wascc-host/src/oci.rs
+++ b/crates/wascc-host/src/oci.rs
@@ -7,7 +7,8 @@ pub(crate) const OCI_VAR_PASSWORD: &str = "OCI_REGISTRY_PASSWORD";
 
 pub(crate) async fn fetch_oci_bytes(img: &str, allow_latest: bool) -> Result<Vec<u8>> {
     if !allow_latest && img.ends_with(":latest") {
-        return Err("Fetching mutable images tagged 'latest' is prohibited".into());
+        return Err(
+            "Fetching images tagged 'latest' is currently prohibited in this host. This option can be overridden".into());
     }
     let cfg = oci_distribution::client::ClientConfig::default();
     let mut c = oci_distribution::Client::new(cfg);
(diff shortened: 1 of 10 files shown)


crates/wascc-host/src/oci.rs
```diff
@@ -5,7 +5,10 @@ use std::str::FromStr;
 pub(crate) const OCI_VAR_USER: &str = "OCI_REGISTRY_USER";
 pub(crate) const OCI_VAR_PASSWORD: &str = "OCI_REGISTRY_PASSWORD";
 
-pub(crate) async fn fetch_oci_bytes(img: &str) -> Result<Vec<u8>> {
+pub(crate) async fn fetch_oci_bytes(img: &str, allow_latest: bool) -> Result<Vec<u8>> {
+    if !allow_latest && img.ends_with(":latest") {
+        return Err("Fetching mutable images tagged 'latest' is prohibited".into());
```
brooksmtownsend, line 10: Do you see value in including in the error something along the lines of `you can override this with the --allow-latest flag`?

brooksmtownsend (commented):
Largely LGTM, one question

crates/wascc-host/src/oci.rs
autodidaddict: Good point. I pushed a change to this, though I don't want to mention `--allow-latest` within the wascc-host code because it doesn't know which binary instantiated it.

brooksmtownsend (approved):
LGTM

PR #4, opened 2020-11-11
Initial refactor for 'no fail' actor start

autodidaddict:
After talking to the people who actually know what they're doing in the Actix gitter/discord, I found that I was using a bit of an anti-pattern. It should be impossible for an actor to fail to start, and there were a couple of places, especially in the capability provider host actor, where I was pulling shenanigans like `try_new()`, which I then hid behind a builder (that also returned a result).

Instead, the recommended practice is that all actors should be 100% guaranteed to start, and if you need to set an initial state that could be rejected or invalid, then you do so through sending that message to the actor, and the actor stops itself if there's an error.

So far I've only refactored the host controller's interaction with the native capability provider host. After we agree on the cleanest way to implement this, I can poke around the rest of all the actors and make them conform to this new pattern.

### DIFF
--- a/crates/wascc-host/src/actors/actor_host.rs
+++ b/crates/wascc-host/src/actors/actor_host.rs
@@ -10,23 +10,34 @@ use futures::executor::block_on;
 use wapc::{WapcHost, WasiParams};
 use wascap::prelude::{Claims, KeyPair};
 
+#[derive(Default)]
 pub(crate) struct ActorHost {
+    state: Option<State>,
+}
+
+struct State {
     guest_module: WapcHost,
     claims: Claims<wascap::jwt::Actor>,
     mw_chain: Vec<Box<dyn Middleware>>,
     image_ref: Option<String>,
 }
 
-impl ActorHost {
-    pub fn new(
-        actor_bytes: Vec<u8>,
-        wasi: Option<WasiParams>,
-        mw_chain: Vec<Box<dyn Middleware>>,
-        signing_seed: String,
-        image_ref: Option<String>,
-    ) -> ActorHost {
-        let buf = actor_bytes.clone();
-        let actor = WasccActor::from_slice(&buf).unwrap();
+#[derive(Message)]
+#[rtype(result = "Result<()>")]
+pub(crate) struct Initialize {
+    pub actor_bytes: Vec<u8>,
+    pub wasi: Option<WasiParams>,
+    pub mw_chain: Vec<Box<dyn Middleware>>,
+    pub signing_seed: String,
+    pub image_ref: Option<String>,
+}
+
+impl Handler<Initialize> for ActorHost {
+    type Result = Result<()>;
+
+    fn handle(&mut self, msg: Initialize, ctx: &mut Self::Context) -> Self::Result {
+        let buf = msg.actor_bytes.clone();
+        let actor = WasccActor::from_slice(&buf)?;
 
         #[cfg(feature = "wasmtime")]
         let engine = wasmtime_provider::WasmtimeEngineProvider::new(&buf, wasi);
@@ -34,11 +45,14 @@ impl ActorHost {
         let engine = wasm3_provider::Wasm3EngineProvider::new(&buf);
 
         let c = actor.token.claims.clone();
+        let c2 = c.clone();
+        let c3 = c.clone(); // TODO: I can't believe I have to do this to make the [censored] borrow checker happy
+        let seed = msg.signing_seed.to_string();
 
         let guest = WapcHost::new(Box::new(engine), move |_id, bd, ns, op, payload| {
             crate::dispatch::wapc_host_callback(
-                KeyPair::from_seed(&signing_seed).unwrap(),
-                actor.token.claims.clone(),
+                KeyPair::from_seed(&seed).unwrap(),
+                c2.clone(),
                 bd,
                 ns,
                 op,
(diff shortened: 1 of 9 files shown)


autodidaddict:
@brooksmtownsend just pushed a new batch that removes potentially panic-inducing unwraps from the actor startup, adopting the new "optional state with initializer" pattern.

crates/wascc-host/src/actors/actor_host.rs
```diff
@@ -10,35 +10,49 @@ use futures::executor::block_on;
...
+        let c2 = c.clone();
+        let c3 = c.clone(); // TODO: I can't believe I have to do this to make the [censored] borrow checker happy
+        let seed = msg.signing_seed.to_string();
 
         let guest = WapcHost::new(Box::new(engine), move |_id, bd, ns, op, payload| {
             crate::dispatch::wapc_host_callback(
-                KeyPair::from_seed(&signing_seed).unwrap(),
-                actor.token.claims.clone(),
+                KeyPair::from_seed(&seed).unwrap(),
+                c2.clone(),
```
brooksmtownsend, line 55: This needs to be cloned again too? Man this borrow checker 😢

crates/wascc-host/src/actors/actor_host.rs
```diff
@@ -47,15 +61,59 @@ impl ActorHost {
...
+                        subscriber: recipient,
+                    })
+                    .await
+                });
+                let pc = PutClaims{ claims: c.clone() };
+                let r = block_on(async move {
+                    if let Err(e) = b2.send(pc).await {
+                        error!("Actor failed to advertise claims to bus: {}", e);
+                        ctx.stop();
+                        Err("Failed to advertise bus claims".into())
```
brooksmtownsend, line 82: So the only situation where an actor _should_ fail to start is if it fails to advertise its claims, e.g. capability contract IDs that it's allowed to interact with? What if we started it with no claims in the event that it fails to advertise? I'm specifically referring to your comment that an actor should be able to start 100% of the time even with invalid state

crates/wascc-host/src/actors/actor_host.rs
```diff
@@ -47,15 +61,59 @@ impl ActorHost {
...
+            }
             Err(e) => {
-                error!("Failed to instantiate waPC host: {}", e);
-                panic!();
+                error!(
+                    "Failed to create a WebAssembly host for actor {}",
+                    actor.token.claims.subject
+                );
+                ctx.stop();
+                Err("Failed to create a raw WebASsembly host".into())
```
brooksmtownsend, line 116: mega nitpick: Extra capital letter here😉

brooksmtownsend (changes requested):
I might be taking the "Actor should always be able to start" too literally, I don't want to encourage creating weird program state to satisfy that. I am interested especially in the actor host Initialize handler about what this should look like.

crates/wascc-host/src/actors/actor_host.rs
autodidaddict: An actor host now has 3 states - "just started (no state)", "running (state)", and "stopping/stopped"
there's no downside to the microscopic gap between just started and running with state, because if converting to running with state fails, the actor host stops itself.

autodidaddict: there's a semantic difference here that's likely confusing. The `ActorHost` _actix actor_ must never fail to start. An actor _can_ fail to initialize its state and take care of all the prep (subscribng to the bus, creating an actor object, etc)

crates/wascc-host/src/actors/actor_host.rs
autodidaddict: It's really disgusting how many times I have to build a "preamble" full of clones all so that I can move detached versions of self-owned fields into a closure.

autodidaddict: It's harder in some places with actix because `block_on(async move { ... }))` is actually a _double closure_, since the input to `block_on` is a closure, and then inside the async block is also a closure.

crates/wascc-host/src/actors/actor_host.rs
brooksmtownsend: The semantics are getting to me a bit. Our previous validation of waSCC actors is not affected by the Actix sentiment, and it's totally fine for us to deny a waSCC actor from starting, but undesirable for the Actix actor (ActorHost) to fail to start.

brooksmtownsend: Just confirming what I'm understanding now... maybe I can call them wactors / wascctors and aactors / actixcors to keep them straight in my head

crates/wascc-host/src/actors/actor_host.rs
brooksmtownsend: Wild. Good to know

crates/wascc-host/src/actors/actor_host.rs
autodidaddict: That's correct. In Actix, you should consider it a _requirement_ that an actor be able to "start" because that really just requires you supplying an instance to a factory. In other words, the factory cannot fail to produce an "empty" actor. Once the "empty" actor has been cut from the factory, it can then have any state you like, which can include "failed to initialize"

autodidaddict: I struggled as well, the `*Host` structs are all Actix actors that wrap wascc abstractions like actors and providers.

autodidaddict: And yes, our previous validation should still exist - you should be able to find the calls to the authorization checker once we establish an instance of an actor and its claims.

brooksmtownsend: Great, thanks for the explanation 😄

brooksmtownsend (approved):

brooksmtownsend (approved):
Those code pieces look a lot cleaner without the header referenced constantly.

PR #14, opened 2020-11-16
First test of implementation of lattice RPC over nats

autodidaddict:
The code should be ready to roll, though I will likely be adding more lattice RPC tests over time. Remember this is only the lattice RPC - the control plane is now a completely separate thing.


autodidaddict:
With this latest commit, if you run `cargo test` at the root of the project directory with a vanilla NATS server running, you'll see the current issue I'm having, which is a NATS timeout attempting to make the client request. The flow is as follows:

1. test essentially curls the http server provider
1. provider uses its dispatch instance to to send an `Invocation` to the bus
1. bus attempts to find a local subscriber and fails
1. bus defers the invocation to the lattice provider 
1. lattice provider (the `crates/nats-rpc` crate) publishes the invocation on the RPC subject
1. the registered subscriber to the actor RPC subject grabs the inbound invocation
1. invocation is then sent to the bus via the `BusDispatcher` (you'll see "local bus invocation" among the 80 million println statements)
1. actor then _successfully_ handles the request and returns an appropriate response
1. It doesn't look like the response makes it into the NATS server (run `nats -DV` to see the detailed flow)
1. the client waiting back a few steps ago times out
1. the timeout produces an "error" style invocation response
1. the http server provider returns an HTTP 500 because it was given invalid response (there's no HttpResponse struct to deserialize from the body of the invocation response because it's an error)

So, the problem at hand ... :thinking: why is the NATS RPC client timing out and, possibly related, why does it look like the reply to the inbound RPC isn't making it back out onto NATS?

autodidaddict:
It actually may appear now that the NATS part is working.. Running the test I see the "all good" statement from , but then I don't see either the success or response message from https://github.com/wascc/wasccd/blob/lattice-rpc/crates/wascc-host/src/messagebus/handlers.rs#L242 . I do see this message just fine when I'm making a single-pass invocation running the no-lattice tests... Maybe it's the head blocked on the tail problem here?

autodidaddict:
The more I look at it (and the more println statements I inject) the more it looks like that line of code there on 242 doesn't finish.. I see none of my debug statements after that line, which incurs the timeout, which then....etc

crates/wascc-host/src/messagebus/rpc_client.rs
```diff
@@ -151,19 +190,54 @@ impl Handler<LinkInbound> for RpcClient {
         }
     }
 }
-// Publish a link to the RPC bus
+// Publish a link definition to the RPC bus
 impl Handler<AdvertiseBinding> for RpcClient {
```
brooksmtownsend, line 194: Should this be renamed to AdvertiseLink, or is that work already being done in https://github.com/wascc/wasccd/pull/12?

crates/wascc-host/src/messagebus/rpc_client.rs
```diff
@@ -151,19 +190,54 @@ impl Handler<LinkInbound> for RpcClient {
...
+        let ld = LinkDefinition {
+            actor: msg.actor,
+            contract_id: msg.contract_id,
+            link_name: msg.binding_name,
+            provider_id: msg.provider_id,
+            values: msg.values,
+        };
+        let nc = self.nc.clone().unwrap();
+        let subject = links_subject(&self.ns_prefix);
+        let bytes = serialize(&ld).unwrap(); // we should never fail or own serialize
```
brooksmtownsend, line 208: should this be "our own serialize"

cmon it's not a PR review from me if I don't nitpick spelling somewhere right

crates/wascc-host/src/middleware/runner.rs
```diff
@@ -36,7 +36,6 @@ pub(crate) async fn invoke_actor(
         error!("Middleware pre-invoke failure: {}", e);
         return Err(e);
     } else {
-        println!("invoking");
         match target.send(inv.clone()).await {
             Ok(ir) => {
                 println!("done");
```
brooksmtownsend, line 41: This seems like a temporary `println` but it should be converted to a `trace!`

brooksmtownsend (changes requested):
A few questions for my own understanding, a few changes requested, largely LGTM 👍

crates/wascc-host/src/auth.rs
```diff
@@ -10,7 +10,7 @@ use wascap::jwt::{Actor, Claims};
 /// impossible to override the base behavior of checking that an actor's embedded JWT contains the right
 /// capability attestations.
 pub trait Authorizer: CloneAuthorizer + Sync + Send {
-    /// This check is performed during the `add_actor` call, allowing the custom authorizer to do things
+    /// This check is performed during the `[start_actor](crate::Host::start_actor)` call, allowing the custom authorizer to do things
```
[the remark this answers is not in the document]
autodidaddict: It doesn't yet, but in the next stable drop of Rust rust docs will support using rust paths to link to related docs. It's going to be amazing.

crates/wascc-host/src/host_controller.rs
```diff
@@ -104,6 +102,52 @@ impl Actor for HostController {
...
+                    // * Send directly to the provider rather than via the bus
+                    // * so we can guarantee we'll never infinite loop this invocation
+                    if let Err(_) = target.send(inv).await {
+                        error!("Capability provider failed to handle link enable call");
+                    }
+                }
+                .into_actor(self),
+            )
+        } else {
+            Box::pin(async {}.into_actor(self))
```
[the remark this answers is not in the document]
autodidaddict: Basically. It's a future that will complete immediately and when completed, returns `()`

crates/wascc-host/src/messagebus/rpc_client.rs
autodidaddict: It should be renamed but you're right, there's already an open PR for that.

crates/wascc-host/src/messagebus/rpc_client.rs
autodidaddict: (╯°□°)╯︵ ┻━┻

crates/wascc-host/src/middleware/runner.rs
autodidaddict: Good catch, I thought i'd gotten rid of all the temporary `println!`

brooksmtownsend (approved):

Issue #19, opened 2020-12-01
Control interface needs auction and live update support


autodidaddict:
Implement auction and live update support in the control interface.

This issue came out of the need to stabilize the main branch to unblock other activities.

Issue #31, opened 2020-12-08
Need the ability to use a local OCI registry for CI builds and offline development


autodidaddict:
This is also needed for testing. There are a couple of use case scenarios where I'll need to put a particular revision of a file into an OCI registry so that  I can then attempt to do a live update , and I can't really pull this off using a public OCI registry.

Additionally, having this ability to have a "ready to use" OCI in our build pipelines will cut down on build times and probably keep costs down  :smile: 

@billyoung Any thoughts on approaches here?

autodidaddict:
It appears as though this is what we need to do in order to have a local registry: https://docs.docker.com/registry/deploying/

billyoung:
The local development piece of this should be solved by https://github.com/wasmCloud/wasmCloud/pull/43

Happy to assist in adding similar functionality to the CI pipeline, but would need to understand more about what we want the flow to look like. @autodidaddict

brooksmtownsend:
This looks like it was closed in #43, @autodidaddict?

autodidaddict:
Yeah I think that docker-compose works for manual, ad-hoc local availability.

As we discussed, we _could_ fake a cache so that the wasmcloud host running in the CI pipeline just uses the binaries we deem appropriate from the cache.

PR #32, opened 2020-12-08
Initial implementation of actor update functionality

autodidaddict:
There's no integration test (yet) for the live update stuff because we need our own local OCI (check out the corresponding issue #31 ) that we can populate with multiple test modules. Right now the integration test for updating actors live requires us to replace an actor with that same actor's image from an OCI, and I can't mess up the public OCI with test signing keys, etc.

All the goodies should be here, however.

Also note that the host builder now has an option for enabling live updates. They will be _disabled_ by default.

### DIFF
--- a/crates/wasmcloud-host/src/actors/actor_host.rs
+++ b/crates/wasmcloud-host/src/actors/actor_host.rs
@@ -1,16 +1,15 @@
 use crate::actors::WasccActor;
 use crate::control_interface::ctlactor::{ControlInterface, PublishEvent};
-use crate::control_interface::events::TerminationReason;
+
 use crate::dispatch::{Invocation, InvocationResponse, WasccEntity};
 use crate::hlreg::HostLocalSystemService;
-use crate::messagebus::{
-    AdvertiseClaims, EnforceLocalActorLinks, MessageBus, PutClaims, Subscribe,
-};
+use crate::messagebus::{AdvertiseClaims, MessageBus, PutClaims, Subscribe};
 use crate::middleware::{run_actor_post_invoke, run_actor_pre_invoke, Middleware};
 use crate::{ControlEvent, Result};
 use actix::prelude::*;
 use futures::executor::block_on;
-use wapc::{WapcHost, WasiParams};
+use wapc::WapcHost;
+use wascap::jwt::TokenValidation;
 use wascap::prelude::{Claims, KeyPair};
 
 #[derive(Default)]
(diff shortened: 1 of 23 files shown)


autodidaddict:
@brooksmtownsend going to defer creating the integration test for live updates until after the oci-distribution crate has stabilized push support since that'll be required for the test.

autodidaddict:
Given the defer, is this PR good to go @brooksmtownsend ?

autodidaddict:
Closes #19

crates/wasmcloud-host/src/actors/actor_host.rs
```diff
@@ -24,113 +23,284 @@ struct State {
...
+            if new_md.rev.unwrap_or(0) <= old_md.rev.unwrap_or(0) {
+                return Err(
+                    "Cannot live update if the new module is not a higher revision number".into(),
+                );
+            }
+            if new_md.caps.as_ref().unwrap_or(&vec![]).len()
+                > old_md.caps.as_ref().unwrap_or(&vec![]).len()
+            {
+                warn!("Live update warning: new actor has more claims than the previous revision.");
             }
```
brooksmtownsend, line 295: I don't think we need it now, but I wonder if a `strict` mode can be enforced with a running host in the scenario where I'm managing my own host, and running 3rd party actors. In that case, I may want to deny a live update to an actor that decides to ask for more permissions. Thoughts?

crates/wasmcloud-host/src/capability/native_host.rs
```diff
@@ -58,12 +58,12 @@ impl Actor for NativeCapabilityHost {
             //warn!("Stopped a provider host that had no state. Something might be amiss, askew, or perchance awry");
             return;
         }
-        let mut state = self.state.as_mut().unwrap();
+        let state = self.state.as_mut().unwrap();
```
brooksmtownsend, line 61: Does this implicitly make a mutable reference, or does the `.as_mut()` not make a difference here

crates/wasmcloud-host/src/host.rs
```diff
@@ -42,6 +43,14 @@ impl HostBuilder {
...
             rpc_client: None,
             cplane_client: None,
+            allow_live_update: false,
+        }
+    }
+
+    pub fn enable_live_updates(self) -> HostBuilder {
+        HostBuilder {
+            allow_live_update: true,
+            ..self
```
brooksmtownsend, line 53: Something like this was what I was referring to in my previous comment about not allowing updates with increased claims.

crates/wasmcloud-host/src/actors/actor_host.rs
brooksmtownsend: Also, I suppose, this statement doesn't check the contents of the claims, so an actor with `keyvalue` and `http_client` could ditch those, add in a `fs` capability and read files without triggering this warning. Could a `.iter().map()` be used in combination with https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.eq to achieve strict comparison?

tests/control.rs
```diff
@@ -229,3 +229,17 @@ fn webrequirements() -> HashMap<String, String> {
     hm.insert("web-friendly".to_string(), "yes".to_string());
     hm
 }
+
+fn embed_revision(source: &[u8], kp: &KeyPair, rev: i32, subject: &str, issuer: &str) -> Vec<u8> {
+    let claims = wascap::jwt::Claims::<wascap::jwt::Actor>::new(
+        "Testy McTestFace".to_string(),
```
brooksmtownsend, line 235: nice

brooksmtownsend (changes requested):

crates/wasmcloud-host/src/actors/actor_host.rs
autodidaddict: Right now I'm just sort of putting the warn in the logs so that someone can maybe check it out. We can always add issues in the future to strengthen the types of rules we have around live updates. We do also disable them by default

crates/wasmcloud-host/src/capability/native_host.rs
autodidaddict: yeah, it's an &mut

crates/wasmcloud-host/src/actors/actor_host.rs
autodidaddict: I like the idea of having an issue to look into a "strict mode" for live update support

brooksmtownsend (approved):

Issue #36, opened 2020-12-11
Support for `call` and `link` actions in control interface


brooksmtownsend:
Currently there are two wasmCloud actions that are present in the Host API, `call` for invoking an operation on an actor, and `link` which connects an actor to a capability provider running in a lattice. These actions are absent from the control interface which necessitates a local host object to invoke them, even if the entities (actor / provider) are not running on that host.

With `call` and `link` available in the control interface, we would be able to fully interact with entities in a lattice without the use of an additional host. Specifically for `wash`, the `wash ctl` command could then support links and calls, and the `wash up` command would be able to run without launching a host for the REPL (if one exists in the lattice already)

PR #37, opened 2020-12-11
Adding support for RPC invocations over control interface

autodidaddict:
First of two features to be added: remote actor invocation via access to the RPC NATS subject.

It's subtle but worth noting that up to this point, the extras provider has been subscribing to the RPC subject whenever there was an RPC connection available. Since we check local-first subscriptions this never actually happened, but I've removed the NATS subscription now so that we don't waste the effort with the RPC proxy on a provider we know is always present.


autodidaddict:
This PR, when complete, will close #36

crates/control-interface/src/generated/ctliface.rs
```diff
@@ -205,3 +206,30 @@ pub struct ProviderDescription {
     #[serde(rename = "image_ref")]
     pub image_ref: Option<String>,
 }
+
+/// The standard function for serializing codec structs into a format that can be
+/// used for message exchange between actor and host. Use of any other function to
```
brooksmtownsend, line 211: I appreciate this comment 😄 Good to distinguish here the proper use of serialize

crates/control-interface/src/inv.rs
```diff
@@ -0,0 +1,134 @@
+// Philosophical note: These two main structs (Invocation and InvocationResponse) exist inside
+// the wasmcloud-host crate. If this crate were to refer to that, we'd create a cyclical reference
+// and all perish inside a black hole.
+//
+// The other alternative is to create a new, third crate shared by the two, but that creates a brand
+// new "reason to change" now for three crates and produces a "dependency as a liability"
+//
+// Therefore, it's easier to simply create copies of the data types since we only need them to be
+// able to push them on the wire.
+
```
brooksmtownsend, line 10: Maybe it would be useful to include a RUSTDOC annotation here with the file that this is a copy of for extra doc info? Something like
```
[Invocation](../wasmcloud-host/struct.Invocation)
```
I'm not sure if that's the correct syntax or not, but I think this would at least provide a reference and help us keep track of the copy

brooksmtownsend (approved):
LGTM, a minor suggestion that may be unnecessary

crates/control-interface/src/generated/ctliface.rs
autodidaddict: This comment actually comes "out of the box" when you use the wapc code generator ;)

crates/control-interface/src/inv.rs
autodidaddict: It's hard to create an extra-crate link when the crate you're pointing to hasn't been published yet. I'll add some context, though.

PR #38, opened 2020-12-11
Make inv mod public for control_interface imports

brooksmtownsend:
Currently, consumers of the `control_interface` crate will be unable to use the `InvocationResponse` type as the struct is hidden behind a private module. If the `InvocationResponse` type is imported from `wasmcloud-host`, Rust interprets them as two different structs with the same name. This PR just adds the `pub` keyword to the `inv` module which makes those types public

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -1,6 +1,6 @@
 pub mod broker;
 mod generated;
-mod inv;
+pub mod inv;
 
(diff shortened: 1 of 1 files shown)


autodidaddict (approved):
You can either make the inv module public, or you can leave that module private and then do `pub use inv::{Invocation, InvocationResponse}` which would make those two types available at the root of the control interface crate without exposing that `inv` module.

PR #42, opened 2020-12-14
Integrated structopt into wasmcloud main

### LINKED ISSUE #30: Add structopt commands to wasmcloud binary
The default wasmcloud binary should have commands that determine whether the RPC or control interface components are enabled and, if so, what their NATS URLs are and credentials. Additionally, there should be a flag that indicates whether live updates are enabled.

* RPC NATS configuration
* Control Interface nats configuration
* Toggle for allowing live updates
* Toggle for allowing latest for OCI pulls

brooksmtownsend:
Fixes #30


autodidaddict (changes requested):
Thanks for diving into this... I was planning on getting here eventually.

A couple of things that I was also planning on doing here:
1. Add `control_host` and `control_port` to allow for the control interface to use a different NATS connection
1. Add `control_jwt` and `control_seed` (optionals). If they're supplied, use this for login instead of anonymous
1. Add `rpc_jwt` and `rpc_seed` (optionals). If they're supplied, use this for login instead of anonymous.

We can default all of the connections to anonymous/localhost but they should be different connections.

autodidaddict:
:thinking: we should probably allow people to specify a `control_credsfile` and a `rpc_credsfile` so that the seed values do not need to be passed on the command-line and you can instead point to the location of a `.creds` file

brooksmtownsend:
> Thanks for diving into this... I was planning on getting here eventually.
> 
> 
> 
> A couple of things that I was also planning on doing here:
> 
> 1. Add `control_host` and `control_port` to allow for the control interface to use a different NATS connection
> 
> 1. Add `control_jwt` and `control_seed` (optionals). If they're supplied, use this for login instead of anonymous
> 
> 1. Add `rpc_jwt` and `rpc_seed` (optionals). If they're supplied, use this for login instead of anonymous.
> 
> 
> 
> We can default all of the connections to anonymous/localhost but they should be different connections.

These all sound good, we can definitely take a similar approach to wash where there are a few ways to supply these sensitive values, including completely hidden from bash history / without reliance on files

brooksmtownsend:
If #48 is merged before this PR, the `disable-strict-update-check` should also be implemented as a structopt flag

autodidaddict (approved):

brooksmtownsend:
@autodidaddict Rebased with main to include the `disable_strict_updates_check` flag as well, for https://github.com/wasmCloud/wasmCloud/pull/48. Any other thoughts before I merge?

autodidaddict:
LGTM. We'll go over all of this a few more times prior to the 0.15.0 release

PR #45, opened 2020-12-15
add parameter 'allow_insecure' to 'fetch_oci_bytes'

tiptop96:


crates/wasmcloud-host/src/control_interface/handlers.rs
```diff
@@ -48,7 +48,7 @@ pub(crate) async fn handle_update_actor(host: &str, msg: &nats::asynk::Message)
             if let Some(a) = a {
                 ack.accepted = true;
                 let _ = msg.respond(&serialize(ack).unwrap()).await;
-                let bytes = fetch_oci_bytes(&req.new_actor_ref, false).await;
+                let bytes = fetch_oci_bytes(&req.new_actor_ref, false, false).await;
```
brooksmtownsend, line 51: In the control interface handlers, I think it makes sense to source the `allow-insecure` value here from the `ControlOptions` struct, that way when we initialize a control interface we can denote if we want to allow insecure registries rather than only deny

crates/wasmcloud-host/src/control_interface/handlers.rs
```diff
@@ -277,7 +277,7 @@ pub(crate) async fn handle_start_actor(host: &str, msg: &nats::asynk::Message, a
         }
     }
 
-    let bytes = crate::oci::fetch_oci_bytes(&cmd.actor_ref, allow_latest).await;
+    let bytes = crate::oci::fetch_oci_bytes(&cmd.actor_ref, allow_latest, false).await;
```
brooksmtownsend, line 280: Same comment as above, you can see how the `allow_latest` value here is from the `ControlOptions` struct

crates/wasmcloud-host/src/host.rs
```diff
@@ -224,7 +224,7 @@ impl Host {
         link_name: Option<String>,
     ) -> Result<()> {
         let hc = HostController::from_hostlocal_registry(&self.id.borrow());
-        let bytes = fetch_oci_bytes(cap_ref, self.allow_latest).await?;
+        let bytes = fetch_oci_bytes(cap_ref, self.allow_latest, false).await?;
```
brooksmtownsend, line 227: Similar to above, the `allow_insecure` could be included on the Host struct (you will also want to include a method in the HostBuilder struct to allow insecure connections. If you put the option on the Host, then each host can configure if they allow insecure registries.

crates/wasmcloud-host/src/host.rs
```diff
@@ -248,7 +248,7 @@ impl Host {
 
     pub async fn start_actor_from_registry(&self, actor_ref: &str) -> Result<()> {
         let hc = HostController::from_hostlocal_registry(&self.id.borrow());
-        let bytes = fetch_oci_bytes(actor_ref, self.allow_latest).await?;
+        let bytes = fetch_oci_bytes(actor_ref, self.allow_latest, false).await?;
```
brooksmtownsend, line 251: Same as above comment

crates/wasmcloud-host/src/manifest.rs
```diff
@@ -96,7 +96,7 @@ pub(crate) async fn generate_actor_start_messages(
             }
         } else {
             // load actor from OCI
-            if let Ok(a) = fetch_oci_bytes(&actor_ref, allow_latest)
+            if let Ok(a) = fetch_oci_bytes(&actor_ref, allow_latest, false)
```
brooksmtownsend, line 99: If you include a parameter for `allow_insecure`, the place where this function (`generate_actor_start_messages`) is called is from a context where the Host is `self`, so you'd have access to the Host `allow_insecure` option mentioned in the above comment

crates/wasmcloud-host/src/manifest.rs
```diff
@@ -130,7 +130,7 @@ pub(crate) async fn generate_provider_start_messages(
             }
         } else {
             // read PAR from OCI
-            if let Ok(prov) = fetch_oci_bytes(&cap.image_ref, allow_latest)
+            if let Ok(prov) = fetch_oci_bytes(&cap.image_ref, allow_latest, false)
```
brooksmtownsend, line 133: See above

crates/wasmcloud-host/src/oci.rs
```diff
@@ -71,7 +75,7 @@ pub(crate) async fn fetch_provider_archive(
     img: &str,
     allow_latest: bool,
 ) -> Result<ProviderArchive> {
-    let bytes = fetch_oci_bytes(img, allow_latest).await?;
+    let bytes = fetch_oci_bytes(img, allow_latest, false).await?;
```
brooksmtownsend, line 78: You should be able to make this also a parameter of this function, you'll need to follow the call stack a few layers but it leads to `impl Handler<NatsMessage> for ControlInterface`

brooksmtownsend (changes requested):
Thanks for getting this issue fixed and PR in so quick! This will be very useful for local development, I have a few requests to allow us to configure the `allow_latest` value from the context of a Host and the context of a ControlInterface, but they should be simple fixes.

brooksmtownsend (approved):

brooksmtownsend:
Looks great @tiptop96 !😄 @autodidaddict any comments?

autodidaddict (approved):

PR #66, opened 2021-01-20
Allow specified list of HTTP connections for OCI interaction

### LINKED ISSUE #63: fetch_oci_bytes forces only HTTP or HTTPS connections
When a host is instantiated, you have the option of providing the `oci_allow_insecure` option to allow for insecure connections to OCI registries (e.g. over http, like the `localhost:5000` docker registry included in the `tools/` directory). However, the logic in `fetch_oci_bytes` chooses the connection protocol based on that option, so if insecure connections are allowed then they will be forced.

This is an issue because when connecting to secure registries, fetching the `wasm` or `par` can fail if the protocol does not match. In the event where insecure connections are allowed and you attempt to connect to a secure registry (e.g. `wasmcloud.azurecr.io`), the following error is shown:

```
15:05:56:ERROR:wasmcloud_host::oci:/Users/btownsend/.cargo/git/checkouts/wasmcloud-2f24a1426d406474/97c66bb/crates/wasmcloud-host/src/oci.rs:52:Failed to fetch OCI bytes: resgistry did not return a digest header
15:05:56:ERROR:wasmcloud_host::control_interface::handlers:/Users/btownsend/.cargo/git/checkouts/wasmcloud-2f24a1426d406474/97c66bb/crates/wasmcloud-host/src/control_interface/handlers.rs:292:Failed to retrieve actor image from OCI registry: Failed to fetch OCI bytes
```

We either need additional information here for if an actor/provider is coming from an insecure registry, or we need some other way to determine the correct protocol to use.

brooksmtownsend:
Fixes #63

My solution here ended up allowing for the user instantiating the host to provide a list of registries that are allowed for insecure (HTTP) connections. This option is valid regardless of authentication scheme. This way, instantiating a host with a host builder now can include the option:
```
host_builder = host_builder.oci_allow_insecure(vec!["localhost:5000"]);
```
to allow for insecure connections to a local registry, but then to use HTTPS for all other connections. The intended use of this is to let users define specific registries for testing / development purposes that they can connect via HTTP.

### DIFF
--- a/crates/wasmcloud-host/src/control_interface/ctlactor.rs
+++ b/crates/wasmcloud-host/src/control_interface/ctlactor.rs
@@ -26,7 +26,7 @@ pub struct Initialize {
 #[derive(Clone, Debug, Default)]
 pub struct ControlOptions {
     pub oci_allow_latest: bool,
-    pub oci_allow_insecure: bool,
+    pub oci_allowed_insecure: Vec<String>,
     pub host_labels: HashMap<String, String>,
     pub max_actors: u16,    // Currently unused
     pub max_providers: u16, // Currently unused
@@ -96,7 +96,7 @@ impl Handler<NatsMessage> for ControlInterface {
         let msg = msg.msg;
         let subject = msg.subject.to_string();
         let allow_latest = self.options.oci_allow_latest;
-        let allow_insecure = self.options.oci_allow_insecure;
+        let allowed_insecure = self.options.oci_allowed_insecure.clone();
         let nc = self.client.clone();
         Box::pin(
             async move {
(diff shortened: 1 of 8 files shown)


brooksmtownsend:
I've tested this with a local docker registry and pulling artifacts from `wascc.azurecr.io` and `wasmcloud.azurecr.io` and they work well in tandem. I was successfully able to schedule actors and providers from both registries

autodidaddict:
So, to recap for the slow to grasp (me)... if you supply no parameters to the host builder, the host will be unable to make _any_ insecure connection to any OCI registry.

If you specify a vec, then any of the hosts (including port) in that list can be used w/out SSL.

The net result is that I can run a single wasmCloud host that is capable of scheduling (and downloading) from any number of secure registries, and from an allow-listed vector of insecure registries?

brooksmtownsend:
> So, to recap for the slow to grasp (me)... if you supply no parameters to the host builder, the host will be unable to make _any_ insecure connection to any OCI registry.
> 
> If you specify a vec, then any of the hosts (including port) in that list can be used w/out SSL.
> 
> The net result is that I can run a single wasmCloud host that is capable of scheduling (and downloading) from any number of secure registries, and from an allow-listed vector of insecure registries?

Correct, with no parameters to the host builder all OCI connections will be carried out over HTTPS. The port is only necessary if you are not communicating over standard HTTP/HTTPS ports (e.g. you can add `wasmcloud.azurecr.io` without the port).

And your last point you are correct. Allow-list of insecure registries keeps us secure by default while also allowing development environments to support quick-and-easy local registries. (The `wash` REPL host will include `localhost:5000` in the insecure registry list, that way you can use the provided registry in `tools/docker-compose.yml` without any modification)

autodidaddict (approved):

Issue #65, opened 2021-01-20
Acknowledge commands early for control interface


autodidaddict:
In the current implementation the control interface waits to acknowledge the receipt of the "start actor" and "start provider" commands until after the start has completed. In situations with high-latency networks or where the download of an OCI image may take a while after not having been cached, this can present consumers and CLI clients (e.g. `wash`) with long delays that can exceed timeout periods, giving consumers a false sense of failure.

If clients want deterministic notification of when actors and providers start, they should listen to the appropriate events on the control interface's `events` stream.

autodidaddict:
This is required for 0.15.0.

brooksmtownsend:
Related to #64

PR #67, opened 2021-01-21
Control interface start actor and provider now acknowledge prior to downloading OCI bytes

autodidaddict:
The new behavior of the wasmCloud control interface is that the `StartActor` and `StartProvider` commands will now be acknowledged as early as possible, either if a validation failure is detected or immediately _before_ the OCI bytes are downloaded.

This has a downstream effect of requiring that any consumer of the control interface must monitor the control event stream in order to determine if an actor or provider did successfully complete the startup process.

### DIFF
--- a/crates/control-interface/src/generated/ctliface.rs
+++ b/crates/control-interface/src/generated/ctliface.rs
@@ -55,8 +55,6 @@ pub struct StartActorAck {
     pub host_id: String,
     #[serde(rename = "actor_ref")]
     pub actor_ref: String,
-    #[serde(rename = "actor_id")]
-    pub actor_id: String,
     #[serde(rename = "failure")]
     pub failure: Option<String>,
 }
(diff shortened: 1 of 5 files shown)


autodidaddict:
This addresses #65

brooksmtownsend (approved):
LGTM. I think monitoring of the control event stream will be a great piece of docs to have so developers can monitor these events, but the big thing is that now we won't be shown false errors due to network conditions. Looking forward to brainstorming on the monitoring piece

autodidaddict:
I also think we can make it easy by adding an event monitor function to the control interface client.

brooksmtownsend:
@autodidaddict Also worth mentioning I've run this with our providers & actors, and even with no cache I don't see a timeout with a value of 5 seconds. Even with 2 seconds I don't see it, so for demos everything will be a lot snappier.

PR #69, opened 2021-01-25
Normalizing Native Capability link names

### LINKED ISSUE #68: Links can be initialized as empty strings
A link, which is a way for us to provide logical separation within a lattice with resources running in the same lattice, can currently be defined as an empty string (""). Links can be thought of as namespaces as well, so an empty string can cause confusion. We currently have a default link name ("default") that we use when one is not supplied, but in order to capture the edge case where a blank string is supplied both `None` values (no link name supplied) and an empty string should result in the link name "default".

brooksmtownsend:
Fixes #68

`link_name` is littered across the project (162 results in 16 files) so I've attempted to do this in the most efficient way possible. I tackled two main areas of issue when creating links across wasmCloud:
1. Starting a capability provider with a link name of `""`
  Tackled with `normalize_link_name` within the `capability` module. Feasibly, we could surround every occurrence of `link_name` with this function, but I see less value in that if we stop the irregular naming at the source. (If this is desired, `normalize_link_name` could be moved to a common `util` module, and I can do that refactor brute-force 😄 
2. Starting a native capability host (e.g. accessing a capability from within an actor) with a link name of `""`
  This functionality was actually already present, just at a later point in the callstack. I moved that logic up so that an actor invoking a provider can't make the mistake of invoking a provider on the link `""`

### DIFF
--- a/crates/wasmcloud-host/src/dispatch.rs
+++ b/crates/wasmcloud-host/src/dispatch.rs
@@ -318,6 +318,12 @@ pub(crate) fn wapc_host_callback(
         operation
     );
 
+    let link_name = if link_name.trim().is_empty() {
+        // Some actor SDKs may not specify a link field by default
+        "default"
+    } else {
+        link_name
+    };
     // Look up the public key of the provider bound to the origin actor
     // for the given capability contract ID.
     let bus = MessageBus::from_hostlocal_registry(&kp.public_key());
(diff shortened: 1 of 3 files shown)


crates/wasmcloud-host/src/dispatch.rs
```diff
@@ -362,17 +368,11 @@ fn invocation_from_callback(
...
-        "default".to_string()
-    } else {
-        bd.to_string()
-    };
     let target = if ns.len() == 56 && ns.starts_with("M") {
         WasmCloudEntity::Actor(ns.to_string())
     } else {
         WasmCloudEntity::Capability {
-            link_name,
+            link_name: bd.to_string(),
```
autodidaddict, line 375: Shouldn't this be `link_name` rather than the unmodified `bd` ?

autodidaddict (changes requested):

crates/wasmcloud-host/src/dispatch.rs
brooksmtownsend: That change was because the logic that created the `link_name` variable was moved up to `wapc_host_callback`, there's no `link_name` in that function anymore

brooksmtownsend:
Any additional comments here @autodidaddict ?

autodidaddict:
> Any additional comments here @autodidaddict ?

Not as long as the tests are all good.

autodidaddict (approved):

PR #76, opened 2021-02-03
Update packaging

billyoung:
This refactors how we manage pushing to many distributions.

Packagecloud uses "distribution version id's" which means in order to support different versions of ubuntu, debian, fedora, etc, we need to push a package with the specific id.

Using the matrix for deb and rpm distribution id's, we can simplify this action considerably. If or when packagecloud adds support for a new distribution, we can simply add the ID to the matrix.


brooksmtownsend (approved):
Nice use of matrix action, LGTM

PR #81, opened 2021-02-11
Remove git dependencies

### LINKED ISSUE #58: Test modules are outdated
Need to replace the parJEEZy files with the up-to-date versions. Case in point - there's one test that will pass when it runs against the main branch of the http server provider but it fails when run against the .par.gz file sitting in `tests/modules`.

brooksmtownsend:
Fixes #58

This PR
- Removes all `git` dependencies, allowing for `control-interface`, `wasmcloud-host` and `wasmcloud` to be published on crates.io
- Updates `wascap`, `provider-archive`, and `oci-distribution` to their latest published versions
- Updates test modules to conform to new `wasmcloud:*` contract IDs, and updated test artifacts with their latest versions
- Includes README badges because I love them

On my local all tests pass except for `distributed_echo`, which has failed for me for a while now. That warrants some more investigation unless that's a local problem for macos (possibly related to firewall rules?)


autodidaddict (changes requested):
It all looks good. Can you go through and fix all the clippy warnings? (btw, I love that the pull requests show clippy warnings)

brooksmtownsend:
@autodidaddict I addressed all clippy warnings except for unused functions, and I didn't want to litter `#[allow(unused)]` everywhere in the project. If you see any functions that you think are fine to go I can remove them, otherwise I think we can either mark them unused or leave them be for the time being

autodidaddict:
We can leave them be for now and then the next time we're in the codebase for something other than a release, we can cut out all the unused stuff.

brooksmtownsend:
@autodidaddict is this good to merge then?

autodidaddict:
SHIP IT :rocket:

PR #88, opened 2021-02-16
Adding a benchmark-friendly actor

autodidaddict:
Benchmark-friendly actor that can be used to test round-trip times for direct calls or through the HTTP provider, depending on what we want to test.

**NOTE** - I could not get the Rust official `bench` macro to work anywhere in our test framework, so right now all I've got committed here is the actor on which we can run benchmarks, not the actual benches.


samples/bench-actor/README.md
```diff
@@ -0,0 +1,21 @@
+# New Actor Template
```
brooksmtownsend, line 1: We should probably update this README to at least just have a different heading, and then a small description about this being a benchmark actor

brooksmtownsend (changes requested):
Largely looks good just calling for removing the "new actor" README template

samples/bench-actor/README.md
autodidaddict: Sigh. I had this readme updated locally and just hadn't pushed it.

autodidaddict:
Resolved merge conflict from main.

brooksmtownsend (changes requested):
I'm confused, did something change about the tests to make them fail? I think we should re-run the job, but this looks good to me when the tests are passing.

autodidaddict:
```
test kvcounter_basic ... Got host
thread 'main' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.2.0/src/runtime/blocking/pool.rs:85:33
```

It's another timing issue. :(

Cargo.toml
```diff
@@ -42,16 +42,16 @@ wasmcloud-host = { path = "crates/wasmcloud-host" }
...
-rmp-serde = "0.14.4"
+rmp-serde = "0.15.4"
 serde = {version = "1.0.117", features = ["derive"] }
 serde_bytes = "0.11.5"
 serde_json = "1.0.59"
 provider-archive = "0.4.0"
-redis = "0.17.0"
-reqwest = { version = "0.10"}
+redis = "0.19.0"
+reqwest = { version = "0.11"}
```
brooksmtownsend, line 54: This `reqwest` update is causing the tests to fail, potentially due to a mismatch in tokio runtime versions? If you revert to `0.10` the tests pass locally on my machine, so I'd bet that will help it here.

brooksmtownsend (changes requested):

brooksmtownsend (approved):

PR #100, opened 2021-02-19
Initial commit of code that invokes the benchmark actor

autodidaddict:
Make sure you run the `bencher` binary/project in release mode, otherwise you get the super-slow parsing/interpreting behavior on the wasm files.

### DIFF
--- a/samples/bencher/src/main.rs
+++ b/samples/bencher/src/main.rs
@@ -10,7 +10,15 @@ use wasmcloud_host::{Actor, HostBuilder};
 async fn main() -> Result<(), Box<dyn Error + Sync + Send>> {
     let h = HostBuilder::new().build();
     h.start().await?;
-    let echo = Actor::from_file("../../target/wasm32-unknown-unknown/release/bench_actor_s.wasm")?;
+    let echo = match Actor::from_file(
+        "../../target/wasm32-unknown-unknown/release/bench_actor_s.wasm",
+    ) {
+        Ok(e) => e,
+        Err(e) => {
+            println!("Unable to locate bench_actor_s.wasm. Please run 'make release' in samples/bench-actor/");
+            return Err(e);
+        }
+    };
     let actor_id = echo.public_key();
     h.start_actor(echo).await?;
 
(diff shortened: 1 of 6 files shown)


samples/bencher/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+use std::time::Instant;
+use wasmcloud_actor_core::serialize;
+use wasmcloud_actor_http_server as http;
+use wasmcloud_host::{Actor, HostBuilder};
+
+#[actix_rt::main]
+async fn main() -> Result<(), Box<dyn Error + Sync + Send>> {
+    let h = HostBuilder::new().build();
+    h.start().await?;
+    let echo = Actor::from_file("../../target/wasm32-unknown-unknown/release/bench_actor_s.wasm")?;
```
brooksmtownsend, line 13: Since this depends on running `make release` in the bench-actor directory, I think this could be better off as a match statement so we can provide some context to the failure in case someone just runs this sample. It will error with a "not found", but then the user will need to check out the code to figure out why. I think, in case of error, a message like:
```
"Error loading bench actor, please run `make release` in samples/bench-actor/"
```
will add some clarity.

samples/bencher/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+        body: serialize(&data)?, // this is msgpack encoding inside an HTTP body.
+    })?;
+
+    let now = Instant::now();
+    for _i in 0..10_000 {
+        h.call_actor(&actor_id, "HandleRequest", &payload).await?;
+    }
+    let done = now.elapsed().as_secs();
+    println!(
+        "10,000 iterations took {}s, {}/iteration",
```
brooksmtownsend, line 36: My `release` report:
```
10,000 iterations took 4s, 0.0004/iteration
```
nice.

brooksmtownsend (changes requested):
One change requested, works great

samples/bencher/src/main.rs
autodidaddict: 2500 messages per second ;)

brooksmtownsend (approved):

samples/bencher/src/main.rs
brooksmtownsend: That's, what the professionals call "fast"

PR #104, opened 2021-02-26
Stopactor

autodidaddict:
Adding this test as a regression detector. Also curious if it fails on GH actions even though it "works on my machine".


brooksmtownsend (approved):
LGTM

Issue #103, opened 2021-02-25
[BUG] wasmcloud does not seem to properly stop actors


jsoverson:
**Describe the bug**

Wasmcloud seems to keep actors hanging around even after stopping them through wash

**To Reproduce**

With wasmcloud running somewhere:

```bash
# Convenient alias
$ alias get-inventory="wash ctl get hosts -o json | jq -r '.hosts[].id' | xargs -n 1 wash ctl get inventory -o json | jq -s 'map({host:.inventory.host_id, actors: (.inventory.actors)})'"
# Show no actors running on hosts
$ get-inventory
# start arbitrary actor
$ wash ctl start actor wasmcloud.azurecr.io/echo:0.2.0
# show actor is running
$ get-inventory
# get actor id to reference later
$ ACTOR_ID=$(wash claims inspect -o json wasmcloud.azurecr.io/echo:0.2.0 | jq -r '.module')
# make a call to the actor 
$ wash ctl call $ACTOR_ID HandleRequest '{"method":"GET", "body":"", "queryString":"", "path":"/test", "header":{}}'
# stop actor across all hosts
$ wash ctl get hosts -o json | jq -r '.hosts[].id' | xargs -n 1 -I{} wash ctl stop actor {} $ACTOR_ID
# show that the actor is reported as stopped across all hosts
$ get-inventory
# Actor should be gone, yet the following call succeeds:
$ wash ctl call $ACTOR_ID HandleRequest '{"method":"GET", "body":"", "queryString":"", "path":"/should_be_dead", "header":{}}'
# Final inventory to show that the actor hasn't been automatically started again by wash.
$ get-inventory
```

**Example output**
```bash
jsoverson@pop-os: ~/development/src/scratch
$ alias get-inventory="wash ctl get hosts -o json | jq -r '.hosts[].id' | xargs -n 1 wash ctl get inventory -o json | jq -s 'map({host:.inventory.host_id, actors: (.inventory.actors)})'"
 
jsoverson@pop-os: ~/development/src/scratch
$ get-inventory                                                                                                                              [14:32:08]
[
  {
    "host": "NDCNHJL32ZD6MNK5Y4WBJ7BWZMX54B5LAJQOCU3AU6T5U6TTQJOLZMHC",
    "actors": []
  }
]
 
jsoverson@pop-os: ~/development/src/scratch
$ wash ctl start actor wasmcloud.azurecr.io/echo:0.2.0                                                                                       [14:32:13]
[...]

PR #105, opened 2021-02-26
Closing #103 explicitly terminates actor and provider hosts

autodidaddict:
.. and the RPC proxies after a message bus unsubscribe.


brooksmtownsend (approved):
LGTM. I saw some clippy warnings, but I think the rename from wascc_codec -> provider_core can be in a separate PR

PR #107, opened 2021-03-01
First commit for allowing invocation by alias.

autodidaddict:
Call alias support:

1. When an actor is started, if its claims contain the `call_alias` field (non-None), that alias is then published into the lattice cache as a key-value pair (`call_{alias}` : `{public_key}`).
2. When you invoke `call_actor` on a host now, if the reference string you pass is not a 56-character `M`-prefixed key, it will attempt to look that string up in the lattice cache as a call alias. If that passes, we invoke by public key, otherwise the call fails.
3. Inside the wapc callback (the function used by actors to invoke against the outside world), we now attempt to resolve the target of the outbound call as follows:
    a. if it's a 56-character `M`-prefixed key, call that actor
    b. if that reference turns up in the call alias cache, call the public key associated with that alias
    c. If that reference is a provider's contract ID that can be found in the actor links cache, call that provider
    d. fail with "no target" otherwise.

To test this, we'll want to create the following actors:
1. `PingTest` , which accepts an incoming HTTP request and in response, calls the `Ping` operation on the actor aliases with `wasmcloud/samples/ping`
2. `Ping`, an actor that is signed with the alias `wasmcloud/samples/ping` and responds to the `Ping` operation by returning the set of bytes `Pong`

Then we can create a host that loads both those actors and the HTTP provider and curl the endpoint and we will get "Pong". This will eventually be part of the integration suite, so this PR shouldn't be merged until these two actors exist in the OCI registry.


tests/no_lattice.rs
```diff
@@ -107,6 +107,47 @@ pub async fn kvcounter_basic() -> Result<()> {
     Ok(())
 }
 
+pub async fn actor_to_actor_call_alias() -> Result<()> {
+    ::std::env::remove_var("KVCACHE_NATS_URL");
+    let h = HostBuilder::new().with_namespace("actor2actor").build();
+    h.start().await?;
+    let pinger = Actor::from_file("./tests/modules/pinger.wasm")?;
+    let ponger = Actor::from_file("./tests/modules/ponger.wasm")?;
```
brooksmtownsend, line 115: Subtle but cool, this host has no NATS but the actors can still call each other. Awesome.

brooksmtownsend (approved):

brooksmtownsend (changes requested):
Unsure why it only shows me a few files occasionally... This LGTM once the inline clippy warnings are fixed 👍🏻

tests/no_lattice.rs
autodidaddict: actors and capability providers can all talk to each other locally within a host. In fact, they even do that as a prioritized call when NATS _is_ running (e.g. if the target of an invocation is both locally inside a host and outside on the lattice, the local one is invoked).

autodidaddict:
Which clippy warnings are there? I don't see them.

brooksmtownsend:
@autodidaddict I see them inline in the PR, small things like redundant clones and using a str in place of a char. If you're running clippy locally you may need:
```
cargo clippy --all-features
```

brooksmtownsend (approved):

PR #109, opened 2021-03-04
Extras provider update

brooksmtownsend:
Updated JWT of embedded extras provider to `wasmcloud:extras` contract ID. I also updated the unit test and included a unit test to ensure that we know if these operations are working well in the future.

The only part of this PR that is not ready is the `extras.wasm` file, which is signed with my local keys. I'll PR in the extras example actor, and once we get that release with our GH action I can put it in this PR.


brooksmtownsend:
The example PR in question: https://github.com/wasmCloud/examples/pull/26

autodidaddict (approved):

brooksmtownsend:
@autodidaddict I just updated the extras actor to be the one from `wasmcloud.azurecr.io/extras:0.2.0`

autodidaddict (approved):

PR #115, opened 2021-03-14
Fixed manifest cache behavior for local par files, added log output

jsoverson:
Manifests support referencing providers as local files, but wasmcloud-host defers to the cache at `$TEMP/wasmcloudcache` for all cases, even if a local provider has changed on disk. This behavior makes sense for immutable sources, but par files on disk don't make that guarantee. This leads to very difficult-to-troubleshoot behavior when the expectation is that the referenced file is the source of truth, not a previously loaded version.

This PR changes cache behavior when loading a manifest that references local provider archives. It does not affect cache behavior for pars from a registry.

This PR also adds logging for previously silent errors and places that would have made troubleshooting this easier.

### DIFF
--- a/crates/wasmcloud-host/src/manifest.rs
+++ b/crates/wasmcloud-host/src/manifest.rs
@@ -105,6 +105,8 @@ pub(crate) async fn generate_actor_start_messages(
                     image_ref: Some(actor_ref.to_string()),
                     actor: a,
                 });
+            } else {
+                error!("Actor {} not found on disk or in registry", actor_ref);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


crates/wasmcloud-host/src/manifest.rs
```diff
@@ -125,6 +127,9 @@ pub(crate) async fn generate_provider_start_messages(
                 .and_then(|bytes| ProviderArchive::try_load(&bytes))
                 .and_then(|par| NativeCapability::from_archive(&par, cap.link_name.clone()))
             {
+                if let Err(e) = crate::capability::native_host::write_provider_to_disk(&prov) {
+                    error!("Could not cache provider to disk: {}", e);
+                }
```
brooksmtownsend, line 132: Just to confirm, this is re-writing the par that was sourced from disk to the cache location, and this is done because otherwise we would just be loading the par from the cache even if it had changed?

brooksmtownsend (commented):
Thanks for this PR! I just had one question, otherwise it LGTM

crates/wasmcloud-host/src/manifest.rs
jsoverson: Yes, that's it.

jsoverson: Oops, no, I read that backwards. The original version loads the cache even if the par has changed, this version updates the cache with the par contents if the par is defined in manifest as a local file.

brooksmtownsend: I may have written that backwards 😅, we're on the same page!

PR #126, opened 2021-03-19
add --label to wasmcloud

StuartHarris:
This PR adds `--label` (`-l`) to the `wasmcloud` CLI in order to pass labels to the instantiated `wasmcloud-host`. The flag can be passed multiple times, for example:

```
wasmcloud --label x=1 --label y=2 -l z=3
```

I know that labels can be set via a manifest, but I think this is more convenient when using the CLI and it prevents overwriting all the labels at once.

### DIFF
--- a/src/wasmcloud.rs
+++ b/src/wasmcloud.rs
@@ -75,6 +75,7 @@ struct Cli {
     manifest: Option<PathBuf>,
 
     /// attach a label to the host - can be used multiple times
+    // `number_of_values = 1` means that each occurrence of the flag accepts only one key-value pair
     #[structopt(long = "label", short = "l", parse(try_from_str = parse_key_val), number_of_values = 1)]
     labels: Vec<(String, String)>,
 }
@@ -204,8 +205,8 @@ mod test {
     #[test]
     fn parse_key_val_ok() {
         let expected = ("x".to_string(), "1".to_string());
-        let actual = parse_key_val::<String, String>("x=1").unwrap();
-        assert!(actual == expected);
+        let actual: (String, String) = parse_key_val("x=1").unwrap();
+        assert_eq!(actual, expected);
     }
 
     #[test]
--- a/src/wasmcloud.rs
+++ b/src/wasmcloud.rs
@@ -1,5 +1,4 @@
-use std::io::prelude::*;
-use std::{fs::File, path::PathBuf};
+use std::{error::Error, fs::File, io::prelude::*, path::PathBuf, str};
 use structopt::{clap::AppSettings, StructOpt};
 use wasmcloud_host::{HostBuilder, HostManifest, Result};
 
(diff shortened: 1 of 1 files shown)


src/wasmcloud.rs
```diff
@@ -175,3 +182,35 @@ fn extract_arg_value(arg: &str) -> Result<String> {
...
+        let expected = ("x".to_string(), "1".to_string());
+        let actual = parse_key_val::<String, String>("x=1").unwrap();
+        assert!(actual == expected);
+    }
+
+    #[test]
+    fn parse_key_val_err() {
+        let actual = parse_key_val::<String, String>("bad string");
+        assert!(actual.is_err());
+    }
```
brooksmtownsend, line 215: Thank you for adding tests! This is awesome 😄

src/wasmcloud.rs
```diff
@@ -74,6 +73,10 @@ struct Cli {
     /// Specifies a manifest file to apply to the host once started
     #[structopt(long = "manifest", short = "m", parse(from_os_str))]
     manifest: Option<PathBuf>,
+
+    /// attach a label to the host - can be used multiple times
+    #[structopt(long = "label", short = "l", parse(try_from_str = parse_key_val), number_of_values = 1)]
```
brooksmtownsend, line 78: If multiple values can be specified, what does `number_of_values` do here? Does it just indicate that I shouldn't be providing labels in the form of:
```
--label foo=bar,foobar=barfoo
```
?
I have tested it and it does work when I specify multiple labels as an argument (e.g. `-l this=that -l  that=this`) but I was just curious.

brooksmtownsend (commented):
Just one thing I'm curious about, otherwise LGTM 👍🏻

src/wasmcloud.rs
StuartHarris: Yep that's it. It restricts each occurrence of the flag to just one pair. Without it, you could do `wasmcloud --label x=1 y=2` but that would prevent you from following it with any other arguments.

StuartHarris: (although a comma would fix that, I guess). Either way, I think it's more intuitive to use the flag once for each label pair.

src/wasmcloud.rs
StuartHarris: pleasure :-) I'm just going to tweak that first one, to make it a bit more idiomatic.

src/wasmcloud.rs
StuartHarris: added a comment in the code to clarify

brooksmtownsend: Definitely, one flag at a time is what I would expect.

src/wasmcloud.rs
StuartHarris: just removed the turbofish, and changed the assert. Reads better, and a failure would be better reported

brooksmtownsend (approved):
As long as tests are good, LGTM

autodidaddict (approved):

PR #132, opened 2021-03-22
Make Host Sendable

autodidaddict:
Fixes an issue where the Host struct was not Send because of leftover and unnecessary use of Refcells


thomastaylor312 (approved):
Looks great! Also got rid of some verbosity along the way

crates/wasmcloud-host/src/host.rs
```diff
@@ -517,3 +511,16 @@ impl Host {
...
+mod test {
+    use crate::HostBuilder;
+
+    #[test]
+    fn is_send() {
+        let h = HostBuilder::new().build();
+        assert_is_send(h);
+    }
+
+    fn assert_is_send<T: Send>(_input: T) {}
```
brooksmtownsend, line 525: Never thought about this as a test, neat

brooksmtownsend (approved):

PR #138, opened 2021-03-29
wasmcloud binary crates.io release action

brooksmtownsend:
While I was in here editing actions, I also made the sub-crates depend on successful build to release.


billyoung (approved):
looks good! :rocket:

Issue #15, opened 2020-11-16
Host allows activities without being started


autodidaddict:
You can start actors and perform other activities without first invoking `start()` on the host. In a number of scenarios the host will actually still "work", but the more complicated your host the less this will work. For example, if you invoke without calling `start()` then your actors won't be able to make network RPC calls (no lattice provider configured), the custom authorizer won't be configured, etc.

We should ensure that the  hosts rejects the appropriate calls when the host is not in a started state.

PR #142, opened 2021-03-29
Closing #15

autodidaddict:
Requires an active actix rt system and a previously successful call to start before allowing operations on a live host.

### DIFF
--- a/crates/wasmcloud-host/src/host.rs
+++ b/crates/wasmcloud-host/src/host.rs
@@ -20,8 +20,11 @@ use crate::oci::fetch_oci_bytes;
 use crate::{ControlEvent, HostManifest, NativeCapability, WasmCloudEntity};
 use crate::{Result, SYSTEM_ACTOR};
 use provider_archive::ProviderArchive;
-use std::collections::HashMap;
 use std::time::Duration;
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
 use wascap::prelude::KeyPair;
 
 /// A host builder provides a convenient, fluid syntax for setting initial configuration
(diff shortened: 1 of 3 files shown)


crates/wasmcloud-host/src/host.rs
```diff
@@ -506,6 +526,16 @@ impl Host {
         Ok(())
     }
 
+    fn ensure_started(&self) -> Result<()> {
+        if *self.started.read().unwrap() == false {
```
brooksmtownsend, line 530: per-clippy warning, this can be refactored to look like:
```
if !*self.started.read().unwrap() {
```

crates/wasmcloud-host/src/host.rs
```diff
@@ -215,6 +219,7 @@ pub struct Host {
     allow_live_updates: bool,
     lattice_cache_provider_ref: Option<String>,
     strict_update_check: bool,
+    started: Arc<RwLock<bool>>,
```
brooksmtownsend, line 222: Is there a reason why this `started` variable is an `Arc<RwLock<bool>>` rather than just a `bool`? I see the benefit of ensuring multiple resources don't modify it at the same time, but it's really only touched when ensuring it has started and appears that a simple `bool` would suffice.

brooksmtownsend (changes requested):
Clippy warning and optimization

crates/wasmcloud-host/src/host.rs
autodidaddict: because the host, by rule, doesn't have any `mut` methods. Mutation has to be kept inside. The host also has requirements by consumers that it be placeable inside refcells, which means it can't do anything internally that isn't threadsafe. So, hiding the mutation of the started field inside an rwlock is the only way to go. Though, we can actually remove the `Arc`.

brooksmtownsend (approved):
happy clippy happy me

PR #144, opened 2021-03-30
Re-advertises claims upon live update

autodidaddict:


autodidaddict:
Tests pass now that this repo no longer has a hard dev dependency on the http server provider.

brooksmtownsend (approved):

Issue #59, opened 2021-01-14
Capability providers segfault on initialization for Raspberry Pi


brooksmtownsend:
Raspberry Pi 4 Model B Rev 1.2
```
OS: Raspbian GNU/Linux 10 (buster) armv7l
Arch: 32bit
Kernel: 5.4.72-v7l+
```

Currently experiencing Segmentation faults when scheduling capability providers on wasmCloud hosts running on raspberry pi's. This happens both on my RPI4 and my RPI3, they both have the same OS / Architecture pair. 

To reproduce:
1. Install `wasmcloud` via cargo, or by cloning the repository and building manually
2. Launch NATS on the pi
3. run wasmcloud with the arguments `RUST_LOG=info KVCACHE_NATS_URL=0.0.0.0:4222 wasmcloud`
4. Attempt to schedule a provider with `wash ctl start provider wasmcloud.azurecr.io/redis:0.9.2` (`redis`, `nats` and `httpserver` all fail)
5. segfault

These same parJEEzy's that are contained in the OCI registry work from my testing on the `x86_64-macos`, `x86-64-linux` and `aarch64-linux`, and each of the targets are built in the same fashion, using `cross` in our GH Actions CI/CD.

Additionally, I've attempted to build the `libwascc_redis.so` and form a parJEEzy on the PI to eliminate any architecture mismatches, and that also causes a segfault in the same initialization code.

This is the exact line of code that the program experiences a segfault: https://github.com/wasmCloud/wasmCloud/blob/main/crates/wasmcloud-host/src/capability/native_host.rs#L248. (just in case it changes later on, I'll paste the snippet here:
```
let plugin = unsafe {
    let constructor: Symbol<PluginCreate> = library.get(b"__capability_provider_create")?;
    let boxed_raw = constructor(); // RPI Segfault

    Box::from_raw(boxed_raw)
};
```

Looking for more information here, perhaps we need to do additional testing with the `constructor()` call for specific capability providers.

autodidaddict:
Since every single provider fails during the call to `constructor()`, this feels like it might be build related or dispatch/invoke related, and not likely being caused by something inside the constructor itself. This is also supported by the fact that nearly all of the providers have virtually empty constructors (they do the bulk of their initialization _per actor_ during a `OP_BIND_ACTOR` handler).

autodidaddict:
[...]

Issue #60, opened 2021-01-15
Loading providers from cache on Windows expects .dll extension


brooksmtownsend:
When we download a provider archive from an OCI registry with wasmCloud, we store the provider archive in `$TMP/wasmcloud_ocicache` and the specific shared library in `$TMP/wasmcloudcache` under the structure `PUBLICKEY/REVISION/target`. for example, `VDTVCIL4V3WHSPR3BWCP6E4GFD6GKZ7OE5NDW3RAEUWSOQQR7FPZEYLR/0/x86_64-linux`.

On Linux and macos, this has worked just fine. Recently I've tested this out in a Windows 10 environment, and Windows expects the library to include the `.dll` extension, and will not load only the `x86_64-windows` library. Since the pattern for the majority of other operating systems is to omit the extension, for Windows targets we should append the `.dll` either when loading or writing the provider.

brooksmtownsend:
So progress here, turns out it was expecting the DLL extension because we were using the OS agnostic libloading library. If we use the windows specific one (check here https://docs.rs/libloading/0.7.0/libloading/struct.Library.html#platform-specific-behaviour) then the following statement applies:
```
If the filename specifies a library filename without path and with extension omitted, .dll extension is implicitly added on Windows.
```
I'm working on a PR to get this in, but just wanted to update.

brooksmtownsend:
Actually, re-reading this, it appears that the `.dll` extension is implicitly ADDED, so we must append the `.dll` for windows. We don't have control to slice off the `.dll` extension so we have to add it at the `.par.gz` unpack step

brooksmtownsend:
Closing this for now, as we've taken a stance to not support windows for our current system. Our official recommendation for those on windows machines will be to use WSL as a workaround, at least until we have full native capability provider support.

NOTE: We **DO** support Windows as an OS when `NativeCapability`'s are instantiated as static plugins, as this doesn't involve loading shared libraries from disk. If there is a desire from a user to run wasmcloud on Windows without WSL we will just need to advocate for creating their own custom Rust host. You can verify that this is the case by running the `distributed_echo` test in a Powershell environment, as the httpserver provider is instantiated as a static plugin instead of being loaded from disk. (You can also run the `link_on_third_host` test to verify that loading from a provider archive segfaults. You may need to adjust the NATS connection from `0.0.0.0:4222` to `127.0.0.1:4222` to run these tests.

PR #145, opened 2021-03-30
Windows debugging

brooksmtownsend:
This PR was created in an attempt to further debug the Windows issues that we are running into for wasmcloud.

TL;DR: use WSL

Re: #60, the libloading library automatically appends `.dll` which I confirmed in their documentation, so we have to extract the provider with the `.dll` extension or windows machines will not be able to work.

Refactored the libloading section to have a linux, windows, and "other" section for loading capability providers. I was able to discover that the `.dll` section is not required so long as you use the `libloading::os::windows` module for the library loading portion. I was not able to resolve https://github.com/wasmCloud/wasmCloud/issues/59, and windows still can segfault upon initialization of the capability provider. I was able to confirm with `dumpbin` that the downloaded httpserver provider (I'm running the echo example) does export `__capability_provider_create`, but regardless the code panics with a `LoadLibraryExW` error. From what I can read, this is due to missing shared libraries elsewhere on my system, but as far as I know I have a well functioning Windows development environment. 

I've used `depends` to try and track down the shared module that is failing, however my `depends` lists hundreds of missing `.dll`s.

I've used ProcessMonitor to track syscalls for the wasmcloud program, but I can't find anything out of the ordinary other than a failure to `CreateFileMapping`. I've attempted to open the library with the specified protections, but unfortunately this doesn't fix the issue. 

Here's my current error:
```
[2021-03-30T18:14:31Z DEBUG] GetProcAddress { source: Os { code: 126, kind: Other, message: "The specified module could not be found." } }
[2021-03-30T18:14:31Z ERROR] Failed to extract plugin from provider: GetProcAddress failed
```

Interested if anyone else on windows can reproduce, if you have `nats-server` and I stole the manifest for the program from https://github.com/wasmCloud/examples/blob/main/echo/manifest.yaml

Reference:
https://docs.microsoft.com/en-us/windows/win32/memory/memory-protection-constants
https://docs.rs/libloading/0.7.0/libloading/os/windows/struct.Library.html

### DIFF
--- a/crates/wasmcloud-host/src/capability/native_host.rs
+++ b/crates/wasmcloud-host/src/capability/native_host.rs
@@ -250,22 +250,36 @@ fn extrude(
         } else {
             write_provider_to_disk(cap)?;
         }
-        #[cfg(target_os = "linux")]
-        let library: Library = {
+
+        let library: Library = if cfg!(target_os = "linux") {
             // On linux the library must be loaded with `RTLD_NOW | RTLD_NODELETE` to fix a SIGSEGV
             // that happens when a thread terminates AFTER a library is unloaded. This has the
             // effect of keeping the library loaded in-memory until the process ends.
             //
             // https://github.com/nagisa/rust_libloading/issues/41
-            unsafe { ::libloading::os::unix::Library::open(Some(&path), 0x2 | 0x1000)? }.into()
+            #[cfg(target_os = "linux")]
+            return unsafe { ::libloading::os::unix::Library::open(Some(&path), 0x2 | 0x1000)? }.into();
+            #[cfg(not(target_os = "linux"))]
+            unsafe { Library::new(&path)? }
+        } else if cfg!(target_os = "windows") {
+            unsafe { ::libloading::os::windows::Library::load_with_flags(&path, 0x1000)? }.into()
+        } else {
+            unsafe { Library::new(&path)? }
         };
-        #[cfg(not(target_os = "linux"))]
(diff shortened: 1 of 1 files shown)


autodidaddict:
w/r/t missing shared files - have you installed a windows equivalent of the C runtime? On linux we need that for libloading to work.

brooksmtownsend:
@autodidaddict Unfortunately I've spent hours trying to make sure my runtime environment is correct, and I'm like 90% sure that my configuration is the correct one. Originally I started on `x86_64-pc-windows-gnu`, only to find out that the recommended way is to install visual studio (not code) and get all of the build tools through there. Which I was able to do, and I even use the whole cross tools command prompt, but its still possible I don't have the proper toolchain installed. If there's any easy resource to check that I'm happy to try.

autodidaddict:
Thanks for the diligence on this. I think you're right in that, at least for now, we recommend people use WSL when working on windows.

brooksmtownsend:
> Thanks for the diligence on this. I think you're right in that, at least for now, we recommend people use WSL when working on windows.

This does narrow our scope, but as you mentioned (potentially in slack) this only really affects users that intend to use windows in production as WSL is a small change for a development machine. We'll just have to take this stance in our documentation somewhere as far as OS/platforms that we officially support

brooksmtownsend:
Since https://github.com/wasmCloud/wasmCloud/issues/60 is closed, closing this PR. Picking up any related docs in https://github.com/wasmCloud/wasmcloud-dev-site/issues/28

PR #148, opened 2021-04-07
Added logging, allowed manifests support missing fields

jsoverson:
This is mostly a logging PR to help diagnose issues and understand the flow of wasmcloud via trace logging. Some logs don't look valuable, but without them it's confusing to know if anything is happening when you poke at the host.

The large chunk of code changed in CreateSubscription is just logging. There was a lot going on there that failed silently on errors.

Piggybacking onto this PR is support for empty links, actors, and capabilities in manifest files. LMK if I should change any of the message text or back anything out.

- added logging messages
- propagated ser/des error messages  
- implemented Display for WasmCloudEntity
- allowing empty links, capabilities, and actors in manifest

### DIFF
--- a/crates/wasmcloud-control-interface/src/lib.rs
+++ b/crates/wasmcloud-control-interface/src/lib.rs
@@ -323,7 +323,7 @@ pub fn serialize<T>(
 where
     T: Serialize,
 {
-    serde_json::to_vec(&item).map_err(|_e| "JSON serialization failure".into())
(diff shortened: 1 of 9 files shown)


autodidaddict (approved):
This is all great stuff. Some of the work you've done here will overlap with some of the things we'll get for free when we adopt the `tracing` crate, but that's no reason to put off having diagnostic detail now. Thanks!

brooksmtownsend (approved):
This is awesome, these `trace` logs will be invaluable for debugging exactly where something goes wrong.

Issue #131, opened 2021-03-22
[BUG] Rust documentation is outdated and inaccurate


autodidaddict:
**Describe the bug**
There are a number of places in the rust documentation for the host API where the information given is incorrect or outdated. There are even places where it still references wascc.

**Environment**
- wasmcloud version [e.g. `0.15.3`]

brooksmtownsend:
Adding to this, all of the functions on the wasmcloud-host `Host` struct should be documented extensively, including the meaning of all parameters and return types.

For example, with `set_link`, we ask for a parameter called `actor`. It's unclear if `actor` should be the actor reference (OCI) or the ID (public key)

autodidaddict:
Yeah, excellent call out. Every public function that appears in docs.rs needs to be evaluated for accurate and _complete_ documentation. We should also try and include example invocations of every function as well, using all of the rustdoc goodies.

PR #151, opened 2021-04-12
Closing #131. Documentation updates.

autodidaddict:
CC @stevelr

### DIFF
--- a/crates/wasmcloud-host/src/host.rs
+++ b/crates/wasmcloud-host/src/host.rs
@@ -159,8 +159,8 @@ impl HostBuilder {
 
     /// Allows the host to pull actor and capability provider images from the supplied list
     /// of valid OCI registries without using a secure (SSL/TLS) connection. This option is empty by default and we recommend
-    /// it not be used in production environments. For local testing, supplying `["localhost:5000"]`
-    /// as an argument for the local OCI registry will allow for http connections to that registry.
+    /// it not be used in production environments. For local testing, make sure you supply the right
+    /// host name and port number for the locally running OCI registry.
     pub fn oci_allow_insecure(self, allowed_insecure: Vec<String>) -> HostBuilder {
         HostBuilder {
             allowed_insecure,
@@ -337,6 +337,10 @@ impl Host {
     /// images are assumed to be immutable this kind of caching is acceptable). If you need to
     /// purge the OCI image cache, remove the `wasmcloudcache` and `wasmcloud_ocicache` directories
     /// from the appropriate root folder (which defaults to the environment's `TEMP` directory).
+    ///
+    /// # Arguments
+    /// * `cap_ref` - The OCI reference URL of the capability provider.
+    /// * `link_name` - The link name to be used for this capability provider.
     pub async fn start_capability_from_registry(
         &self,
         cap_ref: &str,
@@ -389,6 +393,8 @@ impl Host {
     /// Stops a running actor in the host. This call is assumed to be idempotent and as such will
     /// not fail if you attempt to stop an actor that is not running (though this may result
     /// in errors or warnings in log output).
+    ///
+    /// * `actor_ref` - Either the public key or the OCI reference URL of the actor to stop.
     pub async fn stop_actor(&self, actor_ref: &str) -> Result<()> {
         self.ensure_started()?;
         let hc = HostController::from_hostlocal_registry(&self.id);
@@ -403,15 +409,21 @@ impl Host {
     /// Updates a running actor with the bytes from a new actor module. Invoking this method will
     /// cause all pending messages inbound to the actor to block while the update is performed. It can take a few
     /// seconds depending on the size of the actor and the underlying engine you're using (e.g. JIT vs. interpreter).
-    /// ## ⚠️ Caveats
+    ///
+    /// # Arguments
+    /// * `actor_id` - The public key of the actor to update.
+    /// * `new_oci_ref` - If applicable, a new OCI reference URL that corresponds to the new version.
+    /// * `bytes` - The raw bytes containing the new WebAssembly module.
+    ///
+    /// # ⚠️ Caveats
     /// * Take care when supplying a value for the `new_oci_ref` parameter. You should be consistent
     /// in how you supply this value. Updating actors that were started from OCI references should
     /// continue to have OCI references, while those started from non-OCI sources should not be given
     /// new, arbitrary OCI references. Failing to keep this consistent could cause unforeseen failed
-    /// attempts at subsequent updates.
-    /// * You should only ever use this method when running in standalone/isolated mode. If you have a control
-    /// interface configured, then you should perform live updates by using a control interface client
-    /// and a suitable OCI reference URL.
+    /// attempts at subsequent updates.  
+    /// * If the new version of the actor is actually in an OCI registry, then the preferred method of
+    /// performing a live update is to do so through the lattice control interface and specifying the OCI
+    /// URL. This method is less error-prone and can cause less confusion over time.        
     pub async fn update_actor(
         &self,
         actor_id: &str,
@@ -442,6 +454,11 @@ impl Host {
     /// Stops a running capability provider. This call will not fail if the indicated provider
     /// is not currently running. This call is non-blocking and does not wait for the provider
     /// to finish cleaning up its resources before returning.
+    ///
+    /// # Arguments
+    /// * `provider_ref` - The capability provider's public key or OCI reference URL
+    /// * `contract_id` - The contract ID of the provider to stop
+    /// * `link` - The link name used by the instance of the capability provider
     pub async fn stop_provider(
         &self,
         provider_ref: &str,
@@ -460,7 +477,7 @@ impl Host {
         Ok(())
     }
 
-    /// Retrieves the list of all actors within this host. This function call does _not_
+    /// Retrieves the list of the public keys of all actors within this host. This function call does _not_
     /// include any actors remotely running in a connected lattice.
     pub async fn get_actors(&self) -> Result<Vec<String>> {
         self.ensure_started()?;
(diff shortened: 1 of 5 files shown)


crates/wasmcloud-host/src/host.rs
```diff
@@ -95,7 +95,8 @@ impl HostBuilder {
 
     /// Enables remote control of the host via the lattice control protocol by providing an instance
     /// of a NATS connection through which control commands flow. Remote control of hosts is
-    /// disabled by default.
+    /// disabled by default. For security reasons, you should not use the same NATS connection or
+    /// security context for both RPC and lattice control.
```
brooksmtownsend, line 99: I'm interested in this, why not? What I can think of immediately is that for some reason if your connection to the RPC was compromised, you can shut it down without affecting the lattice. Is there more to the story there?

crates/wasmcloud-host/src/host.rs
```diff
@@ -147,18 +149,18 @@ impl HostBuilder {
...
     }
 
-    /// Allows the host to pull actor and capability provider images from these registries without
-    /// using a secure (SSL/TLS) connection. This option is empty by default and we recommend
-    /// it not be used in production environments. For local testing, supplying ["localhost:5000"]
-    /// as an argument for the local docker reigstry will allow for http connections to that registry.
+    /// Allows the host to pull actor and capability provider images from the supplied list
+    /// of valid OCI registries without using a secure (SSL/TLS) connection. This option is empty by default and we recommend
+    /// it not be used in production environments. For local testing, supplying `["localhost:5000"]`
+    /// as an argument for the local OCI registry will allow for http connections to that registry.
```
brooksmtownsend, line 163: The `localhost:5000` is how we have our `docker-compose` setup, but perhaps others won't know that context. What do you think about rephrasing the last sentence as something like:
```
For local testing, supplying `["localhost:5000"]` as an argument will allow for HTTP connections to a local OCI registry listening on port 5000.
```
Definitely will entertain some wording updates, but this may be a bit more explicit for the newbie that has their registry on port 420 without knowing

crates/wasmcloud-host/src/host.rs
```diff
@@ -201,7 +204,7 @@ impl HostBuilder {
     }
 }
 
-/// A wasmCloud `Host` is a secure runtime container responsible for scheduling
+/// A wasmCloud host is a secure runtime responsible for scheduling
```
brooksmtownsend, line 207: YEAH NO CONTAINERS

crates/wasmcloud-host/src/host.rs
```diff
@@ -445,14 +470,14 @@ impl Host {
...
-    /// lattice
+    /// lattice.
     pub async fn get_providers(&self) -> Result<Vec<String>> {
         self.ensure_started()?;
         let b = MessageBus::from_hostlocal_registry(&self.id);
         Ok(b.send(QueryProviders {}).await?.results)
     }
 
-    /// Perform a raw waPC-style invocation on the given actor by supplying an operation
+    /// Perform a raw [waPC](https://wapc.io)-style invocation on the given actor by supplying an operation
```
brooksmtownsend, line 480: I'm a little concerned with the `https://wapc.io` link here because the docs on that site are all placeholders, I was going to suggest https://crates.io/crates/wapc instead but I wouldn't want to give off the impression we only support Rust. Thoughts?

crates/wasmcloud-host/src/host.rs
```diff
@@ -376,8 +401,8 @@ impl Host {
     }
 
     /// Updates a running actor with the bytes from a new actor module. Invoking this method will
-    /// cause all messages inbound to the actor to block and it can take a second or two depending
-    /// on the size of the actor and the underlying engine you're using (e.g. JIT vs. interpreter).
+    /// cause all pending messages inbound to the actor to block while the update is performed. It can take a few
+    /// seconds depending on the size of the actor and the underlying engine you're using (e.g. JIT vs. interpreter).
     /// ## ⚠️ Caveats
     /// * Take care when supplying a value for the `new_oci_ref` parameter. You should be consistent
     /// in how you supply this value. Updating actors that were started from OCI references should
```
brooksmtownsend, line 408: Unfortunately I can't comment on it as it wasn't changed in this PR, but for the snippet a few lines down:
```
    /// * You should only ever use this method when running in standalone/isolated mode. If you have a control
    /// interface configured, then you should perform live updates by using a control interface client
    /// and a suitable OCI reference URL.
```
What are the consequences of this?

crates/wasmcloud-host/src/host.rs
```diff
@@ -487,13 +512,16 @@ impl Host {
...
+    /// is uniquely keyed by the provider's contract ID, public key, and link name.
+    ///
+    /// You can set a link before or after either the actor or provider are started. Links are automatically established
     /// when both parties are present in a lattice, and re-established if a party temporarily
-    /// leaves the lattice and rejoins (which can happen during a crash or a partition event). Link
-    /// data is exactly as durable as your choice of lattice cache provider
+    /// leaves the lattice and rejoins (which can happen during a crash or a network partition event). The resiliency and
+    /// reliability of link definitions within a lattice are inherited from your choice of lattice cache provider.
+    ///
+    /// The value of `actor` for a link definition can either be the actor's public key _or_ an OCI reference URL.
```
brooksmtownsend, line 524: For this, you can specifically annotate the argument like so:
```
/// Links links blah blah resilient lattice
/// 
/// * `actor` - Actor public key _or_ OCI reference URL
fn set_link (actor: &str) {}
```
Unfortunately it won't show live until we republish interfaces, but here's what it looks like with the logging interface:
image

crates/wasmcloud-host/src/host.rs
```diff
@@ -517,7 +545,11 @@ impl Host {
     /// Removes a link definition from the host (and accompanying lattice if connected). This
     /// will remove the link definition from the lattice cache and it will also invoke the "remove actor"
     /// operation on all affected capability providers, giving those providers an opportunity to clean
-    /// up resources and terminate any child processes associated with the actor
+    /// up resources and terminate any child processes associated with the actor. This call does not
+    /// wait for all providers to finish cleaning up associated resources.
+    ///
+    /// To remove a link definition, the value of `actor` must be the actor's public key. You _cannot_ remove
+    /// a link definition with an actor's OCI reference URL.
```
brooksmtownsend, line 552: Same comment as above for the `actor` argument

crates/wasmcloud-host/src/host.rs
brooksmtownsend: This kind of annotation may also help out for any other arguments that are named `provider_ref`, `actor_id`, etc

crates/wasmcloud-host/src/lib.rs
```diff
@@ -1,29 +1,63 @@
...
-//! then loads an actor that simply echoes back incoming HTTP requests as outbound HTTP responses.
-//! The HTTP server capability provider is loaded so that the actor can receive web requests.
-//! Note that the link definition (configuration of the link between the actor and the
-//! capability provider) can be defined _in any order_. The host runtime automatically
-//! establishes links as soon as all related parties are up and running inside a host or
-//! a lattice.
+//! The HTTP server capability provider is loaded so that the actor can receive web requests. A [link definition](https://wasmcloud.dev/reference/host-runtime/links/) is required
+//! between the HTTP server capability provider and the actor in order to verify actor privileges and supply configuration values
+//! (such as the port on which to listen). This link definition can be established _before or after_ the actor and capability
+//! provider have been started, as link definitions are first-class data cached throughout a lattice.
```
brooksmtownsend, line 60: Thoughts on linking the `lattice` keyword here to https://wasmcloud.dev/reference/lattice/?

brooksmtownsend (changes requested):
A couple of requested changes and questions. Aside from that, let me tell you, the documentation on this looks **INCREDIBLE** and I think these things are going to pay massive dividends when people are looking up wasmcloud.

crates/wasmcloud-host/src/host.rs
autodidaddict: That's basically it. Separate the connections and it minimizes the blast radius due to compromise. Someone could compromise _either_ the RPC connection _or_ the lattice control connection, but not both.

crates/wasmcloud-host/src/host.rs
autodidaddict: :fire:

crates/wasmcloud-host/src/host.rs
autodidaddict: going to remove this and just use the opportunity to give people a heads-up to make sure they supply the *correct* host name and port.

crates/wasmcloud-host/src/host.rs
autodidaddict: I like the idea of pointing to the crate. They'll make their way to the wapc documentation one way or another, but at least we control the URLs in the crate and the wapc URL (or documentation) could change without us knowing.

crates/wasmcloud-host/src/host.rs
autodidaddict: `links links blah blah resilient lattice` feels like our new marketing tagline.

brooksmtownsend: `r/wasmcloud_circlejerk` incoming

crates/wasmcloud-host/src/lib.rs
autodidaddict: I like it.

autodidaddict:
Thanks for the tip on using the bullet lists for the arguments. I like how it looks much better than just using a paragraph.

brooksmtownsend (approved):

brooksmtownsend:
Looks like a merge may have gone wrong 😬

autodidaddict:
> Looks like a merge may have gone wrong grimacing

Yep. Not sure what happened, other than me being foolish enough to try and fix the merge using the github editor rather than doing it old school.

autodidaddict:
@brooksmtownsend merge problem is all fixed.

brooksmtownsend (approved):
LG TM

Issue #56, opened 2021-01-06
Add features to wasmCloud Cargo.toml


brooksmtownsend:
The `wasmcloud-host` crate includes feature flags for selecting underlying wasm engines. The top level `wasmcloud` crate (e.g. the Cargo.toml in the root of this repository) should also include the feature flags there so users can select their wasm engine directly with the `wasmcloud` binary.

ChrisRx:
I just ran into this and wanted to leave this as something to keep in mind for addressing this issue later:

[Cargo feature rules](https://doc.rust-lang.org/cargo/reference/features.html#rules) make all features opt-in except for the default feature, which must be opted out of for the feature to not be included.  So specifying the feature `wasmtime` for `wasmcloud-host` when building will still included the `wasm3-provider` dependency unless something like `--no-default-features` is specified.  For crates like `wasmcloud` which specify `wasmcloud-host` as a dependency, this cargo flag will not have an effect on disabling the default features of the `wasmcloud-host`.  This is not a problem but just something to keep in mind going forward with this issue.

PR #152, opened 2021-04-13
Implemented feature flags for wasmcloud binary

brooksmtownsend:
This PR allows you to run the wasmcloud binary with either the `wasm3` or `wasmtime` engine. The default is wasm3, and in order to run with wasmtime you must use the following command:
```
cargo run --features wasmtime --no-default-features
```
In order to enforce a default, as @ChrisRx noted in #56, we have to explicitly remove the default feature in order to use the other engine type. I've adjusted the default log verbosity for the `wasmtime-provider` as in `info` mode you will get a lot of output as wasmtime JITs the actor module, in `debug` mode you get an unusable amount of wasm debug output.

This PR also uncomments our integration tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,11 @@ exclude = [
 ]
 autotests = false
 
+[features]
+default = ["wasm3"]
+wasmtime = ["wasmcloud-host/wasmtime"]
+wasm3 = ["wasmcloud-host/wasm3"]
+
 [[test]]
 name = "integration"
 path = "tests/lib.rs"
@@ -41,7 +46,7 @@ log = "0.4.11"
 env_logger = "0.8.2"
 anyhow = "1.0.34"
 actix-rt = "2.1.0"
-wasmcloud-host = { path = "crates/wasmcloud-host", version = "0.16.1" }
+wasmcloud-host = { path = "crates/wasmcloud-host", version = "0.16.1", default-features = false }
 nats = "0.8.6"
 structopt = "0.3.21"
 nkeys = "0.1.0"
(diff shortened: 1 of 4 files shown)


billyoung:
@brooksmtownsend are we happy with the default `wasm3` being packaged into our deb/rpm packages? If a customer wants to use `wasmtime` I assume they would have to compile and package themselves?

brooksmtownsend:
@billyoung Yeah that's correct, everything will be packaged as wasm3 (which is what we were doing before). We could certainly just adjust the default, or even just the release action to release wasmtime, but I don't think it's necessary to publish both engines as a binary (at least for now when both engines are essentially the same). ADR time for default wasm engine for published binary? @autodidaddict

brooksmtownsend:
We seem to go back and forth on this, and while wasm3 seems to be better for the low cold start penalty it does end up slower than wasmtime in execution in long running wasmcloud hosts. Perhaps now is the time to make our released wasmcloud binary use wasmtime instead.

autodidaddict:
"slower" than wasmtime is a very subjective number. wasm3 per-function-call latency via wapc is a matter of micros.

autodidaddict:
if we're going to change how we release then we essentially add the engine to the matrix, so that we build for OS+CPU+engine and release `wasmcloud-wasm3` and `wasmcloud-wasmtime`, etc. Or we continue just releasing the default wasm3 which is the better general purpose engine and let those who want wasmtime choose it as a feature-flag when they cargo install.

autodidaddict (dismissed):

brooksmtownsend:
Apparently by merging the wasmcloud-host version it dismissed your review @autodidaddict

autodidaddict (approved):

PR #160, opened 2021-04-14
Ensure live update of actor puts new OCI reference

### LINKED ISSUE #156: [BUG] Live updating an actor does not update its OCI reference in the lattice cache
**Describe the bug**
Live updating an actor does not update its OCI reference in the lattice cache

**To Reproduce**
1. Start an actor
2. Perform live update on the actor
3. Observe that the list of OCI references does _not_ contain that actor's new reference

**Expected behavior**
We should see in the host inventory all of the OCI references. Additionally, when retrieving the actor from a host inventory query over lattice control interface, it should return an array of all the OCI references that belong to that actor. Ideally we'd like to sort the OCI references by semver, but there is no requirement that actors published in an OCI registry conform to semver. So, rather than "hoping" that there is a viable sort order for the OCI images, we should simply return them all. This means changing the oci reference field for actors and providers in the host inventory query response from `Option<String>` to `Vec<String>` which can be empty.

brooksmtownsend:
Fixes #156 

A few notes:
1. My suggestion to skip serializing the field if the Vec is empty fails to deserialize (exactly as the documentation says) so I removed it
2. I added a test to ensure we don't regress in the future, which tests both `ctl` and host API

### DIFF
--- a/tests/control.rs
+++ b/tests/control.rs
@@ -180,6 +180,73 @@ pub(crate) async fn live_update() -> Result<()> {
     Ok(())
 }
 
+/// Ensures a live update of an actor places the new OCI reference in the lattice cache
+pub(crate) async fn multiple_ocirefs() -> Result<()> {
+    // Ensure that we're not accidentally using the replication feature on KV cache
+    ::std::env::remove_var("KVCACHE_NATS_URL");
+    const NS: &str = "liveupdate_ctl";
+    const ECHO_0_2_0: &str = "wasmcloud.azurecr.io/echo:0.2.0";
+    const ECHO_0_2_1: &str = "wasmcloud.azurecr.io/echo:0.2.1";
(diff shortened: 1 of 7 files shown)


tests/control.rs
```diff
@@ -180,6 +180,80 @@ pub(crate) async fn live_update() -> Result<()> {
...
+    ctl_client
+        .start_actor(&hid, "wasmcloud.azurecr.io/echo:0.2.0")
+        .await?;
+
+    for _ in 0..MAX_RETRY {
+        let inv = ctl_client.get_host_inventory(&hosts[0].id).await?;
+        if !inv.actors.is_empty() && inv.actors[0].image_refs.contains(&ECHO_0_2_0.to_string()) {
+            break;
+        }
+        actix_rt::time::sleep(Duration::from_millis(1000)).await;
```
autodidaddict, line 221: :clock1:

autodidaddict (approved):
LGTM!

tests/control.rs
brooksmtownsend: I probably could set that to less because I'm doing a few iterations but I just needed some time 😄

PR #161, opened 2021-04-15
Adding CLI option for namespace prefix

autodidaddict:


brooksmtownsend (approved):
There's the old saying, give an engineer 10 lines of code to review and he'll find 10 problems, give an engineer 100 lines of code to review and he'll say looks good to me.

But this looks good 😄

PR #163, opened 2021-04-15
Associate actor/provider with its image_ref, HostAPI parity

brooksmtownsend:
This PR brings the HostAPI functionality up-to-par with the control interface's `HostInventory` operation. A consumer of the Host API or the control interface can now retrieve the list of running actors and providers including their name, revision, OCI image reference (if applicable), ID, and link_name (for providers).

As we haven't released `0.18.0` yet, I did not bump cargo versions. These changes were tested with `wash` for the control interface and Host API along with the integration tests

### DIFF
--- a/crates/wasmcloud-control-interface/src/generated/ctliface.rs
+++ b/crates/wasmcloud-control-interface/src/generated/ctliface.rs
@@ -188,8 +188,9 @@ pub struct HostInventory {
 pub struct ActorDescription {
     #[serde(rename = "id")]
     pub id: String,
-    #[serde(rename = "image_refs")]
-    pub image_refs: Vec<String>,
+    #[serde(rename = "image_ref")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub image_ref: Option<String>,
     #[serde(rename = "name")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
@@ -203,8 +204,9 @@ pub struct ProviderDescription {
     pub id: String,
     #[serde(rename = "link_name")]
     pub link_name: String,
-    #[serde(rename = "image_refs")]
-    pub image_refs: Vec<String>,
+    #[serde(rename = "image_ref")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub image_ref: Option<String>,
     #[serde(rename = "name")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
--- a/crates/wasmcloud-host/src/actors/actor_host.rs
+++ b/crates/wasmcloud-host/src/actors/actor_host.rs
@@ -85,7 +85,8 @@ impl Handler<LiveUpdate> for ActorHost {
         if !self.state.as_ref().unwrap().can_update {
             error!(
                 "Rejecting attempt to update actor ({}) - live updates disabled",
-                msg.image_ref.unwrap_or("No OCI Ref Supplied".into())
+                msg.image_ref
+                    .unwrap_or_else(|| "No OCI Ref Supplied".into())
             );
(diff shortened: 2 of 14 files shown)


brooksmtownsend:
The cargo check failed because of docker!??!?! That's a first 
```
  /usr/bin/docker pull nats
  Using default tag: latest
  Error response from daemon: received unexpected HTTP status: 503 Service Unavailable
  Error: Docker pull failed with exit code 1
```

crates/wasmcloud-control-interface/src/generated/ctliface.rs
```diff
@@ -188,8 +188,9 @@ pub struct HostInventory {
 pub struct ActorDescription {
     #[serde(rename = "id")]
     pub id: String,
-    #[serde(rename = "image_refs")]
-    pub image_refs: Vec<String>,
+    #[serde(rename = "image_ref")]
```
autodidaddict, line 191: :cry: I can't believe you're removing my plural and making it singular again. All that hard work down the drain.

crates/wasmcloud-host/src/actors/actor_host.rs
```diff
@@ -63,6 +63,7 @@ impl Handler<GetIdentity> for ActorHost {
         let state = self.state.as_ref().unwrap();
 
         IdentityResponse {
+            image_ref: state.image_ref.clone(),
```
autodidaddict, line 66: :+1:

crates/wasmcloud-host/src/host.rs
```diff
@@ -521,6 +521,47 @@ impl Host {
             .collect())
     }
 
+    /// Retrieves human-friendly identity information for a provider
+    /// Return value is `(image_reference, name, revision)`
```
autodidaddict, line 525: Should be explicit here that the `id` parameter is the public key.

crates/wasmcloud-host/src/host.rs
```diff
@@ -521,6 +521,47 @@ impl Host {
...
+        } else {
+            Err("Actor is not running on this host".into())
+        }
+    }
+
+    /// Retrieves human-friendly identity information for a provider
+    /// Return value is `(image_reference, name, revision)`
+    pub async fn get_provider_identity(
+        &self,
+        id: &str,
```
autodidaddict, line 546: Explicitly call out in the rustdoc comment that this `id` parameter is the public key of the provider.

tests/control.rs
```diff
@@ -122,9 +122,8 @@ pub(crate) async fn basics() -> Result<()> {
 
     let inv = ctl_client.get_host_inventory(&hosts[0].id).await?;
     println!("Got host inventory: {:?}", inv);
-    //assert_eq!(3, inv.providers.len());
     assert_eq!(1, inv.actors.len());
-    assert_eq!(inv.actors[0].image_refs, vec![KVCOUNTER_OCI.to_string()]);
+    assert_eq!(inv.actors[0].image_ref, Some(KVCOUNTER_OCI.to_string()));
```
autodidaddict, line 126: :sob: all that hard work ... ruined.

autodidaddict (changes requested):
Just a couple tiny recommendations on being more explicit in the public host API docs.

crates/wasmcloud-host/src/host.rs
brooksmtownsend: Good call. I'll rename it to `actor_id` as well to make it immediately clear

crates/wasmcloud-host/src/host.rs
brooksmtownsend: Good call. I'll rename it to `provider_id` as well to make it immediately clear

autodidaddict (approved):

PR #166, opened 2021-04-16
Optimizing Gitpod Deployment

jordan-rash:
Working on getting the gitpod deployment up and running a little quicker.  This PR includes a new Action that will build `.gitpod.Dockerfile` any time that it is updated and push it to azurecr.  Additionally, it changes the gitpod configuration to use that image.  This method eliminated the need to install wasmcloud and wash at start time (as it is included in the container).  

I still feel there is a good deal of optimizing that can be done (by not using that gitpod/workspace-full) as the base image, but for now, this seems a good deal faster to me.  

Note: I was NOT able to test this using Azure as I dont have ant access.  I did test is using GHCR.io and it worked well.

### DIFF
--- a/.github/workflows/gitpod_docker.yml
+++ b/.github/workflows/gitpod_docker.yml
@@ -1,4 +1,4 @@
-name: Package
+name: Build Gitpod Docker Artifact
 on:
   push:
     branches:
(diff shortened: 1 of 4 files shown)


jordan-rash:
All the secrets that I am using in the action, I got from the secrets page in settings.  I can not verify they are right, I just assumed by what they were named.  Need to verify before merged

autodidaddict (approved):

brooksmtownsend (approved):

PR #167, opened 2021-04-18
Bump wasmcloud-host/wasmtime-provider to 0.0.3

matthewtgilbride:
This change should enable installing using the wasmtime feature on Apple M1 hosts.


autodidaddict (approved):

brooksmtownsend (approved):
I've cloned this locally and verified it worked so this isn't an issue, but we should be running tests both with the `wasm3` provider and `wasmtime` provider so we can verify this works in the future 😄 I'll add in a PR for this today

PR #173, opened 2021-04-20
Increased max time to download echo actor from 5s to 30s

brooksmtownsend:
With 5 seconds as a maximum time to wait for `echo` to be scheduled, a computer/runner with constrained resources may not be able to download the ~500k actor and have it available in the host inventory quick enough. I've increased the time to 30 seconds as a maximum to ensure we don't fail to update an actor because it didn't download fast enough.

### DIFF
--- a/tests/control.rs
+++ b/tests/control.rs
@@ -6,6 +6,7 @@ use crate::common::{
 use ::wasmcloud_control_interface::Client;
 use std::collections::HashMap;
 use wasmcloud_actor_http_server::{deserialize, serialize};
+use log::info;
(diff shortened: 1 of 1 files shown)


autodidaddict:
Gotta run `fmt` @brooksmtownsend  :)

autodidaddict (approved):

brooksmtownsend:
@autodidaddict Oh yeah I just wanted to make sure it would pass two times, that was definitely intentional

autodidaddict (approved):

PR #178, opened 2021-04-22
Dockerfile + Automation

jordan-rash:
The action is set to build and push to both dockerhub and azure.  I figured there was no reason not to keep the history in both places.  Also, I think I think i should add an additional action for when the repo is tagged so we have a tag of the tag as well.


billyoung (approved):
looks good!

autodidaddict (approved):

brooksmtownsend (approved):

PR #180, opened 2021-04-22
Checking in wasmcloud binary cargo.lock

### LINKED ISSUE #179: Add Cargo.lock to repo
Per this comment from the generated Cargo `.gitignore` and slack discussion, the repo should have a `Cargo.lock` file.

https://github.com/wasmCloud/wasmCloud/blob/a5fba66c64bdf308b0159c648182cc9736706aae/.gitignore#L6-L8

brooksmtownsend:
Fixes #179


jordan-rash (dismissed):
LGTM

autodidaddict (changes requested):
I think if we blanket ignore `Cargo.lock` then it will leave the cargo lock for all of the `crates/` , which are libraries that we do _not_ want to have checked in cargo.lock files.

Can we explicitly ignore the cargo lock for _just_ the binary and leave it in place for the crates?

brooksmtownsend:
Good call @autodidaddict , this glob pattern will allow `Cargo.lock` at the root, but not in any subdirectories:
```
*/**/Cargo.lock
```
Just changed to that.

autodidaddict (approved):

PR #183, opened 2021-04-27
Lattice control events receiver

### LINKED ISSUE #140: [FEATURE] Control interface event subscription
Currently, the control interface does not have functionality that allows us to subscribe to the lattice event stream. Most of the time we can get away with this, as it's not always necessary with a wasmcloud host to have fine-grained control over the event stream. However, the ability to monitor the event stream allows us to write interactions with a host with a better idea of the state of the lattice.

For example:
1. You start an actor that you have on disk that uses the `telnet` provider
1. You issue a command to a host to start the `telnet` provider.
1. The host must then (assuming it doesn't have that provider cached) download the provider from our OCI registry
1. While this is happening, you may want to write a line of code that ensures that your actor is present, your provider is present, and the link between them has been established before you proceed to make any requests.

Step 4 above is not possible today, but if you are able to monitor for events you would be able to ensure that those resources were appropriately started and linked.

Following the above example again, it would be desirable to either have a `wait_for_event` handler on the control interface or an equivalent method of ensuring that an event occurred.
For example (to wait for an actor, provider, and link worked appropriately):
1. Start resources as described in steps 1-3 above
1. `control_interface::wait_for_event(ActorStarted)`
1. `control_interface::wait_for_event(ProviderStarted)`
1. `control_interface::wait_for_event(LinkEstablished)` (note: I do **not** see this in the current ControlEvent enum)

An alternative solution with a similar gain would be to implement a `wait` option on control interface commands. This, unfortunately, does introduce a blocking call to the command itself, requiring one command to fully complete before the next one may begin.

**Additional context**
This solution should enable an elegant fix for #83 

When approaching this issue, it will be important to consider that if implemented as above the user will need to be careful not to call the `wait_for_event` function after the event has occurred, or you will block forever. Keeping this in mind, it may be appropriate to implement the function instead as `event_exists` to simply check for the existence of a specific event.

cc @autodidaddict

brooksmtownsend:
Fixes #140 

This PR adds the method `events_receiver` for `wasmcloud_control_interface`. This method returns an unbounded crossbeam_channel receiver, which can be used to receive `PublishedEvent`s, containing the control event objects for events like `ActorStarted`, `HostStarted`, `ActorStopped`, etc.

I've also included a doctest to show a possible way to handle events, and an integration test to show it in practice. It's important that you don't block the control interface thread waiting for events, so I've shown an `async` and a `thread::spawn` way to avoid that.

This PR also moves the events.rs from the wasmcloud-host control interface module to the control interface. This might technically break the `wasmcloud_host` API, but I don't believe there was a reason to use those events until now.

### DIFF
--- a/crates/wasmcloud-control-interface/src/events.rs
+++ b/crates/wasmcloud-control-interface/src/events.rs
@@ -3,22 +3,27 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
-#[doc(hidden)]
+/// Contains metadata about a control event
 pub struct EventHeader {
+    /// Origin host id
     pub host_origin: String,
+    /// Timestamp in UTC seconds since the epoch
     pub timestamp: u64,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
-#[doc(hidden)]
+/// A published control event including the event itself and
+/// additional metadata
 pub struct PublishedEvent {
+    /// Event object
     pub event: ControlEvent,
+    /// Event metadata
     pub header: EventHeader,
 }
+
 /// Represents an event that may occur on the lattice control plane. All timestamps
 /// are to be considered as Unix timestamps in UTC in seconds since the epoch.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
-#[doc(hidden)]
 pub enum ControlEvent {
     HostStarted,
     HostStopped,
@@ -71,6 +76,7 @@ pub enum TerminationReason {
 }
 
 impl ControlEvent {
+    /// Converts a control event into a published event with additional metadata
     pub fn into_published(self, origin: &str) -> PublishedEvent {
         let header = EventHeader {
             host_origin: origin.to_string(),
--- a/crates/wasmcloud-control-interface/src/lib.rs
+++ b/crates/wasmcloud-control-interface/src/lib.rs
@@ -1,11 +1,16 @@
 pub mod broker;
+pub mod events;
 mod generated;
 mod inv;
 mod sub_stream;
 
+use crate::events::PublishedEvent;
 pub use crate::generated::ctliface::*;
+use crossbeam_channel::{unbounded, Receiver};
+use futures::stream::StreamExt;
 use inv::Entity;
 pub use inv::{Invocation, InvocationResponse};
+use nats::asynk::Connection;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, time::Duration};
 use sub_stream::SubscriptionStream;
@@ -13,8 +18,9 @@ use wascap::prelude::KeyPair;
 
 type Result<T> = ::std::result::Result<T, Box<dyn ::std::error::Error + Send + Sync>>;
 
+/// Lattice control interface client
 pub struct Client {
-    nc: nats::asynk::Connection,
+    nc: Connection,
     nsprefix: Option<String>,
     timeout: Duration,
     key: KeyPair,
@@ -22,7 +28,7 @@ pub struct Client {
 
 impl Client {
     /// Creates a new lattice control interface client
-    pub fn new(nc: nats::asynk::Connection, nsprefix: Option<String>, timeout: Duration) -> Self {
+    pub fn new(nc: Connection, nsprefix: Option<String>, timeout: Duration) -> Self {
         Client {
             nc,
             nsprefix,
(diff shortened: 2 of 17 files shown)


crates/wasmcloud-control-interface/src/events.rs
```diff
@@ -3,22 +3,27 @@ use serde::{Deserialize, Serialize};
...
+/// A published control event including the event itself and
+/// additional metadata
 pub struct PublishedEvent {
+    /// Event object
     pub event: ControlEvent,
+    /// Event metadata
     pub header: EventHeader,
 }
+
 /// Represents an event that may occur on the lattice control plane. All timestamps
```
autodidaddict, line 24: nitpick, and it's my own fault - but we should say `control interface` rather than `control plane`. I think some kubernetes / CNCF folks feel quite strongly about this.

crates/wasmcloud-control-interface/src/lib.rs
```diff
@@ -312,6 +319,47 @@ impl Client {
...
+        std::thread::spawn(move || loop {
+            if let Some(msg) = block_on(&mut sub.next()) {
+                match deserialize::<PublishedEvent>(&msg.data) {
+                    Ok(evt) => sender.send(evt).unwrap(),
+                    _ => error!("Object received on event stream was not a PublishedEvent"),
+                }
+            }
+        });
+        Ok(receiver)
+    }
```
autodidaddict, line 362: We should probably add a corresponding function that unsubscribes/clears the event receiver. This makes it so the control interface can choose to be concerned with events for a finite amount of time (or until something meaningful happens according to the developer).

We should also include in the documentation that `events_receiver` creates a subscription and will begin buffering events into a channel so that devs can be aware of the resource cost.

autodidaddict: Maybe we should make this function more of a verb, e.g. `receive_events` , in which case the corresponding stop/halt function could be called `stop_events` or `ignore_events`.

crates/wasmcloud-host/src/host_controller/hc_actor.rs
```diff
@@ -341,6 +342,11 @@ impl Handler<StopActor> for HostController {
                     interest: WasmCloudEntity::Actor(pk.to_string()),
                 });
                 act.actors.remove(&pk);
+                let _ = cp.do_send(PublishEvent {
+                    event: ControlEvent::ActorStopped {
+                        actor: pk.to_string(),
```
autodidaddict, line 347: per clippy: this can just be `pk`, doesn't need to be cloned here since this is the last thing that will use pk.

autodidaddict (changes requested):
Just a minor nit and would like a function to cancel/clear the event subscription.

crates/wasmcloud-control-interface/src/events.rs
brooksmtownsend: I agree, with the varying definitions of `control plane` it's probably best if we don't mention it at all

crates/wasmcloud-control-interface/src/lib.rs
brooksmtownsend: Because each call of `events_receiver` creates a single sender/receiver pair, we may be able to just be explicit here that you can call `drop(receiver)` to close the channel 🤔 I'm going to test that that works before I suggest we go that way.

My only concern with implementing `stop_events` would be that we'd either just `drop` the receiver for them, or we'd have to keep track of the thread that's sending the events so that we can terminate it down the line.

autodidaddict: I think the concern I have is that dropping the receiver doesn't actually remove the subscription from the nats client, unless you drop the nats client as well.

autodidaddict: :thinking: it looks like `sub` actually does _move_ into the closure, so the sub should clean itself up. Above `return`, you could explicitly call `sub.unsubscribe();` and that should avoid the continued background CPU activity from the event subscription.

autodidaddict (changes requested):

crates/wasmcloud-control-interface/src/lib.rs
brooksmtownsend: re: your comment above (below?) yes I think the `sub` should be dropped, and unsubscribed https://docs.rs/nats/0.8.6/src/nats/asynk/subscription.rs.html#64-68. I think explicitly unsubscribing is a good idea to make sure it's not hanging.

autodidaddict: nevermind, I just confirmed that NATS will do this for you on drop:
```rust
impl Drop for Inner {
    fn drop(&mut self) {
        self.client.unsubscribe(self.sid).ok();
    }
}
```

autodidaddict (dismissed):

brooksmtownsend:
@autodidaddict Apologies for dismissing your review, but I added the explicit unsubscribe so we don't rely on `drop` implementations too much 🙂

autodidaddict (approved):

Issue #222, opened 2021-07-30
Update wasmcloud-control-interface to 0.50/OTP level


autodidaddict:
There are a number of known incompatibilities that exist between the "old" (0.18) control interface and the new.

* Events are no longer Rust-serialized `PublishedEvent`s and are instead more flexible `CloudEvent` JSON payloads, with fields that differ based on the event type.
* Events are now published on the `wasmbus.evt.{prefix}` topic
* Link definitions are no longer published via the "advertise links" subject suffix `.links`
* Link defs, claims, OCI references will all need to be queried from the JetStream topics rather than by querying psuedo-random hosts within the lattice

PR #226, opened 2021-08-02
Update control-interface for wasmCloud OTP

brooksmtownsend:
Fixing #222 

## Things to fix
- [ ] Events are no longer Rust-serialized PublishedEvents and are instead more flexible CloudEvent JSON payloads, with fields that differ based on the event type.
- [x] Events are now published on the wasmbus.evt.{prefix} topic
- [ ] Link definitions are no longer published via the "advertise links" subject suffix .links
- [ ] Link defs, claims, OCI references will all need to be queried from the JetStream topics rather than by querying psuedo-random hosts within the lattice


brooksmtownsend:
Closing this for now, planning to pick up updating the control interface once we further define the scope and topics the control interface is concerned about

PR #227, opened 2021-08-03
Switch docker image to debian-slim

alsuren:
@StuartHarris can take most of the credit for this (I just added the debug layer).

The from-scratch image did not have certificates in the right place, so it couldn't pull actors from `eu.gcr.io/wasmcloud-k8s-demo/todo-backend:0.2`.
Alpine's libc is also not capable of loading capability provider .so files.

When setting up a kubernetes cluster, we had a few problems connecting to redis and nats, and our http capability provider would sometimes not come up correctly. I have added a `debug` target that includes the utilities needed to debug these things.

I was also thinking that it would be nice to have a `test` target, that runs some sanity integration checks (spin up a wasmcloud with a manifest and check that it works)

This has been pushed to `eu.gcr.io/wasmcloud-k8s-demo/wasmcloud:0.18.2` and `eu.gcr.io/wasmcloud-k8s-demo/wasmcloud:0.18.2-debug` and tested as part of https://github.com/redbadger/wasmcloud-k8s-demo

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,3 +36,23 @@ USER user:user
 
 ENTRYPOINT ["/usr/local/bin/wasmcloud"]
 CMD ["-V"]
+
+# Sometimes you want to be able to debug the cluster connectivity,
+# or the http/redis capability providers. Build and upload this target with:
+#
+#   docker build --target debug -t wasmcloud/wasmcloud:0.18.2-debug .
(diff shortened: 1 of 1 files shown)


autodidaddict (approved):

brooksmtownsend (approved):
builds and works for me 😄

PR #229, opened 2021-08-05
Migrate control-interface to smithy-IDL and code generated structures

stevelr:
Complete:
- create smithy IDL for control interface api and code-generated "src/control.rs"
- remove actor send rpc messages from this crate (including Invocation,InvocationResponse)
- upgraded dependencies. 
   (removed all dependencies on actix, but that required adding one use of a tokio-timer)
- did not port the Event structures to smithy. There's a long 'enum' for event type and we don't have code generation for that yet.

Incomplete/needs attention:
- clarify nats subject(topic) for add links and remove links
- did not edit any of the other rust code in this repo, so build in related crates such as wasmcloud-host is broken
- pick one field name: Host.uptime or Host.uptime_seconds (cannot port as-is)
- double-check that all control messages are supposed to use json serialization instead of msgpack serialization

This is a breaking api change so bumped crate version to 0.5.0.
This is a DRAFT PR.


brooksmtownsend:
@stevelr just a quick note, we can keep this at `0.4.0` since we never published that one, I think `0.4.0` begins the breaking changes for OTP

stevelr:
this branch is obsolete - the code is in https://github.com/wasmCloud/interfaces/tree/main/control

stevelr:
moved again to new repo github.com/wasmcloud/control-interface.

Closing the PR as the migration is done. There is a new open issue of needing to update nats subjects for creating and deleting links.

PR #269, opened 2023-01-23
docs(roadmap): Adds a proposed roadmap document for wasmCloud projects

thomastaylor312:
I also fixed a license issue where we didn't have the date set

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -18,7 +18,14 @@ This category contains things we are currently (or will very soon be) working on
 is very narrow so you know exactly what is coming next 
 
 - SQL V2 contract
+  - We are trying to align with current webassembly standards for this contract
+  - Mitigate the possibility of injection attacks
 - Finishing up [wadm](https://github.com/wasmCloud/wadm)
+  - Validating design assumptions
+  - Supporting custom scaling algorithms
+  - Deployment and infra docs
+  - Getting started and advanced guides and examples
+  - Removing dependency on Redis (NATS only if possible)
 
 ## Next
 
@@ -45,8 +52,7 @@ as we move them into the [Next](#next) section.
 
 - JavaScript Actors (probably with QuickJS)
 - Unit tests for actors (of any language preferably)
-- Separate actor and provider SDKs (wasmbus refactor)
-- Components and the elimination of wasmbus RPC
+- Components implementation and SDK
 - JS host
 - Embedded-first host
 - Multi-tenant wasmcloud host
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,52 @@
+# wasmCloud Roadmap and Vision
+
+This document covers the current roadmap and vision for wasmCloud, as defined by the current project
+and org maintainers. Each category below has more information on what each thing means. As we start
+to define more of the items covered, we will link to the relevant issues. This may also continue to
+evolve as we use things like GitHub milestones or other tools for tracking our progres. 
+
+Overall, we want the community to have a good idea of where things are headed and what things they
+can contribute to. To be clear, this list is not meant as the end all, be all of what we are doing
+is wasmCloud. Not everything has to be on here to be considered "official work." It is meant more as
+a [guideline](https://tenor.com/bcCX3.gif) for what the maintainers are thinking about. If you don't
+see something on here that you think should be on here, feel free to [open a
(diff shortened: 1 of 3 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,52 @@
...
+see something on here that you think should be on here, feel free to [open a
+PR!](https://github.com/wasmCloud/wasmCloud/pulls)
+
+## Now
+
+This category contains things we are currently (or will very soon be) working on. By definition it
+is very narrow so you know exactly what is coming next 
+
+- SQL V2 contract
+- Finishing up [wadm](https://github.com/wasmCloud/wadm)
```
brooksmtownsend, line 21: If we can flesh these out a little bit, saying that we're doing V2 to further align with the Webassembly standard and protect against injection-type attacks that would be good

Re: wadm, could we add a few high level bullet points like improving the reconciliation algorithm, add a proper error state, and more that you've thought of?

brooksmtownsend (commented):
Love it, just a little suggestion for the "Now" section. Once this is merged we can create a project at the org level

LiamRandall (commented):
LGTM

stephenandary:
I have questions about LinkDef 2.0 because of incorrectness in how I have been thinking about them that was not valid due to lattice requirements. 

I may be thinking in terms of microservices and this may be something I need to stop doing. 

How I was thinking of a Link Def: 

- I planned to group some actors with specific link defs to an explicitly hosted provider and group the actors, interfaces, and providers as a unit for a feature I wanted to enable for my app. 

- Why I wanted to do this was to be able to assign specific infrastructure to some functionality so that I could scale it more discretely, and make sure all traffic on that provider was related to the specific feature for troubleshooting/logging, transparency, and cost. I also felt it made it easier to understand what was required in code to deploy that feature to keep track of dependencies and maybe make it easier to understand. 

- How infrastructure is purchased would be a large factor in determining if this could actually reduce operational cost, or if compute was just being wasted. I am thinking that the current system can use total traffic to scale instead of worrying about more discrete requests based on features enabled. 

-  An example scenario is a load balancer "feature" that may require an HTTP Provider, the LB actor, and the http-interface. A load balancer could potentially be processing millions of requests per second, and maybe those requests have erroneous data.  My ignorance is related to the pros/cons of having a dedicated HTTP provider host for these requests such that one may have dedicated hardware for the feature. Is there merit to scaling based on more discrete types of traffic? Or does it add complexity and not value?

stephenandary:
Also curious more about provisioning infra. Are there plans for a Terraform Provider or Ansible Playbooks? Etc?

thomastaylor312:
@stephenandary Actually that is a good point. I had infra guides in my head but hadn't captured that in the doc. I'll add it in there

As for linkdefs, I think the grouping stuff you are looking for is more suited to [wadm](https://github.com/wasmCloud/wadm), unless I am totally misunderstanding

stephenandary:
@thomastaylor312 let me try to describe better. 

Within a single lattice is it possible to do option A or only option B? 
This is where the cost, transparency, logging, and troubleshooting questions I am wondering about originate. Is this what you understood? 

Option A: 
``` mermaid
graph TD
    A[Actor A] -->|LD Name = featureA| B{Http Provider}
    C[Actor B] --> |LD Name = featureB|D{Http Provider}
```
vs

Option B: 
``` mermaid
graph TD
    A[Actor A] -->|LD Name = default| B{Http Provider}
    C[Actor B] --> |LD Name = default|B{Http Provider}
```

thomastaylor312:
It is possible to do both of them. You can link with the same link name to the same provider (so a single http provider can be routing calls to multiple actors) or you can have ones separated by features. Between that and a wadm manifest, I think you can leverage things like different groupings of hardware. The linking feature is already there (no extra work needed). Linkdefs 2.0 is making it easier to discover what kind of configuration is needed for a given link definition (as well as handling errors better). Wadm being finished is the full unlock for what I think you are describing

brooksmtownsend (approved):

mattwilkinsonn (approved):
LGTM! One collaboration question - do we want to keep this in the repository here?  I'm thinking this might be hard to comment on / have multiple people editing in the future.

Perhaps a GitHub Project could work for this? Or a Trello board / Google doc etc. Project is probably best for discoverability purposes.

PR #283, opened 2023-03-15
feat: add wasi-stderr impl

vados-cosmonic:
## Feature or Problem

Add `wasi-stderr` impl to runtime host so we can move towards supporting WASI components.

## Related Issues

N/A

## Release Information

`next`

## Consumer Impact

Consumers can run WASI components which rely on `wasi-stderr`

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

1. Prepare a WASI component (`wasm-tools component new ..`)
2. Sign the component with a local build of `wash` from main (which supports signing components)
3. Attempt to execute an Actor call 

```console
echo '' | cargo run -- /path/to/your/component_s.wasm Any.Any
```
Instead of seeing an error about `wasm-stderr` being the wrong type, you should see one about `wasi-io` being the wrong type


Makefile
```diff
@@ -0,0 +1,9 @@
+.PHONY: build
+
+CARGO ?= cargo
+
+build: ## Build the project
+	$(CARGO) build
+
+build-watch: ## Build the project (continuously)
+	$(CARGO) watch -- $(MAKE) build
```
rvolosatovs, line 9: Why not just run `cargo watch` with defaults? In any case, that's local developer tooling, which does not belong in this repo

rvolosatovs (changes requested):

rvolosatovs (changes requested):
I will not approve this PR until there is a minimized test case added in `tests/actors`, which fails with current `main`
Once that's done the host changes can be rebased on top of that commit

vados-cosmonic:
The [most recent preview adapter](https://github.com/bytecodealliance/preview2-prototyping/releases/tag/latest) is not the  is not the one used by the host -- the issue is that the host uses something off of main, which is why that version does not work. 

The fix is now to build the adapter off of the version in Cargo.toml, so this can be closed

PR #289, opened 2023-03-16
feat: avoid capability checking by default

rvolosatovs:
## Feature or Problem

Avoid capability checking by default, since host does not have enough context

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Unit tests

### DIFF
--- a/src/actor/mod.rs
+++ b/src/actor/mod.rs
@@ -18,8 +18,8 @@ use tracing::instrument;
 use wascap::jwt;
 use wascap::wasm::extract_claims;
 
-/// Extracts and validates claims contained within `WebAssembly` module
-fn actor_claims(wasm: impl AsRef<[u8]>) -> Result<jwt::Claims<jwt::Actor>> {
+/// Extracts and validates claims contained within `WebAssembly` binary
+fn claims(wasm: impl AsRef<[u8]>) -> Result<jwt::Claims<jwt::Actor>> {
     let claims = extract_claims(wasm)
         .context("failed to extract module claims")?
         .context("execution of unsigned Wasm modules is not allowed")?;
--- a/src/actor/module/wasmbus.rs
+++ b/src/actor/module/wasmbus.rs
@@ -4,9 +4,8 @@ use core::fmt::{self, Debug};
 
 use std::sync::Arc;
 
-use anyhow::{ensure, Context, Result};
+use anyhow::{Context, Result};
 use tracing::{instrument, trace, trace_span, warn};
-use wascap::jwt;
 
 pub mod guest_call {
     use super::wasm;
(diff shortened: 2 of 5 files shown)


src/actor/mod.rs
```diff
@@ -36,6 +36,29 @@ fn actor_claims(wasm: impl AsRef<[u8]>) -> Result<jwt::Claims<jwt::Actor>> {
...
+#[instrument]
+fn authorized_call(
+    claims: &jwt::Claims<jwt::Actor>,
+    binding: &str,
+    namespace: &String,
+) -> Result<()> {
+    if binding.is_empty() && namespace.starts_with('M') && namespace.len() == 56 {
+        // This is an actor-to-actor invocation, skip capability check
+        return Ok(());
+    }
```
brooksmtownsend, line 49: Quick check @autodidaddict , will the namespace be populated with the real actor ID here if I use a call alias?

autodidaddict (approved):
LGTM

src/actor/mod.rs
```diff
@@ -36,6 +36,29 @@ fn actor_claims(wasm: impl AsRef<[u8]>) -> Result<jwt::Claims<jwt::Actor>> {
     Ok(claims.claims)
 }
 
+/// Returns true if the host call is authorized and false otherwise
+#[instrument]
+fn authorized_call(
```
brooksmtownsend, line 41: nit: this seems to return a `Result<()>` so it might be more appropriate to call it `authorize_call` and note that it returns `Ok(())` if authorized, and an error otherwise

brooksmtownsend (changes requested):

src/actor/mod.rs
autodidaddict: yeah was just about to suggest that this be a verb not a noun, but that's a super-nit

rvolosatovs: Yeah, I changed it from `bool`, good point

autodidaddict:
I think the logic here will still break on actor-to-actor calls _via call alias_. A call alias isn't a 56-character public key that begins with "M" so this logic will still block this scenario.

I think the logic needs to be something more like the following psuedocode:

```
if target.starts_with("V") && target.len() == 56 {
  enforce_claims_check()
} else {
  // let it through unchecked
}
```

autodidaddict (dismissed):
we need to change the logic to only ever perform the claims check when the target is a provider public key , e.g. **Vxxxx**

src/actor/module/wasmbus.rs
```diff
@@ -250,18 +249,6 @@ async fn host_call(
...
-            .data()
-            .claims
-            .metadata
-            .as_ref()
-            .map(|jwt::Actor { caps, .. }| caps.as_ref().map(|caps| caps.contains(&ns)))
-            .unwrap_or_default()
-            .unwrap_or_default(),
-        "`{ns}` capability request unauthorized"
-    );
-
```
brooksmtownsend, line 264: Please create followup task so we can investigate ways to do this in this library in the future 👍🏻

brooksmtownsend (dismissed):

src/actor/module/wasmbus.rs
rvolosatovs: https://github.com/wasmCloud/wasmCloud/issues/290

ricochet (approved):

PR #273, opened 2023-03-09
feat: add initial host implementation

rvolosatovs:
*Please review commit-by-commit (https://github.com/wasmCloud/wasmCloud/pull/273/commits) and mind the commit messages/descriptions*

Crate documentation available at https://rvolosatovs.github.io/wasmCloud/wasmcloud/index.html

This PR contains:
- initial host implementation, which:
	- executes actors via wasmtime
	- implements the `wasmbus` FFI interface
	- contains built-in provider implementation adaptors for `log` and `rand` crates
- CI setup for linting, testing and build of cross-compiled binaries along with OCI images containing them

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [x] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [x] aarch64-linux
- [x] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)

Yes, see `tests` within `src/actor.rs`

### Acceptance or Integration

See `tests`

### Manual Verification

Tested with a custom `http-log-rng` actor utilizing builtin providers and https://github.com/wasmCloud/examples/tree/main/actor/echo integrated into this repository

PR #298, opened 2023-03-28
Added `call_context` to HostInvocations and additional `call_with_context` functions

brooksmtownsend:
## Feature or Problem
This PR adds an additional parameter to the component and module `call` functions, with another function head called `call_with_context`, and another parameter `HostInvocation` struct called `call_context`, which is an optional byte vector. Host implementers can use to add additional context onto invocations which will be propagated the entire length of the call.

The original problem this addresses is the fact that there was no way to propagate a parent span ID through a system in order to construct a full trace from the host side. I'd be happy to entertain different naming schemes for the context, different API strategies instead of a `call` and `call_with_context` function, etc. I decided to implement first to show a glimpse at what needed to be done so we can discuss.

## Release Information
`next`

## Consumer Impact
Consumers of this crate won't need to change their code as the `call_with_context` functions are pure additions.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
None

### Acceptance or Integration
None

### Manual Verification
I used this git branch in order to test https://github.com/wasmCloud/wasmcloud-otp/pull/594 which is a full implementation of adding a `call_context` with a span context and then reconstituting the trace context on the other end

image

### DIFF
--- a/src/actor/component.rs
+++ b/src/actor/component.rs
@@ -20,6 +20,8 @@ pub(super) struct Ctx<'a> {
     pub wasi: ::host::WasiCtx,
     pub claims: &'a jwt::Claims<jwt::Actor>,
     pub handler: Arc<Box<dyn Handle<Invocation>>>,
+    /// Opaque context that can be supplied upon initial invocation of an actor
+    pub call_context: Option<Vec<u8>>,
 }
 
 impl Debug for Ctx<'_> {
@@ -39,6 +41,7 @@ impl<'a> Ctx<'a> {
             wasi,
             claims,
             handler,
+            call_context: None,
         }
     }
 }
@@ -55,7 +58,11 @@ impl host::Host for Ctx<'_> {
         let invocation = (namespace, operation, payload)
             .try_into()
             .context("failed to parse invocation")?;
-        match self.handler.handle(self.claims, binding, invocation).await {
+        match self
+            .handler
+            .handle(self.claims, binding, invocation, &self.call_context)
+            .await
+        {
             Err(err) => Ok(Err(err.to_string())),
             Ok(res) => Ok(Ok(res)),
         }
@@ -142,6 +149,31 @@ impl Component {
             .await
             .with_context(|| format!("failed to call operation `{operation}` on module"))
     }
+
+    /// Instantiate a [Component] producing an [Instance] and invoke an operation on it using [Instance::call]
+    /// The `call_context` argument is an opaque byte array that can be used for additional
+    /// metadata around an actor call, like a parent span ID or invocation ID.
+    #[instrument(skip(operation, payload, call_context))]
+    pub async fn call_with_context(
+        &self,
+        operation: impl AsRef<str>,
+        payload: Option<impl AsRef<[u8]>>,
+        call_context: impl AsRef<[u8]>,
+    ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
+        let operation = operation.as_ref();
+        let mut instance = self
+            .instantiate()
+            .await
+            .context("failed to instantiate component")?;
+        if !call_context.as_ref().is_empty() {
+            instance.store.data_mut().call_context = Some(call_context.as_ref().to_vec());
+        }
+
+        instance
+            .call(operation, payload)
+            .await
+            .with_context(|| format!("failed to call operation `{operation}` on module"))
+    }
 }
 
 /// An instance of a [Component]
@@ -168,5 +200,22 @@ impl Instance<'_> {
             )
             .await
             .context("failed to call `guest-call`")
+    }
+
+    /// Invoke an operation on an [Instance] producing a result, where outermost error represents
+    /// a WebAssembly execution error and innermost - the component operation error
+    /// The `call_context` argument is an opaque byte array that can be used for additional
+    /// metadata around an actor call, like a parent span ID or invocation ID.
+    #[instrument(skip_all)]
+    pub async fn call_with_context(
+        &mut self,
+        operation: impl AsRef<str>,
+        payload: Option<impl AsRef<[u8]>>,
+        call_context: impl AsRef<[u8]>,
+    ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
+        if !call_context.as_ref().is_empty() {
+            self.store.data_mut().call_context = Some(call_context.as_ref().to_vec());
+        }
+        self.call(operation, payload).await
     }
 }
--- a/src/actor/mod.rs
+++ b/src/actor/mod.rs
@@ -118,4 +118,31 @@ impl Actor {
                 .with_context(|| format!("failed to call operation `{operation}` on component")),
         }
     }
+
+    /// Instantiate the actor and invoke an operation on it.
+    /// The `call_context` argument is an opaque byte array that can be used for additional
+    /// metadata around an actor call, like a parent span ID or invocation ID.
+    #[instrument(skip(operation, payload, call_context))]
+    pub async fn call_with_context(
+        &self,
+        operation: impl AsRef<str>,
+        payload: Option<impl Into<Vec<u8>> + AsRef<[u8]>>,
+        call_context: impl Into<Vec<u8>> + AsRef<[u8]>,
+    ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
+        let operation = operation.as_ref();
+        match self {
+            Actor::Module(module) => module
+                .call_with_context(
+                    operation,
+                    payload.map(Into::into).unwrap_or(vec![]),
+                    call_context,
+                )
+                .await
+                .with_context(|| format!("failed to call operation `{operation}` on module")),
+            Actor::Component(component) => component
+                .call_with_context(operation, payload, call_context)
+                .await
+                .with_context(|| format!("failed to call operation `{operation}` on component")),
+        }
+    }
 }
--- a/src/actor/module/mod.rs
+++ b/src/actor/module/mod.rs
@@ -40,6 +40,8 @@ pub(super) struct Ctx<'a> {
     pub wasi: wasmtime_wasi::WasiCtx,
     pub claims: &'a jwt::Claims<jwt::Actor>,
     pub wasmbus: wasmbus::Ctx,
+    /// Opaque context that can be supplied upon initial invocation of an actor
+    pub call_context: Option<Vec<u8>>,
 }
 
 impl Debug for Ctx<'_> {
@@ -67,6 +69,7 @@ impl<'a> Ctx<'a> {
             wasi,
             claims,
             wasmbus,
+            call_context: None,
         })
     }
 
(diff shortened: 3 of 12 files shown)


src/actor/component.rs
```diff
@@ -52,9 +55,18 @@ impl host::Host for Ctx<'_> {
         operation: String,
         payload: Option<Vec<u8>>,
     ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
-        let invocation = (namespace, operation, payload)
+        let invocation: Invocation = (namespace, operation, payload)
             .try_into()
             .context("failed to parse invocation")?;
+        //TODO: ugly hacky try make better, maybe better .into above
```
autodidaddict, line 61: Agree with the comment here. I think you should be able to modify the `From<(a,b,c)>` implementation to add `From<(a,b,c,d)>` and then add the optional context to the last position (should be `Some(x)` or `None`).

src/actor/component.rs
```diff
@@ -20,6 +20,8 @@ pub(super) struct Ctx<'a> {
     pub wasi: ::host::WasiCtx,
     pub claims: &'a jwt::Claims<jwt::Actor>,
     pub handler: Arc<Box<dyn Handle<Invocation>>>,
+    /// Opaque context that can be supplied upon initial invocation of an actor
+    pub call_context: Option<Vec<u8>>,
```
rvolosatovs, line 24: How about a generic

```suggestion
    pub call_context: Option<C>,
```

Where `C` defaults to `()` (`C = ()`)?

I really do not like the "opaque optional byte blob" here. What if I want to store a struct here? A number?

src/actor/component.rs
brooksmtownsend: Yep! Doing that right now

src/actor/component.rs
```diff
@@ -52,9 +55,18 @@ impl host::Host for Ctx<'_> {
...
             .context("failed to parse invocation")?;
+        //TODO: ugly hacky try make better, maybe better .into above
+        let invocation: Invocation = if let Invocation::Host(inv) = invocation {
+            Invocation::Host(HostInvocation {
+                call_context: self.call_context.clone(),
+                ..inv
+            })
+        } else {
+            invocation
+        };
```
rvolosatovs: None of this should be necessary, instead, the context should be passed as additional parameter to `handle`, since that's a property of an instance of an actor and not the individual invocation

src/actor/component.rs
```diff
@@ -131,14 +143,25 @@ impl Component {
...
+        self.call_with_context(operation, payload, vec![]).await
+    }
+
+    /// Instantiate a [Component] producing an [Instance] and invoke an operation on it using [Instance::call]
+    #[instrument(skip(operation, payload, call_context))]
+    pub async fn call_with_context(
+        &self,
+        operation: impl AsRef<str>,
+        payload: Option<impl AsRef<[u8]>>,
+        call_context: impl AsRef<[u8]>,
```
rvolosatovs:
```suggestion
        call_context: C
```

src/actor/component.rs
```diff
@@ -159,6 +182,19 @@ impl Instance<'_> {
         operation: impl AsRef<str>,
         payload: Option<impl AsRef<[u8]>>,
     ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
+        self.call_with_context(operation, payload, vec![]).await
```
rvolosatovs: Use a different code path completely

src/capability/mod.rs
```diff
@@ -261,6 +263,8 @@ impl TryFrom<(String, String, Option<Vec<u8>>)> for Invocation {
                 namespace,
                 operation,
                 payload,
+                //TODO: find out if this is fine
+                call_context: None,
```
rvolosatovs: This should not be here

src/actor/component.rs
```diff
@@ -52,9 +55,18 @@ impl host::Host for Ctx<'_> {
         operation: String,
         payload: Option<Vec<u8>>,
     ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
-        let invocation = (namespace, operation, payload)
+        let invocation: Invocation = (namespace, operation, payload)
```
rvolosatovs:
```suggestion
        let invocation = (namespace, operation, payload)
```

The bound will be redundant once you rework the code below

src/actor/component.rs
brooksmtownsend: much cleaner now

src/capability/mod.rs
```diff
@@ -226,6 +226,8 @@ pub struct HostInvocation {
     pub operation: String,
     /// Capability provider invocation payload
     pub payload: Option<Vec<u8>>,
+    /// Invocation call context
+    pub call_context: Option<Vec<u8>>,
```
rvolosatovs, line 230: This should not be here

src/actor/component.rs
brooksmtownsend: I like the idea of a generic 👍🏻

src/actor/component.rs
autodidaddict: Personally I think making generic types in the call context is overkill and forces the consumer to do more gymnastics to get stuff done. The appeal of this type of context is that it's opaque and doesn't push complexity where it doesn't belong.

I get that it looks "cleaner" if you love generics but having been down this road before it feels like it might make things harder down the road.

src/actor/component.rs
autodidaddict: If the consumer wants to store a struct, they can serialize it themselves. Why do we need to inject monomorphic type generation for the context to hold a value it will never inspect?

If the runtime was going to inspect it or call some standardized function on it I can see the desire for generics. but the runtime never interrogates the context so it doesn't need to know its data type.

src/actor/component.rs
```diff
@@ -52,10 +55,14 @@ impl host::Host for Ctx<'_> {
...
             .try_into()
             .context("failed to parse invocation")?;
-        match self.handler.handle(self.claims, binding, invocation).await {
+        match self
+            .handler
+            .handle(self.claims, binding, invocation, self.call_context.clone())
```
rvolosatovs, line 63:
```suggestion
            .handle(self.claims, binding, invocation, &self.call_context)
```

src/actor/component.rs
```diff
@@ -159,6 +180,24 @@ impl Instance<'_> {
...
+    /// Invoke an operation on an [Instance] producing a result, where outermost error represents
+    /// a WebAssembly execution error and innermost - the component operation error
+    /// The `call_context` argument is an opaque byte array that can be used for additional
+    /// metadata around an actor call, like a parent span ID or invocation ID.
+    #[instrument(skip_all)]
+    pub async fn call_with_context(
+        &mut self,
+        operation: impl AsRef<str>,
+        payload: Option<impl AsRef<[u8]>>,
+        call_context: Option<impl AsRef<[u8]>>,
```
rvolosatovs, line 196:
```suggestion
        call_context: impl AsRef<[u8]>,
```

`call_with_context` should always take a context

src/actor/component.rs
```diff
@@ -159,6 +180,24 @@ impl Instance<'_> {
...
+    ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
+        if let Some(call_context) = call_context {
+            self.store.data_mut().call_context = Some(call_context.as_ref().to_vec());
+        }
```
rvolosatovs, line 200:
```suggestion
            self.store.data_mut().call_context = Some(call_context);
```

src/actor/mod.rs
```diff
@@ -105,15 +105,33 @@ impl Actor {
         &self,
         operation: impl AsRef<str>,
         payload: Option<impl Into<Vec<u8>> + AsRef<[u8]>>,
+    ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
+        self.call_with_context(operation, payload, None::<Vec<u8>>)
+            .await
```
rvolosatovs, line 110: Use `call` directly

src/actor/module/mod.rs
```diff
@@ -216,6 +234,20 @@ impl Instance<'_> {
...
+
+    /// Invoke an operation on an [Instance] producing a [Response], supplying a `call_context`
+    /// argument. The `call_context` is an opaque byte array that can be used for additional
+    /// metadata around an actor call, like a parent span ID or invocation ID.
+    #[instrument(skip_all)]
+    pub async fn call_with_context(
+        &mut self,
+        operation: impl Into<String>,
+        payload: impl Into<Vec<u8>>,
+        call_context: Option<impl Into<Vec<u8>>>,
```
rvolosatovs, line 250: Still do not think this should be optional

rvolosatovs (changes requested):
Please make sure there's only one commit in this PR

src/actor/mod.rs
brooksmtownsend: I thought here it would be better to reduce code duplication by having this function call `call_with_context` internally

src/actor/component.rs
brooksmtownsend: Opting to do this later as an enhancement https://github.com/wasmCloud/wasmCloud/issues/299

src/actor/component.rs
brooksmtownsend: See above about replacing the generic in a future PR

src/actor/component.rs
```diff
@@ -131,14 +143,25 @@ impl Component {
         &self,
         operation: impl AsRef<str>,
         payload: Option<impl AsRef<[u8]>>,
+    ) -> anyhow::Result<Result<Option<Vec<u8>>, String>> {
+        self.call_with_context(operation, payload, vec![]).await
```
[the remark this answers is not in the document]
brooksmtownsend: I could separate the logic for `call` and `call_with_context`, but all it does is dictate whether or not a context is set in the `store` a few functions down. I felt like I would end up copy/pasting code all over the place to make that happen, happy to discuss

src/actor/component.rs
brooksmtownsend: Done!

brooksmtownsend:
> Please make sure there's only one commit in this PR

@rvolosatovs I'll make sure to rebase locally once reviewing is done, was planning to leave commits separate for reviewing simplicity

src/actor/mod.rs
brooksmtownsend: Revisiting this one, actually

brooksmtownsend:
@rvolosatovs marked as Ready for Review since I addressed comments / tests / clippy, it's ready for another pass

rvolosatovs (approved):
I am fine with this as a temporary hack to unblock https://github.com/wasmCloud/wasmcloud-otp release

Please be aware that as the capability handling functionality is currently being reworked, this PR will be fully reverted going forward. The functionality you're looking for was already available in the first iteration of this crate https://github.com/wasmCloud/wasmCloud/pull/273/files#diff-a4dc3cd6c97c9e07a2c14e649638feb13828f5842772968728799a1367b51b00R32-R40, but removed on @autodidaddict request. This functionality will be brought back and should fulfill your use case

Issue #317, opened 2023-05-04
[RFC] Separate Actor and Provider SDKs


autodidaddict:
# [RFC] Separation of Guest and Provider SDKs

## Summary
This Request for Comment is soliciting feedback on the plan for separating the wasmCloud guest and provider SDKs. The suggested changes would create separate libraries for guests (actors) and capability providers and do so in a way that takes into account a component-based future while being an example of when not to create shared dependencies.

## Rationale
Today, we have [wasmbus_rpc](https://github.com/wasmCloud/weld/tree/main/rpc-rs), a library that provides an RPC foundation to allow capability providers to communicate with actors via hosts and perform routine functions like applying link definitions. Additionally, this exact same library also provides a foundation on which code is generated to give guest modules (actors) the ability to communicate bi-directionally with capability providers.

This creates an unfortunate tight coupling where both actors and providers share a common dependency or even a sub-tree of dependencies. The library is chock full of conditional compilation that checks for a `wasm` target architecture or non-wasm. At first this may not seem like that bad of a pattern because of the convenience of "one crate to rule them all", however, the real workflow-breaking problems occur with transitive dependencies. `wasmbus-rpc` forces both providers and actors to share a common dependency tree when the vast majority of projects don't need those common libraries.

### Problematic Use Cases
The most common problem case is when the provider and actor both depend on _different versions_ of the same shared dependency. This breaks builds in Rust and would also break builds for Go-based actors and providers.

One of the worst examples of how this pattern causes problems is the [lattice-controller](https://github.com/wasmCloud/capability-providers/tree/main/lattice-controller) capability provider, a provider that gives an actor the ability to interact with a lattice's management interface.

In this scenario, the lattice provider uses `wasmbus-rpc` which, in turn, uses a fixed version of `async-nats` (the NATS client library for Rust). However, the wasmCloud control interface is _also_ exposed over NATS, and so the capability provider needs to create its own instances of types in `async-nats`. At **best** this forces the provider and `wasmbus-rpc` to be in lock-step on the same version of the shared dependency, and at **worst** actually prevents features and bug fixes from being used by either of these crates. If **Go** follows the same shared common library pattern, it too will have conflicting NATS dependencies in the shared dependency tree.

[...]

PR #345, opened 2023-05-30
feat(actor): introduce module support

rvolosatovs:
This is the first step in addition of a "transitional" module support to `wasmcloud_actor` crate. When `module` feature is enabled, `wasmcloud_actor` relies on legacy wasmbus API. The module-based API mimics functionality exposed by component model-based support, meaning, that switch to components will be a breeze for users.

## Related Issues

#317 

## Consumer Impact

- All traits are now good-old `sync`, that may change in the future depeding on how the component model API will shape up, TBD. It seems that `sync` is actually, explicitly, desired
- No direct `wasmbus-rpc` usage in actors
- Fail in builtin execution will result in runtime panic on the actor side, since component model API for these functions does not provide a way to handle errors. That's fine, since users should transition to component model anyway and that's what we're facilitating here
- No code generation, i.e. no custom interfaces (yet?). We probably should not support custom interfaces using Smithy at all as part of this effort as by the time we're done, users should be transitioning to WIT and component-model

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Unit tests

### DIFF
--- a/crates/actor/src/module/host.rs
+++ b/crates/actor/src/module/host.rs
@@ -1,0 +1,56 @@
+/// Invoke an operation on the host
+pub fn call(bd: &str, ns: &str, op: &str, pld: Option<&[u8]>) -> Result<Vec<u8>, String> {
+    #[link(wasm_import_module = "wasmbus")]
+    extern "C" {
+        pub fn __host_call(
+            bd_ptr: *const u8,
+            bd_len: usize,
+            ns_ptr: *const u8,
+            ns_len: usize,
+            op_ptr: *const u8,
+            op_len: usize,
+            pld_ptr: *const u8,
+            pld_len: usize,
+        ) -> usize;
+
+        pub fn __host_response(ptr: *mut u8);
+        pub fn __host_response_len() -> usize;
+
+        pub fn __host_error(ptr: *mut u8);
+        pub fn __host_error_len() -> usize;
+    }
+
+    let (pld_ptr, pld_len) = pld
+        .map(|pld| (pld.as_ptr(), pld.len()))
+        .unwrap_or(([].as_ptr(), 0));
+    match unsafe {
+        __host_call(
+            bd.as_ptr(),
+            bd.len(),
+            ns.as_ptr(),
+            ns.len(),
+            op.as_ptr(),
+            op.len(),
+            pld_ptr,
+            pld_len,
+        )
+    } {
+        1 => {
+            // call succeeded
+            let mut buf = vec![0; unsafe { __host_response_len() }];
+            unsafe { __host_response(buf.as_mut_ptr()) };
+            Ok(buf)
+        }
+        _ => {
+            // call failed
+            let mut buf = vec![0; unsafe { __host_error_len() }];
+            unsafe { __host_error(buf.as_mut_ptr()) };
+            match String::from_utf8(buf) {
+                Ok(e) => Err(e),
+                Err(e) => Err(format!(
+                    "the host provided an error, which is not valid UTF-8: {e}"
+                )),
+            }
+        }
+    }
+}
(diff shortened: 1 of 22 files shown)


rvolosatovs:
Note the similarities between
https://github.com/rvolosatovs/wasmCloud/blob/e792ef48090008ed1c5ee0c56c256e3e9a8c83f8/tests/actors/rust/http-log-rng-module/src/lib.rs
and 
https://github.com/rvolosatovs/wasmCloud/blob/feat/actor-sdk/tests/actors/rust/http-log-rng-component/src/lib.rs

The two are nearly identical, although the former is a usual Wasm module, whereas the latter ulitizes component model APIs generated via WIT

crates/actor/src/module/random.rs
```diff
@@ -0,0 +1,70 @@
...
+pub fn get_random_bytes(n: u64) -> Vec<u8> {
+    let n = n.try_into().expect("too many bytes requested");
+    iter::repeat_with(random32)
+        .map(u32::to_ne_bytes)
+        .flatten()
+        .take(n)
+        .collect()
+}
+
+/// Return a 128-bit value that may contain a pseudo-random value.
```
thomastaylor312, line 23: Nit: I assume this function by definition has to return the two `u64`s rather than a `u128`. If that assumption is correct, it might be worth documenting that this returns two u64s that can be combined into a u128

tests/actors/rust/http-log-rng-component/src/lib.rs
```diff
@@ -2,8 +2,7 @@ wit_bindgen::generate!("actor");
 
 use serde::Deserialize;
 use serde_json::json;
-use wasmbus_rpc::common::{deserialize, serialize};
-use wasmcloud_actor::*;
+use wasmcloud_actor::{logging, HostRng};
 use wasmcloud_interface_httpserver::{HttpRequest, HttpResponse};
```
thomastaylor312, line 6: Nit: Should this test be using the types reexported from `wasmcloud_actor` instead?

thomastaylor312 (approved):
This is all extremely cool. This is good to go pending passing tests.

crates/actor/src/module/host.rs
```diff
@@ -0,0 +1,56 @@
+/// Invoke an operation on the host
+pub fn call(bd: &str, ns: &str, op: &str, pld: Option<&[u8]>) -> Result<Vec<u8>, String> {
```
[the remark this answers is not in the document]
rvolosatovs: Followed the existing naming for consistency, but will update all occurrences across the codebase

crates/actor/src/module/random.rs
rvolosatovs: This is duplicate of what `wit-component` generates for component-model case, (more recent) source: https://github.com/WebAssembly/wasi-random/blob/257e4d73b0e3d034d3c147876d51d1b3db92b17b/wit/insecure-seed.wit#L6-L16

tests/actors/rust/http-log-rng-component/src/lib.rs
rvolosatovs: Yes, eventually, still not sure how to go about http serving in component-model case, I think the most reasonable way to handle this is to wait until `wasi-http` really becomes a thing and then use something that's similar to whatever will happen upstream.
Component model should not be using smithy-generated structs (or at least not directly), this is a temporary solution

autodidaddict (approved):

PR #335, opened 2023-05-19
Create devcontainer.json

lee-orr:
## Feature or Problem

Add dev container support to the wasmCloud repo

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Ran tests within the dev container

PR #346, opened 2023-05-31
feat: add devcontainer support

lee-orr:
## Feature or Problem

Adds support for dev containers - allowing for easily pulling in a pre-configured environment to work on wasmcloud

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

ran the repo & tests within the dev container defined here

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,22 @@
+{
+    "name": "wasmCloud",
+    "image": "rust:latest",
+    "features": {
+        "ghcr.io/devcontainers/features/nix": {}
+    },
+    "customizations": {
+        "vscode": {
+            "settings": {
+                "git.alwaysSignOff": true
+            },
+            "extensions": [
+                "rust-lang.rust-analyzer",
+                "panicbit.cargo",
+                "serayuzgur.crates",
+                "bbenoist.Nix",
+                "tamasfe.even-better-toml"
+            ]
+        }
+    },
+    "postAttachCommand": "rustup show"
+}


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "wasmCloud",
+    "image": "rust:bullseye",
+    "features": {
+        "devwasm.azurecr.io/dev-wasm/dev-wasm-feature/rust-wasi:0": {},
+        "ghcr.io/lee-orr/rusty-dev-containers/cargo-binstall:latest": {},
+        "ghcr.io/lee-orr/rusty-dev-containers/cargo-watch:latest": {},
+        "ghcr.io/lee-orr/rusty-dev-containers/cargo-nextest:latest": {},
```
rvolosatovs, line 8: Like mentioned in https://github.com/wasmCloud/wasmCloud/pull/335 , these seem like non-essential functionality, so I'm curious why it's here?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "wasmCloud",
+    "image": "rust:bullseye",
+    "features": {
+        "devwasm.azurecr.io/dev-wasm/dev-wasm-feature/rust-wasi:0": {},
```
rvolosatovs, line 5: Like mentioned in https://github.com/wasmCloud/wasmCloud/pull/335, I'd expect "anything WASI" to be automatically provisioned by `rustup`, given that we have `rust-toolchain.toml` file

.devcontainer/devcontainer.json
lee-orr: I'll replicate your comments & my responses from the original PR here - note that it seems that github considered my responses as though they were an un-submitted review, so they were only visible to me...

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
...
+        "ghcr.io/lee-orr/rusty-dev-containers/cargo-binstall:latest": {},
+        "ghcr.io/lee-orr/rusty-dev-containers/cargo-watch:latest": {},
```
lee-orr, line 7: (original: https://github.com/wasmCloud/wasmCloud/pull/335#discussion_r1203944859)

> Why is this necessary?
This allows for running builds or tests continuously, whenever a file changes, using [cargo watch](https://crates.io/crates/cargo-watch)
It is a useful development tool, and is recommended in the wasmCloud/wash repository here: https://github.com/wasmCloud/wash/blob/main/docs/DEVELOPMENT.md

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
...
+        "ghcr.io/lee-orr/rusty-dev-containers/cargo-nextest:latest": {},
+        "ghcr.io/devcontainers/features/nix": {}
```
lee-orr, line 9: (original: https://github.com/wasmCloud/wasmCloud/pull/335#discussion_r1203946157)

> If we're adding this, does it support flakes out-of-the-box? Where can I find documentation around this?

Since dev containers are there for development purposes, I included nix since there is a flake.nix file in this repo, as well as tests running around nix in the CI, so I wanted to make sure that using files is supported within the dev container by default.
However - I have never used nix, so that is something you'd need to test separately.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "wasmCloud",
+    "image": "rust:bullseye",
+    "features": {
+        "devwasm.azurecr.io/dev-wasm/dev-wasm-feature/rust-wasi:0": {},
+        "ghcr.io/lee-orr/rusty-dev-containers/cargo-binstall:latest": {},
```
lee-orr, line 6: (original: https://github.com/wasmCloud/wasmCloud/pull/335#discussion_r1203946397)

> What does this do?

It installs [Cargo Binstall](https://github.com/cargo-bins/cargo-binstall) - an alternative to cargo install that will download binaries if they exist, and only build from source if binaries aren't found.
This is used by some of the other features to speed up their install process, since they may not need to build if binaries exist.

lee-orr:
(original: https://github.com/wasmCloud/wasmCloud/pull/335#issuecomment-1560997184)

> Thank you for contribution @lee-orr ! I don't use either dev containers, nor vscode, so not sure how to test this. Could you provide a quick guide on how I could try this out? Otherwise, @ricochet maybe you have a moment to test?
> 
> Is `.devcontainer/devcontainer.json` a default well-known location for stuff like that? Is there a spec somewhere? How will this be maintained and, more importantly, who will do that?
> 
> Apart from that, could you squash your changes into a single commit, preferably following https://www.conventionalcommits.org/en/v1.0.0/ ?

To answer your questions:
- There are a few ways to test it:
    - You can open the PR in github codespaces - that does run it as a VSCode instance, but it is hosted so it doesn't require installing anything locally
    - You can install the dev container CLI (https://github.com/devcontainers/cli), clone the PR locally and run "devcontainer up" within the repository. Then you'd use "devcontainer exec" to execute stuff within the container. They have a vim example here: https://github.com/devcontainers/cli/tree/main/example-usage

As for the `.devcontainer/devcontainer.json` - that is the default location, as per the spec: https://containers.dev/implementors/spec/

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
rvolosatovs: Sure, we use nextest in CI as well here, but it's not required for development, so I'd rather not add it to a "generic" development image. Isn't here a way for users to extend this with functionality they want?

.devcontainer/devcontainer.json
rvolosatovs: Again, it's not required for development, so I don't think it belongs here

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "wasmCloud",
+    "image": "rust:bullseye",
```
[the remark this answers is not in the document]
rvolosatovs: Right, and why we don't just use e.g. `latest`?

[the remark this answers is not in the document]
rvolosatovs: Sure, but that's maintenance burden I'd like to avoid

.devcontainer/devcontainer.json
lee-orr: The features run while the docker image is being built, before the repository is mounted - so they don't have access to that context... We could use a "postAttachCommand" to run rustup though - I need to test it, and ATM am having some issues getting any tests to compile (both from this branch and on upstream main, both within the containers and outside of them) and have other things I need to work on. I'll try again tomorrow.

.devcontainer/devcontainer.json
rvolosatovs: I understand, but I still do not think we should be adding non-standard components here. It's really a rabbit hole where everyone will want to add their favorite piece of tooling here. Let's start small, with something that is minimal and easy to maintain

.devcontainer/devcontainer.json
rvolosatovs, line 3:
```suggestion
    "image": "rust:latest",
```

.devcontainer/devcontainer.json
rvolosatovs, line 8:
```suggestion
```

Let's start with just minimal toolset required to develop

rvolosatovs (changes requested):

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
rvolosatovs: > latest to a different OS base, like bookworm, slim, or alpine

I haven't seen a Docker Hub image switching `latest` tag base from any debian/ubuntu derivative to alpine and I highly doubt anyone would do it, especially not official `rust` image. We should stick to upstream defaults here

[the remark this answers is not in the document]
lee-orr: Sounds good - I'll adjust that

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
lee-orr: Replaced by a `rustup update` on attaching

.devcontainer/devcontainer.json
lee-orr: removed

.devcontainer/devcontainer.json
lee-orr: removed

.devcontainer/devcontainer.json
lee-orr: replaced with running `rustup show` in a post attach command. Note that originally I looked at other options like `rustup update` - but they didn't install the active toolchain & targets, while `show` did.

.devcontainer/devcontainer.json
lee-orr: @rvolosatovs - can I consider this resolved?

rvolosatovs: I think so, if it does not work out-of-the-box, then we (I) can address that in a follow-up!

rvolosatovs (approved):

PR #351, opened 2023-06-02
feat(actor): add `compat` feature

rvolosatovs:
## Feature or Problem

Introduce `compat` feature to `wasmcloud_actor`, when enabled along `component`, all WIT interfaces provided by wasmCloud host are implemented in terms of the `wasmcloud:bus/host`

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)

Added a `http-log-rng-compat` test identical to existing `http-log-rng-component`

### Manual Verification

`cargo test`


crates/actor/src/compat/wasmcloud/bus/host.rs
```diff
@@ -44,7 +44,7 @@ pub fn call(
             // call succeeded
             let mut buf = vec![0; unsafe { __host_response_len() }];
             unsafe { __host_response(buf.as_mut_ptr()) };
-            Ok(buf)
+            Ok(Some(buf))
```
connorsmith256, line 47: When will we return an `Ok(None)`?

connorsmith256 (approved):
Re: `http-log-rng-compat`, how about `builtins-compat`?

crates/actor/src/compat/wasmcloud/bus/host.rs
rvolosatovs: Here (module using wasmbus) - never, but a component (using wasmCloud WIT) might

rvolosatovs:
> Re: `http-log-rng-compat`, how about `builtins-compat`?

Yes! The plan is to rename indeed. Need to see more on how it develops, but I think we should be able to just have `compat`, `module` and `component` crates (they would do more than just call builtins)

PR #361, opened 2023-06-16
feat: begin host implementation

rvolosatovs:
## Feature or Problem

First step towards the Rust host with important core runtime changes. The intention is to partially stabilize `wasmcloud_runtime` API

Everything only runs local for now. NATS layer is in-progress. The focus here is on runtime side changes to allow for what's coming next and proper component model support as well. This is all validated by the local (in-memory) lattice implementation.

- make claims optional (at least for now)
- add streaming support to `wasmcloud:bus`
- rename `wasmcloud_host` -> `wasmcloud_runtime`
- remove all `wasmcloud-interface-*` usages
- add support for `command` executables (I/O actors)

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)

See examples addition and various test actor additions

### Manual Verification

`cargo test`

PR #366, opened 2023-07-11
feat: introduce `wasmbus` lattice

rvolosatovs:
## Related Issues

 #361 

This does not actually handle any lattice commands, just the core functionality and some command parsing

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Tested alongside OTP as part of bigger changeset

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -164,15 +164,15 @@ impl Lattice {
             KeyPair::new(KeyPairType::Server)
         };
 
-        let labels = HashMap::from([
+        let mut labels = HashMap::from([
             ("hostcore.arch".into(), ARCH.into()),
             ("hostcore.os".into(), OS.into()),
             ("hostcore.osfamily".into(), FAMILY.into()),
-            (
-                "path".into(),
-                env::var("WASMCLOUD_HOST_PATH").unwrap_or_default(),
-            ),
         ]);
+        labels.extend(env::vars().filter_map(|(k, v)| {
+            let k = k.strip_prefix("HOST_")?;
+            Some((k.to_lowercase(), v))
+        }));
         let friendly_name = names::Generator::default()
             .next()
             .context("failed to generate friendly name")?;
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1,0 +1,600 @@
+/// Wasmbus lattice configuration
+pub mod config;
+
+pub use config::Lattice as LatticeConfig;
+
+use core::future::Future;
+use core::pin::Pin;
+use core::task::{Context, Poll};
+use core::time::Duration;
+
+use std::collections::{BTreeMap, HashMap};
+use std::env;
+use std::env::consts::{ARCH, FAMILY, OS};
+use std::sync::Arc;
+
+use anyhow::{bail, ensure, Context as _};
+use bytes::Bytes;
+use cloudevents::{EventBuilder, EventBuilderV10};
+use futures::stream::{AbortHandle, Abortable};
+use futures::{try_join, Stream, StreamExt};
+use nkeys::{KeyPair, KeyPairType};
+use serde::Deserialize;
+use serde_json::json;
+use time::format_description::well_known::Rfc3339;
+use time::OffsetDateTime;
+use tokio::spawn;
+use tokio::time::{interval_at, Instant};
+use tokio_stream::wrappers::IntervalStream;
+use tracing::{debug, error, info, instrument, trace, warn};
+use ulid::Ulid;
+use uuid::Uuid;
+
+#[derive(Debug)]
+struct Queue {
+    auction: async_nats::Subscriber,
+    commands: async_nats::Subscriber,
+    pings: async_nats::Subscriber,
+    inventory: async_nats::Subscriber,
+    links: async_nats::Subscriber,
+    queries: async_nats::Subscriber,
+    registries: async_nats::Subscriber,
+}
+
+impl Stream for Queue {
+    type Item = async_nats::Message;
+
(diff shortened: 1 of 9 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -0,0 +1,642 @@
...
+            nats.subscribe("wasmbus.ctl.default.auction.>".into()),
+            nats.subscribe(format!(
+                "wasmbus.ctl.default.cmd.{}.*",
+                host_key.public_key()
+            )),
+            nats.subscribe(format!(
+                "wasmbus.ctl.default.get.{}.inv",
+                host_key.public_key()
+            )),
+        )
```
brooksmtownsend, line 155: The `default` in each of these should eventually be a parameter that's passed in and defined at runtime, to match the current functionality. It's the lattice prefix, but as the PR text mentions this isn't the full functionality so I'm just calling attention to it 👍🏻

crates/host/src/wasmbus/mod.rs
```diff
@@ -0,0 +1,642 @@
...
+        };
+
+        let labels = HashMap::from([
+            ("hostcore.arch".into(), ARCH.into()),
+            ("hostcore.os".into(), OS.into()),
+            ("hostcore.osfamily".into(), FAMILY.into()),
+            (
+                "path".into(),
+                env::var("WASMCLOUD_HOST_PATH").unwrap_or_default(),
+            ),
```
brooksmtownsend, line 216: Is this just for debugging purposes? Or to overwrite the nix store PATH variable?

crates/host/src/wasmbus/mod.rs
```diff
@@ -0,0 +1,642 @@
...
+        let wasmbus = Arc::new(wasmbus);
+        let queue = spawn({
+            let wasmbus = Arc::clone(&wasmbus);
+            async {
+                Abortable::new(queue, queue_abort_reg)
+                    .for_each(move |msg| {
+                        let wasmbus = Arc::clone(&wasmbus);
+                        async move { wasmbus.handle_message(msg).await }
+                    })
+                    .await;
```
brooksmtownsend, line 273: I have not used this syntax before, does this allow this host to handle multiple wasmbus messages simultaneously or is it handled in sequence?

brooksmtownsend (approved):
Real solid bones for the wasmbus lattice functionality 🚀 I really only had a few questions, none of them blocking for the scope of this PR

crates/host/src/wasmbus/mod.rs
rvolosatovs: This one is sequential, there's https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each_concurrent for concurrent iteration

crates/host/src/wasmbus/mod.rs
brooksmtownsend: By default, the only labels that I have on a vanilla OTP host is this:
```
  hostcore.osfamily                     unix
  hostcore.os                           macos
  hostcore.arch                         aarch64
```

brooksmtownsend: So I think unless you wanted to patch that bug with the `HOST_PATH` environment variable, you do not need the `path` label

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I think we should use the concurrent iteration unless you can see some consequences there, one scenario I can think of is `wadm` sending out 10 commands at the same time and it would be beneficial to handle those concurrently (maybe up to a certain limit) to not delay responses. If it's easy to do now then I think it makes sense to

rvolosatovs: Since this is an optimization, and handling requests out-of-order can have multiple unforeseen side-effects, I think we should wait and do it right instead when we're in position to do so (i.e. the lattice functionality is fully in-place).

connorsmith256 (approved):
Left suggestions but no blockers from merging

crates/host/src/wasmbus/mod.rs
```diff
@@ -0,0 +1,600 @@
...
+            nats.subscribe("wasmbus.ctl.default.get.*".into()),
+            nats.subscribe("wasmbus.ctl.default.auction.>".into()),
+            nats.subscribe(format!(
+                "wasmbus.ctl.default.cmd.{}.*",
+                host_key.public_key()
+            )),
+            nats.subscribe(format!(
+                "wasmbus.ctl.default.get.{}.inv",
+                host_key.public_key()
+            )),
```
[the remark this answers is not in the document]
rvolosatovs: This seems reasonable, but that is not what OTP host does and I think we should not make any changes in this regard until we're at full feature-parity. I do not know what unexpected side-effects this change of behavior could have.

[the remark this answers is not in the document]
autodidaddict: The no responders behavior is actually what we want here. If there are no hoses listening in a lattice for control requests, some clients actually count on getting no responders back

autodidaddict:
I can't tell from my phone but it looks like the 'default' lattice has been hard coded. Just want to mention here that lattice ids other than default need to be supported

rvolosatovs:
> I can't tell from my phone but it looks like the 'default' lattice has been hard coded. Just want to mention here that lattice ids other than default need to be supported

@brooksmtownsend already pointed that out in https://github.com/wasmCloud/wasmCloud/pull/366#discussion_r1260157830

This will be addressed in a follow-up

PR #372, opened 2023-07-14
feat(wasmbus): implement provider start/stop

rvolosatovs:
## Feature or Problem

Implement provider auction, start and stop

## Related Issues

Similar to #371 , but incomplete

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

You should be able to start and stop a provider from the washboard, that's about it. The providers should show up in the inventory and on the washboard with some missing details.

```
wash start provider wasmcloud.azurecr.io/httpserver:0.16.0
wash stop provider <HOST> VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M default wasmcloud:httpserver
```

Note, that the start command times out due to semantic difference here from OTP. That will be addressed later, since that's not an isolated case.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -30,17 +30,18 @@ use serde_json::json;
 use time::format_description::well_known::Rfc3339;
 use time::OffsetDateTime;
 use tokio::io::{stderr, AsyncWrite};
-use tokio::spawn;
 use tokio::sync::{watch, RwLock};
 use tokio::time::{interval_at, Instant};
+use tokio::{process, spawn};
 use tokio_stream::wrappers::IntervalStream;
 use tracing::{debug, error, info, instrument, trace, warn};
 use ulid::Ulid;
 use uuid::Uuid;
 use wascap::jwt;
 use wasmcloud_control_interface::{
-    ActorAuctionRequest, ActorDescription, ScaleActorCommand, StartActorCommand,
-    StartProviderCommand, StopActorCommand, StopHostCommand, UpdateActorCommand,
+    ActorAuctionRequest, ActorDescription, ProviderAuctionRequest, ProviderDescription,
+    ScaleActorCommand, StartActorCommand, StartProviderCommand, StopActorCommand, StopHostCommand,
+    StopProviderCommand, UpdateActorCommand,
 };
 use wasmcloud_runtime::{ActorInstancePool, Runtime};
 
@@ -308,7 +309,21 @@ type Annotations = BTreeMap<String, String>;
 struct Actor {
     pool: ActorInstancePool,
     instances: RwLock<HashMap<Option<Annotations>, Vec<Arc<ActorInstance>>>>,
-    actor_ref: String,
+    image_ref: String,
+}
+
+#[derive(Debug)]
+struct ProviderInstance {
+    child: process::Child,
+    id: Ulid,
+    annotations: Option<Annotations>,
+}
+
+#[derive(Debug)]
+struct Provider {
+    claims: jwt::Claims<jwt::CapabilityProvider>,
+    instances: HashMap<String, ProviderInstance>,
+    image_ref: String,
 }
 
 /// Wasmbus lattice
@@ -323,6 +338,7 @@ pub struct Lattice {
     host_key: KeyPair,
     labels: HashMap<String, String>,
     nats: async_nats::Client,
+    providers: RwLock<HashMap<String, Provider>>,
     runtime: Runtime,
     start_at: Instant,
     stop_tx: watch::Sender<Option<Instant>>,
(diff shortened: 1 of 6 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -495,12 +509,42 @@ impl Lattice {
...
+            .iter()
+            .flat_map(
+                |(
+                    public_key,
+                    Provider {
+                        claims, instances, ..
+                    },
+                )| {
+                    instances.keys().map(move |link_name| {
+                        if let Some(jwt::CapabilityProvider { capid, .. }) = &claims.metadata {
```
connorsmith256, line 525: Do you need to duplicate the json macro, or can you move the capid to its own variable to make the branch logic smaller?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1039,19 +1104,71 @@ impl Lattice {
...
+                        link_name,
+                        ProviderInstance {
+                            child,
+                            id: Ulid::new(),
+                        },
+                    )]),
+                });
+            }
+            hash_map::Entry::Occupied(mut entry) => {
+                debug!(
```
connorsmith256, line 1133: This should be moved into the if statement below

crates/host/src/wasmbus/mod.rs
```diff
@@ -1039,19 +1104,71 @@ impl Lattice {
...
+                let Provider { instances, .. } = entry.get_mut();
+                if let hash_map::Entry::Vacant(entry) = instances.entry(link_name) {
+                    let child = command
+                        .spawn()
+                        .context("failed to spawn provider process")?;
+                    entry.insert(ProviderInstance {
+                        child,
+                        id: Ulid::new(),
+                    });
+                }
```
connorsmith256, line 1146: There should be an else here, for the error case: the host is already running this provider on this link name

crates/host/src/wasmbus/mod.rs
```diff
@@ -1039,19 +1104,71 @@ impl Lattice {
...
-        bail!("TODO");
+        let (path, claims) = crate::fetch_provider(&provider_ref, &link_name)
+            .await
+            .context("failed to fetch provider")?;
+        let mut command = process::Command::new(&path);
+
+        // TODO: Publish events
+        match self.providers.write().await.entry(claims.subject.clone()) {
+            hash_map::Entry::Vacant(entry) => {
+                debug!(?path, link_name, "launch a new provider");
```
connorsmith256, line 1117: Nit: if I start two NATS providers, one on link name `foo` and the other `bar`, this will print:

```
path=X link_name=foo launch a new provider
path=X link_name=bar launch a new instance of an existing provider
```

IMO there's no reason to emit different logs between the first provider started and subsequent providers; these should be treated as independent

crates/host/src/wasmbus/mod.rs
```diff
@@ -1111,13 +1228,51 @@ impl Lattice {
...
+                )| {
+                    instances.iter().filter_map(
+                        move |(link_name, ProviderInstance { child, id })| {
+                            let pid = child.id()?;
+                            let instance_id = Uuid::from_u128((*id).into());
+                            if let Some(jwt::CapabilityProvider { capid, .. }) = &claims.metadata {
+                                Some(json!({
+                                    "public_key": public_key,
+                                    "link_name": link_name,
+                                    "contract_id": capid,
```
connorsmith256, line 1251: Same comment about not duplicating the JSON as I made before. I prefer to not have to visually diff the structures

connorsmith256 (changes requested):

crates/host/src/wasmbus/mod.rs
rvolosatovs: > this provider

The host in this branch is not running this provider yet (assuming that the provider claim "subject" field uniquely identifies a provider), so I am not sure what you mean

rvolosatovs: Ah, I see now - I have misinterpreted https://github.com/wasmCloud/wasmcloud-otp/blob/4feeb1ff9b4821690a3e42023a926f4ea5c24ede/host_core/lib/host_core/providers/provider_supervisor.ex#L277-L285

connorsmith256: I'm talking about a case like this:
```
wash start provider -l foo wasmcloud.azurecr.io/httpserver:0.17.0
wash start provider -l foo wasmcloud.azurecr.io/httpserver:0.17.0
```

The second command should be rejected by the host, since `(httpserver, foo)` is already running

connorsmith256 (approved):

PR #376, opened 2023-07-14
build: update dependencies

rvolosatovs:
## Feature or Problem

Update dependencies

The version bounds are relaxed to simplify downstream usage for those depending on crates within the workspace as a library

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`

PR #378, opened 2023-07-18
feat: introduce `wasmcloud-component-adapters`

rvolosatovs:
## Feature or Problem

- Introduce a new `wasmcloud-component-adapters` crate, which exposes the WASI adapters
- Remove the need for `nightly` Rust toolchain
- Switch to upstream wasmtime

Rather than building the adapters, they are now fetched from wasmtime release page and can be overriden during build.

## Related Issues

Using nightly prevents us from updating deps today (https://github.com/wasmCloud/wasmCloud/pull/376) due to https://github.com/rust-lang/rust/issues/113152 https://github.com/dtolnay/proc-macro2/pull/391

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


brooksmtownsend (approved):
So there are components in wasi-preview-2, and then there are _adapters_ for both the command and reactor pattern, which is what this crate handles right? Just making sure I understand the different options.

rvolosatovs:
> So there are components in wasi-preview-2, and then there are _adapters_ for both the command and reactor pattern, which is what this crate handles right? Just making sure I understand the different options.

This crate exposes component adapters, which translate WASI preview1 calls to WASI preview2 calls. It's required, because WASI preview1 is what language toolchains link towards today, but the host in "component mode" only exports WASI preview2 functions.
Note, however, that there is also an experimental opt-in host WASI adapter, which makes these adapters redundant.

brooksmtownsend (approved):

crates/actor/derive/src/lib.rs
```diff
@@ -58,7 +58,7 @@ pub fn export_actor(input: TokenStream) -> TokenStream {
             let op = match String::from_utf8(op) {
                 Ok(op) => op,
                 Err(e) => {
-                    let e = format!("operation is not valid UTF-8: {e}");
+                    let e = format!("operation is not valid UTF-8: {}", e);
```
brooksmtownsend, line 61: Did clippy get angry with you over these?

Issue #381, opened 2023-07-19
[RFC] Use labels to indicate RFC status, and add GitHub template


brooksmtownsend:
# RFC: Use labels to indicate RFC status, and add GitHub template

## Introduction

To encourage a formal process to manage RFCs and clearly communicate their status, I propose that we add labels to our github organization to better indicate their status and encourage more community members to take implementation of RFCs.

## Motivation

Without the labels, it can be unclear if an RFC is recently proposed, still in conversation, accepted, or denied. Additionally, there is no issue template, which can discourage RFC contribution

## Detailed Design

I propose we add three labels for use with RFCs:
1. rfc-proposed - For when an RFC is newly proposed
2. rfc-discussion-needed - For when active discussion is going on in the RFC and it's not scoped enough for work to begin
3. rfc-accepted - For when an RFC is adequately scoped and ready for implementation. It's up to maintainers to decide if an RFC is scoped enough for work, and affected projects should be indicated.

Additionally, the template I used for this RFC is what I would propose we create as an issue template.
```markdown
# RFC: (Title of your proposal)

## Introduction

(Here, provide a brief summary of the feature or idea that you're proposing. This should be a couple of sentences or a short paragraph.)

## Motivation

(Explain the background behind this RFC. What problem are you trying to solve? Why is this solution needed? Why is it relevant for your project? This is your chance to provide a persuasive argument for your proposal.)

## Detailed Design

(This is the heart of your RFC. Describe your proposed solution in detail. If you're suggesting a new feature, how will it work? If it's a new policy or process, what are the steps involved? Be as specific as possible. Include any code samples, diagrams, or other materials that would help others understand your proposal.)

## Backwards Compatibility

(Discuss how your proposal will affect existing parts of your project. Will it introduce breaking changes? How will you mitigate these issues?)

## Alternatives Considered

(Describe any alternative solutions that you considered and why you chose your current proposal over them. This demonstrates that you've thought carefully about your proposal and helps others understand your decision-making process.)

## Unresolved Questions

(List any questions or issues that you weren't able to address in your proposal. This helps others understand what feedback you're looking for.)

## Conclusion

(End with a summary of your proposal and the next steps. What kind of feedback are you looking for? When do you hope to make a decision? Be clear about the process moving forward.)

```

## Backwards Compatibility

N/A

## Alternatives Considered

[...]

PR #382, opened 2023-07-20
chore: add RFC issue template

brooksmtownsend:
This PR adds the RFC template as proposed in #381


rvolosatovs (commented):
Thanks for the contribution, if we're following conventional commits, could you please:
1. ensure that the sentence starts with lowercase character
2. use verbs in imperative form

See examples at https://www.conventionalcommits.org/en/v1.0.0/

Issue #387, opened 2023-07-21
Replace `vendor/weld` by `crates/provider`


rvolosatovs:
In a single commit, `vendor/weld` should be removed and `crates/provider`, containing a `wasmcloud-provider` crate should be added. (draft at https://github.com/thomastaylor312/provider-sdk)
This will ensure a meaningful diff and git history, for example `git blame` should (mostly) still work.

PR #393, opened 2023-07-25
feat(provider-sdk): Adds new provider SDK to repo

thomastaylor312:
This is now manually tested and in a state where I think we should have it in the repo. We should be able to keep iterating from there


Cargo.toml
```diff
@@ -39,6 +39,7 @@ test-actors = { workspace = true }
 [workspace]
 members = [
     "crates/providers/*",
+    "crates/provider-sdk",
```
rvolosatovs, line 42: Rather than doing this, could you migrate the http server provider to using this crate? Adding it here would become redundant

crates/provider-sdk/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
+/Cargo.lock
+wit/deps/
```
rvolosatovs, line 3: As discussed offline, this should not be ignored for now

crates/provider-sdk/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
+/Cargo.lock
```
rvolosatovs, line 2: Since we're in a workspace, this should be irrelevant

crates/provider-sdk/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
```
rvolosatovs, line 1: Redundant, since we're in a workspace

crates/provider-sdk/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "provider-sdk"
```
rvolosatovs, line 2: This name is way too generic, please use `wasmcloud-provider`, as outlined in https://github.com/wasmCloud/wasmCloud/issues/387 (also the directory should be renamed)

crates/provider-sdk/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[features]
+default = [ ]
+otel = ["opentelemetry", "tracing-opentelemetry", "opentelemetry-otlp"]
+
+[dependencies]
```
rvolosatovs, line 10: All of these should be using workspace dependencies and not declare custom versions in this crate

crates/provider-sdk/wit/deps.toml
```diff
@@ -0,0 +1,2 @@
+logging = "https://github.com/WebAssembly/wasi-logging/archive/main.tar.gz"
+wasmcloud-core = "."
```
rvolosatovs, line 2: Why would this ever be necessary?

rvolosatovs (changes requested):
This PR should remove the vendored `weld`, as outlined in https://github.com/wasmCloud/wasmCloud/issues/387

Did not review the implementation itself

crates/provider-sdk/wit/deps.toml
thomastaylor312: I was just copy pasting from previous examples

crates/provider-sdk/Cargo.toml
thomastaylor312: Some of them can be, sure. But I don't think optional deps are, so we'll have to keep those

crates/provider-sdk/Cargo.toml
thomastaylor312: I think the `sdk` is important. Will preface `wasmcloud` though

crates/provider-sdk/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "provider-sdk"
+version = "0.1.0"
+edition = "2021"
```
rvolosatovs, line 4: This should be inherited along with other metadata, just like in all other crates in the workspace

rvolosatovs (changes requested):

Cargo.toml
thomastaylor312: I'd rather do that in a follow up

thomastaylor312:
I've addressed all other nits. I understand the desire for #387, but I honestly don't think a clean diff is really important for this. This is "net new" for all intents and purposes and so matters less to me to do it now (along with the fact that I'd have to change the httpserver as well. It'd be easier with all the ongoing work to make this available for people and then come back to update the other things

crates/provider-sdk/wit/wasmcloud.wit
connorsmith256, line 1: This is in `wit` and `wit/deps/wasmcloud-core`. Should it be in both?

crates/provider-sdk/src/provider.rs
```diff
@@ -0,0 +1,541 @@
...
+                    reply: Some(reply_to),
+                    payload,
+                    ..
+                }) = msg
+                {
+                    let shutmsg: ShutdownMessage = serde_json::from_slice(&payload).unwrap_or_default();
+                    // Backwards compatibility - if no host (or payload) is supplied, default
+                    // to shutting down unconditionally
+                    if shutmsg.host_id.is_empty() {
+                        warn!("Please upgrade your wasmcloud host to >= 0.59.0 if you use lattices with multiple hosts.")
```
connorsmith256, line 357: It might be time to assume all hosts will provide a host ID on shutdown messages

connorsmith256 (approved):
I noticed you dropped the OTEL headers on outbound RPC messages, and added them to `trace_context` on the invocation. Does it hurt to leave it on both for backwards-compatibility purposes for a while?

rvolosatovs:
> I've addressed all other nits. I understand the desire for #387, but I honestly don't think a clean diff is really important for this. This is "net new" for all intents and purposes and so matters less to me to do it now (along with the fact that I'd have to change the httpserver as well. It'd be easier with all the ongoing work to make this available for people and then come back to update the other things

There is no *need* to change the HTTP server, but - there is a lot of duplicated code here and there is no way to correlate the two if the `vendor/weld` is not removed in the same commit. That directory exists purely to be removed.

Cargo.toml
rvolosatovs: Sure

rvolosatovs (commented):
Waiting on the first commit in this PR to be amended with removal of `vendor/weld`

crates/provider-sdk/wit/wasmcloud.wit
thomastaylor312: Yes, the deps folder contains everything. Kinda like a vendor directory

crates/provider-sdk/src/provider.rs
thomastaylor312: Oh right. I tried to remove most of the old stuff. Let me go remove this

thomastaylor312:
> I noticed you dropped the OTEL headers on outbound RPC messages, and added them to trace_context on the invocation. Does it hurt to leave it on both for backwards-compatibility purposes for a while?

@connorsmith256 It makes it really easy to not have the NATS dep exposed publicly anymore. I can add both, but I'll need to do a little investigating. I'll file an issue to come back and add that back in

@rvolosatovs I know it won't be easy to correlate, but I am arguing that it doesn't matter long term. Am I wrong in assuming that the http server provider is depending on those files? If so, I probably shouldn't just remove them until I come back to update the http server. If it isn't and I can just remove, that is easy. Context: I am trying to get this in because a whole bunch of stuff is in flight on this. At a minimum @vados-cosmonic and @ricochet are working on stuff that will likely touch this, so I care much less about cleanliness of git history in cases such as these if that is the only reason

thomastaylor312:
@connorsmith256 added #394 to capture your comment

rvolosatovs:
> @rvolosatovs I know it won't be easy to correlate, but I am arguing that it doesn't matter long term. Am I wrong in assuming that the http server provider is depending on those files? If so, I probably shouldn't just remove them until I come back to update the http server. If it isn't and I can just remove, that is easy. Context: I am trying to get this in because a whole bunch of stuff is in flight on this repo. At a minimum @vados-cosmonic and @ricochet are working on stuff that will likely touch this, so I care much less about cleanliness of git history in cases such as these if that is the only reason

Nothing depends on code in `vendor`. Please remove in the same commit, in which you add the "new" stuff

rvolosatovs:
> @rvolosatovs I know it won't be easy to correlate, but I am arguing that it doesn't matter long term. Am I wrong in assuming that the http server provider is depending on those files? If so, I probably shouldn't just remove them until I come back to update the http server. If it isn't and I can just remove, that is easy. Context: I am trying to get this in because a whole bunch of stuff is in flight on this repo. At a minimum @vados-cosmonic and @ricochet are working on stuff that will likely touch this, so I care much less about cleanliness of git history in cases such as these if that is the only reason

This has nothing to do with "cleanliness of git history", rather, by having a clear single commit transitioning from the legacy implementation to the new one, we can:
1. See what changed, which means:
   a. It's easier to meaningfully review current PR
   b. It's easier to debug an issue, which may have been introduced as a side-effect of this transition
2. Give credit where credit is due, by copy-pasting the existing implementation all code attribution goes to you personally, which is false - e.g. the "witty comment" written at the top of a file I see in the diff was (most likely) not written by you, or maybe it was - we will never know if we lose the git history.
3. Most importantly, we can "go back in time" and see when and *why* was a particular change made by looking at the commit, which changed a particular LoC (which is why "clean" git history is important and commit messages akin to "address feedback" are unacceptable in any serious open-source project)

rvolosatovs:
@thomastaylor312 please see https://github.com/wasmCloud/wasmCloud/pull/396/commits/ada90674df5130be6320788bcb08b7868f3b67a5 and note how the diff now looks very different and is much more meaningful. 
There is now clear indication of what have changed in relation to the old version, which will also make migration way simpler for provider developers

rvolosatovs (changes requested):
Addressed my own feedback (based on top of #395) in https://github.com/wasmCloud/wasmCloud/pull/396/commits/ada90674df5130be6320788bcb08b7868f3b67a5 it would be beneficial to merge #395 first and then rebase this PR (or just cherry-pick the commit I linked)

Alternatively, we could simply merge #396, which includes both PRs.

Requesting changes to prevent merge of this PR in current state

PR #395, opened 2023-07-26
feat: implement linkdef addition and removal

rvolosatovs:
## Feature or Problem

- Implement linkdef `put`, `delete` and most things related
- Add a test case for Wasmbus lattice

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`


crates/host/src/wasmbus/mod.rs
```diff
@@ -425,6 +445,38 @@ impl Lattice {
             .build()
             .context("failed to build runtime")?;
         let event_builder = EventBuilderV10::new().source(host_key.public_key());
+
+        let jetstream = async_nats::jetstream::new(nats.clone());
+        // TODO: Use prefix
+        let bucket = format!("LATTICEDATA_{prefix}", prefix = "default");
+        jetstream
+            .create_stream(async_nats::jetstream::stream::Config {
```
connorsmith256, line 453: This isn't quite right:
1) We shouldn't be creating a stream manually, we should be creating a bucket
2) We should only attempt to create the bucket if it doesn't already exist. E.g. the first host started in a lattice can create the bucket; subsequent hosts should re-use the existing bucket
3) When we construct the JS client we need to pass the JS domain

crates/host/src/wasmbus/mod.rs
```diff
@@ -425,6 +445,38 @@ impl Lattice {
...
+                max_messages: -1,
+                max_messages_per_subject: 1,
+                name: format!("KV_{bucket}"),
+                num_replicas: 1,
+                subjects: vec![format!("$KV.{bucket}.>")],
+                ..async_nats::jetstream::stream::Config::default()
+            })
+            .await
+            .map_err(|e| anyhow!(e).context("failed to create data bucket"))?;
+        let data = jetstream
```
connorsmith256, line 471: Naming nit: `store` (or `kv_store`, or `lattice_data_store`) matches the underlying type and is more specific than `data`

crates/host/src/wasmbus/mod.rs
```diff
@@ -1101,12 +1179,70 @@ impl Lattice {
                 instances: HashMap::default(),
             });
         if let hash_map::Entry::Vacant(entry) = instances.entry(link_name.clone()) {
-            debug!(?path, "spawn provider process");
-            let child = process::Command::new(&path)
+            let id = Ulid::new();
+            let async_nats::ServerInfo { host, port, .. } = self.nats.server_info();
```
connorsmith256, line 1183: The provider RPC settings (host, port, JWT, seed, etc) are not necessarily the same as the the CTL interface settings, or the RPC settings for the host. See https://wasmcloud.com/docs/hosts/elixir/host_configure for details

I'm okay with assuming everything is the same for now — we've hard-coded `default` for the lattice prefix too — but please file an issue to add support for different settings

crates/host/src/wasmbus/mod.rs
```diff
@@ -1101,12 +1179,70 @@ impl Lattice {
...
+            let invocation_seed = self
+                .cluster_key
+                .seed()
+                .context("cluster key seed missing")?;
+            let links = self.links.read().await;
+            let link_definitions: Vec<_> = links
+                .values()
+                .filter(|ld| ld.provider_id == claims.subject && ld.link_name == link_name)
+                .collect();
+            let data = serde_json::to_vec(&json!({
```
connorsmith256, line 1193: Naming nit: `host_data`, which should also be a struct type (defined in `provider-sdk/src/core.rs` for reference)

crates/host/src/wasmbus/mod.rs
```diff
@@ -1101,12 +1179,70 @@ impl Lattice {
...
+                .write_all(STANDARD_NO_PAD.encode(&data).as_bytes())
+                .await
+                .context("failed to write provider data")?;
+            stdin
+                .write_all(b"\r\n")
+                .await
+                .context("failed to write newline")?;
+            stdin.shutdown().await.context("failed to close stdin")?;
+
+            let child = spawn(async move {
```
connorsmith256, line 1236: Do we want to shadow `child`? We're returning it as part of the `ProviderInstance` below, so we can issue shutdown messages to the provider. If we pass this child handle instead, won't we kill this process, rather than the provider process?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1431,4 +1606,151 @@ impl Lattice {
...
+        let mut links = self.links.write().await;
+        links.insert(id.to_string(), ld.clone());
+
+        self.publish_event(
+            "linkdef_set",
+            event::linkdef_set(id, actor_id, provider_id, link_name, contract_id, values),
+        )
+        .await?;
+
+        // TODO: Broadcast `linkdef_added`
```
connorsmith256, line 1650: I don't think this is needed. `process_linkdef_put` is called from `process_entry`, which is triggered by watching the KV bucket. Every host in the lattice should receive the same watch event and process it locally

This may have been needed in the OTP host, where links could be created through two different ways: direct writes to the bucket, or (legacy) messages on `wasmbus.ctl`. We won't need to support the latter here

(Note each host in the lattice will publish the `linkdef_set` event, resulting in duplicates. That's a known ugly behavior of wasmCloud hosts)

connorsmith256 (changes requested):
Mostly there, but I have a few non-nits

crates/host/src/wasmbus/mod.rs
rvolosatovs: 1. This is exactly what OTP does, if we do not do it exactly the same way - OTP and Rust host cannot coexist in the same lattice
2. That's exactly how it works - it does not do anything if the stream already exists
3. That will come later together with all the other config (like lattice prefixes)

crates/host/src/wasmbus/mod.rs
rvolosatovs: The key is "LATTICEDATA", `data` directly corresponds to the key. `store` is too opaque, `kv_store` does not convey any semantic meaning, `lattice_data_store` is redundant, since the struct *is* called `Lattice`

I'm happy to rename to `data_store` if you prefer

crates/host/src/wasmbus/mod.rs
rvolosatovs: Yes, that will be fixed together with all the config TODOs

crates/host/src/wasmbus/mod.rs
rvolosatovs: For context, please read the implementation of this method https://docs.rs/async-nats/latest/async_nats/jetstream/context/struct.Context.html#method.create_key_value

The method itself does not provide means to specify all the options that OTP specifies (which is a requirement for this effort) - we have to use a lower-level one

connorsmith256: 1) The code you referenced for OTP should have the same effect as creating a bucket. See the equivalent from the NATS CLI commands:

```
nats stream add KV_foobar
? Subjects $KV.foobar.>
? Storage file
? Replication 1
? Retention Policy Limits
? Discard Policy New
? Stream Messages Limit -1
? Per Subject Messages Limit 1
? Total Stream Size -1
? Message TTL -1
? Max Message Size -1
? Duplicate tracking time window 2m0s
? Allow message Roll-ups Yes
? Allow message deletion No
? Allow purging subjects or the entire stream Yes
Stream KV_foobar was created

Information for Stream KV_foobar created 2023-07-26 12:34:43

             Subjects: $KV.foobar.>
             Replicas: 1
              Storage: File

Options:

            Retention: Limits
     Acknowledgements: true
       Discard Policy: New
     Duplicate Window: 2m0s
    Allows Msg Delete: false
         Allows Purge: true
       Allows Rollups: true

Limits:

     Maximum Messages: unlimited
  Maximum Per Subject: 1
        Maximum Bytes: unlimited
          Maximum Age: unlimited
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 0
              LastSeq: 0
     Active Consumers: 0
```

```
nats stream ls -a
╭────────────────────────────────────────────────────────────────────────────────╮
│                                    Streams                                     │
├───────────┬─────────────┬─────────────────────┬──────────┬──────┬──────────────┤
│ Name      │ Description │ Created             │ Messages │ Size │ Last Message │
├───────────┼─────────────┼─────────────────────┼──────────┼──────┼──────────────┤
│ KV_foobar │             │ 2023-07-26 12:24:23 │ 0        │ 0 B  │ never        │
╰───────────┴─────────────┴─────────────────────┴──────────┴──────┴──────────────╯
```

```
nats kv add foobar
nats: error: nats: stream name already in use
```

```
nats kv del foobar
? Deleted bucket foobar? Yes

nats kv add foobar
Information for Key-Value Store Bucket foobar created 2023-07-26T12:31:35-06:00

Configuration:

          Bucket Name: foobar
         History Kept: 1
        Values Stored: 0
   Backing Store Kind: JetStream
          Bucket Size: 0 B
  Maximum Bucket Size: unlimited
   Maximum Value Size: unlimited
     JetStream Stream: KV_foobar
              Storage: File

  Cluster Information:

                Name:
              Leader: NA3ASHNOFP6SXS2ZVBTBLAXGFF2YJS3FFS2VZFZ7TZR6B5LENER3RK7H

nats stream info KV_foobar
Information for Stream KV_foobar created 2023-07-26 12:35:18

             Subjects: $KV.foobar.>
             Replicas: 1
              Storage: File

Options:

            Retention: Limits
     Acknowledgements: true
       Discard Policy: New
     Duplicate Window: 2m0s
           Direct Get: true
    Allows Msg Delete: false
         Allows Purge: true
       Allows Rollups: true

Limits:

     Maximum Messages: unlimited
  Maximum Per Subject: 1
        Maximum Bytes: unlimited
          Maximum Age: unlimited
 Maximum Message Size: unlimited
    Maximum Consumers: unlimited


State:

             Messages: 0
                Bytes: 0 B
             FirstSeq: 0
              LastSeq: 0
     Active Consumers: 0
```

The only difference between the two is `Direct Get: true`, which we don't care about.

KV buckets _are_ streams under the hood. The OTP code was creating a stream IIRC because the jetstream client we used in Elixir didn't have support for an option we wanted; something like that. I can confirm creating a bucket works as intended because there are wasmCloud lattices I've pre-provisioned using the bucket setup, both with the CLI and with async-nats.

2) If you try to create a stream that already exists, the server will return an error:

```
nats stream add KV_foobar
? Subjects $KV.foobar.>
? Storage file
? Replication 1
? Retention Policy Limits
? Discard Policy New
? Stream Messages Limit -1
? Per Subject Messages Limit -1
? Total Stream Size -1
? Message TTL -1
? Max Message Size -1
? Duplicate tracking time window 2m0s
? Allow message Roll-ups No
? Allow message deletion Yes
? Allow purging subjects or the entire stream Yes
nats: error: could not create Stream: stream name already in use with a different configuration (10058)
```

Unless you use the exact same configuration as what already exists, in which case it is a no-op. To be safe, we should check for the existence of the bucket first before trying to create it.

3) This can wait, yes. Are you tracking TODOs somewhere?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1431,4 +1606,151 @@ impl Lattice {
...
+        let mut key_parts = key.split('_');
+        #[allow(clippy::single_match_else)]
+        let res = match (operation, key_parts.next(), key_parts.next()) {
+            (Operation::Put, Some("LINKDEF"), Some(id)) => {
+                self.process_linkdef_put(id, value).await
+            }
+            (Operation::Delete, Some("LINKDEF"), Some(id)) => {
+                self.process_linkdef_delete(id, value).await
+            }
+            _ => {
```
connorsmith256, line 1739: You'll want a TODO here for operations to put/delete `CLAIMS` and `REFMAP` entries as well

PR #397, opened 2023-07-26
feat: integrate `wasmcloud-provider-sdk`

### LINKED ISSUE #391: Integrate `wasmcloud-provider` in HTTP server provider
Once #387 is done, `wasmcloud-provider-httpsever` should be adapted to use that

rvolosatovs:
## Feature or Problem

This PR switches HTTP server provider to using `wasmcloud-provider-sdk`. Following merge of this PR, all subsequent providers added to the repo MUST be using `wasmcloud-provider-sdk` and *not* `wasmbus-rpc`

## Related Issues

Closes #391 

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


Cargo.toml
```diff
@@ -21,10 +21,24 @@ status = "actively-developed"
 
 [dependencies]
 anyhow = { workspace = true }
-clap = { workspace = true, features = ["color", "derive", "error-context", "help", "std", "suggestions", "usage"] }
+clap = { workspace = true, features = [
```
rvolosatovs, line 24: These changes seem out of scope for the PR

crates/providers/Cargo.toml
```diff
@@ -11,11 +11,7 @@ opt-level = "z"
 lto = true
 
 [workspace]
-members = [
-    "http-server",
-    "kv-redis",
-    "nats",
-]
+members = ["http-server", "kv-redis", "nats"]
```
rvolosatovs, line 14: This seems out of scope

Cargo.toml
thomastaylor312: This is applied by my toml formatter, so it is likely best practice.

crates/providers/Cargo.toml
thomastaylor312: Same as above

Cargo.toml
thomastaylor312: Oh and generally a lot of us other maintainers followers of what I call the "campfire" or "campground" rule which is "leave it cleaner than you found it." This was mostly unintentional because it was automatic, but I don't think it is a problem to leave it cleaner than it was

rvolosatovs: This change makes it very difficult to lexicographically sort the imports and copy-paste this line

crates/providers/Cargo.toml
rvolosatovs: This will introduce merge conflicts on each change, e.g. if two providers are added in separate PRs, whereas that would not happen with the existing approach

rvolosatovs (commented):
Cannot approve PR I have filed myself, but LGTM

Cargo.toml
thomastaylor312: We can change back later if it becomes a thorn in our sides

connorsmith256 (approved):
LGTM

Cargo.toml
rvolosatovs: Sure, but do note that had you not made this change, there would be no merge conflict with #432 - this is primarily why I'm against out-of-scope changes in PRs and prefer PRs to be focused on doing one thing at a time - it lets us move faster with less friction

connorsmith256 (dismissed):

rvolosatovs (commented):
LGTM, good catch about the `STANDARD` vs `STANDARD_NO_PAD`, I chose the wrong engine looking at https://docs.rs/base64/latest/base64/fn.decode.html

connorsmith256 (approved):

Issue #417, opened 2023-07-31
Add support for pre-provisioned NATS buckets


rvolosatovs:
NATS buckets may be pre-provisioned and should be accepted even if the config is different https://github.com/wasmCloud/wasmCloud/pull/395#discussion_r1275130181

The nicest way to handle this should probably be not throwing an error on creation failure, but rather check if the underlying error code is https://docs.rs/async-nats/latest/async_nats/jetstream/struct.ErrorCode.html#associatedconstant.STREAM_NAME_EXIST, and if so - print an `INFO` log statement stating that the stream is reused with different config.
All other errors should fail initialization

PR #427, opened 2023-08-03
refactor: create bucket explicitly instead of stream

connorsmith256:
## Feature or Problem
This PR creates the lattice metadata bucket explicitly, rather than implicitly via a stream. It also gracefully handles errors where the bucket has already been provisioned with custom settings, allowing multiple hosts to run in the same pre-provisioned lattice

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/417

## Release Information
Next

## Consumer Impact
This allows hosts to start in a lattice with pre-provisioned resources, e.g. the lattice metadata bucket

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
The existing `wasmbus` test covers creating a bucket successfully. I locally edited the code to pre-create a bucket beforehand and verified the host gracefully handles the error from the NATS client. I also tested the error case by attempting to create a bucket with an invalid name. 

### Manual Verification
See above

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -670,9 +670,9 @@ async fn create_lattice_metadata_bucket(
                 info!("lattice metadata bucket {bucket} already exists. Skipping creation.");
                 Ok(())
             } else {
-                Err(anyhow!(
-                    "failed to create lattice metadata bucket {bucket}: {err}"
-                ))
+                Err(anyhow!(err).context(format!(
+                    "failed to create lattice metadata bucket '{bucket}'"
+                )))
             }
         }
     }
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -22,7 +22,7 @@ use std::process::Stdio;
 use std::sync::Arc;
 
 use anyhow::{anyhow, bail, ensure, Context as _};
-use async_nats::jetstream::kv;
+use async_nats::jetstream::{context::Context as JetstreamContext, kv};
 use async_trait::async_trait;
 use base64::engine::general_purpose::STANDARD_NO_PAD;
 use base64::Engine;
@@ -644,6 +644,40 @@ fn linkdef_hash(
     hex::encode_upper(hash.finalize())
 }
 
+#[instrument(skip(jetstream))]
+async fn create_lattice_metadata_bucket(
+    jetstream: &JetstreamContext,
+    bucket: &str,
+) -> anyhow::Result<()> {
+    match jetstream
+        .create_key_value(kv::Config {
+            bucket: bucket.to_string(),
+            ..Default::default()
+        })
+        .await
+    {
+        Ok(_) => {
+            info!("created lattice metadata bucket {bucket} with 1 replica");
+            Ok(())
+        }
+        Err(err) => {
+            // Ignore the error if the bucket was pre-provisioned
+            // This is a terrible hack, but the error we get from the NATS client is a nested
+            // dyn Error. The NATS server will return an error containing the code
+            // 10058 if the underlying stream already exists with different configuration,
+            // which is the case we care about
+            if err.to_string().contains("10058") {
+                info!("lattice metadata bucket {bucket} already exists. Skipping creation.");
+                Ok(())
+            } else {
+                Err(anyhow!(
+                    "failed to create lattice metadata bucket {bucket}: {err}"
+                ))
+            }
+        }
+    }
+}
+
 impl Lattice {
     const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(30);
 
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -369,6 +369,40 @@ fn linkdef_hash(
...
+            // This is a terrible hack, but the error we get from the NATS client is a nested
+            // dyn Error. The NATS server will return an error containing the code
+            // 10058 if the underlying stream already exists with different configuration,
+            // which is the case we care about
+            if err.to_string().contains("10058") {
+                info!("lattice metadata bucket {bucket} already exists. Skipping creation.");
+                Ok(())
+            } else {
+                Err(anyhow!(
+                    "failed to create lattice metadata bucket {bucket}: {err}"
```
autodidaddict, line 399: I love the interpolation syntax. It seems crazy that it took rust so long to get it.

autodidaddict (approved):

crates/host/src/wasmbus/mod.rs
```diff
@@ -369,6 +369,40 @@ fn linkdef_hash(
...
+            info!("created lattice metadata bucket {bucket} with 1 replica");
+            Ok(())
+        }
+        Err(err) => {
+            // Ignore the error if the bucket was pre-provisioned
+            // This is a terrible hack, but the error we get from the NATS client is a nested
+            // dyn Error. The NATS server will return an error containing the code
+            // 10058 if the underlying stream already exists with different configuration,
+            // which is the case we care about
+            if err.to_string().contains("10058") {
```
rvolosatovs, line 394: Can't we just downcast?

crates/host/src/wasmbus/mod.rs
```diff
@@ -22,7 +22,7 @@ use std::process::Stdio;
 use std::sync::Arc;
 
 use anyhow::{anyhow, bail, ensure, Context as _};
-use async_nats::jetstream::kv;
+use async_nats::jetstream::{context, kv};
```
rvolosatovs, line 25: nit; can we just use `jetstream::context` directly please? The `context` namespace is way too ambiguous

Once we refactor this file into multiple smaller chunks, this would make sense, but for now it's way too opaque

crates/host/src/wasmbus/mod.rs
```diff
@@ -369,6 +369,40 @@ fn linkdef_hash(
...
+            // Ignore the error if the bucket was pre-provisioned
+            // This is a terrible hack, but the error we get from the NATS client is a nested
+            // dyn Error. The NATS server will return an error containing the code
+            // 10058 if the underlying stream already exists with different configuration,
+            // which is the case we care about
+            if err.to_string().contains("10058") {
+                info!("lattice metadata bucket {bucket} already exists. Skipping creation.");
+                Ok(())
+            } else {
+                Err(anyhow!(
```
rvolosatovs, line 398:
```suggestion
                bail!(
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -369,6 +369,40 @@ fn linkdef_hash(
...
+                    "failed to create lattice metadata bucket {bucket}: {err}"
+                ))
```
rvolosatovs, line 400:
```suggestion
                Err(anyhow!(err).context(format!("failed to create lattice metadata bucket `{bucket}`)))
```

crates/host/src/wasmbus/mod.rs
rvolosatovs: Looking at the docs, you need to call https://docs.rs/async-nats/latest/async_nats/jetstream/context/struct.CreateKeyValueError.html#method.kind first to find out the kind of the error and then use https://docs.rs/async-nats/latest/async_nats/jetstream/context/struct.CreateKeyValueError.html#method.source to acquire the internal error, downcast it to https://docs.rs/async-nats/latest/async_nats/jetstream/struct.Error.html and check the code

rvolosatovs (changes requested):

crates/host/src/wasmbus/mod.rs
connorsmith256: I looked into this before opening the PR and talked it through with @thomastaylor312

Rather than chain those functions to extract the error, we decided to just flatten the error hierarchy to a string and search for the error code

rvolosatovs: Why? This clearly not the right approach, what's the motivation behind this?

connorsmith256: Because while flattening and searching for a substring is brittle, so is chaining and downcasting, and the latter is much harder to read

crates/host/src/wasmbus/mod.rs
connorsmith256: imported as `JetstreamContext`

crates/host/src/wasmbus/mod.rs
rvolosatovs: This is a very subjective comment and I disagree about the "brittleness" of downcasting and checking the exact error code, please expand on why exactly you think it's brittle?

crates/host/src/wasmbus/mod.rs
```diff
@@ -644,6 +644,38 @@ fn linkdef_hash(
...
+            // Ignore the error if the bucket was pre-provisioned
+            // This is a terrible hack, but the error we get from the NATS client is a nested
+            // dyn Error. The NATS server will return an error containing the code
+            // 10058 if the underlying stream already exists with different configuration,
+            // which is the case we care about
+            if err.to_string().contains("10058") {
+                info!("lattice metadata bucket {bucket} already exists. Skipping creation.");
+                Ok(())
+            } else {
+                Err(anyhow!(err).context("failed to create lattice metadata bucket {bucket}"))
```
rvolosatovs, line 673: `{bucket}` is not interpolated here - you should be using `format!` macro.
Also, please surround the variable with backticks.

crates/host/src/wasmbus/mod.rs
thomastaylor312: I walked through this with @connorsmith256 and I agree with the shortcut here. We have to downcast at least 2 (though I think it was 3) times, which involves a lot of option handling and is just a pain to read. Given that this was for a _single_ use case at the beginning of startup, this felt like a great corner to cut and was more readable. These NATS error codes are pretty stable so I don't worry about "cheating" here

crates/host/src/wasmbus/mod.rs
rvolosatovs, line 673: this is still not correct

connorsmith256: Forgot to push

rvolosatovs (approved):

PR #431, opened 2023-08-04
style: rename most instances of lattice and wasmbus to host

connorsmith256:
## Feature or Problem
There were several instances where we referred to "wasmbus" or "lattice" where I think "host" is the more appropriate term. I was getting confused so I replaced most of these with "host"

## Related Issues
N/A

## Release Information
Next

## Consumer Impact
N/A

## Testing
No additional testing performed

### DIFF
--- a/crates/host/src/wasmbus/config.rs
+++ b/crates/host/src/wasmbus/config.rs
@@ -2,8 +2,8 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
-/// Wasmbus lattice configuration
-pub struct Lattice {
+/// wasmCloud Host configuration
+pub struct Host {
     /// URL to connect to
     pub url: Url,
     /// Optional host seed
@@ -12,7 +12,7 @@ pub struct Lattice {
     pub cluster_seed: Option<String>,
 }
 
-impl Default for Lattice {
+impl Default for Host {
     fn default() -> Self {
         Self {
             url: Url::parse("nats://localhost:4222").expect("failed to parse URL"),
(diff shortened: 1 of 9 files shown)


rvolosatovs (dismissed):
This will have to wait until we're done with most of the functionality, since this will cause merge conflicts with work that is in-flight

connorsmith256:
Rebased off main and fixed the one merge conflict in `tests/wasmbus.rs`

crates/host/src/wasmbus/config.rs
```diff
@@ -2,8 +2,8 @@ use serde::{Deserialize, Serialize};
 use url::Url;
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
-/// Wasmbus lattice configuration
-pub struct Lattice {
+/// wasmCloud Host configuration
```
rvolosatovs, line 5: This should still mention wasmbus, since it's the wasmbus lattice protocol, that is implemented by this host

crates/host/src/wasmbus/mod.rs
```diff
@@ -1,7 +1,7 @@
-/// Wasmbus lattice configuration
+/// wasmCloud host configuration
```
rvolosatovs, line 1: This is `wasmcloud-host` crate, so "wasmCloud" here has no meaning. It's wasmbus protocol that is implemented as is important here

crates/host/src/wasmbus/mod.rs
```diff
@@ -609,9 +609,9 @@ struct Provider {
     image_ref: String,
 }
 
-/// Wasmbus lattice
+/// wasmCloud Host
```
rvolosatovs, line 612: same

thomastaylor312:
@rvolosatovs I agree with @connorsmith256 here. What these structs currently include are more than just what is needed for `wasmbus` stuff right now (it is configuration needed for a lot of host implementation details). Combined with the fact that this is _incredibly_ minor, I'd prefer not to block on it

thomastaylor312 (approved):

Issue #428, opened 2023-08-03
[Rust Host] - Support custom lattice prefixes


connorsmith256:
The Rust host currently uses `default` everywhere for lattice prefixes. This value should be able to be supplied via an env var with `WASMCLOUD_LATTICE_PREFIX`

PR #432, opened 2023-08-04
Custom lattice prefixes

connorsmith256:
## Feature or Problem
Note: this PR is based on top of https://github.com/wasmCloud/wasmCloud/pull/431

This PR adds support for setting a custom lattice prefix via args

I also moved a couple other settings into `HostConfig`. If we like this approach, I can easily add support for the rest of the host options.

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/428

## Release Information
Next

## Consumer Impact
Users can now start the host with a `-x foo-bar` and the `lattice_prefix` for incoming/outgoing messages will match

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
I updated the wasmbus test to use a custom prefix

### Manual Verification
I also tested manually with different prefixes and NATS hosts/ports

### DIFF
--- a/crates/host/src/wasmbus/config.rs
+++ b/crates/host/src/wasmbus/config.rs
@@ -4,18 +4,22 @@ use url::Url;
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 /// wasmCloud Host configuration
 pub struct Host {
-    /// URL to connect to
-    pub url: Url,
-    /// Optional host seed
+    /// NATS URL to connect to for control interface connection
+    pub ctl_nats_url: Url,
+    /// The lattice the host belongs to
+    pub lattice_prefix: String,
+    /// The seed key (a printable 256-bit Ed25519 private key) used by this host to generate its public key
     pub host_seed: Option<String>,
-    /// Optional cluster seed
+    /// The seed key (a printable 256-bit Ed25519 private key) used by this host to sign all invocations
     pub cluster_seed: Option<String>,
 }
 
 impl Default for Host {
     fn default() -> Self {
         Self {
-            url: Url::parse("nats://localhost:4222").expect("failed to parse URL"),
+            ctl_nats_url: Url::parse("nats://localhost:4222")
+                .expect("failed to parse control NATS URL"),
+            lattice_prefix: "default".to_string(),
             host_seed: None,
             cluster_seed: None,
         }
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -161,23 +161,19 @@ impl Queue {
     #[instrument]
     async fn new(
         nats: &async_nats::Client,
+        lattice_prefix: &str,
         cluster_key: &KeyPair,
         host_key: &KeyPair,
     ) -> anyhow::Result<Self> {
+        let host_id = host_key.public_key();
         let (registries, pings, links, queries, auction, commands, inventory) = try_join!(
-            nats.subscribe("wasmbus.ctl.default.registries.put".into()),
-            nats.subscribe("wasmbus.ctl.default.ping.hosts".into()),
-            nats.subscribe("wasmbus.ctl.default.linkdefs.*".into()),
-            nats.subscribe("wasmbus.ctl.default.get.*".into()),
-            nats.subscribe("wasmbus.ctl.default.auction.>".into()),
-            nats.subscribe(format!(
-                "wasmbus.ctl.default.cmd.{}.*",
-                host_key.public_key()
-            )),
-            nats.subscribe(format!(
-                "wasmbus.ctl.default.get.{}.inv",
-                host_key.public_key()
-            )),
+            nats.subscribe(format!("wasmbus.ctl.{lattice_prefix}.registries.put",)),
+            nats.subscribe(format!("wasmbus.ctl.{lattice_prefix}.ping.hosts",)),
+            nats.subscribe(format!("wasmbus.ctl.{lattice_prefix}.linkdefs.*",)),
+            nats.subscribe(format!("wasmbus.ctl.{lattice_prefix}.get.*",)),
+            nats.subscribe(format!("wasmbus.ctl.{lattice_prefix}.auction.>",)),
+            nats.subscribe(format!("wasmbus.ctl.{lattice_prefix}.cmd.{host_id}.*",)),
+            nats.subscribe(format!("wasmbus.ctl.{lattice_prefix}.get.{host_id}.inv",)),
         )
         .context("failed to subscribe to queues")?;
         Ok(Self {
@@ -313,6 +309,7 @@ struct InvocationResponse {
 #[derive(Clone, Debug)]
 struct Handler {
     nats: async_nats::Client,
+    lattice_prefix: String,
     cluster_key: Arc<KeyPair>,
     origin: WasmCloudEntity,
     interfaces: Arc<RwLock<HashMap<String, WasmCloudEntity>>>,
@@ -350,13 +347,12 @@ impl Handler {
         let invocation = self.invocation(operation, request).await?;
         let request =
             rmp_serde::to_vec_named(&invocation).context("failed to encode invocation")?;
+        let lattice_prefix = &self.lattice_prefix;
+        let provider_id = invocation.target.public_key;
         let res = self
             .nats
             .request(
-                format!(
-                    "wasmbus.rpc.default.{}.default",
-                    invocation.target.public_key
-                ),
+                format!("wasmbus.rpc.{lattice_prefix}.{provider_id}.default",),
                 request.into(),
             )
             .await
@@ -416,6 +412,7 @@ impl Bus for Handler {
         let (res_r, mut res_w) = socket_pair()?;
 
         let nats = self.nats.clone();
+        let lattice_prefix = self.lattice_prefix.clone();
         let provider_id = target.public_key.clone();
         let origin = self.origin.clone();
         let target = target.clone();
@@ -438,7 +435,7 @@ impl Bus for Handler {
                     .map_err(|e| e.to_string())?;
                 let res = nats
                     .request(
-                        format!("wasmbus.rpc.default.{provider_id}.default"),
+                        format!("wasmbus.rpc.{lattice_prefix}.{provider_id}.default"),
                         request.into(),
                     )
                     .await
@@ -838,6 +835,7 @@ pub struct Host {
     event_builder: EventBuilderV10,
     friendly_name: String,
     heartbeat: AbortHandle,
+    host_config: HostConfig,
     host_key: KeyPair,
     labels: HashMap<String, String>,
     nats: async_nats::Client,
@@ -904,14 +902,10 @@ impl Host {
     /// Construct a new [Host] returning a tuple of its [Arc] and an async shutdown function.
     #[instrument]
     pub async fn new(
-        HostConfig {
-            url,
-            cluster_seed,
-            host_seed,
-        }: HostConfig,
+        config: HostConfig,
     ) -> anyhow::Result<(Arc<Self>, impl Future<Output = anyhow::Result<()>>)> {
-        let cluster_key = if let Some(cluster_seed) = cluster_seed {
-            let kp = KeyPair::from_seed(&cluster_seed)
+        let cluster_key = if let Some(cluster_seed) = config.cluster_seed.as_ref() {
+            let kp = KeyPair::from_seed(cluster_seed)
                 .context("failed to construct key pair from seed")?;
             ensure!(kp.key_pair_type() == KeyPairType::Cluster);
             kp
@@ -919,9 +913,9 @@ impl Host {
             KeyPair::new(KeyPairType::Cluster)
         };
         let cluster_key = Arc::new(cluster_key);
-        let host_key = if let Some(host_seed) = host_seed {
+        let host_key = if let Some(host_seed) = config.host_seed.as_ref() {
             let kp =
-                KeyPair::from_seed(&host_seed).context("failed to construct key pair from seed")?;
+                KeyPair::from_seed(host_seed).context("failed to construct key pair from seed")?;
             ensure!(kp.key_pair_type() == KeyPairType::Server);
             kp
         } else {
@@ -948,11 +942,15 @@ impl Host {
             "version": env!("CARGO_PKG_VERSION"),
         });
 
-        let nats = async_nats::connect(url.as_str())
+        debug!(
+            ctl_nats_url = config.ctl_nats_url.as_str(),
+            "connecting to NATS control server"
+        );
+        let nats = async_nats::connect(config.ctl_nats_url.as_str())
             .await
-            .context("failed to connect to NATS")?;
+            .context("failed to connect to NATS control server")?;
 
-        let queue = Queue::new(&nats, &cluster_key, &host_key)
+        let queue = Queue::new(&nats, &config.lattice_prefix, &cluster_key, &host_key)
             .await
             .context("failed to initialize queue")?;
         nats.flush().await.context("failed to flush")?;
@@ -977,8 +975,7 @@ impl Host {
         let event_builder = EventBuilderV10::new().source(host_key.public_key());
 
         let jetstream = async_nats::jetstream::new(nats.clone());
-        // TODO: Use prefix
-        let bucket = format!("LATTICEDATA_{prefix}", prefix = "default");
+        let bucket = format!("LATTICEDATA_{}", config.lattice_prefix);
         create_lattice_metadata_bucket(&jetstream, &bucket).await?;
 
         let data = jetstream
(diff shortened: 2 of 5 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -159,22 +159,40 @@ impl Queue {
...
         let (registries, pings, links, queries, auction, commands, inventory) = try_join!(
-            nats.subscribe("wasmbus.ctl.default.registries.put".into()),
-            nats.subscribe("wasmbus.ctl.default.ping.hosts".into()),
-            nats.subscribe("wasmbus.ctl.default.linkdefs.*".into()),
-            nats.subscribe("wasmbus.ctl.default.get.*".into()),
-            nats.subscribe("wasmbus.ctl.default.auction.>".into()),
             nats.subscribe(format!(
-                "wasmbus.ctl.default.cmd.{}.*",
-                host_key.public_key()
+                "wasmbus.ctl.{prefix}.registries.put",
```
rvolosatovs, line 168:
```suggestion
                "wasmbus.ctl.{lattice_prefix}.registries.put",
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -159,22 +159,40 @@ impl Queue {
...
+            )),
+            nats.subscribe(format!(
+                "wasmbus.ctl.{prefix}.cmd.{host}.*",
+                prefix = lattice_prefix,
+                host = host_key.public_key()
+            )),
+            nats.subscribe(format!(
+                "wasmbus.ctl.{prefix}.get.{host}.inv",
+                prefix = lattice_prefix,
+                host = host_key.public_key()
```
rvolosatovs, line 195: This should have been called outside the macro

crates/host/src/wasmbus/mod.rs
```diff
@@ -334,12 +352,13 @@ struct Provider {
     image_ref: String,
 }
 
-/// Wasmbus lattice
+/// wasmCloud Host
```
rvolosatovs, line 355: Out of scope

crates/host/src/wasmbus/mod.rs
```diff
@@ -334,12 +352,13 @@ struct Provider {
...
 #[derive(Debug)]
-pub struct Lattice {
+pub struct Host {
     // TODO: Clean up actors after stop
     actors: RwLock<HashMap<String, Arc<Actor>>>,
     cluster_key: KeyPair,
+    config: HostConfig,
```
rvolosatovs, line 361: The config should have been parsed on initialization, only the relevant parts from it should be stored in the resulting struct

crates/host/src/wasmbus/mod.rs
```diff
@@ -369,27 +388,23 @@ fn linkdef_hash(
...
     pub async fn new(
-        LatticeConfig {
-            url,
-            cluster_seed,
-            host_seed,
-        }: LatticeConfig,
+        config: HostConfig,
     ) -> anyhow::Result<(Arc<Self>, impl Future<Output = anyhow::Result<()>>)> {
-        let cluster_key = if let Some(cluster_seed) = cluster_seed {
+        let cluster_key = if let Some(cluster_seed) = config.cluster_seed.clone() {
```
rvolosatovs, line 399: The `clone` is redundant

crates/host/src/wasmbus/mod.rs
```diff
@@ -369,27 +388,23 @@ fn linkdef_hash(
...
             let kp = KeyPair::from_seed(&cluster_seed)
                 .context("failed to construct key pair from seed")?;
             ensure!(kp.key_pair_type() == KeyPairType::Cluster);
             kp
         } else {
             KeyPair::new(KeyPairType::Cluster)
         };
-        let host_key = if let Some(host_seed) = host_seed {
+        let host_key = if let Some(host_seed) = config.host_seed.clone() {
```
rvolosatovs, line 407: Redundant `clone`

crates/host/src/wasmbus/mod.rs
```diff
@@ -447,8 +464,8 @@ impl Lattice {
         let event_builder = EventBuilderV10::new().source(host_key.public_key());
 
         let jetstream = async_nats::jetstream::new(nats.clone());
-        // TODO: Use prefix
-        let bucket = format!("LATTICEDATA_{prefix}", prefix = "default");
+        let lattice_prefix = config.lattice_prefix.clone();
```
rvolosatovs, line 467: Redundant `clone`

crates/host/src/wasmbus/mod.rs
```diff
@@ -1509,7 +1529,7 @@ impl Lattice {
           "ctl_host": "TODO",
           "prov_rpc_host": "TODO",
           "rpc_host": "TODO",
-          "lattice_prefix": "default",
+          "lattice_prefix": self.config.lattice_prefix,
```
rvolosatovs, line 1532: This should have just been a field on the struct

rvolosatovs:
Please drop the lattice -> host renaming commit, since it's out-of-scope for this PR and will cause merge conflicts with other work

tests/wasmbus.rs
```diff
@@ -49,9 +49,10 @@ async fn wasmbus() -> anyhow::Result<()> {
         )
         .init();
 
+    let nats_host = "localhost";
```
rvolosatovs, line 52: redundant, this is already available from https://docs.rs/url/latest/url/struct.Url.html#method.host

tests/wasmbus.rs
```diff
@@ -81,18 +82,21 @@ async fn wasmbus() -> anyhow::Result<()> {
...
-    let client = async_nats::connect_with_options(
+    let nats_client = async_nats::connect_with_options(
         nats_url.as_str(),
         async_nats::ConnectOptions::new().retry_on_initial_connect(),
     )
     .await
     .context("failed to connect to NATS")?;
-    let client = ClientBuilder::new(client)
+
+    let test_prefix = "test-prefix";
```
rvolosatovs, line 92: `const`

crates/host/src/wasmbus/mod.rs
connorsmith256: Same as above

crates/host/src/wasmbus/mod.rs
connorsmith256: Fixed

tests/wasmbus.rs
connorsmith256: I don't think I understand. Are you suggesting we call `Url::parse` first, then extract the host with `.host_str().unwrap()`?

connorsmith256:
@rvolosatovs I think I addressed all your review comments. I left a few reply comments for clarification. There's one comment I want to talk about here:

> The config should have been parsed on initialization, only the relevant parts from it should be stored in the resulting struct

I think we should do this, but not right now. Most of what I anticipate will be in `HostConfig` should end up in the `Host` struct, but probably not everything, and I'd rather punt on deciding which things should go where. So for now I'd rather keep those values contained in the `host_config` field, so it will be easy to refactor later

crates/host/src/wasmbus/config.rs
```diff
@@ -1,21 +1,27 @@
...
-use url::Url;
 
+#[allow(clippy::module_name_repetitions)]
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 /// wasmCloud Host configuration
 pub struct Host {
-    /// URL to connect to
-    pub url: Url,
-    /// Optional host seed
+    /// NATS server host to connect to
```
rvolosatovs, line 7: I think internally, this should not be split into "host" and "port", but be just a single URL.

Apart from simplifying usage internally, a more important aspect of this is the scheme - while I understand that in current OTP implementation it is not possible to force a particular scheme to be used, a developer should be able to e.g. enforce a connection to NATS via TLS, which can only be done by specifying the scheme

src/config.rs
```diff
@@ -0,0 +1,17 @@
+/// Default NATS server host
```
rvolosatovs, line 1: Is a separate module really needed for this?

Especially, since you're importing all of these

src/main.rs
```diff
@@ -4,15 +4,37 @@ use anyhow::{self, Context};
 use clap::Parser;
 use tokio::{select, signal};
 use tracing_subscriber::prelude::*;
-use wasmcloud_host::WasmbusHostConfig;
+
+mod config;
+
+#[allow(clippy::wildcard_imports)]
+use config::*;
+use wasmcloud_host::wasmbus::HostConfig;
```
rvolosatovs, line 12: Could you move this import back together with all the other imports?

src/main.rs
```diff
@@ -4,15 +4,37 @@ use anyhow::{self, Context};
 use clap::Parser;
 use tokio::{select, signal};
 use tracing_subscriber::prelude::*;
-use wasmcloud_host::WasmbusHostConfig;
+
+mod config;
```
rvolosatovs, line 8: nit; could you move this above the `use` block, that's so far the consistent style used throughout the codebase in this repository.

(It should be made a guideline)

rvolosatovs (commented):
This is looking much better now, thanks!

src/main.rs
```diff
@@ -4,15 +4,37 @@ use anyhow::{self, Context};
...
 #[derive(Debug, Parser)]
 #[command(version, about, long_about = None)]
-struct Args;
+
+struct Args {
+    /// NATS server host to connect to
+    #[clap(long = "nats-host", default_value = DEFAULT_NATS_HOST, env = ENV_NATS_HOST)]
+    pub nats_host: String,
+    /// NATS server port to connect to
+    #[clap(long = "nats-port", default_value = DEFAULT_NATS_PORT, env = ENV_NATS_PORT)]
```
rvolosatovs, line 22: What do we gain from indirection in all of the `DEFAULT_*` and `ENV*`? Why not define everything directly in the attribute?

tests/wasmbus.rs
rvolosatovs: Yes, for example. It's confusing to have two "sources of truth", which are likely to diverge over time as we work on this test resulting in very hard-to-debug issues

rvolosatovs: Well, ideally we would not need the `nats_host` at all https://github.com/wasmCloud/wasmCloud/pull/432#discussion_r1289650029

crates/host/src/wasmbus/config.rs
connorsmith256: Can we revisit this later? I mirrored the options we [use in wash](https://github.com/wasmCloud/wash/blob/b0e746be713d070b4400294ec401b87444bd5741/src/up/mod.rs#L94), and this will make it easier to maintain backwards compatibility with tooling that starts hosts. (Whether we use a single URL or a host + port, we'll need to refactor anyway to split that into 3 sets of args: `WASMCLOUD_CTL_*`, `WASMCLOUD_RPC_*`, and `WASMCLOUD_PROV_RPC_*`)

src/main.rs
connorsmith256: Sure

src/main.rs
connorsmith256: I followed the pattern we used in wash, and personally, I think it's easier to view the defaults and env vars as their own lists.

src/main.rs
rvolosatovs: @connorsmith256 I understand that there is a particular order in other repositories, in ordering in this repository is the following:

- pub use super...
- pub use crate...
- pub use core...
- pub use std...
- pub use everything-else (including other crates in the same workspace)

- use super...
- use crate...
- use core...
- use std...
- use everything-else (including other crates in the same workspace)

Let's please follow the format already used throughout the codebase in this repository and we can always refactor it later. There is no reason to introduce inconsistencies here.

As a part of the refactoring we should introduce a development guideline to avoid such situations in the future

src/config.rs
rvolosatovs: I would rather not try to prematurely optimize this and keep it simple for now

crates/host/src/wasmbus/config.rs
rvolosatovs: Well, we can, but that would mean we have to revert this particular change, since we already are using a URL here, which to me seems as unnecessary churn.

Note, that I'm not asking to change the binary arguments and/or environment variables, just keeping the type of the field on the struct the same as it already was.

crates/host/src/wasmbus/mod.rs
```diff
@@ -350,12 +347,13 @@ impl Handler {
         let invocation = self.invocation(operation, request).await?;
         let request =
             rmp_serde::to_vec_named(&invocation).context("failed to encode invocation")?;
+        let lattice_prefix = self.lattice_prefix.clone();
```
rvolosatovs, line 350:
```suggestion
        let lattice_prefix = &self.lattice_prefix;
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -416,6 +414,7 @@ impl Bus for Handler {
         let (res_r, mut res_w) = socket_pair()?;
 
         let nats = self.nats.clone();
+        let lattice_prefix = self.lattice_prefix.clone();
```
rvolosatovs, line 417:
```suggestion
        let lattice_prefix = &self.lattice_prefix;
```

tests/wasmbus.rs
rvolosatovs: We also use it here https://github.com/wasmCloud/wasmCloud/blob/77d663d3e1fd5590177ac8003a313a3edf29ab1f/tests/wasmbus.rs#L191 and https://github.com/wasmCloud/wasmCloud/blob/77d663d3e1fd5590177ac8003a313a3edf29ab1f/tests/wasmbus.rs#L391

If you'd address https://github.com/wasmCloud/wasmCloud/pull/432#discussion_r1289650029, we'd also use it there.

If you would like to define the host name - fine, but please use a constant for that then

crates/host/src/wasmbus/mod.rs
```diff
@@ -948,11 +944,13 @@ impl Host {
             "version": env!("CARGO_PKG_VERSION"),
         });
 
-        let nats = async_nats::connect(url.as_str())
+        let nats_url = format!("{}:{}", config.nats_host, config.nats_port);
```
rvolosatovs, line 947: We already had a rich type here, which was known to be valid, why would we introduce a worse approach here?

crates/host/src/wasmbus/config.rs
```diff
@@ -1,21 +1,27 @@
 use serde::{Deserialize, Serialize};
-use url::Url;
 
+#[allow(clippy::module_name_repetitions)]
```
rvolosatovs, line 3: is this really necessary? Could we add a comment here about why?

tests/wasmbus.rs
```diff
@@ -226,15 +231,18 @@ async fn wasmbus() -> anyhow::Result<()> {
...
+
     let (host, shutdown) = Host::new(HostConfig {
-        url: nats_url.clone(),
-        cluster_seed: Some(cluster_key.seed().unwrap()),
-        host_seed: Some(host_key.seed().unwrap()),
+        nats_host: nats_host.to_string(),
+        nats_port,
+        lattice_prefix: TEST_PREFIX.to_string(),
+        cluster_seed: cluster_key.seed().ok(),
+        host_seed: host_key.seed().ok(),
```
rvolosatovs, line 240: Can we revert this change? This should panic if the seed key is missing and not silently ignore it - it would introduce a very hard-to-debug issue if any of these keys was not a seed key

crates/host/src/wasmbus/mod.rs
rvolosatovs: This still does not look addressed, but I don't mind fixing that in a follow-up

src/main.rs
```diff
@@ -33,3 +68,15 @@ async fn main() -> anyhow::Result<()> {
     shutdown.await.context("failed to shutdown host")?;
     Ok(())
 }
+
+impl From<Args> for WasmbusHostConfig {
+    fn from(args: Args) -> Self {
+        Self {
+            ctl_nats_url: Url::parse(&format!("nats://{}:{}", args.nats_host, args.nats_port))
+                .expect("failed to parse ctl_nats_url"),
```
rvolosatovs, line 76: This should have happened in `main()`, but I don't mind fixing that up in a follow-up myself

rvolosatovs (approved):

crates/host/src/wasmbus/mod.rs
connorsmith256: Look below, the usage is in an `async move`. Try to compile this as a ref and you'll get a lifetime error

rvolosatovs: You're right, I've missed that part

Issue #421, opened 2023-08-01
Support standardized WASI worlds


brooksmtownsend:
The core WASI worlds, specifically [wasi-cloud](https://github.com/WebAssembly/wasi-cloud-core), are as follows: 
- [x] [wasi-keyvalue](https://github.com/WebAssembly/wasi-keyvalue)
	- [ ] https://github.com/wasmCloud/wasmCloud/issues/456
- [x] [wasi-messaging](https://github.com/WebAssembly/wasi-messaging)
	- [ ] https://github.com/wasmCloud/wasmCloud/issues/682
- [x] [wasi-http](https://github.com/WebAssembly/wasi-http)
	- [ ] #681
	- [x] #680
- [ ] [wasi-runtime-config](https://github.com/WebAssembly/wasi-runtime-config)
- [ ] [wasi-distributed-lock-service](https://github.com/WebAssembly/wasi-distributed-lock-service)
- [ ] [wasi-sql](https://github.com/WebAssembly/wasi-sql)
- [x] [wasi-blob-store](https://github.com/WebAssembly/wasi-blob-store)
	- [ ] https://github.com/wasmCloud/wasmCloud/issues/679
- [x] [wasi-sockets](https://github.com/WebAssembly/wasi-sockets)
	- [x] https://github.com/wasmCloud/wasmCloud/issues/683

This issue encapsulates supporting these interfaces as a part of the host runtime. The host doesn't have to support the implementations directly and should opt to send out messages over NATS to a capability provider instead.

rvolosatovs:
`wasi-logging` is already supported, rest are TODO

brooksmtownsend:
We do this by and large with our releases of 0.82 which included WASI 0.2 support, and 1.0 which will even better support these interfaces with wRPC

PR #449, opened 2023-08-08
feat: implement `wasi:keyvalue/readwrite`

rvolosatovs:
## Feature or Problem

implement `wasi:keyvalue/readwrite`

## Related Issues

#421 
#448 
https://github.com/WebAssembly/wasi-keyvalue/pull/18
Blocking https://github.com/wasmCloud/wasmCloud/pull/459

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`


crates/host/src/wasmbus/mod.rs
```diff
@@ -450,6 +450,108 @@ impl Bus for Handler {
...
+    ) -> anyhow::Result<(Box<dyn AsyncRead + Sync + Send + Unpin>, u64)> {
+        #[derive(Deserialize)]
+        struct GetResponse {
+            #[serde(default)]
+            value: String,
+            #[serde(default)]
+            exists: bool,
+        }
+
+        const METHOD: &str = "wasmcloud:keyvalue/KeyValue.Get";
```
thomastaylor312, line 469: Just a heads up that currently the methods are not prefaced with the contract name as it is part of the linkdef. So this should just be `KeyValue.Get`

rvolosatovs: Methods may not contain the package specification in current (legacy) wasmbus invocation scheme, but that's not relevant here, implementation requires the "fully qualified" method name, which must include the package specification by definition.

Currently, links are established per-package, but that will have to change to per-interface links for component model support

thomastaylor312: Ok that makes sense. We should just make sure to update @vados-cosmonic's codegen work before it lands to make sure it does handle the fully qualified messages because I know my hand rolled coded didn't do that. Also, did we capture that change in wasmbus in an issue or quick RFC? Want to make sure that is documented as an implementation detail in our docs page eventually

rvolosatovs: There's no change in wasmbus protocol yet, this is all internal to the host.

_eventually_ we'll have to make changes, I was planning to file an issue today

rvolosatovs: https://github.com/wasmCloud/wasmCloud/issues/458

tests/wasmbus.rs
```diff
@@ -281,40 +360,35 @@ async fn wasmbus() -> anyhow::Result<()> {
...
-            &nats_provider_key.public_key(),
+            format!(r#"{{"address":"[{}]:{http_port}"}}"#, Ipv6Addr::UNSPECIFIED),
+        ),
+        assert_advertise_link(
+            &client,
+            &actor_claims,
+            &kvredis_provider_key,
+            "wasmcloud:keyvalue",
+            format!(r#"{{"url":"{redis_url}"}}"#),
+        ),
```
brooksmtownsend, line 382: I think, according to the provider code, `url` should be `URL` to properly parse as the KvRedisConfig

brooksmtownsend (approved):

Issue #367, opened 2023-07-13
Handle `wasmbus` requests concurrently


rvolosatovs:
Curently, all `wasmbus` lattice requests are handled sequentially, which eliminates possibility of race conditions and weird behaviors caused by out-of-order request handling. This should be parallelized, but the decision making should happen on per-request basis.
For the most trivial example, no other requests should be handled after `host stop` command is received, most likely, in fact, all other requests should finish processing before that command is executed.

brooksmtownsend:
Partially addressed by #716. Not handling requests after `host stop` isn't implemented but concurrent handling is

connorsmith256:
@rvolosatovs @brooksmtownsend thoughts on closing this one and opening a new issue to track the host not handling requests after a `stop` message? The current title feels outdated

brooksmtownsend:
I'm a fan of another issue that captures the desire to not accept more requests after host stop and finish other requests that we're currently executing when host stop is handled

PR #451, opened 2023-08-09
feat(wasmbus): handle launch commands concurrently

rvolosatovs:
## Feature or Problem

handle launch commands concurrently

## Related Issues

#367 - it could potentially close it, but we still need to do more research about what should/could be parallelized 
This is a blocker for #449 , since without this change starting three providers is just too slow

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1343,20 +1343,15 @@ impl Host {
         Ok(SUCCESS.into())
     }
 
-    #[instrument(skip(self, payload))]
-    async fn handle_launch_actor(
+    #[instrument(skip(self))]
+    async fn handle_launch_actor_task(
         &self,
-        payload: impl AsRef<[u8]>,
+        actor_ref: String,
+        annotations: Option<HashMap<String, String>>,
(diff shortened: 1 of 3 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1400,6 +1410,40 @@ impl Host {
                 instances.entry(annotations).or_default().append(&mut delta);
             }
         }
+        Ok(())
+    }
+
+    #[instrument(skip(self, payload))]
+    async fn handle_launch_actor(
+        self: Arc<Self>,
```
brooksmtownsend, line 1418: Arc<Self>... that's pretty neat.

brooksmtownsend (approved):

PR #466, opened 2023-08-13
fix(logs): allow debug and trace logs in release mode

connorsmith256:
## Feature or Problem
This removes a feature from the tracing crate, allowing higher log verbosity than info for release builds

Allowing users to specify debug/trace-level logs is important for debugging release builds. The size of the debug build is prohibitively large in some environments where viewing debug/trace logs is still helpful

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ clap = { workspace = true, features = [
     "usage",
 ] }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread", "signal"] }
-tracing = { workspace = true, features = ["release_max_level_info"] }
+tracing = { workspace = true } # TODO: revisit the 'release_max_level_info' feature https://github.com/wasmCloud/wasmCloud/issues/468
 tracing-subscriber = { workspace = true, features = [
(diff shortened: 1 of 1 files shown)


rvolosatovs (changes requested):
I am against this change, since I consider this a security vulnerability. This would cause potentially sensitive user data to be leaked unprotected in any way, which is unacceptable for a system that calls itself "secure".

Why not deploy a development profile build to a staging/development environment? Debugging does not belong in production.

rvolosatovs:
As discussed offline, I'm fine with allowing debug (preferable) or even trace level logs, as done here (last resort) in release builds as long as there's a clear TODO stating that this should be revisited (because at present trace every single Wasm call with full payloads cleartext are printed to trace) and an issue filed.

The issue should have "technical debt" and "security" labels (or appropriate alternatives) associated with it and have a link to this PR and discussion.

@connorsmith256

connorsmith256:
@rvolosatovs Issue captured in https://github.com/wasmCloud/wasmCloud/issues/468

rvolosatovs (approved):

PR #467, opened 2023-08-14
feat: add wit-bindgen for provider binaries

vados-cosmonic:
## Feature or Problem

Add `wit-bindgen-wasmcloud` and `wit-bindgen-wasmcloud-provider-binary` which enable generation of Provider boilerplate via proc-macro.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


Cargo.toml
```diff
@@ -130,3 +130,5 @@ wasmtime = { version = "11", default-features = false }
 wasmtime-wasi = { version = "11", default-features = false }
 wit-bindgen = { version = "0.9", default-features = false }
 wit-component = { version = "0.12", default-features = false }
+wit-bindgen-wasmcloud = { path = "./crates/wit-bindgen-wasmcloud" }
+wit-bindgen-wasmcloud-provider-binary = { path = "./crates/wit-bindgen-wasmcloud-provider-binary" }
```
rvolosatovs, line 134: could we keep this sorted?

crates/wit-bindgen-wasmcloud-provider-binary/.gitignore
```diff
@@ -0,0 +1,5 @@
+/target/
+target/
+.idea
+.direnv
+.envrc
```
rvolosatovs, line 5: Is there a reason to add this on a sub-directory level and not reuse the root one?

crates/wit-bindgen-wasmcloud-provider-binary/Cargo.lock
```diff
@@ -0,0 +1,396 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 3
```
rvolosatovs, line 3: Why is there a `Cargo.lock`, shouldn't this be a part of the workspace?

crates/wit-bindgen-wasmcloud-provider-binary/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+proc-macro2 = "1.0.66"
+syn = { version = "1.0.92", features = [ "parsing", "full", "visit-mut", "extra-traits" ] }
+wasm-metadata = "0.10.1"
+heck = "0.4.1"
+wasmtime-component-macro = { version = "11.0.1", features = [ "async" ] }
+wasmtime-wit-bindgen = "11.0.1"
+wit-parser = "0.8.0"
+serde = "1.0.183"
+anyhow = "1.0.72"
+quote = "1.0.32"
```
rvolosatovs, line 31: Please use workspace dependencies, like all other crates

crates/wit-bindgen-wasmcloud/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+pub mod provider {
+    pub mod binary {
+        pub use wit_bindgen_wasmcloud_provider_binary::generate;
+    }
+}
```
rvolosatovs, line 5: Could you help me understand why all the nesting here is necessary?

rvolosatovs (commented):
Could you add a provider utilizing the binding generator as part of this PR?

crates/wit-bindgen-wasmcloud-provider-binary/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+description = """
+Bindgen macro for binary-based wasmCloud capability providers written in Rust
+"""
+
+[lib]
+proc-macro = true
+
+[features]
+default = []
+debug = []
```
rvolosatovs, line 19: What does this do?

vados-cosmonic: old code here, this feature was removed, will remove

crates/wit-bindgen-wasmcloud/src/lib.rs
vados-cosmonic: Since there are different types of providers (rust-only, wasm guest, wasm host if we ever do that) I prefer `wit-bindgen-wasmcloud::provider::<type>::generate`

crates/wit-bindgen-wasmcloud-provider-binary/Cargo.lock
vados-cosmonic: - [ ] Remove Cargo.lock

Cargo.toml
vados-cosmonic: - [ ] sort `wit-*` deps

vados-cosmonic:
@rvolosatovs a provider using the bindgen will have to be added later -- neither the messaging or kvredis providers that are in the repo right now match the WIT interfaces (stuff like the `Transport` usage and `RpcResult` will be gone, from what I have heard).

@thomastaylor312 is working on rewriting the messaging stuff to match the provider SDK, but even then it won't use the bindgen right away. 

Messaging is the first target for using the bindgen, then kvredis -- and they're both very different now/not maching the WIT.

crates/wit-bindgen-wasmcloud/src/lib.rs
thomastaylor312: @vados-cosmonic I think we're only going to ever have binary provider gen stuff until we move to normal wasi providers, which will use the open source toolchain (with some of the modifications or similar to what you added). Let's just call this `wit_bindgen_wasmcloud_provider` and not have any nesting for now. If we do end up with more stuff we can nest then

vados-cosmonic: OK, sure -- will change this to `wit_bindgen_wasmcloud_provider`, and remove the second project completely
- [ ] rename to `wit_bindgen_wasmcloud_provider`

vados-cosmonic:
Oh also note that an example of the original messaging provider built with this bindgen can be found here:

https://github.com/vados-cosmonic/wasmcloud-provider-bindgen-test/tree/main/messaging-binary

crates/wit-bindgen-wasmcloud-provider/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+anyhow = { workspace = true }
+heck = { workspace = true }
+proc-macro2 = { workspace = true }
+quote = { workspace = true }
+serde = { workspace = true }
+syn = { version = "1.0.92", features = [ "parsing", "full", "visit-mut", "extra-traits" ] }
+wasm-metadata = "0.10.1"
+wasmtime-component-macro = { version = "11.0.1", features = [ "async" ] }
+wasmtime-wit-bindgen = "11.0.1"
+wit-parser = "0.8.0"
```
rvolosatovs, line 27: These should have been workspace dependencies. I'm fine with fixing that later, but could you add a `TODO` here to fix that? Preferably, with a reference to an issue (which you could maybe file?)

Cargo.toml
```diff
@@ -129,4 +129,7 @@ wasmparser = { version = "0.110", default-features = false }
 wasmtime = { version = "11", default-features = false }
 wasmtime-wasi = { version = "11", default-features = false }
 wit-bindgen = { version = "0.9", default-features = false }
+wit-bindgen-wasmcloud-provider = { path = "./crates/wit-bindgen-wasmcloud-provider" }
 wit-component = { version = "0.12", default-features = false }
+heck = { version = "0.4.1" }
+proc-macro2 = { version = "1.0.66" }
```
rvolosatovs, line 135: All other deps in this file are (or, at least, should be) sorted lexicographically. If you choose not to do this now - that's fine, but could you add a TODO here then? These also should not blindly enable all default features, but only the features required, just like all other dependencies in this repository

rvolosatovs:
> @rvolosatovs a provider using the bindgen will have to be added later -- neither the messaging or kvredis providers that are in the repo right now match the WIT interfaces (stuff like the `Transport` usage and `RpcResult` will be gone, from what I have heard).
> 
> @thomastaylor312 is working on rewriting the messaging stuff to match the provider SDK, but even then it won't use the bindgen right away.
> 
> Messaging is the first target for using the bindgen, then kvredis -- and they're both very different now/not maching the WIT.

Sounds good, thanks for clarifying, could you file an issue for this?

crates/wit-bindgen-wasmcloud-provider/Cargo.toml
vados-cosmonic: Ah these deps aren't used by any other crates yet -- is the rule here that all deps need to be hoisted?

rvolosatovs: Currently, that's how it's done for maintenance - all versions are defined at a single location

Cargo.toml
vados-cosmonic: - [ ] remove default feature usage from deps

crates/wit-bindgen-wasmcloud-provider/Cargo.toml
vados-cosmonic: note also that the `syn` dep *cannot* use the top level one -- others can, though.

crates/wit-bindgen-wasmcloud-provider/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[lib]
+proc-macro = true
+
+[dependencies]
+anyhow = { workspace = true }
+heck = { workspace = true }
+proc-macro2 = { workspace = true }
+quote = { workspace = true }
+serde = { workspace = true }
+syn = { version = "1.0.92", features = [ "parsing", "full", "visit-mut", "extra-traits" ] }
```
rvolosatovs, line 23: Is there a reason to use a different version from `2`, which is already in the workspace? Could you add a comment here describing it, if there is one? (and file an issue to use a consistent version?)

rvolosatovs (approved):
Looking forward to see this used and tested. Approving to keep the momentum

crates/wit-bindgen-wasmcloud-provider/Cargo.toml
vados-cosmonic: Yes, there is a reason, otherwise the version would not be `1.0.92` -- will add comment here so curious people can tell at a glance.

- [ ] Add comment around `syn` v1.0.92` version

rvolosatovs (approved):

PR #476, opened 2023-08-15
ref(providers): Finishes refactoring all providers to use the provider SDK

thomastaylor312:
Each commit in this PR refactors the kvredis and nats providers to use the provider SDK. `wasmbus-rpc` is no longer in the dep tree for providers


crates/providers/nats/Cargo.toml
```diff
@@ -15,11 +15,14 @@ async-trait = { workspace = true }
 base64 = { workspace = true }
 bytes = { workspace = true }
 futures = { workspace = true }
+lazy_static = "1"
```
rvolosatovs, line 18: Please use workspace dependencies. (`crates/providers`) contain their own workspace.

This particular dependency should have a TODO about replacing it with `once_cell`

crates/providers/nats/src/lib.rs
```diff
@@ -1,2 +1,3 @@
 // TODO: These types should be defined via WIT
+pub mod otel;
```
rvolosatovs, line 2: The comment above does not correspond to the right module anymore

rvolosatovs (approved):
It passes tests and I'm fine with cleaning it up later

crates/providers/nats/Cargo.toml
thomastaylor312: This is actually where I am curious. I've never done a Rust monorepo (just one with a few crates). I had understood workspace dependencies to be the ones shared across all crates and the ones in here are specific to this provider. Is the proper procedure to put _all_ of them at the top level?

crates/providers/nats/src/lib.rs
thomastaylor312: Oh whoops good catch

crates/providers/nats/Cargo.toml
thomastaylor312: I lifted them all, but still curious about my question

crates/providers/kv-redis/src/bin/kvredis.rs
```diff
@@ -346,35 +319,186 @@ impl KvRedisProvider {
...
+fn get_redis_url(link_values: &[(String, String)], default_connect_url: &str) -> String {
     link_values
         .iter()
         .find(|(key, _value)| key.eq_ignore_ascii_case(REDIS_URL_KEY))
         .map(|(_key, url)| url.to_owned())
         .unwrap_or_else(|| default_connect_url.to_owned())
 }
 
+#[async_trait::async_trait]
+impl wasmcloud_provider_sdk::MessageDispatch for KvRedisProvider {
```
connorsmith256, line 353: Does dispatch belong here? Not in the generated interface file?

crates/providers/nats/src/otel.rs
connorsmith256, line 1: Are these copy-pasted here for the NATS provider to support extracting/setting headers on application-level messages? If so, a comment explaining would help. I was surprised to see these here, since this is in the internals of the SDK

crates/providers/kv-redis/src/bin/kvredis.rs
thomastaylor312: Yeah because generated interface in this case is in a separate "crate" (the lib part) than the actual provider impl and struct

crates/providers/nats/src/otel.rs
thomastaylor312: Yes they are. These are not in the SDK anymore because the provider SDK sets the tracing stuff on the Invocation now.

I did call this out in the comment

```
I am just putting it here since we no longer have
// this code in the SDK provider (which was being used previously)
```

Were you looking for more than that?

Iceber:
It may be better to be consistent in the directory structure of the providers.

```bash
$ tree .
.
├── http-server
│   ├── bin
│   │   └── httpserver.rs
│   └── src
│       ├── *
├── kv-redis
│   └── src
│       ├── bin
│       │   └── kvredis.rs
│       ├── lib.rs
│       └── wasmcloud_interface_keyvalue.rs
└── nats
    └── src
        ├── bin
        │   └── nats_messaging.rs
        ├── lib.rs
        └── wasmcloud_interface_messaging.rs

rvolosatovs (approved):

thomastaylor312:
@Iceber That is probably a good idea, but I think out of scope for this PR. I figure when we move these all the way over to the wasi-cloud contracts, we'll take the time to restructure as much as possible

crates/providers/nats/src/otel.rs
connorsmith256: Nope, missed this somehow 🤦 Disregard!

Issue #483, opened 2023-08-15
Implement invocation verification


brooksmtownsend:
Invocations should be verified before they are handled, namely (checkboxes for #528)
- [x] The cluster issuer of the invocation should be in the host's cluster issuer list
- [x] The hash of the invocation should match
- [x] Actors initiating an invocation should have the proper capability claim to make that invocation
- [x] Actors receiving an invocation should have the proper capability claim to receive that invocation

In addition to the above requirements, we should reach out to a policy server as well, which will be handled separately with #480

rvolosatovs:
Isn't this a duplicate of https://github.com/wasmCloud/wasmCloud/issues/447 ?

brooksmtownsend:
> Isn't this a duplicate of #447 ?

Not quite, will add more detail to specify

PR #490, opened 2023-08-16
Implement cluster issuers, log level, provider shutdown, js_domain, and OCI CLI vars

brooksmtownsend:
## Feature or Problem
This PR continues work towards OTP feature parity by using more of the CLI args provided. Specifically, this PR implements using:
1. `cluster_issuers` list by propagating the list as a part of host config. Will need to be utilized in #483 
2. `log_level` by parsing as a `tracing::Level` and passing to the `tracing_subscriber` setup
3. `provider_shutdown_delay` by adding a NATS request to gracefully shutdown the provider with this timeout
4. `allow_latest`, `allowed_insecure`, `oci_registry`, `oci_user`, and `oci_password` by adding an `OciOpts` struct that the `Host` holds as a field and will use for all OCI operations
5. `js_domain` by using the optional domain to create the Jetstream context when accessing the LATTICECACHE KV bucket.

There are still more args to implement, but I'm trying to keep the PRs more organized and only add a few features at a time.

## Related Issues
N/A

## Release Information
N/A

## Consumer Impact
Nothing really quite yet, but you can now run the host with a variety of options 

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
- Validated that log level is parsed properly and affects emitted logs
- Validated that the provider shutdown delay is respected and the message publishes
- Validated that JS domain does properly look for KV buckets in the correct domain (required check for bucket existence in cases where creation of buckets is blocked at the NATS account level)
- Validated that allow_latest, and the 3-tuple of oci_registry, oci_user, and oci_password allow for authenticating to a registry

### DIFF
--- a/crates/host/src/lib.rs
+++ b/crates/host/src/lib.rs
@@ -77,14 +77,16 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
 
 /// Fetch an actor from a reference.
 #[instrument(skip(actor_ref))]
-pub async fn fetch_actor(actor_ref: impl AsRef<str>) -> anyhow::Result<Vec<u8>> {
+pub async fn fetch_actor(
+    actor_ref: impl AsRef<str>,
+    oci_opts: &oci::Config,
+) -> anyhow::Result<Vec<u8>> {
     match ResourceRef::try_from(actor_ref.as_ref())? {
         ResourceRef::File(actor_ref) => fs::read(actor_ref).await.context("failed to read actor"),
         ResourceRef::Bindle(actor_ref) => crate::bindle::fetch_actor(None, &actor_ref)
             .await
             .with_context(|| format!("failed to fetch actor under Bindle reference `{actor_ref}`")),
-        // TODO: Set config
-        ResourceRef::Oci(actor_ref) => crate::oci::fetch_actor(None, &actor_ref, true, vec![])
+        ResourceRef::Oci(actor_ref) => crate::oci::fetch_actor(None, &actor_ref, oci_opts)
             .await
             .with_context(|| format!("failed to fetch actor under OCI reference `{actor_ref}`")),
     }
@@ -95,6 +97,7 @@ pub async fn fetch_actor(actor_ref: impl AsRef<str>) -> anyhow::Result<Vec<u8>> 
 pub async fn fetch_provider(
     provider_ref: impl AsRef<str>,
     link_name: impl AsRef<str>,
+    oci_opts: &oci::Config,
 ) -> anyhow::Result<(PathBuf, jwt::Claims<jwt::CapabilityProvider>)> {
     match ResourceRef::try_from(provider_ref.as_ref())? {
         ResourceRef::File(provider_ref) => par::read(provider_ref, link_name)
@@ -107,9 +110,8 @@ pub async fn fetch_provider(
                     format!("failed to fetch provider under Bindle reference `{provider_ref}`")
                 })
         }
-        // TODO: Set config
         ResourceRef::Oci(provider_ref) => {
-            crate::oci::fetch_provider(None, &provider_ref, true, vec![], link_name)
+            crate::oci::fetch_provider(&provider_ref, link_name, None, oci_opts)
                 .await
                 .with_context(|| {
                     format!("failed to fetch provider under OCI reference `{provider_ref}`")
--- a/crates/host/src/oci.rs
+++ b/crates/host/src/oci.rs
@@ -6,45 +6,57 @@ use crate::par;
 use core::str::FromStr;
 
 use std::collections::HashMap;
-use std::env::{temp_dir, var};
+use std::env::temp_dir;
 use std::path::{Path, PathBuf};
 
 use anyhow::{bail, Context as _};
 use oci_distribution::client::{ClientConfig, ClientProtocol, ImageData};
 use oci_distribution::secrets::RegistryAuth;
 use oci_distribution::{Client, Reference};
+use serde::{Deserialize, Serialize};
 use tokio::fs;
 use tokio::io::AsyncWriteExt;
 use wascap::jwt;
-
-const OCI_VAR_REGISTRY: &str = "OCI_REGISTRY";
-const OCI_VAR_USER: &str = "OCI_REGISTRY_USER";
-const OCI_VAR_PASSWORD: &str = "OCI_REGISTRY_PASSWORD";
 
 const PROVIDER_ARCHIVE_MEDIA_TYPE: &str = "application/vnd.wasmcloud.provider.archive.layer.v1+par";
 const WASM_MEDIA_TYPE: &str = "application/vnd.module.wasm.content.layer.v1+wasm";
 const OCI_MEDIA_TYPE: &str = "application/vnd.oci.image.layer.v1.tar";
 
+/// Configuration options for OCI operations.
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub struct Config {
+    /// Whether or not to allow downloading OCI artifacts with the tag `latest`
+    pub allow_latest: bool,
+    /// A list of OCI registries that are allowed to be accessed over HTTP
+    pub allowed_insecure: Vec<String>,
+    /// Used in tandem with `oci_user` and `oci_password` to override credentials for a specific OCI registry.
+    pub oci_registry: Option<String>,
+    /// Username for the OCI registry specified by `oci_registry`.
+    pub oci_user: Option<String>,
+    /// Password for the OCI registry specified by `oci_registry`.
+    pub oci_password: Option<String>,
+}
+
 fn determine_auth(
     image_reference: &str,
     creds_override: Option<HashMap<String, String>>,
+    oci_config: &Config,
 ) -> RegistryAuth {
-    if let Some(hm) = creds_override {
-        match (hm.get("username"), hm.get("password")) {
+    match (creds_override, oci_config) {
+        (
+            _,
+            Config {
+                oci_registry: Some(reg),
+                oci_user: Some(u),
+                oci_password: Some(p),
+                ..
+            },
+        ) if image_reference.starts_with(reg) => RegistryAuth::Basic(u.clone(), p.clone()),
+        (Some(hm), _) => match (hm.get("username"), hm.get("password")) {
             (Some(un), Some(pw)) => RegistryAuth::Basic(un.to_string(), pw.to_string()),
             _ => RegistryAuth::Anonymous,
-        }
-    } else {
-        match (
-            var(OCI_VAR_REGISTRY),
-            var(OCI_VAR_USER),
-            var(OCI_VAR_PASSWORD),
-        ) {
-            (Ok(reg), Ok(u), Ok(p)) if image_reference.starts_with(&reg) => {
-                RegistryAuth::Basic(u, p)
-            }
-            _ => RegistryAuth::Anonymous,
-        }
+        },
+        _ => RegistryAuth::Anonymous,
     }
 }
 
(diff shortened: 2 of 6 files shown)


src/main.rs
```diff
@@ -197,41 +193,57 @@ struct Args {
...
         cluster_seed,
+        cluster_issuers,
+        provider_shutdown_delay,
+        allow_latest,
+        allowed_insecure,
+        oci_registry,
+        oci_user,
+        oci_password,
+        js_domain,
         ..
```
rvolosatovs, line 239: nit; if all the args are used now, can we remove the `..` to force a compile-time error when an unused arg is added?

crates/host/src/oci.rs
```diff
@@ -17,34 +18,32 @@ use tokio::fs;
...
 ) -> RegistryAuth {
-    if let Some(hm) = creds_override {
-        match (hm.get("username"), hm.get("password")) {
-            (Some(un), Some(pw)) => RegistryAuth::Basic(un.to_string(), pw.to_string()),
-            _ => RegistryAuth::Anonymous,
+    match (
+        host_oci_opts.oci_registry.as_ref(),
+        host_oci_opts.oci_user.as_ref(),
+        host_oci_opts.oci_password.as_ref(),
+    ) {
```
rvolosatovs, line 34: nit; you could match on `(creds_override, host_oci_opts)` tuple to avoid the `as_ref()` and the `is_some()` below

crates/host/src/wasmbus/mod.rs
```diff
@@ -2189,8 +2214,8 @@ impl Host {
           "uptime_seconds": uptime.as_secs(),
           "uptime_human": "TODO",
           "version": env!("CARGO_PKG_VERSION"),
-          "cluster_issuers": "TODO",
-          "js_domain": "TODO",
+          "cluster_issuers": self.host_config.cluster_issuers.clone().unwrap_or_else(|| vec![self.cluster_key.public_key()]).join(","),
```
rvolosatovs, line 2217: nit; could we split this into a separate `cluster_issuers` binding above?

rvolosatovs (approved):

crates/host/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub mod oci;
 mod par;
 
 pub use local::{Host as LocalHost, HostConfig as LocalHostConfig};
+use wasmbus::config::OciOptions;
```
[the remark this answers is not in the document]
rvolosatovs: Actually, looks like this should be in the `pub use` below as `WasmbusOciOptions`?

crates/host/src/oci.rs
```diff
@@ -2,11 +2,12 @@
 // https://github.com/wasmCloud/wasmcloud-otp/blob/5f13500646d9e077afa1fca67a3fe9c8df5f3381/host_core/native/hostcore_wasmcloud_native/src/oci.rs
 
 use crate::par;
+use crate::wasmbus::config::OciOptions;
```
rvolosatovs, line 5: Shouldn't there be a `Config` struct defined in this module instead? Why is it in `wasmbus::config`?

brooksmtownsend: Moved to `Config` struct in this module and refactored around this.

tests/wasmbus.rs
```diff
@@ -30,6 +30,7 @@ use wasmcloud_control_interface::{
     ActorAuctionAck, ActorDescription, ActorInstance, ClientBuilder, CtlOperationAck,
     Host as HostInfo, HostInventory, ProviderAuctionAck,
 };
+use wasmcloud_host::oci::Config as OciConfig;
```
rvolosatovs, line 33: nit; should probably just re-export as `OciConfig` from `wasmcloud_host` root

rvolosatovs (approved):

tests/wasmbus.rs
```diff
@@ -269,10 +274,10 @@ async fn wasmbus() -> anyhow::Result<()> {
             [],
         ) => {
             // TODO: Validate `issuer`
-            ensure!(cluster_issuers == Some("TODO".into()));
+            ensure!(cluster_issuers == Some(cluster_key.public_key()));
```
connorsmith256, line 277: Could you file an issue to test that hosts can accept invocations from another issuer?

crates/host/src/oci.rs
```diff
@@ -177,17 +186,15 @@ pub async fn fetch_actor(
 /// Returns an error if either fetching fails or reading the fetched OCI path fails
 #[allow(clippy::implicit_hasher)]
 pub async fn fetch_provider(
```
connorsmith256, line 188: Nit: `fetch_actor` uses `(creds, ref, opts)` for its arg order, but now `fetch_provider` uses `(ref, link, creds, opts)`. Can we be consistent?

crates/host/src/wasmbus/mod.rs
```diff
@@ -869,6 +869,12 @@ async fn create_lattice_metadata_bucket(
     jetstream: &JetstreamContext,
     bucket: &str,
 ) -> anyhow::Result<()> {
+    // Don't create the bucket if it already exists
```
connorsmith256, line 872: This technically introduces a race condition: consider the scenario of starting multiple hosts close to the same time/while another host/client is in the process of creating the bucket.

I think it's better to always try to create the bucket: if the settings are the same as what already exists, the NATS client will act like nothing bad happened. If the settings differ, the NATS client will return a relevant error

crates/host/src/wasmbus/mod.rs
```diff
@@ -1502,7 +1505,8 @@ impl Host {
         ) {
             (hash_map::Entry::Vacant(_), None) => {}
             (hash_map::Entry::Vacant(entry), Some(count)) => {
-                let actor = fetch_actor(&actor_ref)
+                // FIXME: avoid duplicate actor fetch
```
connorsmith256, line 1508: Can we link to an issue?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1977,14 +1982,34 @@ impl Host {
...
             return Ok(SUCCESS.into());
         };
         let provider = entry.get_mut();
         let instances = &mut provider.instances;
         if let hash_map::Entry::Occupied(entry) = instances.entry(link_name.clone()) {
             if entry.get().annotations == annotations {
                 let ProviderInstance { id, child, .. } = entry.remove();
+
+                // Send a request to the provider, requesting a graceful shutdown
+                if let Ok(payload) = serde_json::to_vec(&json!({ "host_id": host_id })) {
```
connorsmith256, line 1995: Since we know the contents of this payload and it's simple, we should probably assert we can serialize and bail if we can't

crates/host/src/wasmbus/mod.rs
```diff
@@ -1977,14 +1982,34 @@ impl Host {
...
+                                .payload(payload.into())
+                                .timeout(self.host_config.provider_shutdown_delay),
+                        )
+                        .await
+                    {
+                        warn!(?e, "Provider didn't gracefully shut down in time, shutting down forcefully");
+                    }
+                }
+
                 child.abort();
```
connorsmith256, line 2013: Should we wait for `provider_shutdown_delay` before sending this? Does an issue exist for that yet?

connorsmith256 (changes requested):
I didn't get my review in before auto-merge :/ No showstoppers here, but things we should file/clean up

tests/wasmbus.rs
brooksmtownsend: Added more information into #483, we still have to implement the actual invocation verification and it should be done

crates/host/src/oci.rs
brooksmtownsend: Unfortunate, I thought I made that change. Yes, I wanted it to be component information -> creds -> opts

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I chose to do it this way because even if we were to fail to send the message or serialize (which should be impossible) we definitely want to proceed to aborting the child process. For that reason I don't think we should bail, and this destructuring here is really just out of an abundance of caution

crates/host/src/wasmbus/mod.rs
brooksmtownsend: We only have to wait for `provider_shutdown_delay` in terms of waiting for a _reply_ from the capability provider, we don't always have to wait for that time if it only takes, say, 100ms for the provider to cleanup. So having it as the request timeout feels correct, unless I'm thinking about it wrong 🤔

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Will file

brooksmtownsend: https://github.com/wasmCloud/wasmCloud/issues/495

crates/host/src/wasmbus/mod.rs
brooksmtownsend: The only issue with trying to create the bucket always is if the ability to create a stream is blocked at a NATS account level then the error we get back is that Jetstream is unavailable. There's no way that I could see to tell the difference from not being allowed to create a stream vs not having Jetstream connectivity.

I do agree with the small race condition though, and wouldn't mind a better way to handle this.

tests/wasmbus.rs
connorsmith256: I think what I'm describing is a little different. #483 captures that the host needs to verify invocations. I was proposing we add the happy path to our tests, ie start two hosts with different cluster seeds but with shared issuers and confirm that invocations work across hosts

The motivation for this comes from doing some manual e2e testing during the weekend: I realized setting cluster issuers wasn't implemented yet when I ran into errors from a _provider_ refusing to process an invocation, because it hadn't been given a list of valid issuers by the host

crates/host/src/wasmbus/mod.rs
connorsmith256: Fair enough

crates/host/src/wasmbus/mod.rs
connorsmith256: Ah okay. I wasn't sure if it was possible for a provider to immediately ack-ed the shutdown message before actually cleaning up its resources/shutting down gracefully

crates/host/src/wasmbus/mod.rs
connorsmith256: Mm yeah good point. I think we'll have to accept this as a race condition then. The good news is even in most race conditions, multiple racing clients should be attempting to create the bucket with the same settings, so it shouldn't matter which wins. The only edge case is a scenario like:

1) client A (host) checks if the bucket exists -> false
2) client B creates the bucket with different settings (e.g. multiple replicas)
3) client A tries to create the bucket -> receives an error from the NATS client -> bails

In this case the recovery steps would be to restart the host, so it's not _that_ bad

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Hm, yeah I think it is possible for providers to do that but we wouldn't know if they did or not 🤔

crates/host/src/oci.rs
brooksmtownsend: Implemented in https://github.com/wasmCloud/wasmCloud/pull/496

tests/wasmbus.rs
brooksmtownsend: Implemented in https://github.com/wasmCloud/wasmCloud/pull/496

Issue #469, opened 2023-08-14
Link namespaces and actor-to-actor links


rvolosatovs:
Actors should be able to select a target of the upcoming interface method call, whether that's an actor or a provider.

Providers should be identified by their link name and actors - by their public key.

The suggestion is to:
1. Introduce a `set_target` method to our `wasmcloud:bus/host` interface, which would take a strongly-typed interface identifier and a target identifier as a parameter. Enums could be the simplest way forward here.
2. Add a target parameter to `wasmcloud:bus/host.call` method. This target would take an enum of either a provider link name string or actor public key

Note, that host will only be able to expose a finite set of interfaces, which is known upfront, any "custom" interface calls would go through `wasmcloud:bus/host.call`, which takes a target in itself. Therefore, the set of interfaces, for which a target can be selected can be represented with a strongly-typed enum, fully removing the need for host-side verification for of the interface being selected by `set_target` (e.g. `wasi:io/types` interface should not be selectable using this mechanism)

Refs https://github.com/wasmCloud/wasmCloud/issues/317

ricochet:
Today there are actor call aliases: https://github.com/wasmCloud/examples/tree/main/actor/actor-to-actor

In addition to targetting interfaces with the actor public key, we should allow for actor call aliases.

PR #493, opened 2023-08-16
feat: implement runtime link name selection and a2a calls

### LINKED ISSUE #317: [RFC] Separate Actor and Provider SDKs

### LINKED ISSUE #463: Multiple links per contract ID
Currently, the host only stores a single link per contract ID, it should be able to store multiple and take link name into account.
First, we should figure out how this would work in component model, refs https://github.com/wasmCloud/wasmCloud/issues/461#issuecomment-1673678571

### LINKED ISSUE #494: [Rust Host] Actor (module) -> provider invocations don't make it to the bus
From local debugging/print lines, the await on [read_to_end](https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L426) never resolves, so invocations don't make it onto `wasmbus.rpc`

Example to reproduce:
```
wash start actor wasmcloud.azurecr.io/kvcounter:0.4.2
wash start provider wasmcloud.azurecr.io/httpserver:0.19.1
wash start provider wasmcloud.azurecr.io/kvredis:0.22.0

wash link put MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M wasmcloud:httpserver address=0.0.0.0:1234

wash link put -x default -p 4222 MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E VAZVC4RX54J2NVCMCW7BPCAHGGG5XZXDBXFUMDUXGESTMQEJLC3YVZWB wasmcloud:keyvalue

curl 127.0.0.1:1234/api/counter
```

If you `nats sub wasmbus.rpc.>`, you'll see the actor module invocation on the bus, but won't see the provider call

rvolosatovs:
## Feature or Problem

Define the API for https://github.com/wasmCloud/wasmCloud/issues/469 and implement

## Related Issues

Closes https://github.com/wasmCloud/wasmCloud/issues/469
Closes #463 
Closes #317 
Closes #494 (please test)

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


thomastaylor312:
@rvolosatovs So I think this is close but there might have been some confusion when we last synced on this offline. I think it will be much cleaner if we keep the current `call` stuff exactly as it is, but expose a new interface that can included in the actor world.

```
// This is just a guess at names. Not strongly opinionated on interface or package name
package wasmcloud:lattice

interface lattice {
    variant actor-identifier {
        /// Actor public key
        public-key(string),
        /// Actor call alias
        alias(string),
    }

    // Note that this doesn't impose any restraints on the interface/contract name as we need to be able to support custom contracts
    set-link-name: func(interface: string, link-name: string)

    set-target: func(actor: actor-identifier)

    clear-link-name: func()

    clear-target: func()
}
```

Basically the host will provide that functionality and then we don't have to worry about passing interfaces or link names in the host call. What I envision the code looking like from an actor perspective is something like this:

```rust
set_link_name("wasmcloud:mycustom", "foobar");
// This will now be using the link name under the hood
custom::thing_i_call("blah");
clear_link_name();
```

An actor to actor call would be similar:

```rust
set_target(ActorIdentifier::CallAlias("foobar");
actor::thing_i_call("blah");
clear_target();
```

In the host implementation it would be something like this (very psuedocode-y, just trying to explain reasoning):

```rust
impl Lattice for Host {
    fn set_link_name(interface: String, link_name: String) {
        let current_interface = self.interfaces.get(interface);
        current_interface.link_name = link_name;
        self.interfaces.insert(interface, current_interface);
    }
}
    

/// Then when host.call comes
let current = self.interfaces.get(interface);
let inv = Invocation {
    link_name: current.link_name.unwrap_or("default")
    // All the rest of the stuff
}
```

Does that make any sense?

rvolosatovs:
> @rvolosatovs So I think this is close but there might have been some confusion when we last synced on this offline. I think it will be much cleaner if we keep the current `call` stuff exactly as it is, but expose a new interface that can included in the actor world.
> 
> ```
> // This is just a guess at names. Not strongly opinionated on interface or package name
> package wasmcloud:lattice
> 
> interface lattice {
>     variant actor-identifier {
>         /// Actor public key
>         public-key(string),
>         /// Actor call alias
>         alias(string),
>     }
> 
>     // Note that this doesn't impose any restraints on the interface/contract name as we need to be able to support custom contracts
>     set-link-name: func(interface: string, link-name: string)
> 
>     set-target: func(actor: actor-identifier)
> 
>     clear-link-name: func()
> 
>     clear-target: func()
> }
> ```
> 
> Basically the host will provide that functionality and then we don't have to worry about passing interfaces or link names in the host call. What I envision the code looking like from an actor perspective is something like this:
> 
> ```rust
> set_link_name("wasmcloud:mycustom", "foobar");
> // This will now be using the link name under the hood
> custom::thing_i_call("blah");
> clear_link_name();
> ```
> 
> An actor to actor call would be similar:
> 
> ```rust
> set_target(ActorIdentifier::CallAlias("foobar");
> actor::thing_i_call("blah");
> clear_target();
> ```
> 
> In the host implementation it would be something like this (very psuedocode-y, just trying to explain reasoning):
> 
> ```rust
> impl Lattice for Host {
>     fn set_link_name(interface: String, link_name: String) {
>         let current_interface = self.interfaces.get(interface);
>         current_interface.link_name = link_name;
>         self.interfaces.insert(interface, current_interface);
>     }
> }
>     
> 
> /// Then when host.call comes
> let current = self.interfaces.get(interface);
> let inv = Invocation {
>     link_name: current.link_name.unwrap_or("default")
>     // All the rest of the stuff
> }
> ```
> 
> Does that make any sense?

I disagree with your proposal, let's discuss it in detail offline tomorrow.

crates/actor/src/lib.rs
```diff
@@ -68,11 +68,30 @@ mod test {
...
-                wasi::io::streams::InputStream,
-                wasi::io::streams::OutputStream,
-            ) = wasmcloud::bus::host::call("mycompany:mypackage/interface.operation").unwrap();
+            {
+                let (_, _, _): (
+                    wasmcloud::bus::host::FutureResult,
+                    wasi::io::streams::InputStream,
+                    wasi::io::streams::OutputStream,
+                ) = wasmcloud::bus::host::call(None, "mycompany:mypackage/interface.operation")
+                    .unwrap();
```
brooksmtownsend, line 77: Flagging because I see `unwrap` but I'm fine with this if it's just temporary code.

crates/host/src/wasmbus/mod.rs
```diff
@@ -340,9 +340,14 @@ impl Handler {
     #[instrument(skip(operation, request))]
     async fn call_provider_with_payload(
         &self,
+        target: Option<bus::host::TargetEntity>,
         operation: impl AsRef<str>,
         request: Vec<u8>,
     ) -> anyhow::Result<Result<Vec<u8>, String>> {
+        if target.is_some() {
+            // TODO: Implement
```
brooksmtownsend, line 348: nit: would be good to mark this and other TODOs as `TODO(#469)`, some GitHub plugins can link that up with the PR and later we can search for all the relevant TODOs

brooksmtownsend (dismissed):

crates/actor/src/lib.rs
rvolosatovs: That's a test and this is never actually called

crates/runtime/wit/deps/cli/exit.wit
```diff
@@ -1,4 +1,4 @@
 interface exit {
-  /// Exit the current instance and any linked instances.
+  /// Exit the curerent instance and any linked instances.
```
thomastaylor312, line 2: Nit: Looks like a fat fingered key

thomastaylor312 (commented):
This is good to go from a code review perspective. Not giving the full approval so @connorsmith256 can give final LGTM once he tests that #494 is resolved

connorsmith256:
I haven't reviewed the updated code yet, but it looks like #494 is still happening :( I followed the steps to reproduce in the issue, and still don't see an outbound RPC call from the actor to the provider

tests/wasmbus.rs
```diff
@@ -390,24 +390,21 @@ expected: {expected_labels:?}"#
         .context("failed to parse `rust-httpserver` provider key")?;
     let httpserver_provider_url = Url::from_file_path(test_providers::RUST_HTTPSERVER)
         .expect("failed to construct provider ref");
-    let httpserver_provider_link_name = "link-name-1";
```
connorsmith256, line 393: Why switch to magic strings?

crates/host/src/wasmbus/mod.rs
```diff
@@ -264,12 +266,39 @@ impl Invocation {
...
         msg: Vec<u8>,
     ) -> anyhow::Result<Invocation> {
+        let operation = operation.into();
+        let (package, operation) = operation
+            .split_once('/')
+            .context("failed to parse operation")?;
+        let target = match target {
+            None => interfaces
+                .get(package)
+                .and_then(|links| links.get("default"))
```
connorsmith256, line 283: Consider a named constant, to disambiguate among references to default link names and default lattice prefixes

crates/host/src/wasmbus/mod.rs
```diff
@@ -2442,6 +2514,12 @@ impl Host {
         } = links
             .remove(id)
             .context("attempt to remove a non-existent link")?;
+        if let Some(actor) = self.actors.read().await.get(actor_id) {
+            let mut links = actor.handler.links.write().await;
+            if let Some(links) = links.get_mut(contract_id) {
+                links.remove(link_name);
+            }
+        }
```
connorsmith256, line 2522: Nice, thanks for fixing this

connorsmith256 (changes requested):
No big issues, only requesting changes until outbound invocations from modules are confirmed to work. I suggest we add to the wasmbus tests: start the kvcounter actor, increment via HTTP request, and confirm the data in redis is updated

crates/host/src/wasmbus/mod.rs
rvolosatovs: I don't really see what's there to disambiguate, `links` is keyed by link names, not lattice prefixes

tests/wasmbus.rs
rvolosatovs: For clarity, simplifying debugging and being explicit.
Note, that in doing so I've fixed an invalid link name specified in the link deletion.

crates/host/src/wasmbus/mod.rs
connorsmith256: What I mean is distinguishing between semantics of a common string (`"default"`) in the repo. Right now, if someone wants to find all the places we fall back on `default` for a link name, they have to search for this string, and then go through the mental effort of excluding uses where `default` was being used in the context of a lattice prefix. (I had to go through this effort a week ago when adding support for custom lattice prefixes and custom link names)

Swapping to named constants like `DEFAULT_LINK_NAME` and `DEFAULT_LATTICE_PREFIX` would remove any ambiguity

tests/wasmbus.rs
connorsmith256: Not going to block on this since it's a test, but my $0.02: copy-paste errors are copy-paste errors, regardless of whether they're a named var or a string. The advantage of a named var over a string is to prevent fat-finger typos, like passing `htpserver` as a link name when it should be `httpserver`. If it's a named variable, typos won't compile

rvolosatovs: Sure, the advantage of plain strings is maintenance, this string has to be duplicated in at least 5 different crates and indirection here just makes it more difficult to keep track of these. Looking at a single call in isolation should be enough to figure out what link name is used

connorsmith256: Did you mean to respond to the other thread? I'm not sure I follow the reference to multiple crates, since this is all in one test file

crates/runtime/wit/deps/cli/exit.wit
rvolosatovs: Indeed, but we should not fix a dependency

rvolosatovs:
@connorsmith256 I cannot reproduce #494 on this branch:
```
$ wash start actor wasmcloud.azurecr.io/kvcounter:0.4.2

Actor [MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E] (ref: [wasmcloud.azurecr.io/kvcounter:0.4.2]) started on host [NDCCV4YKFKHUMTRFVQ6ZU4TQRP7ZKO236TF5XICAQVB32RHHN6FMAQ6D]
$ wash start provider wasmcloud.azurecr.io/httpserver:0.19.1

Provider [VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M] (ref: [wasmcloud.azurecr.io/httpserver:0.19.1]) started on host [NDCCV4YKFKHUMTRFVQ6ZU4TQRP7ZKO236TF5XICAQVB32RHHN6FMAQ6D]
$ wash start provider wasmcloud.azurecr.io/kvredis:0.22.0

Provider [VAZVC4RX54J2NVCMCW7BPCAHGGG5XZXDBXFUMDUXGESTMQEJLC3YVZWB] (ref: [wasmcloud.azurecr.io/kvredis:0.22.0]) started on host [NDCCV4YKFKHUMTRFVQ6ZU4TQRP7ZKO236TF5XICAQVB32RHHN6FMAQ6D]
$ wash link put MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M wasmcloud:httpserver address=0.0.0.0:1234

Published link (MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E) <-> (VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M) successfully
$ wash link put -x default -p 4222 MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E VAZVC4RX54J2NVCMCW7BPCAHGGG5XZXDBXFUMDUXGESTMQEJLC3YVZWB wasmcloud:keyvalue

Published link (MCFMFDWFHGKELOXPCNCDXKK5OFLHBVEWRAOXR5JSQUD2TOFRE3DFPM7E) <-> (VAZVC4RX54J2NVCMCW7BPCAHGGG5XZXDBXFUMDUXGESTMQEJLC3YVZWB) successfully
$ curl 127.0.0.1:1234/api/counter
{"counter":1}
$ curl 127.0.0.1:1234/api/counter
{"counter":2}
$ curl 127.0.0.1:1234/api/counter
$ redis-cli
127.0.0.1:6379> keys *
1) "counter:default"
```

tests/wasmbus.rs
rvolosatovs: See the updated diff, names must match in 3 Wasm test binaries and in 2 tests - one you commented upon and a `wasmcloud_runtime`-level test

connorsmith256 (approved):
Love all the new test cases ❤️

Issue #491, opened 2023-08-16
[FEAT] Transfer friendly name generation parameters from OTP


brooksmtownsend:
I have no issue _specifically_ with the `names` crate, but I did just run a host that generated a friendly name of **dirty-women**, which of course is not acceptable. 

We should bring the adjectives and nouns over from https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/namegen.ex, and change the naming scheme with `names` to `Numbered` so we can retain the 4 digit number suffix.

PR #501, opened 2023-08-17
feat: generate host name based on a random number

Iceber:
## Feature or Problem

Generate a friendly name for the host based on a random number.
Names we pulled from a list of friendly or neutral adjectives and nouns suitable for use in public and on hosts/domain names

bring the adjectives and nouns over from https://github.com/wasmCloud/wasmcloud-otp/blob/main/host_core/lib/host_core/namegen.ex
## Related Issues

fix: https://github.com/wasmCloud/wasmCloud/issues/491
## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -895,6 +895,38 @@ async fn create_lattice_metadata_bucket(
 impl Host {
     const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(30);
 
+    const NAME_ADJECTIVES: &str = "
+    autumn hidden bitter misty silent empty dry dark summer
+    icy delicate quiet white cool spring winter patient
+    twilight dawn crimson wispy weathered blue billowing
+    broken cold damp falling frosty green long late lingering
+    bold little morning muddy old red rough still small
+    sparkling bouncing shy wandering withered wild black
+    young holy solitary fragrant aged snowy proud floral
+    restless divine polished ancient purple lively nameless
+    gray orange mauve
+    ";
+
+    const NAME_NOUNS: &str = "
+    waterfall river breeze moon rain wind sea morning
+    snow lake sunset pine shadow leaf dawn glitter forest
+    hill cloud meadow sun glade bird brook butterfly
+    bush dew dust field fire flower firefly ladybug feather grass
+    haze mountain night pond darkness snowflake silence
+    sound sky shape stapler surf thunder violet water wildflower
+    wave water resonance sun wood dream cherry tree fog autocorrect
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -895,6 +895,38 @@ async fn create_lattice_metadata_bucket(
...
+    haze mountain night pond darkness snowflake silence
+    sound sky shape stapler surf thunder violet water wildflower
+    wave water resonance sun wood dream cherry tree fog autocorrect
+    frost voice paper frog smoke star hamster ocean emoji robot
+    ";
+
+    /// Generate a friendly name for the host based on a random number.
+    /// Names we pulled from a list of friendly or neutral adjectives and nouns suitable for use in
+    /// public and on hosts/domain names
+    fn generate_friendly_name() -> Option<String> {
```
Iceber, line 924: I considered putting it in the new names.mod file, but it looks like it's only necessary for use here, so it's in Host.

crates/host/src/wasmbus/mod.rs
```diff
@@ -895,6 +895,38 @@ async fn create_lattice_metadata_bucket(
...
+    const NAME_ADJECTIVES: &str = "
+    autumn hidden bitter misty silent empty dry dark summer
+    icy delicate quiet white cool spring winter patient
+    twilight dawn crimson wispy weathered blue billowing
+    broken cold damp falling frosty green long late lingering
+    bold little morning muddy old red rough still small
+    sparkling bouncing shy wandering withered wild black
+    young holy solitary fragrant aged snowy proud floral
+    restless divine polished ancient purple lively nameless
+    gray orange mauve
```
Iceber, line 907: I didn't convert it to `&[&str]`, string look better maintained.
```rust
const NAME_ADJECTIVES: &[&str] = [
"autumn",
"hidden",
"bitter",
... // A very long list
]

rvolosatovs (approved):
Thank you for contribution!

crates/host/src/wasmbus/mod.rs
rvolosatovs: I would prefer an array here, but I understand the reasoning and we can always optimize later if we need to (we most probably will not)

Iceber: I tried the array and it is very long after formatting.:joy:

Issue #495, opened 2023-08-16
Avoid fetching actor bytes twice when scaling actor


brooksmtownsend:
This issue corresponds to this FIXME comment: https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L1508

When trying to get the actor ID, if we don't already know it we fetch the actor https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L1491 and then just a few lines later we fetch again https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L1509. Unless I'm missing a reason, we may as well avoid doing multiple fetches here.

PR #502, opened 2023-08-17
avoid fetching actor bytes twice when scaling actor

Iceber:
## Related Issues

fix: https://github.com/wasmCloud/wasmCloud/issues/495
## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1454,6 +1454,16 @@ impl Host {
         Ok(Some(buf.into()))
     }
 
+    #[instrument(skip(self))]
+    async fn fetch_actor(&self, actor_ref: &str) -> anyhow::Result<wasmcloud_runtime::Actor> {
+        let actor = fetch_actor(&actor_ref, &self.host_config.oci_opts)
+            .await
+            .context("failed to fetch actor")?;
+        let actor = wasmcloud_runtime::Actor::new(&self.runtime, actor)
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1454,6 +1454,16 @@ impl Host {
         Ok(Some(buf.into()))
     }
 
+    #[instrument(skip(self, payload))]
+    async fn fetch_actor(&self, actor_ref: &str) -> anyhow::Result<wasmcloud_runtime::Actor> {
```
rvolosatovs, line 1458: Could we avoid adding this helper for now? It's confusing to have `fetch_actor`, which calls `fetch_actor` and also initializes it.

There are only 2 places where we need this functionality and it's small enough that I think it's fine to inline

rvolosatovs (dismissed):
Thank you for contribution, this looks good, just a small organizational comment!

crates/host/src/wasmbus/mod.rs
Iceber: Sorry there were some other changes not committed, this function is used in several places

PR #506, opened 2023-08-17
Implement connection args for RPC/CTL/Prov RPC arguments

### LINKED ISSUE #478: Connect to NATS and utilize RPC, CTL, and PROV_RPC separated connections
The OTP host uses connection variables for NATS for the following parameters:
1. host
2. port
3. JWT
4. seed
5. timeout
6. TLS

We should use that as well, across our RPC, CTL, and PROV_RPC connections.

brooksmtownsend:
## Feature or Problem
This PR implements all of the connection options for RPC, CTL, and Provider RPC connections. Additionally, it adds support for the control interface topic prefix, which just defaults to `wasmbus.ctl` as before.

I made a few choices + stances in this PR:
1. RPC, CTL, and provider RPC connection arguments are _advanced_ arguments. Therefore, they are hidden by default. (I would support showing these arguments with some kind of --help --advanced flag)
2. RPC, CTL, and provider RPC connection args should fall back on the supplied NATS connection variables if not provided, instead of maintaining a separate all-around NATS connection. This way we're always internally using the correct connection, even if it's using the same parameters.

## Related Issues
Fixes #478 

## Release Information
N/A

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
Modified wasmbus test to use the new connection parameters.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1,7 +1,9 @@
 /// wasmCloud host configuration
 pub mod config;
+mod nats;
 
 pub use config::Host as HostConfig;
+use nats::connection_options;
 
 mod event;
 
@@ -985,20 +987,11 @@ impl Host {
             ctl_nats_url = config.ctl_nats_url.as_str(),
             "connecting to NATS control server"
         );
-        let ctl_nats = match (config.ctl_jwt.as_ref(), config.ctl_seed.as_ref()) {
-            (Some(jwt), Some(seed)) => {
-                let kp = std::sync::Arc::new(
-                    nkeys::KeyPair::from_seed(seed)
-                        .context("failed to construct key pair from seed")?,
-                );
-                async_nats::ConnectOptions::with_jwt(jwt.to_string(), move |nonce| {
-                    let key_pair = kp.clone();
-                    async move { key_pair.sign(&nonce).map_err(async_nats::AuthError::new) }
-                })
-            }
-            _ => async_nats::ConnectOptions::new(),
-        }
-        .require_tls(config.ctl_tls)
+        let ctl_nats = connection_options(
+            config.ctl_jwt.as_ref(),
+            config.ctl_seed.as_ref(),
+            config.ctl_tls,
+        )?
         .connect(config.ctl_nats_url.as_str())
         .await
         .context("failed to connect to NATS control server")?;
@@ -1018,20 +1011,11 @@ impl Host {
             rpc_nats_url = config.rpc_nats_url.as_str(),
             "connecting to NATS RPC server"
         );
-        let rpc_nats = match (config.rpc_jwt.as_ref(), config.rpc_seed.as_ref()) {
-            (Some(jwt), Some(seed)) => {
-                let kp = std::sync::Arc::new(
-                    nkeys::KeyPair::from_seed(seed)
-                        .context("failed to construct key pair from seed")?,
-                );
-                async_nats::ConnectOptions::with_jwt(jwt.to_string(), move |nonce| {
-                    let key_pair = kp.clone();
-                    async move { key_pair.sign(&nonce).map_err(async_nats::AuthError::new) }
-                })
-            }
-            _ => async_nats::ConnectOptions::new(),
-        }
-        .require_tls(config.rpc_tls)
+        let rpc_nats = connection_options(
+            config.rpc_jwt.as_ref(),
+            config.rpc_seed.as_ref(),
+            config.rpc_tls,
+        )?
         .request_timeout(Some(config.rpc_timeout))
         .connect(config.rpc_nats_url.as_str())
         .await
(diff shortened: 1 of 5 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -978,14 +985,79 @@ impl Host {
             ctl_nats_url = config.ctl_nats_url.as_str(),
             "connecting to NATS control server"
         );
-        let nats = async_nats::connect(config.ctl_nats_url.as_str())
-            .await
-            .context("failed to connect to NATS control server")?;
+        let ctl_nats = match (config.ctl_jwt.as_ref(), config.ctl_seed.as_ref()) {
```
connorsmith256, line 988: Since this logic is duplicated for ctl, rpc, and prov_rpc, can we extract to a function?

crates/host/src/wasmbus/mod.rs
```diff
@@ -840,7 +841,13 @@ pub struct Host {
...
     host_key: KeyPair,
     labels: HashMap<String, String>,
-    nats: async_nats::Client,
+    ctl_topic_prefix: String,
+    /// NATS client to use for control interface subscriptions and jetstream queries
+    ctl_nats: async_nats::Client,
+    /// NATS client to use for actor RPC calls
+    rpc_nats: async_nats::Client,
+    /// NATS client to use for communicating with capability providers
+    prov_rpc_nats: async_nats::Client,
```
connorsmith256, line 850: Nit: these (almost?) sorted before

connorsmith256 (changes requested):

crates/host/src/wasmbus/mod.rs
```diff
@@ -978,14 +985,79 @@ impl Host {
             ctl_nats_url = config.ctl_nats_url.as_str(),
             "connecting to NATS control server"
         );
-        let nats = async_nats::connect(config.ctl_nats_url.as_str())
-            .await
-            .context("failed to connect to NATS control server")?;
+        let ctl_nats = match (config.ctl_jwt.as_ref(), config.ctl_seed.as_ref()) {
+            (Some(jwt), Some(seed)) => {
+                let kp = std::sync::Arc::new(
```
rvolosatovs, line 990:
```suggestion
                let kp = Arc::new(
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -978,14 +985,79 @@ impl Host {
...
+                    nkeys::KeyPair::from_seed(seed)
```
rvolosatovs, line 991:
```suggestion
                    KeyPair::from_seed(seed)
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -978,14 +985,79 @@ impl Host {
             ctl_nats_url = config.ctl_nats_url.as_str(),
             "connecting to NATS control server"
         );
-        let nats = async_nats::connect(config.ctl_nats_url.as_str())
-            .await
-            .context("failed to connect to NATS control server")?;
+        let ctl_nats = match (config.ctl_jwt.as_ref(), config.ctl_seed.as_ref()) {
+            (Some(jwt), Some(seed)) => {
```
thomastaylor312, line 989: We should probably return an error if only one of these is set

crates/host/src/wasmbus/mod.rs
```diff
@@ -978,14 +985,79 @@ impl Host {
...
+        let queue = Queue::new(
+            &ctl_nats,
+            &config.ctl_topic_prefix,
+            &config.lattice_prefix,
+            &cluster_key,
+            &host_key,
+        )
+        .await
+        .context("failed to initialize queue")?;
+        ctl_nats.flush().await.context("failed to flush")?;
```
thomastaylor312, line 1015: Nit: Might want to note why we are flushing here for future maintainers and future selves (I know this was already here, but hence the nit)

crates/host/src/wasmbus/mod.rs
```diff
@@ -2275,7 +2349,14 @@ impl Host {
             ..
         }: async_nats::Message,
     ) {
-        let mut parts = subject.split('.').skip(3); // skip `wasmbus.ctl.{prefix}`
+        // Skip the topic prefix and then the lattice prefix
+        // e.g. `wasmbus.ctl.{prefix}`
+        let mut parts = subject
+            .trim()
+            .trim_start_matches(&self.ctl_topic_prefix)
+            .trim_start_matches('.')
```
thomastaylor312, line 2357: Nit: Could you just go straight to the `split` and then skip 2 to skip the empty string and then prefix?

src/main.rs
```diff
@@ -219,12 +214,15 @@ struct Args {
...
 }
 
+#[allow(clippy::too_many_lines)]
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     let Args {
         log_level,
         nats_host,
         nats_port,
+        nats_jwt,
```
thomastaylor312, line 224: This is totally a preference thing, but can we just switch this to `let args = Args::parse()` and then just use `arg.<var>` instead? This just looks kinda ridiculous at this point (and triggers a clippy lint apparent) 😆

thomastaylor312 (commented):
Mostly nits, but agree with @connorsmith256 about putting the NATS connection stuff into a function

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Yeah I'm not actually sure why, but happy to add a comment if there's reasoning

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Do you mean something like `subject.split(&self.ctl_topic_prefix).skip(1).split('.')`? I did it this way because someone could pick a prefix with an arbitrary number of parts

src/main.rs
brooksmtownsend: I don't mind either way

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I just put them in this order because that's the order I put em in everywhere else (assuming you're asking for ctl -> prov RPC -> RPC)

src/main.rs
brooksmtownsend: Did it, can always drop the commit if we hate it.

connorsmith256 (approved):
I know there are a couple outstanding nits but my comments were addressed. LGTM

Issue #481, opened 2023-08-15
Implement get operations for link definitions and claims


brooksmtownsend:

rvolosatovs:
Before we can implement claim queries we need to store them https://github.com/wasmCloud/wasmCloud/issues/492

brooksmtownsend:
Implemented by #510

PR #510, opened 2023-08-17
feat: add claims and link query functionality

connorsmith256:
## Feature or Problem
This PR:
- stores actor and provider claims when they are started
- adds the ability to query claims and links via the host

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/492 and https://github.com/wasmCloud/wasmCloud/issues/481

## Consumer Impact
The host will now respond accurately to requests on `wasmbus.ctl.{lattice_prefix}.get.claims` and `.get.links`

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
I added tests to `tests/wamsbus.rs` to confirm that the hosts responds to requests for links and claims with the correct data, and also that the data returned is exactly what the control interface client returns when querying the latticedata bucket directly

### Manual Verification
I also did manual testing with `wash get claims/links` and `nats req ...`

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -56,8 +56,6 @@ use wasmcloud_runtime::capability::{messaging, Bus, KeyValueReadWrite, Messaging
 use wasmcloud_runtime::{ActorInstancePool, Runtime};
 
 const SUCCESS: &str = r#"{"accepted":true,"error":""}"#;
-const CLAIMS_PREFIX: &str = "CLAIMS_";
-const LINKDEF_PREFIX: &str = "LINKDEF_";
 
 #[derive(Debug)]
 struct Queue {
@@ -2184,14 +2182,14 @@ impl Host {
             claims: Vec<StoredClaims>,
         }
 
-        let claims: Vec<StoredClaims> = self.scan_latticedata(CLAIMS_PREFIX).await?;
+        let claims: Vec<StoredClaims> = self.scan_latticedata("CLAIMS_").await?;
         let resp = ClaimsResponse { claims };
         Ok(serde_json::to_vec(&resp)?.into())
     }
 
     #[instrument(skip(self))]
     async fn handle_links(&self) -> anyhow::Result<Bytes> {
-        let links: Vec<LinkDefinition> = self.scan_latticedata(LINKDEF_PREFIX).await?;
+        let links: Vec<LinkDefinition> = self.scan_latticedata("LINKDEF_").await?;
         let resp = LinkDefinitionList { links };
         Ok(serde_json::to_vec(&resp)?.into())
     }
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2200,30 +2200,36 @@ impl Host {
         &self,
         key_prefix: &str,
     ) -> anyhow::Result<Vec<T>> {
-        let filtered_keys = self
+        self
             .data
             .keys()
             .await
             .context("failed to scan lattice data keys")?
+            .map_err(|e| anyhow!(e).context("failed to read lattice data stream"))
             .try_filter(|key| futures::future::ready(key.starts_with(key_prefix)))
-            .try_collect::<Vec<String>>()
-            .await
-            .context("failed to collect lattice data keys")?;
+            .and_then(|key| async move {
+                let maybe_bytes = self.data.get(&key).await?; // if we get an error here, we do want to fail
 
-        let futs = filtered_keys.into_iter().map(|key| self.data.get(key));
-        let list: Vec<T> = futures::future::join_all(futs)
-            .await
-            .into_iter()
-            .filter_map(|resp| {
-                // TODO: we should probably handle when we get an error from NATS or encountering
-                // malformed data in the bucket entries https://github.com/wasmCloud/wasmCloud/issues/509
-                resp.ok()
-                    .and_then(|bytes| bytes)
-                    .and_then(|bytes| serde_json::from_slice::<T>(&bytes).ok())
+                match maybe_bytes {
+                    None => {
+                        // this isn't the responsibility of the host, so warn and continue
+                        warn!(key, "latticedata entry was empty. Data may have been deleted during processing");
+                        anyhow::Ok(None)
+                    }
+                    Some(bytes) => {
+                        if let Ok(t) = serde_json::from_slice::<T>(&bytes) {
+                            anyhow::Ok(Some(t))
+                        } else {
+                            // this isn't the responsibility of the host, so warn and continue
+                            warn!(key, "failed to deserialize latticedata entry");
+                            anyhow::Ok(None)
+                        }
+                    }
+                }
             })
-            .collect();
-
-        Ok(list)
+            .try_filter_map(|v| futures::future::ready(Ok(v)))
+            .try_collect::<Vec<T>>()
+            .await
     }
 
     #[instrument(skip(self, payload))]
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -31,7 +31,7 @@ use cloudevents::{EventBuilder, EventBuilderV10};
 use futures::stream::{AbortHandle, Abortable};
 use futures::{stream, try_join, FutureExt, Stream, StreamExt, TryStreamExt};
 use nkeys::{KeyPair, KeyPairType};
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_json::json;
 use sha2::{Digest, Sha256};
 use time::format_description::well_known::Rfc3339;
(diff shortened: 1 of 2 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -2504,3 +2596,75 @@ impl Host {
...
+    type Error = anyhow::Error;
+
+    fn try_from(claims: jwt::Claims<jwt::CapabilityProvider>) -> Result<Self, Self::Error> {
+        let issuer = claims.issuer.clone();
+        let subject = claims.subject.clone();
+
+        let metadata = claims
+            .metadata
+            .as_ref()
+            .ok_or(anyhow!("no metadata found on provider claims"))?
```
rvolosatovs, line 2655:
```suggestion
            .context("no metadata found on provider claims")?
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2504,3 +2596,75 @@ impl Host {
...
+    type Error = anyhow::Error;
+
+    fn try_from(claims: jwt::Claims<jwt::Actor>) -> Result<Self, Self::Error> {
+        let issuer = claims.issuer.clone();
+        let subject = claims.subject.clone();
+
+        let metadata = claims
+            .metadata
+            .as_ref()
+            .ok_or(anyhow!("no metadata found on actor claims"))?
```
rvolosatovs, line 2628:
```suggestion
            .context("no metadata found on actor claims")?
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2504,3 +2596,75 @@ impl Host {
...
+impl TryFrom<jwt::Claims<jwt::Actor>> for StoredClaims {
+    type Error = anyhow::Error;
+
+    fn try_from(claims: jwt::Claims<jwt::Actor>) -> Result<Self, Self::Error> {
+        let issuer = claims.issuer.clone();
+        let subject = claims.subject.clone();
+
+        let metadata = claims
+            .metadata
+            .as_ref()
```
rvolosatovs, line 2627: Isn't `as_ref()` followed by `clone()` completely redundant?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2504,3 +2596,75 @@ impl Host {
...
+    tags: String,
+    version: String,
+}
+
+impl TryFrom<jwt::Claims<jwt::Actor>> for StoredClaims {
+    type Error = anyhow::Error;
+
+    fn try_from(claims: jwt::Claims<jwt::Actor>) -> Result<Self, Self::Error> {
+        let issuer = claims.issuer.clone();
+        let subject = claims.subject.clone();
```
rvolosatovs, line 2623: Aren't these clones redundant?

You could even destruct this as `let jwt::Claims{ issuer, subject, metadata } = claims`, right? (or even do that in the function parameter)

crates/host/src/wasmbus/mod.rs
```diff
@@ -2504,3 +2596,75 @@ impl Host {
...
+            version: metadata.ver.unwrap_or_default(),
+        })
+    }
+}
+
+impl TryFrom<jwt::Claims<jwt::CapabilityProvider>> for StoredClaims {
+    type Error = anyhow::Error;
+
+    fn try_from(claims: jwt::Claims<jwt::CapabilityProvider>) -> Result<Self, Self::Error> {
+        let issuer = claims.issuer.clone();
```
rvolosatovs, line 2649: same as above

crates/host/src/wasmbus/mod.rs
```diff
@@ -2355,6 +2403,42 @@ impl Host {
...
+        let key = format!("CLAIMS_{}", stored_claims.subject);
+
+        let bytes = serde_json::to_vec(&stored_claims)
+            .map_err(anyhow::Error::from)
+            .context("failed to serialize claims")?
+            .into();
+        self.data
+            .put(key, bytes)
+            .await
+            .map(|_| ())
```
rvolosatovs, line 2437: I think it would be much clearer to just remove this one and add `Ok(())` at the end

crates/host/src/wasmbus/mod.rs
```diff
@@ -2355,6 +2403,42 @@ impl Host {
...
+    ) -> anyhow::Result<()> {
+        self.store_claims(
+            claims
+                .try_into()
+                .context("failed to convert provider claims")?,
+        )
+        .await
+    }
+
+    async fn store_claims(&self, stored_claims: StoredClaims) -> anyhow::Result<()> {
```
rvolosatovs, line 2427: Why not take `impl TryInto<StoredClaims>` as argument instead and remove both `store_actor_claims` and `store_provider_claims` ?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2164,16 +2175,55 @@ impl Host {
...
+    async fn scan_latticedata<T: DeserializeOwned>(
+        &self,
+        key_prefix: &str,
+    ) -> anyhow::Result<Vec<T>> {
+        let filtered_keys = self
+            .data
+            .keys()
+            .await
+            .context("failed to scan lattice data keys")?
+            .try_filter(|key| futures::future::ready(key.starts_with(key_prefix)))
```
rvolosatovs, line 2208:
```suggestion
            .try_filter(|key| async { key.starts_with(key_prefix) })
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2164,16 +2175,55 @@ impl Host {
...
+            .try_collect::<Vec<String>>()
+            .await
+            .context("failed to collect lattice data keys")?;
+
+        let futs = filtered_keys.into_iter().map(|key| self.data.get(key));
```
rvolosatovs, line 2213: `try_collect` here is redundant, since immediately after allocating a whole vector you're turning it into an iterator and then allocating yet another vector of futures.

Try something like this:
```suggestion
           .and_then(|key| async {
              let buf = self.data.get(key).await.context("failed to get lattice key")?;
              serde_json::from_slice::<T>(&buf)
           });
```

crates/host/src/wasmbus/mod.rs
connorsmith256: Whoops yes, nice catch

crates/host/src/wasmbus/mod.rs
connorsmith256: Yep, should have cleaned this up after iterating 👍

crates/host/src/wasmbus/mod.rs
connorsmith256: I'm a fan of the toilet closure `|_| ()` 😉 IMO it's more clear that we're returning the result from `put` either way: either the error, or a mapped version of the ok. But I don't feel strongly about it, so I'll change it

crates/host/src/wasmbus/mod.rs
connorsmith256: I had fun with lifetime errors earlier. I started with that, but:
```
error: lifetime may not live long enough
    --> crates/host/src/wasmbus/mod.rs:2191:31
     |
2191 |             .try_filter(|key| async { key.starts_with(CLAIMS_PREFIX) })
     |                          ---- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
     |                          |  |
     |                          |  return type of closure `[async block@crates/host/src/wasmbus/mod.rs:2191:31: 2191:71]` contains a lifetime `'2`
     |                          has type `&'1 std::string::String````

File: crates/host/src/wasmbus/mod.rs (hunk 7 above)
Comment (connorsmith256) [reply crates/host/src/wasmbus/mod.rs:2427]: I couldn't get this to work earlier, but you got me to go back to it 👍 `impl TryInto<StoredClaims>` isn't enough, but this is:
```T: TryInto<StoredClaims, Error = anyhow::Error>```

File: crates/host/src/wasmbus/mod.rs (hunk 9 above)
Comment (connorsmith256) [reply crates/host/src/wasmbus/mod.rs:2213]: This approach resolves https://github.com/wasmCloud/wasmCloud/issues/509, but I'm not sure it's the best behavior: if we fail to retrieve any value — say a key is deleted while processing, or there's a malformed entry in the bucket — do we want to fail to return anything to the client? Wouldn't it be better to exclude bad values and emit a warning?

Comment (rvolosatovs) [reply crates/host/src/wasmbus/mod.rs:2213]: Yeah, I think we should print an `error` log if we fail to deserialize an entry here, but I'm not suggesting any change in behavior here, just removing the redundant steps

File: crates/host/src/wasmbus/mod.rs [hunk 10]
```diff
@@ -48,14 +48,16 @@ use uuid::Uuid;
...
+    LinkDefinitionList, ProviderAuctionRequest, ProviderDescription, RemoveLinkDefinitionRequest,
+    ScaleActorCommand, StartActorCommand, StartProviderCommand, StopActorCommand, StopHostCommand,
     StopProviderCommand, UpdateActorCommand,
 };
 use wasmcloud_runtime::capability::{messaging, Bus, KeyValueReadWrite, Messaging};
 use wasmcloud_runtime::{ActorInstancePool, Runtime};
 
 const SUCCESS: &str = r#"{"accepted":true,"error":""}"#;
+const CLAIMS_PREFIX: &str = "CLAIMS_";
+const LINKDEF_PREFIX: &str = "LINKDEF_";
```
rvolosatovs, line 60: these are only used once in the whole codebase, is there really any benefit we get from not just inlining the values?

crates/host/src/wasmbus/mod.rs
connorsmith256: I was confused by your code snippet, where failing to deserialize would also result in a top-level error. I've refactored the code and added comments clarifying which errors I think we should gracefully bypass

rvolosatovs (approved):

PR #517, opened 2023-08-18
Propagate allow_file_load, implement structured logging

brooksmtownsend:
## Feature or Problem
This PR:
1. Propagates and respects the `allow_file_load` parameter
2. Enables structured logging with the `json()` layer, and passes that field and `log_level` to providers as well

Marked as draft for testing with `allow_file_load`, and because I could use some advice on the structured logging and if this is sufficient. I'd also like #506 to merge first 

## Related Issues
N/A

## Release Information
N/A

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/config.rs
+++ b/crates/host/src/wasmbus/config.rs
@@ -50,6 +50,12 @@ pub struct Host {
     pub provider_shutdown_delay: Option<std::time::Duration>,
     /// Configuration for downloading artifacts from OCI registries
     pub oci_opts: OciConfig,
+    /// Whether to allow loading actor or provider components from the filesystem
+    pub allow_file_load: bool,
+    // Whether or not structured logging is enabled
+    // pub enable_structured_logging: bool,
+    // Log level to pass to capability providers to use. Should be parsed from a [`tracing::Level`]
+    // pub log_level: String,
 }
 
 impl Default for Host {
@@ -79,6 +85,9 @@ impl Default for Host {
             cluster_issuers: None,
             provider_shutdown_delay: None,
             oci_opts: OciConfig::default(),
+            allow_file_load: false,
+            // enable_structured_logging: false,
+            // log_level: "INFO".to_string(),
         }
     }
 }
--- a/crates/host/src/wasmbus/config.rs
+++ b/crates/host/src/wasmbus/config.rs
@@ -5,6 +5,7 @@ use url::Url;
 
 use crate::oci::Config as OciConfig;
 
+#[allow(clippy::struct_excessive_bools)]
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 /// wasmCloud Host configuration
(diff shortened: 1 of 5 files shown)


crates/host/src/wasmbus/config.rs
```diff
@@ -50,6 +50,12 @@ pub struct Host {
     pub provider_shutdown_delay: Option<std::time::Duration>,
     /// Configuration for downloading artifacts from OCI registries
     pub oci_opts: OciConfig,
+    /// Whether to allow loading actor or provider components from the filesystem
+    pub allow_file_load: bool,
+    /// Whether or not structured logging is enabled
+    pub enable_structured_logging: bool,
+    /// Log level to pass to capability providers to use. Should be parsed from a [`tracing::Level`]
+    pub log_level: String,
```
connorsmith256, line 58: How about an enum?

src/main.rs
```diff
@@ -218,17 +217,21 @@ struct Args {
...
-            }),
-        )
-        .init();
+    let env_filter = tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+        tracing_subscriber::EnvFilter::new(format!("{},cranelift_codegen=warn", args.log_level))
+    });
+
+    if args.enable_structured_logging {
+        tracing_subscriber::registry()
+            .with(tracing_subscriber::fmt::layer().json().without_time())
```
connorsmith256, line 226: I think there are some other values that need to be set, or we run into weird terminal issues as well as parsing issues from the tracing crate at runtime. See https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-sdk/src/provider_main.rs#L400

crates/host/src/wasmbus/config.rs
brooksmtownsend: I wanted to use `tracing::Level` and then just `.to_string()` it but it didn't meet the wasmbus_rpc Serialize trait requirements 🤔

src/main.rs
brooksmtownsend: Awesome thank you, I knew I was missing something

connorsmith256: Discussed offline: we haven't reproduced issues with interleaving lines (yet?), so we think this is good to go. If we run into issues we can use an approach like above to resolve it

brooksmtownsend: I'm going to push this off for the future because I don't think I'm introducing anything different in this PR other than just the basic inclusion of JSON logging, but if we see interleaved logs again we can file an issue + fix

connorsmith256 (approved):

PR #522, opened 2023-08-18
Fix/list parsing

connorsmith256:
## Feature or Problem
Providers should not implicitly receive environment variables from the host, since this is a security risk. Values providers need should be passed via `HostData`

Note this means we need to add `OTEL_TRACES_EXPORTER` and `OTEL_EXPORTER_OTLP_ENDPOINT` to `HostData`, which I've filed in https://github.com/wasmCloud/wasmCloud/issues/523

## Related Issues
N/A

## Release Information
Next

## Consumer Impact
As we rewrite providers, we should remove uses of `std::env::var`, since these values will now be empty

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
I started the host with an env var, launched a provider, inspected the env vars the child process had, and confirmed the host's var was not proxied to the provider

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1922,6 +1922,7 @@ impl Host {
                 "spawn provider process"
             );
             let mut child = process::Command::new(&path)
+                .env_clear()
                 .stdin(Stdio::piped())
                 .kill_on_drop(true)
                 .spawn()


rvolosatovs:
Why was this closed?

connorsmith256:
I pushed from the wrong branch and this included commits from another branch. The commit I intended for this PR has now been merged

PR #532, opened 2023-08-22
feat: delete claims when actors or providers are stopped

Iceber:
## Feature or Problem

Claims need to be deleted in the following cases:
1. stop an actor and instances is empty
2. scale an actor to 0
3. stop provider

related pr: https://github.com/wasmCloud/wasmCloud/pull/510
## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1474,6 +1474,10 @@ impl Host {
     ) -> anyhow::Result<()> {
         let actor = entry.remove();
         let claims = actor.pool.claims().context("claims missing")?;
+        self.delete_claims(claims.clone())
+            .await
+            .context("failed to delete claims")?;
+
         let mut instances = actor.instances.write().await;
         let remaining: usize = instances.values().map(Vec::len).sum();
         stream::iter(instances.drain())
@@ -1679,6 +1683,9 @@ impl Host {
                     }
                     if remaining == 0 {
                         drop(instances);
+                        self.delete_claims(claims.clone())
+                            .await
+                            .context("failed to delete claims")?;
                         entry.remove();
                     }
                 }
@@ -1837,6 +1844,9 @@ impl Host {
                 }
                 if instances.len() == 0 {
                     drop(instances);
+                    self.delete_claims(claims.clone())
+                        .await
+                        .context("failed to delete claims")?;
                     entry.remove();
                 }
             }
@@ -2142,6 +2152,9 @@ impl Host {
             }
         }
         if instances.is_empty() {
+            self.delete_claims(provider.claims.clone())
+                .await
+                .context("failed to delete claims")?;
             entry.remove();
         }
         Ok(SUCCESS.into())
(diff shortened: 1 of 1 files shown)


rvolosatovs (approved):

brooksmtownsend:
This PR should be reverted, unfortunately I missed the review window.

Claims should stick around even if an actor or provider stops because there can always be another host in the same lattice that is running the same actor or provider that still depends on those claims.

Cc @rvolosatovs

rvolosatovs:
> This PR should be reverted, unfortunately I missed the review window.
> 
> Claims should stick around even if an actor or provider stops because there can always be another host in the same lattice that is running the same actor or provider that still depends on those claims.
> 
> Cc @rvolosatovs 

Could you open a revert PR?

Iceber:
@brooksmtownsend Sorry for overlooking this, so when do we need to remove claims from nats?

Or rather, claims are only increasing, not decreasing.

Iceber:
Or put the claims information into a heartbeat broadcast? Each host caches claims information based on other hosts' heartbeats.

brooksmtownsend:
Hey @Iceber! No worries at all, this was great thinking in context of the rest of the codebase, it just happens to cause undetermined behavior in advanced cases.

For now, claims are only increasing, and I left a little comment in #534 to address this. Claims cache is relatively small so I'm not too worried about it growing too large, but wouldn't mind adding functionality in `wash drain` to remove the claims cache.

connorsmith256:
@Iceber we don't need to explicitly manage the lifetime of entries in the lattice data bucket, including claims, at least for now.

Outdated claims will be replaced when an updated actor/provider is started, so we don't need to worry about data integrity. As for storage usage, claims entries are an insignificant number of bytes, and the number of claims scales with the number of unique actors/providers started in the lattice, which is likely to be low.

Eventually, the lattice data bucket would grow to a point where cleanup is justified. But I'm not aware of a lattice which has ever gotten to this point. It's also possible for operators to pre-provision the bucket/underlying stream before the host starts with a max size and eviction policy. The default policy for NATS streams is to evict oldest entries first.

So for now I don't think the wasmCloud host ever needs to delete claims entries from the bucket :)

Iceber:
@brooksmtownsend @connorsmith256 Thanks a lot for your replies!

I have one last question, Sorry if I ask silly doubts at times

When a host is shut down, this may leave behind some claims that were only used on that host or an actor may be deleted and never created again. These can lead to a lot of meaningless claims when getting claims, isn't this something to consider?

I haven't used wasmcloud in production yet, so I'm not sure if this scenario makes sense!

brooksmtownsend:
I think that makes sense @Iceber and makes me think even more that it would be worth it to add a claims/linkdefs purge inside of `wash drain`. https://github.com/wasmCloud/wash/issues/719

PR #533, opened 2023-08-22
feat: implement `wasmcloud:http/incoming-handler` support

rvolosatovs:
## Feature or Problem

Implement `wasi-http:/incoming-handler` support for components using the legacy Smithy `HttpServer` contract

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/421

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`


crates/compat/src/http.rs
```diff
@@ -23,6 +26,87 @@ pub struct Request {
...
+            query_string,
+            header,
+            body,
+        }: Request,
+    ) -> Result<Self, Self::Error> {
+        let req = http::Request::builder().method(method.as_str());
+        let req = header
+            .into_iter()
+            .filter_map(|(name, mut values)| {
+                let value = values.pop()?;
```
connorsmith256, line 97: Can we add context here?

crates/compat/src/http.rs
```diff
@@ -23,6 +26,87 @@ pub struct Request {
...
+                Some((name, value))
+            })
+            .fold(req, |req, (name, value)| req.header(name, value));
+        let req = if !path.is_empty() || !query_string.is_empty() {
+            req.uri(format!("{path}{query_string}"))
```
connorsmith256, line 102: If path is empty, is this valid?

crates/compat/src/http.rs
```diff
@@ -48,3 +132,73 @@ impl Default for Response {
...
+            status_code,
+            header,
+            body,
+        }: Response,
+    ) -> Result<Self, Self::Error> {
+        let res = http::Response::builder().status(status_code);
+        let res = header
+            .into_iter()
+            .filter_map(|(name, mut values)| {
+                let value = values.pop()?;
```
connorsmith256, line 198: Same question here, can we add context?

crates/host/src/wasmbus/mod.rs
```diff
@@ -747,6 +749,66 @@ impl Messaging for Handler {
...
+                    .into_incoming_http()
+                    .await
+                    .context("failed to instantiate `wasi:http/incoming-handler`")?
+                    .handle(req.map(|body| -> Box<dyn AsyncRead + Send + Sync + Unpin> {
+                        Box::new(Cursor::new(body))
+                    }))
+                    .await
+                {
+                    Ok(res) => res,
+                    Err(err) => return Ok(Err(format!("{err:#}"))),
```
connorsmith256, line 787: Is this for pretty printing?

crates/compat/src/http.rs
rvolosatovs: No, because this needs to return an `Option` and I think that empty header values should just be ignored

crates/host/src/wasmbus/mod.rs
rvolosatovs: This is how errors are printed throughout runtime crate, which provides the error chain

crates/compat/src/http.rs
rvolosatovs: No, see answer below

crates/compat/src/http.rs
rvolosatovs: It may be - there's nothing else we can really do here.
If the resulting URI is not valid, then the request construction below will fail, so I don't think this really matters

crates/host/src/wasmbus/mod.rs
rvolosatovs: See https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations

connorsmith256 (approved):

PR #539, opened 2023-08-22
feat(wasmbus)!: implement provider health checks

brooksmtownsend:
## Feature or Problem
This PR implements periodic provider health checks, publishing a `health_check_passed` event when the provider succeeds and a `health_check_failed` when the provider fails. This PR also preserves the OTP behavior which avoids spammy health check updates by only publishing a health check event if the status changes. 

I also refactored the `publish_event` functionality to be inside of the `events` module as an alternate way to publish an event, given a `ctl_nats` and `eventbuilder` as arguments. This is useful because the closure that monitors provider health checks (and will exit once the provider stops) doesn't really need a full clone of the host object, and instead we can pass a cheaply cloned eventbuilder + ctl_nats client.

I found this by testing wadm and realizing that wadm essentially relies on health checks to know that a provider is healthy and ready for a link definition.

## Related Issues
Requires https://github.com/wasmCloud/wadm/pull/164 to "fully" support applications

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
I tested this working with our kvcounter + echo applications.


crates/host/src/wasmbus/mod.rs
```diff
@@ -2000,13 +1993,84 @@ impl Host {
...
-                        error!("failed to wait for `{}` to execute: {e}", path.display());
+                // Check the health of the provider every 30 seconds
+                let mut health_check = tokio::time::interval(Duration::from_secs(30));
+                let mut previous_healthy = false;
+                // Allow the provider 5 seconds to initialize
+                health_check.reset_after(Duration::from_secs(5));
+                let health_topic =
+                    format!("wasmbus.rpc.{health_lattice_prefix}.{health_provider_id}.{health_link_name}.health");
+                loop {
+                    tokio::select! {
```
rvolosatovs, line 2013:
```suggestion
                    select! {
```

crates/host/src/wasmbus/event.rs
```diff
@@ -210,3 +215,43 @@ pub fn provider_stopped(
...
+) -> serde_json::Value {
+    json!({
+        "public_key": public_key.as_ref(),
+        "link_name": link_name.as_ref(),
+        "contract_id": contract_id.as_ref(),
+    })
+}
+
+#[instrument(skip(event_builder, ctl_nats, name))]
+pub(crate) async fn publish_event(
```
rvolosatovs, line 232:
```suggestion
pub(crate) async fn publish(
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2000,13 +1993,84 @@ impl Host {
...
+                                                    &health_provider_id,
+                                                    &health_link_name,
+                                                    &health_contract_id,
+                                                )
+                                            ).await {
+                                                warn!(?e, "failed to publish provider health check failed event");
+                                            }
+                                        }
+                                        // If the provider health status didn't change, we don't need to publish an event
+                                        (Ok(_), true | false) => (),
```
rvolosatovs, line 2056:
```suggestion
                                        (Ok(_), _) => (),
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2000,13 +1993,84 @@ impl Host {
...
+                                    warn!("failed to request provider health, retrying in 30 seconds");
+                                }
+                        }
+                        exit_status = child.wait() => match exit_status {
+                            Ok(status) => {
+                                debug!("`{}` exited with `{status:?}`", path.display());
+                                break;
+                            }
+                            Err(e) => {
+                                error!("failed to wait for `{}` to execute: {e}", path.display());
```
rvolosatovs, line 2070:
```suggestion
                                warn!("failed to wait for `{}` to execute: {e}", path.display());
```

No need to spam production logs with errors caused by broken providers (user error)

crates/host/src/wasmbus/mod.rs
```diff
@@ -2000,13 +1993,84 @@ impl Host {
...
             stdin.shutdown().await.context("failed to close stdin")?;
 
+            let prov_nats = self.prov_rpc_nats.clone();
+            let ctl_nats = self.ctl_nats.clone();
+            let event_builder = self.event_builder.clone();
+            // NOTE: health_ prefix here is to allow us to move the variables into the closure
+            let health_lattice_prefix = self.host_config.lattice_prefix.clone();
+            let health_provider_id = claims.subject.to_string();
+            let health_link_name = link_name.to_string();
+            let health_contract_id = claims.metadata.clone().map(|m| m.capid).unwrap_or_default();
```
rvolosatovs, line 2003: Nothing necessarily needs to change here, but just want to point out that we could easily just change method receiver to `Arc<Self>` and then clone the Arc of the host itself and move it into the closure.

Note that it would make `event::publish` redundant and remove a bunch of clones here

crates/host/src/wasmbus/mod.rs
```diff
@@ -2000,13 +1993,84 @@ impl Host {
...
+                                        }
+                                        // If the provider health status didn't change, we don't need to publish an event
+                                        (Ok(_), _) => (),
+                                        _ => warn!("failed to deserialize provider health check response"),
+                                    }
+                                }
+                                else {
+                                    warn!("failed to request provider health, retrying in 30 seconds");
+                                }
+                        }
```
rvolosatovs, line 2063: There's way too much nesting here. I'm fine with this if we add a `TODO` to refactor, otherwise, could we adapt an approach similar to https://github.com/wasmCloud/wasmCloud/blob/b77767e6d3c32ceba0b4e5b421b532ac0788dc15/tests/wasmbus.rs#L96-L115 perhaps?

brooksmtownsend: I'll add a TODO here so we can tackle it, along with the idea of cloning Arc<Self> to clean up the code

rvolosatovs (approved):

PR #545, opened 2023-08-23
store claims when the scale triggers the startup actor

Iceber:
## Feature or Problem

Claims also need to be stored when scale triggers the start of an actor.
https://github.com/wasmCloud/wasmCloud/blob/4e0c5b02b50759c01c2e6689d02342f86d6c5bc7/crates/host/src/wasmbus/mod.rs#L1675-L1683

Storing claims in the `start_actor` function ensures that claims are stored for the newly started actor.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1407,6 +1407,10 @@ impl Host {
 
         let annotations = annotations.map(Into::into);
         let claims = actor.claims().context("claims missing")?;
+        self.store_claims(claims.clone())
+            .await
+            .context("failed to store claims")?;
+
         let links = self.links.read().await;
         let links = links
             .values()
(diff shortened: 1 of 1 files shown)


Iceber:
@rvolosatovs @brooksmtownsend @connorsmith256 PTAL, I hope I'm not overlooking something.

brooksmtownsend (approved):
I think this makes sense! Refactor looks good, just moving from outside of the control interface so we make sure to always store the claims of a newly started actor 👍🏻

connorsmith256 (approved):
Nice catch!

PR #546, opened 2023-08-23
refactor: error message for instantiation failure

vados-cosmonic:
## Feature or Problem

When methods of instantiation fail, show all related errors from attempts rather that only the last one.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/runtime/src/actor/component/mod.rs
+++ b/crates/runtime/src/actor/component/mod.rs
@@ -8,7 +8,7 @@ use core::ops::{Deref, DerefMut};
 
 use std::sync::Arc;
 
-use anyhow::{ensure, Context as _};
+use anyhow::{bail, ensure, Context as _};
 use async_trait::async_trait;
 use bytes::{Bytes, BytesMut};
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
@@ -418,16 +418,30 @@ impl Instance {
 
     /// Instantiates and returns [`GuestBindings`] if exported by the [`Instance`].
     async fn as_guest_bindings(&mut self) -> anyhow::Result<GuestBindings> {
-        if let Ok((bindings, _)) =
-            guest_bindings::Guest::instantiate_async(&mut self.store, &self.component, &self.linker)
-                .await
+        // Attempt to instantiate using guest bindings
+        let guest_err = match guest_bindings::Guest::instantiate_async(
+            &mut self.store,
+            &self.component,
+            &self.linker,
+        )
+        .await
         {
-            Ok(GuestBindings::Interface(bindings))
-        } else {
-            let (bindings, _) = Command::instantiate_async(&mut self.store, &self.component, &self.linker).await.context(
-                    "failed to instantiate either `wasmcloud::bus/guest` interface or get `wasi:command/command`",
-                )?;
-            Ok(GuestBindings::Command(bindings))
+            Ok((bindings, _)) => return Ok(GuestBindings::Interface(bindings)),
+            Err(e) => e,
+        };
+
+        // Attempt to instantiate using only bindings available in command
+        match Command::instantiate_async(&mut self.store, &self.component, &self.linker).await {
+            Ok((bindings, _)) => Ok(GuestBindings::Command(bindings)),
+            // If neither of the above instantiations worked, the instance cannot be run
+            Err(command_err) => bail!(
+                r#"failed to instantiate instance (no bindings satisfied exports):
+
+`wasmcloud:bus/guest` error: {guest_err:?}
+
+`wasi:command/command` error: {command_err:?}
+"#,
+            ),
         }
     }
 


crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,17 +418,37 @@ impl Instance {
 
     /// Instantiates and returns [`GuestBindings`] if exported by the [`Instance`].
     async fn as_guest_bindings(&mut self) -> anyhow::Result<GuestBindings> {
-        if let Ok((bindings, _)) =
-            guest_bindings::Guest::instantiate_async(&mut self.store, &self.component, &self.linker)
-                .await
+        let mut errors = Vec::new();
+
```
rvolosatovs, line 422:
```suggestion
```

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,17 +418,37 @@ impl Instance {
...
+        // Attempt to instantiate using only bindings available in command
+        match Command::instantiate_async(&mut self.store, &self.component, &self.linker).await {
+            Ok((bindings, _)) => return Ok(GuestBindings::Command(bindings)),
+            Err(e) => errors.push(e),
+        };
+
+        // If neither of the above instantiations worked, the instance cannot be run
+        bail!(
+            r#"no bindings could satisfy instance instantiation:
```
rvolosatovs, line 443: Could we stick to `failed to ...` like the rest of the errors in the codebase?

rvolosatovs (approved):

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,17 +418,37 @@ impl Instance {
 
     /// Instantiates and returns [`GuestBindings`] if exported by the [`Instance`].
     async fn as_guest_bindings(&mut self) -> anyhow::Result<GuestBindings> {
-        if let Ok((bindings, _)) =
-            guest_bindings::Guest::instantiate_async(&mut self.store, &self.component, &self.linker)
-                .await
+        let mut errors = Vec::new();
```
rvolosatovs, line 421: How about we only store a singular `guest_error: Option<anyhow::Error>` here?

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,17 +418,37 @@ impl Instance {
...
-            Ok(GuestBindings::Command(bindings))
-        }
+            Ok((bindings, _)) => return Ok(GuestBindings::Interface(bindings)),
+            Err(e) => errors.push(e),
+        };
+
+        // Attempt to instantiate using only bindings available in command
+        match Command::instantiate_async(&mut self.store, &self.component, &self.linker).await {
+            Ok((bindings, _)) => return Ok(GuestBindings::Command(bindings)),
+            Err(e) => errors.push(e),
```
rvolosatovs, line 438: Could we directly `bail` here?

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,17 +418,37 @@ impl Instance {
...
-                )?;
-            Ok(GuestBindings::Command(bindings))
-        }
+            Ok((bindings, _)) => return Ok(GuestBindings::Interface(bindings)),
+            Err(e) => errors.push(e),
+        };
+
+        // Attempt to instantiate using only bindings available in command
+        match Command::instantiate_async(&mut self.store, &self.component, &self.linker).await {
+            Ok((bindings, _)) => return Ok(GuestBindings::Command(bindings)),
```
rvolosatovs, line 437: If we bail in the other match case, 
```suggestion
            Ok((bindings, _)) => Ok(GuestBindings::Command(bindings)),
```

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: Sure I can change to only storing the `guest_error` -- I think that's what most of the suggestions are about so will address them together

- [ ] store only guest error as an `Option` rather than the list

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: - [x] make error text consistent with rest of codebase (addressed this with the other comment)

vados-cosmonic:
@rvolosatovs would you mind taking another look? had to change the code a bit to make clippy happy with the `Option` usage.

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,16 +418,39 @@ impl Instance {
...
+
+        // Attempt to instantiate using only bindings available in command
+        match (
+            Command::instantiate_async(&mut self.store, &self.component, &self.linker).await,
+            guest_err,
+        ) {
+            (Ok((bindings, _)), _) => Ok(GuestBindings::Command(bindings)),
+            // If neither of the above instantiations worked, the instance cannot be run
+            (Err(command_err), Some(guest_err)) => bail!(
+                r#"failed to instantiate instance (no bindings satisfied imports/exports):
```
rvolosatovs, line 441:
```suggestion
                r#"failed to instantiate instance (no bindings satisfied exports):
```

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,16 +418,39 @@ impl Instance {
...
+
+`wasmcloud:bus/guest` error: {:?}
```
rvolosatovs, line 443:
```suggestion
`wasmcloud:bus/guest` error: {guest_err:?}
```

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,16 +418,39 @@ impl Instance {
...
+
+`wasi:command/command` error: {:?}
```
rvolosatovs, line 445:
```suggestion
`wasi:command/command` error: {command_err:?}
```

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,16 +418,39 @@ impl Instance {
...
+        .await
         {
-            Ok(GuestBindings::Interface(bindings))
-        } else {
-            let (bindings, _) = Command::instantiate_async(&mut self.store, &self.component, &self.linker).await.context(
-                    "failed to instantiate either `wasmcloud::bus/guest` interface or get `wasi:command/command`",
-                )?;
-            Ok(GuestBindings::Command(bindings))
+            Ok((bindings, _)) => return Ok(GuestBindings::Interface(bindings)),
+            Err(e) => Some(e),
```
rvolosatovs, line 430: I would just nest the second call here to avoid the `unreachable` below

rvolosatovs (dismissed):

crates/runtime/src/actor/component/mod.rs
rvolosatovs, line 430: Wait, but you don't actually need an `Option` here, right?

rvolosatovs, line 430:
```suggestion
            Err(e) => e,
```

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: addressed!

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: good suggestion, updated to use the short syntax now that we're not using the array

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: removed since you're right we don't need the `Option` now

vados-cosmonic: yup that's right -- refactored this

vados-cosmonic: made this change!

crates/runtime/src/actor/component/mod.rs
```diff
@@ -418,16 +418,33 @@ impl Instance {
...
-                )?;
-            Ok(GuestBindings::Command(bindings))
+            Ok((bindings, _)) => return Ok(GuestBindings::Interface(bindings)),
+            Err(e) => e,
+        };
+
+        // Attempt to instantiate using only bindings available in command
+        match (
+            Command::instantiate_async(&mut self.store, &self.component, &self.linker).await,
+            guest_err,
```
rvolosatovs, line 436:
```suggestion
```

rvolosatovs (approved):

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: just removed the second bit of this match

Issue #511, opened 2023-08-17
[Rust Host] Support supplemental config


connorsmith256:
The OTP host supports loading supplemental config at start time via an external "config service" over NATS. There are two uses for supplemental config currently:
- starting a set of actors/providers as part of the host's startup
- managing a dynamic set of registry credentials

The documentation of this feature has always been lacking. The [smithy interface](https://github.com/wasmCloud/interfaces/blob/main/configservice/config-service.smithy) is one of the clearest resources.

I'm not aware of any users depending on auto-starting actors/providers, and with the release of wadm, I think it would be acceptable to **not** port this functionality to the new host.

However, there are users of the config service for initializing registry credentials from an external source, so we'll need to continue supporting that feature

PR #547, opened 2023-08-23
feat: support registry settings via config service and command-line args

connorsmith256:
## Feature or Problem
This PR:
- adds support for the config service
- unifies registry settings retrieved via config service with the OCI command-line args
- adds support for `.registries.put` on the control interface

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/511

## Release Information
Next

## Consumer Impact
If the config service is enabled, upon startup the host will make a request for supplemental configuration, including registry settings

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
I performed extensive e2e testing manually, using a combination of the CLI args and a config service I have running. I'm open to ideas for adding to our automated integration tests. Currently these use file references, but I suppose we could spin up a registry locally and push artifacts to that?


crates/host/src/lib.rs
```diff
@@ -63,34 +67,59 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
+                }
+            }
             Err(e) => {
                 bail!(anyhow!(e).context(format!("failed to parse reference `{s}`")))
             }
         }
     }
 }
 
+fn extract_server<'a>(resource_ref: &ResourceRef<'a>) -> Option<&'a str> {
```
rvolosatovs, line 91: Why can't we use `authority` from `Url` for this?

crates/host/src/lib.rs
```diff
@@ -63,34 +67,59 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
 #[instrument(skip(actor_ref))]
 pub async fn fetch_actor(
     actor_ref: impl AsRef<str>,
-    oci_opts: &oci::Config,
     allow_file_load: bool,
+    all_registry_settings: &HashMap<String, RegistrySettings>,
 ) -> anyhow::Result<Vec<u8>> {
-    match ResourceRef::try_from(actor_ref.as_ref())? {
+    let actor_ref = ResourceRef::try_from(actor_ref.as_ref())?;
+    let server = extract_server(&actor_ref).unwrap_or_default();
```
rvolosatovs, line 108: I would expect to see something like 
```suggestion
    let server = actor_ref.authority().unwrap_or_default();
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -966,6 +971,103 @@ async fn create_lattice_metadata_bucket(
...
+                    warn!(
+                        ?e,
+                        "failed to deserialize supplemental config. Defaulting to empty config"
+                    );
+                    Ok(SupplementalConfig::default())
+                }
+            }
+        }
+        Err(e) => {
+            warn!(
```
rvolosatovs, line 1015:
```suggestion
            error!(
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -1612,9 +1725,9 @@ impl Host {
     #[instrument(skip(self))]
     async fn fetch_actor(&self, actor_ref: &str) -> anyhow::Result<wasmcloud_runtime::Actor> {
         let actor = fetch_actor(
-            &actor_ref,
-            &self.host_config.oci_opts,
+            actor_ref,
             self.host_config.allow_file_load,
+            &*self.registry_settings.read().await,
```
rvolosatovs, line 1730: Could we extract this onto a separate line for clarity? (I.e. `let registry_settings = self.registry_settings.read().await`)

crates/host/src/wasmbus/mod.rs
```diff
@@ -1984,8 +2097,8 @@ impl Host {
         let (path, claims) = crate::fetch_provider(
             provider_ref,
             link_name,
-            &self.host_config.oci_opts,
             self.host_config.allow_file_load,
+            &*self.registry_settings.read().await,
```
rvolosatovs, line 2101: Same as above

rvolosatovs (approved):

brooksmtownsend (commented):
Overall I'm a big fan of the refactor and changes to an easier-to-understand `registry::Settings` struct. Just commenting since @rvolosatovs has requested changes that should be resolved but other than that LGTM

crates/host/src/lib.rs
```diff
@@ -63,34 +67,59 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
+    let default_registry_settings = RegistrySettings::default();
+    let registry_settings = all_registry_settings
+        .get(server)
+        .unwrap_or(&default_registry_settings);
```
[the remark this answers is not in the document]
connorsmith256: I don't think this will work. `get` returns `Option<&T>`, whereas `Default` is `T`. I can't return a reference to a `T` allocated inside `unwrap_or_else`; that will produce `cannot return reference to temporary value`.

The default value for the struct is two enum values and two bools, so eagerly allocating doesn't cost us anything significant. I agree the code would be cleaner if I could do what you're suggesting, of course

crates/host/src/lib.rs
connorsmith256: Hmm, `ResourceRef` contains a `&str`, which most of the time won't be a valid URL, since OCI and Bindle references don't contain a scheme. I could change `ResourceRef` to store a `Url` instead, with the type prefix as a pseudo-scheme, but then I would need to strip the scheme before passing an `impl AsRef<str>` to `fetch_actor`/`fetch_provider`, since the resource string we give to the underlying OCI/bindle clients shouldn't have them.

I think it's simpler to pull off the first part of the string before a `/`, since that defines the server for OCI and Bindle references. In fact, that's [how the OCI crate does it](https://github.com/krustlet/oci-distribution/blob/main/src/reference.rs#L264)

crates/host/src/bindle.rs
```diff
@@ -62,39 +59,22 @@ impl TokenManager for Auth {
...
+fn get_bindle_auth(registry_settings: &RegistrySettings) -> Auth {
+    match &registry_settings.auth {
+        RegistryAuth::Basic(username, password) => Auth::Http(HttpBasic::new(username, password)),
+        RegistryAuth::Token(token) => Auth::LongLived(LongLivedToken::new(token)),
+        RegistryAuth::Anonymous => Auth::None(NoToken),
     }
 }
 
 /// Returns a bindle client configured to cache to disk
 #[allow(clippy::implicit_hasher)]
```
rvolosatovs, line 71: This is now redundant, isn't it? (Same for other occurrences)

rvolosatovs (approved):

rvolosatovs (approved):

PR #552, opened 2023-08-23
fix(host): ignore empty responses

rvolosatovs:
## Feature or Problem

It appears that providers using `wasmcloud-provider-sdk` differ from the ones using `wasmbus-rpc` in that the latter return an empty response on RPCs that do not define a response, whereas the former encodes `()`. Account for both.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -623,7 +623,11 @@ impl KeyValueReadWrite for Handler {
                 },
             )
             .await?;
-        rmp_serde::from_slice(&res).context("failed to decode response")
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
Weird. LGTM

PR #564, opened 2023-08-24
refactor(host): clean-up

rvolosatovs:
## Feature or Problem

Various clean-ups for consistency and clarity

## Related Issues

Mostly fixup #547 

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`

### DIFF
--- a/crates/host/src/bindle.rs
+++ b/crates/host/src/bindle.rs
@@ -1,7 +1,7 @@
 // Adapted from
 // https://github.com/wasmCloud/wasmcloud-otp/blob/5f13500646d9e077afa1fca67a3fe9c8df5f3381/host_core/native/hostcore_wasmcloud_native/src/client.rs
 
-use crate::{par, RegistryAuth};
+use crate::{par, RegistryAuth, RegistryConfig};
 
 use std::env;
 use std::env::consts::{ARCH, OS};
@@ -113,7 +113,7 @@ pub async fn get_client(
         .verification_strategy(VerificationStrategy::MultipleAttestation(vec![
             SignatureRole::Host,
         ]))
-        .build(&bindle_url, auth, std::sync::Arc::new(keyring))?;
+        .build(&bindle_url, auth, Arc::new(keyring))?;
     let local = FileProvider::new(bindle_dir, NoopEngine::default()).await;
     Ok(DumbCache::new(client, local))
 }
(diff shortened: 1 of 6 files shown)


crates/host/src/lib.rs
```diff
@@ -88,40 +90,47 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
-fn extract_server<'a>(resource_ref: &ResourceRef<'a>) -> Option<&'a str> {
-    match resource_ref {
-        ResourceRef::File(_) => None,
-        ResourceRef::Bindle(resource_ref) | ResourceRef::Oci(resource_ref) => {
-            resource_ref.split('/').next()
-        }
-    }
+fn oci_authority(s: &str) -> Option<&str> {
+    let (l, _) = s.split_once('/')?;
+    Some(l)
```
brooksmtownsend, line 95: This is used for bindle and OCI, should we name it `registry_authority` or `registry_url`?

crates/host/src/oci.rs
```diff
@@ -37,29 +36,47 @@ pub struct Config {
...
+
 /// Fetch an OCI path
 #[allow(clippy::missing_errors_doc)] // TODO: document errors
 pub async fn fetch_oci_path(
-    img: &str,
-    registry_settings: &RegistrySettings,
+    img: impl AsRef<str>,
+    auth: &RegistryAuth,
+    allow_latest: bool,
+    allow_insecure: bool,
```
brooksmtownsend, line 65: These refactors from `RegistrySettings` -> using auth and the two boolean flags feels like a fine refactor for now but might just get reverted if we add one or two more fields. No strongly held opinion here.

brooksmtownsend (changes requested):

crates/host/src/lib.rs
```diff
@@ -88,40 +90,47 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
-    let actor_ref = ResourceRef::try_from(actor_ref.as_ref())?;
-    let server = extract_server(&actor_ref).unwrap_or_default();
-    let default_registry_settings = RegistrySettings::default();
-    let registry_settings = all_registry_settings
-        .get(server)
-        .unwrap_or(&default_registry_settings);
-
-    match actor_ref {
-        ResourceRef::File(actor_ref) if allow_file_load => {
+    match ResourceRef::try_from(actor_ref.as_ref())? {
```
connorsmith256, line 105: $0.02: With the additional nesting, branching logic, indentation, and line breaks, I find these match blocks harder to follow after this refactor

connorsmith256 (approved):

crates/host/src/oci.rs
rvolosatovs: Yeah, the `OciConfig` should be propagated here, but it's lost along the way.
There's still a bunch of work to do here

crates/host/src/bindle.rs
```diff
@@ -36,12 +35,36 @@ const KEYRING_FILE: &str = "keyring.toml";
...
-    None(NoToken),
     /// HTTP authentication
     Http(HttpBasic),
     /// Token authentication
     LongLived(LongLivedToken),
+    /// No authentication
+    None(NoToken),
+}
+
+impl Default for Auth {
```
thomastaylor312, line 46: Small nit: Probably better to derive default and then just use the `#[default]` attr to pick this one

crates/host/src/lib.rs
thomastaylor312: I agree with this, but also consider it a nit as well. Generally I prefer readability over conciseness (and it can go too far in either direction)

thomastaylor312 (approved):
All things are minor nits. Good work here

crates/host/src/bindle.rs
rvolosatovs: That was not possible with the existing code, but I refactored the enum to allow for this

crates/host/src/lib.rs
rvolosatovs: @connorsmith256 I think this should not be addressed

crates/host/src/lib.rs
rvolosatovs: It's now a method on the ref!

crates/host/src/oci.rs
rvolosatovs: Updated, PTAL, I think you'll like the new approach

brooksmtownsend (approved):

PR #567, opened 2023-08-24
feat: implement `wasi:blobstore`

rvolosatovs:
## Feature or Problem

Implement `wasi:blobstore` sans 2 methods not supported by existing providers.
Testing also would need some love, but let's take care of that in a follow-up - will file an issue

## Related Issues

#421 
https://github.com/wasmCloud/wasi-blob-store/pull/3
https://github.com/WebAssembly/wasi-blob-store/pull/11

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`


crates/host/src/wasmbus/mod.rs
```diff
@@ -305,6 +306,280 @@ impl Handler {
...
+                        offset: 0,
+                        is_last: true,
+                    },
+                    ..Default::default()
+                },
+            )
+            .await?;
+        let wasmcloud_compat::blobstore::PutObjectResponse { stream_id } =
+            decode_provider_response(res)?;
+        ensure!(stream_id.is_none());
```
connorsmith256, line 494: Nit: add a string here explaining that multipart uploads aren't supported(?)

rvolosatovs: Do you mean a comment?
Because this is not user-facing (or at least it should not be)

rvolosatovs: Do providers return a `stream_id` even when the complete blob is sent?

rvolosatovs: looks like it should not https://github.com/wasmCloud/wasmCloud/blob/a82847f4e4a2637bc9d6f9c75be0a902f7944212/crates/providers/blobstore-fs/src/bin/blobstore_fs.rs#L590-L599

connorsmith256 (approved):

Issue #568, opened 2023-08-24
Make OTEL settings discoverable


connorsmith256:
As of https://github.com/wasmCloud/wasmCloud/pull/565, the host will send OTEL configuration to providers, based off env vars provided to the host. While this is a "standard" way of specifying OTEL config, not everyone who wants tracing configured will know that. We should explore making this information more discoverable to users. Some ideas:
- add config to the CLI arguments
- use custom help text on the host?

Note that a documentation page [exists already](https://wasmcloud.com/docs/production/tracing#tracing-setup-for-your-project)

brooksmtownsend:
This might be a scenario where we draw the line in the sand for what we add to the CLI configuration, so I'm totally open to debate on this one. Perhaps documentation is sufficient here for anyone configuring tracing, I just want to avoid a scenario where we're looking for an environment variable in the codebase and that makes it difficult to know how to enable a feature

Iceber:
I see that provider-sdk in pr is no longer compatible with environment variables, which will result in products created by provider-sdk not being able to use OTEL on wasmcloud-otp, and providers created by wasmbus-rpc will not be able to use OTEL on WasmCloud.

Of course, if the above mentioned repositories are updated in a unified way, they will be compatible with each other, but of course user-defined providers will still need to upgrade their dependencies.

Can our plan users choose to use WasmClould-OTP and WasmCloud at the same time, or will WasmCloud replace WasmCloud-OTP?

connorsmith256:
Providers built using the new SDK won't be compatible with the OTP host. That's not _only_ because env vars (including OTEL-related ones) were removed, but also because the new SDK creates providers from wit definitions

However, we would like to be able to continue running legacy providers on the Rust host, at least for a while. This is a good callout and is something I should be able to handle in https://github.com/wasmCloud/wasmCloud/pull/572. We will have the host pass the two OTEL environment variables to legacy providers 👍

Issue #563, opened 2023-08-24
[Rust Host] Emit logs to stderr


connorsmith256:
Currently the host emits logs to stdout. Instead, the host should emit logs to stderr by default

PR #572, opened 2023-08-25
feat: support OTEL traces end-to-end

connorsmith256:
## Feature or Problem
This PR:
- adds OTEL config values to the CLI args for the host
- sets up tracing for the host
- proxies OTEL config values to providers via `HostData`
- uses a strongly-typed `HostData` in the host to pass data to providers
- adds a `core` crate for types shared across the wasmCloud ecosystem (primarily between host and provider)
- adds a `tracing` crate for setting up tracing and extracting/injecting trace context on invocations (this code was migrated from the provider SDK)

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/568
Resolves https://github.com/wasmCloud/wasmCloud/issues/577
Resolves https://github.com/wasmCloud/wasmCloud/issues/563

## Release Information
Next

## Consumer Impact
Providers built against the updated provider SDK will need to run on a host that supplies `OtelConfig`

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
I set up tempo locally along with the kvcounter actor + HTTP server and KV Redis provider built using these changes. Here is an example where you can see spans from the host and both providers as part of a round-trip invocation:

```
docker exec -it 8a6428c77728 /bin/sh -c "wget -q -O - localhost:8000/api/traces/cef9a8119b196d111bc2f5956f22cd4e" | jq -r '.batches | .[] | .instrumentationLibrarySpans | .[] | .spans | .[] | .name' | tr '\n' ' '

increment dispatch publish rpc instantiate instantiate instantiate write_bytes write_bytes guest_request read_bytes read_string read_bytes read_string read_bytes read_string read_bytes identify_wasmbus_target identify_wasmbus_target call_operation_with_payload call_sync handle set_host_response host_call host_response_len write_bytes host_response read_bytes guest_response call call handle handle_invocation handle_call request inner_rpc call_actor http request request 
```


crates/tracing/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    Resource,
+};
+#[cfg(feature = "otel")]
+use opentelemetry::trace::TraceError;
+#[cfg(feature = "otel")]
+use opentelemetry_otlp::{Protocol, WithExportConfig};
+use tracing::{Event, Subscriber};
+use tracing_subscriber::fmt::format::{DefaultFields, JsonFields};
+use tracing_subscriber::fmt::{
+    format::{Format, Full, Json, Writer},
```
rvolosatovs, line 17: nit; this should have been a single `use` statement

crates/tracing/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl<'a> Drop for LockedWriter<'a> {
+    fn drop(&mut self) {
+        let _ = self.flush();
+    }
+}
+
+lazy_static::lazy_static! {
+    static ref STDERR: std::io::Stderr = std::io::stderr();
+}
```
rvolosatovs, line 62: Can we migrate to `once_cell` please? I'm fine with adding a TODO here and in `Cargo.toml`

crates/tracing/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[features]
+default = []
+otel = ["opentelemetry", "tracing-opentelemetry", "opentelemetry-otlp"]
+
+[dependencies]
+lazy_static = { workspace = true }
```
rvolosatovs, line 16: Add a TODO to remove (or replace by `once_cell` already now)

crates/provider-sdk/src/lib.rs
```diff
@@ -19,6 +16,12 @@ pub use provider::ProviderConnection;
 pub use provider_main::{load_host_data, run_provider, start_provider};
 pub use rpc_client::RpcClient;
 
+pub mod core {
+    pub use wasmcloud_core::{
+        HealthCheckRequest, HealthCheckResponse, HostData, LinkDefinition, WasmCloudEntity,
+    };
+}
```
rvolosatovs, line 23: I would expect this to be removed and any usages of these to be replaced by imports from `wasmcloud_core`

crates/host/src/wasmbus/mod.rs
```diff
@@ -265,71 +219,72 @@ fn invocation_hash(
...
-            encoded_claims,
-            host_id,
-        })
-    }
+/// Creates a new invocation. All invocations are signed with the host key as a way
+/// of preventing them from being forged over the network when connected to a lattice,
+/// so an invocation requires a reference to the host (signing) key
+#[allow(clippy::implicit_hasher)]
+#[allow(clippy::missing_errors_doc)]
+pub fn new_invocation(
```
rvolosatovs, line 227: Why the change from `Invocation::new`?

crates/host/src/wasmbus/config.rs
```diff
@@ -5,6 +5,7 @@ use nkeys::KeyPair;
 use url::Url;
 
 use crate::oci::Config as OciConfig;
+use wasmcloud_core::{logging::Level as LogLevel, OtelConfig};
```
rvolosatovs, line 8: This belongs in a separate import group

crates/host/src/wasmbus/mod.rs
connorsmith256: There were types referenced in `new` that weren't easily moved into `wasmcloud_core`. I refactored the first commit to account for this, so it's back to `Invocation::new`

crates/provider-sdk/src/lib.rs
connorsmith256: I was conflicted on this. Do you think it's better for providers to explicitly depend on `wasmcloud_core`, or re-export the types providers need so they only have to depend on `wasmcloud_provider_sdk`?

rvolosatovs: I'd definitely re-export, but I would not do the inner nesting, i.e. I would do `pub use wasmcloud_core as core` and call it a day

connorsmith256: Fixed

PR #574, opened 2023-08-25
update limits of ActorInstancePool when scaling actor

Iceber:
## Feature or Problem

The pool limit should be the same as the number of scale

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1885,6 +1885,7 @@ impl Host {
                 let current = instances.values().map(Vec::len).sum();
                 let claims = actor.pool.claims().context("claims missing")?;
                 if let Some(delta) = count.checked_sub(current).and_then(NonZeroUsize::new) {
+                    actor.pool.set_limit(NonZeroUsize::new(count)).await;
                     let mut delta = self
                         .instantiate_actor(
                             claims,
@@ -1926,6 +1927,8 @@ impl Host {
                     if remaining == 0 {
                         drop(instances);
                         entry.remove();
+                    } else {
+                        actor.pool.set_limit(NonZeroUsize::new(count)).await;
                     }
                 }
             }


rvolosatovs (changes requested):
The instance pool is a really low-level structure, which, at the present time, I don't think we should touch - we explicitly *want* the pooled instance amount to be more than the "scale", at least because we cannot put instantiated components back to the pool yet (that should change in the future as component model support in Wasmtime is better supported)

We also would want the amortization in possible actor start "burst" scenario at scale.

I think that if we do want to limit the pool size (and it's waaaaaay too early to do that now, we have to profile and benchmark first), we'd set it to a constant value we determine during profiling

Iceber:
> The instance pool is a really low-level structure, which, at the present time, I don't think we should touch - we explicitly *want* the pooled instance amount to be more than the "scale", at least because we cannot put instantiated components back to the pool yet (that should change in the future as component model support in Wasmtime is better supported)
> 
> We also would want the amortization in possible actor start "burst" scenario at scale.
> 
> I think that if we do want to limit the pool size (and it's waaaaaay too early to do that now, we have to profile and benchmark first), we'd set it to a constant value we determine during profiling

 I see that it is put back into the pool when drop the instance, I may have misinterpreted it
https://github.com/wasmCloud/wasmCloud/blob/04fb0e1befc097c00aeabba29adfcf944ea44a3b/crates/runtime/src/actor/mod.rs#L570-L575

Iceber:
I'm testing it.

rvolosatovs:
> > The instance pool is a really low-level structure, which, at the present time, I don't think we should touch - we explicitly _want_ the pooled instance amount to be more than the "scale", at least because we cannot put instantiated components back to the pool yet (that should change in the future as component model support in Wasmtime is better supported)
> > We also would want the amortization in possible actor start "burst" scenario at scale.
> > I think that if we do want to limit the pool size (and it's waaaaaay too early to do that now, we have to profile and benchmark first), we'd set it to a constant value we determine during profiling
> 
> I see that it is put back into the pool when drop the instance, I may have misinterpreted it
> 
> https://github.com/wasmCloud/wasmCloud/blob/04fb0e1befc097c00aeabba29adfcf944ea44a3b/crates/runtime/src/actor/mod.rs#L570-L577

Instances on which component model bindings were instantiated are not put back into the pool, because the underlying instance is moved in the process - that is a limitation of current Wasmtime implementation.

We should not limit the pool size yet

Iceber:
@rvolosatovs Thanks

PR #576, opened 2023-08-25
Add ADR for detached UI

### LINKED ISSUE #321: [RFC] Decouple the Washboard from the Host
## Summary
This request for comment seeks commentary on the idea of extricating the wasmCloud dashboard (aka `washboard`) functionality from the OTP host and ensuring that functionality is held elsewhere. Note that the washboard functionality will not be retired or deleted, only made available on its own without requiring it be embedded in a host.

## Rationale
The first and probably easiest to identify issue with the status quo is that both the core wasmCloud host (`host_core`) and the Phoenix server that provides the dashboard functionality (`wasmcloud_host`) exist in the same distribution. This means that, by default, every deployed version of a wasmCloud host carries with it the extra dependencies, size, and networking port requirements of a full web application.

This tight coupling presents problems above and beyond the deployment footprint. In our recent efforts to create static builds, having the Phoenix application be a required piece of the target output makes the pipelines more difficult. This tight coupling also means that every time we want to change, enhance, or fix the `washboard`, we have to release a new version of the host, even though the two logical components should be able to have their own independent release cadences.

A fairly unknown and esoteric bit of code in the OTP host actually involves a special internal pub/sub mechanism that compensates for the dashboard being within the same process as a running host. In other words, we're actually writing code to counteract negative consequences of the pairing. There are code paths we actually need to test that only arise from having the dashboard and core host in the same process.

## Implementation
This section enumerates a number of possible implementations of the washboard functionality as a separate entity from the host. In all of the below cases, washboard will be removed from the OTP host. This is an objective list with identified, practical pros and cons.

Implicit in the implementation is the addition of support in the washboard UI for interacting with [wadm](https://github.com/wasmcloud/wadm)

### Embed React App in Rust+Hyper
This implementation option would follow the same pattern we use for things like our [petclinic](https://github.com/wasmCloud/examples/tree/main/petclinic) UI. Embed the various bundled UI assets inside a Rust binary, and use a built-in web server like the Hyper crate to serve up the assets. Hyper would then also have to expose an API that the React application would use.

#### Pros
* Easy distribution - single Rust binary, easily cross-compiled
* React ecosystem is robust and accessible
* Leverage pre-existing team skills

#### Cons
* We'd have to do the heavy lifting on building websockets+realtime into the server, proxying NATS
  * Potential counter-point is to use NATS native websockets endpoint 
* Non-traditional React development environment could cause friction/frustration and take more effort than desired 

### Yew and Tauri
[Yew](https://yew.rs/) is a Rust-based web front end framework that has undergone exponential growth since its original creation. It gives developers a familiar component-based UI framework while also using pure Rust. The [Tauri](https://tauri.app/) project takes pretty much any front-end (including Yew) and bundles it all up inside an easily distributed Rust binary.

#### Pros
* Tauri takes away the manual bundling complexity from the React+Hyper manual embedding option
* Rust all the way down, leveraging team skills

#### Cons
* Tauri is still in its infancy, and Yew, despite its age, is not yet 1.0. Both projects are, despite their popularity within the Rust community, relatively niche solutions

### Wash WASI Plugin
This option involves sitting on top of [wash plugins](https://github.com/wasmCloud/wash/issues/542) implemented as WASI components. Wash, which would serve as an ephemeral bespoke host, would provide this plugin with the `lattice-controller` provider suitably configured with a NATS connection, as well as a bound HTTP server provider on a port. The actor component would then respond to HTTP requests and interact with the lattice via the lattice controller provider.

#### Pros
* Absolute bleeding edge
* Easy portable distribution
* Looks like it's part of wash

#### Cons
* Requires the wash plugin RFC to be built
* Relies on stability in the component model
* Would require Rust implementation of the plugin
* Likely requires completely one-off way of embedding the app assets (js/css/html/etc)

### Phoenix (Elixir) App
This involves extracting the `wasmcloud_host` top-level OTP application, removing the tentacles that require the underlying core host, and building and distributing it as a separate application.

#### Pros
* Would take "almost no work", the web UI is already done
* OTP wasmcloud_host application already has a NATS connection and client

#### Cons
* Still would have the burden of building and deploying OTP applications
* The web UI is stale and outdated and doesn't properly handle many hosts (e.g. it's semi-tightly-coupled to the old days when it administered one host at a time)

### Flutter App
[Flutter](https://flutter.dev/) is a cross-platform framework that allows you to define your user interface in terms of platform agnostic primitives and it takes care of building the target-native binary, including iOS, Mac, Linux, Windows, and Android. This should **not** be confused with other bundling tools that take a web application, cram them into a bundle, and then run them inside an iframe in a native browser component.

The flutter app would communicate with NATS over WebSockets

#### Pros
* Access to rich, interactive, reactive UIs that can be optimized for many different platforms and screens
* Platform-ready easily-distributed binaries (Flutter builds the platform distros for you)
* Unlock some potentially very cool user experiences around things like tablets, etc. Who doesn't want to walk around a coffee shop while administering their lattice on an iPad?
* Easier for "regular devs" to write good UIs, lowers barrier
* Super popular right now with a massive ecosystem
  * pub.dev - massive repository of reusable bits 
* Could be seamlessly launched by wash, e.g. `wash console`

#### Cons
* New technology, lack of team familiarity
  * Starting from scratch, can't reuse existing Rust or Elixir libraries

## Risks
Typically the biggest risk is time. We could invest a large amount of time in this effort and ultimately not please the wasmCloud community. Otherwise this seems pretty low risk. The idea that the admin dashboard should be decoupled from the actual server daemons is fairly uncontroversial. The only real risk is in being unable to decide on which of the alternatives we want to use.

lachieh:
Closes #321

### DIFF
--- a/adr/0014-detach-ui-from-host.md
+++ b/adr/0014-detach-ui-from-host.md
@@ -1,0 +1,76 @@
+# Detach UI from Host
+
+| Status | Deciders | Date |
+|--|--|--|
+| accepted | wasmCloud Maintainers | 07-12-2023 |
+
+## Context and Problem Statement
+
+With the decision to transition the primary host to Rust, we have the ability to rethink the Washboard UI for the host. The current UI is attached directly to the OTP host, and is written in Elixir. This means that the UI can only be used when an OTP host is running, and that the UI is not portable to other hosts.
+
+## Decision Drivers <!-- optional -->
+
+* OTP host is being deprecated
+* UI should be compatible with all hosts
+* Community contributions to the UI should be easy
+* UI should be easy to run anywhere
+
+## Considered Options
+
+* Static HTML/JS/CSS
+* Rust based UIs (Yew, Seed, etc)
+* Wash WASI Plugin
+* Phoenix (Elixir) App
+* Platform-agnostic App (Flutter, React Native, etc)
+* Web wrapper App (Electron, Tauri, etc)
+
+## Decision Outcome
+
+The decision is to use a static HTML/JS/CSS UI. This will allow the UI to be run anywhere that can display HTML, and will allow the UI to be easily contributed to by the community. The UI will be written in React, Typescript, and Tailwind CSS. The build will be handled with Vite, though this is an implementation detail.
+
+### UI Framework
+
+There are a number of UI frameworks that could be used to build the UI. The most popular of these are React, Vue, and Svelte. Angular is not an option given the majority negative sentiment amongst the wider community. The decision to use React was made based on the fact that currently has the widest number of contributors and the largest ecosystem of libraries. Svelte was a close second, but the ecosystem is not as mature, and the community is not as large as React's.
+
+### Styling and Component Library
+
+For styling we wanted the most flexibility with as much core functionality covered as possible. For that reason we decided to use Radix-UI along with Tailwind CSS. This means that accessibility is a primary consideration while allowing for a large number of components to be used out of the box that can be styled to fit within the wasmCloud brand guidelines.
+
+### Build tooling
+
+The build tooling is not a primary consideration, but it is important to note that the build tooling should be as simple as possible. The decision to use Vite was made because it is a simple build tool that is easy to configure because of with opinionated defaults and has a large number of plugins available given that it is mostly built on top of Rollup.
+
+## Pros and Cons of the Options
+
+### Static HTML/JS/CSS
+
+This consideration was not initially brought up in the RFC, but came out through the discussion of the other options. Many of the other options require some sort of HTML/JS/CSS UI to be built, and their benefits come only with how the project is distributed. Almost every device that can run a UI already has a browser installed, and the flexibility of where the UI is hosted cannot be understated.
+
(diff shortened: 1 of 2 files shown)


connorsmith256 (approved):
Thank you!

rajsite:
Can't fault the decision but would have been awesome to land on Web Components for the UI (via lit, microsoft fast, shoelace, stencil, etc) to parallel the awesome work being done with WASM Components for the backend.

Componentize all the things on top of standards! 🧩🧹

brooksmtownsend:
> Can't fault the decision but would have been awesome to land on Web Components for the UI (via lit, microsoft fast, shoelace, stencil, etc) to parallel the awesome work being done with WASM Components for the backend.
> 
> 
> 
> Componentize all the things on top of standards! 🧩🧹

I see no reason to keep this out of our future looks! 🔮 This is a very fun idea. I think for the core dashboard itself we landed in the right place to attract contributors and iterate quickly. But this sounds like a fantastic hackathon project that could turn into the main way to deploy the dashboard if it performs well!

Issue #482, opened 2023-08-15
Implement chunking for large RPC payloads


brooksmtownsend:
For any payload >1MB, it should be chunked using the jetstream object store to avoid NATS message limits.

PR #579, opened 2023-08-27
Feat/chunking

connorsmith256:
## Feature or Problem
**Note:** left in draft until https://github.com/wasmCloud/wasmCloud/pull/572 is merged, since this is based on top of it

This PR adds chunking to RPC messages over the size limit (900KB)

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/482

## Release Information
Next

## Consumer Impact
"Large" RPC payloads should now work in the new host

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
I updated the test actors to respond with a "large" payload (~10KB), in conjunction with lowering the chunking threshold in tests to 1KB

### Manual Verification
I also temporarily updated the HTTP server provider to include a "large" header (~10KB) in the invocation it makes to the echo actor, and confirmed the host was able to process this invocation successfully


crates/host/src/wasmbus/mod.rs
```diff
@@ -492,31 +281,44 @@ impl Handler {
...
-            ensure!(content_length == msg.len(), "message size mismatch");
+
+        let resp_length = usize::try_from(content_length.unwrap_or_default())
+            .context("content length does not fit in usize")?;
+        if needs_chunking(resp_length) {
+            msg = self
+                .chunk_endpoint
+                .get_unchunkified(&invocation_id)
+                .await
+                .context("failed to dechunk response")?;
```
rvolosatovs, line 309: There's no need to call `needs_chunking` with a `content_length` set to `None`, so please bring the `if let Some....` block back

crates/host/src/wasmbus/mod.rs
```diff
@@ -627,30 +446,41 @@ impl Bus for Handler {
...
                     .context("failed to decode invocation response")
                     .map_err(|e| e.to_string())?;
                 if invocation_id != invocation.id {
                     return Err("invocation ID mismatch".into());
                 }
-                if let Some(content_length) = content_length {
-                    let content_length = usize::try_from(content_length)
-                        .context("content length does not fit in usize")
+
+                let resp_length = usize::try_from(content_length.unwrap_or_default())
```
rvolosatovs, line 471: content_length unwrapped to a 0 is really confusing

crates/host/src/wasmbus/mod.rs
```diff
@@ -877,7 +707,7 @@ impl Messaging for Handler {
         }
     }
 
-    #[instrument]
+    #[instrument(skip(self, __arg1))]
```
rvolosatovs, line 710: What is this trying to achieve? Maybe just `skip_all`?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2326,46 +2200,72 @@ impl Host {
...
+                link_name: link_name.to_string(),
+                lattice_rpc_user_jwt: self.host_config.prov_rpc_jwt.clone().unwrap_or_default(),
+                lattice_rpc_user_seed: lattice_rpc_user_seed.unwrap_or_default(),
+                lattice_rpc_url: self.host_config.prov_rpc_nats_url.to_string(),
+                env_values: vec![],
+                instance_id: Uuid::from_u128(id.into()).to_string(),
+                provider_key: claims.subject.clone(),
+                link_definitions,
+                config_json: configuration,
+                default_rpc_timeout_ms: Some(
```
rvolosatovs, line 2215: This belongs on a separate line (`let default_rpc_timeout = ...`)

crates/host/src/wasmbus/mod.rs
```diff
@@ -2326,46 +2200,72 @@ impl Host {
...
+                        .rpc_timeout
+                        .as_millis()
+                        .try_into()
+                        .context("failed to convert rpc_timeout to u64")?,
+                ),
+                cluster_issuers: self.cluster_issuers.clone(),
+                invocation_seed,
+                log_level: Some(self.host_config.log_level.clone()),
+                structured_logging: self.host_config.enable_structured_logging,
+                otel_config: OtelConfig {
```
rvolosatovs, line 2226: Please extract on a separate line

crates/host/src/wasmbus/mod.rs
```diff
@@ -3218,6 +3113,33 @@ fn human_friendly_uptime(uptime: Duration) -> String {
     .to_string()
 }
 
+fn to_core_entity(
```
rvolosatovs, line 3116: Looking at the name of this function, its purpose is not clear

crates/core/src/chunking.rs
```diff
@@ -15,46 +15,51 @@
...
+#[cfg(not(test))]
 const CHUNK_THRESHOLD_BYTES: usize = 1024 * 900; // 900KB
+#[cfg(test)]
+const CHUNK_THRESHOLD_BYTES: usize = 1024 * 1; // 1KB
 
 /// check if message payload needs to be chunked
-pub(crate) fn needs_chunking(payload_size: usize) -> bool {
+pub fn needs_chunking(payload_size: usize) -> bool {
     payload_size > CHUNK_THRESHOLD_BYTES
 }
```
rvolosatovs, line 43: Why is a function necessary for this, wouldn't it be much clearer and useful to just export the constant?

crates/host/src/wasmbus/mod.rs
connorsmith256: This skips the `body` field on `BrokerMessage`. Since we're destructuring that arg, I can't do `skip(self, body)`, but `__arg1` allows me to skip `BrokerMessage`. The alternative was not destructuring. Would you prefer that?

crates/core/src/chunking.rs
connorsmith256: I think the function call is more clear, and this was existing code, but I don't care strongly either way. I'll change callers to compare against `CHUNK_THRESHOLD_BYTES`

crates/host/src/wasmbus/mod.rs
connorsmith256: I agree, I'm not happy with the name. Do you have an idea for a better name?

crates/host/src/wasmbus/mod.rs
connorsmith256: See above response and let me know whether that resolves this

crates/host/src/wasmbus/mod.rs
rvolosatovs: Doesn't it depend on provider implementation?
If we expect `content_length` to always be set here, then I'm fine with the type change and failing when content length is not set

rvolosatovs: actually, since we enforce that the payload length is `0` if the field is not set, I'm ok with this

rvolosatovs (approved):

crates/host/src/wasmbus/mod.rs
connorsmith256: No, invocations/invocation responses shouldn't be manually constructed by a provider implementation; it should be common to the provider SDK, e.g. https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-sdk/src/rpc_client.rs#L178

I changed the type, and tested with legacy HTTP and Redis KV providers 👍

rvolosatovs (approved):

crates/core/src/chunking.rs
```diff
@@ -15,46 +15,46 @@
...
     object_store::{Config, ObjectStore},
-    Context,
+    Context as JetstreamContext,
 };
 use futures::TryFutureExt;
 use tokio::io::{AsyncRead, AsyncReadExt};
 use tracing::{debug, error, instrument};
 
-use crate::error::{InvocationError, InvocationResult};
+/// Amount of time to add to rpc timeout if chunkifying
```
thomastaylor312, line 30: Nit: Maybe add a note about why 13

crates/host/src/wasmbus/mod.rs
```diff
@@ -1676,7 +1754,7 @@ impl Host {
 
     /// Instantiate an actor and publish the actor start events.
     #[allow(clippy::too_many_arguments)] // TODO: refactor into a config struct
-    #[instrument(skip(self, host_id, actor_ref))]
+    #[instrument(skip(self, claims, annotations, host_id, actor_ref, pool, handler))]
```
thomastaylor312, line 1757: Super nit: Maybe just `skip_all` and add back in the fields you need?

thomastaylor312 (approved):

PR #597, opened 2023-08-30
refactor: flatten optional annotations to always be set

connorsmith256:
## Feature or Problem
This PR flattens uses of options for annotations, both for the host's internal `Annotations` type (`BTreeMap<String, String>`) as well as incoming annotations from the control interface client (`HashMap<String, String>`)

This is to avoid cases where e.g. a provider is started via a request where `"annotations" :{}`, but then the user tries to stop the provider via a request where `"annotations": null` (or where `annotations` is omitted entirely)

The host will now treat an omitted annotations field, null annotations, and an empty annotations map on requests the same way

The host will now always emit an empty annotations map in events as well. Since consumers of the events previously had to handle the omitted field, null, or an empty map, this change is backwards-compatible

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
Existing tests still pass

### DIFF
--- a/tests/wasmbus.rs
+++ b/tests/wasmbus.rs
@@ -1077,7 +1077,7 @@ expected: {expected_name:?}"#
                 compat_instances.is_empty(),
                 "more than one compat actor instance found"
             );
-            ensure!(annotations == None);
+            ensure!(annotations == Some(HashMap::default()));
             ensure!(Uuid::parse_str(&instance_id).is_ok());
             ensure!(revision == expected_revision.unwrap_or_default());
 
@@ -1109,7 +1109,7 @@ expected: {expected_name:?}"#
                 component_instances.is_empty(),
                 "more than one component actor instance found"
             );
-            ensure!(annotations == None);
+            ensure!(annotations == Some(HashMap::default()));
             ensure!(Uuid::parse_str(&instance_id).is_ok());
             ensure!(revision == expected_revision.unwrap_or_default());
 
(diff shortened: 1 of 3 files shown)


tests/wasmbus.rs
```diff
@@ -1242,7 +1242,7 @@ expected: {expected_labels:?}"#
     match (providers.pop(), providers.as_slice()) {
         (Some(nats), []) => {
             // TODO: Validate `constraints`
-            ensure!(nats.annotations == None);
+            ensure!(nats.annotations.unwrap_or_default() == HashMap::new());
```
rvolosatovs, line 1245:
```suggestion
            ensure!(nats.annotations.is_empty());
```

tests/wasmbus.rs
```diff
@@ -1207,7 +1207,7 @@ expected: {expected_name:?}"#
             ensure!(httpserver.revision == 0);
 
             // TODO: Validate `constraints`
-            ensure!(kvredis.annotations == None);
+            ensure!(kvredis.annotations.unwrap_or_default() == HashMap::new());
```
rvolosatovs, line 1210:
```suggestion
            ensure!(kvredis.annotations.is_empty());
```

tests/wasmbus.rs
```diff
@@ -1198,7 +1198,7 @@ expected: {expected_name:?}"#
             ensure!(blobstore_fs.revision == 0);
 
             // TODO: Validate `constraints`
-            ensure!(httpserver.annotations == None);
+            ensure!(httpserver.annotations.unwrap_or_default() == HashMap::new());
```
rvolosatovs, line 1201:
```suggestion
            ensure!(httpserver.annotations.is_empty());
```

rvolosatovs (approved):

PR #598, opened 2023-08-30
ci: setup release process

rvolosatovs:
## Feature or Problem

Push multi-arch Docker images tagged by the commit rev on each push to this repo and each workflow dispatch. Tag pushes will be handled later as part of the release process.

## Related Issues

just like https://github.com/bytecodealliance/wit-deps/pull/54 https://github.com/bytecodealliance/wit-deps/pull/55

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

https://github.com/rvolosatovs/wasmCloud/actions/runs/6037286627
https://github.com/rvolosatovs/wasmCloud/pkgs/container/wasmcloud/123726133?tag=0.78.0-rc1

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -3,6 +3,11 @@ name: wasmCloud
 on:
   pull_request:
   push:
+    branches:
+    - main
+    tags:
+    - 'v[0-9].[0-9]+.[0-9]+'
+    - 'v[0-9].[0-9]+.[0-9]+-rc[0-9]+'
   workflow_dispatch:
 
 concurrency:
@@ -81,6 +86,16 @@ jobs:
         name: wasmcloud-universal-darwin
         path: wasmcloud-universal-darwin
 
+  test-linux:
+    runs-on: ubuntu-22.04
+    needs: build-bin
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: wasmcloud-x86_64-unknown-linux-musl
+    - run: chmod +x ./wasmcloud-x86_64-unknown-linux-musl
+    - run: ./wasmcloud-x86_64-unknown-linux-musl --version
+
   test-windows:
     runs-on: windows-2022
     needs: build-bin
@@ -127,7 +142,7 @@ jobs:
           <meta charset="utf-8">
           <title>Redirecting to wasmcloud/index.html</title>
           <meta http-equiv="refresh" content="0; URL=wasmcloud/index.html">
-          <link rel="canonical" href="https://${{ github.repository.owner }}.github.io/wasmCloud/wasmcloud/index.html">
+          <link rel="canonical" href="https://${{ github.repository_owner }}.github.io/wasmCloud/wasmcloud/index.html">
         EOF
     - uses: actions/upload-pages-artifact@v2.0.0
       with:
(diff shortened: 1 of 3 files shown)


.github/workflows/wasmcloud.yml
```diff
@@ -81,6 +81,40 @@ jobs:
...
+    - uses: docker/login-action@v2
+      if: ${{ github.event_name != 'pull_request' }}
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - run: nix run -L .#build-wasmcloud-oci wasmcloud
+    - run: nix profile install --inputs-from . 'nixpkgs#buildah'
+      if: ${{ github.event_name != 'pull_request' }}
+    - run: buildah push --storage-driver=vfs --all --format 'v2s2' wasmcloud docker://ghcr.io/${{ github.repository_owner }}/wasmcloud:${{ github.sha }}
```
connorsmith256, line 103: I'm not familiar with this. Is there a reason we can't push to wasmcloud.azurecr.io, where our release images are?

.github/workflows/wasmcloud.yml
```diff
@@ -81,6 +81,40 @@ jobs:
         name: wasmcloud-universal-darwin
         path: wasmcloud-universal-darwin
 
+  oci:
```
connorsmith256, line 84: Can we gate this whole job instead of only some of the steps below?

.github/workflows/wasmcloud.yml
```diff
@@ -81,6 +81,40 @@ jobs:
...
+    permissions:
+      packages: write
+    needs: build-bin
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./.github/actions/install-nix
+      with: 
+        cachixAuthToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+    - uses: docker/login-action@v2
+      if: ${{ github.event_name != 'pull_request' }}
```
connorsmith256, line 95: I tried searching for the docs. Is this the right page? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

If so it looks like there are a lot of event names we don't want to match on. Do we want to run for any other reason than merging to main, i.e. `${{ github.ref == 'refs/heads/main' }}` ?

connorsmith256 (dismissed):

.github/workflows/wasmcloud.yml
rvolosatovs: We have already agreed to push to GitHub registry and Docker Hub.
Docker Hub will be the next step.
Please file an issue for pushing to azure

.github/workflows/wasmcloud.yml
rvolosatovs: We can, but we should not - we should verify that the build works on every build

.github/workflows/wasmcloud.yml
rvolosatovs: This job can only be triggered by push, PR or manual dispatch. Pushes and dispatches should trigger an upload.

Why would we want to make it harder to test this?

.github/workflows/wasmcloud.yml
connorsmith256: Can you help me understand the reasoning for this/point me to where we agreed to push to GHCR? I'm unfamiliar with `buildah` and whether changing to ACR is more than swapping the string here

.github/workflows/wasmcloud.yml
connorsmith256: Okay, how about building/running the image in one job always, and uploading an artifact we download in a push job?

The separate `if` checks on the steps makes this job difficult to follow

.github/workflows/wasmcloud.yml
rvolosatovs: We have discussed that at standup.
No, this is not much different from `docker push`

.github/workflows/wasmcloud.yml
rvolosatovs: I'll refactor the step into a multiline script, but it will only get more "difficult to follow" as we add pushing of tags and whatnot.
That's just how GitHub Actions work and there's nothing we can do about that.
What you are suggesting would not work

.github/workflows/wasmcloud.yml
brooksmtownsend: From what I remember from the discussion, we decided on GHCR so that the image could be discoverable from the packages tab on GitHub and that all of the release assets could be centralized. Also, Dockerhub is good because when people think of a container image they go to dockerhub. I think we put AzureCR as a third priority because the content discovery isn't great there, so people would only know what's there based on what we explicitly document.

Good to know buildah is similar to a docker push so we can just add those here as we put the correct credentials into secrets.

.github/workflows/wasmcloud.yml
brooksmtownsend: What we've done here before is ensure that the job is something that triggers on tag push or manual dispatch, to make an allow list of events that should trigger the action. Based on Connor's documentation link, it does look like there's plenty of other events, how is it guaranteed that other events aren't triggering this?

I think that pushing an artifact with the commit sha on each merge to main / tag push makes sense, I just want to understand that there are not other events that can trigger this (e.g. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release)

.github/workflows/wasmcloud.yml
rvolosatovs: Yes, in fact, that's already done in the PR, we push to all three of them, e.g. https://hub.docker.com/repository/docker/rvolosatovs/wasmcloud/general ( don't have an Azure account)

.github/workflows/wasmcloud.yml
rvolosatovs: The workflow is only triggered by these three events https://github.com/wasmCloud/wasmCloud/pull/598/files#diff-b62bad75f377e25a10e7a10690fb12edea66f96e7df69d1de7873a0324f52b66R4-R11, there is not way any other event (like `release`) could appear here

brooksmtownsend (commented):
LGTM

brooksmtownsend (approved):

PR #612, opened 2023-09-01
fix: attach traces on inbound and outbound messages

connorsmith256:
Parse headers from CTL interface and RPC messages, and publish tracing headers on CTL and RPC responses

Leaving in draft for now, since there's probably a way to avoid making as many allocations as I did. But it does work:
Screenshot 2023-08-31 at 16 19 10

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -23,6 +23,7 @@ use std::env;
 use std::env::consts::{ARCH, FAMILY, OS};
 use std::io::Cursor;
 use std::process::Stdio;
+use std::str::FromStr;
 use std::sync::Arc;
 
 use anyhow::{anyhow, bail, ensure, Context as _};
@@ -67,7 +68,7 @@ use wasmcloud_runtime::capability::{
     KeyValueReadWrite, Messaging, TargetEntity, TargetInterface,
 };
 use wasmcloud_runtime::{ActorInstancePool, Runtime};
-use wasmcloud_tracing::context::{attach_span_context, OtelHeaderInjector};
+use wasmcloud_tracing::context::TraceContextInjector;
 
 const SUCCESS: &str = r#"{"accepted":true,"error":""}"#;
 
@@ -281,6 +282,8 @@ impl Handler {
             .context("failed to parse operation")?;
         let inv_target = resolve_target(target, links.get(package), &aliases).await?;
         let needs_chunking = request.len() > CHUNK_THRESHOLD_BYTES;
+        let injector = TraceContextInjector::default_with_span();
+        let headers = injector_to_headers(&injector);
         let mut invocation = Invocation::new(
             &self.cluster_key,
             &self.host_key,
@@ -288,7 +291,7 @@ impl Handler {
             inv_target,
             operation,
             request,
-            OtelHeaderInjector::default_with_span().into(),
+            injector.into(),
         )?;
 
         // Validate that the actor has the capability to call the target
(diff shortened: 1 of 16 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -3727,6 +3780,30 @@ fn ensure_actor_capability(
     Ok(())
 }
 
+fn injector_to_headers(injector: &OtelHeaderInjector) -> HeaderMap {
```
rvolosatovs, line 3783: Please spell out the whole type `async_nats::...`
Otherwise it's very opaque

crates/host/src/wasmbus/mod.rs
```diff
@@ -3727,6 +3780,30 @@ fn ensure_actor_capability(
...
+    injector
+        .as_ref()
+        .iter()
+        .filter_map(|(k, v)| {
+            // There's not really anything we can do about headers that don't parse
+            let name = HeaderName::from_str(k.as_str()).ok();
+            let value = HeaderValue::from_str(v.as_str()).ok();
```
rvolosatovs, line 3790: same as above, I'm fine with `use async_nats::header::{HeaderName, HeaderValue}` inside the function if you really want it, but I'd rather have all of these types spelled out completely

crates/host/src/wasmbus/mod.rs
```diff
@@ -3727,6 +3780,30 @@ fn ensure_actor_capability(
...
+            match (name, value) {
+                (Some(name), Some(value)) => Some((name, value)),
+                _ => None,
+            }
```
rvolosatovs, line 3794:
```suggestion
            Some((name, value))
```

crates/host/src/wasmbus/mod.rs
rvolosatovs, line 3783: Also looks like you wanted to implement `From<&OtelHeaderInjector> for HeaderMap` in `tracing` crate (it's already bloated with dependencies, but if you really want, you could make `async_nats` an optional dep there)

crates/host/src/wasmbus/mod.rs
```diff
@@ -3727,6 +3780,30 @@ fn ensure_actor_capability(
...
+        .collect()
+}
+
+fn headermap_to_headers(headermap: Option<HeaderMap>) -> HashMap<String, String> {
+    headermap
+        .unwrap_or_default()
+        .iter()
+        .map(|(k, v)| (k.to_string(), v.into()))
+        .collect()
+}
```
rvolosatovs, line 3805: It's weird for this function to take an `Option`, especially since `attach_span_context` does not do anything with an empty map, so IMO the caller should check if it's `None` and only call this when it's `Some`

crates/host/src/wasmbus/mod.rs
rvolosatovs, line 3783:
```suggestion
fn injector_to_headers(injector: OtelHeaderInjector) -> HeaderMap {
```

Looking at the usages of this, you're always creating the value and then immediately converting it, so it seems that at least this should consume the value

crates/host/src/wasmbus/mod.rs
thomastaylor312: Oh actually then we bring back in a dependency on nats in the wasmcloud_tracing crate, which is what we've been trying to avoid

crates/host/src/wasmbus/mod.rs
connorsmith256: I think Taylor alluded to this in another comment, but I kept this out of `tracing` to avoid "polluting" that crate with the async-nats deps, to avoid transitive dependency issues down the line, when/if we want to update async-nats in one place (since `tracing` is used by both the host as well as providers)

thomastaylor312 (approved):

thomastaylor312 (dismissed):

thomastaylor312 (approved):

PR #624, opened 2023-09-05
fix(host): expose registry as a public module

Iceber:
## Feature or Problem

The public function `fetch_*` takes `register::Settings` as a parameter, but the *register module* is not public, so it is now exposed so that `fetch_*` can be called externally.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/lib.rs
+++ b/crates/host/src/lib.rs
@@ -19,14 +19,14 @@ pub mod oci;
 /// wasmCloud policy service
 pub mod policy;
 
+/// Common registry types
+pub mod registry;
+
 /// Provider archive functionality
 mod par;
 
-/// Common registry types
-mod registry;
-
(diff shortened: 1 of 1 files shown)


rvolosatovs (approved):
Thanks!
Please be aware that #564 exists, which will do a bit of refactoring here

Iceber:
> Please be aware that https://github.com/wasmCloud/wasmCloud/pull/564 exists, which will do a bit of refactoring here

Thanks

Issue #633, opened 2023-09-08
`wasmcloud-host` FFI


rvolosatovs:
We had a brief discussion at WasmCon about FFI support with @charypar and it seems that we could introduce that using https://github.com/mozilla/uniffi-rs

charypar:
What I had in mind specifically was the idea of embedding a wasmCloud host in a client-side app, mobile or web, using the rust host as a library, through and FFI from Swift/Kotlin/Java/JS.

I'll try to find some time to have a play around with the idea, see if I hit any serious difficulties.

rvolosatovs:
@pgray just got NATS and Cosmonic running on Android! :tada:

pgray:
I used [termux](https://f-droid.org/en/packages/com.termux/) for running them.

compile steps:
```
git clone git@github.com:nats-io/nats-server.git && cd nats-server
CGO_ENABLED=0 GOOS=android GOARCH=arm64 go build .

# assumes android studio, ndk, etc is setup
# example env vars:
# ANDROID_SDK_ROOT=/Users/pgray/Library/Android/sdk
# ANDROID_API=33
# ANDROID_NDK_ROOT=/Users/pgray/Library/Android/sdk/ndk
cd .. && git clone git@github.com/wasmcloud/wasmcloud.git && cd wasmcloud
cargo ndk -t arm64-v8a build --release
```

rvolosatovs:
With #636 merged we now have the minimal foundation to build this feature upon, it is already possible to link to wasmCloud-the-library and start a complete host using default config, which I was able to do from C.
It, of course, gets much more complicated when we want to add things like config and/or support for lower-level functionality, @charypar would https://github.com/mozilla/uniffi-rs help us address that?

charypar:
@rvolosatovs I've been getting my head around the rust host codebase to understand what it would take to run 'as a library'. It's a learning curve, not going to lie, but I'm getting there slowly.

In terms of UniFFI, I think it can help pass more complex data across the FFI boundary for something like configuration, but its IDL (called UDL) type system [is somewhat limited](https://mozilla.github.io/uniffi-rs/udl_file_spec.html).

My open questions for embedding are largely around how much the host can (and how much it should) control the host, e.g. can it prevent it from downloading capability providers, since starting separate processes may not be possible in the 'library' scneario. 

[...]

PR #636, opened 2023-09-09
feat: add host C FFI bindings

rvolosatovs:
## Feature or Problem

just a first step in adding proper C FFI

## Related Issues

#633 

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

```
cargo build -p wasmcloud-host-sys
cc test.c -o test -lwasmcloud_host_sys -Ltarget/debug/
LD_LIBRARY_PATH=target/debug ./test
```
This showed up as a valid host

test.c:
```c
extern void run_host();

int main(void) {
    run_host();
    return 0;
}
```

### DIFF
--- a/crates/host-sys/src/lib.rs
+++ b/crates/host-sys/src/lib.rs
@@ -1,0 +1,16 @@
+use wasmcloud_host::wasmbus::{Host, HostConfig};
+
+// TODO: Proper error handling
+
+#[no_mangle]
+pub extern "C" fn run_host() {
+    tokio::runtime::Runtime::new()
+        .expect("failed to create runtime")
+        .block_on(async {
+            let (host, shutdown) = Host::new(HostConfig::default())
+                .await
+                .expect("failed to initialize host");
+            host.stopped().await.expect("failed to await host stop");
+            shutdown.await.expect("failed to shutdown host");
+        })
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -524,6 +524,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "cbindgen"
+version = "0.25.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "faeaa693e5a727975a79211b8f35c0cb09b031fdb6eaa4a788bc6713d01488ca"
+dependencies = [
+ "heck",
+ "indexmap 1.9.3",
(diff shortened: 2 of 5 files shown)


crates/host-sys/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+use wasmcloud_host::wasmbus::{Host, HostConfig};
+
+// TODO: Proper error handling
+
+#[no_mangle]
+pub extern "C" fn run_host() {
+    tokio::runtime::Builder::new_multi_thread()
+        .enable_time()
+        .enable_io()
```
vados-cosmonic, line 9: `.enable_all()` would do both here

crates/host-sys/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+use wasmcloud_host::wasmbus::{Host, HostConfig};
+
+// TODO: Proper error handling
```
brooksmtownsend, line 3: YOLO

brooksmtownsend (dismissed):
This looks good to me 👍🏻

crates/host-sys/src/lib.rs
brooksmtownsend: Only concern there would be the doc comment
```
If additional components are added to Tokio in the future, enable_all will include these future components.
```

That also would mean we get additional components without changing the code, depends on what we'd rather have

vados-cosmonic: yeah that's definitely true -- if a breaking change were added to `tokio` this might need to be changed when upgrading the dependency could be extra work to do/needing to be careful around the upstream releases

rvolosatovs: I just switched to https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.new, which should enable everything we need here

vados-cosmonic (approved):
thanks for addressing the nit!

brooksmtownsend (approved):

PR #606, opened 2023-08-31
feat: ignore `stop_provider` annotations

rvolosatovs:
## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

PR #656, opened 2023-09-18
[FEAT] Support annotation filters

### LINKED ISSUE #607: [FEAT] Treat annotations on `stop` commands as a filter, rather than an exact match
Relevant after #606 

## Background

When you start actors and capability providers on wasmCloud hosts, you have the option of attaching additional annotations to those individual actors/providers (abbreviated as components for future references.) This is treated as additional metadata for the instance and persisted as long as the instance runs. [wadm](https://github.com/wasmcloud/wadm) uses annotations in order to uniquely identify instances that are being managed by wadm so that it can only start and stop resources that it's managing, without affecting developer started components.

When stopping a component, you can supply a map of annotations to _exactly match_ the resource you want to stop. This is how wadm can stop instances that match its expected set of annotations.

## Current behavior
The OTP host, and Rust host, currently support annotations for starting components. When stopping components, supplying annotations can result in different behavior based on the state of the host.

Take a host that's running 2 instances of the `echo` actor, one with no annotations and one with an annotation `foo=bar`. Additionally, we're running the httpserver provider with no annotations, and keyvalue provider with annotation `bar=baz`.

**Scenarios:**
1. Send stop command for echo with no annotations, count 1
    1. Actor with no annotations stops, actor with annotation foo=bar remains
1. Send stop command for echo with annotation foo=bar, count 1
    1. Actor with annotations stops, actor with no annotations remains
1. Send stop command for echo with no annotations, count 2
    1. Actor with no annotations stops, actor with annotation foo=bar remains
1. Send stop command for echo with annotation foo=bar, count 2
    1. Actor with annotations stops, actor with no annotations remains
1. Send stop command for httpserver provider with no annotations
    1. httpserver provider stops
1. Send stop command for httpserver provider with annotation hee=haw
    1. httpserver provider stops
1. Send stop command for keyvalue provider with no annotations
    1. keyvalue provider stops
1. Send stop command for keyvalue provider with annotation bar=baz
    1. keyvalue provider stops

## Desired behavior
Annotations should act as a restrictive filter, where providing no annotations will match with _any_ resource with the provided public key (and link name for providers) and each additional annotation will filter out non-matching resources.

**Scenarios(changes bolded):**
1. Send stop command for echo with no annotations, count 1
    1. Actor with no annotations stops, actor with annotation foo=bar remains
    2. **NOTE:** This scenario is now non-deterministic given the fact that we could equally choose to terminate either resource. I lean towards trying to stop the actor with no annotations in this case.
1. Send stop command for echo with annotation foo=bar, count 1
    1. Actor with annotations stops, actor with no annotations remains
1. Send stop command for echo with no annotations, count 2
    1. **Both echo actors stop**
1. Send stop command for echo with annotation foo=bar, count 2
    1. Actor with annotations stops, actor with no annotations remains
1. Send stop command for httpserver provider with no annotations
    1. httpserver provider stops
1. Send stop command for httpserver provider with annotation hee=haw
    1. **httpserver provider does not stop**
1. Send stop command for keyvalue provider with no annotations
    1. keyvalue provider stops
1. Send stop command for keyvalue provider with annotation bar=baz
    1. keyvalue provider stops

brooksmtownsend:
## Feature or Problem
This PR treats incoming annotations on `stop_actor`, `scale_actor` (to stop), and `stop_provider` as a filter instead of a strict set (or ignored in the case of #606). Please see the description in #607 to see a few examples of this filter in use.

This PR also removes an unnecessary variable `actor_id` from the scale command, and I think we should just completely remove it from the command but that will take place in the control interface.

## Related Issues
Fixes #607 

## Release Information
v0.78.0

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
Modified the wasmbus test to include a few test cases for scale and filtering by annotation.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -3988,8 +3988,7 @@ fn annotations_match_filter(annotations: &Annotations, filter: &Annotations) -> 
     filter.iter().all(|(k, v)| {
         annotations
             .get(k)
-            .map(|instance_value| instance_value == v)
-            .unwrap_or(false)
+            .is_some_and(|instance_value| instance_value == v)
     })
 }
 
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2292,16 +2292,16 @@ impl Host {
             .map(anyhow::Result::<_>::Ok)
             .try_fold(
                 remaining,
-                |remaining, (annotations, mut instances)| async move {
+                |remaining, (instance_annotations, mut instances)| async move {
                     let Some(count) = NonZeroUsize::new(instances.len()) else {
-                        return Ok(remaining)
+                        return Ok(remaining);
                     };
                     let remaining = remaining
                         .checked_sub(count.into())
                         .context("invalid instance length")?;
                     self.uninstantiate_actor(
                         claims,
-                        &annotations,
+                        &instance_annotations,
                         host_id,
                         &mut instances,
                         count,
@@ -2442,7 +2442,6 @@ impl Host {
         host_id: &str,
     ) -> anyhow::Result<Bytes> {
         let ScaleActorCommand {
-            actor_id,
             actor_ref,
             count,
             annotations,
@@ -2450,7 +2449,7 @@ impl Host {
         } = serde_json::from_slice(payload.as_ref())
             .context("failed to deserialize actor scale command")?;
 
-        debug!(actor_id, actor_ref, count, "scale actor");
+        debug!(actor_ref, count, "scale actor");
 
         let actor = self.fetch_actor(&actor_ref).await?;
         let claims = actor.claims().context("claims missing")?;
@@ -2471,7 +2470,7 @@ impl Host {
             )
         };
 
-        let annotations = annotations.unwrap_or_default().into_iter().collect();
+        let annotations: Annotations = annotations.unwrap_or_default().into_iter().collect();
         match (
             self.actors.write().await.entry(actor_id),
             NonZeroUsize::new(count.into()),
@@ -2481,14 +2480,23 @@ impl Host {
                 self.start_actor(entry, actor, actor_ref, count, host_id, annotations)
                     .await?;
             }
-            (hash_map::Entry::Occupied(entry), None) => {
-                self.stop_actor(entry, host_id).await?;
-            }
-            (hash_map::Entry::Occupied(entry), Some(count)) => {
+            (hash_map::Entry::Occupied(entry), _) => {
                 let actor = entry.get();
-                let mut instances = actor.instances.write().await;
+                let mut actor_instances = actor.instances.write().await;
                 let count = usize::from(count);
-                let current = instances.values().map(Vec::len).sum();
+                // Only consider instances that match the requested annotations
+                let matching_instances = actor_instances
+                    .iter()
+                    .filter_map(|(instance_annotations, instances)| {
+                        annotations_match_filter(instance_annotations, &annotations)
+                            .then(|| (instance_annotations.clone(), instances.clone()))
+                    })
+                    .collect::<Vec<_>>();
+                let current = matching_instances
+                    .iter()
+                    .map(|(_annotations, instances)| instances.len())
+                    .sum();
+
                 let claims = actor.pool.claims().context("claims missing")?;
                 if let Some(delta) = count.checked_sub(current).and_then(NonZeroUsize::new) {
                     let mut delta = self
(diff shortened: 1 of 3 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -3944,6 +3961,18 @@ fn injector_to_headers(injector: &TraceContextInjector) -> async_nats::header::H
...
 
+/// Helper function to ensure an individual instance's annotations contain all the
+/// filter annotations, determining if an instance matches a filter.
+///
+/// See #607 for more details.
+fn annotations_match_filter(annotations: &Annotations, filter: &Annotations) -> bool {
+    filter.iter().all(|(k, v)| {
+        annotations
+            .get(k)
+            .is_some_and(|instance_value| instance_value == v)
```
rvolosatovs, line 3972: nit; couldn't we do something like 
```suggestion
        annotations.get(k).as_deref() == Some(v)
```
instead?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3944,6 +3961,18 @@ fn injector_to_headers(injector: &TraceContextInjector) -> async_nats::header::H
         .collect()
 }
 
+/// Helper function to ensure an individual instance's annotations contain all the
+/// filter annotations, determining if an instance matches a filter.
+///
+/// See #607 for more details.
+fn annotations_match_filter(annotations: &Annotations, filter: &Annotations) -> bool {
```
rvolosatovs, line 3968: nit; what if we had a function, which would take the instances and set of annotations to filter by and return an iterator instead?
This function could be as generic as desired, it could even be implemented as an extension on a generic actor instance collection, but all I'm trying to achieve here, is basically usage like this:
`actor_instances.filter_by_annotations(&annotations).for_each(...)`

it does not need to happen in this PR, but would be nice to perhaps add an issue to refactor this?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2481,14 +2480,20 @@ impl Host {
...
-                self.stop_actor(entry, host_id).await?;
-            }
-            (hash_map::Entry::Occupied(entry), Some(count)) => {
+            (hash_map::Entry::Occupied(entry), _) => {
                 let actor = entry.get();
-                let mut instances = actor.instances.write().await;
+                let mut actor_instances = actor.instances.write().await;
                 let count = usize::from(count);
-                let current = instances.values().map(Vec::len).sum();
+                let current = actor_instances
```
rvolosatovs, line 2487: Cannot comment on an unchanged line, but this will change the value of `remaining` actors in the event(s) published, will that break any consumer usages? (this is a breaking change potentially)

crates/host/src/wasmbus/mod.rs
```diff
@@ -2481,14 +2480,20 @@ impl Host {
...
+                    .iter()
+                    .filter_map(|(instance_annotations, instances)| {
+                        if annotations_match_filter(instance_annotations, &annotations) {
+                            Some(instances.len())
+                        } else {
+                            None
+                        }
```
rvolosatovs, line 2494:
```suggestion
                        annotations_match_filter(instance_annotations, &annotations).then(|| instances.len())
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2481,14 +2480,20 @@ impl Host {
...
+                    })
+                    .sum();
```
connorsmith256, line 2496: Suggestion: instead of filter_map + summing the count here, store a `current_instances` that uses the `annotations_match_filter`, and a separate `current_count`. Then in the scale-down case below, we can iterate over this filtered collection (which _likely_ only has a single member) instead of all `actor_instances` again. That also removes the need for the comment and check you added below

crates/host/src/wasmbus/mod.rs
brooksmtownsend: We could, feels like personal preference and I don't mind either way

crates/host/src/wasmbus/mod.rs
```diff
@@ -2529,10 +2542,11 @@ impl Host {
                             break;
                         }
                     }
-                    if remaining == 0 {
-                        drop(instances);
-                        entry.remove();
-                    }
+                }
+
+                if actor_instances.is_empty() {
```
connorsmith256, line 2547: Is this right? Previously we would drop the individual `instances` if a group of annotated-instances became empty. Now it looks like we only drop if _all_ instances are gone for the actor

Though it also looks like there's a bug in the current implementation. I don't think we want to `remove` when _any_ group becomes empty. So it looks like we might need two checks:
- remove groups from `actor_instances` if they become empty as a result of scaling down
- remove the overall actor entry from `actors` if all instances are stopped

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Ah I see, if an instance filter matches multiple instance lists based on annotations then it would result in multiple `actor_stopped` and `actors_stopped` events 🤔 This might require some thought, but based on your above comment I think refactoring to filter the instances by annotations and then uninstantiating all of those would be a better way to handle this.

crates/host/src/wasmbus/mod.rs
```diff
@@ -3106,58 +3120,61 @@ impl Host {
...
         );
 
+        let annotations: Annotations = annotations.unwrap_or_default().into_iter().collect();
         let mut providers = self.providers.write().await;
         let hash_map::Entry::Occupied(mut entry) = providers.entry(provider_ref.clone()) else {
             return Ok(SUCCESS.into());
         };
         let provider = entry.get_mut();
         let instances = &mut provider.instances;
         if let hash_map::Entry::Occupied(entry) = instances.entry(link_name.clone()) {
```
connorsmith256, line 3130: Should we add an `else` to this that emits a stop-failed event because nothing matched? (same for scaling down/stopping actors)

connorsmith256 (changes requested):

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I thought the behavior you described is actually what I implemented, but will take another look back through because that's definitely desired behavior.

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I don't even know if we have a stop_failed event! That would be a decent use for the new event and it would be nice to know when you did something wrong, though I would lean more in the direction of returning a failed `CtlOperationAck`

connorsmith256: Agreed, I think a failed ack is a good addition

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Fairly sure my latest commit now implements this behavior as described

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I do like this idea, I couldn't quite get it to work nicely with my latest push in order to properly update the actors list as we remove across potentially different annotations. I think in general we really need to revisit this Annotations -> List of instances way of bookkeeping, once we can get rid of keeping track of instance counts this likely gets a whole lot easier.

For that reason I'd lean away from another issue, but this feedback should be considered when getting rid of the instances list.

crates/host/src/wasmbus/mod.rs
```diff
@@ -2481,14 +2480,23 @@ impl Host {
...
-            }
-            (hash_map::Entry::Occupied(entry), Some(count)) => {
+            (hash_map::Entry::Occupied(entry), _) => {
                 let actor = entry.get();
-                let mut instances = actor.instances.write().await;
+                let mut actor_instances = actor.instances.write().await;
                 let count = usize::from(count);
-                let current = instances.values().map(Vec::len).sum();
+                // Only consider instances that match the requested annotations
+                let current_instances = actor_instances
```
connorsmith256, line 2488: Naming supernit: something like `matching_instances` might fit better. `current` kinda makes it sound like it's all "running" instances

crates/host/src/wasmbus/mod.rs
```diff
@@ -2503,37 +2511,57 @@ impl Host {
...
                             continue;
                         };
+
                         remaining = remaining
                             .checked_sub(count.into())
                             .context("invalid instance length")?;
                         delta = delta.checked_sub(count.into()).context("invalid delta")?;
+
+                        // Update actor instance list, removing instances that are being scaled down
+                        let mut new_instances = instances.clone();
```
connorsmith256, line 2536: Naming supernit: `remaining_instances`?

connorsmith256 (approved):
A couple suggestions around naming, but the logic LGTM

connorsmith256 (approved):

PR #661, opened 2023-09-19
refactor: remove instance pooling

rvolosatovs:
## Feature or Problem

Remove instance pooling functionality, since Wasmtime instantiation is designed to be fast and cheap. *not pooling* is the part of Wasmtime security model

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


brooksmtownsend (commented):
@rvolosatovs I know this is in draft so disregard if this is planned work.

I believe we talked about the `count` field supplied on the actor start/scale events serving as a sort of ceiling for maximum number of concurrent requests, rather than deciding how many "Instances" to run concurrently. This PR so far looks like it removes the notion of count entirely, simply letting us instantiate on request. Is it a planned piece of this PR that will allow us to instantiate an actor instance for a request as long as we're not above that threshold?

Secondary comment, I think this should be merged as a part of the 0.79.0 release due to the breaking implications this has on actor instances, we'll want to change a couple of things across the org to support no instance lists

rvolosatovs:
> @rvolosatovs I know this is in draft so disregard if this is planned work.
> 
> I believe we talked about the `count` field supplied on the actor start/scale events serving as a sort of ceiling for maximum number of concurrent requests, rather than deciding how many "Instances" to run concurrently. This PR so far looks like it removes the notion of count entirely, simply letting us instantiate on request. Is it a planned piece of this PR that will allow us to instantiate an actor instance for a request as long as we're not above that threshold?
> 
> Secondary comment, I think this should be merged as a part of the 0.79.0 release due to the breaking implications this has on actor instances, we'll want to change a couple of things across the org to support no instance lists

Sorry, did not have time to update the OP originally, but this PR does not actually have any functional changes - the reason for this PR is actually this talk by @fitzgen (thanks for the awesome talk!) https://www.youtube.com/watch?v=FFPoOR_5urw&list=PLbzoR-pLrL6pdO2yYg4jAQAW3zd15CYO0&index=28, where he talks about Wasmtime security model and how instantiation is designed to be cheap, which means that instance pooling as designed is simply not necessary (and does not work well with components anyway)
That's the only thing being changed here

crates/host/src/wasmbus/mod.rs
```diff
@@ -2820,15 +2832,14 @@ impl Host {
         .await
         .context("failed to uninstantiate running actor")?;
 
-        let new_pool = ActorInstancePool::new(new_actor.clone(), Some(count));
```
rvolosatovs, line 2823: This change does appear like it would remove the upper limit on instances, but do note that it never actually behaved that way - I was not aware this was introduced and is not how it was designed to be used. See https://github.com/wasmCloud/wasmCloud/pull/574 for more context

This simply defines the maximum amount of instances pooled, which has no effect on the amount of instances that could be active at any point of time

rvolosatovs:
> Note that to get the fastest instantiation out of Wasmtime, you'll want to configure Wasmtime's pooling instance allocator (this pools the resources necessary for allocating instances, but not Wasm-level instances themselves, so each new instantiation is an isolated instance but is cheap to instantiate):
> 
>     * [docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.allocation_strategy](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.allocation_strategy)
> 
>     * [docs.rs/wasmtime/latest/wasmtime/enum.InstanceAllocationStrategy.html#variant.Pooling](https://docs.rs/wasmtime/latest/wasmtime/enum.InstanceAllocationStrategy.html#variant.Pooling)
> 
>     * [docs.rs/wasmtime/latest/wasmtime/struct.PoolingAllocationConfig.html](https://docs.rs/wasmtime/latest/wasmtime/struct.PoolingAllocationConfig.html)

Thank you so much for this @fitzgen ! :heart:

brooksmtownsend (approved):

PR #663, opened 2023-09-19
fix(host): proxy SYSTEMROOT to providers on Windows

connorsmith256:
## Feature or Problem
Apparently child processes on Windows need a `SYSTEMROOT` env var set, or else they fail to load DLLs with:
```
Error: ConnectError { kind: Io, source: Some(Os { code: 10106, kind: Uncategorized, message: "The requested service provider could not be loaded or initialized." }) }
{"timestamp":"2023-09-19T17:03:48.992363Z","level":"DEBUG","fields":{"message":"`C:\\Users\\Connor\\AppData\\Local\\Temp\\wasmcloudcache\\VCCVLH4XWGI3SGARFNYKYT2A32SUYA2KVAIV2U2Q34DQA7WWJPFRKIKM\\1685738546\\wasmcloud_httpclient_default.exe` exited with `ExitStatus(ExitStatus(1))`"},"target":"wasmcloud_host::wasmbus"}
```

This PR proxies this env var from the host to provider processes on Windows

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [x] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [x] x86_64-windows

### Manual Verification
Built the host and a provider on windows, and confirmed the provider starts and listens on the expected topic:
```

C:\Users\Connor\Downloads\nats-0.0.35-windows-amd64> ./nats.exe req wasmbus.rpc.default.VDKTWC5Z2SWMWZV5S4F6AQDVZFJE7FLVASN35RPPUZ6TGCGUY3TBJP2B.default '{}'
12:12:29 Sending request on "wasmbus.rpc.default.VDKTWC5Z2SWMWZV5S4F6AQDVZFJE7FLVASN35RPPUZ6TGCGUY3TBJP2B.default"
12:12:29 Received with rtt 1.599ms
��msg� �invocation_id��error�Udeser error: deserialization: invalid type: integer `123`, expected struct Invocation�content_length
```

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2900,7 +2900,9 @@ impl Host {
                 host_data = &*String::from_utf8_lossy(&host_data),
                 "spawn provider process"
             );
-            let mut child = process::Command::new(&path)
+
+            let mut child_cmd = process::Command::new(&path);
+            child_cmd
                 .env_clear()
                 // TODO: remove these once all providers are updated to use the new SDK
                 .env(
@@ -2918,7 +2920,18 @@ impl Host {
                         .exporter_otlp_endpoint
                         .clone()
                         .unwrap_or_default(),
-                )
+                );
+
+            if cfg!(windows) {
+                // Proxy SYSTEMROOT to providers. Without this, providers on Windows won't be able to start
+                child_cmd.env(
+                    "SYSTEMROOT",
+                    env::var("SYSTEMROOT")
+                        .context("SYSTEMROOT is not set. Providers cannot be started")?,
+                );
+            }
+
+            let mut child = child_cmd
                 .stdin(Stdio::piped())
                 .kill_on_drop(true)
                 .spawn()


crates/host/src/wasmbus/mod.rs
```diff
@@ -2918,7 +2920,14 @@ impl Host {
                         .exporter_otlp_endpoint
                         .clone()
                         .unwrap_or_default(),
-                )
+                );
+
+            if cfg!(windows) {
+                // Proxy SYSTEMROOT to providers. Without this, providers on Windows won't be able to start
+                child_cmd.env("SYSTEMROOT", env::var("SYSTEMROOT").unwrap_or_default());
```
brooksmtownsend, line 2927: Thoughts on unwrapping to the default of `C:\\Windows` just in case it's not found for whatever reason to _hopefully_ catch that case for the majority of users?

brooksmtownsend: I think bailing works just fine and we can encourage users to set `SYSTEMROOT` if they for some reason do not have it set.

connorsmith256: Updated!

ricochet:
I don't have `SYSTEMROOT` set and I've been doing rust, cpp, and python dev on this machine for a while.

To determine what value should be set, run `wmic os get WindowsDirectory`.
```bash
wmic os get WindowsDirectory
WindowsDirectory
C:\WINDOWS
```

brooksmtownsend:
@ricochet I wasn't aware this isn't a common environment variable, do you think we should revisit the idea of bailing if not set?

ricochet:
If I follow old guides, I don't get set anything even when running ` set systemroot`. I'm running windows 11 so perhaps this a deprecated env?

brooksmtownsend (approved):

PR #701, opened 2023-09-28
fix: rework host shutdown

rvolosatovs:
## Feature or Problem

- Always include a timeout for graceful shutdown (e.g. if NATS connection dies, it will never finish)
- Stop if one of the core wasmbus tasks dies
- Flush NATS queues concurrently on shutdown
- Handle `stopped` method errors

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

build and run, then:
- try starting the host, killing NATS and `^C` - it will shutdown after 10 seconds of trying to gracefully shutdown
- the usual `wash stop host`

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1921,13 +1921,25 @@ impl Host {
         let host = Arc::new(host);
         let queue = spawn({
             let host = Arc::clone(&host);
-            async {
-                Abortable::new(queue, queue_abort_reg)
-                    .for_each(move |msg| {
+            async move {
+                let mut queue = Abortable::new(queue, queue_abort_reg);
+                queue
+                    .by_ref()
+                    .for_each({
                         let host = Arc::clone(&host);
-                        async move { host.handle_message(msg).await }
+                        move |msg| {
+                            let host = Arc::clone(&host);
+                            async move { host.handle_message(msg).await }
+                        }
                     })
                     .await;
+                let deadline = { *host.stop_rx.borrow() };
+                host.stop_tx.send_replace(deadline);
+                if queue.is_aborted() {
+                    info!("control interface queue task gracefully stopped");
+                } else {
+                    error!("control interface queue task unexpectedly stopped");
+                }
             }
         });
         let data_watch: JoinHandle<anyhow::Result<_>> = spawn({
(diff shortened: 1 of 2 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1921,13 +1921,20 @@ impl Host {
...
+                    Abortable::new(queue, queue_abort_reg)
+                        .for_each(move |msg| {
+                            let host = Arc::clone(&host);
+                            async move { host.handle_message(msg).await }
+                        })
+                        .await;
+                }
+                let deadline = { *host.stop_rx.borrow() };
+                host.stop_tx.send_replace(deadline);
+                info!("control interface queue task stopped");
```
thomastaylor312, line 1937: Nit: Will we only ever get here (and in the other tasks) on clean exit or could we also get here if the task aborts early? If we can abort early, should this be a `warn` or `error`?

thomastaylor312 (dismissed):
Just one small nit here

crates/host/src/wasmbus/mod.rs
rvolosatovs: both and IMO it'd be pretty confusing to get a `warn`/`error` on clean exit

rvolosatovs: We will get an error before that from the actual stream though

connorsmith256: To clarify: I'm wondering whether there's a case where the underlying NATS client's pollable returns a `Poll::Ready(None)` without us intentionally unsubscribing

rvolosatovs: > I have the same question, but I think the answer is: this and the other tasks below can be stopped from receiving a signal (ctrl-C, or a `.stop` command on the control interface), or abnormally if the underlying stream finishes (which shouldn't ever happen). @rvolosatovs is there a way to handle when the stream unexpectedly closes?

did not see this before answering.
we can actually just check https://docs.rs/futures/latest/futures/stream/struct.Abortable.html#method.is_aborted
It won't always be true, there's a race condition, but it should be good enough - let me do that.

connorsmith256: Oo I like that idea 👍

connorsmith256 (approved):

thomastaylor312 (approved):
Looking great!

PR #709, opened 2023-10-03
chore: reduce verbosity of instrumented functions

connorsmith256:
## Feature or Problem
This updates many instrumented functions to be at the debug or trace level. As a rule of thumb, I moved runtime internals to trace, and host functions to debug

### DIFF
--- a/crates/host/src/policy.rs
+++ b/crates/host/src/policy.rs
@@ -260,7 +260,7 @@ impl Manager {
     }
 
     /// Constructs a
-    #[instrument(skip(self))]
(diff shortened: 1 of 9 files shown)


rvolosatovs (approved):
I'd just move all instrumentation to `trace`

brooksmtownsend (approved):

connorsmith256:
@rvolosatovs we still want most spans enabled when the wasmcloud logging level is set to info/debug. But the ones I changed to trace are deep in the runtime of the host and result in many more spans than are needed in typical use

Issue #696, opened 2023-09-27
[RFC] Actor Autoscaling & Scale to Zero


brooksmtownsend:
# Introduction

This RFC proposes changes to our verbiage around starting, stopping, scaling application components (actor or provider) in wasmCloud. With recent support for the latest wasmtime and the ability to instantiate on-request, we no longer need the notion of a prevailing “count” of instances that are running. This allows us to reimagine our approach to Wasm apps, improving the developer experience, even if it means breaking a few things.

## Motivation

My goal is for developers using wasmCloud to never think about scaling. They should have an application that scales up and down, on demand, based on load. WebAssembly components are cheap to instantiate and do so quickly, effectively leaving us with no reason to run “hot” instances based on a predetermined `count` . I also want to use as much terminology from current paradigms as possible so new wasmCloud developers easily understand what they are doing when they start running application components. 

## Detailed Design

The **********Scale********** operation on the control interface and CLI becomes the main mechanism for scheduling a resource (actor or provider) with an optional `--max` parameter. The `--count` field is deprecated entirely.

Inside of the wasmCloud host, we no longer track lists of instances of individual actors. An “Actor” now refers to a compiled WebAssembly component that's **instantiated** upon receiving a request on its RPC topic. After that instance finishes execution, it is discarded and its resources disposed of. wasmCloud components now completely scale down when not actively executing and can automatically scale to handle load and demand.

Though we technically do not need to provide an upper bound to the number of concurrently executing actors, it’s a good idea to do so in order to avoid an infinitely growing scope of execution. The primary reason that I see it as important is to prevent a thundering herd of requests affecting the ability for other actors to handle requests. It also has the benefit of bounding execution on devices that are constrained in order to prevent unexpected CPU or memory pressure. The resource allocation for a WebAssembly component is measurable and calculable, so when running on a severely constrained device we could reasonably tell each actor component to execute at a maximum concurrency of `100` to throttle unexpected bursts. **********************************************************************************By default, the maximum concurrency will be unbounded, represented as a max scale of 0.**********************************************************************************

This requires changes in a few crates and projects, laid out below.

**wasmCloud control interface client:** 

- Start will be deprecated as we release `0.30.0` and will be removed in a future release, I’d be fine with this being `0.31.0`.
- Stop remains as-is, though the `count` field will be removed in `0.30.0`. Stopping an actor stops the entire entity (respecting annotations) rather than stopping some instances.
[...]

PR #710, opened 2023-10-03
[FEAT] Actor Autoscaling and scale to zero

### LINKED ISSUE #746: [BUG] Scaling an actor has ambiguous filtering logic
**Describe the bug**
In the implementation of #656, we added support for annotations filters which lets wasmCloud users stop and scale actors based on supplied annotations. This lets users not need to supply an exact set of annotations in order to stop running actors to avoid confusing messages. Unfortunately, in the implementation of `scale`, there's an ambiguous scenario that remains and should be fixed.

**To Reproduce**
1. Start NATS and wasmCloud, take note of the host ID
1. Scale actor to 5 instances: `wash ctl scale actor <HOST_ID> MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5 wasmcloud.azurecr.io/echo:0.3.4 --count 5 -a foo=bar`
Now, the following command has two possible meanings: `wash ctl scale actor <HOST_ID> MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5 wasmcloud.azurecr.io/echo:0.3.4 --count 3`
1. The host should scale down to 3 instances of the actor with the annotation `foo=bar` because unsupplied annotations technically filter matches the `foo=bar` annotation (this is what happens today)
2. The host should scale up another 3 instances of the actor with an empty annotations map as unsupplied annotations is an empty map and does not match the `foo=bar` annotation

**Expected behavior**
Resolving the ambiguity goes slightly against our previous implementation but results in better defined behavior. On the scale actor command, annotations should not be a filter and instead they should be an exact match on the individual actor instance. This means that in the above reproduction steps the second scenario is the desired behavior. The stop command will continue to use a filter as a convenience.

**Additional context**
This should be implemented as a fix to 0.79.0 after #710 merges.

brooksmtownsend:
## Feature or Problem
This PR features a few prominent changes to the handling of control interface requests, namely `scale`, `start`, and `stop` when it comes to actor components.
**External facing things**
1. Scale and Start now use `max_concurrent` instead of `count` to describe the handling of multiple instances of an individual actor. Per RFC #696, this accurately describes the way we instantiate actors, instantiating on request to handle an invocation and then exiting rather than running a certain number of things.
2. The `ActorInstance` is now the entity that stores an image reference rather than an `Actor`. This functionally doesn't change anything at this time but enables us in the future to consider running different image references of the same actor. We still report the image reference on the host inventory as before and ensure you cannot run duplicate instances. 
3. Requesting to stop an actor with provided annotations will stop that actor entirely and ignore the `count` field.
4. Scaling an actor to 0 will no longer remove that actor, instead it will ensure that the actor has an unbounded level of concurrency. **This is a breaking change and clients like wadm will need to use `stop_actor` instead of scale to 0**
**Internal changes**.
1. Actor instances internal `calls` abortable tasks now streams messages from the RPC topic and handles them concurrently up to the `max` specified when the actor is started/scaled.
2. Any commands that come in on the `la` aka Start actor topic will be immediately proxied into scale commands by adding the requested number to the current number of max instances. This preserves start functionality for now in a nice backwards compatible way.

## Related Issues
Depends on: https://github.com/wasmCloud/control-interface-client/pull/58
RFC: #696 
Fixes #746 

## Release Information
v0.79.0

## Consumer Impact
This PR builds off of v0.30.0 of the wasmCloud control interface client and it makes every choice possible to be backwards compatible with <v0.30.0. Once this is released there should be no required changes on the client side, though some code paths will result in deprecation warnings.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,13 +70,8 @@ wascap = { workspace = true }
 wasmcloud-control-interface = { workspace = true }
 
 [workspace]
-members = [
-    "crates/*",
-    "examples/rust/*",
-]
-exclude = [
-    "crates/providers"
-]
+members = ["crates/*", "examples/rust/*"]
+exclude = ["crates/providers"]
 
 [workspace.dependencies]
 anyhow = { version = "1", default-features = false }
@@ -137,7 +132,7 @@ wasmcloud-actor = { version = "*", path = "./crates/actor" }
 wasmcloud-actor-macros = { version = "*", path = "./crates/actor/macros" }
 wasmcloud-compat = { version = "*", path = "./crates/compat" }
 wasmcloud-component-adapters = { version = "0.2.3" }
-wasmcloud-control-interface = { version = "0.29", default-features = false }
+wasmcloud-control-interface = { version = "0.30", default-features = false }
 wasmcloud-core = { version = "*", path = "./crates/core" }
 wasmcloud-host = { version = "*", path = "./crates/host" }
 wasmcloud-provider-sdk = { version = "*", path = "./crates/provider-sdk", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,7 +137,7 @@ wasmcloud-actor = { version = "*", path = "./crates/actor" }
 wasmcloud-actor-macros = { version = "*", path = "./crates/actor/macros" }
 wasmcloud-compat = { version = "*", path = "./crates/compat" }
 wasmcloud-component-adapters = { version = "0.2.3" }
-wasmcloud-control-interface = { version = "0.28", default-features = false }
+wasmcloud-control-interface = { version = "0.29", default-features = false }
 wasmcloud-core = { version = "*", path = "./crates/core" }
 wasmcloud-host = { version = "*", path = "./crates/host" }
 wasmcloud-provider-sdk = { version = "*", path = "./crates/provider-sdk", default-features = false }
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2654,7 +2654,7 @@ impl Host {
                             .instantiate_actor(
                                 claims,
                                 &annotations,
-                                host_id.clone(),
+                                host_id,
                                 &actor_ref,
                                 max,
                                 actor.actor.clone(),
@@ -2674,7 +2674,7 @@ impl Host {
                         .instantiate_actor(
                             claims,
                             &annotations,
-                            host_id.clone(),
+                            host_id,
                             &actor_ref,
                             max,
                             actor.actor.clone(),
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -34,7 +34,7 @@ use base64::Engine;
 use bytes::{BufMut, Bytes, BytesMut};
 use cloudevents::{EventBuilder, EventBuilderV10};
 use futures::stream::{AbortHandle, Abortable};
-use futures::{join, stream, try_join, FutureExt, Stream, StreamExt, TryFutureExt, TryStreamExt};
+use futures::{join, stream, try_join, FutureExt, Stream, StreamExt, TryFutureExt};
 use nkeys::{KeyPair, KeyPairType};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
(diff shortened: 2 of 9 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1394,6 +1399,26 @@ impl Deref for Actor {
...
 }
 
+// NOTE: this is specifically a function instead of an [Actor] method
+// to avoid deadlocks on the instances field.
+/// Returns the first instance that matches the provided annotations
+fn matching_instance(
+    instances: &HashMap<BTreeMap<String, String>, Arc<ActorInstance>>,
+    annotations: &Annotations,
+) -> Option<Arc<ActorInstance>> {
+    // for loop here so we can short circuit
```
connorsmith256, line 1409: Nit: `find` should short-circuit too, right?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3588,6 +3518,7 @@ impl Host {
             .trim_start_matches('.')
             .split('.')
             .skip(1);
+        info!("handling ctl request on {}", subject);
```
connorsmith256, line 3521: This should probably be a `trace`. Maaaaybe debug, but it's going to generate a lot of log lines

crates/host/src/wasmbus/mod.rs
```diff
@@ -3644,6 +3575,7 @@ impl Host {
         if let Err(e) = &res {
             warn!("failed to handle `{subject}` request: {e:?}");
         }
+        info!("handled ctl request on {}", subject);
```
connorsmith256, line 3578: Same here: `trace`?

connorsmith256 (approved):
LGTM

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Oh yes this was leftover from debugging

crates/host/src/wasmbus/mod.rs
brooksmtownsend: fine I'll make it functional

Cargo.toml
```diff
@@ -137,8 +137,8 @@ wasmcloud-actor = { version = "*", path = "./crates/actor" }
 wasmcloud-actor-macros = { version = "*", path = "./crates/actor/macros" }
 wasmcloud-compat = { version = "*", path = "./crates/compat" }
 wasmcloud-component-adapters = { version = "0.2.3" }
-wasmcloud-control-interface = { version = "0.28", default-features = false }
 wasmcloud-core = { version = "*", path = "./crates/core" }
+wasmcloud-control-interface = { version = "0.29", default-features = false }
```
rvolosatovs, line 141: Please avoid breaking the lexicographic ordering here

tests/wasmbus.rs
```diff
@@ -47,13 +47,14 @@ async fn free_port() -> anyhow::Result<u16> {
...
 async fn assert_start_actor(
-    ctl_client: &wasmcloud_control_interface::Client,
+    ctl_client: &wasmcloud_control_interface::Client<DirectKvStore>,
     _nats_client: &async_nats::Client, // TODO: This should be exposed by `wasmcloud_control_interface::Client`
     _lattice_prefix: &str,
     host_key: &KeyPair,
     url: impl AsRef<str>,
     count: u16,
 ) -> anyhow::Result<()> {
+    #[allow(deprecated)]
```
rvolosatovs, line 57: I feel like there should be a comment here about what is deprecated and when we'll remove it

crates/host/src/wasmbus/mod.rs
```diff
@@ -49,9 +49,10 @@ use tracing::{debug, error, info, instrument, trace, warn};
 use ulid::Ulid;
 use uuid::Uuid;
 use wascap::{jwt, prelude::ClaimsBuilder};
+#[allow(deprecated)]
```
rvolosatovs, line 52: Please add a comment about what is deprecated, when we'll remove it and ideally add an issue ref as well

crates/host/src/wasmbus/mod.rs
rvolosatovs, line 1409: is this comment still relevant?

brooksmtownsend: Nope 😮‍💨

crates/host/src/wasmbus/mod.rs
```diff
@@ -2336,38 +2366,25 @@ impl Host {
...
+            instances.remove(&to_stop.annotations);
+
+            self.uninstantiate_actor(claims, host_id, to_stop)
+                .await
+                .context("failed to uninstantiate actor")?;
+            Ok(())
+        } else {
+            Err(anyhow::anyhow!(
+                "no actors with matching annotations found to stop"
+            ))
```
rvolosatovs, line 2386:
```suggestion
            bail!("no actors with matching annotations found to stop")
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2799,54 +2757,21 @@ impl Host {
...
-            }
-            _ => {
-                // NOTE: This mimics OTP behavior
-                // TODO: What does OTP do?
+        match self.actors.write().await.entry(actor_ref) {
+            hash_map::Entry::Occupied(entry) => {
+                self.stop_actor(entry, &annotations, host_id).await?;
+                Ok(ACCEPTED.into())
             }
+            hash_map::Entry::Vacant(_) => Err(anyhow::anyhow!("actor is not running on this host")),
```
rvolosatovs, line 2773:
```suggestion
            hash_map::Entry::Vacant(_) => bail!("actor is not running on this host"),
```

crates/host/src/wasmbus/mod.rs
rvolosatovs: If that's just for the `start_actor`, then can we remove the import here and refer to it using absolute path to avoid having to do this?

connorsmith256 (dismissed):

connorsmith256 (approved):

PR #716, opened 2023-10-04
fix(wasmbus): download actor to scale in task

brooksmtownsend:
## Feature or Problem
This PR moves the heavy-IO work of downloading the wasmCloud actor before scaling into a tokio task (and subsequently the rest of the work) in order to quick-ack the control interface request and be able to handle additional commands.

## Related Issues
Part of this work is done in #710 but it's more essential to release and can go with 0.79.0 directly.

## Release Information
v0.79.0-rc2

## Consumer Impact
Consumers of the control interface client will no longer receive an error unless the actual scale payload is invalid. Validation of running actors and downloading of the image reference is done within the task, so the only thing we can do there is send out `actors_start_failed` commands.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2492,7 +2492,7 @@ impl Host {
 
     #[instrument(skip(self, payload))]
     async fn handle_scale_actor(
-        &self,
+        self: Arc<Self>,
         payload: impl AsRef<[u8]>,
         host_id: &str,
     ) -> anyhow::Result<Bytes> {
@@ -2506,7 +2506,30 @@ impl Host {
 
         debug!(actor_ref, count, "scale actor");
 
-        let actor = self.fetch_actor(&actor_ref).await?;
+        let host_id = host_id.to_string();
+        let annotations = annotations.unwrap_or_default().into_iter().collect();
+        spawn(async move {
+            if let Err(e) = self
+                .handle_scale_actor_task(&actor_ref, &host_id, count, annotations)
+                .await
+            {
+                error!("{e:?}");
+            }
+        });
+        Ok(SUCCESS.into())
+    }
+
+    #[instrument(skip(self))]
+    async fn handle_scale_actor_task(
+        &self,
+        actor_ref: &str,
+        host_id: &str,
+        count: u16,
+        annotations: Annotations,
+    ) -> anyhow::Result<()> {
+        debug!(actor_ref, count, "scale actor task");
+
+        let actor = self.fetch_actor(actor_ref).await?;
         let claims = actor.claims().context("claims missing")?;
         let actor_id = claims.subject.clone();
         let resp = self
@@ -2525,7 +2548,7 @@ impl Host {
             )
         };
 
-        let annotations: Annotations = annotations.unwrap_or_default().into_iter().collect();
+        let actor_ref = actor_ref.to_string();
         match (
             self.actors.write().await.entry(actor_id),
             NonZeroUsize::new(count.into()),
@@ -2621,7 +2644,7 @@ impl Host {
                         .await
                         .context("failed to uninstantiate actor")?;
                     } else {
-                        return Ok(r#"{"accepted":false,"error":"no actors with matching annotations found to stop"}"#.into());
+                        bail!("no actors with matching annotations found to stop");
                     }
                 }
 
@@ -2631,7 +2654,7 @@ impl Host {
                 }
             }
         }
-        Ok(SUCCESS.into())
+        Ok(())
     }
 
     #[instrument(skip(self))]
@@ -3583,9 +3606,10 @@ impl Host {
             (Some("cmd"), Some(host_id), Some("sa"), None) => {
                 self.handle_stop_actor(payload, host_id).await.map(Some)
             }
-            (Some("cmd"), Some(host_id), Some("scale"), None) => {
-                self.handle_scale_actor(payload, host_id).await.map(Some)
-            }
+            (Some("cmd"), Some(host_id), Some("scale"), None) => Arc::clone(&self)
+                .handle_scale_actor(payload, host_id)
+                .await
+                .map(Some),
             (Some("cmd"), Some(host_id), Some("sp"), None) => {
                 self.handle_stop_provider(payload, host_id).await.map(Some)
             }
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -68,7 +68,7 @@ use wasmcloud_runtime::capability::{
 use wasmcloud_runtime::Runtime;
 use wasmcloud_tracing::context::TraceContextInjector;
 
-const SUCCESS: &str = r#"{"accepted":true,"error":""}"#;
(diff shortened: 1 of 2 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -2506,131 +2506,144 @@ impl Host {
...
-            .await?;
-        if !resp.permitted {
-            bail!(
-                "Policy denied request to start actor `{}`: `{:?}`",
-                resp.request_id,
-                resp.message
-            )
-        };
+        let host_id = host_id.to_string();
+        spawn(async move {
```
rvolosatovs, line 2510: Can we introduce a `handle_scale_actor_task` like for `launch_actor`?

brooksmtownsend: Sure 👍

brooksmtownsend: Done!

crates/host/src/wasmbus/mod.rs
```diff
@@ -1933,7 +1933,7 @@ impl Host {
                 let mut queue = Abortable::new(queue, queue_abort_reg);
                 queue
                     .by_ref()
-                    .for_each({
+                    .for_each_concurrent(0, {
```
thomastaylor312, line 1936: Nit: having this parameter be 0 is a little confusing if you go look at the docs because it seems like you're telling it not to have _any_ concurrency. Consider setting this to `None` instead

crates/host/src/wasmbus/mod.rs
```diff
@@ -2506,7 +2506,30 @@ impl Host {
...
+        let annotations = annotations.unwrap_or_default().into_iter().collect();
+        spawn(async move {
+            if let Err(e) = self
+                .handle_scale_actor_task(&actor_ref, &host_id, count, annotations)
+                .await
+            {
+                error!("{e:?}");
+            }
+        });
+        Ok(SUCCESS.into())
```
thomastaylor312, line 2519: Nit: We might want to rename this const to `ACCEPTED` as when I first read it I was like "wait no, we don't know if it is successful yet"

tests/wasmbus.rs
```diff
@@ -67,20 +67,35 @@ async fn assert_start_actor(
 
 async fn assert_scale_actor(
     ctl_client: &wasmcloud_control_interface::Client,
-    _nats_client: &async_nats::Client, // TODO: This should be exposed by `wasmcloud_control_interface::Client`
-    _lattice_prefix: &str,
+    nats_client: &async_nats::Client, // TODO: This should be exposed by `wasmcloud_control_interface::Client`
```
thomastaylor312, line 70: Should we open an issue against the control interface and link here?

thomastaylor312 (approved):
All nits here. None of them are blocking

thomastaylor312 (dismissed):

tests/wasmbus.rs
```diff
@@ -67,20 +67,35 @@ async fn assert_start_actor(
...
     ensure!(error == "");
     ensure!(accepted);
 
+    // Naive wait for at least a stopped / started event before exiting this function. This prevents
+    // assuming we're done with scaling too early since scale is an early-ack ctl request.
+    while let Some(message) = sub.next().await {
+        if let Ok(evt) = serde_json::from_slice::<serde_json::Value>(&message.payload) {
+            if let Some(serde_json::Value::String(event_type)) = evt.get("type") {
+                if event_type.contains("actors_stopped") || event_type.contains("actors_started") {
+                    break;
```
connorsmith256, line 93: This looks potentially flaky, since we're not filtering on which actor was started/stopped. Does the payload include the ref/ID we can check for?

connorsmith256 (changes requested):

tests/wasmbus.rs
brooksmtownsend: The payload includes the ref on the started event but not the stopped, and since we're building these actors locally I can't assume a specific ID. Potentially flaky though at this time we're only issuing one of these assertions at a time, not try_join-ing all of them like the start actor assertion

thomastaylor312: Was typing the same thing @brooksmtownsend just said when it came through. That is why I wasn't worried

crates/host/src/wasmbus/mod.rs
```diff
@@ -1933,7 +1933,7 @@ impl Host {
                 let mut queue = Abortable::new(queue, queue_abort_reg);
                 queue
                     .by_ref()
-                    .for_each({
+                    .for_each_concurrent(None, {
```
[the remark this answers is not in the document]
brooksmtownsend: It honestly at this time does the same thing since starting an actor goes into a background task.

To hopefully address your point without nitpicking the example, e.g. the host receives a `stop_actor` message and get inventory at the same time, yes you're right that we now have to consider that we don't have a FIFO queue style of handling CTL requests now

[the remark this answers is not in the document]
brooksmtownsend: It honestly at this time does the same thing since starting an actor goes into a background task.

To hopefully address your point without nitpicking the example, e.g. the host receives a `stop_actor` message and get inventory at the same time, yes you're right that we now have to consider that we don't have a FIFO queue style of handling CTL requests now

crates/host/src/wasmbus/mod.rs
```diff
@@ -2506,7 +2506,30 @@ impl Host {
...
 
-        let actor = self.fetch_actor(&actor_ref).await?;
+        let host_id = host_id.to_string();
+        let annotations = annotations.unwrap_or_default().into_iter().collect();
+        spawn(async move {
+            if let Err(e) = self
+                .handle_scale_actor_task(&actor_ref, &host_id, count, annotations)
+                .await
+            {
+                error!("{e:?}");
```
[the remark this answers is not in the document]
thomastaylor312: Note to the above, I think most of our error fields are called `error` so make sure it is 

```
error!(%actor_ref, error = ?e, "failed to scale actor");
```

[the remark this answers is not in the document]
brooksmtownsend: Great callout here

tests/wasmbus.rs
brooksmtownsend: I'd like to include that information, though, no reason not to

crates/host/src/wasmbus/mod.rs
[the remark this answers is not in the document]
rvolosatovs: I really think we have to revert this change in the future, but I'm fine with it for now as a hotfix

rvolosatovs (dismissed):

connorsmith256 (approved):

connorsmith256 (approved):

rvolosatovs (approved):

PR #728, opened 2023-10-06
chore: update CODEOWNERS

connorsmith256:
## Feature or Problem
This updates the CODEOWNERS file (and removes redundant files)

Note the `.github/CODEOWNERS` path is the first place github checks: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location

To be a valid identifier in CODEOWNERS, the owner must have write access to the repository. Thus this change grants write access to this repo to more individuals than before. Since not every individual who maintains part of the wasmCloud ecosystem should be able to merge all pull requests, in I've enabled a new setting for the branch protection on this repo: "Require review from Code Owners." Since all paths are owned (by the org-maintainers, by default), in practice this should only allow owners to approve code they own


.github/CODEOWNERS
```diff
@@ -1,2 +1,38 @@
...
+# shared crates
+crates/core @wasmCloud/org-maintainers
+crates/opentelemetry-nats @thomastaylor312 @connorsmith256
+crates/provider-sdk @thomastaylor312 @connorsmith256
+crates/provider-wit-bindgen @thomastaylor312 @vados-cosmonic
+crates/tracing @thomastaylor312 @connorsmith256
+
+# projects
+crates/providers @wasmCloud/capability-provider-maintainers
+examples @wasmCloud/examples-maintainers
```
rvolosatovs, line 37: I feel like host maintainers belong here as well, these examples are (at least for now) specific to `wasmCloud` host and will even be run by `cargo` as part of project tests

.github/CODEOWNERS
```diff
@@ -1,2 +1,38 @@
...
+vendor/capability-providers @wasmCloud/capability-provider-maintainers
```
rvolosatovs, line 38: I don't think we need this one, the whole `vendor` directory should be removed and this may complicate getting that to happen

rvolosatovs (approved):
This looks good to me overall, but I'm a bit concerned that if we enforce code owner reviews for merges, we will immensely slow down the progress.
For example, I don't want non-breaking dependency update in one of the "shared" crates to require one of 2 owners to approve the PR for merge.
Let's use code owners to ping the right people for review, but not require code owner reviews for merges (at least, not yet)

.github/CODEOWNERS
```diff
@@ -1,2 +1,38 @@
...
+* @wasmCloud/org-maintainers
+
+# .github CI
+.github @wasmCloud/ci-maintainers
+
+# nix
+.github/actions/build-nix @rvolosatovs
+.github/actions/install-nix @rvolosatovs
+.github/workflows/nix.yml @rvolosatovs
+flake.* @rvolosatovs
```
rvolosatovs, line 11: Doesn't this mean that any change to flakes would require my review for merge?
Doesn't this mean that if we enforce code owner reviews and I'm on PTO, we won't ever be able to merge anything related to flakes? (e.g. a component adapter update)

rvolosatovs:
> @rvolosatovs I agree with your concern, but I don't _think_ that's what's going to happen in practice. I would like to merge this and open a PR to test, to confirm the org maintainers will maintain the ability to approve and merge any PR (which is the status quo). In other words, my goal in setting up CODEOWNERS is to be strictly additive in who can help us make progress 👍 If this doesn't work as configured, I have two ideas for tweaks we can layer on top:
> 
>     * explicitly add the org-maintainers to every line (ew...)
> 
>     * create a separate ruleset protecting main which only applies to org-maintainers and does not require codeowner approval
> 
> 
> If neither of those work, we can revert this and try something else

Org maintainers would indeed need to be added to each line in order for them to keep being code owners, that's, unfortunately, how CODEOWNERS files work

> If you want to match two or more code owners with the same pattern, all the code owners must be on the same line. If the code owners are not on the same line, the pattern matches only the last mentioned code owner.

connorsmith256:
In that case I'll set up a separate rule for org-maintainers

.github/CODEOWNERS
thomastaylor312: I think the main issue is most of the rest of us aren't comfortable enough with how nix works to be confident. I am pretty sure I can check the flakes, but I also don't know that I'd spot issues

thomastaylor312 (approved):

connorsmith256:
I set up rulesets which require approval from codeowners but which can be bypassed by the `org-maintainers` group. Let's see how that works with this 👍

PR #730, opened 2023-10-06
chore: add ci-maintainers to Nix-related paths

connorsmith256:
This way if Roman goes on vacation others can still review and enable auto-merge for changes against these paths :)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,11 +5,11 @@
 .github @wasmCloud/ci-maintainers
 
 # nix
-.github/actions/build-nix @rvolosatovs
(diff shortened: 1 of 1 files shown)


rvolosatovs (approved):

rvolosatovs:
Can I please be made `ci-maintainer`?

thomastaylor312 (changes requested):
Based on what I am seeing here, wouldn't it be easier to just add @rvolosatovs to the ci-maintainers group? That isn't an actual project maintainers group, so I don't think there would be any problem with adding

connorsmith256:
I'm good with that!

brooksmtownsend (approved):

Issue #680, opened 2023-09-25
implement `wasi:http/outgoing-handler`


rvolosatovs:
Mostly done already locally, but we probably want to wait for Wasmtime to complete parts of this as well

PR #745, opened 2023-10-12
refactor: convert httpclient provider to bindgen

vados-cosmonic:
## Feature or Problem

This PR converts the `httpclient` provider to use `provider-wit-bindgen`, utilizing a couple features and bugfixes from other PRs:

https://github.com/wasmCloud/wasmCloud/pull/963
https://github.com/wasmCloud/wasmCloud/pull/983

Those PRs *should* land before this one, but it's not absolutely required.

## Related Issues

#743 
#680 

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration

Pending implementation of `wasi:outgoing-handler` in host

### Manual Verification

Tested locally


rvolosatovs (changes requested):
requesting changes to prevent (accidental) merge until we have https://github.com/wasmCloud/wasmCloud/issues/680 done

Regarding the backwards-incompatibility, I expected we would serialize lists of tuples as maps (that could be a parameter to the macro invocation or we can just hardcode it for now)

rvolosatovs:
> As of 2023/10/13, this code doesn't work due to conversions between `map` (which doesn't exist in WIT) as expressed by the Smithy contract and the `list<tuple<string, list<u8>>>` that WIT has to take. The error looks like this:
> 
> ```
> {"error":"actor: sending req: Host send error failed to call `wasmcloud:bus/host.call`: provider call failed: Error when handling invocation: Error when deserializing invocation: Syntax(\"invalid type: map, expected a sequence\")"}
> ```
> 
> This should be resolved by #680, as the host will likely have to munge invocations and responses to translate maps into/out of what WIT can actually understand.

#680 is not expected to be anyhow WIT-aware, message format should be fully backwards-compatible, meaning that `wasi:http/outgoing-handler` will be sending msgpack maps, just like before. Bindgen, however, should be able to handle that and decode a msgpack map into a list of tuples and other way round

crates/providers/http-client/Cargo.toml
```diff
@@ -1,16 +1,33 @@
...
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
+[badges.maintenance]
+status = "actively-developed"
+
 [dependencies]
 async-trait = { workspace = true }
+base64 = { workspace = true }
```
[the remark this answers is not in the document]
vados-cosmonic: Thanks for the catch, removed these unused crates

vados-cosmonic:
> https://github.com/wasmCloud/wasmCloud/issues/680 is not expected to be anyhow WIT-aware

Well glad I brought this up early -- could you expand on how #680 is not WIT-aware? Maybe the scope of work isn't well defined, there's not much in the issue, but `wasi:http/outgoing-handler` is an intrinsically WIT-based concept.

I thought that the work represented by that issue was implementing the WIT interfaces that are specified by `wasi:http/outgoing-handler` so they can be used by wasmCloud actors?

> , message format should be fully backwards-compatible, meaning that wasi:http/outgoing-handler will be sending msgpack maps, just like before. Bindgen, however, should be able to handle that and decode a msgpack map into a list of tuples and other way round

While it is *possible* for bindgen to generate an object that converts to a msgpack map, it's not clear that it *should* be there, rather than in the host which is doing the `wasi:http/outgoing-handler` to `HttpClient.<operation>` conversions. 

The host intercepts every call, for every operation and does some transformations already.

The reason we *know* that the value should be a map is because of the contract. These kinds of transformations are [already in the code](https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L486), so could you explain why this time should be different?

Even if it was something that should be totally handled in WIT at the bindgen level, it's not safe to convert every `list<tuple<string, list<u8>>>` into a `HashMap` (or whatever type we choose to serialize to a msgpack map), as it's possible that the intended value *is* a list of tuples. 

WIT doesn't currently have a way of telling you "this thing should have been a HashMap for Smithy", that's a requirement that we're injecting by trying to make WIT speak across Smithy interfaces. 

For example, if WIT was talking to WIT, both sides would expect the list of tuples, not a map.

vados-cosmonic:
> meaning that wasi:http/outgoing-handler will be sending msgpack maps, just like before. 

This seems to be a bit confusingly worded -- `wasi:http/outgoing-handler` *cannot* send out msgpack maps, because WIT doesn't have a map type. Is what you mean here is that #680 will make sure that the *host* intercepts `wasi:http/outgoing-handler` calls and converts them into objects that serialize as msgpack maps?

> Bindgen, however, should be able to handle that and decode a msgpack map into a list of tuples and other way round

This is going to require some deeper changes to bindgen -- it's not clear that is safe to turn every `list<tuple<string, list<u8>>>` or any such type into something that serializes to a type that serializes to a msgpack map, and this would break normal WIT <-> WIT semantics. 

This will likely have to be some kind of configuration option.

rvolosatovs:
> > meaning that wasi:http/outgoing-handler will be sending msgpack maps, just like before.
> 
> This seems to be a bit confusingly worded -- `wasi:http/outgoing-handler` _cannot_ send out msgpack maps, because WIT doesn't have a map type. Is what you mean here is that #680 will make sure that the _host_ intercepts `wasi:http/outgoing-handler` calls and converts them into objects that serialize as msgpack maps?

yes

> > Bindgen, however, should be able to handle that and decode a msgpack map into a list of tuples and other way round
> 
> This is going to require some deeper changes to bindgen -- it's not clear that is safe to turn every `list<tuple<string, list<u8>>>` or any such type into something that serializes to a type that serializes to a msgpack map, and this would break normal WIT <-> WIT semantics.
> 
> This will likely have to be some kind of configurable option.

Yes, ideally it's a parameter to the macro, as outlined in my original comment https://github.com/wasmCloud/wasmCloud/pull/745#pullrequestreview-1674510237, but for now we can just hard-code it, because we are the only users of this crate

crates/providers/Cargo.toml
```diff
@@ -32,6 +32,7 @@ futures = { version = "0.3", default-features = false }
 http = { version = "0.2", default-features = false }
 opentelemetry = { version = "0.20", default-features = false }
 opentelemetry-nats = { path = "../opentelemetry-nats" }
+once_cell = { version = "1.18", default-features = false }
```
rvolosatovs, line 35: still looks unused to me

vados-cosmonic: Removed!

vados-cosmonic:
Thanks for the clarifications -- I'm not too keen on hard-coding it, so I'll take a little time to think about a reasonable way to do it flexibly.

rvolosatovs:
@vados-cosmonic any idea why the provider build artifacts might have changed in this PR?
Does e.g.
```
cargo clippy --release -j 2 --all-targets --workspace -- --deny warnings
```
work locally for you with this change set?

vados-cosmonic:
Not sure, but it looks like the `test-providers` build in `build.rs` is failing to build after the rebase -- will look into it and see what changed.

rvolosatovs:
looks like this is blocked on #869 , but I have no permissions to approve/merge, so we'll have to wait until someone wakes up in US

rvolosatovs:
feel free to just commit the regenerated `Cargo.lock`

brooksmtownsend:
@rvolosatovs / @vados-cosmonic merged that PR!

brooksmtownsend (commented):
@rvolosatovs This looks good to me at a high level! Let me know if you need an additional approval here to merge but leaving it to you since you've been following

vados-cosmonic:
Hey @brooksmtownsend this isn't ready to merge yet -- I still haven't been able to get the tests working even after rebasing -- custom build isn't passing still even with the `Cargo.lock` changes.

Regardless of that, given that the host implementation changed there will likely be changes required here.

Going to put this back in Draft so people aren't reviewing it.

crates/providers/http-client/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[instrument(level = "debug", skip(self, _ctx, req), fields(actor_id = ?_ctx.actor, method = %req.method, url = %req.url))]
+    async fn request(
+        &self,
+        _ctx: Context,
+        req: HttpRequest,
+    ) -> ProviderInvocationResult<HttpResponse> {
+        let headers: HeaderMap = build_http_header_map(&req.headers)?;
+
+        let method = reqwest::Method::from_str(&req.method).map_err(|e| {
+            ProviderInvocationError::Provider(format!("method: {}:{}", &req.method, e))
```
connorsmith256, line 43: Nit: 
```suggestion
            ProviderInvocationError::Provider(format!("failed to convert method: {}:{}", &req.method, e))
```

crates/providers/http-client/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+            .headers(headers)
+            .body(req.body)
+            .send()
+            .await
+            .map_err(|e| {
+                // send() can fail if there was an error while sending request,
+                // a redirect loop was detected, or redirect limit was exhausted.
+                // For now, we'll return an error (not HttpResponse with error
+                // status) and the caller should receive an error
+                // (needs to be tested).
```
connorsmith256, line 58: It looks like this was migrated from the existing code, but I'm trying to figure out what "needs to be tested" means. Surely we've checked this behavior? 😅

crates/providers/http-client/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        let status_code = response.status().as_u16();
+        let resp_headers = convert_header_map_to_hashmap(response.headers());
+
+        let body = response
+            .bytes()
+            .await
+            // error receiving the body could occur if the connection was closed before it was fully received
+            .map_err(|e| {
+                ProviderInvocationError::Provider(format!("receiving response body: {}", e))
```
connorsmith256, line 74:
```suggestion
                ProviderInvocationError::Provider(format!("error receiving response body: {}", e))
```

crates/providers/http-client/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+                // send() can fail if there was an error while sending request,
+                // a redirect loop was detected, or redirect limit was exhausted.
+                // For now, we'll return an error (not HttpResponse with error
+                // status) and the caller should receive an error
+                // (needs to be tested).
+                error!(
+                    error = %e,
+                    "httpclient network error attempting to send"
+                );
+                ProviderInvocationError::Provider(format!("sending request: {}", e))
```
connorsmith256, line 63:
```suggestion
                ProviderInvocationError::Provider(format!("error sending request: {}", e))
```

connorsmith256 (changes requested):
One blocking request, and a few nits. I'm assuming you'll split out the bindgen code from this commit once https://github.com/wasmCloud/wasmCloud/pull/963 is merged?

crates/providers/http-client/src/bin/httpclient.rs
```diff
@@ -1,156 +1,16 @@
...
-use wasmcloud_interface_httpclient::{HttpClient, HttpClientReceiver, HttpRequest, HttpResponse};
-
-// main (via provider_main) initializes the threaded tokio executor,
-// listens to lattice rpcs, handles actor links,
-// and returns only when it receives a shutdown message
-//
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    provider_main(
-        HttpClientProvider::default(),
+    tracing_subscriber::fmt().init();
```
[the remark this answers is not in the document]
vados-cosmonic: Sure, will do! I do think that people would want to be able to configure the OTEL stuff on their providers (possibly writing  out to a different place), but we'll cross that bridge when we get there I guess!

- [ ] remove `tracing_subscriber::fmt().init()`

crates/providers/http-client/src/lib.rs
vados-cosmonic: Went with a slightly different message here, but thanks for this!

crates/providers/http-client/src/lib.rs
vados-cosmonic: Fixed this with a slightly different message!

crates/providers/http-client/src/lib.rs
vados-cosmonic: Yeah I assume this is how the behavior was found... Maybe worth trying to add a test that does this explicitly in a follow up?

connorsmith256 (approved):

crates/providers/http-client/src/bin/httpclient.rs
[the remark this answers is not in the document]
connorsmith256: Do you think people want to use different otel settings for providers vs hosts? The providers currently inherit their otel settings from the host settings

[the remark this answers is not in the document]
vados-cosmonic: Yeah I would imagine that some providers are more important than others and you'd want to know more about one in particular. 

Like I'm thinking of specifically someone running a provider that they wrote w/ the existing host and wanting to shuttle their o11y outside to some other service to see how this provider was doing specifically. I think of it like the host is mostly boring and the providers are where more action happens -- and they would change more frequently.

That said, it's totally reasonable to hook the external o11y solution to the host and filter out for the provider too... Definitely a "we can cross that bridge when we get there" thing.

crates/providers/http-client/src/lib.rs
vados-cosmonic: Made [an issue](https://github.com/wasmCloud/wasmCloud/issues/995) for this

vados-cosmonic:
Ah sorry @connorsmith256 refactoring bug made it's way in there... hopefully you don't have to re-approve (I just rebased to one commit again)

crates/providers/http-client/src/bin/httpclient.rs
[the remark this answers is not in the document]
connorsmith256: IMO single-provider-only traces aren't very useful, since you won't see anything preceding or following those spans. IME distributed tracing provides the most value when you have end-to-end traces

But if we're thinking of just logs, I do see what you mean. Do you want to file an issue related to this? I'll leave this thread for you to resolve either way

thomastaylor312:
@connorsmith256 asked me to merge this for him

PR #748, opened 2023-10-13
chore: update to Wasmtime 15 and latest WIT

rvolosatovs:
## Feature or Problem

- Update to Wasmtime 15
- Since I had to update deps generally anyway, also ran `cargo update` in all workspaces

## Related Issues

https://github.com/wasmCloud/wasmcloud-component-adapters/pull/14
https://github.com/WebAssembly/wasi-logging/pull/16
https://github.com/WebAssembly/wasi-blob-store/pull/13
https://github.com/WebAssembly/wasi-keyvalue/pull/29
https://github.com/WebAssembly/wasi-blob-store/pull/14

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`


crates/actor/wit/deps/logging/world.wit
```diff
@@ -1,4 +1,4 @@
-package wasi:logging@0.2.0-rc-2023-10-18;
+package wasi:logging;
```
brooksmtownsend, line 1: These don't need to be pinned anymore because they are stable?

crates/runtime/src/actor/component/logging.rs
```diff
@@ -18,7 +18,7 @@ pub mod logging_bindings {
         world: "logging",
         async: true,
         with: {
-           "wasi:logging/logging@0.2.0-rc-2023-10-18": crate::capability::logging,
+           "wasi:logging/logging@0.2.0-rc-2023-11-10": crate::capability::logging,
         },
```
brooksmtownsend, line 22: Should this just be wasi:logging/logging or is this different than the wit

brooksmtownsend (commented):
Just a question for clarity but overall all looks good to me 👍🏻

crates/actor/wit/deps/logging/world.wit
rvolosatovs: It appears that it was simply removed from Wasmtime 15, so this is essentially a downgrade to upstream, which is not tagged

crates/runtime/src/actor/component/logging.rs
rvolosatovs: It just should not *be*, it's removed from Wasmtime and is completely ignored actually. I'll remove it

brooksmtownsend (approved):

PR #756, opened 2023-10-17
Chore/merge wash

connorsmith256:
This migrates https://github.com/wasmcloud/wash into the `wash` subdirectory of this repo, while preserving git history

Steps taken:
- clone wasmCloud
- clone wash
- move wash's history into a `wash` subdir using https://github.com/newren/git-filter-repo
```
cd wash
git-filter-repo.py --to-subdirectory-filter wash
cd -
```
- add a local "remote" for wash to wasmCloud
```
cd wasmCloud
git remote add wash ../wash
git fetch wash --no-tags
```
- merge wash into wasmCloud
```
git merge --allow-unrelated-histories wash/main
```


rvolosatovs (commented):
Is there a reason this is not a subtree merge https://stackoverflow.com/questions/6426247/merge-git-repository-in-subdirectory?

Done as-is we have to freeze all changes to `wash`, since merging them here if made in the other repo will be extremely difficult and we don't have CI for `wash` in this repo yet

connorsmith256:
@rvolosatovs what is the advantage of doing a subtree merge? Requiring a freeze is a theoretical downside of this approach, but I chose to do this now since there are no open PRs to wash other than dependabot, so there should be no practical impact. As for missing CI, that's something I can move out of `wash/.github` as soon as this is complete.

Is there an important difference between the two approaches I'm missing?

rvolosatovs:
> @rvolosatovs what is the advantage of doing a subtree merge? Requiring a freeze is a theoretical downside of this approach, but I chose to do this now since there are no open PRs to wash other than dependabot, so there should be no practical impact. As for missing CI, that's something I can move out of `wash/.github` as soon as this is complete.
> 
> Is there an important difference between the two approaches I'm missing?

With subtrees we preserve the original commit history (including hashes), which means e.g. we can merge from the original repo if necessary and things like commit references can still be correctly resolved.

connorsmith256:
@rvolosatovs I didn't realize this had changed the SHAs. That's something I explicitly wanted to keep. I assumed since the commit history was maintained so were the SHAs. Let's use a subtree merge instead 👍

PR #765, opened 2023-10-19
ci: publish latest tag

connorsmith256:
## Feature or Problem
I had a sad:
```
docker run wasmcloud/wasmcloud:latest
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
wasmcloud 0.18.2
```


connorsmith256:
I'm not sure how `buildah` works, so correct me if this is not the right approach 🙂

rvolosatovs (changes requested):
Originally, I did not add this until we were still working on stabilizing the Rust host.
Approach LGTM, but we need to make sure to only publish `latest` on non-RC versions, so I am afraid we will have to add a `if`-gated step which would check for a non-RC tag to publish latest

connorsmith256:
Thanks @rvolosatovs, I hadn't considered publishing over `latest` on `rc` tags 👍

rvolosatovs (approved):

Issue #608, opened 2023-08-31
Expand `wasi:blobstore` testing procedure


rvolosatovs:
Only the most basic operations are tested, that should be expanded and look similar to `keyvalue` tests

rimbi:
Where are these `keyvalue` tests exactly?

vados-cosmonic:
Closing as this isn't the case anymore!

PR #772, opened 2023-10-20
refactor: convert blobstore-fs to bindgen

vados-cosmonic:
## Feature or Problem

This PR implements `blobstore-fs` w/ the provider bindgen crate

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


crates/providers/blobstore-fs/src/lib.rs
```diff
@@ -1 +1,750 @@
...
+    Context,
+};
+
+mod fs_utils;
+use fs_utils::all_dirs;
+
+wasmcloud_provider_wit_bindgen::generate!({
+    impl_struct: FsProvider,
+    contract: "wasmcloud:blobstore",
+    wit_bindgen_cfg: "provider-blobstore"
```
rvolosatovs, line 34: what does this do?

rvolosatovs (approved):
passes tests, therefore it works (right?)

crates/providers/blobstore-fs/src/lib.rs
vados-cosmonic: The value on this line is actually passed through to the bindgen *underneath* `wasmcloud_provider_wit_bindgen` -- i.e. passed to the upstream code, so that the same configurability is available to the person calling generate.

Not all the forms are well documented, but it's using the form that is equivalent to the world name only, without any other special options being passed.

vados-cosmonic:
@rvolosatovs note that there is a TODO in the existing tests to *expand* the `blobstore` use cases/test cases... 

https://github.com/wasmCloud/wasmCloud/issues/608 should probably be prioritized as a follow up, but for now it is as tested as the current blobstore.

crates/providers/blobstore-fs/Cargo.toml
```diff
@@ -15,7 +18,9 @@ status = "actively-developed"
 async-trait = { workspace = true }
 path-clean = { workspace = true }
 serde = { workspace = true }
+serde_bytes = { workspace = true }
```
rvolosatovs, line 21: where is this used?

vados-cosmonic: it's a requirement that becomes used [during bindgen](https://github.com/wasmCloud/wasmCloud/pull/772/files#diff-c35e6db0016f48f0c6fe14869e51b1fbffd9c6d8db1be7933222b5d519e6045aR1608), because of [how the host currently expects binary data to be provided](https://github.com/wasmCloud/wasmCloud/blob/main/crates/compat/src/blobstore.rs#L15), which is [the more efficient way to pass bytes](https://docs.rs/serde_bytes/latest/serde_bytes/)

rvolosatovs: I see. Does not have to happen in this PR, but let's at least file an issue to get rid of the external dependency here and instead re-export and use `serde_bytes` from the bindgen crate itself here

vados-cosmonic: That would make this macro unhygenic -- generally we want our downstreams to choose the versions of the libs that we require. It's not like it's impossible, but if we risk incompatibility with a different dep (like `serde`) here if we do that. 

@thomastaylor312 should be able to cross-check this

rvolosatovs: From the guide you linked to, @vados-cosmonic:

> Procedural macros are _unhygienic_

Which seems to assert that "lack of hygiene" is an inherent property of all procedural macros.

Regardless, I don't really see how this is relevant here - I do not think that we should require users of this macro to add custom opaque dependencies - relying on re-exported dependencies is a pretty common practice in the ecosystem, which eliminates the guesswork and ensures compatibility.

rvolosatovs: Filed an issue

connorsmith256 (approved):

connorsmith256 (approved):

rvolosatovs (approved):

thomastaylor312 (approved):

Issue #776, opened 2023-10-21
Improve reference parsing


connorsmith256:
There are some common reference formats which the host does not gracefully parse right now. A few examples:

localhost:
```
wash start actor localhost:5000/v2/echo:0.1.0
```

docker container names:
```
wash start actor registry:5000/v2/echo:0.1.0
```

http/https registries:
```
wash start actor https://some-registry/v2/echo:0.1.0
```

If the reference looks like a URL, the host checks for `file`, `oci`, and `bindle` as schemes. The latter two aren't technically schemes, though these prefixes can help the host determine which client to use to connect to the given URL

Instead, the host should allow any "scheme". `oci`, `http`, and `https` should all be stripped and treated as OCI references, for simplicity. A more robust solution would be to interact with the provided server to determine if it's OCI, bindle, or something else, but for now defaulting to OCI is fine.

PR #777, opened 2023-10-21
chore: improve reference parsing

connorsmith256:
## Feature or Problem
Resolves https://github.com/wasmCloud/wasmCloud/issues/776

## Related Issues
https://github.com/wasmCloud/wasmCloud/issues/776

## Release Information
Next

## Consumer Impact
Users can use a broader set of references to start resources on a host

### Unit Test(s)
Added a unit test

### DIFF
--- a/crates/host/src/lib.rs
+++ b/crates/host/src/lib.rs
@@ -42,7 +42,7 @@ use std::path::PathBuf;
 
 use anyhow::{anyhow, bail, ensure, Context as _};
 use tokio::fs;
-use tracing::{instrument, warn};
+use tracing::{debug, instrument, warn};
 use url::Url;
 use wascap::jwt;
 
@@ -56,6 +56,7 @@ fn socket_pair() -> anyhow::Result<(tokio::io::DuplexStream, tokio::io::DuplexSt
     Ok(tokio::io::duplex(8196))
 }
 
+#[derive(PartialEq)]
 enum ResourceRef<'a> {
     File(PathBuf),
     Bindle(&'a str),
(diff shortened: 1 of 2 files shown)


crates/host/src/lib.rs
```diff
@@ -73,16 +74,29 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
+                            .map(Self::Oci)
+                            .context("invalid OCI reference")
+                    }
+                    scheme @ ("http" | "https") => {
+                        debug!(%url, "interpreting reference as OCI");
+                        s.strip_prefix(&format!("{scheme}://"))
+                            .map(Self::Oci)
+                            .context("invalid OCI reference")
+                    }
+                    _ => {
```
thomastaylor312, line 95: Nit: This is a minor concern since I consider this overall an improvement: but won’t this cause error messages to the user that say things like “invalid OCI reference” when they didn’t think they passed an OCI reference?

thomastaylor312 (approved):

crates/host/src/lib.rs
```diff
@@ -73,16 +74,29 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
+                        // handle strings like `registry:5000/v2/foo:0.1.0`
```
rvolosatovs, line 96: The ref in the comment should already be handled by the `Err` case below 

Can you give an example of a ref that would be parsed here?

crates/host/src/lib.rs
```diff
@@ -73,16 +74,29 @@ impl<'a> TryFrom<&'a str> for ResourceRef<'a> {
...
+                            .map(Self::Bindle)
+                            .context("invalid Bindle reference")
+                    }
+                    "oci" => {
+                        // Note: oci is not a scheme, but using this as a prefix takes out the guesswork
+                        s.strip_prefix("oci://")
+                            .map(Self::Oci)
+                            .context("invalid OCI reference")
+                    }
+                    scheme @ ("http" | "https") => {
```
rvolosatovs, line 89: Can you give an example of when this code path would be taken? What's the use case of passing an OCI ref with a `http(s)` scheme?

crates/host/src/lib.rs
connorsmith256: I'm not sure if you meant to comment a couple lines up? The `.context("invalid OCI reference")` lines should never be processed in practice, since they're handling the `None` case of `s.strip_prefix(...)`, and we've already matched against the `scheme` of the parsed string

The one potentially odd case for users is the fallback `_`, where we will print a debug message saying `unknown scheme in reference, assuming OCI` for URLs like `registry:5000/v2/echo:0.1.0`, but I think this is the best we can do without actually adding logic to intelligently query the server

crates/host/src/lib.rs
connorsmith256: Sure, consider a URL like in the test: https://github.com/wasmCloud/wasmCloud/blob/ff0f52c11e3d94e3fd03ee8bc7bc211de785d453/crates/host/src/lib.rs#L237

The intent here is to gracefully process more patterns of URLs. Most references shouldn't need an `http` or `https`, but while exercising this code, I as a user tried to get URLs like `registry:5000/v2/foo:0.1.0` or `localhost:5000/v2/foo:0.1.0` to parse, including appending `http://` to the front of them

crates/host/src/lib.rs
thomastaylor312: So I did mean to leave the comment here. I know that right now it will debug print the message, but most people won't have it set to debug. My main concern is that a "bare" reference (not passed with a schema) will be then passed to the oci-distribution crate, and if it is an invalid reference it will then say "invalid OCI reference" or "unknown image <reference>." Does that make any more sense?

thomastaylor312: Synced through this with @connorsmith256 offline and like we thought, there isn't much we can do here for the weird error cases

PR #778, opened 2023-10-21
chore: add logs related to registry config

connorsmith256:
This just adds some logs related to setting registry credentials


rvolosatovs (approved):
I'd make all of these logs `trace`, but sure

Issue #692, opened 2023-09-26
[REQUEST] Runtime Host Tags


jordan-rash:
I would like to be able to create tags on the host at runtime.

Something along the lines of 
`wash host <host-id> my:tag`

This would be a powerful way to start tuning the host without restart.  Currently, a problem I have is that if I forget to add a tag/label to the host, it requires a restart of the host. 

Ref: https://wasmcloud.slack.com/archives/C04RVMJJH17/p1695735058510099

connorsmith256:
I love this idea 🙌 

I think we'll want some more technical details ironed out:
- when you say "tags", do you mean the existing host labels, or a new, separate set of metadata?
- how will the host receive updates?
  - presumably a new topic on the control interface?
  - will the payload be treated as a full replacement, or an "upsert"?

jordan-rash:
By tags, I mean host labels (my mistake)

I don't know how the host should best receive them, but I do think they should be an insert and not a replace.

`wash host add-label <host-id> my:tag`
`wash host del-label <host-id> my`

connorsmith256:
@jordan-rash check out https://github.com/wasmCloud/wasmCloud/pull/780 and let me know if you have any feedback! Once the host/control interface supports this API, we can add it to wash pretty easily

connorsmith256:
@jordan-rash bump?

jordan-rash:
#780 looks great to me

PR #780, opened 2023-10-22
feat: enable updating host labels via the control interface

connorsmith256:
## Feature or Problem
This adds two new topics to the host's control interface: `wasmbus.ctl.{lattice_id}.labels.{host_id}.{put/del}`. Publishing on these topics allows users to dynamically adjust the target host's labels

## Related Issues
Relates to https://github.com/wasmCloud/wasmCloud/issues/692

## Release Information
Minor bump to the control interface crate as well as the host. Functionality via wash to be added after both of the former are released

## Consumer Impact
This is a potentially a step in the direction of enabling lame-duck mode on the host (we could use a special label to convey that intent)

### Acceptance or Integration
I exercised put and delete via the wasmbus tests

### Manual Verification
```
wash get hosts -o json | jq -r '.hosts | .[] | .labels'
{
  "hostcore.arch": "aarch64",
  "hostcore.os": "macos",
  "hostcore.osfamily": "unix"
}

nats req wasmbus.ctl.default.labels.NARZ43WTEW56DC3Q42FXIND2WJV5TFVTBYJBJ2T42S5OB2AHIPITDDWW.put '{"label": "foo", "value": "bar"}'

18:31:08 Sending request on "wasmbus.ctl.default.labels.NARZ43WTEW56DC3Q42FXIND2WJV5TFVTBYJBJ2T42S5OB2AHIPITDDWW.put"
18:31:08 Received with rtt 501.459µs
{"accepted":true,"error":""}

wash get hosts -o json | jq -r '.hosts | .[] | .labels'
{
  "foo": "bar",
  "hostcore.arch": "aarch64",
  "hostcore.os": "macos",
  "hostcore.osfamily": "unix"
}

nats req wasmbus.ctl.default.labels.NARZ43WTEW56DC3Q42FXIND2WJV5TFVTBYJBJ2T42S5OB2AHIPITDDWW.del '{"label": "foo"}'

18:31:20 Sending request on "wasmbus.ctl.default.labels.NARZ43WTEW56DC3Q42FXIND2WJV5TFVTBYJBJ2T42S5OB2AHIPITDDWW.del"
18:31:20 Received with rtt 551.917µs
{"accepted":true,"error":""}

wash get hosts -o json | jq -r '.hosts | .[] | .labels'
{
  "hostcore.arch": "aarch64",
  "hostcore.os": "macos",
  "hostcore.osfamily": "unix"
}
```

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -511,13 +511,13 @@ impl Client {
     pub async fn put_label(
         &self,
         host_id: &str,
-        label: &str,
+        key: &str,
         value: &str,
     ) -> Result<CtlOperationAck> {
         let subject = broker::put_label(&self.topic_prefix, &self.lattice_prefix, host_id);
         debug!(%subject, "putting label");
         let bytes = json_serialize(HostLabel {
-            label: label.to_string(),
+            key: key.to_string(),
             value: value.to_string(),
         })?;
         match self.request_timeout(subject, bytes, self.timeout).await {
@@ -531,11 +531,11 @@ impl Client {
     /// # Errors
     ///
     /// Will return an error if there is a communication problem with the host
-    pub async fn delete_label(&self, host_id: &str, label: &str) -> Result<CtlOperationAck> {
+    pub async fn delete_label(&self, host_id: &str, key: &str) -> Result<CtlOperationAck> {
         let subject = broker::delete_label(&self.topic_prefix, &self.lattice_prefix, host_id);
         debug!(%subject, "removing label");
         let bytes = json_serialize(HostLabel {
-            label: label.to_string(),
+            key: key.to_string(),
             value: String::new(), // value isn't parsed by the host
         })?;
         match self.request_timeout(subject, bytes, self.timeout).await {
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -503,6 +503,47 @@ impl Client {
         }
     }
 
+    /// Put a new (or update an existing) label on the given host.
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if there is a communication problem with the host
+    pub async fn put_label(
+        &self,
+        host_id: &str,
+        label: &str,
+        value: &str,
+    ) -> Result<CtlOperationAck> {
+        let subject = broker::put_label(&self.topic_prefix, &self.lattice_prefix, host_id);
+        debug!(%subject, "putting label");
+        let bytes = json_serialize(HostLabel {
+            label: label.to_string(),
+            value: value.to_string(),
+        })?;
+        match self.request_timeout(subject, bytes, self.timeout).await {
+            Ok(msg) => Ok(json_deserialize(&msg.payload)?),
+            Err(e) => Err(format!("Did not receive put label acknowledgement: {e}").into()),
+        }
+    }
+
+    /// Removes a label from the given host.
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if there is a communication problem with the host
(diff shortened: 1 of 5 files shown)


rvolosatovs (approved):

crates/control-interface/src/lib.rs
```diff
@@ -503,6 +503,47 @@ impl Client {
...
+    }
+
+    /// Removes a label from the given host.
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if there is a communication problem with the host
+    pub async fn delete_label(&self, host_id: &str, label: &str) -> Result<CtlOperationAck> {
+        let subject = broker::delete_label(&self.topic_prefix, &self.lattice_prefix, host_id);
+        debug!(%subject, "removing label");
```
vados-cosmonic, line 536: Hey could you add the `host_id` here? Maybe something like `removing label [{}] from host [{}]`?

crates/control-interface/src/lib.rs
```diff
@@ -503,6 +503,47 @@ impl Client {
...
+        let bytes = json_serialize(HostLabel {
+            label: label.to_string(),
+            value: String::new(), // value isn't parsed by the host
+        })?;
+        match self.request_timeout(subject, bytes, self.timeout).await {
+            Ok(msg) => Ok(json_deserialize(&msg.payload)?),
+            Err(e) => Err(format!("Did not receive remove label acknowledgement: {e}").into()),
```
vados-cosmonic, line 543: NIT: should this be capital (control-interface is internal right?), and could we `context` here?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3675,7 +3695,7 @@ impl Host {
 
     #[instrument(level = "trace", skip_all)]
     async fn handle_config_get_one(&self, entity_id: &str, key: &str) -> anyhow::Result<Bytes> {
-        trace!("getting config");
+        trace!("handling config");
```
vados-cosmonic, line 3698:
```suggestion
        trace!(%entity_id, "handling config");
```
To match the other one?

crates/control-interface/src/types.rs
```diff
@@ -418,3 +418,9 @@ pub struct LinkDefinition {
     pub contract_id: String,
     pub values: LinkSettings,
 }
+
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub struct HostLabel {
+    pub label: String,
```
brooksmtownsend, line 424: I'm going back and forth on this being `label` or `key`, did you consider `key` so it's just submitting kv pairs?

brooksmtownsend (approved):
I'm happy with this, and I think the topics fit into our current structure. We may want to change them when we audit the control interface topics but that's okay

crates/control-interface/src/types.rs
connorsmith256: Hmm I didn't think about this, but I like what you're saying. A "label" is really a key + value

I'll rename 👍

vados-cosmonic (approved):
LGTM :rocket:

crates/control-interface/src/lib.rs
vados-cosmonic: Ah thanks for clarifying this -- the output doesn't look too unreasonable

crates/control-interface/src/lib.rs
vados-cosmonic: hmnnn I know that as someone debugging something perhaps I'd really want to know *what* the label was removed from, but definitely shouldn't hold up this PR -- we'll cross that :bridge_at_night: if we get there

PR #782, opened 2023-10-23
ci: fix `if` conditional for crate release

rvolosatovs:
## Feature or Problem

See example run https://github.com/wasmCloud/wasmCloud/actions/runs/6614576526/job/17964950149

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -7,8 +7,12 @@ on:
     branches:
     - main
     tags:
+    - 'provider-archive-v[0-9].[0-9]+.[0-9]+'
+    - 'provider-archive-v[0-9].[0-9]+.[0-9]+-rc[0-9]+'
     - 'v[0-9].[0-9]+.[0-9]+'
     - 'v[0-9].[0-9]+.[0-9]+-rc[0-9]+'
+    - 'wascap-v[0-9].[0-9]+.[0-9]+'
+    - 'wascap-v[0-9].[0-9]+.[0-9]+-rc[0-9]+'
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/wasmcloud.yml
```diff
@@ -303,13 +307,14 @@ jobs:
...
 
-    name: publish ${{ matrix.crate }} to crates.io
     if: startsWith(github.ref, 'refs/tags/') && github.event_name == 'push'
     needs: cargo
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4.1.1
-      if: startsWith(github.ref, 'refs/tags/${{ matrix.crate }}-v')
-    - run: cargo publish --token ${{ secrets.CRATES_PUBLISH_TOKEN }}
+      if: startsWith(github.ref, format('refs/tags/{0}-v', matrix.crate))
```
connorsmith256, line 315: This looks equivalent to what we had before 🤔 

I know I must be missing something, since this step ran when we pushed the v0.79.1 tag, but I can't understand how, because the ref would have been `refs/tags/v0.79.1`, and didn't include a `-` at all

Can you point out what I'm missing?

rvolosatovs: Yes, I think `${{ .. }}` reppacement does not work here directly, we need to use a function

connorsmith256: Ohh, got it!

connorsmith256 (approved):

.github/workflows/wasmcloud.yml
connorsmith256: I just found out about `actionlint`, which gives a helpful explanation for this:
```
.github/workflows/wasmcloud.yml:312:11: if: condition "startsWith(github.ref, 'refs/tags/${{ matrix.crate }}-v')" is always evaluated to true because extra characters are around ${{ }} [if-cond]
    |
312 |       if: startsWith(github.ref, 'refs/tags/${{ matrix.crate }}-v')
```

Issue #757, opened 2023-10-17
[FEATURE] Move host label parsing to clap


thomastaylor312:
**Is your feature request related to a problem? Please describe.**

The current way we parse labels from environment variables is inside of library code rather than being at the level of command line arg definition with clap. This is an artifact from moving things over from the OTP host and makes it hard to follow what is going on with parsing data

**Describe the solution you'd like**

The label parsing code should be moved from https://github.com/wasmCloud/wasmCloud/blob/b07302fe0befb3c04cc4f3f4c3c58f8fd35be04a/crates/host/src/wasmbus/mod.rs#L1782 to the top level clap args in the `src` directory

MeenuyD:
Hello @thomastaylor312 I would like to work on this issue can you please tell me where should change needs for this issue

thomastaylor312:
Hey there @MeenuyD! Thanks for being willing to help. So the logic inside of this block: https://github.com/wasmCloud/wasmCloud/blob/b07302fe0befb3c04cc4f3f4c3c58f8fd35be04a/crates/host/src/wasmbus/mod.rs#L1770-L1793 should be handled somewhere right after here: https://github.com/wasmCloud/wasmCloud/blob/e4889ae8af1d404521ce0c823399e768f59733c8/src/main.rs#L345. If this is your first time contributing to a Rust project, please feel free to open a draft PR and we can help walk you though!

MeenuyD:
Should I do the whole logic from the mod.rs to the main.rs

MeenuyD:
Or we can create it as the function

professorabhay:
hey @thomastaylor312, I tried to resolve the issue and raised the [PR](https://github.com/wasmCloud/wasmCloud/pull/783).
Please take a look and let me know any changes you want. I am newbie with rust and its first time to contributing to WASMCLOUD.

PR #783, opened 2023-10-23
Solve #757 [DRAFT]

professorabhay:
Make the required changes as per need in #757


src/main.rs
```diff
@@ -343,6 +343,30 @@ async fn main() -> anyhow::Result<()> {
         .map(|labelpair| parse_label(labelpair))
         .collect::<anyhow::Result<HashMap<String, String>, anyhow::Error>>()
         .context("failed to parse labels")?;
+    let mut labels = HashMap::from([
```
thomastaylor312, line 346: So when you declare this variable, you are actually overriding the `labels` value declared on 339. The other thing is that these are the default labels so they should be the ones you start from and then `extend` them with the other user provided data. Generally the order of precedence in most projects is:

1. Command line flags
2. Env vars
3. Defaults

So what you'll probably want to do is start with:

```rust
let mut labels = HashMap::from([
    ("hostcore.arch".into(), ARCH.into()),
    ("hostcore.os".into(), OS.into()),
    ("hostcore.osfamily".into(), FAMILY.into()),
]);
```

Then `.extend` it with the stuff from 339-345

src/main.rs
```diff
@@ -343,6 +343,30 @@ async fn main() -> anyhow::Result<()> {
...
+            ("hostcore.arch".into(), ARCH.into()),
+            ("hostcore.os".into(), OS.into()),
+            ("hostcore.osfamily".into(), FAMILY.into()),
+        ]);
+        labels.extend(config.labels.clone().into_iter());
```
thomastaylor312, line 351: This line is no longer needed because it is coming from `args` above

thomastaylor312 (changes requested):
Great start here! There are a few changes needed around ordering. Let me know if you need any more clarification on Rust things!

professorabhay:
hey @thomastaylor312, Kindly take a look now !

crates/host/src/wasmbus/mod.rs
```diff
@@ -1766,31 +1766,7 @@ impl Host {
         } else {
             Arc::new(KeyPair::new(KeyPairType::Server))
         };
-
-        let mut labels = HashMap::from([
```
thomastaylor312, line 1770: You'll need to make sure you update places that were using `labels` to use `config.labels` instead in this function. You'll also need to remove some old imports you no longer need. If you run `cargo check --workspace`, you should see all of the compile errors and the warnings of imports you should remove

src/main.rs
```diff
@@ -336,13 +336,37 @@ async fn main() -> anyhow::Result<()> {
         policy_changes_topic: args.policy_changes_topic,
         policy_timeout_ms: args.policy_timeout_ms,
     };
-    let labels = args
+    let mut labels = HashMap::from([
+        ("hostcore.arch".into(), ARCH.into()),
```
thomastaylor312, line 340: You'll probably need to add this imports here. Once again, `cargo check` is your friend!

thomastaylor312 (changes requested):
The code itself is looking correct, but you're missing some imports and updates to where the labels are being used!

thomastaylor312:
Hey @professorabhay, just circling back to see if you needed any help

thomastaylor312:
Since it has been a bit since we’ve heard anything, I’m going to go ahead and close for now. Few free to reopen if you have more time!

PR #788, opened 2023-10-24
Migrate helm chart from wasmcloud-otp

joonas:
## Feature or Problem

Enables the installation of Rust-based wasmCloud host via `helm`.

## Related Issues

#684, also supersedes #785.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Issue #684, opened 2023-09-25
Migrate chart to repo


thomastaylor312:
The wasmCloud chart should be migrated from the old OTP repo: https://github.com/wasmCloud/wasmcloud-otp/tree/main/wasmcloud_host/chart. Now that we have rolled out the new rust host, it should also be updated to use the latest rust host (and actually tested that it works)

Needed steps

- [ ] Move chart to this repo
- [ ] Update app version to use the latest rust host images
- [ ] Make sure the build pipeline for publishing the chart is also migrates (it will involve GH pages being enabled)
- [ ] Test the chart to make sure it deploys properly

joonas:
Given #788, #802, #806 and #918, I think this can be considered done now?

PR #802, opened 2023-10-25
Add pipeline for validating and publishing the wasmcloud helm chart

joonas:
## Feature or Problem

Adds a pipeline that validates the helm chart (added in #788) and (on `chart-vX.X.X` push) publishes the helm chart to `ghcr.io/wasmcloud/wasmcloud-chart`.

## Related Issues

#684

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/.github/workflows/wasmcloud-chart.yml
+++ b/.github/workflows/wasmcloud-chart.yml
@@ -1,0 +1,83 @@
+name: wasmcloud-chart
+
+env:
+  HELM_VERSION: v3.13.1
+
+on:
+  push:
+    tags:
+      - 'chart-v[0-9].[0-9]+.[0-9]+'
+  pull_request:
+    paths:
+      - 'chart/**'
+      - '.github/workflows/wasmcloud-chart.yml'
+
+jobs:
+  validate:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v4.1.1
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v3.5
+        with:
+          version: ${{ env.HELM_VERSION }}
+
+      # Used by helm chart-testing below
+      - name: Set up Python
+        uses: actions/setup-python@v4.7.1
+        with:
+          python-version: '3.12.0'
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.4.0
+        with:
+          version: v3.9.0
+          yamllint_version: 1.32.0
+          yamale_version: 4.0.4
+
+      - name: Run chart-testing (lint)
+        run: |
+          ct lint --charts chart/ --validate-maintainers=false
+
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.8.0
+
+      - name: Run chart-testing (install)
+        run: ct install --charts chart/
+
+  publish:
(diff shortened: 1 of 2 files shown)


.github/workflows/wasmcloud-chart.yml
```diff
@@ -0,0 +1,83 @@
...
+          ct lint --charts chart/ --validate-maintainers=false
+
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.8.0
+
+      - name: Run chart-testing (install)
+        run: ct install --charts chart/
+
+  publish:
+    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/chart-v') }}
```
connorsmith256, line 51: Nit: isn't checking the event name redundant? I'm not sure there's a way for the ref to be a tag without the event being a push

.github/workflows/wasmcloud-chart.yml
```diff
@@ -0,0 +1,83 @@
...
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish
+        run: |
+          for chart in .helm-charts/*; do
+            if [ -z "${chart:-}" ]; then
+              break
+            fi
+            helm push "${chart}" "oci://ghcr.io/${{ github.repository_owner }}"
```
connorsmith256, line 82: I don't have any experience publishing charts, but did notice for the [old chart](https://github.com/wasmCloud/wasmcloud-otp/blob/main/.github/workflows/release.yml) we use the `helm/chart-releaser-action` action. Is there a reason we shouldn't use that anymore?

connorsmith256 (commented):
LGTM, with the caveat that I don't have domain knowledge on helm charts. Can approve if needed, but will defer to others for now

.github/workflows/wasmcloud-chart.yml
joonas: > Nit: isn't checking the event name redundant? I'm not sure there's a way for the ref to be a tag without the event being a push

It is, I'll drop the `github.event_name` check here.

.github/workflows/wasmcloud-chart.yml
joonas: So the [`helm/chart-releaser-action`](https://github.com/helm/chart-releaser-action) is primarily intended for publishing charts to be self-hosted on GitHub Pages, which is not how we intend to host the chart. Instead, we're looking to publish the chart onto GitHub Container Registry as an OCI artifact.

The `chart-releaser-action` has an [issue open regarding adding support for OCI registries](https://github.com/helm/chart-releaser-action/issues/107), which would let us use it in the future should they choose to implement that functionality.

connorsmith256: Ah, got it! Thanks 👍

thomastaylor312 (approved):

PR #804, opened 2023-10-25
feat(http): implement `wasi:http/outgoing-handler`

### LINKED ISSUE #680: implement `wasi:http/outgoing-handler`

rvolosatovs:
## Feature or Problem

Implement `wasi:http/outgoing-handler`

## Related Issues

Closes #680 

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

`cargo test`


crates/runtime/src/actor/component/http.rs
```diff
@@ -33,23 +35,62 @@ pub mod incoming_http_bindings {
...
+            {
+                Ok(resp) => {
+                    let resp = resp.map(|body| BoxBody::new(AsyncReadBody::new(body, 1024)));
+                    Ok(IncomingResponseInternal {
+                        resp,
+                        worker: preview2::spawn(async { Ok(()) }),
+                        between_bytes_timeout,
+                    })
+                }
+                Err(e) => Err(e),
```
connorsmith256, line 86: If we don't need to do anything with the error case, how about `map` instead of `match`?

connorsmith256 (approved):

thomastaylor312 (approved):

crates/runtime/src/actor/component/http.rs
rvolosatovs: Sounds good, but I will not be addressing that here to avoid extending the review cycle, since this PR is blocking #745 and potentially may cause conflicts with #807

PR #805, opened 2023-10-25
fix: stop sending linkdef events on startup

connorsmith256:
## Feature or Problem
This PR splits processing historical lattice metadata entries into two parts:
- on startup, the host reads all entries out of the bucket and initializes its cache of link defs/claims. It should not publish link def put/deleted events for historical entries
- on startup, the host starts listening to new entries. For new entries, the host should publish link def put/deleted events

## Release Information
Next

## Consumer Impact
The host should simply emit fewer events on startup

### Manual Verification
Ran the host, confirmed a bunch of `process_entry` logs were emitted, but no `linkdef_set` events were, until I ran `wash link put`

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -34,7 +34,7 @@ use base64::Engine;
 use bytes::{BufMut, Bytes, BytesMut};
 use cloudevents::{EventBuilder, EventBuilderV10};
 use futures::stream::{AbortHandle, Abortable};
-use futures::{join, stream, try_join, FutureExt, Stream, StreamExt, TryFutureExt};
+use futures::{join, stream, try_join, FutureExt, Stream, StreamExt, TryFutureExt, TryStreamExt};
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1993,11 +1993,30 @@ impl Host {
                 }
             }
         });
+
+        // Process existing data without emitting events
+        data.keys()
```
rvolosatovs, line 1998: There is a race condition here and since it's probably better to process events at-least-once rather than at-most-once (we cannot do exactly-once I believe due to lack of "transactions" in NATS), this block should be placed *after* the `watch`

rvolosatovs (changes requested):

crates/host/src/wasmbus/mod.rs
connorsmith256: Good point 👍

brooksmtownsend (approved):

PR #812, opened 2023-10-27
feat: add WASI-enabled http echo tinygo example

vados-cosmonic:
## Feature or Problem

Component model and WASI support for TinyGo and `wash` ergonomics are finally at a point where we can introduce examples that are written then built exclusively with `wash build`.

This commit adds a new WASI-enabled component example that adapts and contributes the hard work of @jordan-rash and
wraps in some more utility scripts and configuration.

## Related Issues

https://github.com/wasmCloud/examples/pull/245

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows


examples/golang/actors/http-echo-tinygo/wadm.yaml
```diff
@@ -0,0 +1,44 @@
...
+  annotations:
+    version: v0.0.1
+    description: "HTTP Echo demo in Golang (TinyGo), using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)"
+spec:
+  components:
+    - name: http-echo-tinygo
+      type: actor
+      properties:
+        # TODO: you must replace the path below to match your genreated code in build
+        image: file:///home/mrman/code/work/cosmonic/forks/echo-wasi/build/http-echo-tinygo-component_s.wasm
```
connorsmith256, line 25: Can we make this path generic/a placeholder?

examples/golang/actors/http-echo-tinygo/wasmcloud.toml
```diff
@@ -0,0 +1,34 @@
...
+#
+# As of 2023, components are cutting edge WebAssembly, and are thus experimental
+# features in wasmCloud. 
+#
+# That said, this example *should* compile and run on wasmCloud:
+#
+# 1. Build the project with `wash build`
+# 2. Deploy the project with `wash app deploy wadm.yaml` 
+#
+# Please file an issue if you encounter any friction (https://github.com/wasmCloud/examples/issues/new)
```
connorsmith256, line 17: Let's update this to the main repo

connorsmith256 (commented):
LGTM, just two nits

examples/golang/actors/http-echo-tinygo/wadm.yaml
vados-cosmonic: Ah sorry this didn't make it [over during the copy](https://github.com/wasmCloud/examples/pull/245/files#diff-fe1c89408d46e810c1f20806d2a35c07e718358c887d50c256dc02d19c185330R69-R71). will fix

vados-cosmonic: Done!

examples/golang/actors/http-echo-tinygo/wasmcloud.toml
vados-cosmonic: Fixed!

examples/golang/actors/http-echo-tinygo/wit/deps.toml
```diff
@@ -0,0 +1,6 @@
+blobstore = "https://github.com/WebAssembly/wasi-blob-store/archive/main.tar.gz"
+logging = "https://github.com/WebAssembly/wasi-logging/archive/main.tar.gz"
+http = "https://github.com/wasmCloud/wasi-http/archive/main.tar.gz" # TODO: Switch to upstream once Wasm tooling compatible
+poll = "https://github.com/WebAssembly/wasi-poll/archive/main.tar.gz" # TODO: Remove once keyvalue is compatible
+# io = "https://github.com/WebAssembly/wasi-io/archive/main.tar.gz" # TODO: Remove once keyvalue is compatible
+io = "./overrides/io" # Pin to newer version of streams contract
```
rvolosatovs, line 6: None of this will work with 0.80 (latest `main`), you should directly import from `crates/runtime/wit/deps` for now, but that would pull in resources, which I believe is not supported yet by the tooling (?)

So I guess there's nothing to do here (yet), just be aware that this will not actually run on the latest host?  @ricochet

vados-cosmonic: With the conversation from earlier today it looks like the move was to merge this for now and either use it with v79 or find out if it we can upgrade the example later? @connorsmith256 @ricochet are we aligned on this?

ricochet: Golang examples need to be pinned to v79. There isn't an easy way to do that with the current tooling. Unless we cut a `release-0.79` and reference this in the toml instead of `main`.

The alternative is to use relative paths since they allow for atomic commits with the host. To make this work for golang today (similar to the current version on this branch that was built with an older version of main) is to be in a half-way state where we keep the committed deps on an older version of the WIT definitions. Pinning to main is a weird place in a monorepo where it is the last commit on main.

My preference is the release branch approach as it's very clear about the state of the world.

vados-cosmonic: For this PR, then the idea here is to add a warning about the version. The average user is going to need to configure *wash* to use the right version `0.79.0`, with something like:

```console
WASMCLOUD_VERSION=v0.79.0 wash up
```

So that's what I'll put in the README for the example.

vados-cosmonic: PR has been updated -- I added a warning about the required version, and changed the `wash up` command

connorsmith256 (approved):

rvolosatovs (approved):

PR #838, opened 2023-10-30
[BUG] wash claims sign allows empty versions and revisions

ahmedtadde:
## Feature or Problem
Currently, when signing an actor with wash, the version (--ver) and revision (--rev) args are not required. This will lead to claims that are missing some important information.

## Related Issues

[859](https://github.com/wasmCloud/wasmCloud/issues/861)

## Release Information
v0.79.0

## Consumer Impact
stricter `wash claims sign` cmd that ultimately improves the metadata output of `wash inspect`

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
updated integration tests [integration_claims_sign](https://github.com/ahmedtadde/wash/blob/0939a7d4b1c8c229386baedccbaef9f3621071c6/tests/integration_claims.rs#L12), [integration_claims_call_alias](https://github.com/ahmedtadde/wash/blob/21f9f2f4d751e29e88fd8990ef9014a310ed425a/tests/integration_claims.rs#L328)

### DIFF
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -177,9 +177,8 @@ fn sign_actor_wasm(
         destination: Some(destination),
         metadata: ActorMetadata {
             name: common_config.name.clone(),
-            ver: Some(common_config.version.to_string()),
-            //NOTE(Ahmed Tadde): sign_file requires a revision, using 1 as default
-            rev: Some(1),
+            ver: common_config.version.to_string(),
+            rev: common_config.revision,
             custom_caps: actor_config.claims.clone(),
             call_alias: actor_config.call_alias.clone(),
             issuer: signing_config.issuer,
@@ -697,6 +696,7 @@ world downstream {
                 &CommonConfig {
                     name: "test".into(),
                     version: Version::parse("0.1.0")?,
+                    revision: 0,
                     path: project_dir.path().into(),
                     wasm_bin_name: Some("test.wasm".into()),
                 },
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -178,6 +178,8 @@ fn sign_actor_wasm(
         metadata: ActorMetadata {
             name: common_config.name.clone(),
             ver: Some(common_config.version.to_string()),
+            //NOTE(Ahmed Tadde): sign_file requires a revision, using 1 as default
+            rev: Some(1),
             custom_caps: actor_config.claims.clone(),
             call_alias: actor_config.call_alias.clone(),
             issuer: signing_config.issuer,
--- a/crates/wash-lib/tests/parser/files/rust_actor_with_revision.toml
+++ b/crates/wash-lib/tests/parser/files/rust_actor_with_revision.toml
@@ -1,0 +1,18 @@
+language = "rust"
+type = "actor"
+name = "testactor"
+version = "0.1.0"
+revision = 666
+
+[actor]
+claims = ["wasmcloud:httpserver"]
+registry = "localhost:8080"
+push_insecure = false
+key_directory = "./keys"
+filename = "testactor.wasm"
+wasm_target = "wasm32-unknown-unknown"
+call_alias = "testactor"
+
+[rust]
+cargo_path = "./cargo"
+target_path = "./target"
--- a/crates/providers/Cargo.lock
+++ b/crates/providers/Cargo.lock
@@ -3182,7 +3182,7 @@ dependencies = [
  "serde_json",
  "wasm-encoder 0.36.1",
  "wasm-gen",
- "wasmparser 0.115.0",
+ "wasmparser 0.116.0",
 ]
 
 [[package]]
@@ -3564,9 +3564,9 @@ dependencies = [
 
 [[package]]
 name = "wasmparser"
-version = "0.115.0"
+version = "0.116.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e06c0641a4add879ba71ccb3a1e4278fd546f76f1eafb21d8f7b07733b547cd5"
+checksum = "53290b1276c5c2d47d694fb1a920538c01f51690e7e261acbe1d10c5fc306ea1"
 dependencies = [
  "indexmap 2.0.2",
  "semver",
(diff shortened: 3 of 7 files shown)


crates/wash-lib/src/build.rs
```diff
@@ -180,6 +180,8 @@ fn sign_actor_wasm(
         metadata: ActorMetadata {
             name: common_config.name.clone(),
             ver: Some(common_config.version.to_string()),
+            //NOTE(Ahmed Tadde): sign_file requires a revision, using 1 as default
+            rev: Some(1),
```
brooksmtownsend, line 184: I think what we should do here is use `Some(0)`, as `0` is treated in wascap by grabbing the current timestamp when signing. Basically if a revision isn't supplied or supplied as 0, then we'll default to a monotonically (mostly) increasing number by default. 
Can @connorsmith256 or @autodidaddict check me on this logic?

brooksmtownsend (changes requested):
@ahmedtadde I think we should update the CLI modules in `wash-lib` to make revision and version required, e.g. https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-lib/src/cli/claims.rs#L269. That way you cannot forget to supply the value when using `wash claims sign` directly. Additionally with the build command we should supply the revision from the common_config

autodidaddict:
I don't know about the Rust host, but always setting revision to `1` on signed modules will absolutely break updates. Revisions need to either monotonically increase _or_ stay at zero where they are assigned a value based on the clock time at startup

crates/wash-lib/src/build.rs
autodidaddict: Correct. If we use a `1` here we'll break developer workflows by blocking updates with no version change.

brooksmtownsend: Thanks for double checking me, yeah so I think we should set this to `Some(0)`

crates/wash-lib/tests/parser/files/rust_actor_with_revision.toml
```diff
@@ -0,0 +1,18 @@
+language = "rust"
+type = "actor"
+name = "testactor"
+version = "0.1.0"
+revision = 666
```
brooksmtownsend, line 5: 👹

brooksmtownsend (approved):

brooksmtownsend:
@ahmedtadde Looks good to me once tests are working! I'm glad we can keep revision "optional" in the wasmcloud.toml file and default to `0` for them.

ahmedtadde:
> @ahmedtadde Looks good to me once tests are working! I'm glad we can keep revision "optional" in the wasmcloud.toml file and default to `0` for them.

so, for the tests, i am not able to reproduce this https://github.com/wasmCloud/wasmCloud/actions/runs/6711700776/job/18239654369?pr=838 locally? also, i am confused about the error message... looking at the code, i don't see a `tags` field on the `ActorConfig` struct

ahmedtadde:
> > @ahmedtadde Looks good to me once tests are working! I'm glad we can keep revision "optional" in the wasmcloud.toml file and default to `0` for them.
> 
> so, for the tests, i am not able to reproduce this https://github.com/wasmCloud/wasmCloud/actions/runs/6711700776/job/18239654369?pr=838 locally? also, i am confused about the error message... looking at the code, i don't see a `tags` field on the `ActorConfig` struct

just saw that the `tags` field is from a change in `main`. will try to rebase.

brooksmtownsend:
@ahmedtadde Ah I see that now too. Thanks for doing that rebase, we had a few edits going into `build` at the same time 🤔

crates/wash-lib/tests/parser/files/rust_actor_with_revision.toml
thomastaylor312, line 5: Suggestion here: I think we shouldn't require this in the config file (though we should have it so people can override) and instead just use the unix timestamp or something similar. In the past our makefiles and other things set this so it feels better as something automatic unless the user wants to specify an override.

@brooksmtownsend @autodidaddict any thoughts?

brooksmtownsend: Gah. This looks like something that's present in the OTP host NIF but may not be present in wash or in the wasmCloud runtime. The function is called `revision_or_iat`
```
631:pub(crate) fn revision_or_iat(rev: Option<i32>, iat: u64) -> Option<i32> {
632-    if rev.is_some() && rev.unwrap() > 0 {
633-        rev
634-    } else {
635-        Some(iat as i32)
636-    }
637-}
```

autodidaddict: yep this is the code that produces a revision if there wasn't one by reading the `iat` from the JWT. I'd just assumed the new Rust host did this as well.

brooksmtownsend: I think we have the correct implementation here then `Some(0)`, ideally we should get rid of the ability for that to be optional. In the Rust host we should then treat `0` or `None` revisions as an indicator to grab the `iat`

thomastaylor312: Yep, that sounds right

ahmedtadde: so, in terms of the `iat` fallback logic, should it be in this PR or another one. If it is the former, some pointer(s) on where to look would be great; i looked around and i saw a bunch of `unwrap_or_else_default()` for `revision` and `version` stuff in the `host` crate.

brooksmtownsend: I think in this PR we can only keep the logic to supply `Some(0)` as a default, and we'll need to transfer that revision functionality to wasmCloud. Alternatively, we could set up a nice helper in `wascap` to return the revision or the iat depending on the value. So @ahmedtadde I think there's nothing else to do here for wash.

PR #868, opened 2023-10-30
fix(ci): use wash Dockerfile

connorsmith256:
## Feature or Problem
I noticed the OCI build [references a file that doesn't exist](https://github.com/wasmCloud/wasmCloud/actions/runs/6698481609/job/18201730394), so I updated the workflow


rvolosatovs (approved):
Interesting why there was no error 🤔

PR #869, opened 2023-10-30
chore: sync Cargo.lock

connorsmith256:
I'm not sure how this got out of sync, since the bump was from dependabot...


rvolosatovs (approved):
Yes, this happens, because dependabot does not actually run `cargo build` or something else that would trigger the lock update for these subcrates outside the workspace

brooksmtownsend (approved):

PR #873, opened 2023-10-30
fix(ci): fix parameter name

connorsmith256:
## Feature or Problem
`dockerfile` isn't a valid parameter name. I think we want `file`: https://github.com/wasmCloud/wasmCloud/actions/runs/6698978154


rvolosatovs (approved):
I would expect an error for specifying a nonexistent parameter 🤔

PR #890, opened 2023-10-31
chore: un-hide policy options

connorsmith256:
## Feature or Problem
@jordan-rash pointed out these options were hidden, so I un-hid them 🙂


jordan-rash:
Thank you @connorsmith256

thomastaylor312 (approved):

Issue #641, opened 2023-09-13
Publish new crates


connorsmith256:
As part of the effort to implement the new host, we've created several new crates which allow for code re-use. Most of these crates are used internally, but there are a few which should be published, for at least two reasons:

- allow capability providers to be built outside of the `wasmCloud` repo
- enable code re-use, e.g. wadm could depend on `wasmcloud-tracing` instead of the copy-pasted tracing/logging setup

This is the list of crates I believe need to be published. I think only the last crate depends on any of the others:
- `opentelemetry-nats`
- `wasmcloud-tracing`
- `wasmcloud-core`
- `wasmcloud-provider-wit-bindgen`
- `wasmcloud-provider-sdk`

Also, `opentelemetry-nats` should be moved to its own repo, since it's wasmCloud-agnostic

PR #893, opened 2023-11-01
fix(releases): use version numbers of internal crates to allow publishing to crates.io

connorsmith256:
## Feature or Problem
Publishing `wash-lib` failed because our workspace uses wildcard constraints:
```
  the remote server responded with an error: wildcard (`*`) dependency constraints are not allowed on crates.io. Crate with this problem: `provider-archive` See 
```

https://github.com/wasmCloud/wasmCloud/actions/runs/6723759809/job/18274709187

This updates the workspace dep versions to not use `*` in cases where we want to publish crates for downstream consumption, e.g. `wash-lib`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ opentelemetry-nats = { version = "*", path = "./crates/opentelemetry-nats", defa
 opentelemetry-otlp = { version = "0.13", default-features = false }
 path-absolutize = { version = "3", default-features = false }
 proc-macro2 = { version = "1", default-features = false }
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):

brooksmtownsend (approved):
Are there any others that need to change? Feel free to just de-wildcard all of them 🤔

connorsmith256:
@brooksmtownsend more will need to change in the future, when we work on https://github.com/wasmCloud/wasmCloud/issues/641. But for releasing wash/wash-lib this week, I think this is all of them

PR #896, opened 2023-11-01
chore: fix path aliases tsconfig

yordis:
@lachieh and I discussed about refactoring the codebase to make sure the paths aren't conflicting with packages and it is more explicit what comes from the app-level code.

You can read more about the thought process behind it: https://straw-hat-team.github.io/adr/2068541249/README.html


washboard-ui/src/hosts/index.ts
yordis, line 1: Avoid re-exports like these in app-level code (leverage it in libs)

lachieh (approved):
Thanks @yordis! I'm actually thinking that if you want to reorganize into the structure that matches your ADR, it's a great time to do that! Approved if you want it in a different change but I think that's something we'll want to do soon.

PR #902, opened 2023-11-02
ci: allow any semver compliant tag

connorsmith256:
@ricochet pointed out that semver allows anything after the `-`, so we might as well be more permissive


ricochet (approved):

brooksmtownsend (approved):
What is this designated for, provider-archive-v0.9.0-😎?

rvolosatovs (approved):

PR #917, opened 2023-11-03
Bump Chart.yaml appVersion to 0.80.0

joonas:
## Feature or Problem

Since [`v0.80.0`](https://github.com/wasmCloud/wasmCloud/releases/tag/v0.80.0) has been released, let's bump the Chart.yaml to use that instead of `v0.79.1`.


joonas:
Let's roll with #918

Issue #875, opened 2023-10-30
[FEATURE] Update wash to use pre-built assets from releases


lachieh:
Once #874 is complete, we should update wash so that when running `wash ui` it will:
- check for local assets and download if none exist
- check for updated release and download if newer than local
- set specific release with option like `wash ui --version <washboard release>`

connorsmith256:
@lachieh I propose we remove the second bullet point from the requirements here and file as a separate nice-to-have

lachieh:
Was just going to ask about that—agreed. Want to split it out into a new issue as a `good-first-issue`?

lachieh:
No, nevermind, let's add it to #848

PR #919, opened 2023-11-03
feat(wash): download washboard assets from releases instead of embedding from source

connorsmith256:
## Feature or Problem
This changes `wash ui` to download the washboard assets from a released version instead of embedding local files, which will enable publishing wash

The version defaults to v0.1.0 for now, but can be specified with `--version`:

```
./target/debug/wash ui --version v0.1.0
Washboard running on http://localhost:3030
Hit CTRL-C to stop
```

@lachieh do we need to allow overriding this behavior for local dev? Or should local dev always be done via `make run-ui`?

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/875

## Release Information
v0.22.0

### Manual Verification
Built and ran locally, and confirmed I could serve the washboard UI

### DIFF
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -11,7 +11,6 @@ on:
     - Cargo.toml
     - crates/wash-cli/**
     - crates/wash-lib/**
-    - washboard-ui/**
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


thomastaylor312 (approved):

lachieh:
> @lachieh do we need to allow overriding this behavior for local dev? Or should local dev always be done via make run-ui?

No, we can run through the washboard npm script with `npm run dev`.

lachieh (approved):

PR #920, opened 2023-11-03
Convert the organization name to lowercase for publishing to ghcr.io

joonas:
## Feature or Problem

GitHub Container Registry namespaces are required to be in all lowercase, so we can't use the `github.repository_owner` directly.

### DIFF
--- a/.github/workflows/wasmcloud-chart.yml
+++ b/.github/workflows/wasmcloud-chart.yml
@@ -73,11 +73,15 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Lowercase the organization name for ghcr.io
+        run: |
+          echo "GHCR_REPO_NAMESPACE=${GITHUB_REPOSITORY_OWNER,,}" >>${GITHUB_ENV}
+
       - name: Publish
         run: |
           for chart in .helm-charts/*; do
             if [ -z "${chart:-}" ]; then
               break
             fi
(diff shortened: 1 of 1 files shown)


.github/workflows/wasmcloud-chart.yml
```diff
@@ -73,11 +73,15 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Lowercase the organization name for ghcr.io
+        run: |
+          echo "GHCR_REPO_NAMESPACE=${GITHUB_REPOSITORY_OWNER,,}" >>${GITHUB_ENV}
```
rvolosatovs, line 78: Aren't we missing the package name here? For context, that's what we do for OCI https://github.com/wasmCloud/wasmCloud/blob/e383accd311b14e7bf85c536f37ae09e90a1d2c1/.github/workflows/wasmcloud.yml#L242

joonas: Helm will automatically append the chart name to the remote it's given, so it'll turn `oci://ghcr.io/wasmcloud` into `oci://ghcr.io/wasmcloud/<chart-name>` (in our case ``oci://ghcr.io/wasmcloud/wasmcloud-chart`)

rvolosatovs: Oh, I see, good to know, thanks!

rvolosatovs (approved):

thomastaylor312 (approved):

PR #922, opened 2023-11-03
ci(wash): disable docker build and push temporarily

connorsmith256:
## Feature or Problem
This disables the docker build-and-push step for wash, since we're currently getting 403s: https://github.com/wasmCloud/wasmCloud/actions/runs/6750332769/job/18352964847

Disabling this should allow the github release to proceed


thomastaylor312 (approved):

rvolosatovs (approved):
Maybe instead of re-enabling this we should just switch to exact same OCI publish process we use for `wasmCloud`, which does not seem to be affected by this issue?

Issue #724, opened 2023-10-05
[FEATURE] add RISCV64 build


ricochet:
[wasmtime has tier 3 support](https://docs.wasmtime.dev/stability-tiers.html) for riscv64. It would be amazing if wasmCloud provided a host binary for `riscv64gc-unknown-linux-gnu`

**Describe the solution you'd like**
I noticed that wasmtime has recently had several contributions related to riscv64 and I'd love to be able to run wasmCloud on some of my microcontrollers.

rvolosatovs:
Unfortunately, this is currently blocked on `ring` being updated to 0.17, this PR specifically is what we need, which landed in that release https://github.com/briansmith/ring/pull/1627

`ring` is our transitive dependency:

```
$ cargo tree -i ring
ring v0.16.20
├── provider-archive v0.7.0
│   └── wasmcloud-host v0.79.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud/crates/host)
│       └── wasmcloud v0.79.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud)
│   [build-dependencies]
│   └── test-providers v0.1.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud/tests/providers)
│       [dev-dependencies]
│       └── wasmcloud v0.79.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud)
└── wascap v0.11.2
    ├── provider-archive v0.7.0 (*)
    ├── wasmcloud-core v0.1.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud/crates/core)
    │   ├── wasmcloud v0.79.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud)
    │   ├── wasmcloud-host v0.79.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud/crates/host) (*)
    │   └── wasmcloud-tracing v0.1.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud/crates/tracing)
    │       ├── wasmcloud v0.79.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud)
    │       └── wasmcloud-host v0.79.0 (/home/rvolosatovs/src/github.com/wasmCloud/wasmCloud/crates/host) (*)
[...]

PR #927, opened 2023-11-06
integrate `wasmcloud-control-interface` into the repository

rvolosatovs:
## Feature or Problem

Integrate https://github.com/wasmCloud/control-interface-client into the repo. While working on that, I have removed a bunch of unused dependencies and even unused code.

Funny enough, the `ring` dependency in this crate was blocking us from implementing #724, but it was never actually used by the crate


.github/workflows/wasmcloud.yml
```diff
@@ -493,18 +495,23 @@ jobs:
     strategy:
       matrix:
         crate:
+        - control-interface
         - provider-archive
         - wascap
         - wash-cli
         - wash-lib
 
-    if: startsWith(github.ref, 'refs/tags/') && github.event_name == 'push'
```
brooksmtownsend, line 501: nit: Now that we have the dry run, we could restrict the publish to only happen on the tag push right?

Does this action not being gated by tags mean we're just going to release new versions of the crate on merge? I don't think we've designed our workflow for that

.github/CODEOWNERS
```diff
@@ -28,6 +28,7 @@ Cargo.* @wasmCloud/host-maintainers
 .cargo @wasmCloud/host-maintainers
 
 # shared crates
+crates/control-interface @autodidaddict @brooksmtownsend @stevelr @connorsmith256
```
brooksmtownsend, line 31:
```suggestion
crates/control-interface @wasmCloud/control-interface-maintainers
```

Cargo.toml
```diff
@@ -166,25 +166,25 @@ wadm = { version = "0.7", default-features = false }
 walkdir = { version = "2", default-features = false }
 warp = { version = "0.3", default-features = false }
 warp-embed = { version = "0.4", default-features = false }
-wascap = { version = "0.11", path = "./crates/wascap", default-features = false }
-wash-cli = { version = "0.22", path = "./crates/wash-cli", default-features = false }
-wash-lib = { version = "0.13", path = "./crates/wash-lib", default-features = false }
+wascap = { version = "0", path = "./crates/wascap", default-features = false }
+wash-cli = { version = "0", path = "./crates/wash-cli", default-features = false }
+wash-lib = { version = "0", path = "./crates/wash-lib", default-features = false }
```
brooksmtownsend, line 171: The reason why we explicitly versioned these from `*` was so that crates.io would publish dependents and know what version to use, does using `version = "0"` let this happen? If I remember correctly the dry run didn't catch it before

brooksmtownsend (changes requested):
Requests for process reasons but the crate move looks good

.github/workflows/wasmcloud.yml
rvolosatovs: Crates are only published when the matching tag is pushed

Cargo.toml
rvolosatovs: It should, but there's really only one way to find out if dry run does not catch it

brooksmtownsend (approved):
Once we merge and attempt to cut a tag, let's see if specifying 0 works for us. If it's the semver 0 major version then I would assume that works

PR #938, opened 2023-11-06
feat: wash build enhancements --sign-only and build_command

brooksmtownsend:
- feat(wash-cli): enable only signing actors
- feat(wash)!: add support for custom build command
- chore(wash-cli): enumerate wasm targets in README

## Feature or Problem
This PR adds a few beneficial changes to the `wash build` workflow. Supplying `wash build --sign-only` will skip the build process which can be nice for using the project config to simply sign an existing artifact. 

This PR also adds the escape hatch in `wasmcloud.toml` to run a custom build command for the project instead of using the default build command. This is _really_ nice for running little scripts or supporting languages that we don't technically have first class support for. Right now we'll still bail out with an error `unknown language in wasmcloud.toml` but once I can test this code on a real example I'll remove this.

## Release Information
v0.25.0

## Consumer Impact
This PR includes new features in a backwards compatible way so if it's not released in v0.22 it will go out as v0.23.

### Manual Verification
I manually ran these commands to see how the build command worked!


crates/wash-lib/src/build.rs
```diff
@@ -96,19 +99,48 @@ pub fn build_actor(
...
+        let mut split_command = raw_command.split_ascii_whitespace();
+        let command = split_command
+            .next()
+            .context("build command is supplied but empty")?;
+
+        // Change directory into the project directory
+        std::env::set_current_dir(&common_config.path)?;
+
+        let mut command = process::Command::new(command);
+        command.args(split_command).status().map_err(|e| {
```
connorsmith256, line 112: Readability nit: instead of continuing to use the mut iterator where we've consumed the first token, I would collect the rest of the iterator to an `args` var above, and pass that collection here. At first thought this would include the first token

crates/wash-lib/src/build.rs
```diff
@@ -117,28 +149,30 @@ pub fn build_actor(
             &actor_wasm_path,
             &actor_wasm_path,
         )?;
-    };
+        };
+
+        // If the actor has been configured as WASI Preview2, adapt it from preview1
+        if actor_config.wasm_target == WasmTarget::WasiPreview2 {
```
connorsmith256, line 155: Do we not want to try to adapt actors built with a custom command? If not, I think we should call this out in the docs on `build_command`

crates/wash-lib/tests/parser/main.rs
```diff
@@ -49,6 +49,7 @@ fn rust_actor() {
                 .canonicalize()
                 .unwrap(),
             wasm_bin_name: None,
+            build_command: None,
```
connorsmith256, line 52: Can we add a test case where `build_command` is set?

connorsmith256 (changes requested):

crates/wash-cli/src/build.rs
```diff
@@ -68,15 +72,29 @@ pub async fn handle_command(command: BuildCommand) -> Result<CommandOutput> {
                 })
             };
 
-            let actor_path = build_project(&config, sign_config)?;
+            let actor_path = if command.sign_only {
+                std::env::set_current_dir(&config.common.path)?;
+                let signed_path = sign_actor_wasm(
+                    &config.common,
+                    actor_config,
+                    sign_config.context("cannot supply --build-only and --sign-only")?,
```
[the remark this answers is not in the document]
brooksmtownsend: This definitely shouldn't be possible with the conflicts. I just did it because the unwrap here feels wrong, e.g. may as well save face here with a nice error, but I can just leave a SAFETY comment if that's preferred 😄

crates/wash-lib/src/build.rs
brooksmtownsend: But but efficiency!

Settle for a middleground of a clarifying comment?

crates/wash-lib/src/build.rs
brooksmtownsend: Good call. Basically, the thought is if you're adding a custom build command you're telling wash that you know what you're doing and not to interfere.

brooksmtownsend: For context, one such application of this is using `cargo component build --release` to build a straight up wasip2 component, and in that case no adapting is needed

brooksmtownsend:
@connorsmith256 I changed just a couple things in my last force push 😄 Turns out we never used the `filename` config anywhere so I removed it in favor of `destination`, and added in `build_artifact` in addition to moving `build_command` under actor config.

TL;DR check out `crates/wash-lib/tests/parser/files/rust_actor_custom_build.toml` for an actual reasonable way to encapsulate a custom build command into a wasmcloud.toml file

crates/wash-lib/src/build.rs
connorsmith256: I feel firmly that the readability is harmed, and the cost to allocate a vec of a few bytes in a CLI command is negligible, so a comment would make me think "change the source then?"

But that's just like, my opinion, man. And I respect your opinion, so if you disagree, I won't be upset 🤝

crates/wash-cli/README.md
```diff
@@ -94,49 +94,56 @@ nix run github:wasmCloud/wash
...
+| name          | string |                               | Name of the project                                                                    |
 | version       | string |                               | Semantic version of the project                                                        |
-| path          | string | `{pwd}`                       | Path to the project directory to determine where built and signed artifacts are output | 
-| wasm_bin_name | string | "name" setting                | Expected name of the wasm module binary that will be generated                         |
+| path          | string | `{pwd}`                       | Path to the project directory to determine where built and signed artifacts are output |
 | language      | enum   | [rust, tinygo]                | Language that actor or provider is written in                                          |
 | type          | enum   | [actor, provider, interface ] | Type of wasmcloud artifact that is being generated                                     |
-
+| wasm_bin_name | string | "name" setting                | Expected name of the wasm module binary that will be generated                         |
```
connorsmith256, line 109: Just a note, since I'm guessing refactoring would be non-trivial, but your comment about moving the build command to actor config made me ask: shouldn't this also be under actor config? I.e. it's not relevant for building providers (until we have WASI providers)

crates/wash-lib/tests/parser/files/rust_actor_custom_build.toml
```diff
@@ -0,0 +1,9 @@
+language = "rust"
+type = "actor"
+name = "custombuildactorcomponent"
+
+[actor]
+claims = ["wasmcloud:httpserver"]
+build_artifact = "target/wasm32-wasi/release/custombuildactorcomponent.wasm"
+build_command = "cargo component build --release --target wasm32-wasi"
```
connorsmith256, line 8: Do we have a test case that confirms this gets parsed as expected?

crates/wash-cli/README.md
brooksmtownsend: Yeah this was a really weird one. It's strange because this isn't used in the project config most of the time, usually it's used in the context of a Rust project that we infer the binary name from the Cargo.toml `[lib]` section. Let me see if I could easily refactor it

crates/wash-lib/tests/parser/files/rust_actor_custom_build.toml
brooksmtownsend: ugh FINE

brooksmtownsend: Did it 😄

connorsmith256 (approved):

crates/wash-lib/src/build.rs
```diff
@@ -96,19 +98,57 @@ pub fn build_actor(
...
+                } else {
+                    common_config.path.join(p)
+                }
+            })
+            .unwrap_or_else(|| {
+                common_config
+                    .path
+                    .join(format!("build/{}.wasm", common_config.wasm_bin_name()))
+            });
+        if let Err(_e) = std::fs::metadata(actor_path.as_path()) {
```
vados-cosmonic, line 131:
```suggestion
        if std::fs::metadata(actor_path.as_path()).is_err() {
```

What do you think about this?

crates/wash-lib/src/build.rs
```diff
@@ -96,19 +98,57 @@ pub fn build_actor(
...
+    } else {
+        // Build actor based on language toolchain
+        let actor_wasm_path = match language_config {
+            LanguageConfig::Rust(rust_config) => {
+                build_rust_actor(common_config, rust_config, actor_config)
+            }
+            LanguageConfig::TinyGo(tinygo_config) => {
+                build_tinygo_actor(common_config, tinygo_config, actor_config)
+            }
+        }?;
```
vados-cosmonic, line 147: how do you feel about matching on `raw_comand` *and* the `langauge_config` in a tuple here? And maybe moving the logic in the custom stuff (`(Some(cmd), _) => ...`) to another function? 

We way we could keep this pretty easy-to-understand section here as a `match` statement?

thomastaylor312:
One thing about the `--sign-only` flag: Would it be better user experience for `wash sign` to be `wasmcloud.toml` aware? It feels like that would be more discoverable and ergonomic from my POV

crates/wash-lib/src/build.rs
brooksmtownsend: Since I wanted to skip the embedding this ended up making it a bit more complicated, but I did refactor a bit into a function instead of keeping this all inline

crates/wash-cli/README.md
```diff
@@ -133,23 +133,28 @@ There are three main sections of a `wasmcloud.toml` file: common config, languag
...
+
+| Setting        | Type    | Default                                   | Description                                                                                                                                                                                                                                                                                                                   |
+| -------------- | ------- | ----------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| claims         | list    | []                                        | The list of provider claims that this actor requires. eg. ["wasmcloud:httpserver", "wasmcloud:blobstore"]                                                                                                                                                                                                                     |
```
vados-cosmonic, line 139:
```suggestion
| claims         | list    | []                                        | The list of provider claims that this actor requires. eg. `["wasmcloud:httpserver", "wasmcloud:blobstore"]`                                                                                                                                                                                                                     |
```

vados-cosmonic (approved):
LGTM :rocket:

PR #951, opened 2023-11-07
ci: setup crate publish pipeline

rvolosatovs:
## Feature or Problem

- Add (almost) all crates in the workspace to publish pipeline
- Dry-run publish on all non-tag builds (on tags, we just publish the crate we need)
- Only publish non-RC tags

## Related Issues

Closes https://github.com/wasmCloud/wasmCloud/issues/641

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -7,16 +7,38 @@ on:
     branches:
     - main
     tags:
+    - 'actor-v[0-9].[0-9]+.[0-9]+'
+    - 'actor-v[0-9].[0-9]+.[0-9]+-*'
+    - 'compat-v[0-9].[0-9]+.[0-9]+'
+    - 'compat-v[0-9].[0-9]+.[0-9]+-*'
     - 'control-interface-v[0-9].[0-9]+.[0-9]+'
     - 'control-interface-v[0-9].[0-9]+.[0-9]+-*'
+    - 'core-v[0-9].[0-9]+.[0-9]+'
+    - 'core-v[0-9].[0-9]+.[0-9]+-*'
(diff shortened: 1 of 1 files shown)


.github/workflows/wasmcloud.yml
```diff
@@ -7,16 +7,38 @@ on:
     branches:
     - main
     tags:
+    - 'actor-v[0-9].[0-9]+.[0-9]+'
```
connorsmith256, line 10: Since there's a long list here now, what if we narrowed down to host and crates generally:

crates:
```
*-v[0-9].[0-9]+.[0-9]+
*-v[0-9].[0-9]+.[0-9]+-*
```

host:
```
v[0-9].[0-9]+.[0-9]+
v[0-9].[0-9]+.[0-9]+-*
```

Do we get anything out of enumerating?

connorsmith256 (commented):
LGTM, just didn't hit approve until my question is resolved

.github/workflows/wasmcloud.yml
rvolosatovs: I just like to be explicit to ensure we're not accidentally hitting some case we did not consider, but this is probably fine - I'll apply this change

rvolosatovs: No, I will not be changing this, because this workflow should not run on e.g. chart tags or UI tags:
- https://github.com/wasmCloud/wasmCloud/blob/7c8ee2af3ce17b5515c673358d82634639286796/.github/workflows/wasmcloud-chart.yml#L9
- https://github.com/wasmCloud/wasmCloud/blob/7c8ee2af3ce17b5515c673358d82634639286796/.github/workflows/washboard.yml#L10

connorsmith256 (approved):

PR #953, opened 2023-11-07
ci: publish `wash` OCI images

rvolosatovs:
## Feature or Problem

- Integrate `wash` into existing OCI release pipeline
- Publish `canary` tags for wasmCloud (for consistency with `wash`)
- Add labels to wasmCloud image (for consistency with `wash`)
- Unlike before, OCI image version label value for `wash` is now taken directly from `Cargo.toml` independent from the git tag
- Remove `wash` binary from wasmCloud OCI image
- Test Debian-based OCI images after build
- Always build Debian-based OCI images
- Always run `docker run` with `--rm`
- Do not override `$PATH` in images
- Fix prerelease check for crate publish
- Fix `wash` release tag match

### Manual Verification

https://github.com/rvolosatovs/wasmCloud/actions/runs/6789376691
https://github.com/rvolosatovs/wasmCloud/actions/runs/6789374979
https://github.com/rvolosatovs/wasmCloud/actions/runs/6789373032
https://github.com/rvolosatovs/wasmCloud/actions/runs/6789371434

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -265,9 +265,16 @@ jobs:
     - name: Extract tag context
       id: ctx
       run: |
-        echo 'owner=${GITHUB_REPOSITORY_OWNER,,}' >> $GITHUB_OUTPUT
-        echo 'sha_short=$(git rev-parse --short HEAD)' >> $GITHUB_OUTPUT
-        echo 'version=${GITHUB_REF_NAME#${{ matrix.prefix }}v}' >> $GITHUB_OUTPUT
+        echo "owner=${GITHUB_REPOSITORY_OWNER,,}" >> "$GITHUB_OUTPUT"
+        echo "sha_short=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
+
+        version=${GITHUB_REF_NAME#${{ matrix.prefix }}v}
+        echo "version=${version}" >> "$GITHUB_OUTPUT"
+        echo "version is ${version}"
+        if [[ $version == *"-"* ]]; then
+          echo "version ${version} is a pre-release"
+          echo "prerelease=true" >> "$GITHUB_OUTPUT"
+        fi
 
     - name: Login to AzureCR
       if: startswith(github.ref, format('refs/tags/{0}v', matrix.prefix)) || github.ref == 'refs/heads/main'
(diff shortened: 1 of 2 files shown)


.github/workflows/wasmcloud.yml
```diff
@@ -242,159 +242,123 @@ jobs:
...
+        echo "version is ${version}"
+        if [[ $version == *"-"* ]]; then
+          echo "version ${version} is a pre-release"
+          echo "prerelease=true" >> "$GITHUB_OUTPUT"
+        fi
 
     - name: Login to AzureCR
-      if: startswith(github.ref, 'refs/tags/v') || startswith(github.ref, 'refs/tags/wash-cli-v')
+      if: startswith(github.ref, format('refs/tags/{0}v', matrix.prefix)) || github.ref == 'refs/heads/main'
+      continue-on-error: ${{ github.repository_owner != 'wasmCloud' }}
```
connorsmith256, line 281: Is this to handle forks? I'm not sure I understand

.github/workflows/wasmcloud.yml
```diff
@@ -566,13 +533,24 @@ jobs:
...
     - name: dry-run publish ${{ matrix.crate }} to crates.io
-      run: cargo publish --dry-run
-      working-directory: ./crates/${{ matrix.crate }}
       if: ${{ !startsWith(github.ref, 'refs/tags/') }}
       continue-on-error: ${{ matrix.workspace-dependencies }} # publish may fail due to workspace crates not being published yet
+      run: cargo publish --dry-run
+      working-directory: ./crates/${{ matrix.crate }}
 
     - name: publish ${{ matrix.crate }} to crates.io
+      if: startsWith(github.ref, format('refs/tags/{0}-v', matrix.crate)) && !steps.ctx.outputs.prerelease
```
connorsmith256, line 553: Just noting: I was thrown off at first by the inclusion of the `-` here, since the OCI job includes the `-` in `matrix.prefix`. I see why we can't do that here, I just hope we don't make silly errors between the two in the future 😕

flake.nix
```diff
@@ -233,31 +241,95 @@
                 mv $out/usr/etc $out/etc
               '';
             };
+
+            version =
+              if name == "wasmcloud"
+              then (readTOML ./Cargo.toml).package.version
```
connorsmith256, line 247: Help me understand why we read from the Cargo.toml instead of using the git ref?

connorsmith256 (commented):
LGTM, just a couple questions to help me understand

.github/workflows/wasmcloud.yml
rvolosatovs: Yes, see commit https://github.com/wasmCloud/wasmCloud/pull/953/commits/e579676fc40e018fb417caeadf831771ae48fdcd

flake.nix
rvolosatovs: The version in `Cargo.toml` must match the version of the binary and it's the simplest approach

.github/workflows/wasmcloud.yml
rvolosatovs: well, ideally we should never really have to change this

flake.nix
rvolosatovs: For reference, here's the spec on these https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys, which is very loosely defined (which is generally true for all of these labels and they really should not be used as an API of any sorts)

>    The version MAY match a label or tag in the source code repository
>    version MAY be Semantic versioning-compatible

.github/workflows/wasmcloud.yml
connorsmith256: I'm tempted to save this comment 😉

connorsmith256 (approved):

Issue #864, opened 2023-10-11
[BUG] `wash app` commands do not respect wash context


connorsmith256:
**Describe the bug**
`wash app` commands do not use the lattice prefix from the current wash context

**To Reproduce**
1. `wash up -d -x foobar`
2. `wash ctx new foobar`
2a. `wash ctx edit --editor code foobar` (update lattice_prefix to `foobar`)
1. `wash app deploy ~/Downloads/hello.yaml`
1. _wait a few seconds for wadm to reconcile_
2. `wash get inventory`
3. _notice the host is empty_
4. `wash app deploy -x foobar ~/Downloads/hello.yaml`
5. _wait a few seconds for wadm to reconcile_
6. `wash get inventory`
7. _notice the host is now running things_

**Expected behavior**
When a context is set, `wash app` commands should respect that context

**Steps to Fix**
This one should be pretty easy to fix. Note for most commands, we get a [WashConnectionOptions](https://github.com/wasmCloud/wash/blob/44b56b22bf67aeb164a713eb50242ca8477a3eb7/crates/wash-lib/src/cli/get.rs#L59), which takes the context into account. For app commands, however, we're explicitly [using the lattice prefix on the command itself](https://github.com/wasmCloud/wash/blob/44b56b22bf67aeb164a713eb50242ca8477a3eb7/src/app/mod.rs#L179), which is ignoring the context. We should use the prefix from the `WashConnectionOptions` for all the app commands

ahmedtadde:
hey @connorsmith256 ,

so, do we want to ignore/remove the `lattice_prefix` cmd option altogether? or do we want to do the following:
- if `lattice_prefix` cmd option is provided, use it and ignore `WashConnectionOptions.lattice_prefix`
- else, use `WashConnectionOptions.lattice_prefix` as the default

connorsmith256:
@ahmedtadde I think the priority order you described is right 👍

connorsmith256:
Resolved by #954

PR #954, opened 2023-11-07
[BUG] wash app commands do not respect wash context

ahmedtadde:
## Feature or Problem
When a context is set, `wash app` commands should respect that context

## Related Issues
#864 

## Release Information
`wash-cli v0.22.0`

## Consumer Impact
improved correctness and DX for `wash app`  commands.

### Unit Test(s)
[test_lattice_prefix_and_nats_client_from_cmd_opts](https://github.com/ahmedtadde/wasmCloud/blob/ba31f26c62c50ac5f202282361949bf7ae4187ed/crates/wash-cli/src/app/mod.rs#L339)

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -176,10 +176,8 @@ pub async fn handle_command(
 }
 
 async fn undeploy_model(cmd: UndeployCommand) -> Result<DeployModelResponse> {
-    let lattice_prefix = cmd.opts.lattice_prefix.clone();
-    let client = <CliConnectionOpts as TryInto<WashConnectionOptions>>::try_into(cmd.opts)?
-        .into_nats_client()
-        .await?;
+    let (lattice_prefix, client) =
+        get_lattice_prefix_and_nats_client_from_cmd_opts(cmd.opts).await?;
 
     wash_lib::app::undeploy_model(
         &client,
@@ -191,10 +189,8 @@ async fn undeploy_model(cmd: UndeployCommand) -> Result<DeployModelResponse> {
 }
 
 async fn deploy_model(cmd: DeployCommand) -> Result<DeployModelResponse> {
-    let lattice_prefix = cmd.opts.lattice_prefix.clone();
-    let client = <CliConnectionOpts as TryInto<WashConnectionOptions>>::try_into(cmd.opts)?
-        .into_nats_client()
-        .await?;
+    let (lattice_prefix, client) =
+        get_lattice_prefix_and_nats_client_from_cmd_opts(cmd.opts).await?;
 
     let app_manifest = match cmd.application {
         Some(source) => load_app_manifest(source.parse()?).await?,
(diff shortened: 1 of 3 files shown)


crates/wash-cli/src/app/mod.rs
```diff
@@ -328,3 +323,68 @@ fn show_model_history(results: VersionResponse) -> CommandOutput {
...
+    use anyhow::Result;
+    use std::env;
+    use wash_lib::{
+        cli::CliConnectionOpts,
+        config::{DEFAULT_CTX_DIR_NAME, DEFAULT_LATTICE_PREFIX, WASH_DIR},
+        context::{fs::ContextDir, ContextManager, WashContext},
+    };
+
+    #[tokio::test]
+    async fn test_lattice_prefix_and_nats_client_from_cmd_opts() -> Result<()> {
```
connorsmith256, line 339: This is a great test to have. Since it's not specific to the `app` commands, would you mind moving it to `wash-lib/src/cli/mod.rs`, where we define `impl TryFrom<CliConnectionOpts> for WashConnectionOptions` ?

crates/wash-lib/src/cli/mod.rs
```diff
@@ -240,9 +240,12 @@ impl TryFrom<CliConnectionOpts> for WashConnectionOptions {
             ctl_seed,
             ctl_credsfile,
             js_domain,
-            lattice_prefix,
-            timeout_ms,
+            lattice_prefix: match lattice_prefix {
```
connorsmith256, line 243: I don't think we should override just `lattice_prefix` and not the other fields which can be set via context (e.g. `ctl_host`). Since the current struct has `ctx` as a separate field, let's leave that as-is. Instead, I would suggest we define a `get_lattice_prefix` on `WashConnectionOptions`, which could look something like:

```rust
pub fn get_lattice_prefix(&self) -> String {
    self.lattice_prefix.clone().unwrap_or_else(|| {
        self.ctx
            .as_ref()
            .map(|c| c.lattice_prefix.clone())
            .unwrap_or(DEFAULT_LATTICE_PREFIX.to_string())
    })
}
```

connorsmith256 (changes requested):
Looks great, just requesting a couple small refactors 🙏

crates/wash-lib/src/cli/mod.rs
ahmedtadde: i made some [changes ](https://github.com/wasmCloud/wasmCloud/pull/954/commits/f4807e8fe4b8be6099259bdb89678b60245e52ff) based on your feedback. let me know if it works.

ahmedtadde: actually nvm... need to refactor to get tests passing

connorsmith256:
@ahmedtadde the latest changes look great. It looks like part of your new test is failing, but only on Windows. I suspect `set_default_context` might not be working on Windows(?) If you add `#[cfg(not(target_family = "windows"))]` above the new test and it passes, I think we should merge this and file a separate issue to investigate wash contexts on Windows

connorsmith256 (approved):

connorsmith256:
Filed the suspected windows bug: https://github.com/wasmCloud/wasmCloud/issues/966

PR #963, opened 2023-11-08
feat(bindgen): support witified maps

vados-cosmonic:
## Feature or Problem

WIT can only represent maps as `list<tuple<K,V>>`, which we refer to as "WIT-ified maps" in the codebase.

Up until now, WIT-ified maps were not explicitly supported by provider-wit-bindgen. This means that these structures were passed through *as-is*, expecting that the code on the other end also expects to deal with WIT-ified maps (accepting a list, instead of a proper map).

Legacy (smithy) interfaces use *real* maps (simplest of which being a HashMap), and as such the legacy interfaces (like http-client) can't use code that returns list parading as a map.

This commit adds support for WIT-ified maps, converting them during the bindgen process to `std::collections::HashMap`s

### DIFF
--- a/crates/provider-wit-bindgen/src/lib.rs
+++ b/crates/provider-wit-bindgen/src/lib.rs
@@ -26,17 +26,9 @@
 //! For more information on the options available to underlying bindgen, see the [wasmtime-component-bindgen documentation](https://docs.rs/wasmtime/latest/wasmtime/component/macro.bindgen.html).
 //!
 
-mod vendor;
-
-use anyhow::{bail, ensure, Context};
-use tracing::{debug, warn};
-use tracing_subscriber::EnvFilter;
-use vendor::wasmtime_component_macro::bindgen::{
-    expand as expand_wasmtime_component, Config as WitBindgenConfig,
-};
-
 use std::{collections::HashMap, str::FromStr};
 
+use anyhow::{bail, ensure, Context};
 use heck::{ToSnakeCase, ToUpperCamelCase};
 use proc_macro2::{Ident, Punct, Spacing, Span, TokenStream, TokenTree};
 use quote::{format_ident, ToTokens, TokenStreamExt};
@@ -48,6 +40,13 @@ use syn::{
     visit_mut::{visit_item_mut, VisitMut},
     FnArg, Item, ItemMod, ItemStruct, ItemType, LitStr, PathSegment, ReturnType, Token, TraitItem,
     TraitItemFn, Type,
+};
+use tracing::{debug, trace, warn};
+use tracing_subscriber::EnvFilter;
+
+mod vendor;
+use vendor::wasmtime_component_macro::bindgen::{
+    expand as expand_wasmtime_component, Config as WitBindgenConfig,
 };
 
 /// Rust module name that is used by wit-bindgen to generate all the modules
@@ -94,13 +93,22 @@ struct ProviderBindgenConfig {
     pub(crate) exposed_interface_deny_list: Vec<LatticeExposedInterface>,
 
     /// wit-bindgen configuration that is passed straight through (uses vendored wit-bindgen)
-    pub(crate) wit_bindgen_cfg: WitBindgenConfig,
+    ///
+    /// During an actual parse run, configuration to pass through to wit bindgen *must* be provided
+    /// this means that full parse runs can only be run when you can set up the filesystem as necessary.
+    ///
+    /// This allows local tests to easily generate `ProviderBindgenConfig` objects, without building the
+    /// file tree (and related WIT) that wit-bindgen would attempt to read
+    pub(crate) wit_bindgen_cfg: Option<WitBindgenConfig>,
 
     /// Translation strategy for functions on imported interfaces when generating lattice data structures and code
     pub(crate) import_fn_lattice_translation_strategy: WitFunctionLatticeTranslationStrategy,
 
     /// Translation strategy for functions on exported interfaces when generating lattice data structures and code
     pub(crate) export_fn_lattice_translation_strategy: WitFunctionLatticeTranslationStrategy,
+
+    /// Whether to replace WIT-ified maps (`list<tuple<T, T>>`) with a Map type (`std::collections::HashMap`)
+    pub(crate) replace_witified_maps: bool,
 }
 
 /// Keywords that are used by this macro
@@ -114,6 +122,7 @@ mod keywords {
     syn::custom_keyword!(export_fn_lattice_translation_strategy);
     syn::custom_keyword!(exposed_interface_allow_list);
     syn::custom_keyword!(exposed_interface_deny_list);
+    syn::custom_keyword!(replace_witified_maps);
 }
 
 /// Wrapper for a list of qualified WIT function names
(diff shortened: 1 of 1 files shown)


crates/provider-wit-bindgen/src/lib.rs
```diff
@@ -94,13 +93,16 @@ struct ProviderBindgenConfig {
...
+    pub(crate) wit_bindgen_cfg: Option<WitBindgenConfig>,
 
     /// Translation strategy for functions on imported interfaces when generating lattice data structures and code
     pub(crate) import_fn_lattice_translation_strategy: WitFunctionLatticeTranslationStrategy,
 
     /// Translation strategy for functions on exported interfaces when generating lattice data structures and code
     pub(crate) export_fn_lattice_translation_strategy: WitFunctionLatticeTranslationStrategy,
+
+    /// Whether to replace WIT-ified maps (`list<tuple<T, T>>`) with a Map type (`std::collections::HashMap`)
+    pub(crate) replace_witified_maps: Option<bool>,
```
connorsmith256, line 105: Why `Option` instead of just a bool?

crates/provider-wit-bindgen/src/lib.rs
```diff
@@ -94,13 +93,16 @@ struct ProviderBindgenConfig {
     pub(crate) exposed_interface_deny_list: Vec<LatticeExposedInterface>,
 
     /// wit-bindgen configuration that is passed straight through (uses vendored wit-bindgen)
-    pub(crate) wit_bindgen_cfg: WitBindgenConfig,
+    pub(crate) wit_bindgen_cfg: Option<WitBindgenConfig>,
```
connorsmith256, line 96: Suggestion: I think the comment [here](https://github.com/wasmCloud/wasmCloud/pull/963/files#diff-c35e6db0016f48f0c6fe14869e51b1fbffd9c6d8db1be7933222b5d519e6045aR922) would actually be better if it were here (assuming I've understood correctly)

crates/provider-wit-bindgen/src/lib.rs
```diff
@@ -1749,3 +1812,174 @@ fn convert_wit_type(t: &wit_parser::Type, cfg: &ProviderBindgenConfig) -> anyhow
...
 }
+
+/// Attempt to extract key and value types from a tree of tokens that is a witified map
+///
+/// For example, the following Rust type, submitted as a list of tokens would be parsed successfully:
+///
+/// ```rust,ignore
+/// Vec<(String, String)>
+/// ```
+fn extract_witified_map_types(input: &[TokenTree]) -> Option<(TokenStream, TokenStream)> {
```
connorsmith256, line 1823: I'm leaning toward thinking `Result` would be more clear than `Option` here. Thoughts?

I'm trying to determine if we only call this where we expect a map type to be, and therefore if we don't find one, it's an "error", vs if we optimistically call this and are okay if the following tokens don't match

crates/provider-wit-bindgen/src/lib.rs
```diff
@@ -1749,3 +1812,174 @@ fn convert_wit_type(t: &wit_parser::Type, cfg: &ProviderBindgenConfig) -> anyhow
...
+    match input {
+        [
+            TokenTree::Ident(vec_ty),
+            TokenTree::Punct(p1), // <
+            TokenTree::Group(g),
+            TokenTree::Punct(p2), // >
+        ] if *vec_ty == "Vec" && p1.to_string() == "<" && p2.to_string() == ">" => {
```
connorsmith256, line 1830: Is it possible to have the tuple types themselves contain `<`, e.g. `Vec<(GenT<Foo>, GenT<Bar>)>`?

connorsmith256 (approved):
LGTM. Left a few comments for consideration

crates/provider-wit-bindgen/src/lib.rs
vados-cosmonic: Sure, will move this!

- [ ] move docs on `wit_bindgen_cfg`

crates/provider-wit-bindgen/src/lib.rs
vados-cosmonic: This can be a bool -- will update

- [ ] change `replace_witified_maps` to a bool

crates/provider-wit-bindgen/src/lib.rs
vados-cosmonic: Yeah it's more that I optimistically call this if the tokens don't match I want to use the existing code/fallback, so it's not quite an "error". This is written to work at places were there *could* be a map type:

- in exported WIT functions (as the first argument or otherwise)
- in types of members of WIT-defined structs

Most of the time we'll have a clue (e.x. `process_fn_arg`) with the given naming of the argument, but that's not always the case (ex. processing structs from bindgen, someone could h).

As far as the type goes, it could be a `Result`, but an `Option` in the case of extraction failure fit a little better to me -- I'm open to changing it

crates/provider-wit-bindgen/src/lib.rs
vados-cosmonic: Yeah it's possible -- the Group here should capture both of those.

That said, the comma-finding index right below works on simple cases of that, but it *could* get mixed up with a `Vec(GenT<Foo, Bar>, GenT<Baz, Bazr>)` -- *if* they don't show up as a `TokenTree::Group` (i.e. if there is more than one "top level" comma)

crates/provider-wit-bindgen/src/lib.rs
connorsmith256: Your reasoning makes sense to me 👍 Thanks for the explanation

crates/provider-wit-bindgen/src/lib.rs
connorsmith256: Ahh, I was missing that `Group` would handle more complicated structures

connorsmith256 (approved):

crates/provider-wit-bindgen/src/lib.rs
vados-cosmonic: Oh just a word on this, it's entirely possible that I could get some tokens that *don't* behave like that... But I figure we can cross that bridge when we get there. I think it *is* possible that something comes in with some other combination of syntax possibilities that makes a bad comma show up.

connorsmith256 (approved):

crates/provider-wit-bindgen/src/lib.rs
```diff
@@ -782,6 +803,8 @@ Please ensure that types for exported interface functions (normally a single rec
         _iface_fn: &wit_parser::Function,
         _cfg: &ProviderBindgenConfig,
     ) -> anyhow::Result<TokenStream> {
+        // TODO: add bundling of exported args to single struct
+        // TODO: handle WIT-map -> Map translation for arguments in iface
```
thomastaylor312, line 807: Is this going to be a follow up? If yes, do we need to file an issue before merging?

thomastaylor312 (approved):
This is good to go, just one minor question

crates/provider-wit-bindgen/src/lib.rs
vados-cosmonic: Yup this is a follow up -- just made [an issue](https://github.com/wasmCloud/wasmCloud/issues/999), so we can use that to track it

Issue #849, opened 2023-06-26
[FEATURE] `wash up --nats-connect-only` respect context


jordan-rash:
If I have a context selected and run `wash up --nats-connect-only`, it would be nice if the host would start with the settings inside the context.  As it sits now, I have to manually set the environment or create a separate host config.

PR #971, opened 2023-11-09
Refactor/context polish

connorsmith256:
## Feature or Problem
This was meant to be a small change to respect the current context during `wash up`, but it didn't turn out to be simple. I refactored the context logic quite a bit:
- always initialize the context directory before trying to perform any actions related to contexts
- create a `default` file in the context directory to contain the name of the default context, rather than an `index.json` file, which is not an index, and only ever contained a single name
  - include backwards-compatibility code to gracefully migrate defaults from the old to the new system
- **BREAKING CHANGE:** update the `context` options on `call` and all the ctl commands to take the name of a context, rather than a path. This isn't strictly necessary, but I suspect no users have been using this option yet
- and finally, update `wash up` to take the current context into account
  - this revealed a bug where wash was no longer passing the `PROV_RPC` vars, but the host still needs them, if the RPC traffic is not on the default `127.0.0.1:4222`

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/849

## Release Information
v0.23.0, since this is a breaking change

### Unit Test(s)
See several updated tests

### Manual Verification
Lots of `wash ctx` and `wash up` testing


crates/wash-cli/src/call.rs
```diff
@@ -231,14 +227,12 @@ async fn rpc_client_from_opts(
...
-        let ctx_dir = ContextDir::new(ctx_dir)?;
-        ensure_host_config_context(&ctx_dir)?;
-        Some(ctx_dir.load_default_context()?)
+    // Attempt to load a context, falling back on the default if not supplied
+    let ctx_dir = ContextDir::new()?;
+    let ctx = if let Some(context_name) = opts.context {
+        Some(ctx_dir.load_context(&context_name)?)
     } else {
-        None
+        ctx_dir.load_default_context().ok()
```
vados-cosmonic, line 235: NIT: `ctx_dir.load_context` and `load_default_context` have the same `Result<WashContext>` -- do we need to wrap in `Some` if we know it will always be there? seems like a bunch of code after this could be simplified as well if we know we always load either the given context or the default one.

Also, it would probably be nice to add a `.context("failed to load context @ [...]")` or similar for failures here.

This isn't critical though, biased for :racehorse: here.

crates/wash-cli/src/completions.rs
```diff
@@ -45,18 +44,15 @@ pub enum ShellSelection {
...
     }
-    let _ = std::fs::File::create(token).map_err(|e| {
-        anyhow!(
-            "can't create completion first-run token in {}: {}",
-            // unwrap() ok because cfg_dir worked above
-            cfg_dir().unwrap().display(),
-            e
-        )
-    })?;
+    let _ = std::fs::File::create(token).context(format!(
```
vados-cosmonic, line 52:
```suggestion
    let _ = std::fs::File::create(token).with_context(|| format!(
```

crates/wash-cli/src/up/config.rs
```diff
@@ -92,27 +91,28 @@ pub async fn configure_host_env(
 
     // NATS RPC connection configuration
     if let Some(host) = wasmcloud_opts.rpc_host {
-        host_config.insert(WASMCLOUD_RPC_HOST.to_string(), host);
-    } else {
-        host_config.insert(WASMCLOUD_RPC_HOST.to_string(), nats_opts.nats_host.clone());
+        host_config.insert(WASMCLOUD_RPC_HOST.to_string(), host.clone());
+        // TODO: remove me after the host removes support for PROV_RPC connections
```
vados-cosmonic, line 95: This note and the other similar ones are probably worth adding to https://github.com/wasmCloud/wasmCloud/pull/774 so we don't forget?

crates/wash-cli/src/up/mod.rs
```diff
@@ -607,15 +644,14 @@ async fn stop_wasmcloud(mut wasmcloud_child: Child) -> Result<()> {
...
-    )
-    .await?;
+async fn is_wadm_running(
+    nats_host: &str,
+    nats_port: u16,
+    credsfile: Option<PathBuf>,
+    lattice_prefix: &str,
+) -> Result<bool> {
+    let client =
+        nats_client_from_opts(nats_host, &nats_port.to_string(), None, None, credsfile).await?;
```
vados-cosmonic, line 654: This feels like it could be moved to `wash-lib`, what do you think?

crates/wash-lib/src/cli/mod.rs
```diff
@@ -223,14 +219,11 @@ impl TryFrom<CliConnectionOpts> for WashConnectionOptions {
...
-        } else if let Ok(ctx_dir) = context_dir(None) {
-            let ctx_dir = ContextDir::new(ctx_dir)?;
-            ensure_host_config_context(&ctx_dir)?;
-            Some(ctx_dir.load_default_context()?)
+        let ctx_dir = ContextDir::new()?;
+        let ctx = if let Some(context_name) = context {
+            Some(ctx_dir.load_context(&context_name)?)
         } else {
-            None
+            ctx_dir.load_default_context().ok()
```
vados-cosmonic, line 226: NIT: `ContextDir::load_name_or_default` ? `ContextDir::load_named`? `ContextDir::load_by_name` ?

crates/wash-lib/src/config.rs
```diff
@@ -24,41 +20,26 @@ pub const DEFAULT_START_PROVIDER_TIMEOUT_MS: u64 = 60_000;
...
-        )
-    })?;
+pub fn cfg_dir() -> Result<PathBuf> {
+    let home = dirs::home_dir().context("no home directory found. Please set $HOME")?;
 
     let wash = home.join(WASH_DIR);
 
     if !wash.exists() {
-        fs::create_dir_all(&wash)?;
+        fs::create_dir_all(&wash).context("failed to create .wash directory")?;
```
vados-cosmonic, line 29: How do you feel about noting the path that it failed to create? `with_context` and then using the path would help make it much clearer the file that failed to create (which might explain why it failed)

crates/wash-lib/src/context/fs.rs
```diff
@@ -29,42 +31,64 @@ impl Deref for ContextDir {
...
-        let p = path.as_ref();
-        let exists = p.exists();
-        if exists && !p.is_dir() {
-            anyhow::bail!("{} is not a directory (or cannot be accessed)", p.display())
+    /// Creates a new ContextDir at ~/.wash/contexts
+    pub fn new() -> Result<ContextDir> {
+        Self::new_with_dir(None::<&Path>)
+    }
+
+    /// Creates a new ContextDir at the specified path. If a path is not provided, defaults to ~/.wash/contexts
```
vados-cosmonic, line 39:
```suggestion
    /// Creates and initializes a new [`ContextDir`] at the specified path. If a path is not provided, defaults to ~/.wash/contexts
```

crates/wash-lib/src/context/fs.rs
```diff
@@ -29,42 +31,64 @@ impl Deref for ContextDir {
...
+    pub fn new_with_dir(path: Option<impl AsRef<Path>>) -> Result<ContextDir> {
```
vados-cosmonic, line 40: This is really similar to the other case of going from a context name or the default but how do you feel about:

NIT: `ContextDir::from_dir` ?

I also like the `TryFrom` approach, but this unfortunately clashes w/ named contexts and file paths since they both can be interpreted as paths
```rust
impl<T> TryFrom<Option<T>> for ContextDir where T: AsRef<std::path::Path> {
    type Error = anyhow::Error;
    fn try_from(other: Option<T>) -> anyhow::Result<ContextDir> { ... }
}
```

crates/wash-lib/src/context/fs.rs
```diff
@@ -74,67 +98,93 @@ impl ContextDir {
...
+fn default_context_dir() -> Result<PathBuf> {
+    Ok(cfg_dir()?.join(DEFAULT_CTX_DIR_NAME))
+}
 
-    fn index_path(&self) -> PathBuf {
-        self.0.join(INDEX_JSON)
+fn initialize_context_dir(context_dir: &Path, default_path: &PathBuf) -> Result<()> {
+    let mut default_context_name = HOST_CONFIG_NAME.to_string();
+
+    // TEMPORARY (TM): look for and parse existing index.json, to preserve backwards compatibility
```
vados-cosmonic, line 110: :smile:

crates/wash-lib/src/context/fs.rs
```diff
@@ -74,67 +98,93 @@ impl ContextDir {
...
+            default_context_name = old_default_context.name;
+        }
     }
+    // END TEMPORARY (TM)
+
+    let mut default_file =
+        File::create(default_path).context("failed to create default context file")?;
+    default_file
+        .write_all(default_context_name.as_bytes())
+        .context("failed to write to default context file")?;
```
vados-cosmonic, line 129: Do you like the explicitness of the errors here w/ failure to create files  and/or write to them vs `std::fs::write`?

Also, it would be great to have more details about the paths that couldn't be created/written

crates/wash-lib/src/context/fs.rs
```diff
@@ -190,24 +233,24 @@ mod test {
...
-            .path();
+            .filter_map(|entry| entry.unwrap().file_name().to_os_string().into_string().ok())
+            .collect();
+        let expected_filenames: std::collections::HashSet<String> = vec![
+            "default".to_string(),
+            "host_config.json".to_string(),
+            "happy_path.json".to_string(),
+        ]
+        .into_iter()
+        .collect();
```
vados-cosmonic, line 247:
```suggestion
        let expected_filenames = std::collections::HashSet::from([
            "default".to_string(),
            "host_config.json".to_string(),
            "happy_path.json".to_string(),
        ]);
```

vados-cosmonic (approved):
LGTM, Nothin but NITS :rocket:

crates/wash-cli/src/call.rs
connorsmith256: Nope, you're totally right. Didn't think of this during late-night refactoring 😴

crates/wash-cli/src/up/config.rs
connorsmith256: Merging #774 occupies a permanently reserved portion of my mind 😉

crates/wash-cli/src/up/mod.rs
connorsmith256: Turns out we have `create_nats_client_from_opts` in wash-lib already. I'm going to delete `nats_client_from_opts`

crates/wash-lib/src/context/fs.rs
connorsmith256: I like `from_dir` 👍

crates/wash-lib/src/cli/mod.rs
connorsmith256: I think I prefer keeping `load_context` and `load_default_context` separate

crates/wash-lib/src/context/fs.rs
connorsmith256: I tried this and was confused why it didn't work. It's because I had the `vec!` 🙂

crates/wash-lib/src/context/fs.rs
connorsmith256: To be honest, no, the only reason most of this code didn't change more is because I didn't originally intend to change any of this code 😅 Fixed

crates/wash-lib/src/context/fs.rs
vados-cosmonic: oh I meant the "(tm)" here, was funny

vados-cosmonic (approved):
LGTM :rocket:

PR #972, opened 2023-11-09
fix(docs): update helm chart install to use artifacthub

vados-cosmonic:
## Feature or Problem

While the official wasmcloud Helm chart was recently updated and published to ArtifactHub, the README was still referencing the old OTP host's Helm chart.

This commit updates the helm chart README to use recent Helm install instructions that are posted to the ArtifactHub page, and also reference that page so users can find the listing.

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -20,26 +20,37 @@ recommend deploying your main NATS server with another chart.
 
 ## Running the chart
 
-The chart is available in a hosted Helm repository. To add the repository, run the following:
+You can install the chart via [the `wasmcloud-chart` package on ArtifactHub][artifacthub-wasmcloud]:
 
 ```console
-$ helm repo add wasmcloud https://wasmcloud.github.io/wasmcloud-otp/
+$ helm install wasmcloud oci://ghcr.io/wasmcloud/wasmcloud-chart --version 0.7.0
 ```
 
-Then you can install the chart:
-
-```console
-$ helm install <RELEASE_NAME> wasmcloud/wasmcloud-host
-```
+> [!NOTE]
+>
+> You can replace `wasmcloud` with whatever you'd like to call your instantiation
+> of the helm chart.
+>
+> Version `0.7.0` of the helm chart (released Nov 9th, 2023) deploys version [`v0.80.0`][wasmcloud-v0.80.0] of the host by default.
+>
+> To change the version of the wasmcloud host that is deployed, set `wasmcloud.image.tag` in [`values.yaml`][values-yaml] to the
+> [docker image tag][wasmcloud-docker-tags] you'd like to deploy instead.
 
 This will get you up and going with a wasmCloud Host and NATS server so you can kick the tires and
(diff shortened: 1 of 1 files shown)


chart/README.md
```diff
@@ -20,17 +20,18 @@ recommend deploying your main NATS server with another chart.
 
 ## Running the chart
 
-The chart is available in a hosted Helm repository. To add the repository, run the following:
+You can install the chart via [the `wasmcloud-chart` package on ArtifactHub][artifacthub-wasmcloud]:
 
 ```console
-$ helm repo add wasmcloud https://wasmcloud.github.io/wasmcloud-otp/
+$ helm install wasmcloud oci://ghcr.io/wasmcloud/wasmcloud-chart --version 0.7.0
```
rvolosatovs, line 26: Will we have to be updating this every time we bump wasmCloud version? That seems like an awful lot of maintenance burden

chart/README.md
```diff
@@ -20,17 +20,18 @@ recommend deploying your main NATS server with another chart.
...
-
-```console
-$ helm install <RELEASE_NAME> wasmcloud/wasmcloud-host
-```
+> [!NOTE]
+>
+> You can replace `wasmcloud` with whatever you'd like to call your instantiation
+> of the helm chart.
+>
+> Version `0.7.0` of the helm chart (released Nov 9th, 2023) deploys version [`v0.80.0`][wasmcloud-v0.80.0] of the host by default.
```
rvolosatovs, line 34: same as above

rvolosatovs (commented):
Overall, I'm fine with this, but concerned about the maintenance burden of keeping this doc in sync with latest version of wasmCloud (which is something we will inevitably forget to do at one point)

I think @joonas should take a look here

chart/README.md
joonas: Not necessarily, the Chart should be independently versioned based on changes to the underlying software it deploys that necessitate changes to the Chart.

For example, if we choose to add or remove configuration options, we should update the Chart to reflect that.

However, new versions of wasmCloud being released where nothing else changes, it would be fine (imo) to keep the Chart as is, because the Chart provides a way to override the version of wasmCloud host it's running via `image.tag` option (and falling back to the Chart's appVersion if not provided): https://github.com/wasmCloud/wasmCloud/blob/main/chart/templates/deployment.yaml#L35

chart/README.md
joonas: Would it be worth mentioning that the default version (as provided by `appVersion`) can be overridden by setting `wasmcloud.image.tag` value in values.yaml?

vados-cosmonic:
Hey @joonas thanks for taking a look -- I just made that change, if you wouldn't mind looking again?

chart/README.md
vados-cosmonic: Done! mentioned the values override

connorsmith256 (approved):
LGTM, thank you @vados-cosmonic !

joonas:
> Hey @joonas thanks for taking a look -- I just made that change, if you wouldn't mind looking again?

Looks good!

PR #976, opened 2023-11-09
chore(wash-cli): remove `wasmbus_rpc` dependency

rvolosatovs:
## Feature or Problem

Replace `wasmbus_rpc` by `wasmcloud_core` and `wasmcloud_provider_sdk`

This is the last occurence of usage of `wasmbus_rpc` within the root workspace. Only providers left (e.g. #745 )

## Related Issues

#641

### DIFF
--- a/crates/wash-cli/Cargo.toml
+++ b/crates/wash-cli/Cargo.toml
@@ -59,8 +59,9 @@ warp = { workspace = true }
 warp-embed = { workspace = true }
 wascap = { workspace = true }
 wash-lib = { workspace = true, features = ["cli", "parser", "nats", "start"] }
-wasmbus-rpc = { workspace = true }
 wasmcloud-control-interface = { workspace = true }
+wasmcloud-core = { workspace = true }
+wasmcloud-provider-sdk = { workspace = true }
 wasmcloud-test-util = { workspace = true }
 weld-codegen = { workspace = true, features = ["wasmbus"] }
 which = { workspace = true }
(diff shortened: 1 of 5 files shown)


crates/wash-cli/src/util.rs
```diff
@@ -28,7 +28,7 @@ pub fn default_timeout_ms() -> u64 {
 /// Transform a json string (e.g. "{"hello": "world"}") into msgpack bytes
 pub fn json_str_to_msgpack_bytes(payload: &str) -> Result<Vec<u8>> {
     let json = serde_json::from_str::<serde_json::Value>(payload)?;
-    let payload = wasmbus_rpc::common::serialize(&json)?;
+    let payload = rmp_serde::to_vec_named(&json)?;
     Ok(payload)
```
vados-cosmonic, line 32:
```suggestion
    rmp_serde::to_vec_named(&json).context("failed to encode json value as msgpack")
```

vados-cosmonic (approved):
LGTM :rocket:

crates/wash-cli/Cargo.toml
```diff
@@ -59,8 +59,9 @@ warp = { workspace = true }
 warp-embed = { workspace = true }
 wascap = { workspace = true }
 wash-lib = { workspace = true, features = ["cli", "parser", "nats", "start"] }
-wasmbus-rpc = { workspace = true }
 wasmcloud-control-interface = { workspace = true }
+wasmcloud-core = { workspace = true }
+wasmcloud-provider-sdk = { workspace = true }
```
connorsmith256, line 64: This is okay for now, but I don't love bringing in this dependency. It looks like it's just for `wash call`, which doesn't currently work anyway, so we could also leave this out for now

rvolosatovs: I don't feel like removing functionality is in-scope for this PR, let's keep functional changes for a different PR. This is just removing the problematic and unmaintained `wasmbus-rpc`

connorsmith256 (approved):

PR #977, opened 2023-11-09
build: update `async-nats` to 0.33

rvolosatovs:
## Feature or Problem

Update `async-nats` to latest release

## Related Issues

This unblocks #724  (or at least it should)

### DIFF
--- a/crates/provider-sdk/src/provider.rs
+++ b/crates/provider-sdk/src/provider.rs
@@ -274,7 +274,7 @@ impl ProviderConnection {
                                     if let Some(reply) = msg.reply {
                                         // send reply
                                         if let Err(err) = this.rpc_client
-                                            .publish_invocation_response(reply.to_string(), resp).in_current_span().await {
+                                            .publish_invocation_response(reply, resp).in_current_span().await {
                                             error!(%err, "rpc sending response");
                                         }
                                     }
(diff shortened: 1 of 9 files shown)


vados-cosmonic (approved):
LGTM :rocket:

crates/provider-sdk/src/provider.rs
```diff
@@ -274,15 +274,15 @@ impl ProviderConnection {
                                     if let Some(reply) = msg.reply {
                                         // send reply
                                         if let Err(err) = this.rpc_client
-                                            .publish_invocation_response(reply, resp).in_current_span().await {
+                                            .publish_invocation_response(reply.to_string(), resp).in_current_span().await {
```
connorsmith256, line 277: Why were these `to_string()` needed? Isn't `reply` a `Subject` already?

rvolosatovs: `reply` is `Subject` indeed, however `Subject` is a type defined in `async_nats` and therefore it does not belong in public API of this crate

connorsmith256: I don't think `publish_invocation_response` and `publish` should be public. They're part of the SDK's internal handling of invocations. Can we use the NATS type internally and avoid republishing? Normally I wouldn't care, but the RPC internals are where performance and allocations matter most, so it feels like a step backward to add these

rvolosatovs: We're dealing with network I/O here, so this really should not matter, but sure

rvolosatovs: `publish_invocation_response` is, indeed, public, so I am against this change

connorsmith256: I know, I'm saying these _shouldn't_ be public

rvolosatovs: Sure, but that change is out-of-scope for `async-nats` update PR, right?

thomastaylor312 (approved):

PR #983, opened 2023-11-10
fix(tests): avoid saving data when starting local redis instances

vados-cosmonic:
## Feature or Problem

Up until now, Redis has generated dump files which can be loaded depending on configuration of the local redis server. The files (by default `dump.rdb` get created when tests are run, and are gitignored)

This commit avoids generating dump files by disabling save *and* randomizes the file names that are generated by embedding the port to avoid overlaps.

### DIFF
--- a/tests/wasmbus.rs
+++ b/tests/wasmbus.rs
@@ -241,10 +241,17 @@ async fn start_redis() -> anyhow::Result<(
     JoinHandle<anyhow::Result<ExitStatus>>,
     oneshot::Sender<()>,
     Url,
+    tempfile::NamedTempFile,
 )> {
     let port = free_port().await?;
     let url =
         Url::parse(&format!("redis://localhost:{port}")).context("failed to parse Redis URL")?;
+    let dump_file =
+        tempfile::NamedTempFile::new().context("failed to create temporary dump file for redis")?;
+    let dump_file_path = dump_file
+        .path()
+        .to_str()
+        .context("failed to retrieve path to created redis dump file")?;
     let (server, stop_tx) = spawn_server(
         Command::new(
             env::var("WASMCLOUD_REDIS")
@@ -260,12 +267,12 @@ async fn start_redis() -> anyhow::Result<(
             "--save",
             "",
             "--dbfilename",
-            format!("test-redis-{port}.rdb").as_str(),
+            dump_file_path,
         ]),
     )
     .await
     .context("failed to start Redis")?;
-    Ok((server, stop_tx, url))
+    Ok((server, stop_tx, url, dump_file))
 }
 
 async fn start_nats() -> anyhow::Result<(
@@ -494,8 +501,13 @@ async fn wasmbus() -> anyhow::Result<()> {
     // FIXME: we should be using separate NATS clients for CTL, RPC, and PROV_RPC
 
     let (
-        (component_redis_server, component_stop_redis_tx, component_redis_url),
-        (module_redis_server, module_stop_redis_tx, module_redis_url),
+        (
+            component_redis_server,
+            component_stop_redis_tx,
+            component_redis_url,
+            _component_redis_dump_file,
+        ),
+        (module_redis_server, module_stop_redis_tx, module_redis_url, _module_redis_dump_file),
     ) = try_join!(start_redis(), start_redis())?;
 
     let mut component_redis_client =
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@
 .vscode/
 /tests/fixtures
 /tmp/
-dump.rdb
 host_config.json
 node_modules
 target/


tests/wasmbus.rs
```diff
@@ -251,7 +251,17 @@ async fn start_redis() -> anyhow::Result<(
...
+        .args([
+            "--port",
+            &port.to_string(),
+            // Ensure that no data is saved locally, since users with
+            // redis-server installed on their machines may have default
+            // configurations which normally specify a persistence directory
+            "--save",
+            "",
+            "--dbfilename",
+            format!("test-redis-{port}.rdb").as_str(),
```
rvolosatovs, line 263: I'm pretty sure we don't need this, from official docs:
> # Note that in order to read the configuration file, Redis must be
> # started with the file path as first argument:
> #
> # ./redis-server /path/to/redis.conf

And

> # Snapshotting can be completely disabled with a single empty string argument
> # as in following example:
> #
> # save ""

From https://github.com/redis/redis/blob/cf6ed3feeb95059817d7fa8685aa04a8a33a1f92/redis.conf

So if we set `save ""`, then it won't write the snapshot and the system-level config is never loaded anyway.

Do note that from my experience, however, I've only seen `dump.rdb` files appear on crashes (e.g. ^C of a running test), but it appears that should be addressed

vados-cosmonic: Redis loads a default configuration, and in that configuration `dump.rdb` is the default name of the snapshot file, so it recognizes and loads it, if present. The documentation is confusing there so it's hard to get right.

Restated, if you have a `dump.rdb` in the folder where you run `redis-server`, it gets loaded at startup, you can try it yourself if you start `redis-server` like so:

```console
redis-server --save ""
```

To confirm this yourself you can create a `dump.rdb` to try with by starting the server, running some `SET`s (and optionally some `GET`s), then the [SAVE command](https://redis.io/commands/save/), then restart your `redis-server` with the `dump.rdb` in the same folder. 

When restarting the `redis-server` you can either use no args or use the `--save ""` args there and you'll be able to read any values you set.

Let me know if you get something different!

rvolosatovs: Right, I see what you mean.
How about we use `tempfile` crate to create a unique temporary file and set `dbfilename` to that path to avoid possibility of this happening altogether?
Let's also delete the file once we close Redis.

vados-cosmonic: I think it's unlikely that two instances of redis get the same port, and crash (or do some other operation) that causes them to write dump files that overwrite.

That said -- I'm fine with using a `NamedTempFile` to make it even *more* unlikely. 

- [ ] use `NamedTempFile` to provide redis a dump file

vados-cosmonic: Nope, actually I can't do this, because `dbfilename` needs to be a *filename*, not a path -- this won't change for now.

vados-cosmonic: @rvolosatovs note the above -- this should be ready for review as-is

rvolosatovs: I don't think this is ready for review - there is only the gitignore change in this PR currently.
I disagree that renaming the dump file really improves anything here, on the contrary, it would:
1. Needlessly populate developer's working directory with up to ~2^16 dump files (which are now also likely to get checked in)
2. Make issue harder to debug, because it now it *may* pull in the previous state, but also *may not*.

I think we should either:
1. Not do anything at all here, leave everything as-is
2. Set `save ""` and leave the rest as-is (my preferred solution)
3. Set `save ""` and configure Redis in such a way that dump is written in a unique-per-run, temporary location *outside* the repository

rvolosatovs (changes requested):
Requesting changes to prevent merge of the PR in the current state

tests/wasmbus.rs
vados-cosmonic: > I don't think this is ready for review - there is only the gitignore change in this PR currently.

It looks like the change actually is *already in* `wasmcloud/wasmcloud` -- I'm going to close this PR. 

With regards to your points, 

>  I disagree that renaming the dump file really improves anything here, on the contrary, it would:

The dump file having the same name as the default `dump.rdb` caused lost hours of debugging time, and existing code was not written to attempt to randomize this file or prevent dumps from being written at all.

That's a bug, and it's fixed with this PR -- it is both unlikely that you will *have a dump file*, and that they will collide. 

> Needlessly populate developer's working directory with up to ~2^16 dump files (which are now also likely to get checked in)

There is no world in which files named `test-redis-<...>.rdb` are not easy to notice, and easily identifiable by a developer running tests, unlike `dump.rdb`. The previous world where the file was ignored was actually *harder* to notice -- the file does not stick out in any way compared to the rest of the codebase, and it was ignored so no new untracked files show up.

If we really want the generated files to be ignored and the issue here is with the `gitignore` pattern that can be solved separately by updating the `gitignore`, but that doesn't seem to be your suggestion.

> Make issue harder to debug, because it now it may pull in the previous state, but also may not.

The probability of a collision in a space of 1/2^16 is not a realistic concern, for developers running tests which only leave behind files in a failure scenario. In any case this is much better than the 100% chance (assuming the file existed) with the previous code.

> I think we should either:

Thanks for the suggestions here! I think they could be addressed in a follow up PR -- at this point, since this seems to have already merged with no adverse effects, I'm going to resolve this and close *this* PR

Issue #978, opened 2023-11-09
[FEATURE] Improve WasmCloud Helm Chart for modern k8s security expectations


VestigeJ:
**Is your feature request related to a problem? Please describe.**
When deploying wasmcloud as a helm chart on most recent k8s versions you're met with security warnings violations.  

```
Pulled: ghcr.io/wasmcloud/wasmcloud-chart:0.7.0
Digest: sha256:e89daefd2590a024acee95bdbed660366d9b417419cc90659c2522b4164378c5
W1109 19:30:00.954074  111030 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (containers "wasmcloud-chart", "nats" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (containers "wasmcloud-chart", "nats" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or containers "wasmcloud-chart", "nats" must set securityContext.runAsNonRoot=true), seccompProfile (pod or containers "wasmcloud-chart", "nats" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

**Describe the solution you'd like**

Ideally the helm chart deployment incorporates the security best practices expected on modern k8s deployments to reflect the already awesome internal security considerations.

**Describe alternatives you've considered**
Alternatives are to create a less secure more permissive namespace for wasmcloud/nats to function with root permissions inside of the clusters pods.

**Additional context**

joonas:
@VestigeJ first off, thanks for taking the time to write this feedback down, it's always great to hear from folks how we can make things better!

Let me take a look at this and see what we can do with our chart to make the story better.

I'll make sure to include this issue and ping you on the PR once we've got the necessary improvements ready to go.

PR #984, opened 2023-11-10
Adopt Restricted profile recommendations as default securityContext for wasmcloud-chart containers

joonas:
## Feature or Problem

It was pointed out in #978 that the defaults we have in our chart's `values.yaml` for securityContext that gets configured for the both of the containers included in the Chart does not meet the recommendations from a security standpoint.

This PR adds a set of defaults that meets the [Restricted profile as outlined in Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/).

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/978

## Testing

I've tested this locally against a [KinD cluster running 1.27.3 (latest available KinD version)](https://github.com/kubernetes-sigs/kind/releases/tag/v0.20.0), and from review the [`1.28` changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.28.md), it did not appear anything had changed in a material way that would impact this set of configuration.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -113,8 +113,8 @@ podSecurityContext:
   # fsGroup: 2000
 
 securityContext:
-  runAsUser: 10000
-  runAsGroup: 10000
+  runAsUser: 1000
+  runAsGroup: 1000
   runAsNonRoot: true
   allowPrivilegeEscalation: false
   capabilities:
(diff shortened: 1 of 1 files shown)


joonas:
@VestigeJ please feel free to chime in if this misses the mark on what you would expect to have seen here

connorsmith256:
@thomastaylor312 can you handle the review for this?

thomastaylor312:
Yep just coming back from Kubecon week, so catching up on stuff now

thomastaylor312 (approved):

PR #990, opened 2023-11-10
feat(*): Adds support for actor config

thomastaylor312:
This is a fairly large PR because it is adding several new control interface topics as well as actually adding the actor config feature.

This feature was motivated by 2 major reasons:

1. We have been needing something like this for a while, at the very least for being able to configure link names in an actor at runtime
2. There aren't currently any active (yes there were some in the past) efforts to add a generic `wasi:cloud/guest-config` interface that can allow any host to provide config values to a component. I want to use this as a springboard for the conversation in wasi-cloud as we will start to use it and can give active feedback as to how the interface should be shaped

With that said, note that this is only going to be added for actors built against the component model. Since this is net new functionality, I didn't think it was worth it to try to backport.

As for testing, I have tested that an actor can import the functions and get the values via the various e2e tests and also manually validated that all of the new topics work.


crates/runtime/tests/runtime.rs
```diff
@@ -426,8 +443,11 @@ fn new_runtime(
         .expect("failed to construct runtime")
 }
 
-async fn run(wasm: impl AsRef<Path>) -> anyhow::Result<Vec<(logging::Level, String, String)>> {
-    const BODY: &str = r#"{"min":42,"max":4242,"port":42424}"#;
+async fn run(
+    wasm: impl AsRef<Path>,
+    is_module: bool,
```
vados-cosmonic, line 448: Maybe you could use [`wasmparser::is_core_wasm`](https://docs.rs/wasmparser/latest/wasmparser/struct.Parser.html#method.is_core_wasm) here?

thomastaylor312: I forgot that existed, but since this is a test, it is probably easier to specify a bool directly. I'll give it a try when I push any other updates from code review

tests/wasmbus.rs
```diff
@@ -1250,6 +1302,7 @@ expected: {expected_labels:?}"#
                 module_http_port,
                 module_nats_client.clone(),
                 &mut module_redis_client,
+                true,
```
rvolosatovs, line 1305: like mentioned above, I'd make the function *return* the values and assert after the call directly here

crates/host/src/wasmbus/mod.rs
```diff
@@ -181,20 +188,24 @@ impl Queue {
...
+        let (registries, pings, links, queries, auction, commands, inventory, config) = try_join!(
             nats.subscribe(format!("{topic_prefix}.{lattice_prefix}.registries.put",)),
             nats.subscribe(format!("{topic_prefix}.{lattice_prefix}.ping.hosts",)),
             nats.queue_subscribe(
                 format!("{topic_prefix}.{lattice_prefix}.linkdefs.*"),
                 format!("{topic_prefix}.{lattice_prefix}.linkdefs",)
             ),
             nats.queue_subscribe(
-                format!("{topic_prefix}.{lattice_prefix}.get.*"),
+                format!("{topic_prefix}.{lattice_prefix}.get.>"),
```
rvolosatovs, line 199: This seems out-of-scope 🤔 why is this change made?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2374,12 +2481,13 @@ impl Host {
                     links
                 },
             );
-        let origin = WasmCloudEntity {
+        let origin: WasmCloudEntity = WasmCloudEntity {
```
rvolosatovs, line 2484: Why do we need to specify a type here now?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3651,6 +3803,68 @@ impl Host {
...
+            return Ok(ACCEPTED.into());
+        };
+        let futs = all_data
+            .keys()
+            .map(|key| self.handle_config_delete(entity_id, key));
+        futures::future::join_all(futs)
+            .await
+            .into_iter()
+            .collect::<Result<Vec<_>, anyhow::Error>>()
+            .context("Unable to delete all config keys. Some keys may remain")?;
```
rvolosatovs, line 3863: How about https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_for_each_concurrent instead? It would avoid allocating a vector here and would be a bit easier to read

rvolosatovs (commented):
Good job, thanks!
Overall, LGTM, but please fix the clippy issues and I think could improve the testing a bit.

crates/runtime/tests/runtime.rs
```diff
@@ -591,20 +621,32 @@ async fn run(wasm: impl AsRef<Path>) -> anyhow::Result<Vec<(logging::Level, Stri
...
         long_value: _,
+        config_value,
+        all_config,
     } = serde_json::from_slice(&body).context("failed to decode body as JSON")?;
     ensure!(Uuid::from_str(&guid).is_ok());
     ensure!(
         (42..=4242).contains(&random_in_range),
         "{random_in_range} should have been within range from 42 to 4242 inclusive"
     );
+    if !is_module {
```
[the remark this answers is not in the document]
thomastaylor312: Actually that is what I started with but then I ran into...something that I can't remember. I'll probably do it again

crates/host/src/wasmbus/mod.rs
thomastaylor312: It isn't out of scope actually! If you look at the handler for getting config values, it matches multiple parts of the remaining subject so it wasn't actually listening for requests

crates/host/src/wasmbus/mod.rs
thomastaylor312: Honestly I might have tab completed something on accident. I'll just remove

crates/host/src/wasmbus/mod.rs
thomastaylor312: So I did attempt to do this, however, `all_data.keys()` is not a stream, just an iterator. I could wrap it into a stream but it was making it more unreadable, so I opted to use `try_join_all` (which I always forget about). If we're worried about the `Vec` it allocates for results, I can always put everything in a `FuturesUnordered` and then do the stream stuff that way

crates/host/src/wasmbus/mod.rs
```diff
@@ -1604,15 +1649,16 @@ fn linkdef_hash(
...
 }
 
-#[instrument(level = "debug", skip_all)]
-async fn create_lattice_metadata_bucket(
+#[instrument(skip(jetstream, bucket_message_name))]
+async fn create_bucket(
     jetstream: &async_nats::jetstream::Context,
     bucket: &str,
-) -> anyhow::Result<()> {
+    bucket_message_name: &str,
```
connorsmith256, line 1656: Nit: is this needed? I think I'd prefer to see the actual name for debug purposes

crates/host/src/wasmbus/mod.rs
```diff
@@ -1622,12 +1668,12 @@ async fn create_lattice_metadata_bucket(
         })
         .await
     {
-        Ok(_) => {
-            info!(%bucket, "created lattice metadata bucket with 1 replica");
-            Ok(())
+        Ok(store) => {
+            info!("created {bucket_message_name} bucket with 1 replica");
```
connorsmith256, line 1672: Due to how log ingestion querying (and billing) tends to work, in general we should emit log messages that have a low cardinality, and put dynamic data in separate fields

crates/host/src/wasmbus/mod.rs
```diff
@@ -1604,15 +1649,16 @@ fn linkdef_hash(
     hex::encode_upper(hash.finalize())
 }
 
-#[instrument(level = "debug", skip_all)]
-async fn create_lattice_metadata_bucket(
+#[instrument(skip(jetstream, bucket_message_name))]
```
connorsmith256, line 1652: Does this need to trace at the info level or can it stay debug?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3578,6 +3686,49 @@ impl Host {
         Ok(res.into())
     }
 
+    #[instrument(skip(self))]
```
connorsmith256, line 3689: Let's set the instrument level to debug (or trace?) on these

crates/host/src/wasmbus/mod.rs
```diff
@@ -3578,6 +3686,49 @@ impl Host {
...
+    async fn handle_config_get_one(&self, entity_id: &str, key: &str) -> anyhow::Result<Bytes> {
+        trace!("getting config");
+        let json = match self
+            .config_data_cache
+            .read()
+            .await
+            .get(entity_id)
+            .and_then(|map| map.get(key))
+        {
+            Some(data) => json!({
```
connorsmith256, line 3699: Can we re-use `GetConfigKeyResponse` here?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3651,6 +3802,66 @@ impl Host {
         Ok(ACCEPTED.into())
     }
 
+    #[instrument(skip(self, data))]
```
connorsmith256, line 3805: Debug? I would also consider `skip_all` and putting the fields on the log line. It renders more clearly

crates/host/src/wasmbus/mod.rs
```diff
@@ -3651,6 +3802,66 @@ impl Host {
...
+            .await?;
+
+        Ok(ACCEPTED.into())
+    }
+
+    #[instrument(skip(self))]
+    async fn handle_config_clear(&self, entity_id: &str) -> anyhow::Result<Bytes> {
+        debug!("handle config deletion");
+
+        let config_cache = self.config_data_cache.read().await;
```
connorsmith256, line 3851: Why do we use `self.config_data` in the other functions but `self.config_data_cache` here?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3676,92 +3887,106 @@ impl Host {
...
-        let mut parts = subject
+        let mut parts = message
+            .subject
             .trim()
             .trim_start_matches(&self.ctl_topic_prefix)
             .trim_start_matches('.')
             .split('.')
             .skip(1);
-        trace!(%subject, "handling control interface request");
+        trace!("handling control interface request");
```
connorsmith256, line 3905: See other comment, but I think this makes the logs harder to read

crates/host/src/wasmbus/mod.rs
```diff
@@ -4038,6 +4261,31 @@ impl Host {
         }
     }
 
+    #[instrument(skip_all, fields(bucket = %entry.bucket, key = %entry.key, revision = %entry.revision, operation = ?entry.operation))]
```
connorsmith256, line 4264: Let's use trace level here

crates/host/src/wasmbus/mod.rs
```diff
@@ -181,20 +188,24 @@ impl Queue {
         host_key: &KeyPair,
     ) -> anyhow::Result<Self> {
         let host_id = host_key.public_key();
-        let (registries, pings, links, queries, auction, commands, inventory) = try_join!(
+        let (registries, pings, links, queries, auction, commands, inventory, config) = try_join!(
```
connorsmith256, line 191: Since we're considering adding several topics to the control interface, should we put this functionality behind an `--experimental` flag on the host? Separate/fine-grained feature flags would be better, but we don't have a way to do that yet

connorsmith256 (changes requested):
Overall approach looks great. A few nits on logging/tracing, and a question about how we want to handle experimental features

crates/host/src/wasmbus/mod.rs
thomastaylor312: The rebase with your changes was a bit difficult so this got lost in the shuffle. Will change

crates/host/src/wasmbus/mod.rs
thomastaylor312: So I did consider that here. All this is actually doing is making sure the message is correct. It is a little awkward yes, but almost all of the other logic is the same, so this only ends up with 2 messages. However, I think just logging the bucket name is probably good enough at this point

crates/host/src/wasmbus/mod.rs
thomastaylor312: I think we should use the types everywhere, but there are two main reasons I didn't:

1. We should have these types in a separate crate to be used by both the client and the server (we could put them in the host as well, but that is a heavy crate to pull in)
2. I was matching the style done elsewhere in this code to maintain consistency

crates/host/src/wasmbus/mod.rs
thomastaylor312: This is something we should maybe decide on, but I personally don't like manually setting fields every time I need them. If there is something I think should be on every log line for context, I don't want to manually put it in there

crates/host/src/wasmbus/mod.rs
thomastaylor312: I was trying to shortcut a call to the database to list all keys, I can change this though

crates/host/src/wasmbus/mod.rs
thomastaylor312: I don't really like the idea of needing to specify a flag to turn this on unless we truly think this is too experimental for people to use normally. We really should have some versioned APIs instead, but that would require changing all topics

crates/host/src/wasmbus/mod.rs
thomastaylor312: Oh I remember now, I was trying to delete without listing all keys, so I would prefer not to change this if that's ok. Will add a comment though

crates/host/src/wasmbus/mod.rs
connorsmith256: I'm okay with this for now, and agree we should move common types somewhere in the future 👍

crates/host/src/wasmbus/mod.rs
connorsmith256: We synced offline and agreed in theory putting the field son the span makes more sense, but the formatter we're using makes the logs hard to read that way 😢

crates/host/src/wasmbus/mod.rs
connorsmith256: Synced offline and agreed this one doesn't need an experimental flag

crates/host/src/wasmbus/mod.rs
connorsmith256: Leaving this one unresolved because it's still instrumented at an info level

crates/host/src/wasmbus/mod.rs
connorsmith256: Oh, I see now, we're still using `self.config_data` inside `handle_config_delete` 👍

connorsmith256 (approved):
LGTM, just a reminder there are still a few functions instrumented at info level

crates/host/src/wasmbus/mod.rs
thomastaylor312: Should be done now

thomastaylor312:
So, we had a great conversation in the community call around actor config. I've done some more thinking through what I think are all the options and wanted to enumerate them here along with what I think is the best option.

## Option 0: Leave it as is with no config

I listed this first because I think this is a non-starter. @LiamRandall gave a great real life example in the community call about why he needed it for a Cosmonic customer. We also have had plenty of people ask for it to configure things like link names dynamically. In addition, other efforts around how to configure things (such as with env vars) are already underway and being used in other wasm projects

## Option 1: Pull-based (aka `get`) config

**Pros**
- Simplest option as all an actor needs to do is call the exposed `get` function to get a config value. The simpler something is, the easier it is to get accepted as a standard wasm community interface
- Closest match to some of the other things people are already doing (like the aforementioned env vars)

**Cons**
- This is a footgun for "reactor" (i.e. long running) components as you could cache a value in a static variable. After thinking about it, this is not actually a footgun for actors as actors now scale back to 0 when they're not being run so you won't actually run into this. It will be a possible footgun for future wasi-based providers

## Option 2: Push-based config

**Pros**
- No footguns here as config is always pushed to an actor. This means it should always have the latest code and the host has complete control over when things are sent

**Cons**
- This puts a lot of complexity on the writer of a component. They will have to write a handler that gets called on push and then store that data, which will be a lot of copy paste (or a common library that becomes a dep they have to update)
- Probably most importantly: This method _will not_ work for "command" style actors (i.e. one shot or run once) as these cannot receive push data. 

## Option 3: Push and pull (Option 1 + Option 2)

**Pros**
- People have a choice of what they can use and which problem you have to deal with (the footgun or the overhead)

**Cons**
- It is much more complex to handle two different config interfaces simultaneously
- This is less likely to be accepted at the community level

## Additional notes

Please note that in all options there is a chance for race conditions where an actor gets/receives a config item before/after a call starts and it has a stale value. But c'est la vie when it comes to distributed systems. If you need strict guarantees, use a DB instead. Most config will be set once and not changed too often.

## Conclusion

Based on the options presented above, there are a few key takeaways:

1. We definitely need actor config, so we can't sidestep with Option 0
2. We support reactor and command style actors, so Option 2 doesn't work
3. Option 2 (and 3) puts more burden on devs and would be less likely to be accepted as a wasi-cloud interface

Option 1 (which is what this PR supports) has a footgun that _can be clearly documented/communicated_ and will only apply to wasi-based providers. If you really want to cache config in those cases, that is something we can address with users down the road based on clearly-defined and concrete use cases

Thoughts?

Issue #962, opened 2023-11-08
[BUG] wasmCloud UI doesn't load in browser


jackcarey:
**Describe the bug**
The web UI for wasmCloud doesn't load.

**To Reproduce**
0. Folowing the commands on https://wasmcloud.com/docs/installation?os=windows
1. Open Powershell as administrator. Run: `choco install wash`
2. Run `wash new actor -t hello hello` to generate a template project
3. change directory into '/hello'.
4. Run `wash app deploy wadm.yaml`, to deploy the app.
5. Navigate to `localhost:8080` in your browser. It should load, displaying "Hello World" text.
6. In a new Powershell admin window, run `wash ui --experimental`. The terminal should say `Washboard running on http://localhost:3030 Hit CTRL-C to stop`.
7. Navigate to `localhost:3030` in your browser. You'll receive the error message below.

**Expected behavior**
The web UI should load to a usable state.

**Screenshots**
```
Unexpected Application Error!
Failed to execute 'createElement' on 'Document': The tag name provided ('/assets/logo-wide-0e72ceaf.svg') is not a valid name.
Error: Failed to execute 'createElement' on 'Document': The tag name provided ('/assets/logo-wide-0e72ceaf.svg') is not a valid name.
    at UE (http://localhost:3030/assets/index-3587af0c.js:40:16783)
    at Gx (http://localhost:3030/assets/index-3587af0c.js:40:40071)
    at Wx (http://localhost:3030/assets/index-3587af0c.js:40:39824)
    at GE (http://localhost:3030/assets/index-3587af0c.js:40:39691)
    at gc (http://localhost:3030/assets/index-3587af0c.js:40:39545)
    at Fh (http://localhost:3030/assets/index-3587af0c.js:40:35913)
    at Bx (http://localhost:3030/assets/index-3587af0c.js:40:34864)
    at P (http://localhost:3030/assets/index-3587af0c.js:25:1602)
    at MessagePort.L (http://localhost:3030/assets/index-3587af0c.js:25:1973)
```
![image](https://github.com/wasmCloud/wasmCloud/assets/35701383/5e36579f-dd96-4098-9f7d-ae7db9ff400e)

**Environment (please complete the following information) **
 - OS: Windows 11
 - Shell: Powershell
 - wasmcloud Version: 0.80.0

**Additional context**
N/A

connorsmith256:
[...]

PR #996, opened 2023-11-11
build(deps): added missing dependencies to washboard-ui

jackcarey:
## Feature or Problem

`radix-ui/react-collapsible` and `react-router` are used in the front end but didn't appear in `package.json`. This means they weren't installed alongside other dependencies, causing an error in the terminal and the browser to hang when trying to load the UI. This PR hotfixes that by adding them to the package list so that the UI renders.

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/962

## Release Information

next

## Consumer Impact

See details above, no steps are needed.

## Testing

n/a

### Unit Test(s)
 
n/a

### Acceptance or Integration

n/a

### Manual Verification

1. Clone repo.
2. cd to `/washboard-ui`.
3. Run `npm run dev`.
4. Check the terminal for error messages.
5. Navigate to `localhost:5173` in a browser.


thomastaylor312:
This looks good to me, but deferring to @lachieh for final review since I don't trust my JS

lachieh:
Thanks @jackcarey! I appreciate the investigation. While this does technically solve the issue, it isn't the root cause.  Both of these packages are listed in the `package-lock.json` file so the error is actually a result of not installing _any_ dependencies.  We actually just need to update the instructions to include `npm install` before running `npm run dev`.

You can test this yourself by reverting to `main`, deleting the `node_modules` folder, and running `npm install` before then running `npm run dev`.

`react-router` is a dependency of `react-router-dom` (see 6745):
https://github.com/wasmCloud/wasmCloud/blob/50d50b710b4dabfdf33d712a63b06417b07ce988/washboard-ui/package-lock.json#L6739-L6754

`@radix-ui/react-collapsible` is a dependency of `@radix-ui/react-accordion` (see 1116):
https://github.com/wasmCloud/wasmCloud/blob/50d50b710b4dabfdf33d712a63b06417b07ce988/washboard-ui/package-lock.json#L1109-L1124

Issue #939, opened 2023-10-23
control-interface client should check for empty parameters


thomastaylor312:
All of our functions do no validation to make sure that the passed host IDs and OCI refs are not empty. It would be a nicer user experience if we returned early with an error in cases where an empty string is passed

PR #1001, opened 2023-11-13
control-interface client should check for empty parameters

ahmedtadde:
## Feature or Problem
All of our functions do no validation to make sure that the passed host IDs and OCI refs are not empty. It would be a nicer user experience if we returned early with an error in cases where an empty string is passed

## Related Issues
#939

## Release Information
`wasm-cloud-control-interface v0.31.0`

## Consumer Impact
better DX

### Unit Test(s)
[test_parse_identifier](https://github.com/ahmedtadde/wasmCloud/blob/2947dd023923a564af1853589f276677743e4c03/crates/control-interface/src/lib.rs#L783)

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -761,7 +761,6 @@ fn parse_identifier<T: AsRef<str>>(kind: &IdentifierKind, value: T) -> Result<St
     };
 
     match kind {
-        IdentifierKind::HostId => assert_non_empty_string(value, Some("Host ID".to_string())),
         IdentifierKind::ActorId => assert_non_empty_string(value, Some("Actor ID".to_string())),
         IdentifierKind::ActorRef => {
             assert_non_empty_string(value, Some("Actor OCI reference".to_string()))
@@ -802,31 +801,5 @@ mod tests {
         });
         println!("Listening to Cloud Events for 120 seconds. Then we will quit.");
         tokio::time::sleep(Duration::from_secs(120)).await;
-    }
-
-    #[test]
-    fn test_parse_identifier() -> Result<()> {
-        assert!(parse_identifier(&IdentifierKind::HostId, "").is_err());
-        assert!(parse_identifier(&IdentifierKind::HostId, " ").is_err());
-        let host_id = parse_identifier(&IdentifierKind::HostId, "             ");
-        assert!(host_id.is_err(), "parsing host id should have failed");
-        assert!(host_id
-            .unwrap_err()
-            .to_string()
-            .contains("Empty string provided for Host ID"));
-        let provider_ref = parse_identifier(&IdentifierKind::ProviderRef, "");
-        assert!(
-            provider_ref.is_err(),
-            "parsing provider ref should have failed"
-        );
-        assert!(provider_ref
-            .unwrap_err()
-            .to_string()
-            .contains("Empty string provided for Provider OCI reference"));
-        assert!(parse_identifier(&IdentifierKind::HostId, "host_id").is_ok());
-        let actor_id = parse_identifier(&IdentifierKind::ActorId, "            iambatman  ")?;
-        assert_eq!(actor_id, "iambatman");
-
-        Ok(())
     }
 }
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -173,8 +173,11 @@ impl Client {
     /// Retrieves the contents of a running host
     #[instrument(level = "debug", skip_all)]
     pub async fn get_host_inventory(&self, host_id: &str) -> Result<HostInventory> {
-        let subject =
-            broker::queries::host_inventory(&self.topic_prefix, &self.lattice_prefix, host_id);
+        let subject = broker::queries::host_inventory(
+            &self.topic_prefix,
+            &self.lattice_prefix,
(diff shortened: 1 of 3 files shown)


crates/control-interface/src/lib.rs
```diff
@@ -674,6 +709,51 @@ pub async fn collect_sub_timeout<T: DeserializeOwned>(
...
+    ProviderRef,
+    ContractId,
+    LinkName,
+}
+
+//NOTE(ahmedtadde): For an initial implementation, we just want to make sure that the identifier is, at very least, not an empty string.
+//This parser should be refined over time as needed.
+fn parse_identifier<T: AsRef<str>>(kind: &IdentifierKind, value: T) -> Result<String> {
+    let value = value.as_ref();
+    let assert_non_empty_string = |input: &str, message: Option<String>| -> Result<String> {
```
thomastaylor312, line 726: Nit: Can we just break this out into an actual function rather than an enclosure? Also, since this is a private function and we are always passing in a message context, let's just remove the `Option<String>` and just make it a `&str` (which also avoids the allocation)

crates/control-interface/src/lib.rs
```diff
@@ -674,6 +709,51 @@ pub async fn collect_sub_timeout<T: DeserializeOwned>(
...
+        if input.trim().is_empty() {
+            Err(format!(
+                "Empty string provided for {}",
+                message.unwrap_or_else(|| "entity identifier".to_string())
+            )
+            .into())
+        } else {
+            Ok(input.trim().to_string())
```
thomastaylor312, line 734: Might be better to just have this return a unit type instead of the string. In the future, we might add additional verifications, which means we should let the caller decide when to clean up the input and convert it to a string

thomastaylor312 (changes requested):
Thanks for tackling this! It was a good idea to think ahead for supporting other validations we might add in the future. There are some small Rust specific changes and organizational requests here, but nothing big

t3hmrman:
Hey also @ahmedtadde @thomastaylor312 what do you think about introducing a tuple struct to manage this complexity/check? Generally leaning on the type system here is the way to go (and letting that change percolate across the code-base gardually), to ensure that we check stuff at the edges of the domain and are safe inside.

It could look something like this:

```rust
use anyhow::{bail, Result};

#[derive(Debug, PartialEq, Eq)]
pub struct HostId(String);

impl HostId {
    pub fn new(s: impl AsRef<str>) -> Result<Self> {
        let s = s.as_ref().trim();
        if s.is_empty() {
            bail!("HostIds cannot be empty (invalid value [{s}])");
        }
        Ok(HostId(s.into()))
    }
}
```

There would be a little repetition across the types but we'd be able to pass around `HostId`s in interfaces which is a lot more legible for the future

vados-cosmonic:
Hey also @ahmedtadde @thomastaylor312 what do you think about introducing a tuple struct to manage this complexity/check? Generally leaning on the type system here is the way to go (and letting that change percolate across the code-base gardually), to ensure that we check stuff at the edges of the domain and are safe inside.

It could look something like this:

```rust
use anyhow::{bail, Result};

#[derive(Debug, PartialEq, Eq)]
pub struct HostId(String);

impl HostId {
    pub fn new(s: impl AsRef<str>) -> Result<Self> {
        let s = s.as_ref().trim();
        if s.is_empty() {
            bail!("HostIds cannot be empty (invalid value [{s}])");
        }
        Ok(HostId(s.into()))
    }
}
```

There would be a little repetition across the types but we'd be able to pass around `HostId`s in interfaces which is a lot more legible for the future

ahmedtadde:
hey @vados-cosmonic, what you mentioned  is something that i thought about as well.
are you envisioning that the consumers/callers will pass in, say `host_id: HostId` as params instead of the current `host_id: &str | string`? or, are these nominal types just internal to the client lib?

one main thing that stopped me from going down the rabbit hole of type refinement for this is that `let host_id = HostId(" 
        ".to_string());` is still a valid declaration... even though, fundamentally, i wouldn't want that code to compile. thought, this matters less if these types are meant to be used just internally for sanity checking... we would just need to make sure the internal code always uses `HostId::new()` (same for the other refinement types).

vados-cosmonic:
To prevent breaking end-users it might make sense to have it be a more internal thing -- we can be liberal in what we accept (basically any string) but convert as soon as possible internally to `HostId`

> one main thing that stopped me from going down the rabbit hole of type refinement for this is that let host_id = HostId("  ".to_string()); is still a valid declaration... even though, fundamentally, i wouldn't want that code to compile. thought, this matters less if these types are meant to be used just internally for sanity checking... we would just need to make sure the internal code always uses HostId::new() (same for the other refinement types).

Hmnn, `HostId("".to_string())` should *not* compile in a sufficiently encapsulated module -- you should be able to move it around such that all the external modules won't be able to manipulate that tuple struct's internals since they're not `pub` or `pub(crate)`.

ahmedtadde:
hey @vados-cosmonic , in the last commit, I made some changes to reflect your suggestion (only for `HostId` though). let me know if this is inline with what you have in mind. If yes, I will go ahead and do the same for the remaining identifiers. it fells like a bit of boilerplate, but further refinements (per each identifier as needed) down the line would make it worth it.

thomastaylor312 (approved):

thomastaylor312 (approved):

PR #1010, opened 2023-11-14
Docs/wash installation instructions

connorsmith256:
## Feature or Problem
Now that the [formula installs from released binaries](https://github.com/wasmCloud/homebrew-wasmcloud/pull/51), it's no longer needed to `tap`

I also updated outdated URLs to wash


thomastaylor312 (approved):

lachieh (approved):

connorsmith256:
Bypassing branch protections because we're only lacking approval from `examples-maintainers`, and the only change I made there was to a README

PR #1022, opened 2023-11-16
feat(wash)!: Removes need for actor/provider/host IDs in almost all cases

thomastaylor312:
This is something that has been bugging me for a while. It has been such a pain to look up and copy paste all the proper IDs to run various wash commands.

This PR is a breaking change for several commands (like stop provider, which reorders some things and most other commands requiring a host ID now have that as an optional flag) and makes it so you can pass a string that it will attempt to match on to find IDs

## Release Information

vNext

## Consumer Impact

Namely the structure of many of the commands have changed, but since everything requires less, I think this is fine

## Testing

I added a few more tests in places where we already had some and then manually verified the rest

### Manual Verification

```
./target/debug/wash link put echo 'http server' wasmcloud:httpserver 'ADDRESS=0.0.0.0:8080'

Published link (echo) <-> (http server) successfully
```

After starting httpclient:

```
./target/debug/wash link put echo http wasmcloud:httpserver 'ADDRESS=0.0.0.0:8080'

Error advertising link: Multiple matches found with the provided search term: [VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M (HTTP Server), VCCVLH4XWGI3SGARFNYKYT2A32SUYA2KVAIV2U2Q34DQA7WWJPFRKIKM (HTTP Client)]
```

Other examples of the updated commands:

```
./target/debug/wash stop actor echo

Actor [Echo] stopped
```

```
./target/debug/wash stop provider VCCVL wasmcloud:httpclient

Provider [HTTP Client] stopped successfully
```

```
./target/debug/wash link del echo wasmcloud:httpserver

Deleted link for echo on wasmcloud:httpserver (default) successfully
```

```
./target/debug/wash stop provider --host-id dark-sun client wasmcloud:httpclient

Provider [HTTP Client] stopped successfully
```

I also tested that passing in IDs everywhere also worked

### DIFF
--- a/crates/wash-cli/src/common/link_cmd.rs
+++ b/crates/wash-cli/src/common/link_cmd.rs
@@ -7,7 +7,7 @@ use wash_lib::cli::link::{
     LinkQueryCommand,
 };
 use wash_lib::cli::{CommandOutput, OutputKind};
-use wash_lib::id::{validate_contract_id, ModuleId, ServiceId};
+use wash_lib::id::validate_contract_id;
 use wasmcloud_control_interface::LinkDefinition;
 
 use crate::appearance::spinner::Spinner;
@@ -15,10 +15,12 @@ use crate::ctl::{link_del_output, links_table};
 
 /// Generate output for link put command
 pub fn link_put_output(
-    actor_id: &ModuleId,
-    provider_id: &ServiceId,
+    actor_id: impl AsRef<str>,
+    provider_id: impl AsRef<str>,
     failure: Option<String>,
 ) -> Result<CommandOutput> {
+    let actor_id = actor_id.as_ref();
+    let provider_id = provider_id.as_ref();
     match failure {
         None => {
             let mut map = HashMap::new();
--- a/crates/wash-cli/tests/wash_start.rs
+++ b/crates/wash-cli/tests/wash_start.rs
@@ -1,5 +1,7 @@
 mod common;
 
+use std::process::Stdio;
+
 use common::{TestWashInstance, ECHO_OCI_REF, PROVIDER_HTTPSERVER_OCI_REF};
 
 use anyhow::{Context, Result};
@@ -13,7 +15,7 @@ use wash_lib::cli::output::StartCommandOutput;
     not(can_reach_wasmcloud_azurecr_io),
     ignore = "wasmcloud.azurecr.io is not reachable"
 )]
-async fn integration_start_actor_serial() -> Result<()> {
+async fn integration_start_stop_actor_serial() -> Result<()> {
     let wash_instance = TestWashInstance::create().await?;
 
     let output = Command::new(env!("CARGO_BIN_EXE_wash"))
@@ -39,12 +41,35 @@ async fn integration_start_actor_serial() -> Result<()> {
         serde_json::from_slice(&output.stdout).context("failed to parse output")?;
     assert!(cmd_output.success, "command returned success");
 
+    // Test stopping using only aliases, yes I know this mixes stop and start, but saves on copied
+    // code
+    let status = Command::new(env!("CARGO_BIN_EXE_wash"))
+        .args([
+            "stop",
+            "actor",
+            "echo",
+            "--output",
+            "json",
+            "--timeout-ms",
+            "40000",
+            "--ctl-port",
+            &wash_instance.nats_port.to_string(),
+        ])
+        .kill_on_drop(true)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()
+        .await
+        .context("failed to start actor")?;
+
+    assert!(status.success(), "Sucessfully stopped actor");
+
     Ok(())
 }
 
 #[tokio::test]
 #[serial]
-async fn integration_start_provider_serial() -> Result<()> {
+async fn integration_start_stop_provider_serial() -> Result<()> {
     let wash_instance = TestWashInstance::create().await?;
 
     let output = Command::new(env!("CARGO_BIN_EXE_wash"))
@@ -69,6 +94,30 @@ async fn integration_start_provider_serial() -> Result<()> {
     let cmd_output: StartCommandOutput =
         serde_json::from_slice(&output.stdout).context("failed to parse output")?;
     assert!(cmd_output.success, "command returned success");
+
+    // Test stopping using only aliases, yes I know this mixes stop and start, but saves on copied
+    // code
+    let status = Command::new(env!("CARGO_BIN_EXE_wash"))
+        .args([
+            "stop",
+            "provider",
+            "server",
+            "wasmcloud:httpserver",
+            "--output",
+            "json",
+            "--timeout-ms",
+            "40000",
+            "--ctl-port",
+            &wash_instance.nats_port.to_string(),
+        ])
+        .kill_on_drop(true)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()
+        .await
+        .context("failed to start actor")?;
+
+    assert!(status.success(), "Sucessfully stopped provider");
 
     Ok(())
 }
--- a/crates/wash-lib/src/cli/get.rs
+++ b/crates/wash-lib/src/cli/get.rs
@@ -2,7 +2,11 @@ use anyhow::{Context, Result};
 use clap::Parser;
 use wasmcloud_control_interface::{Host, HostInventory};
 
-use crate::{common::boxed_err_to_anyhow, config::WashConnectionOptions, id::ServerId};
+use crate::{
+    common::{boxed_err_to_anyhow, get_all_inventories},
+    config::WashConnectionOptions,
+    id::ServerId,
+};
 
 use super::CliConnectionOpts;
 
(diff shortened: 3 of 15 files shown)


crates/wash-cli/src/common/link_cmd.rs
```diff
@@ -7,16 +7,16 @@ use wash_lib::cli::link::{
...
 
 /// Generate output for link put command
 pub fn link_put_output(
-    actor_id: &ModuleId,
-    provider_id: &ServiceId,
+    actor_id: &str,
+    provider_id: &str,
     failure: Option<String>,
 ) -> Result<CommandOutput> {
     match failure {
```
vados-cosmonic, line 22:
```suggestion
    actor_id: impl AsRef<str>,
    provider_id: impl AsRef<str>,
    failure: Option<String>,
) -> Result<CommandOutput> {
    let actor_id = actor_id.as_ref();
    let provider_id = provider_id.as_ref();
    match failure {
```

crates/wash-cli/tests/wash_start.rs
```diff
@@ -39,12 +41,35 @@ async fn integration_start_actor_serial() -> Result<()> {
...
+            "40000",
+            "--ctl-port",
+            &wash_instance.nats_port.to_string(),
+        ])
+        .kill_on_drop(true)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status()
+        .await
+        .context("failed to start actor")?;
```
vados-cosmonic, line 63: Would you mind hanging this off of `impl TestWashInstance`? I think it'd be very generally useful

crates/wash-cli/tests/wash_start.rs
```diff
@@ -70,5 +95,29 @@ async fn integration_start_provider_serial() -> Result<()> {
...
```
vados-cosmonic, line 118: Would you mind hanging the provider stop also off of `impl TestWashInstance` ? Definitely something people would want to do commonly from tests

crates/wash-lib/src/cli/get.rs
```diff
@@ -53,34 +57,25 @@ pub enum GetCommand {
...
-    let host_ids = if let Some(host_id) = cmd.host_id {
-        vec![host_id.to_string()]
+    if let Some(host_id) = cmd.host_id {
+        Ok(vec![client
+            .get_host_inventory(&host_id)
+            .await
+            .map_err(boxed_err_to_anyhow)?])
     } else {
-        let hosts = client.get_hosts().await.map_err(boxed_err_to_anyhow)?;
+        let hosts = get_all_inventories(&client).await?;
```
vados-cosmonic, line 71: Can we put a `.context` on this, *just in case* it fails here?

crates/wash-lib/src/cli/link.rs
```diff
@@ -4,19 +4,19 @@ use wasmcloud_control_interface::LinkDefinition;
...
     #[clap(flatten)]
     pub opts: CliConnectionOpts,
 
-    /// Public key ID of actor
-    #[clap(name = "actor-id", value_parser)]
-    pub actor_id: ModuleId,
+    /// Public key ID or name of actor to match on. If an actor name is given and matches multiple
+    /// actors, an error will be returned with a list of matching actors and their IDs.
+    #[clap(name = "actor-id")]
+    pub actor_id: String,
```
vados-cosmonic, line 19: NIT: Feels like we lost some specificity/protection here... I would have liked an enum to capture both these cases

crates/wash-lib/src/common.rs
```diff
@@ -1,64 +1,125 @@
...
+            write!(f, "{}", self.id)
+        }
+    }
+}
+
+impl Debug for Match {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(&self, f)
+    }
+}
```
vados-cosmonic, line 54: QUESTION: What do we get by using the display impl for debug vs the generated one? Just avoiding the default because it's sometimes hard to read?

crates/wash-lib/src/common.rs
```diff
@@ -1,64 +1,125 @@
...
-
     // Case insensitive searching here to make things nicer
     let value = value.to_lowercase();
     // If it wasn't an ID, get the claims
     let claims = ctl_client
         .get_claims()
         .await
-        .map_err(|e| FindIdError::Error(anyhow::anyhow!("Unable to get claims: {}", e)))?;
+        .map_err(boxed_err_to_anyhow)
+        .context("Unable to get claims for lookup")?;
```
vados-cosmonic, line 105:
```suggestion
        .context("unable to get claims for lookup")?;
```

crates/wash-lib/src/common.rs
```diff
@@ -1,64 +1,125 @@
...
                 .get(CLAIMS_SUBJECT)
                 .map(|s| s.as_str())
                 .unwrap_or_default();
-            // If it isn't a module, just skip
-            let id = match ModuleId::from_str(id) {
+            // If it doesn't parse to our type, just skip
+            let id = match T::from_str(id_str) {
                 Ok(id) => id,
                 Err(_) => return None,
             };
```
vados-cosmonic, line 117: Do you think a warning might be reasonable here? Maybe not

crates/wash-lib/src/cli/stop.rs
```diff
@@ -217,3 +265,58 @@ pub async fn stop_host(cmd: StopHostCommand) -> Result<CommandOutput> {
...
+    } else if all_matching.len() > 1 {
+        let matches = all_matching.into_iter().map(|(id, friendly_name)| Match {
+            id: id.into_string(),
+            friendly_name: Some(friendly_name),
+        });
+        Err(FindIdError::MultipleMatches(matches.collect()))
+    } else {
+        // SAFETY: We know there is exactly one match at this point
+        Ok(all_matching.into_iter().next().unwrap().0)
+    }
```
vados-cosmonic, line 321: How do you feel about using `match all_matching[..]` ? A bit easier to read IMO and then you don't need the safety comment I think (it's self-evident).

Also intermediate `matches` var probably unnecessary

crates/wash-lib/src/common.rs
```diff
@@ -72,12 +133,68 @@ pub async fn find_actor_id(
...
-                        id.into_string()
-                    }
+                .map(|(id, friendly_name)| Match {
+                    id: id.to_string(),
+                    friendly_name,
+                })
+                .collect(),
+        ))
+    } else {
+        // SAFETY: We know we have exactly one match at this point
```
vados-cosmonic, line 143: Same nit about that `[..]` pattern here, it would probably make this bit easier to read/remove requirement of the written assurance (even though it is legitimately impossible to not be the case there)

vados-cosmonic:
@thomastaylor312 mostly nits here, but I did have one over-arching question -- since we are changing the CLI here, is it possible to guide the user at all about the changed options? Feels like a lot of CLI calls are going to break and it would be nice to have an explanation ready. 

Though if people upgrade and read the release notes the breaking change would be called out there

crates/wash-cli/tests/wash_start.rs
thomastaylor312: There are times we are testing the output so I probably don't want to restrict it. Though it would probably be nice to have something that can generate the `Command` for you given your list of args

crates/wash-cli/tests/wash_start.rs
thomastaylor312: Also don't want to add this as part of the fixture stuff because it is explicitly testing that I can stop it, not being used as some way of cleaning it up

crates/wash-lib/src/cli/link.rs
thomastaylor312: I actually don't agree here. All of these go through processing after they are specified that processes them to `ModuleId` (or their relevant type). We need this loose so we can take all sorts of match-y things and having it parse to an enum feels like extra boilerplate with not a lot of benefit

crates/wash-lib/src/common.rs
thomastaylor312: Good question. So I did this on purpose because I wanted it to print the same way when an error was printed whether or not it used `Display` or `Debug`. That way the error always prints nicely

crates/wash-lib/src/common.rs
thomastaylor312: Probably not worth the warning because fetching claims returns _all_ claims. So we will always have some that fail to parse as this type

crates/wash-lib/src/cli/stop.rs
thomastaylor312: I did split up `matches` on purpose to make it a bit more readable, but I don't care either way how we do it. Let me try the match pattern though

thomastaylor312:
I don't think there is anything in clap that would help us guide. We could try to add some logic for detecting when they use the old format, but that feels like it could be really brittle and a heavier lift for something we'd remove in a version or two. I think release notes will be the way to go

crates/wash-lib/src/cli/stop.rs
thomastaylor312: Ok, so I tried this and it doesn't like matching the whole slice (and I am pretty sure it borrows rather than taking ownership, which is an admittedly minor detail here), so I'm not going to fight with it right now

crates/wash-lib/src/common.rs
thomastaylor312: See previous comment

crates/wash-lib/src/common.rs
vados-cosmonic: Ah OK got it, thanks for the explanation!

crates/wash-lib/src/cli/stop.rs
vados-cosmonic: Thanks for trying it, definitely not urgent so all good.

crates/wash-lib/src/common.rs
vados-cosmonic: Resolving, this isn't blocking

crates/wash-lib/src/cli/link.rs
vados-cosmonic: Ah it feels here like when we try to use these values elsewhere we'll want to know if it is one thing or the other (a "fuzzy" name or an explicit ID, or something else). Feels like the enum can also hold information as we enrich it/figure out what the thing is (if something is a provider fuzzy name it's probably going to stay the same).

All that said I'm OK with seeing how this evolves over time -- biasing for speed.

crates/wash-cli/tests/wash_start.rs
vados-cosmonic: Ah so I mean something [like this](https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-cli/tests/common/mod.rs#L272) so we can keep the tests small, and re-use common things to run on the actual instance.

vados-cosmonic (approved):
LGTM :rocket: 

Remaining nits are optional

crates/wash-cli/tests/wash_start.rs
thomastaylor312: I get what you're saying now. I created #1032 to track doing this later as this currently matches what is done for starting the actor

crates/wash-cli/tests/wash_start.rs
thomastaylor312: See #1032

Issue #664, opened 2023-09-19
[RFC] Host metrics


pgray:
# RFC: Host metrics

## Introduction

wasmcloud host metrics implemented using a [prometheus](https://prometheus.io) `/metrics` endpoint on a configurable port

## Motivation

In order to provide users insight into their running workloads wasmcloud needs to expose metrics that reveal the traffic actors and providers have handled over a period of time.

## Detailed Design

Hosts already provide basic metrics in health payloads like number of actors/providers/etc. running/configured

In order to provide insight into running workloads, a wasmcloud metrics implementation should provide the following metrics:

- host uptime (seconds)
- actors
  - invocation count
  - error count
  - latency (actor execution time)
- providers
  - restart count
  - provider running seconds (process uptime)

The main wasmcloud runtime process should instantiate a prometheus [registry](https://github.com/prometheus/client_rust/blob/master/examples/actix-web.rs#L8) and register/deregister actors/providers as they are scheduled on the host.

### tags

wasmcloud should tag metrics with basic identifying tags:

- host id
- lattice id
- oci ref (e.g. azurecr.io/wasmcloud/kv)
- version (e.g. v1.2.3)

## Backwards Compatibility

this functionality would be net-new so no backwards incompatible changes are planned

additionally, this RFC only seeks to implement metrics based on what a host can observe based on its runtime behavior... we leave provider-specific metrics for a later RFC

## Alternatives Considered

A nats publish topic capable of deriving host/provider/actor metrics would provide cluster operators a simplified deployment model for getting metrics from their wasmcloud deployments. Utilizing nats' queueing capabilities, an agent could be designed to support many copies running and sharing work. Depending on the cluster topology and tenancy, a nats metric agent might require elevated privileges in order to collect metrics from all hosts. Wasmcloud hosts would need to publish metrics to a shared topic in nats in order to avoid the nats metric agent inspecting all traffic flowing on wasmcloud topics.

## Unresolved Questions

- how should we implement per-provider metrics? (e.g. connection count, latency, etc)
- should we add user configurable tags?
- what other metrics should exist by default?

## Conclusion
Adding prometheus metrics to wasmcloud hosts would provide operators with a familiar and easy to use method for collecting metrics on their wasmcloud workloads. I hope this RFC sets defines enough to kick off the discussion.

connorsmith256:
CC @protochron and @stevelr who have expressed interest in this RFC before

connorsmith256:
[...]

PR #1025, opened 2023-11-16
Metrics RFC follow-up (DRAFT)

pgray:
## Feature or Problem

Draft of metrics RFC work in response to #664

This adds the first metrics outlined in the comments and a couple possible backends for them

Happy for someone to take it over if I don't get a chance to finish up soon

## Related Issues
 #664 

## Release Information
 next/tbd

## Consumer Impact
 adds host with various backends (debug stdout, otlp grpc/http, and prometheus)

## Testing
 tbd


src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
         env = "OTEL_EXPORTER_OTLP_ENDPOINT"
     )]
     otel_exporter_otlp_endpoint: Option<String>,
+    #[clap(long = "otel-metrics-endpoint", env = "OTEL_EXPORTER_METRICS_ENDPOINT", default_value = metrics::DEFAULT_OTEL_METRICS_ENDPOINT)]
```
vados-cosmonic, line 217: Missing doc here but also some spacing between the entries would help readability here IMO

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+    otel_metrics_endpoint: String,
+    #[clap(long = "otel-metrics-mode", env = "OTEL_EXPORTER_METRICS_MODE", default_value_t = OtlpMode::Http)]
+    otel_metrics_mode: OtlpMode,
+    // Enable metrics
+    #[clap(
+        long = "enable-metrics",
+        default_value_t = false,
+        env = "WASMCLOUD_ENABLE_METRICS"
```
vados-cosmonic, line 225: It's a big change but what do you think about calling this `metrics-enabled`? That would enable you to use `WASMCLOUD_METRICS_ENABLED` which makes `WASMCLOUD_METRICS_` as a prefix consistent. 

That said, I also get wanting to keep the UX of `--enable-<x>`... I personally like it when tools have all their options related to a certain feature/area consistently ENV-var'd. I know it's not a thing we've been doing up until now (otel options are just out there "in space", so to speak), but wanted to note it.

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+        default_value_t = MetricType::Debug,
+        env = "WASMCLOUD_METRICS_TYPE"
+    )]
+    metrics_type: MetricType,
+    #[clap(long = "metrics-prometheus-port", env = "WASMCLOUD_METRICS_PROMETHEUS_PORT", default_value_t = metrics::DEFAULT_PROMETHEUS_PORT)]
+    metrics_prometheus_port: i32,
+}
+
+#[derive(clap::ValueEnum, Clone, Debug)]
+enum OtlpMode {
```
vados-cosmonic, line 240: The repetition here is unfortunate... but maybe call this something like `OtlpModeOpt` ?

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+    Otlp,
+    Http,
+}
+
+impl OtlpMode {
```
vados-cosmonic, line 245: `impl From<OltpModeOpt> for OltpMode` ?

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+    fn to_config(&self) -> metrics::OtlpMode {
+        match &self {
+            OtlpMode::Otlp => metrics::OtlpMode::Otlp,
+            OtlpMode::Http => metrics::OtlpMode::Http,
+        }
+    }
+}
+impl std::fmt::Display for OtlpMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let s = match self {
```
vados-cosmonic, line 255: could you get away with delegating harder by calling `to_config()` or `into()` here and then calling `.fmt(f)` ?

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+            Self::Http => "http",
+        };
+        s.fmt(f)
+    }
+}
+
+impl std::str::FromStr for OtlpMode {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
```
vados-cosmonic, line 266: feels like we could delegate harder here

src/main.rs
```diff
@@ -328,6 +432,13 @@ async fn main() -> anyhow::Result<()> {
         enable_structured_logging: args.enable_structured_logging,
         otel_config,
         policy_service_config,
+        enable_metrics: args.enable_metrics,
+        metric_backend: backend(&BackendConfig {
+            backend_type: args.metrics_type,
+            otel_metrics_endpoint: args.otel_metrics_endpoint,
+            otel_mode: args.otel_metrics_mode,
+            prometheus_port: args.metrics_prometheus_port,
+        }),
```
vados-cosmonic, line 441: How do you feel about building the metrics backend a bit more explicitly above this huge list? feels like it's somewhat buried down here.

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+}
+
+struct BackendConfig {
+    backend_type: MetricType,
+    otel_metrics_endpoint: String,
+    otel_mode: OtlpMode,
+    prometheus_port: i32,
+}
+
+fn backend(config: &BackendConfig) -> MetricBackend {
```
vados-cosmonic, line 311: `impl TryFrom<MetricBackend> for BackendConfig` ? Feels like it's possible this could fail in the future, but maybe that's overkill for now and regular `From` works.

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+        match s {
+            "otlp" => Ok(Self::Otlp),
+            "prometheus" => Ok(Self::Prometheus),
+            "debug" => Ok(Self::Debug),
+            _ => Err(format!("unknown metrics type: {s}")),
+        }
+    }
+}
+
+struct BackendConfig {
```
vados-cosmonic, line 304: I guess we never print or compare this but the usual `#[derive(Debug, Clone, PartialEq, Eq)]` would be useful here for debugging.

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+            Self::Debug => "debug",
+        };
+        s.fmt(f)
+    }
+}
+
+impl std::str::FromStr for MetricType {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
```
vados-cosmonic, line 295: could probably delegate then call `from` ?

Also, should we match on lowercased values here so people can put in `PrOmETheUS`?

src/main.rs
```diff
@@ -213,6 +214,109 @@ struct Args {
...
+#[derive(clap::ValueEnum, Clone, Debug)]
+enum MetricType {
+    Otlp,
+    Prometheus,
+    Debug,
+}
+
+impl std::fmt::Display for MetricType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let s = match self {
```
vados-cosmonic, line 283: More delegation, harder!

vados-cosmonic (commented):
Did a bunch of early review and nit-gen

pgray:
going to close for now

Issue #994, opened 2023-11-11
[FEATURE] Replace `env_logger` with `tracing`


vados-cosmonic:
**Is your feature request related to a problem? Please describe.**

`wash` historically used [`env_logger`](https://crates.io/crates/env_logger) for logging, and while that works it would be ideal to use [`tracing`](https://crates.io/crates/tracing) like the rest of the crates in the monorepo

**Describe the solution you'd like**

Convert the `wash` codebase to use `tracing` rather than `env_logger`

**Describe alternatives you've considered**

N/A

**Additional context**

thomastaylor312:
Hi @Shreya111111! @MeenuyD already opened #997 to do this, but we haven't had a response in a few days. If that PR goes stale, then we'd be more than willing to pass it on to you!

vados-cosmonic:
Ah sorry I missed this (thanks @thomastaylor312)  -- fully agreed @Shreya111111 please go ahead and give it a shot

PR #1039, opened 2023-11-19
replacing env_logger with tracing

Shreya111111:
## Feature or Problem
Replacing `env_logger`  with `tracing` to increase the consistency.
#994

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -154,7 +154,7 @@ tokio-stream = { version = "0.1", default-features = false }
 tokio-tar = { version = "0.3", default-features = false }
 tokio-util = { version = "0.7", default-features = false }
 toml = { version = "0.7", default-features = false }
-tracing = { version = "0.1", default-features = false }
+tracing = { version = "0.1.40", default-features = false }
 tracing-futures = { version = "0.2", default-features = false }
 tracing-opentelemetry = { version = "0.20", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
--- a/crates/wash-cli/src/bin/wash.rs
+++ b/crates/wash-cli/src/bin/wash.rs
@@ -220,7 +220,7 @@ enum CliCommand {
 #[tokio::main]
 async fn main() {
     use clap::CommandFactory;
-    env_logger::init();
+    tracing_subscriber::init();
     let cli: Cli = Parser::parse();
 
     let output_kind = cli.output;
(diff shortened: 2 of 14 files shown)


Shreya111111:
@thomastaylor312 can u plz review it.

Cargo.toml
```diff
@@ -154,7 +154,7 @@ tokio-stream = { version = "0.1", default-features = false }
 tokio-tar = { version = "0.3", default-features = false }
 tokio-util = { version = "0.7", default-features = false }
 toml = { version = "0.7", default-features = false }
-tracing = { version = "0.1", default-features = false }
+tracing = { version = "0.1.40", default-features = false }
```
rvolosatovs, line 157:
```suggestion
tracing = { version = "0.1", default-features = false }
```

vendor/capability-providers/blobstore-s3/tests/large_data.rs
```diff
@@ -28,7 +28,7 @@ fn gen_bytes(len: usize) -> (Vec<u8>, u32) {
 #[tokio::test]
 async fn run_all() {
     let opts = TestOptions::default();
-    env_logger::try_init().ok();
+    tracing::try_init().ok();
```
rvolosatovs, line 31: Unfortunately, this won't work - you should use `tracing-subscriber` crate, like e.g. here https://github.com/wasmCloud/wasmCloud/blob/a457cb033cb25f0c29659224db553021244a4298/tests/wasmbus.rs#L462-L469 except we don't need the `,cranelift_codegen=warn,wasmcloud=trace` part

Please make sure to run `cargo test` from the root of this repository to check that it works

Shreya111111:
> Actually, let's not change anything in `vendor`, because that will complicate subtree merges later. We only need this changed in `crates` for now

@rvolosatovs sure, I will look into it

crates/wash-cli/src/bin/wash.rs
```diff
@@ -220,7 +220,7 @@ enum CliCommand {
 #[tokio::main]
 async fn main() {
     use clap::CommandFactory;
-    env_logger::init();
+    tracing_subscriber::init();
```
thomastaylor312, line 223: Just a heads up, the simplest substitution here is `tracing_subscriber::fmt::init();` That will give you basically the same behavior as `env_logger::init()`. We can add more advanced filtering and options later if we wish

brooksmtownsend:
Closed in favor of #1153

Issue #1038, opened 2023-11-18
[BUG] Can't compile wash-cli on arm mac


zshehov:
Running `cargo install wash-cli` results in a compilation error: 

**To Reproduce**
1. Run command `cargo install wash-cli`
2. See error
```wash-lib-0.14.0/src/config.rs:102:49
    |
102 |         let mut builder = CtlClientBuilder::new(nc)
    |                           --------------------- ^^ expected `Client`, found `async_nats::Client`
    |                           |
    |                           arguments to this function are incorrect
    |
    = note: `async_nats::Client` and `Client` have similar names, but are actually distinct types
note: `async_nats::Client` is defined in crate `async_nats`
   --> .../async-nats-0.31.0/src/client.rs:50:1
    |
50  | pub struct Client {
    | ^^^^^^^^^^^^^^^^^
note: `Client` is defined in crate `async_nats`
   --> .../async-nats-0.33.0/src/client.rs:52:1
    |
52  | pub struct Client {
    | ^^^^^^^^^^^^^^^^^
    = note: perhaps two different versions of crate `async_nats` are being used?
```

**Expected behavior**
Successful installation of `wash-cli`

**Environment (please complete the following information) **
 - OS: MacOS Sonoma
 - Shell: zsh

**Additional context**
2 versions of `async_nats` are compiled during the installation:
- async-nats v0.31.0
- async-nats v0.33.0

rvolosatovs:
Hey @zshehov, thank you for report, until #1042 is merged, you should be able to install `wash-cli` via:
```
cargo install wash-cli --git https://github.com/rvolosatovs/wasmcloud --branch fix/update-wadm
```

Or use our binary releases

vados-cosmonic:
Hey @zshehov would you mind posting the output of [`cargo tree`](https://doc.rust-lang.org/cargo/commands/cargo-tree.html) here?

The workaround noted by @rvolosatovs above should work to get you moving (please let us know if it doesn't), but I'm curious exactly which dependency is causing `async_nats` `0.33` to be pulled in.

rvolosatovs:
Actually, let's keep this one open until we have a new release

zshehov:
Hey, thanks for the quick reaction guys, that's awesome!

connorsmith256:
@zshehov This should be fixed now: `cargo install wash-cli` should successfully install v0.24.0. Thanks for the report and your patience!

PR #1043, opened 2023-11-20
fix(ci): make dry-run publish failures visible

connorsmith256:
## Feature or Problem
We've had a couple instances where we fail to publish a crate, or after it's published, users cannot install it, because a downstream dependency had a breaking change. Since the dry-run step is not currently required to merge, it's okay to let it fail, and this should help make failures much more visible

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -539,7 +539,6 @@ jobs:
 
     - name: dry-run publish ${{ matrix.crate }} to crates.io
       if: ${{ !startsWith(github.ref, 'refs/tags/') }}
-      continue-on-error: ${{ matrix.workspace-dependencies }} # publish may fail due to workspace crates not being published yet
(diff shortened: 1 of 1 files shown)


rvolosatovs (changes requested):
I am against this change - this will fail most of the time, because dry-publishing crates dependent on other (not published yet) crates within workspace may fail

rvolosatovs:
Note that this change would do nothing to prevent https://github.com/wasmCloud/wasmCloud/issues/1038

The CI on unrelated PRs could just suddenly start failing, which is not desired

connorsmith256:
Reopening because this once again bit us. https://github.com/wasmCloud/wasmCloud/pull/1381 was able to be merged even though there was a compilation issue in the crate

brooksmtownsend:
Just closing as we don't have a plan of action here unfortunately, not that we don't plan to do something like this. Looking into https://crates.io/crates/cargo-smart-release as well.

PR #1049, opened 2023-11-20
build: bump wash to 0.24.0

rvolosatovs:
## Feature or Problem

Bump wash version for release. Please verify that that's indeed a patch bump.

## Related Issues

#1038

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4835,7 +4835,7 @@ dependencies = [
 
 [[package]]
 name = "wash-cli"
-version = "0.23.0"
+version = "0.24.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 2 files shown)


vados-cosmonic (changes requested):
I don't think this can be a patch release since we have https://github.com/wasmCloud/wasmCloud/commit/ce7904e6f4cc49ca92ec8dee8e263d23da26afd0 in-tree -- there's a breaking change in here so we will have to update to `0.24.0` if that commit is still in here.

vados-cosmonic (approved):
LGTM :rocket:

connorsmith256 (approved):

PR #1051, opened 2023-11-20
release(wash-lib): v0.15.0

connorsmith256:
## Feature or Problem
https://github.com/wasmCloud/wasmCloud/pull/1049 bumped wash-cli, but wash-lib v0.14 still uses async-nats v0.31, so we need a minor bump here as well

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4897,7 +4897,7 @@ dependencies = [
 
 [[package]]
 name = "wash-lib"
-version = "0.14.0"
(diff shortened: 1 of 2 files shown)


rvolosatovs (approved):

brooksmtownsend (approved):
for the record I would be fine publishing v0.14.1 and yanking v0.14 if the only issue was a messed up dependency. I'm good either way and approving just in case I'm missing a breaking change

Issue #860, opened 2023-10-04
[FEATURE] support for registry config in wasmcloud.toml


thomastaylor312:
**Is your feature request related to a problem? Please describe.**

As a user, I would really like it if I didn't have to manually type out my URL to push my actor every time I need to push a new version.

**Describe the solution you'd like**

I would like for wasmcloud.toml to have a `registry` section that lets you configure the registry you wish to push to. `wash push` can then support reading that file if it exists and pushing.

Additionally, it might be really nice for some way to configure registry credentials that could be stored in the `~/.wash` directory so that I don't have to specify the environment variables every time

ahmedtadde:
so, here's one way to handle this

```toml
# when the `url` arg is specified, the `uri` config field is ignored entirely, only the `credentials` field remains possibly useful.
[registry]
# `WASH_REG_URI` always takes precedence over this field.
uri = "http://localhost:8080"
# `credentials` is an optional field; defaults to `.wash/registry_creds.json`. the file path specified is expected to yield a json object { user, password }
# `WASH_REG_USER` and `WASH_REG_PASSWORD` always take precedence over the values derived from this `credentials` field
# `user` and `password` cmd args always take precedence.
credentials = "./creds.json"
```

of course, the `registry` config entry in `wasmcloud.toml` is itself optional.

let me know if this is adequate.

thomastaylor312:
I think I like that setup @ahmedtadde! The only thing I'd recommend is that credentials be tied to URI. So a creds file would look like this:

```json
"localhost:8080": {
    "user": "blah",
    "password": "blah"
}
```

I'd also note that I think it is standard to not preface the URI with `http/s` for OCI registry URL.

I think that would be enough to get us started. As the project grows we might end up hooking into the normal docker credentials or the system toolchain

thomastaylor312:
Oh honestly I forgot that the `actor_config.registry` field existed! I think the best option would be to deprecate that field and have this new `registry` section override it if both exist. Thoughts @brooksmtownsend?

connorsmith256:
It doesn't look like the current `registry` is used anywhere, so we should be safe to remove/move it without deprecating

vados-cosmonic:
Hey @thomastaylor312 since this got merged:

https://github.com/wasmCloud/wasmCloud/pull/1067

[...]

PR #1067, opened 2023-11-21
[FEATURE] support for registry config in wasmcloud.toml 

ahmedtadde:
## Feature or Problem
As a user, I would really like it if I didn't have to manually type out my URL to push my actor every time I need to push a new version.

## Related Issues
#860 

## Release Information
`wash-cli v0.23.0`

## Consumer Impact
better DX

### Acceptance or Integration
[intergration_reg_config](https://github.com/ahmedtadde/wasmCloud/blob/63d95e548e5fa063bcc7c38ae4009c6f85928f72/crates/wash-cli/tests/wash_reg.rs#L248)

### DIFF
--- a/crates/control-interface/src/types.rs
+++ b/crates/control-interface/src/types.rs
@@ -1,6 +1,7 @@
 #![allow(deprecated)]
 use std::collections::HashMap;
 
+use anyhow::bail;
 use serde::{Deserialize, Serialize};
 
 /// One of a potential list of responses to an actor auction
@@ -237,19 +238,24 @@ pub struct RegistryCredential {
     pub registry_type: String,
 }
 
-impl RegistryCredential {
-    #[must_use]
-    pub fn as_oci_registry_auth(&self) -> Option<oci_distribution::secrets::RegistryAuth> {
-        if self.registry_type != "oci" {
-            return None;
+fn default_registry_type() -> String {
+    "oci".to_string()
+}
+
+impl TryFrom<&RegistryCredential> for oci_distribution::secrets::RegistryAuth {
+    type Error = anyhow::Error;
+
+    fn try_from(cred: &RegistryCredential) -> Result<Self, Self::Error> {
+        if cred.registry_type != "oci" {
+            bail!("Only OCI registries are supported at this time");
         }
 
-        match self {
+        match cred {
             RegistryCredential {
                 username: Some(username),
                 password: Some(password),
                 ..
-            } => Some(oci_distribution::secrets::RegistryAuth::Basic(
+            } => Ok(oci_distribution::secrets::RegistryAuth::Basic(
                 username.clone(),
                 password.clone(),
             )),
@@ -259,17 +265,13 @@ impl RegistryCredential {
                 password: None,
                 token: Some(token),
                 ..
-            } => Some(oci_distribution::secrets::RegistryAuth::Basic(
+            } => Ok(oci_distribution::secrets::RegistryAuth::Basic(
                 username.clone(),
                 token.clone(),
             )),
-            _ => None,
+            _ => bail!("Invalid OCI registry credentials"),
         }
     }
-}
-
-fn default_registry_type() -> String {
-    "oci".to_string()
 }
 
 /// A set of credentials to be used for fetching from specific registries
--- a/crates/control-interface/src/types.rs
+++ b/crates/control-interface/src/types.rs
@@ -237,6 +237,37 @@ pub struct RegistryCredential {
     pub registry_type: String,
 }
 
+impl RegistryCredential {
+    #[must_use]
+    pub fn as_oci_registry_auth(&self) -> Option<oci_distribution::secrets::RegistryAuth> {
+        if self.registry_type != "oci" {
+            return None;
+        }
+
+        match self {
+            RegistryCredential {
+                username: Some(username),
+                password: Some(password),
+                ..
+            } => Some(oci_distribution::secrets::RegistryAuth::Basic(
+                username.clone(),
+                password.clone(),
+            )),
+
+            RegistryCredential {
+                username: Some(username),
+                password: None,
+                token: Some(token),
+                ..
+            } => Some(oci_distribution::secrets::RegistryAuth::Basic(
+                username.clone(),
+                token.clone(),
+            )),
+            _ => None,
+        }
+    }
+}
+
 fn default_registry_type() -> String {
     "oci".to_string()
 }
--- a/crates/control-interface/src/types.rs
+++ b/crates/control-interface/src/types.rs
@@ -233,8 +233,12 @@ pub struct RegistryCredential {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub username: Option<String>,
     /// The type of the registry (only "oci" is supported at this time")
-    #[serde(rename = "registryType")]
+    #[serde(rename = "registryType", default = "default_registry_type")]
     pub registry_type: String,
+}
+
+fn default_registry_type() -> String {
+    "oci".to_string()
 }
 
 /// A set of credentials to be used for fetching from specific registries
--- a/crates/wash-cli/src/common/registry_cmd.rs
+++ b/crates/wash-cli/src/common/registry_cmd.rs
@@ -4,6 +4,7 @@ use anyhow::{bail, Context, Result};
 use log::warn;
 use oci_distribution::{
     client::{Client, ClientConfig, ClientProtocol},
+    secrets::RegistryAuth,
     Reference,
 };
 use serde_json::json;
@@ -92,7 +93,7 @@ pub async fn registry_ping(cmd: RegistryPingCommand) -> Result<CommandOutput> {
         _ => resolve_registry_credentials(image.registry()).await,
     }?;
 
-    let Some(credentials) = credentials.as_oci_registry_auth() else {
+    let Ok(credentials) = RegistryAuth::try_from(&credentials) else {
         bail!("failed to resolve registry credentials")
     };
 
--- a/crates/wash-cli/src/common/registry_cmd.rs
+++ b/crates/wash-cli/src/common/registry_cmd.rs
@@ -10,14 +10,19 @@ use oci_distribution::{
 use serde_json::json;
 use tokio::fs::File;
 use tokio::io::AsyncWriteExt;
-use wash_lib::cli::{
-    labels_vec_to_hashmap,
-    registry::{RegistryCommand, RegistryPingCommand, RegistryPullCommand, RegistryPushCommand},
-    CommandOutput, OutputKind,
-};
 use wash_lib::registry::{
     pull_oci_artifact, push_oci_artifact, validate_artifact, OciPullOptions, OciPushOptions,
     SupportedArtifacts,
+};
+use wash_lib::{
+    cli::{
+        labels_vec_to_hashmap,
+        registry::{
+            RegistryCommand, RegistryPingCommand, RegistryPullCommand, RegistryPushCommand,
+        },
+        CommandOutput, OutputKind,
+    },
+    parser::get_config,
 };
 
 use crate::appearance::spinner::Spinner;
@@ -36,13 +41,31 @@ pub async fn registry_pull(
     let spinner = Spinner::new(&output_kind)?;
     spinner.update_spinner_message(format!(" Downloading {} ...", image.whole()));
 
+    let credentials = match (cmd.opts.user, cmd.opts.password) {
+        (Some(user), Some(password)) => Ok(RegistryAuth::Basic(user, password)),
+        _ => match get_config(None, Some(true)) {
+            Ok(project_config) => {
+                project_config
+                    .resolve_registry_credentials(&artifact_url)
+                    .await
+            }
+            Err(_) => Ok(RegistryAuth::Anonymous),
+        },
+    }?;
+
     let artifact = pull_oci_artifact(
         artifact_url,
         OciPullOptions {
             digest: cmd.digest,
             allow_latest: cmd.allow_latest,
-            user: cmd.opts.user,
-            password: cmd.opts.password,
+            user: match &credentials {
+                RegistryAuth::Basic(user, _) => Some(user.to_owned()),
+                _ => None,
+            },
+            password: match &credentials {
+                RegistryAuth::Basic(_, password) => Some(password.to_owned()),
+                _ => None,
+            },
             insecure: cmd.opts.insecure,
         },
     )
@@ -70,11 +93,15 @@ pub async fn registry_ping(cmd: RegistryPingCommand) -> Result<CommandOutput> {
         },
         ..Default::default()
     });
-    let auth = match (cmd.opts.user, cmd.opts.password) {
-        (Some(user), Some(password)) => RegistryAuth::Basic(user, password),
-        _ => RegistryAuth::Anonymous,
-    };
-    let (_, _) = client.pull_manifest(&image, &auth).await?;
+    let credentials = match (cmd.opts.user, cmd.opts.password) {
+        (Some(user), Some(password)) => Ok(RegistryAuth::Basic(user, password)),
+        _ => match get_config(None, Some(true)) {
+            Ok(project_config) => project_config.resolve_registry_credentials(&cmd.url).await,
+            Err(_) => Ok(RegistryAuth::Anonymous),
+        },
+    }?;
+
+    let (_, _) = client.pull_manifest(&image, &credentials).await?;
     Ok(CommandOutput::from("Pong!"))
 }
 
@@ -119,6 +146,18 @@ pub async fn registry_push(
     let spinner = Spinner::new(&output_kind)?;
     spinner.update_spinner_message(format!(" Pushing {} to {} ...", cmd.artifact, artifact_url));
 
+    let credentials = match (cmd.opts.user, cmd.opts.password) {
+        (Some(user), Some(password)) => Ok(RegistryAuth::Basic(user, password)),
+        _ => match get_config(None, Some(true)) {
+            Ok(project_config) => {
+                project_config
+                    .resolve_registry_credentials(&artifact_url)
+                    .await
+            }
+            Err(_) => Ok(RegistryAuth::Anonymous),
+        },
+    }?;
+
     let annotations = labels_vec_to_hashmap(cmd.annotations.unwrap_or_default())?;
 
     push_oci_artifact(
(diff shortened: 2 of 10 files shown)


crates/wash-lib/src/cli/registry.rs
```diff
@@ -41,8 +41,8 @@ pub enum RegistryCommand {
 #[derive(Parser, Debug, Clone)]
 pub struct RegistryPullCommand {
     /// URL of artifact
-    #[clap(name = "url")]
-    pub url: String,
+    #[clap(name = "url", env = "WASH_REG_URL")]
+    pub url: Option<String>,
```
vados-cosmonic, line 45: Hmnnn I think the ENV variable might be a bit weirdly named. This URL is normally the URL of *the artifact*, which happens to be in a registry.  Maybe this should be `WASH_ARTIFACT_URL` ?

As written, I'd think of a registry URL to be `ghcr.io/user`, and an artifact URL to be `ghcr.io/user/project:0.1.0`... does that match with how you're thinking about it?

crates/wash-lib/src/parser/mod.rs
```diff
@@ -218,6 +234,8 @@ pub struct CommonConfig {
     /// Expected name of the wasm module binary that will be generated
     /// (if not present, name is expected to be used as a fallback)
     pub wasm_bin_name: Option<String>,
+    /// optional artifact registry configuration. Primary use case is for the `wash push` & `wash pull` commands
```
vados-cosmonic, line 237:
```suggestion
    /// Optional artifact OCI registry configuration. Primarily used for `wash push` & `wash pull` commands
```

crates/wash-cli/tests/wash_reg.rs
```diff
@@ -232,3 +246,291 @@ fn integration_reg_push_comprehensive() {
 
     remove_dir_all(push_dir).unwrap();
 }
+
+#[tokio::test]
+#[cfg_attr(
+    not(can_reach_wasmcloud_azurecr_io),
+    ignore = "wasmcloud.azurecr.io is not reachable"
+)]
+async fn intergration_reg_config() -> Result<()> {
```
vados-cosmonic, line 255: Thanks for adding to the test suite! These tests look great, but they're a bit focused on lower level bits -- I think they might be better as unit tests closer to the functions that you're testing.

Also, rather than pull a bunch of tests into one, let's split them out into little chunks -- testing precedence, with the field missing, etc can be their own test functions (in the configuration code, rather than here IMO), with their own `///` comments explaining what they do/what they test, if you're open to it.

vados-cosmonic (commented):
Thanks for taking on such a meaty task @ahmedtadde , this is awesome. This is something that will improve the UX of `wash pull`and `wash push` immensely :rocket: .

I found a few minor things and stuff you might want to think about reorganizing, if you wouldn't mind taking a quick look!

crates/wash-lib/src/cli/registry.rs
ahmedtadde: so, the distinction did cross my mind. Though, looking at the related issue (and its discussion) and the related code, there is no apparent distinction between a registry url and artifact url (ie, a single registry providing multiple artifact repositories). if the distinction does matter (and it probably should), then there needs to be more refinement on the issue to address how we want to handle it across all config options (cli, env, wasmcloud file, custom and default credentials file).

crates/wash-cli/tests/common/mod.rs
```diff
@@ -73,6 +73,21 @@ pub fn test_dir_file(subfolder: &str, file: &str) -> PathBuf {
...
+        path.to_string_lossy()
+    ))?;
+
+    file.write_all(content.as_bytes()).await.context(format!(
+        "failed to write content to test file {}",
+        path.to_string_lossy()
+    ))?;
+    Ok(())
+}
+
```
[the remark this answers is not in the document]
ahmedtadde: is there are particular reason why to `tokio` async variant is wanting here?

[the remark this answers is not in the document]
vados-cosmonic: Hey ahmed I think you were trying to note that [`tokio::fs::write`](https://docs.rs/tokio/latest/tokio/fs/fn.write.html) is preferrable, that's totally right, please use that! that is the better option since it's in an `async fn`

crates/wash-lib/src/cli/registry.rs
vados-cosmonic: Hmnnn, so when I look around the codebase I mostly see uses of URLs in the context of artifacts -- the places I do see a "registry url" (like `oci.rs`), It's intended to be the URL of the registry itself (not an individual manifest or artifact). 

What really made me think that the name might be more geared was two things:

- Doc comment above there that mentions artifacts (my first thought was that maybe we need to update that doc comment, but on second thought it seems like an artifact URL)
- How we use [the `url` parameter on `RegistryPushCommand`](https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-cli/src/common/registry_cmd.rs#L33) -- we call it an `artifact_url` there.

I'm willing to be flexible here though, if you feel like `WASH_REG_URL` is the right name for this. It's not like we can't change it later, and I'd love to get your change merged sooner rather than later!

ahmedtadde: on the decision on how to move forward with this, i will defer to @thomastaylor312 since he is the one who originally created the issue #860. the solution i proposed (see discussion on the issue itself) isn't a good fit for having a common configured registry to which multiple artifact can be pushed/pulled.  so, we can either move forward with this for now or this PR is put on ice until an adequate solution is devised.

vados-cosmonic: Let's move forward with this for now! Given that it wasn't used before I think we've got some space -- I'll leave this open just in case Taylor wants to chime in.

vados-cosmonic:
Hey @ahmedtadde there is one conflict, but please let me know when this is ready to re-review and I'll review and merge!

crates/control-interface/src/types.rs
```diff
@@ -224,10 +224,14 @@ pub struct RegistryCredential {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub username: Option<String>,
     /// The type of the registry (only "oci" is supported at this time")
-    #[serde(rename = "registryType")]
+    #[serde(rename = "registryType", default = "default_registry_type")]
```
connorsmith256, line 227: Nit: this could be `default_value` to avoid the need to define a function

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -30,19 +35,54 @@ pub async fn registry_pull(
     cmd: RegistryPullCommand,
     output_kind: OutputKind,
 ) -> Result<CommandOutput> {
-    let artifact_url = cmd.url.to_ascii_lowercase();
-    let image: Reference = artifact_url.parse()?;
+    let artifact_url = match cmd.url {
+        Some(url) => Ok(url),
+        None => {
+            let project_config = get_config(None, Some(true)).map_err(|e| {
```
connorsmith256, line 41: Nit: Instead of `.map_err` where we call `anyhow!`, you should be able to call `.context`, since the error type is already any anyhow error

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -30,19 +35,54 @@ pub async fn registry_pull(
...
+
+            project_config.resolve_registry_url().map_err(|_| {
+                anyhow!(
+                    "No registry URL configured. Please specify a registry URL with the --url flag"
+                )
+            })
+        }
+    }?;
+
+    let image: Reference = artifact_url.to_ascii_lowercase().parse()?;
```
connorsmith256, line 56: Consider adding a `.context` here in case the parse error isn't very friendly

crates/wash-lib/src/parser/mod.rs
```diff
@@ -510,3 +544,63 @@ impl RawProjectConfig {
...
+            .registry
+            .credentials
+            .to_owned()
+            .or_else(|| default_registry_credentials_file().ok());
+
+        if credentials_file.is_none() {
+            return Ok(RegistryAuth::Anonymous);
+        }
+
+        let credentials_file = credentials_file.clone().unwrap();
```
connorsmith256, line 575: Nit: is this clone necessary? And you can avoid the unwrap by combining this with the above block, like so:
```rust
let Some(credentials_file) = credentials_file else {
  return Ok(RegistryAuth::Anonymous);
};
```

You can repeat this structure a couple times below as well

crates/wash-lib/src/parser/mod.rs
```diff
@@ -510,3 +544,63 @@ impl RawProjectConfig {
...
+
+        let credentials = tokio::fs::read_to_string(&credentials_file).await;
+        if credentials.is_err() {
+            return Ok(RegistryAuth::Anonymous);
+        }
+
+        let credentials =
+            serde_json::from_str::<RegistryCredentialMap>(&credentials.unwrap_or_default());
+        if credentials.is_err() {
+            return Ok(RegistryAuth::Anonymous);
```
connorsmith256, line 589: Consider printing/returning an error here, since this case — the file exists but is malformed — sounds like a user error

crates/wash-lib/src/config.rs
```diff
@@ -43,6 +44,10 @@ pub fn downloads_dir() -> Result<PathBuf> {
     Ok(cfg_dir()?.join(DOWNLOADS_DIR))
 }
 
+pub fn default_registry_credentials_file() -> Result<PathBuf> {
+    Ok(cfg_dir()?.join(DEFAULT_REGISTRY_CREDS_FILE))
```
connorsmith256, line 48: Hmm, I think we might want to associate the creds file with a context, instead of having a global one. What do you think?

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -30,19 +35,54 @@ pub async fn registry_pull(
...
 
+    let credentials = match (cmd.opts.user, cmd.opts.password) {
+        (Some(user), Some(password)) => Ok(RegistryAuth::Basic(user, password)),
+        _ => match get_config(None, Some(true)) {
+            Ok(project_config) => {
+                project_config
+                    .resolve_registry_credentials(&artifact_url)
+                    .await
+            }
+            Err(_) => Ok(RegistryAuth::Anonymous),
```
connorsmith256, line 69: When will `get_config` return an error? Would it be better to return that error instead of defaulting to anonymous auth?

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -30,19 +35,54 @@ pub async fn registry_pull(
...
+    }?;
+
     let artifact = pull_oci_artifact(
         artifact_url,
         OciPullOptions {
             digest: cmd.digest,
             allow_latest: cmd.allow_latest,
-            user: cmd.opts.user,
-            password: cmd.opts.password,
+            user: match &credentials {
```
connorsmith256, line 78: Nit: since we're checking the same condition twice, you could assign to `(user, password)` above this block

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -61,7 +101,24 @@ pub async fn registry_pull(
 }
 
 pub async fn registry_ping(cmd: RegistryPingCommand) -> Result<CommandOutput> {
-    let image: Reference = cmd.url.parse()?;
+    let artifact_url = match cmd.url {
```
connorsmith256, line 104: It looks like we have this same logic used 3 times. Can we move it to a function?

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -70,11 +127,19 @@ pub async fn registry_ping(cmd: RegistryPingCommand) -> Result<CommandOutput> {
         },
         ..Default::default()
     });
-    let auth = match (cmd.opts.user, cmd.opts.password) {
-        (Some(user), Some(password)) => RegistryAuth::Basic(user, password),
-        _ => RegistryAuth::Anonymous,
-    };
-    let (_, _) = client.pull_manifest(&image, &auth).await?;
+    let credentials = match (cmd.opts.user, cmd.opts.password) {
```
connorsmith256, line 130: Same question here: since this is used 3 times, can we move it to a function?

crates/wash-lib/src/cli/registry.rs
```diff
@@ -63,11 +63,11 @@ pub struct RegistryPullCommand {
...
 pub struct RegistryPushCommand {
     /// URL to push artifact to
-    #[clap(name = "url")]
-    pub url: String,
+    #[clap(name = "url", env = "WASH_REG_URL")]
+    pub url: Option<String>,
 
     /// Path to artifact to push
-    #[clap(name = "artifact")]
+    #[clap(long = "artifact")]
```
connorsmith256, line 70: Since `artifact` is always required, doesn't it make more sense to keep this as a positional argument?

crates/wash-lib/src/cli/registry.rs
connorsmith256: Taylor is out for the next week, but I'll comment on his behalf. Up to now, this url was meant to be a specific artifact URL, e.g. `wasmcloud.azurecr.io/echo:0.3.8`, not just a registry url (e.g. `wasmcloud.azurecr.io`). 

Currently, I can do:

```
wash reg pull/push wasmcloud.azurecr.io/echo:0.3.8
```

With your changes, we allow users to set `registry` config in the project file. So that _might_ allow usage like:
```
wash reg pull/push echo:0.3.8
```

However, since the registry config is optional, we won't always know what the registry URL is. Another option is to change the command structure to be something like this:
```
wash reg pull/push --registry wasmcloud.azurecr.io echo:0.3.8
```

And then in cases where I've set up `registry` config in my `wasmcloud.toml`, I could omit the flag and just do:

```
wash reg pull/push echo:0.3.8
```

crates/wash-lib/src/parser/mod.rs
```diff
@@ -510,3 +544,63 @@ impl RawProjectConfig {
...
+                bail!("No registry URL specified in wasmcloud.toml");
+            }
+        };
+
+        Ok(registry_url)
+    }
+
+    pub async fn resolve_registry_credentials(
+        &self,
+        registry: impl AsRef<str>,
```
connorsmith256, line 562: Should this be `registry`, or `artifact_url`? I noticed all the uses of this function pass a full artifact URL, not just the name of the registry

connorsmith256 (changes requested):

crates/control-interface/src/types.rs
ahmedtadde: i think i tried to use `default_value` but i got an error. i know it works with `clap` but couldn't get it to work with `serde`. can you please paste in the annotation you have in mind

connorsmith256: Ah, I was thinking of clap, not serde. Disregard this 🙂

ahmedtadde:
hey @connorsmith256 

so assuming that a registry url is configured (via `WASH_REG_URL` or `wasmcloud.toml`)... how do we deterministically handle

(1) `wash reg pull/push wasmcloud.azurecr.io/echo:0.3.8` vs (2) `wash reg pull/push echo:0.3.8`

That is, how do we distinguish between a full artifact url (1) and a repository path (2) that is supposed to be prepended by a registry url.  I am thinking using `oci_distribution:Reference::try_from()` could work? if it returns an error, then we assume the input value is meant to be just the repository path instead of the full artifact url.

assuming that the above is handled... the current `push/pull` commands would basically stay the same except that we add an additional optional arg `--registry` which fallbacks to `WASH_REG_URL` or `project_config.common.registry.url` when available. if the registry url fallback values are not available, we return an error to the user. 

now, for the credentials, does the credentials file `project_config.common.registry.credentials` take precedence over the current context's credentials file? or vice versa? either way, we only return an error when the credentials file is specified, exists and parsing it from `json` into `RegistryCredentialMap` fails.

connorsmith256:
@ahmedtadde here's my proposal, though I'd want to hear from others, including you, since this is a breaking change to wash, no matter which approach we take:
- `--registry` becomes an optional argument to `push/pull`, instead of being a positional arg
  - (or `--registry-url`? `-r` for shorthand would be nice too)
  - for reference, we use `--oci-registry`/`OCI_REGISTRY` for the host 
- in terms of precedence, we should check in this order, which favors explicit user intent over implicit config:
  - flag
  - env var
  - `wasmcloud.toml`
- if no registry URL is provided at all, I agree, we should print an error
- for credentials, the precedence should follow the same order, for consistency. I don't think we need a default registry credsfile. If you think we should have one, I would prefer it to be per-context, instead of global
- **BONUS**: the full artifact URL could be inferred, from a combination of the registry URL + name + version. The artifact name and version could be provided the same ways as the registry URL. That would enable `wash reg push` without any extra details, for the most common workflow, and commands like `wash reg push -v 0.3.8-rc.1` to override a specific parameter
  - this could be done in subsequent PR(s)

crates/control-interface/src/types.rs
```diff
@@ -233,10 +233,45 @@ pub struct RegistryCredential {
...
     pub registry_type: String,
 }
 
+impl RegistryCredential {
+    #[must_use]
+    pub fn as_oci_registry_auth(&self) -> Option<oci_distribution::secrets::RegistryAuth> {
```
connorsmith256, line 242: Nit: I think this could be accomplished via a `TryFrom`

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -162,6 +194,58 @@ pub async fn handle_command(
...
 }
 
+fn resolve_artifact_ref(url: &str, registry: &str) -> Result<Reference> {
+    let image: Reference = url
+        .trim()
+        .to_ascii_lowercase()
+        .parse()
+        .context("failed to parse artifact url into oci image reference")?;
+
+    if url.trim() == image.whole() {
```
connorsmith256, line 204: I'm not very familiar with the `oci_distribution` crate. When will this not be true?

crates/wash-cli/tests/wash_reg.rs
```diff
@@ -232,3 +238,170 @@ fn integration_reg_push_comprehensive() {
...
     remove_dir_all(push_dir).unwrap();
 }
+
+// NOTE: This test will fail without a local docker registry running
+#[tokio::test]
+#[cfg_attr(
+    not(can_reach_wasmcloud_azurecr_io),
+    ignore = "wasmcloud.azurecr.io is not reachable"
+)]
+async fn intergration_reg_config() -> Result<()> {
```
connorsmith256, line 248: Nit: typo in integration

connorsmith256 (approved):
Sorry for the delay @ahmedtadde, I've had a lot on my plate

LGTM! Just a few optional nits. @vados-cosmonic could you handle getting this merged?

vados-cosmonic:
Sure @connorsmith256 ! Hey @ahmedtadde let me know if you want me to merge this PR immediately, or if you want to address those nits -- I'm fine either way as we can always come back and make the small changes later -- thanks again for the awesome contribution!

crates/wash-cli/src/common/registry_cmd.rs
ahmedtadde: based on our use case and the implementation of `image.whole()` from the `oci_distribution` crate, that predicate will be false when the user provides a repository path rather than a full artifact url.

ahmedtadde: will mark this as resolved but you follow up on this if you think there should be more deliberation on this

ahmedtadde:
hey @vados-cosmonic , i pushed a commit to address the nits from @connorsmith256. from my end, PR should be good to merge at this point.

Issue #1092, opened 2023-11-25
[FEATURE] Replace actor start/stop events with actor_scaled event


connorsmith256:
**Is your feature request related to a problem? Please describe.**
There are currently 4 events related to scaling an actor:
- `actor_started`
- `actor_stopped`
- `actors_started`
- `actors_stopped`

**Describe the solution you'd like**
I propose we use a single event to capture all of these: `actor_scaled`

**Describe alternatives you've considered**
N/A

**Additional context**
I'm not sure how this impacts wadm, so would appreciate input from @brooksmtownsend @thomastaylor312

MeenuyD:
Hello @connorsmith256 I would like to work on this issue can you please assign it to me I have a question should we have to replace all the `actors_started` and `actors_stopped` to `actors_scaled`

brooksmtownsend:
w.r.t. wadm, it would be nice if we introduced the new `actor_scaled` event and deprecate the others so that we can work forwards with backwards compatibility for a release or two. 

Ideally:
1. Add `actor_scaled` to the list of events, emitted whenever an actor is started, stopped, or scaled. If there's a cloudevents way to mark an event deprecated (TBD) we should do that too.
2. Note in that wasmCloud release that the other events are deprecated.
3. Update wadm and wash to expect the `actor_scaled` event.
4. Remove old events

Question: How do we treat the idea of scaling an actor with no maximum concurrency? Does an `actor_scaled` event with a max of 0 mean infinite, or that it stopped?
Strawman answer: Scaling to 0 means that the actor is no longer running. Actually scaling with a maximum level of concurrency should be supported with a big number, e.g. if the scale is a u16 we should represent the max as `65536`.

brooksmtownsend:
#1107 will remove `actor_started` and `actor_stopped` and `actor_start_failed` in favor of `actor_scaled`, but leaves the plural events for now so we can make the compatible changes in wadm, the lattice observer, wash, etc.

brooksmtownsend:
Closed in #1107

PR #1107, opened 2023-11-28
feat(control-interface)!: rename max-concurrent to max-instances, simplify scale

brooksmtownsend:
## Feature or Problem
After a month or so of working with the changes from #696, I've come to a few conclusions about the developer experience of the new scaling model. All the thoughts below are up for discussion I just wanted to propose the changes.

1. This number `--max-concurrent` is describing "the maximum number of concurrent instances allowed to execute" and and I slightly prefer `--max-instances` to describe this number. This is also a more intuitive number when describing capability provider instances, and the eventual WASI provider instances, and will be better to replace the `replicas` field in wadm manifests. This might just be `instances`, but that can be determined later.
2. The difference between scaling to `None`, `Some(0)`, and `Some(10000)` is still confusing, and "unbounded" is a scary number.
3. It's a little weird to work with the `Option<NonZeroUsize>` type all over the place.

In order to address these issues, I made the following changes
1. Renamed from `--max-concurrent` / `max_concurrent` to `--max-instances` / `max_instances` around the codebase.
2. The maximum instances parameter is no longer an optional number, just a u32. In developer facing clients we'll default to `u16::MAX` and it will be easy to see that number.
    a. ~If we run into an issue in the future with an actor needing to scale past 65535 concurrent running instances you are welcome to @ me and I'll update everything to be u32s 😉~ It's u32 now 😉 
3. All max scale numbers are now `NonZeroUsize`s. I considered refactoring this to a `NonZeroU32` as well, but felt that was an unnecessary internal detail to correct before discussion.

## Related Issues
Addresses removing the singular events and adding `actor_scaled` re: #1092 

## Release Information
No rush on releasing these so they can go out with our next regular release.

control-interface v0.33.0
wash-lib v0.16.0
wash-cli v0.26.0
wasmCloud v0.82.0

## Consumer Impact
- control-interface has a few `Option<u16>` to `u32` changes, which should be simple and anyone previously passing `None` can simply pass `u32::MAX` to preserve the same behavior
- wash-lib has a few `Option<u16>` to `u32` changes, same as control-interface
- wash-cli won't have any breaking changes, only a name change in the flag (which has a backwards compatible alias)
- wasmCloud users will need to update their `wash-cli` as omitting the count will no longer scale to unbounded, it will stop the actor.

### Unit Test(s)
Modified tests to use `u32`s instead of `Option<u16>`s

### DIFF
--- a/crates/host/src/wasmbus/event.rs
+++ b/crates/host/src/wasmbus/event.rs
@@ -36,41 +36,7 @@ fn format_actor_claims(claims: &jwt::Claims<jwt::Actor>) -> serde_json::Value {
     }
 }
 
-pub fn actor_started(
-    claims: &jwt::Claims<jwt::Actor>,
-    annotations: &BTreeMap<String, String>,
-    instance_id: Uuid,
-    image_ref: impl AsRef<str>,
-) -> serde_json::Value {
-    json!({
-        "public_key": claims.subject,
-        "image_ref": image_ref.as_ref(),
-        "api_version": "n/a",
-        "instance_id": instance_id,
-        "annotations": annotations,
-        "claims": format_actor_claims(claims),
-    })
-}
-
-pub fn actor_start_failed(actor_ref: impl AsRef<str>, error: &anyhow::Error) -> serde_json::Value {
-    json!({
-        "actor_ref": actor_ref.as_ref(),
-        "error": format!("{error:#}"),
-    })
-}
-
-pub fn actor_stopped(
-    claims: &jwt::Claims<jwt::Actor>,
-    annotations: &BTreeMap<String, String>,
-    instance_id: Uuid,
-) -> serde_json::Value {
-    json!({
-        "public_key": claims.subject,
-        "instance_id": instance_id,
-        "annotations": annotations,
-    })
-}
-
+// TODO(#1092): Remove this event in favor of `actor_scaled`
 pub fn actors_started(
     claims: &jwt::Claims<jwt::Actor>,
     annotations: &BTreeMap<String, String>,
@@ -88,6 +54,7 @@ pub fn actors_started(
     })
 }
 
+// TODO(#1092): Remove this event in favor of `actor_scaled`
 pub fn actors_start_failed(
     claims: &jwt::Claims<jwt::Actor>,
     annotations: &BTreeMap<String, String>,
@@ -104,19 +71,56 @@ pub fn actors_start_failed(
     })
 }
 
+// TODO(#1092): Remove this event in favor of `actor_scaled`
 pub fn actors_stopped(
     claims: &jwt::Claims<jwt::Actor>,
     annotations: &BTreeMap<String, String>,
     host_id: impl AsRef<str>,
     count: NonZeroUsize,
     remaining: usize,
+    image_ref: impl AsRef<str>,
 ) -> serde_json::Value {
     json!({
-        "host_id": host_id.as_ref(),
         "public_key": claims.subject,
+        "annotations": annotations,
+        "host_id": host_id.as_ref(),
         "count": count,
         "remaining": remaining,
+        "image_ref": image_ref.as_ref(),
+    })
+}
+
+pub fn actor_scaled(
+    claims: &jwt::Claims<jwt::Actor>,
+    annotations: &BTreeMap<String, String>,
+    host_id: impl AsRef<str>,
+    max_instances: NonZeroUsize,
+    image_ref: impl AsRef<str>,
+) -> serde_json::Value {
+    json!({
+        "public_key": claims.subject,
         "annotations": annotations,
+        "host_id": host_id.as_ref(),
+        "image_ref": image_ref.as_ref(),
+        "max_instances": max_instances,
+    })
+}
+
+pub fn actor_scale_failed(
+    claims: &jwt::Claims<jwt::Actor>,
+    annotations: &BTreeMap<String, String>,
+    host_id: impl AsRef<str>,
+    image_ref: impl AsRef<str>,
+    max_instances: NonZeroUsize,
+    error: &anyhow::Error,
+) -> serde_json::Value {
+    json!({
+        "public_key": claims.subject,
+        "annotations": annotations,
+        "host_id": host_id.as_ref(),
+        "image_ref": image_ref.as_ref(),
+        "max_instances": max_instances,
+        "error": format!("{error:#}"),
     })
 }
 
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -261,7 +261,8 @@ struct ActorInstance {
     handler: Handler,
     chunk_endpoint: ChunkEndpoint,
     annotations: Annotations,
-    max: NonZeroUsize,
+    /// Maximum number of instances of this actor that can be running at once
+    max_instances: NonZeroUsize,
     /// Cluster issuers that this actor should accept invocations from
     valid_issuers: Vec<String>,
     policy_manager: Arc<PolicyManager>,
@@ -1597,8 +1598,7 @@ pub struct Host {
     queue: AbortHandle,
     aliases: Arc<RwLock<HashMap<String, WasmCloudEntity>>>,
     links: RwLock<HashMap<String, LinkDefinition>>,
-    actor_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::Actor>>>>,
-    // TODO: use a single map once Claims is an enum
+    actor_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::Actor>>>>, // TODO: use a single map once Claims is an enum
     provider_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::CapabilityProvider>>>>,
     config_data_cache: Arc<RwLock<ConfigCache>>,
 }
@@ -2314,7 +2314,7 @@ impl Host {
                 let instances = actor.instances.read().await;
                 let max = instances
                     .iter()
-                    .map(|(_annotations, instance)| instance.max.get())
+                    .map(|(_annotations, instance)| instance.max_instances.get())
                     .sum();
                 if max == 0 {
                     None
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -167,15 +167,15 @@ warp = { version = "0.3", default-features = false }
 warp-embed = { version = "0.4", default-features = false }
 wascap = { version = "0.12", path = "./crates/wascap", default-features = false }
 wash-cli = { version = "0", path = "./crates/wash-cli", default-features = false }
-wash-lib = { version = "0.15", path = "./crates/wash-lib", default-features = false }
+wash-lib = { version = "0.16", path = "./crates/wash-lib", default-features = false }
```
connorsmith256, line 170: Are we ready to release these dependent crates? I ask because e.g. if we needed to do a patch bump of wash-cli, and we've changed the versions here but haven't published them, then publishing wash-cli would fail, right?

crates/host/src/wasmbus/mod.rs
```diff
@@ -70,6 +58,17 @@ use wasmcloud_runtime::capability::{
 use wasmcloud_runtime::Runtime;
 use wasmcloud_tracing::context::TraceContextInjector;
 
+use crate::{
```
connorsmith256, line 61: Why the reorder here?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3928,7 +3917,7 @@ impl Host {
         Ok(buf.into())
     }
 
-    #[instrument(level = "trace", skip_all, fields(subject = %message.subject))]
+    #[instrument(level = "trace", skip_all, fields(subject = % message.subject))]
```
connorsmith256, line 3920: oops?

crates/host/src/wasmbus/mod.rs
```diff
@@ -4308,7 +4297,7 @@ impl Host {
         }
     }
 
-    #[instrument(level = "trace", skip_all, fields(bucket = %entry.bucket, key = %entry.key, revision = %entry.revision, operation = ?entry.operation))]
+    #[instrument(level = "trace", skip_all, fields(bucket = % entry.bucket, key = % entry.key, revision = % entry.revision, operation = ? entry.operation))]
```
connorsmith256, line 4300: oops?

connorsmith256 (commented):
LGTM, but a few questions about possibly-unintended changes

Also, for consideration: 

> If we run into an issue in the future with an actor needing to scale past 65535 concurrent running instances you are welcome to @ me and I'll update everything to be u32s 😉

wanna just make everything a u32 now? 🙂

crates/wash-lib/Cargo.toml
```diff
@@ -29,10 +29,10 @@ cli = [
     "path-absolutize",
 ]
 nats = ["async-nats", "wadm"]
-build-adapters = [ "wasmcloud-component-adapters/build-adapters" ]
+build-adapters = ["wasmcloud-component-adapters/build-adapters"]
 
 [package.metadata.docs.rs]
-no-default-features = true # skip building adapters (which requires internet access)
+no-default-features = true             # skip building adapters (which requires internet access)
```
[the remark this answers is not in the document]
brooksmtownsend: I didn't do this manually it was just my editor auto format, unfortunately I'm having trouble finding the setting that caused this, will try and correct. Definitely don't want inconsistent styling

crates/host/src/wasmbus/mod.rs
brooksmtownsend: It was automatic, maybe new rust-analyzer? I can move back but this seems closer in line with the convention we use, being
1. core
2. std
3. external
4. internal

Cargo.toml
brooksmtownsend: Very true. I don't think we need to release them yet, but I do think that we need to update the versions in these crates because there's new functionality. Put another way, if the newly updated code for `wash-cli` tries to compile against `wash-lib 0.15`, it will fail.

With this method of development in the monorepo, I think we have to bump crates when making breaking changes or pay attention to the conventional commits leading up to a release. This change going in means wash-cli has to bump a minor version, so we can't publish a patch anyways unless we cherry pick this commit out. Does that make sense? I want to make sure I'm thinking about this the right way since we should all take the same actions when working on changes that affect multiple libraries in this repo.

crates/host/src/wasmbus/mod.rs
connorsmith256: The order I saw elsewhere was
1. internal (`crate`)
2. core
3. std
4. external

FWIW I prefer the order you have now. I wish we had consistent tooling to do this automatically so I could stop thinking about it 😅

vados-cosmonic: Just a data point but I personally prefer the order that brooks mentioned:

core 
std
external
Internal

brooksmtownsend:
Just a note that I feel I've addressed the comments here, and I found out why the test is failing 😬 For backwards compatibility with previous events we publish an `actor_started` event for each actor scaled up. Now, by default, the behavior is scaling to `u32::MAX` with wash and the test, and so 4,294,967,296 `actor_started` events get published.

I think I'll wait to finish this one up until #1092 is done, which would be a nice change and resolve this issue. Alternatively I could go ahead and remove the `actor_started` event here, but that's half the work.

brooksmtownsend:
Added another commit which adds the `actor_scaled` event and removes the singular actor events. This was necessary as the default behavior in this PR with wash is to scale to `u32::MAX` actors, which would publish `u32::MAX` `actor_started` events 😅

crates/host/src/wasmbus/mod.rs
```diff
@@ -2380,7 +2381,7 @@ impl Host {
         claims: &jwt::Claims<jwt::Actor>,
         annotations: &Annotations,
         actor_ref: impl AsRef<str>,
-        max: Option<NonZeroUsize>,
+        max: NonZeroUsize,
```
vados-cosmonic, line 2384: I guess there's no use case for forcing actor instantiation but having no max? I might have thought that would be a case for some sort of "preloading"

Also, did we miss `max_instances` rename here?

vados-cosmonic (approved):
LGTM :rocket:

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Yeah this was just shorthand, I can go refactor for clarity, no need to have ambiguity.

What I was thinking is that if we forced instantiation with no max, it would only ever be different if there were more than 2^32 concurrently executing instances of a single component. The benefit of getting rid of the option is clarity, no need to speculate between the difference of `None` and `Some(0)`

vados-cosmonic: ah OK -- yeah that's definitely a simplification, and I guess if we ever want to have some sort of preload operation we can add one explicitly

crates/host/src/wasmbus/event.rs
```diff
@@ -36,87 +36,91 @@ fn format_actor_claims(claims: &jwt::Claims<jwt::Actor>) -> serde_json::Value {
...
     claims: &jwt::Claims<jwt::Actor>,
     annotations: &BTreeMap<String, String>,
-    instance_id: Uuid,
+    host_id: impl AsRef<str>,
+    image_ref: impl AsRef<str>,
+    error: &anyhow::Error,
 ) -> serde_json::Value {
     json!({
         "public_key": claims.subject,
-        "instance_id": instance_id,
```
connorsmith256, line 69: Are any clients depending on this?

crates/wash-lib/Cargo.toml
```diff
@@ -29,7 +29,7 @@ cli = [
     "path-absolutize",
 ]
 nats = ["async-nats", "wadm"]
-build-adapters = [ "wasmcloud-component-adapters/build-adapters" ]
+build-adapters = ["wasmcloud-component-adapters/build-adapters"]
```
connorsmith256, line 32: I'd try to avoid changing this line since https://github.com/wasmCloud/wasmCloud/pull/1146 touches it too

connorsmith256 (approved):

crates/host/src/wasmbus/event.rs
brooksmtownsend: The diff here is actually really misleading, I only removed `instance_id` from actor_started and actor_stopped which no longer exist.

crates/wash-lib/Cargo.toml
brooksmtownsend: Gah, I don't even have a formatter installed anymore, how does this happen

crates/wash-cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.25.0"
+version = "0.24.1"
```
connorsmith256, line 3: whoops

crates/wash-lib/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.16.0"
+version = "0.15.0"
```
connorsmith256, line 3: whoops

crates/host/src/wasmbus/mod.rs
```diff
@@ -1597,7 +1598,8 @@ pub struct Host {
     queue: AbortHandle,
     aliases: Arc<RwLock<HashMap<String, WasmCloudEntity>>>,
     links: RwLock<HashMap<String, LinkDefinition>>,
-    actor_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::Actor>>>>, // TODO: use a single map once Claims is an enum
+    actor_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::Actor>>>>,
+    // TODO: use a single map once Claims is an enum
```
connorsmith256, line 1602: Odd that this got shoved between the two lines

crates/host/src/wasmbus/mod.rs
```diff
@@ -4264,78 +4260,86 @@ impl Host {
         }
     }
 
-    /// Publishes `count` `actor_started` events and one `actors_started` event with the supplied count
-    async fn publish_actor_started_events(
+    // TODO(#1092): only publish actor_scale_failed event after wasmCloud releases 0.81
```
connorsmith256, line 4263: Now that 0.81 is out, can we resolve these TODOs?

connorsmith256 (changes requested):
Just need to fix a couple version decrements

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I actually have to change these to 0.82 😢 The idea was to get these events published but not remove the old ones until a minor bump is done

connorsmith256 (approved):

Issue #1011, opened 2023-11-14
[BUG] call alias clash detection prevents republished actors from using a previous alias


connorsmith256:
**Describe the bug**
If two claims entries attempt to use the same call alias, the host will reject the latter, with a message like:
```
5908:{"timestamp":"2023-11-01T13:06:14.850993Z","level":"ERROR","fields":{"message":"failed to scale actor","actor_ref":"ghcr.io/connorsmith256/inferenceapi:0.2.0","err":"failed to store claims\n\nCaused by:\n    call alias `inferenceapi` clash between `MALZANXA36ZWPZFFDWQZUQVMH3LOE3BD3WUU5ZBSNJ5PFBXBJT4V6ZQQ` and `MA7ZIVAJMZSRVVKMFQQNSH4WQWKGE25MEUY7XRMGEHD3HCK5OAKWNJZH`"},"target":"wasmcloud_host::wasmbus"}
```

This can happen when an actor was built locally that uses a call alias, and then in the same lattice, you attempt to start a published version of the actor, which uses the same alias

**To Reproduce**
1. Start an actor that uses a call alias
1. Re-build or publish the actor (simplest to clear the `.nk` file in `~/.wash/keys`, so you have an artifact that uses the same alias but with a different public key
1. Attempt to start the new actor
2. Notice the host rejects the attempt due to the call alias already being in the lattice cache

**Expected behavior**
While re-using a call alias can be a problem sometimes, rejecting it outright leads to unexpected behavior, and the user has no way forward

Instead, I propose moving the error to a `warn`-level message, and the host should overwrite the previous alias. This scenario shouldn't happen often

**Screenshots**

**Environment (please complete the following information) **
 - OS: [e.g. Linux, MacOS]
 - Shell [e.g. bash, zsh, powershell]
 - wasmcloud Version [e.g. 0.78.0]

**Additional context**

PR #1116, opened 2023-11-29
fix(host): override previous call alias on clash

connorsmith256:
## Feature or Problem
Update the host to override previously cached call aliases when an alias is re-used. The host will emit a warning, since this can result in unintuitive behavior

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/1011

## Release Information
Next

### Manual Verification
Built two copies of the ponger actor, ponger1 and ponger2:
```
wash start actor file://$(realpath pinger/build/pinger_s.wasm)
wash start provider wasmcloud.azurecr.io/httpserver:0.19.1
wash link put MB2SLLXFTWEX7HR6F6PTLPWSU35MJKHR6CQHFN4QCWEIVBVJA2CFXHLY VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M wasmcloud:httpserver address=0.0.0.0:1234
```

```
wash start actor file://$(realpath ponger1/build/ponger_s.wasm)

curl 127.0.0.1:1234
Ping pong1%

wash start actor file://$(realpath ponger2/build/ponger_s.wasm)

curl 127.0.0.1:1234
Ping pong2%
```

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2668,14 +2668,17 @@ impl Host {
         {
             let mut aliases = self.aliases.write().await;
             match aliases.entry(call_alias) {
-                Entry::Occupied(entry) => {
-                    ensure!(
-                        entry.get().public_key == claims.subject,
-                        "call alias `{}` clash between `{}` and `{}`",
-                        entry.key(),
-                        entry.get().public_key,
-                        claims.subject
+                Entry::Occupied(mut entry) => {
+                    warn!(
+                        alias = entry.key(),
+                        existing_public_key = entry.get().public_key,
+                        new_public_key = claims.subject,
+                        "call alias clash. Replacing existing entry"
                     );
+                    entry.insert(WasmCloudEntity {
+                        public_key: claims.subject.clone(),
+                        ..Default::default()
+                    });
                 }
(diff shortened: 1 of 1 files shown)


rvolosatovs (approved):
LGTM, but I'm confused on the purpose of `call_alias` being a part of the claims in the first place, especially if it can (now) be freely overridden.

Shouldn't `call_alias` simply be a runtime config?

connorsmith256:
@rvolosatovs in all "typical" scenarios, I don't think call aliases should be overridden. They're defined at signing time, and usually specified in the `wasmcloud.toml` configuration. Call aliases are part of the architecture of an app, and they're a convenience over making an actor-to-actor call via public-key, since a2a calls are specified in code, and you don't have the actor's ID available prior to signing it. Most actors won't be re-signed in their lifetime, and most actors won't have an alias. This fix is meant for the edge case where this does inadvertently happen. One "legitimate" scenario is:
- work on actor Foo locally, which has a call alias
- start actor Foo on your lattice
- do some e2e testing
- publish actor Foo in CI, so it now has a different public key
- start Foo on your lattice again
- get a clash on alias, since the old Foo's claims are still in the lattice

Issue #740, opened 2023-10-10
[FEAT] Remove deprecated control interface functionality


brooksmtownsend:
In #710 we introduced a few clippy warnings about deprecated control interface functionality in order to support the start actor command in a backwards compatible way. As of the control interface 0.29.0 and applicable wadm/wash versions (TBD), this command isn't used anywhere and can be fully removed instead of just deprecated.

PR #1144, opened 2023-12-06
refactor: remove deprecated code related to start actor cmd

ahmedtadde:
## Feature or Problem
In #710 we introduced a few clippy warnings about deprecated control interface functionality in order to support the start actor command in a backwards compatible way. As of the control interface 0.29.0 and applicable wadm/wash versions (TBD), this command isn't used anywhere and can be fully removed instead of just deprecated.

## Related Issues
#740 

## Release Information
`wasmCloud v0.80` & `wash-cli v0.24.1`

## Consumer Impact
there should be none

## Testing
all unit tests and integration tests pass

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2953,76 +2953,6 @@ impl Host {
     }
 
     #[instrument(level = "debug", skip_all)]
-    async fn handle_launch_actor(
-        self: Arc<Self>,
-        payload: impl AsRef<[u8]>,
-        host_id: &str,
-    ) -> anyhow::Result<Bytes> {
-        // Explicitly allowing the deprecated command so we can offer backwards compatibility
-        // TODO(#740): Remove deprecated once control clients no longer use this command
-        #[allow(deprecated)]
-        let wasmcloud_control_interface::StartActorCommand {
-            actor_ref,
-            annotations,
-            count,
-            ..
-        } = serde_json::from_slice(payload.as_ref())
-            .context("failed to deserialize actor launch command")?;
-
-        warn!("received deprecated start_actor command which will be removed in a future version");
-
-        let host_id = host_id.to_string();
-        spawn(async move {
-            let actor = self.fetch_actor(&actor_ref).await?;
-            let claims = actor.claims().context("claims missing")?;
-            let actor_id = claims.subject.clone();
-
-            let annotations = annotations.unwrap_or_default().into_iter().collect();
-
-            let max = match self.actors.write().await.entry(actor_id) {
-                hash_map::Entry::Vacant(_) => Some(count),
-                hash_map::Entry::Occupied(entry) => {
-                    let actor = entry.get();
-                    let instances = actor.instances.read().await;
-                    if let Some(instance) = matching_instance(&instances, &annotations) {
-                        if instance.image_reference != actor_ref {
-                            let err = anyhow!(
-                                "actor is already running with a different image reference `{}`",
-                                instance.image_reference
-                            );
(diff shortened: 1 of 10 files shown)


thomastaylor312 (approved):
Thanks again @ahmedtadde this is looking good! I think everything is good to go, but because this is a breaking removal, I'd also like at least @brooksmtownsend or @connorsmith256 to take a look and approve as well!

Whichever of you get to it first can go ahead and merge if you also approve

crates/wash-lib/src/cli/dev.rs
```diff
@@ -42,16 +41,8 @@ pub async fn run_dev_loop(
...
-                &host_id,
-                &actor_id,
-                None,
-                default_timeout_ms(),
-                false,
-            )
-            .await?;
+
+            update_actor(ctl_client, &host_id, &actor_id, actor_ref).await?;
             start_actor(StartActorArgs {
```
connorsmith256, line 46: Now that we're using `update_actor`, shouldn't we remove this?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3953,10 +3883,6 @@ impl Host {
             (Some("auction"), Some("provider"), None, None) => {
                 self.handle_auction_provider(message.payload).await
             }
-            (Some("cmd"), Some(host_id), Some("la"), None) => Arc::clone(&self)
```
connorsmith256, line 3956: I think we should leave support in for this for at least one version. Consider a user who is interacting with a lattice where they've upgraded their host versions, but are using v0.23 of wash: wash will still be issuing `.la` requests to the control interface, but the host(s) won't be able to handle them

So I think this PR should remove all uses of start/launch from wash, and then after a new version of wash has been out for a while, we can remove support in the host

This follows a general safe pattern of deprecating and removing functionality from a server:
1. mark as deprecated in API
2. remove all uses from all clients
3. have reasonable knowledge all clients are updated
4. remove support from the server

I think we're jumping from 1) to 4) here

connorsmith256 (dismissed):
Correct me if I'm wrong, but I think this needs to be a multi-step process (see comment)

crates/host/src/wasmbus/mod.rs
thomastaylor312: As far as I understood here, this PR does remove use from all clients (I know we aren't using it in wadm and everything else is in this repo now) and they had been deprecated for a few versions now. However, I understand the concern with keeping support around for older wash versions. I'm honestly ok with cutting it at this point, but if we want to be extra careful, we can keep it for now. The other option is to remove and just return an error that says this endpoint has been removed

brooksmtownsend: I need to double check on this, but I'm fairly sure when I cut control interface 0.29 it actually proxies to `scale_actor` from within the control interface. So it's been removed from clients for quite a while https://docs.rs/wasmcloud-control-interface/0.30.0/src/wasmcloud_control_interface/lib.rs.html#309

connorsmith256: @thomastaylor312 my point is we cannot remove from all clients in the same release tuple (clients, server) as we drop support from the server

But it sounds like @brooksmtownsend covered this, and we _actually_ removed usage from all clients a version ago 👀 if that's the case, then 2) and 3) are already done, and I'm good with this PR moving to step 4) 👍

brooksmtownsend (approved):

thomastaylor312:
@ahmedtadde looks like the only CI failure is from `cargo clippy` and should be easy to fix! Once you do that I'll approve and merge

PR #1151, opened 2023-12-10
feat(wash-lib)!: Add support for model.status wadm operation in wash-lib

joonas:
## Feature or Problem

In the absence of a client library provided by wadm itself, I've started using `wash-lib` for this purpose, but I noticed that [`model.status` operation](https://github.com/wasmCloud/wadm/blob/dbe3c284dd3021b79653d776ddf63ee09b484928/src/server/mod.rs#L188-L198) isn't currently supported, so I thought I'd add it.


brooksmtownsend (approved):
Nice!

brooksmtownsend (approved):

PR #1152, opened 2023-12-11
refactor(providers): convert blobstore-s3 to provider bindgen

vados-cosmonic:
## Feature or Problem

The blobstore-s3 provider provides access to S3-compatible object storage from a given wasmcloud lattice. While in the past the provider was based on Smithy contracts and the somewhat older wasmcloud provider toolchain, WIT support has matured such that it can be supported as a new WIT world.

This commit converts the previously vendored blobstore-s3 provider into one that operates via provider-wit-bindgen.


crates/providers/blobstore-s3/src/lib.rs
```diff
@@ -0,0 +1,468 @@
...
+//     }
+
+//     async fn delete_link(&self, actor_id: &str) {
+//         self.config.write().await.remove(actor_id);
+//     }
+
+//     async fn shutdown(&self) {
+//         self.config.write().await.drain();
+//     }
+// }
```
brooksmtownsend, line 468: If this code is unused, we can just remove it

crates/providers/blobstore-s3/wit/deps/blobstore/blobstore.wit
```diff
@@ -0,0 +1,364 @@
...
+    /// For Smithy, timestamps mean time between January 1, 0001 CE to 23:59:59.999 and
+    /// December 31, 9999 CE, with millisecond granularity. While Smithy (v2.0) intentionally
+    /// does not constrain serialization/deserialiation semantics, we must take a stance here.
+    ///
+    /// This timestamp represents - represents absolute time in UTC,
+    /// as non-leap seconds and nanoseconds since the UNIX EPOCH
+    record timestamp {
+      /// The number of non-leap seconds since UNIX EPOCH in UTC
+      /// TODO(vadossi-cosmonic): this is a break with the previous Timestamp type which used an i64
+      sec: u64,
```
brooksmtownsend, line 19: Is there anything to be todone here?

tests/common/minio.rs
```diff
@@ -0,0 +1,41 @@
...
+use anyhow::{Context, Result};
+use tokio::process::Command;
+use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
+use url::Url;
+
+use super::{free_port, spawn_server};
+
+/// Spawn a minio server to use for testing
+pub async fn start_minio() -> Result<(
```
brooksmtownsend, line 13: Does this now mean minio is a requirement locally for running `cargo nextest run`?

brooksmtownsend (changes requested):

crates/providers/blobstore-s3/src/lib.rs
vados-cosmonic: Yup, this code got ported -- removed this!

crates/providers/blobstore-s3/wit/deps/blobstore/blobstore.wit
vados-cosmonic: Nope, this I thought would be a problem but it turns out it wasn't

tests/common/minio.rs
vados-cosmonic: Yeah, so I'm not sure exactly how we want to handle this, but similar to `redis` and `vault`. 

CI should be fine, as the nix stuff has been updated, but people do have to have this installed locally to run the tests

vados-cosmonic: I could move these tests to under a specific feature flag though, if `minio` is one too far

brooksmtownsend: I would be okay if the result is that that individual test fails, but doesn't bring down the entire test suite

vados-cosmonic: So that's *kind of* what happens -- the whole test suite doesn't go down -- only the part of the suite that does minio stuff which is `tests/blobstore-s3.rs` for now

vados-cosmonic: yeah unlike before/with the other `Makefile` driven stuff, the dependencies are started *during* the test. 

We could reduce our dependencies to one if we... bring in the :whale2: ... Which has been discussed before but people weren't so onboard with it

crates/providers/blobstore-s3/src/client.rs
```diff
@@ -104,246 +121,89 @@ impl StorageClient {
...
+        ctx: &Context,
+        arg: &ContainerId,
+    ) -> ProviderInvocationResult<bool> {
+        let bucket_id = self.unalias(arg);
+        match self.s3_client.head_bucket().bucket(bucket_id).send().await {
+            Ok(_) => Ok(true),
+            Err(se) => match se.into_service_error() {
+                HeadBucketError::NotFound(_) => Ok(false),
+                e => {
+                    error!(error = ?e, "Unable to head bucket");
```
connorsmith256, line 137: Would you mind replacing these `error` fields with `err`? Unfortunately there's an issue somewhere in the tracing stack, where the `error` field gets set as a boolean and ignores whatever the client tried to set

connorsmith256 (approved):

crates/providers/blobstore-s3/src/client.rs
vados-cosmonic: Ahh yes, will do!

brooksmtownsend (approved):

rvolosatovs (approved):
tests pass, lgtm

PR #1160, opened 2023-12-12
feat(host)!: Combine heartbeat and inventory payloads

### LINKED ISSUE #789: [FEATURE] Add enough information to heartbeat to fully describe state
wadm and the lattice observer projects currently have to query the host inventory to retrieve all of the running actors and providers AND their annotations. This is inefficient and any consumer of the wasmbus event stream should be able to synthesize the state of a lattice by just listening to the heartbeats.

Ideally, a host heartbeat and host inventory should contain the same information.

We could add a new event purely for the purpose of observing state, but that just feels excessive unless the heartbeat is slimmed down to purely be a liveness indicator.

**Additional context**
 The original reason why heartbeats were slimmed down to just include an actor ID and count were because we tracked individual instance IDs and with 10000+ actors we would exceed the NATS max message payload of 1mb. This is no longer an issue unless a host is running 10000+ unique actors/providers which is...excessive.

brooksmtownsend:
- feat(host)!: change heartbeat payload to inventory
- feat(control-interface)!: add heartbeat fields to inventory

## Feature or Problem
In https://github.com/wasmCloud/wadm and https://github.com/wasmCloud/lattice-observer, and any other client, after a heartbeat a host inventory query is necessary in order to get enough information to reflect the state of the host. This PR solves this issue, giving any interested client the ability to know the full state of what's running on the host with only the heartbeat or the host inventory.

## Related Issues
Fixes #789 

## Release Information
wasmCloud v0.82.0
control interface v0.33.0

## Consumer Impact
The only consumers of the heartbeat payload, noted above, I'll update to work with both the old payload and the new payload. Worth noting that the only deserialize breaking change here is in the heartbeat, changing the actor map into a list and changing the provider "public_key" into "id" which I mitigated by adding a serde alias.

I see this as a test for future compatibility and will try out a strategy to see how we can work with deserializing two possible versions in the future.

### Unit Test(s)
Modified wasmbus test to ensure that the new inventory variables, version, uptime_seconds, uptime_human come back as sensible values.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2316,8 +2316,6 @@ impl Host {
     async fn inventory(&self) -> HostInventory {
         trace!("generating host inventory");
         let actors = self.actors.read().await;
-        // let actors: HashMap<&String, usize> = stream::iter(actors.iter())
-        // Previous format was pubkey => max
         let actors: Vec<_> = stream::iter(actors.iter())
             .filter_map(|(id, actor)| async move {
                 let instances = actor.instances.read().await;
@@ -2344,9 +2342,9 @@ impl Host {
                         }
                     })
                     .collect();
-                if instances.is_empty() {
+                let Some(image_ref) = instances.first().map(|i| i.image_ref.clone()) else {
                     return None;
-                }
+                };
                 let name = actor
                     .claims()
                     .and_then(|claims| claims.metadata.as_ref())
@@ -2354,13 +2352,7 @@ impl Host {
                     .cloned();
                 Some(ActorDescription {
                     id: id.into(),
-                    // SAFETY: We just checked above and returned `None` if instances were empty
-                    // Preserve backwards compatible payload by returning the first image reference
-                    image_ref: instances
-                        .first()
-                        .expect("actor should have at least one instance")
-                        .image_ref
-                        .clone(),
+                    image_ref,
                     instances,
                     name,
                 })
@@ -2388,8 +2380,6 @@ impl Host {
                         rev: revision,
                         ..
                     } = claims.metadata.as_ref()?;
-                    // In previous heartbeat, was ID
-                    // "public_key": public_key,
                     Some(instances.iter().map(
                         move |(link_name, ProviderInstance { annotations, .. })| {
                             let annotations = Some(annotations.clone().into_iter().collect());
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2200,7 +2200,14 @@ impl Host {
                         move |_| {
                             let host = Arc::clone(&host);
                             async move {
-                                let heartbeat = host.heartbeat().await;
+                                let heartbeat = match host.heartbeat().await {
+                                    Ok(heartbeat) => heartbeat,
+                                    Err(e) => {
+                                        error!("failed to generate heartbeat: {e}");
+                                        return;
+                                    }
+                                };
+
                                 if let Err(e) =
                                     host.publish_event("host_heartbeat", heartbeat).await
                                 {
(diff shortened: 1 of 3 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -2305,21 +2305,57 @@ impl Host {
     }
 
     #[instrument(level = "debug", skip_all)]
-    async fn heartbeat(&self) -> serde_json::Value {
-        trace!("generating heartbeat");
+    async fn inventory(&self) -> HostInventory {
+        trace!("generating host inventory");
         let actors = self.actors.read().await;
-        let actors: HashMap<&String, usize> = stream::iter(actors.iter())
+        // let actors: HashMap<&String, usize> = stream::iter(actors.iter())
```
connorsmith256, line 2311: Remove me?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2305,21 +2305,57 @@ impl Host {
...
+                            image_ref: Some(instance.image_reference.clone()),
+                            // We only accept u16 values on the control interface, so the try_from is a safety measure.
+                            max_concurrent: instance
+                                .max
+                                .and_then(|m| u16::try_from(m.get()).ok())
+                                .unwrap_or(u16::MAX),
+                        }
+                    })
+                    .collect();
+                if instances.is_empty() {
```
connorsmith256, line 2339: I think you could avoid the `expect` below by doing something like
```rust
let Some(image_ref) = instances.first().map(|i| i.image_ref) else {
  return None;
};

```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2328,36 +2364,70 @@ impl Host {
...
-                        })
-                    })
+                    let jwt::CapabilityProvider {
+                        capid: contract_id,
+                        name,
+                        rev: revision,
+                        ..
+                    } = claims.metadata.as_ref()?;
+                    // In previous heartbeat, was ID
+                    // "public_key": public_key,
```
connorsmith256, line 2384: I don't follow what this comment means 🤔

connorsmith256 (commented):
Mostly LGTM. Approval pending on resolution of comments

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Whoops, leftover from testing.

brooksmtownsend:
Adding label to hold this back to merge so we can cut 0.81 first. I know we can cherry pick but that just keeps it simple

crates/host/src/wasmbus/mod.rs
```diff
@@ -2328,36 +2356,68 @@ impl Host {
...
+    async fn heartbeat(&self) -> serde_json::Value {
+        trace!("generating heartbeat");
+        let inventory = self.inventory().await;
+        match serde_json::to_value(inventory) {
+            Ok(value) => value,
+            // SAFETY: Failing to serialize the inventory shouldn't be possible as we control the instantiation of the
+            // struct For safety we'll fall back on the default but that shouldn't ever occur.
+            Err(e) => {
+                error!(%e, "failed to serialize inventory for heartbeat");
+                serde_json::Value::default()
```
connorsmith256, line 2418: If we can't serialize the inventory, I suspect it's better to not publish anything, since this will (probably?) break any client trying to parse the heartbeat payload

crates/host/src/wasmbus/mod.rs
```diff
@@ -3493,103 +3553,8 @@ impl Host {
...
-            host_id: self.host_key.public_key(),
-            issuer: self.cluster_key.public_key(),
-            labels: self.labels.read().await.clone(),
-            friendly_name: self.friendly_name.clone(),
-            actors,
-            providers,
-        })
-        .context("failed to encode reply")?;
+        let inventory = self.inventory().await;
+        let buf = serde_json::to_vec(&inventory).context("failed to encode reply")?;
```
connorsmith256, line 3557:
```suggestion
        let buf = serde_json::to_vec(&inventory).context("failed to encode inventory")?;
```

Better to be more specific?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2328,36 +2356,68 @@ impl Host {
...
-                            "link_name": link_name,
-                            "contract_id": contract_id.unwrap_or("n/a"),
-                        })
-                    })
+                    let jwt::CapabilityProvider {
+                        capid: contract_id,
+                        name,
+                        rev: revision,
+                        ..
+                    } = claims.metadata.as_ref()?;
```
connorsmith256, line 2374: This should never fail, right? Was wondering if we should log if it does

connorsmith256 (commented):
Just a couple questions/suggestions

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Good call

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Yeah it shouldn't, I mean this would have to be a scenario where the claims had no metadata, which I think isn't possible based on how we admit claims 🤔

brooksmtownsend:
@connorsmith256 this one is ready for another look, rebased on  #1160 to bring in the u32 changes

PR #1163, opened 2023-12-12
ci(cargo): run host tests outside Nix sandbox

rvolosatovs:
## Feature or Problem

As host tests are growing, they are making more and more assumptions and so it's becoming increasingly difficult to run them in hermetic environment like Nix sandbox, move host tests outside the sandbox to simplify maintenance

## Related Issues

https://github.com/wasmCloud/wasmCloud/pull/1152


rvolosatovs:
test failed, since the CI machine is slow and unlike within Nix sandbox, we build in development profile here (which is slow to execute)

Possible solutions are:
- Increase timeout for actions
- Build in release (potentially very slow if we cannot cache the build)

Issue #1101, opened 2023-11-28
[FEAT] Make wash claims aware of wasmcloud.toml


brooksmtownsend:
## Feature
Just like `wash build` infers information about a project from the `wasmcloud.toml` file, it should be a one step process to run `wash claims sign` to sign an existing wasm file without building.

## Alternatives considered
Instead of adding this support, we could make the signing process a core piece of the `build` command. I don't see any reason why having flexibility here would be bad, running the sign command without worrying about the build process at all is valuable and this would essentially just reduce the number of parameters necessary to supply with `wash claims sign`. 

## Origin

              One thing about the `--sign-only` flag: Would it be better user experience for `wash sign` to be `wasmcloud.toml` aware? It feels like that would be more discoverable and ergonomic from my POV

_Originally posted by @thomastaylor312 in https://github.com/wasmCloud/wasmCloud/issues/938#issuecomment-1808926592_

connorsmith256:
Closing since this is obviated by https://github.com/wasmCloud/wasmCloud/issues/1219

PR #1166, opened 2023-12-13
[FEAT] Make wash claims aware of wasmcloud.toml

ahmedtadde:
## Feature or Problem
Just like wash build infers information about a project from the `wasmcloud.toml` file, it should be a one step process to run wash claims sign to sign an existing `wasm` file without building.

## Related Issues
#1101

## Release Information
`wash-cli v0.25.0`

## Consumer Impact
better dx

### Unit Test(s)
- all existing tests work as expected; refactor has no regression
- added [rust_actor_metadata_with_project_config_overrides](https://github.com/ahmedtadde/wasmCloud/blob/39d635f907431df5ee814c5d8a13c72f228ec0d5/crates/wash-lib/src/cli/claims.rs#L1384) and [rust_provider_metadata_with_project_config_overrides](https://github.com/ahmedtadde/wasmCloud/blob/39d635f907431df5ee814c5d8a13c72f228ec0d5/crates/wash-lib/src/cli/claims.rs#L1540) to test `wasmcloud.toml` configuration on `wash claims ...` commands

### Acceptance or Integration
all existing tests work as expected; refactor has no regression

### DIFF
--- a/crates/wash-lib/src/cli/mod.rs
+++ b/crates/wash-lib/src/cli/mod.rs
@@ -274,12 +274,15 @@ pub fn extract_keypair(
         let key_dir = match (&keygen_type, directory) {
             (_, Some(d)) => KeyDir::new(d)?,
             (KeyPairType::Module, _) => {
-                let project_config = get_config(None, Some(true))?;
-                match project_config.project_type {
-                    TypeConfig::Actor(ref actor_config) => {
-                        KeyDir::new(&actor_config.key_directory)?
+                if let Ok(project_config) = get_config(None, Some(true)) {
+                    match project_config.project_type {
+                        TypeConfig::Actor(ref actor_config) => {
+                            KeyDir::new(&actor_config.key_directory)?
+                        }
+                        _ => default_key_dir,
                     }
-                    _ => default_key_dir,
+                } else {
+                    default_key_dir
                 }
             }
             _ => default_key_dir,
--- a/crates/wash-lib/src/cli/mod.rs
+++ b/crates/wash-lib/src/cli/mod.rs
@@ -30,6 +30,7 @@ use crate::{
         fs::{read_key, KeyDir},
         KeyManager,
     },
+    parser::{get_config, TypeConfig},
 };
 
 pub mod capture;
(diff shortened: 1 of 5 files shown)


crates/wash-lib/src/cli/mod.rs
```diff
@@ -269,7 +270,31 @@ pub fn extract_keypair(
...
+                                    _ => default(),
+                                }
+                            } else {
+                                default()
+                            }
+                        }
+                    }
+                }
+                _ => KeyDir::new(determine_directory(directory)?),
+            }
```
brooksmtownsend, line 295: This statement may be more human-readable if the output was the directory that needed to be wrapped in `KeyDir::new()`, if that makes sense. 

My main issue with this setup is the function `default`, which is a nice use of a closure but is assigning to a name that seems like a reserved keyword.

brooksmtownsend (commented):
One request regarding the closure.

This issue will partially address #1101 in the ability to sign using claims in a wasmcloud.toml key directory which is awesome. The rest of the issue, signing with capability claims for example, can be completed as a followup or in this PR 😄

brooksmtownsend (approved):
One request regarding the closure.

This issue will partially address #1101 in the ability to sign using claims in a wasmcloud.toml key directory which is awesome. The rest of the issue, signing with capability claims for example, can be completed as a followup or in this PR 😄

brooksmtownsend (changes requested):
Meant to request change, not approve. Apologies for the confusion

ahmedtadde:
hey @brooksmtownsend, you mentioned
> the ability to sign using claims in a wasmcloud.toml key directory
by `claims` here, i assume you mean the signing `nk` key that is found/created in the key directory

> The rest of the issue, signing with capability claims for example, can be completed as a followup or in this PR 😄
i actually didn't get this at all originally. but now that you mentioned it, it sounds like we want to be  able to set fields on `ActorMetadata` and `ProviderMetadata` based on configuration in `wasmcloud.toml` (ie, when the user does specify via cli args or env vars). If that is the case, can you confirm the following configuration keys of interest:
-  `ActorMetadata` => `actor.claims, actor.call_alias, actor.destination, actor.tags?, version, revision, name`
- `ProviderMetadata` => `provider.capability_id, provider.vendor, version, revision, name`

crates/wash-lib/src/cli/mod.rs
ahmedtadde: i updated the logic, let me know if that works

ahmedtadde:
hey @brooksmtownsend  thanks for the feedback on my last inquiry. I have two more questions that came when i resumed word on this...

should `name, rev, ver` become optional on `ActorMetadata` now that the claims command is supposed to be `wasmcloud.toml` aware? The awareness of the project config is, otherwise, not useful for those particular args.
In addition, should the `provider` flag be removed on that same struct at this point?

brooksmtownsend:
@ahmedtadde Great questions, I think both of your assumptions are correct when it comes to unioning the caps + tags, we can handle those cases but it would be nice to just have folks use the wasmcloud.toml file. 

> should name, rev, ver become optional on ActorMetadata now that the claims command is supposed to be wasmcloud.toml aware?
I think they can be optional command line arguments, but I think they are still required when it comes to signing so if they aren't supplied then a project config is required.

> In addition, should the provider flag be removed on that same struct at this point?
I think so, I can't think of any reason why we should keep it

brooksmtownsend (approved):
This looks great, I'm excited to try this out!

Issue #1162, opened 2023-12-12
[CHORE] add wasmtime ignore condition for dependabot


brooksmtownsend:
According to @rvolosatovs, wasmtime dependencies will never be able to simply update on its own, so we shouldn't have Dependabot check on that dependency.

              OK, I won't notify you again about this release, but will get in touch when a new version is available. If you'd rather skip all updates until the next major or minor version, let me know by commenting `@dependabot ignore this major version` or `@dependabot ignore this minor version`. You can also ignore all major, minor, or patch releases for a dependency by adding an [`ignore` condition](https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates#ignore) with the desired `update_types` to your config file.

If you change your mind, just re-open this PR and I'll resolve any conflicts on it.

_Originally posted by @dependabot[bot] in https://github.com/wasmCloud/wasmCloud/issues/1131#issuecomment-1852570241_

PR #1168, opened 2023-12-13
ci(dependabot): group wasmtime dependencies

connorsmith256:
## Feature or Problem
Dependabot supports a `group` setting, which tells dependabot to open PRs [as a group](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-3)

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/1162

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,10 @@ updates:
   directory: "/"
   schedule:
     interval: "daily"
+  groups:
+    wasmtime: # wasmtime dependencies need to be updated together
+      patterns:
+      - "wasmtime*"
 - package-ecosystem: "cargo"
   directory: "/crates/providers"
   schedule:


brooksmtownsend (commented):
LGTM as long as there aren't also `wit*` deps that need to go with it, cc @rvolosatovs

rvolosatovs (approved):
LGTM.
I'd be happy to get major release bumps this way as well, sure they'll need some manual intervention most likely, but it'd be automated and reduce maintenance load anyway

.github/dependabot.yml
```diff
@@ -4,6 +4,12 @@ updates:
   directory: "/"
   schedule:
     interval: "daily"
+  groups:
+    wasmtime: # wasmtime dependencies need to be updated together
+      patterns:
+      - "wasmtime*"
+      update-types:
+      - "patch"
```
rvolosatovs, line 12: to clarify, I'd just drop this

rvolosatovs:
> LGTM as long as there aren't also `wit*` deps that need to go with it, cc @rvolosatovs

That should only happen for major versions, but I don't see why that would be a problem anyway

rvolosatovs (approved):

PR #1178, opened 2023-12-14
feat(example): add tinygo http hello world

brooksmtownsend:
## Feature or Problem
This PR adds a simple TinyGo "hello world" example that works with wasmtime 16 and wasmCloud 0.81. This must be built using the unreleased wash 0.25, so I'm leaving the PR in draft for the time being. It also could use a little cleanup around the project edges, including deciding if we keep `echo` or just use this instead for a minimal HTTP example.

## Related Issues
#1184 

## Release Information
N/A

## Consumer Impact
New example yay


brooksmtownsend:
I'll bring this PR and #1184 out of draft once I can remove my adapter and build the component without the custom things in `wasmcloud.toml`. Not far away!

examples/golang/actors/http-hello-world/wit/world.wit
```diff
@@ -0,0 +1,5 @@
+package wasmcloud:hello;
+
+world hello {
+  include wasi:http/proxy@0.2.0-rc-2023-12-05;
```
vados-cosmonic, line 4: The `proxy` world has a *slightly* different intent -- it's impressive that we can properly implement this, but I think it pulls in stuff you don't need like `outgoing-handler` (since we're not doing any actual proxying yet)

Maybe it's worth considering using logging + http incoming handler?

```wit
    import wasi:logging/logging;
    export wasi:http/incoming-handler@0.2.0-rc-2023-12-05;
```

I mean logging itself also isn't even strictly necessary.

vados-cosmonic (approved):
LGTM :rocket:

examples/golang/actors/http-hello-world/wit/world.wit
brooksmtownsend: Good call! We can just do the handler. I may leave the logging out if wasm time serve doesn't support it

examples/golang/actors/http-hello-world/project-generate.toml
```diff
@@ -0,0 +1,9 @@
+[template]
+
+raw = [
+  "*.wasm"
+]
+exclude = [
+  "gen/",
```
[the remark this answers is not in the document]
brooksmtownsend: Yeah we can exclude keys as well, and no I don't think we would check it in unless we had a good reason to avoid generating the interface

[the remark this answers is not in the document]
vados-cosmonic: The only thing I'm a bit worried about is when the interfaces change out from under us -- but probably not a big deal for now... the rate of change *should* slow going forward :crossed_fingers:

examples/golang/actors/http-hello-world/wit/world.wit
brooksmtownsend: Left out logging since `wasmtime serve -Scommon` doesn't include that as an export but I will add in our other docs!

examples/golang/actors/http-hello-world/wadm.yaml
```diff
@@ -0,0 +1,35 @@
...
+      properties:
+        # TODO: you must replace the path below to match your genreated code in build
+        # Try using `wash build -o json` and use the `actor_path` field, prepended with `file://`
+        image: file:///the/absolute/path/to/build/http-echo-tinygo-component_s.wasm
+      traits:
+        # Govern the spread/scheduling of the actor
+        - type: spreadscaler
+          properties:
+            replicas: 1
+        # Link the HTTP server, and inform it to listen on port 8081
```
connorsmith256, line 22: Nit: 8081 vs 8080

brooksmtownsend: Good catch! fixed.

connorsmith256 (approved):

PR #1185, opened 2023-12-15
refactor: update golang example to wasmtime 16

vados-cosmonic:
## Feature or Problem

With the fast-moving development of WebAssembly ecosystem, WASI, the Component Model, and WIT have seen many changes in the last couple months.

For example, The existing golang echo example in the repo was originally built when wit-bindgen version 0.13.1 was the most important version, and upstream wit-bindgen is now at 0.16.0. As wit-bindgen reflects releases of wasmtime and the ecosystem as a whole, there's been a lot of sales.

This commit updates the golang example echo actor to use the WIT and related generated bindings for newer versions of wasmtime 16 and related WIT definitions, including resources.


examples/golang/actors/http-echo-tinygo/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	echo "echo/gen"
+	echo_types "echo/gen"
```
brooksmtownsend, line 5: do we need this and `echo` here?

brooksmtownsend (commented):
Just one thing that I think can be removed, LGTM

brooksmtownsend (approved):

examples/golang/actors/http-echo-tinygo/main.go
brooksmtownsend: Not a big blocker and can be easily removed

PR #1190, opened 2023-12-17
chore: add playwright

yordis:


.github/workflows/washboard.yml
```diff
@@ -2,8 +2,9 @@ name: washboard
 
 on:
   push:
-    paths: 
+    paths:
       - 'washboard-ui/**'
+      - '.tool-versions'
```
lachieh, line 7: I don't see a problem with this being at the root level but will want one of our @wasmCloud/org-maintainers to confirm.

lachieh (approved):

washboard-ui/playwright.config.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const PORT = 5173;
+
+export default defineConfig({
+  testDir: './tests/playwright',
+  fullyParallel: true,
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 2 : 0,
+  workers: process.env.CI ? 1 : undefined,
+  reporter: 'html',
```
lachieh, line 11: Actually, let's add the  github reporter in CI and switch to the list for local dev.

```suggestion
  reporter: process.env.CI ? [['github'], ['html']] : 'list',
```

lachieh (changes requested):

.github/workflows/washboard.yml
connorsmith256: Is there a downside to moving this under `washboard-ui`? I'd prefer not adding another file to the top level, if possible

.github/workflows/washboard.yml
```diff
@@ -49,6 +56,16 @@ jobs:
           name: washboard
           path: ./washboard-ui/washboard.tar.gz
 
+      - name: Upload Playwright Report
+        uses: actions/upload-artifact@v3
+        if: always()
```
connorsmith256, line 61: Nit: why use `if` if we always upload?

yordis: Because a failed test (therefore a job) will product artifacts that are useful for troubleshooting such as screenshot, videos, and so on.

connorsmith256: Ah, `always()` forces this step to run, even if a previous step (the tests) fails. Gotcha

connorsmith256 (approved):

lachieh (approved):
Wonderful!

PR #1194, opened 2023-12-18
refactor(examples): add type aliases, comments, update golang code

vados-cosmonic:
## Feature or Problem

This commit improves the readability of the golang echo actor in-tree WIT-powered example, adding type aliases and more comments explaining the code.


examples/golang/actors/http-echo-tinygo/main.go
```diff
@@ -1,26 +1,38 @@
 package main
 
 import (
-	echo "echo/gen"
-	echo_types "echo/gen"
 	"encoding/json"
 	"fmt"
+	echo "github.com/wasmcloud/wasmcloud/examples/golang/actors/echo/gen"
```
brooksmtownsend, line 6: nit: `golang/actors/http-echo-tinygo/gen` right?

brooksmtownsend (approved):
Very nice! Optional nit since I'm fairly sure it doesn't affect anything.

You got me thinking with that template... we might be able to do some smart things with file paths and wadm manifests 🤔

examples/golang/actors/http-echo-tinygo/main.go
vados-cosmonic: Ah yes! I need to change the module name too -- thanks.

vados-cosmonic: Fixed!

brooksmtownsend (approved):

PR #1217, opened 2023-12-21
feat(examples): add docker compose examples

connorsmith256:
This migrates the docker compose examples into the main repo. The contents are mostly the same as https://github.com/wasmcloud/examples/tree/main/docker, but with updated versions and a more detailed README

### DIFF
--- a/examples/docker/docker-compose-full.yml
+++ b/examples/docker/docker-compose-full.yml
@@ -1,0 +1,64 @@
+# This docker-compose file starts an entire wasmCloud ecosystem, including:
+#   a NATS server
+#   a local OCI registry
+#   grafana + tempo for tracing
+#   a wasmCloud host
+#   a WADM server for managing applications
+
+version: "3"
+services:
+  nats:
+    image: nats:2.10.7-alpine
+    ports:
+      - "4222:4222"
+      - "6222:6222"
+      - "8222:8222"
+    command: ["-js"]
+  registry:
+    image: registry:2.8
+    ports:
+      - "5000:5000"
+  grafana:
+    image: grafana/grafana:10.0.10
+    ports:
+      - 5050:3000
+    volumes:
+      - ./grafana-datasources.yaml:/etc/grafana/provisioning/datasources/datasources.yaml
+    environment:
+      - GF_AUTH_ANONYMOUS_ENABLED=true
+      - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
+      - GF_AUTH_DISABLE_LOGIN_FORM=true
+    depends_on:
+      - tempo
+  tempo:
+    image: grafana/tempo:2.3.1
+    command: ["-config.file=/etc/tempo.yaml"]
+    volumes:
+      - ./tempo.yaml:/etc/tempo.yaml
+    ports:
+      - 8000:8000 # tempo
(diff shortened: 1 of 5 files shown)


examples/docker/docker-compose-full.yml
```diff
@@ -0,0 +1,64 @@
...
+    environment:
+      RUST_LOG: debug,hyper=info,async_nats=info,oci_distribution=info,cranelift_codegen=warn
+      WASMCLOUD_LOG_LEVEL: debug
+      WASMCLOUD_RPC_HOST: nats
+      WASMCLOUD_CTL_HOST: nats
+      WASMCLOUD_OCI_ALLOWED_INSECURE: registry:5000
+      OTEL_TRACES_EXPORTER: otlp
+      OTEL_EXPORTER_OTLP_ENDPOINT: http://tempo:55681/v1/traces
+    ports:
+      - "8080-8089:8080-8089" # Allows exposing examples on ports 8080-8089
```
vados-cosmonic, line 57: Should we be more explicit about what it's using these ports? May not be clear that examples *running httpservers* are the most likely thing to use those ports.

vados-cosmonic (approved):
LGTM :rocket:

examples/docker/docker-compose-full.yml
connorsmith256: I raised an eyebrow at this too (and briefly got confused when I tried to link the HTTP server on port 1234 and requests failed)

What do you think we should do? Just improve the comment by clarifying these are for examples using the HTTP server? Widen the range? Something else?

t3hmrman: Yeah I think for now improving the comment is the easiest thing... We clearly don't want to expose *all* the ports.

Isn't the default for httpserver 8000? Maybe we should include that one too, though I know often examples use 8081. It might be reasonable to take 8000-8100 or something.

vados-cosmonic: Yeah I think for now improving the comment is the easiest thing... We clearly don't want to expose *all* the ports.

Isn't the default for httpserver 8000? Maybe we should include that one too, though I know often examples use 8081. It might be reasonable to take 8000-8100 or something.

connorsmith256: Yeah the default is 8000. 8000-8100 sounds good to me 👍

examples/docker/docker-compose-full.yml
```diff
@@ -0,0 +1,64 @@
...
+    ports:
+      - 8000:8000 # tempo
+      - 55681:55681 # otlp http
+
+  wasmcloud:
+    depends_on:
+      - "nats"
+      - "grafana"
+      - "tempo"
+    image: wasmcloud/wasmcloud:0.81.0-rc1 # TODO(nomerge!): update back to latest
```
brooksmtownsend, line 47: No merge? 😄 
```suggestion
    image: wasmcloud/wasmcloud:latest
```

brooksmtownsend (changes requested):

examples/docker/docker-compose-full.yml
connorsmith256: Yep! I'm waiting :)

brooksmtownsend (approved):

PR #1223, opened 2023-12-21
feat(wash-cli): prefix absolute path references with file://

connorsmith256:
A few users have been confused trying to run commands like `wash start actor /path/to/file` and seeing an error (from the host). The `file://` scheme is rough devX, so instead, if wash can detect the reference is an absolute file path (i.e. it starts with a `/`), wash can simply prefix the reference for the user

### DIFF
--- a/crates/wash-lib/src/cli/start.rs
+++ b/crates/wash-lib/src/cli/start.rs
@@ -79,24 +79,27 @@ pub async fn handle_start_actor(cmd: StartActorCommand) -> Result<CommandOutput>
         .into_ctl_client(Some(cmd.auction_timeout_ms))
         .await?;
 
+    let actor_ref = if cmd.actor_ref.starts_with('/') {
+        format!("file://{}", &cmd.actor_ref) // prefix with file:// if it's an absolute path
+    } else {
+        cmd.actor_ref.to_string()
+    };
+
     let host = match cmd.host_id {
         Some(host) => find_host_id(&host, &client).await?.0,
         None => {
             let suitable_hosts = client
                 .perform_actor_auction(
-                    &cmd.actor_ref,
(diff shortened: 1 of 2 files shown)


vados-cosmonic (commented):
Changes look good to me, but test is failing, is it related? A flake? We shouldn't (tm) have any more flakes in the wash tests

connorsmith256:
@vados-cosmonic looks like @rvolosatovs reran the job and it passed, so it does look like a flake

vados-cosmonic:
Did you happen to get the error report? I'm trying to find the failed run, but I can't find it, I can only see [the successful run](https://github.com/wasmCloud/wasmCloud/actions/runs/7293737899) now.

Approving but wanted to know if that failure was an actual test failure or some other infra ephemeral failure

vados-cosmonic (approved):

crates/wash-lib/src/cli/start.rs
```diff
@@ -79,24 +79,27 @@ pub async fn handle_start_actor(cmd: StartActorCommand) -> Result<CommandOutput>
         .into_ctl_client(Some(cmd.auction_timeout_ms))
         .await?;
 
+    let actor_ref = if cmd.actor_ref.starts_with('/') {
+        format!("file://{}", &cmd.actor_ref) // prefix with file:// if it's an absolute path
+    } else {
+        cmd.actor_ref.to_string()
+    };
```
brooksmtownsend, line 86: Just a note for something I considered for other reviewers, this is a better check than checking to see if the ref is an absolute path on the system as wash could be starting the actor from a path on a different machine 👍🏻

brooksmtownsend (commented):
Love it 👍🏻

Issue #1102, opened 2023-11-28
[FEATURE] Namespace all command line environment variables with `WASMCLOUD_`


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
Most of our flags are prefixed with `WASMCLOUD_` but there are a few that are not.
```
      --nats-host <NATS_HOST>
          NATS server host to connect to [env: NATS_HOST=] [default: 127.0.0.1]
      --nats-port <NATS_PORT>
          NATS server port to connect to [env: NATS_PORT=] [default: 4222]
      --nats-jwt <NATS_JWT>
          A user JWT to use to authenticate to NATS [env: NATS_JWT=]
      --nats-seed <NATS_SEED>
          A seed nkey to use to authenticate to NATS [env: NATS_SEED=]
      --oci-registry <OCI_REGISTRY>
          Used in tandem with `oci_user` and `oci_password` to override credentials for a specific OCI registry [env: OCI_REGISTRY=]
      --oci-user <OCI_USER>
          Username for the OCI registry specified by `oci_registry` [env: OCI_REGISTRY_USER=]
      --oci-password <OCI_PASSWORD>
          Password for the OCI registry specified by `oci_registry` [env: OCI_REGISTRY_PASSWORD=]

```

**Describe the solution you'd like**
Ideally all of these environment variables are prefixed with `WASMCLOUD_` and serde aliases are added for the previous environment variables to preserve backwards compatibility.

**Describe alternatives you've considered**
Worth mentioning that the OTEL environment variables are not namespaced as per convention, we match the standard OTEL variable syntax. Future variables like this don't have to be namespaced.

hailelagi:
hi @brooksmtownsend would love to contribute this, is this behaviour still desired? Is my assumption correct that the desired cli ux for the env variables parsed by clap would all be prefixed for each variable above -- for example `NATS_HOST` becomes `WASMCLOUD_NATS_HOST` like so:

```rust
/// NATS server host to connect to
#[clap(long = "nats-host", env = "WASMCLOUD_NATS_HOST", alias="NATS_HOST")]
pub nats_host: Option<String>,
```

thomastaylor312:
Hey @hailelagi! So the only thing that needs to change is just the environment variable. The actual field name can remain the same. 

> and lastly some of the env flags such as OCI_USER, OCI_PASSWORD, OCI_REGISTRY, NATS_JWT and NATS_SEED seem have either been renamed or removed?

[...]

PR #1243, opened 2023-12-26
feat: consistently prefix wash-cli arguments

hailelagi:
## Feature or Problem

consistent namespace prefixes for cli flags

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/1102

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -32,10 +32,20 @@ struct Args {
     #[clap(long = "nats-port", default_value_t = 4222, env = "NATS_PORT")]
     nats_port: u16,
     /// A user JWT to use to authenticate to NATS
-    #[clap(long = "nats-jwt", env = "NATS_JWT", requires = "nats_seed")]
+    #[clap(
+        long = "nats-jwt",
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -190,14 +201,16 @@ struct Args {
     #[clap(
         long = "oci-user",
         env = "OCI_REGISTRY_USER",
+        alias = "WASMCLOUD_OCI_REGISTRY_USER",
```
vados-cosmonic, line 204: Any reason you didn't go for `WASMCLOUD_*` being the `env` value here like other places?

vados-cosmonic (approved):
LGTM :rocket: 

Thanks for the contribution @hailelagi, let's run the CI and we should be good to merge.

src/main.rs
hailelagi: good catch! not in particular, glossed over this thanks

brooksmtownsend (approved):
Awesome! One more task checked off the 1.0 list, thank you @hailelagi

Issue #1241, opened 2023-12-26
[BUG] echo-tinygo template broken due to preview1-triggered wit generation


rvolosatovs:
**Describe the bug**
Wash integration tests fail in CI

**To Reproduce**
1. https://github.com/wasmCloud/wasmCloud/actions/runs/7325077061 this seems to be the first occurrence

**Expected behavior**
Success

**Additional context**
It seems pretty clear, that these tests are nondeterministic, so I doubt this is the last time they will break. Is there a reason we cannot avoid relying on fetching from external sources in these tests?

brooksmtownsend:
I believe @vados-cosmonic did a good bit of work to make these deterministic, the issue here of
```
missing `wit_world` in wasmcloud.toml ([actor] section) to run go bindgen generate
```

seems like something we can fix 🤔

vados-cosmonic:
Yeah I don't think this issue actually reflects the failure that was found (and linked), slightly more information around that line that you found, Brooks:

```
 missing `wit_world` in wasmcloud.toml ([actor] section) to run go bindgen generate
thread 'integration_build_tinygo_actor_unsigned' panicked at crates/wash-cli/tests/wash_build.rs:304:5:
assertion failed: status.success()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

So after doing some digging it looks like this is the confluence of a couple things:

- `wash` attempts to perform bindgen on preview1 *or* preview2 components
- the `echo-tinygo` example has a `wash_target` of `wasm32-wasi-preview1`
- the `echo-tinygo` component uses absolutely no component-relevant machinery

Building preview1 or preview2 does not technically mean that WIT is involved, so there's a mismatch there -- but we want to be WIT-first so I'm not sure if that check should be changed. 

The template builds just fine if we *remove* `wasm_target` (so it defaults to `CoreModule` (i.e. `wasm32-unknown-unknown`) which builds and is usable from the host just fine.

There are a few ways we can fix this, but I'm thinking of two things:

**At the `wash` level** we can update the check to at least make sure there *is* a `wit` directory at all (i.e. the project should be built using bindgen). We currently don't expect/let people change the directory where WIT is specified (it could technically be another dir) so that should at least indicate that 

**At the templates level** we should remove that setting ASAP until `wash` is released with the fix, so we don't break current users.

brooksmtownsend:
Fairly sure we fixed this right @vados-cosmonic ?

[...]

PR #1246, opened 2023-12-27
fix(wash): only generate tinygo when wit-dir present

vados-cosmonic:
## Feature or Problem

Golang projects built by wash which have
`wasm32-wasi-preview1` set as their `wasm_target` (in `wasmcloud.toml`) fail to build due to go
bindgen (i.e. `wit-bindgen-go`) being run on them.

While the wasmcloud ecosystem is WIT-first, it is possible to build preview1/preview2 components *without* WIT (i.e. with the legacy Smithy ecosystem), and projects that are built in that way should not have bindgen run on them.

This commit improves the check to use `wit_world` to determine whether to run go-based bindgen

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/1241

## Release Information

next

## Consumer Impact

templates built with tinygo that are using legacy smithy interfaces will build when using `wasm_target = "wasm32-wasi-preview1"`)

### Acceptance or Integration

There's a test that caught this failure -- `integration_build_tinygo_actor_unsigned`

### Manual Verification

Verified manually

### DIFF
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -315,7 +315,14 @@ fn build_tinygo_actor(
     }
 
     // Generate wit-bindgen code for Golang actors which are components-first
-    if let WasmTarget::WasiPreview1 | WasmTarget::WasiPreview2 = &actor_config.wasm_target {
+    //
+    // Running wit-bindgen via go generate is only for WIT-enabled projects, so we must limit
+    // to only projects that have their WIT defined in the expected top level wit directory
+    //
+    // While wasmcloud and it's tooling is WIT-first, it is possible to build preview1/preview2
+    // components that are *not* WIT enabled. To determine whether the project is WIT-enabled
+    // we check for the `wit` directory which would be passed through to bindgen.
+    if actor_config.wit_world.is_some() {
         generate_tinygo_bindgen(
             &output_dir,
             common_config.path.join("wit"),
@@ -323,7 +330,7 @@ fn build_tinygo_actor(
                 "missing `wit_world` in wasmcloud.toml ([actor] section) to run go bindgen generate",
             )?,
         )
-            .context("generating golang bindgen code failed")?;
+                .context("generating golang bindgen code failed")?;
     }
 
     let result = command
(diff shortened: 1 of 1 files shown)


crates/wash-lib/src/build.rs
```diff
@@ -315,15 +316,25 @@ fn build_tinygo_actor(
...
-    if let WasmTarget::WasiPreview1 | WasmTarget::WasiPreview2 = &actor_config.wasm_target {
+    //
+    // Running wit-bindgen via go generate is only for WIT-enabled projects, so we must limit
+    // to only projects that have their WIT defined in the expected top level wit directory
+    //
+    // While wasmcloud and it's tooling is WIT-first, it is possible to build preview1/preview2
+    // components that are *not* WIT enabled. To determine whether the project is WIT-enabled
+    // we check for the `wit` directory which would be passed through to bindgen.
+    let wit_dir = common_config.path.join("wit");
+    if let (WasmTarget::WasiPreview1 | WasmTarget::WasiPreview2, true) =
```
thomastaylor312, line 327: We should probably print out a warning if they have the wasm target set but the wit is _not_ there. It would help users that put that in, then find out they aren't building a component

thomastaylor312 (changes requested):
One minor addition here around UX

crates/wash-lib/src/build.rs
vados-cosmonic: The problem is that it's possible to build a component that isn't WIT-based, it's just that we expect components be WIT based right now. 

Rather than do this, what I'll do is key off of `wit_world`.

vados-cosmonic: To address the possibility of the wit dir being missing I'll ad a specific error message for that (can't do a warning, since this is `wash-lib`)

crates/wash-lib/src/build.rs
```diff
@@ -315,15 +315,22 @@ fn build_tinygo_actor(
     }
 
     // Generate wit-bindgen code for Golang actors which are components-first
-    if let WasmTarget::WasiPreview1 | WasmTarget::WasiPreview2 = &actor_config.wasm_target {
+    //
+    // Running wit-bindgen via go generate is only for WIT-enabled projects, so we must limit
+    // to only projects that have their WIT defined in the expected top level wit directory
+    //
+    // While wasmcloud and it's tooling is WIT-first, it is possible to build preview1/preview2
```
connorsmith256, line 322: Supernit typo:
```suggestion
    // While wasmcloud and its tooling is WIT-first, it is possible to build preview1/preview2
```

connorsmith256 (approved):

thomastaylor312 (approved):

PR #1266, opened 2024-01-02
feat(example): add logging and keyvalue for tutorial use

brooksmtownsend:
## Feature or Problem
This PR adds `wasi:keyvalue` and `wasi:logging` to the wit dependencies for both of our hello world examples for use in the quickstart tutorials. Without this, developers would need to manually modify the `deps.toml` file and run `wit-deps` update, so now during the quickstart documentation they can simply add these imports to their `wit/hello.wit` file and then use the capabilities.

This is only a temporary fix. Ideally these examples don't come with a wit deps directory at all and the only wit modification needed will be to specify custom import/exports, and a crate like `wasmcloud-actor` would be used to wrap wit definitions. I think that it's important to start here as a baseline and then remove some of the low level WebAssembly concepts as we can, as this level of detail will be necessary for languages where we don't have some sort of higher level SDK or library support.

## Related Issues
Enables https://github.com/wasmCloud/wasmcloud.com/pull/250

## Release Information
N/A

### DIFF
--- a/examples/golang/actors/http-hello-world/wit/deps.toml
+++ b/examples/golang/actors/http-hello-world/wit/deps.toml
@@ -1,2 +1,4 @@
 http = "https://github.com/WebAssembly/wasi-http/archive/v0.2.0-rc-2023-12-05.tar.gz"
-
+io = "https://github.com/WebAssembly/wasi-io/archive/v0.2.0-rc-2023-11-10.tar.gz"
+logging = "https://github.com/WebAssembly/wasi-logging/archive/main.tar.gz"
+keyvalue = "https://github.com/WebAssembly/wasi-keyvalue/archive/main.tar.gz"
--- a/examples/golang/actors/http-hello-world/wit/deps/io/poll.wit
+++ b/examples/golang/actors/http-hello-world/wit/deps/io/poll.wit
@@ -3,7 +3,7 @@ package wasi:io@0.2.0-rc-2023-11-10;
 /// A poll API intended to let users wait for I/O events on multiple handles
 /// at once.
 interface poll {
-    /// `pollable` represents a single I/O event which may be ready, or not.
+    /// `pollable` epresents a single I/O event which may be ready, or not.
     resource pollable {
 
       /// Return the readiness of a pollable. This function never blocks.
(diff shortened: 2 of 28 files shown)


examples/golang/actors/http-hello-world/wit/deps.toml
```diff
@@ -1,2 +1,4 @@
 http = "https://github.com/WebAssembly/wasi-http/archive/v0.2.0-rc-2023-12-05.tar.gz"
-
+io = "https://github.com/WebAssembly/wasi-io/archive/v0.2.0-rc-2023-11-10.tar.gz"
+logging = "https://github.com/WebAssembly/wasi-logging/archive/main.tar.gz"
+keyvalue = "https://github.com/WebAssembly/wasi-keyvalue/archive/main.tar.gz"
```
connorsmith256, line 4: Are these being at main at risk of floating? Should they be versioned instead like the others?

examples/golang/actors/http-hello-world/wit/deps/io/poll.wit
```diff
@@ -3,7 +3,7 @@ package wasi:io@0.2.0-rc-2023-11-10;
 /// A poll API intended to let users wait for I/O events on multiple handles
 /// at once.
 interface poll {
-    /// `pollable` represents a single I/O event which may be ready, or not.
+    /// `pollable` epresents a single I/O event which may be ready, or not.
```
connorsmith256, line 6: Supanit

brooksmtownsend: This is pulled from upstream 😭

examples/golang/actors/http-hello-world/wit/deps.toml
brooksmtownsend: Good question. No risk as long as you don't update of course which the tutorial doesn't say to do. This is the reference that we use in the actor and runtime crate 🤔 I just worked with that precedent

connorsmith256: Sounds good to me

connorsmith256 (approved):

PR #1295, opened 2024-01-09
feat(wash-lib): support other build languages

brooksmtownsend:
## Feature or Problem
This PR allows `wash` to support other languages than Rust or TinyGo by using the `build_command` and `build_artifact` information from wasmcloud.toml. The proper usage of these fields is to supply a build command that wash can execute for you, and then the build artifact should be the path to the built Wasm component so wash can then load it to sign the component. For example, a Javascript component will be compilable by wash if you use a wasmcloud.toml like so:
```toml
name = "componentize"
language = "javascript"
version = "0.1.0"
type = "actor"

[actor]
claims = []
build_command = "node componentize.mjs"
build_artifact = "hello.component.wasm"
destination = "hello_s.component.wasm"
```

I left the language as a completely open String, so users could supply any language other than rust or go if that's what they are using.

## Release Information
Patch bump in wash-lib and wash-cli, so whenever we release the next version

### Manual Verification
This worked for my componentize-js demo!

### DIFF
--- a/crates/wash-cli/README.md
+++ b/crates/wash-cli/README.md
@@ -109,7 +109,7 @@ There are three main sections of a `wasmcloud.toml` file: common config, languag
 | name          | string |                               | Name of the project                                                                    |
 | version       | string |                               | Semantic version of the project                                                        |
 | path          | string | `{pwd}`                       | Path to the project directory to determine where built and signed artifacts are output |
-| language      | enum   | [rust, tinygo]                | Language that actor or provider is written in                                          |
+| language      | enum   | [rust, tinygo, other]         | Language that actor or provider is written in                                          |
 | type          | enum   | [actor, provider, interface ] | Type of wasmcloud artifact that is being generated                                     |
 | wasm_bin_name | string | "name" setting                | Expected name of the wasm module binary that will be generated                         |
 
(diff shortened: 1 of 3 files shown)


crates/wash-lib/src/build.rs
```diff
@@ -126,6 +126,17 @@ pub fn build_actor(
                 };
                 actor_wasm_path
             }
+            LanguageConfig::Other(_) if actor_config.build_command.is_some() => {
+                // SAFETY: We checked that the build command is not None above
```
vados-cosmonic, line 130: NIT: don't think you need this comment

crates/wash-cli/README.md
```diff
@@ -132,6 +132,10 @@ There are three main sections of a `wasmcloud.toml` file: common config, languag
 | cargo_path  | string | `which cargo` | The path to the cargo binary            |
 | target_path | string | ./target      | Path to cargo/rust's `target` directory |
 
+#### Language Config - [other]
+
+If you are using a language other than Rust or Go, you can designate your language as any string. Since `wash` will be unable to infer the build toolchain, you must supply the `build_command` and the path to the `build_artifact` so that the artifact can be built and found before signing.
```
vados-cosmonic, line 137:
```suggestion
If you are using a language other than Rust or Go, you can designate your language as any string (ex. `javascript`). 

Since `wash` will be unable to infer the build toolchain from the language you provide, you must supply the `build_command` and the path to the `build_artifact` so that the artifact can be built and found before signing.
```

vados-cosmonic (approved):
LGTM :rocket:

crates/wash-lib/src/build.rs
brooksmtownsend: I figure it doesn't hurt, I just like including them wherever there is an `unwrap` in library code. I can remove if it bugs ya 😉

vados-cosmonic: oh yeah definitely not a big deal! Resolved!

PR #1298, opened 2024-01-10
feat(wash): detect arch when building PAR files

vados-cosmonic:
## Feature or Problem

This commit adds the ability to detect the arch + OS combination when building PAR files, and use that as a default valiue. It's unlikely that people will create PARs from *not* the native toolchain, and in those cases they can specify `--arch` as normal.

### DIFF
--- a/crates/wash-cli/src/par.rs
+++ b/crates/wash-cli/src/par.rs
@@ -13,6 +13,11 @@ use wash_lib::cli::{extract_keypair, inspect, par, CommandOutput, OutputKind};
 
 const GZIP_MAGIC: [u8; 2] = [0x1f, 0x8b];
 
+/// Helper function for detecting the arch used by the current machine
+fn detect_arch() -> String {
+    format!("{}-{}", std::env::consts::ARCH, std::env::consts::OS)
+}
+
 #[derive(Debug, Clone, Subcommand)]
 pub enum ParCliCommand {
     /// Build a provider archive file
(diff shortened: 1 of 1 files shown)


crates/wash-cli/src/par.rs
```diff
@@ -13,6 +13,11 @@ use wash_lib::cli::{extract_keypair, inspect, par, CommandOutput, OutputKind};
 
 const GZIP_MAGIC: [u8; 2] = [0x1f, 0x8b];
 
+/// Helper function for detecting the arch used by the current machine
+fn detect_arch() -> String {
+    format!("{}-{}", std::env::consts::ARCH, std::env::consts::OS)
+}
```
brooksmtownsend, line 19: Just to be doubly sure, does this ARCH/OS convention match our standard set?
```
  x86_64-macos
  aarch64-macos
  x86_64-windows
  aarch64-linux
  x86_64-linux
```
I know in some cases it can report `arm64` here instead of `aarch64` and `darwin` instead of `macos`

vados-cosmonic: It does! I tested on my Linux laptop/desktop (arch btw) and my Macbook

vados-cosmonic: For reference:

https://doc.rust-lang.org/1.72.1/std/env/consts/constant.ARCH.html
https://doc.rust-lang.org/1.72.1/std/env/consts/constant.OS.html

brooksmtownsend (approved):

PR #1299, opened 2024-01-10
feat(examples): add typescript http echo actor

vados-cosmonic:
## Feature or Problem

As Javascript support for WebAssembly progresses along with the `jco` toolchain, it's possible to build actors that work with the wasmcloud ecosystem, leveraging the existing host, provider and actor ecosystem.

This commit adds a Typescript HTTP echo actor which shows how to use the Typescript and `jco` ecosystems together with wasmcloud.


examples/typescript/actors/http-echo/README.md
```diff
@@ -0,0 +1,96 @@
+# Typescript HTTP Echo
+
+This repository
```
brooksmtownsend, line 3: Don't leave us in suspense!

examples/typescript/actors/http-echo/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+This command won't return (as it's the running host process), but you can view the output of the host.
+
+## Build the actor component
+
+To build the [actor component][wasmcloud-actor-component]:
+
+```console
+npm build
+```
```
brooksmtownsend, line 45: Could we set up the `wasmcloud.toml` to run `npm build` instead? That way you could use `wash build` here, and leave a note that it's just running `npm build` behind the scenes

brooksmtownsend: I see further in the package.json that the npm scripts do encapsulate `wash build`, I just think it would be very nice to have wash be the agent that builds all the thigns

brooksmtownsend (changes requested):
This is so cool 💥 I just have a request to make `wash` the main agent of building. Since folks in the TS ecosystem will be used to using `npm`, my opinion isn't too strongly held here as long as a `wash build` will work out of the box to build the component

examples/typescript/actors/http-echo/README.md
vados-cosmonic: Oh hey I didn't know you were looking at this already! Check out the updated README!

examples/typescript/actors/http-echo/README.md
vados-cosmonic: > Could we set up the wasmcloud.toml to run npm build instead? That way you could use wash build here, and leave a note that it's just running npm build behind the scenes

Ah sure! I was going to wait until we had your change in wash, but I can change this early.

> I see further in the package.json that the npm scripts do encapsulate wash build, I just think it would be very nice to have wash be the agent that builds all the thigns

Agreed

vados-cosmonic:
Renaming since this is a `hello-world` actor, not an `echo` actor!

PR #1304, opened 2024-01-11
docs(examples): add README for language level

vados-cosmonic:
## Feature or Problem

While users are perusing examples, it's probably a good idea to add some language-specific information at the root for the given language.

As we current have our example structure set up to be `examples/<language>/(actors|...)`, we have a change to add a README at the `examples/<language>` level.

This commit adds READMEs for language-level folders under examples.

### DIFF
--- a/examples/golang/README.md
+++ b/examples/golang/README.md
@@ -1,0 +1,51 @@
+# Golang Examples
+
+This folder contains examples that can be run with [wasmcloud][wasmcloud] which are written in [Golang][golang].
+
+Most examples compile with [TinyGo][tinygo] rather than the Golang compiler proper given that WebAssembly support landed in TinyGo first and is still slightly better supported there.
+
+## Building individual example actors
+
+Example actors built in Golang can be compiled similarly to any other wasmcloud project:
+
+```console
+wash build
+```
+
+## WebAssembly support
+
+As WebAssembly is intended to be a "compilation target" for native Golang code, upstream work is underway to integrate and improve support for the various standards of WebAssembly.
+
+| Language | Core Modules (`wasm32-unknown-unknown`) | Preview 1 (`wasm32-wasi-preview1`) | WASI Preview 2 (`wasm32-wasi-preview2`)  |
+|----------|-----------------------------------------|------------------------------------|------------------------------------------|
+| Golang   | ✅ (`GOOS=js`,`GOARCH=wasm`)            | ✅ (`GOOS=wasip1`)                 | ✅ (requires [adapter][wasi-p2-adapter]) |
+| TinyGo   | ✅ (`-target=wasm`)                     | ✅ (`-target=wasi`)                | ✅ (requires [adapter][wasi-p2-adapter]) |
+
+Tiny Go WASI support is detailed [on their website][tinygo-wasi].
+Golang WASI preview 1 support was [announced on the Golang blog][golang-blog-wasi]
+
+> [!NOTE]
+> Don't know what `wasm32-unknown-unknown` means versus `wasm32-wasi-preview1`?
+>
+> `wasm32-unknown-unknown` is a compile target which deals in core [WebAssembly modules][wasm-core-modules] (i.e. you're only given access to numbers at this level)
+> [`wasm32-wasi-preview1`][wasi-p1] is a compile target that provides richer types, support for more higher level platform APIs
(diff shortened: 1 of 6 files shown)


examples/golang/README.md
```diff
@@ -0,0 +1,41 @@
...
+Most examples compile with [TinyGo][tinygo] rather than the Golang compiler proper given that WebAssembly support landed in TinyGo first and is still slightly better supported there.
+
+## WebAssembly support
+
+As WebAssembly is intended to be a "compilation target" for native Golang code, upstream work is underway to integrate and improve support for the various standards of WebAssembly.
+
+| Language | Core Modules (`wasm32-unknown-unknown`) | Preview 1 (`wasm32-wasi-preview1`) | WASI Preview 2 (`wasm32-wasi-preview2`) |
+|----------|-----------------------------------------|------------------------------------|-----------------------------------------|
+| Golang   | ✅ (`GOOS=js`,`GOARCH=wasm`)            | ✅ (`GOOS=wasip1`)                 | 🏗                                       |
+| TinyGo   | ✅ (`-target=wasm`)                     | ✅ (`-target=wasi`)                | 🏗                                       |
```
connorsmith256, line 14: I think these tables are technically correct (the best kind of correct?), but will likely create the wrong impression for new users. I.e. while TinyGo, Rust, and JS/TS toolchains currently target WASI P1 from a _compiler_ perspective, the tooling (e.g. jco, cargo component, wash) will automatically adapt to a WASI P2 component, ready to be run in wasmtime, wasmCloud, etc.

So from a user perspective, I would expect to see a ✅ for some of these P2 columns

vados-cosmonic: Yeah I was trying to focus here on the baked-into-the-language functionality -- I'll add a note like "requires adapter"

vados-cosmonic: done, updated!

examples/golang/README.md
```diff
@@ -0,0 +1,42 @@
+# Golang Examples
+
+This folder contains examples that can be run with [wasmcloud][wasmcloud] which are written in [Golang][golang].
+
+Most examples compile with [TinyGo][tinygo] rather than the Golang compiler proper given that WebAssembly support landed in TinyGo first and is still slightly better supported there.
```
brooksmtownsend, line 5: this `tinygo` link didn't resolve properly

examples/golang/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+| Language | Core Modules (`wasm32-unknown-unknown`) | Preview 1 (`wasm32-wasi-preview1`) | WASI Preview 2 (`wasm32-wasi-preview2`)  |
+|----------|-----------------------------------------|------------------------------------|------------------------------------------|
+| Golang   | ✅ (`GOOS=js`,`GOARCH=wasm`)            | ✅ (`GOOS=wasip1`)                 | ✅ (requires [adapter][wasi-p2-adapter]) |
+| TinyGo   | ✅ (`-target=wasm`)                     | ✅ (`-target=wasi`)                | ✅ (requires [adapter][wasi-p2-adapter]) |
+
+Tiny Go WASI support is detailed [on their website][tinygo-wasi].
+Golang WASI preview 1 support was [announced on the Golang blog][golang-blog-wasi]
+
+> [!INFO]
```
brooksmtownsend, line 19: This INFO also didn't show up with the correct formatting 🤔

brooksmtownsend: Same with the other READMEs, I'm not super familiar with why

brooksmtownsend (changes requested):
A couple requests for markdown formatting.

Only other thing I'd love to have in `golang` / `rust` is just a snippet that details that those languages are supported out-of-the-box with `wash build`, and maybe a experimental note for TypeScript since we don't have that yet

examples/golang/README.md
vados-cosmonic: Oh this is weird -- that should be the right syntax for Github's alert...

vados-cosmonic: Ahh fixed, it was `[!NOTE]`

examples/golang/README.md
vados-cosmonic: Oh good catch, fixed!

brooksmtownsend (commented):
Love it!

brooksmtownsend (approved):
I SAID: I LOVE IT!

PR #1305, opened 2024-01-11
refactor(provider-wit-bindgen): macro crate structure

vados-cosmonic:
## Feature or Problem

Up until now, dependencies that were exercised by
`wasmcloud-wit-provider-bindgen` were required of the downstream binaries that were being used.

To make it easier for people to build providers (as they might be surprised by dependencies introduced by the macro), and encourage consistency in the dependencies used, we want to internalize the dependencies that the macro introduces (i.e. hanging them off the macro crate).

Given that Rust has a limitation where macro crates cannot have any other exports, we must create an "umbrella" crate which hosts the macro along with other dependencies, and use *that*.

This commit modifies the macro crate structure, renaming out `wasmcloud-provider-wit-bindgen`
to `wasmcloud-provider-wit-bindgen-macro`, and keeping `wasmcloud-provider-wit-bindgen` as the top level macro without breaking backwards compatibility.

This commit does not *internalize* the deps, it only goes as far as to make the structure where it is possible.


crates/provider-wit-bindgen-macro/src/lib.rs
brooksmtownsend, line 1: I assume this file was mostly moved? Or did this entire thing get modified?

vados-cosmonic: Yup! this PR is basically moving the macro crate to a `-macro` crate, so we can introduce a crate that the deps can be accessed from.

So `provider-wit-bindgen` will *contain* `provider-wit-bindgen-macro` (exposed via `provider-wit-bindgen::generate` as before), and *inside* `provider-wit-bindgen-macro` we can depend on a dep like `provider-wit-bindgen::deps::serde_bytes`.

vados-cosmonic: see this file:  `crates/provider-wit-bindgen/src/lib.rs` I think it's hidden by default

brooksmtownsend: Okay nice! I've used `git mv` for these types of files before and that typically smartens git up a little bit to not report it as 2600 deletes and adds

brooksmtownsend (approved):
Woo macros

PR #1306, opened 2024-01-11
ci(wash-cli): parse version from github ref

brooksmtownsend:
## Feature or Problem
This PR fixes one other issue with the wash-cli release to packagecloud which was a missing environment variable for parsing the reference. The release worked great, just had the default version.

## Related Issues
#1303

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -434,6 +434,7 @@ jobs:
   nfpm:
     if: startsWith(github.ref, 'refs/tags/wash-cli-v')
     env:
+      REF: ${{ github.ref }}
       PACKAGECLOUD_TOKEN: ${{ secrets.PACKAGECLOUD_API_TOKEN }}
     needs:
     - cargo


.github/workflows/wasmcloud.yml
```diff
@@ -434,6 +434,7 @@ jobs:
   nfpm:
     if: startsWith(github.ref, 'refs/tags/wash-cli-v')
     env:
+      REF: ${{ github.ref }}
```
connorsmith256, line 437: Odd that this is okay with `refs/tags`?

connorsmith256 (approved):

.github/workflows/wasmcloud.yml
brooksmtownsend: I don't follow, do you mean that it's okay that this has the refs/tags prefix? We take care of that a little below
```
export VERSION=$(echo $REF | cut -d/ -f3)
```

connorsmith256: Ah missed that, thanks!

PR #1320, opened 2024-01-12
chore(ci): update artifact actions

connorsmith256:
Obviates https://github.com/wasmCloud/wasmCloud/pull/1181 and https://github.com/wasmCloud/wasmCloud/pull/1182

This also groups the upload and download actions in the dependabot config, so future automated bumps should work

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -59,7 +59,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
         with:
           name: SARIF file
           path: results.sarif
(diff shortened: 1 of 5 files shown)


rvolosatovs (commented):
Changes looks good to me, but I have not checked the changelog for these actions - are we sure that all artifacts still appear where we would expect them? Could you make a test release in a fork to verify that all the artifacts we'd expect are still uploaded?

.github/workflows/scorecard.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
```
rvolosatovs, line 62: Is there a reason not to just use the tagged version like all other occurrences?

connorsmith256: I don't know why it was done this way before, I just kept it consistent 🤷

connorsmith256:
@rvolosatovs pushed host and wash tags to my fork, LGTY?
- https://github.com/connorsmith256/wasmCloud/actions/runs/7547250849
- https://github.com/connorsmith256/wasmCloud/actions/runs/7547253562

thomastaylor312 (approved):

PR #1321, opened 2024-01-12
fix(washboard): update lattice service

### LINKED ISSUE #1319: [FEATURE] Consolidate React API for LatticeService
The hooks that interact with the `LatticeService` instance have some overlap. It would be good to be able to consolidate these hooks into a unified interface that can doesn't need to much updating when the `LatticeService` gains more functionality.

We should:
- combine the `useLatticeConfig` and `useReactiveConfig` to use the familiar `useState` pattern.
- update usage of both hooks
- limit API surface of `LatticeService` to only that which is necessary for hooks

lachieh:
## Feature or Problem
This is mostly organization and moving toward familiar patterns found in the React ecosystem.
- Allow `LatticeService` class to be initialized outside as an instance (remove singleton pattern)
- Add `LatticeServiceProvider` component and `useLatticeService` hook, implement
- Add single barrel export for Lattice service "package"

## Related Issues
Closes #1319

## Release Information
`next`

## Consumer Impact
None

### Manual Verification
Confirmed still working with existing `wash up --nats-websocket-port=4001` example


washboard-ui/src/services/lattice/index.ts
```diff
@@ -0,0 +1,6 @@
+export {useLatticeData} from './hooks/use-lattice-data';
+export {useLatticeConfig} from './hooks/use-lattice-config';
+export {canConnect} from './helpers/can-connect';
+export {LatticeServiceProvider} from './context/LatticeServiceProvider';
+export {LatticeService} from './classes/lattice-service';
+export type {WadmActor, WadmHost, WadmLink, WadmProvider} from './classes/lattice-service';
```
yordis, line 6: At the app-level code, avoid re-exporting, this becomes an anti-pattern. At the app-level, go directly to where things are define.

lachieh: Normally I would agree, but this is what it's going for:
> Add single barrel export for Lattice service "package"

Eventually, this will be it's own package and having a single main export means that there's a clear API for using it

connorsmith256 (approved):

PR #1333, opened 2024-01-15
chore: Normalize service.names to use kebab-case

joonas:
## Feature or Problem

Standardizing on using kebab-case makes it easier for downstream consumers to operate on the names of services

### DIFF
--- a/crates/tracing/src/lib.rs
+++ b/crates/tracing/src/lib.rs
@@ -7,6 +7,7 @@ use std::env;
 use std::io::{IsTerminal, StderrLock, Write};
 
 use anyhow::Context;
+use heck::ToKebabCase;
 use once_cell::sync::OnceCell;
 use tracing::{Event, Subscriber};
 use tracing_subscriber::filter::LevelFilter;
@@ -114,6 +115,7 @@ pub fn configure_tracing(
 
 #[cfg(feature = "otel")]
 #[allow(clippy::missing_errors_doc)] // TODO: Document errors
+#[allow(clippy::needless_pass_by_value)]
 pub fn configure_tracing(
     service_name: String,
     otel_config: &OtelConfig,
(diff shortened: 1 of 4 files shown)


crates/tracing/src/lib.rs
```diff
@@ -114,18 +114,21 @@ pub fn configure_tracing(
 
 #[cfg(feature = "otel")]
 #[allow(clippy::missing_errors_doc)] // TODO: Document errors
+#[allow(clippy::needless_pass_by_value)]
```
brooksmtownsend, line 117: Note that we discussed this offline and decided to allow this instead of making a breaking API change at this time, but this is worth removing and resolving whenever we move the wasmcloud-tracing crate to 0.2.0 for legitimate reasons 😄

crates/tracing/src/lib.rs
```diff
@@ -114,18 +114,21 @@ pub fn configure_tracing(
...
 pub fn configure_tracing(
     service_name: String,
     otel_config: &OtelConfig,
     structured_logging_enabled: bool,
     log_level_override: Option<&Level>,
 ) -> anyhow::Result<()> {
+    use heck::ToKebabCase;
```
connorsmith256, line 124: Nit: unless there's a good reason to use inline imports (e.g. purposefully limiting scope to avoid a naming conflict, or conditional compilation), we prefer importing once at the top of the file. Makes things easier to find

joonas: Fair enough, I chose to import it here because this was the only place it was being used, so it would be local to the code that consumes it. I'm happy to hoist it up to the top if that's more appropriate.

connorsmith256: I don't have a strong opinion, but curious how @brooksmtownsend @thomastaylor312 feel

most importantly I wish we had automated tooling that just formatted these things for us :)

joonas: I went ahead and hoisted the use statement up to the top 👍

thomastaylor312: For future reference so my opinion is out in public 😆, I am really not a fan of code that imports in a function unless for the reasons @connorsmith256. I find that they clutter code and make it hard to see what is all actually imported

thomastaylor312 (approved):

connorsmith256 (approved):

Issue #1221, opened 2023-12-21
[RFC] Rename `lattice_prefix` to `lattice`


brooksmtownsend:
## Introduction

In wasmCloud, we have the concept of a Lattice which is a seamless connected network of wasmCloud hosts over NATS. As a measure of isolation and namespacing, wasmCloud hosts start with a `lattice_prefix` argument which is used on each RPC and control interface message that is sent or handled by that host and its actors/providers. 

By default, this is `default`, and covers the 95% use case of developers that are just running in one lattice. 

Some example topics with `default` and `custom` lattice prefixes, showing the namespacing ability:
```
wasmbus.rpc.default.MBSNVBOSBQL6L4G3NZERCQAAEMTFWGJUWOO2X4MV3EIDMVJZNOL2IIZO
wasmbus.evt.default.host_started
wasmbus.rpc.custom.MBSNVBOSBQL6L4G3NZERCQAAEMTFWGJUWOO2X4MV3EIDMVJZNOL2IIZO
wasmbus.evt.custom.host_started
```

## Motivation

In the interest of clarity and transferrable knowledge, this RFC proposes renaming `lattice_prefix` to `namespace`. For a developer new to the wasmCloud ecosystem, lattice is a new term and I think it deserves to be the emphasis of learning. Lattice prefix introduces another term where `namespace` could be equivalent and mean essentially the same thing.

Additionally, other orchestration frameworks like `nomad` and `kubernetes` use namespace as a way of identifying logically separate jobs/workloads, so platform developers looking at wasmCloud will have an immediate transfer of knowledge.

## Detailed Design

The first step of this design is to, at the CLI level of wasmcloud and wash, rename `lattice-prefix` to `namespace` and support `lattice-prefix` as an alias. The same thing should happen for our public APIs like the control interface and wadm, where the name of types should be changed to `namespace` but still serialized as `lattice_prefix`.

Our documentation should be updated to refer to this "namespace" as a way of providing logically separate relationships within a single lattice connection.

## Backwards Compatibility

Mentioned in the above are methods of providing backwards compatibility using serde aliases to preserve over-the-wire compatibility. Lattice prefix should be deprecated and, once wasmCloud hits 1.0, removed in favor of all tooling that uses `namespace`. 

## Alternatives Considered

We could rename this to `lattice_namespace` as well, which would preserve the connection to the lattice. I feel that simply using namespace is simple enough and does not reduce clarity.

## Conclusion

[...]

PR #1341, opened 2024-01-17
refactor(*)!: rename lattice prefix to just lattice

connorsmith256:
## Feature or Problem
This renames `lattice_prefix` to just `lattice` internally, including in user-facing ways:
- `--lattice-prefix` in the host/wash is now just `--lattice`
  - backwards compatibility is maintained via a clap alias, for the time being
- the `WASMCLOUD_LATTICE_PREFIX` env var is now just `WASMCLOUD_LATTICE`
  - backwards compatibility is maintained via manual checking, which emits a warning when the old env var is used

## Related Issues
https://github.com/wasmCloud/wasmCloud/issues/1221

## Release Information
- v0.33 of the control interface (breaking change)
- v0.82 of the host (not a breaking change, but that's the next version)
- v0.26 of wash (breaking change due to pulling in the new control interface)

## Consumer Impact
There shouldn't be any, other than users receiving warnings if they use the old env var

Users of the control interface crate will need to update now that `lattice_prefix` is not a public field

### Acceptance or Integration
The existing test suite has good coverage

### Manual Verification
I also played around with starting the host and using wash with the new flag/env var

### DIFF
--- a/crates/control-interface/src/types.rs
+++ b/crates/control-interface/src/types.rs
@@ -115,7 +115,9 @@ pub struct Host {
     pub labels: Option<KeyValueMap>,
     /// Lattice prefix/ID used by the host
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub lattice_prefix: Option<String>,
+    pub lattice: Option<String>,
+    #[serde(default, skip_serializing)]
+    lattice_prefix: Option<String>, // TODO(pre-1.0): remove me once all clients are updated to control interface v0.33
     /// NATS server host used for regular RPC
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub rpc_host: Option<String>,
@@ -128,6 +130,14 @@ pub struct Host {
     /// Current wasmCloud Host software version
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
+}
+
+// TODO(pre-1.0): remove me once all clients are updated to control interface v0.33
+impl Host {
+    #[must_use]
+    pub fn lattice(&self) -> Option<&String> {
+        self.lattice.as_ref().or(self.lattice_prefix.as_ref())
+    }
 }
 
 /// Describes the known contents of a given host at the time of
--- a/crates/wash-cli/src/bin/wash.rs
+++ b/crates/wash-cli/src/bin/wash.rs
@@ -230,6 +230,15 @@ async fn main() {
         .with_writer(std::io::stderr)
         .with_env_filter(EnvFilter::from_default_env())
         .init();
+
+    // TODO(pre-1.0): remove me
+    if let Ok(lattice) = std::env::var("WASMCLOUD_LATTICE_PREFIX") {
+        eprintln!(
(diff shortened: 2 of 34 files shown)


crates/control-interface/src/types.rs
```diff
@@ -130,6 +132,15 @@ pub struct Host {
     pub version: Option<String>,
 }
 
+// TODO(pre-1.0): remove me once all clients are updated to control interface v0.33
+impl Host {
+    #[must_use] pub fn lattice(&self) -> Option<&String> {
```
thomastaylor312, line 137: Why the `must_use` here?

crates/wash-cli/src/bin/wash.rs
```diff
@@ -230,6 +230,15 @@ async fn main() {
         .with_writer(std::io::stderr)
         .with_env_filter(EnvFilter::from_default_env())
         .init();
+
+    // TODO(pre-1.0): remove me
+    if let Ok(lattice) = std::env::var("WASMCLOUD_LATTICE_PREFIX") {
+        eprintln!(
+            "The `WASMCLOUD_LATTICE_PREFIX` environment variable is deprecated and will be removed. Please use `WASMCLOUD_LATTICE` instead."
+        );
+        std::env::set_var("WASMCLOUD_LATTICE", lattice);
```
thomastaylor312, line 239: Could this end up munging something if they have both vars set? I don't think this is blocking, but I have something deep in my memory (that could be totally false) that setting env vars for people (unless that is the actual intent of the user) is bad practice

src/main.rs
```diff
@@ -233,6 +234,14 @@ const DEFAULT_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
 #[tokio::main]
 #[allow(clippy::too_many_lines)]
 async fn main() -> anyhow::Result<()> {
+    // TODO(pre-1.0): remove me
+    if let Ok(lattice) = std::env::var("WASMCLOUD_LATTICE_PREFIX") {
+        eprintln!(
+            "The `WASMCLOUD_LATTICE_PREFIX` environment variable is deprecated and will be removed. Please use `WASMCLOUD_LATTICE` instead."
+        );
+        std::env::set_var("WASMCLOUD_LATTICE", lattice);
```
thomastaylor312, line 242: Same thing here about setting env vars, in case we decide it matters

thomastaylor312 (approved):
Nothing blocking, but maybe one possible discussion

crates/control-interface/src/types.rs
connorsmith256: No idea, clippy fix automatically added this

crates/wash-cli/src/bin/wash.rs
connorsmith256: I agree, but can you help me come up with an alternative approach? Based on my reading, clap only supports associating a single env var with an option using the macro :/

I'm okay with the grossness, given this is temporary ™️ code that should be released after host version n+1 is out. But if there's a way to avoid it entirely, I'm all ears

brooksmtownsend (approved):

connorsmith256:
@thomastaylor312 merging now given the likelihood of merge conflicts the longer this stays open, but happy to follow up and make the env var handling more graceful if you had an idea there

PR #1352, opened 2024-01-17
fix(examples): update port for tempo to remove conflict with host ports

connorsmith256:
I noticed this error:
```
docker-grafana-1    | logger=plugins.update.checker t=2024-01-17T19:00:23.605767216Z level=info msg="Update check succeeded" duration=119.701458ms
Error response from daemon: driver failed programming external connectivity on endpoint docker-wasmcloud-1 (a763906bc096c57b9addf7cead80f1eb84f61ded4344e69166a56b75f98bd801): Bind for 0.0.0.0:8000 failed: port is already allocated
```

So I simply updated the tempo port from 8000 to 7999

### DIFF
--- a/examples/docker/docker-compose-auxiliary.yml
+++ b/examples/docker/docker-compose-auxiliary.yml
@@ -27,5 +27,5 @@ services:
     volumes:
       - ./tempo.yaml:/etc/tempo.yaml
     ports:
-      - 8000:8000 # tempo
+      - 7999:7999 # tempo
       - 55681:55681 # otlp http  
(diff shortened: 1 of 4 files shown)


brooksmtownsend (approved):
8000 is also our httpserver default port, so we might run into an issue there for things where we're assuming no link defined port/address.

We could also just remap the tempo port to something else, I'm not sure if we depend on that or reference port 8000 in our documentation at the moment. Approving because fixing the issue is better than speculating about a possible problem

connorsmith256:
Great point @brooksmtownsend, I forgot 8000 was the default. Let's see if port 7999 works for tempo 😉

brooksmtownsend (approved):
Ignore my messages I didn't read tempo.yaml for a while and I can't delete them for some reason

PR #1355, opened 2024-01-18
feat(wadm): allow relative paths in file-based WADM manifests

vados-cosmonic:
## Feature or Problem

WADM does not allow non-relative file paths to be used for values like `image:` (which is relevant for actors and providers specified in the manifest).

If a user is using a local file path, it's very likely that the host on which the declarative architecture will be deployed is the same host as the one that is running `wadm`.

To enable users to more conveniently build declarative manifests, we can resolve `file://...` paths based on the path to the WADM file itself (which is known at load time).

The basic scheme is to update the `AppManifest`s to store YAML structure rather than a simple `String`, in order to enable iterating and replacing paths as is necessary.

This commit allows for relative paths in WADM manifests that are fed to commands like `wash app deploy`.

### DIFF
--- a/crates/wash-lib/src/app.rs
+++ b/crates/wash-lib/src/app.rs
@@ -1,10 +1,13 @@
 //! Interact with and manage wadm applications over NATS, requires the `nats` feature
 
-use std::{path::PathBuf, str::FromStr, time::Duration};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+use std::time::Duration;
 
 use anyhow::{bail, Context, Result};
 use async_nats::{Client, Message};
 use regex::Regex;
+use tracing::warn;
 use wadm::server::{
     DeleteModelRequest, DeleteModelResponse, DeployModelRequest, DeployModelResponse,
     GetModelRequest, GetModelResponse, ModelSummary, PutModelResponse, StatusResponse,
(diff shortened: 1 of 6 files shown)


crates/wash-lib/src/app.rs
```diff
@@ -50,10 +52,59 @@ impl ToString for ModelOperation {
...
+            // Convert the base dir + relative path into a file based URL
+            let full_path = base_dir.as_ref().join(
+                s.strip_prefix("file://")
+                    .context("failed to strip prefix")?,
+            );
+            // Build a file based URL and replace the existing one
+            if let Ok(url) = Url::from_file_path(full_path) {
+                *s = url.into();
+            }
+        }
```
brooksmtownsend, line 89: Thoughts on outputting a warning log or similar for when we resolve a full file path, and then nothing is there?

This may or may not be an error depending on where the dev is in their workflow of course, but just a little heads up to say "hey, I know you wanted this Wasm and I didn't find it"

brooksmtownsend (approved):
Quick turnaround from community call, and what a feature!

brooksmtownsend (changes requested):
Actually sorry I'm going to be pedantic here, would you mind amending your commit to note that this is a `feat(wash-lib)!` as that's where the primary change occurs, and the breaking change is because of new functionality/public members?

crates/wash-lib/src/app.rs
vados-cosmonic: Sure, added that! It's possible that it's not a WASM file, could be a par.gz or any other kind of file actually but I added a more vague warning

vados-cosmonic:
No problem, didn't think this was a breaking change but if we're considering it one good thing it's getting in before 1.0

brooksmtownsend:
Yeah I think it's just _technically_ an API breaking change for wash-lib. There's already breaking changes in main so we're going to bump regardless for the next release, I just want to make sure our commits are all setup nicely so we can start evaluating tools for automatic releases, changelogs, etc etc.

crates/wash-lib/src/app.rs
vados-cosmonic: It ideally *shouldn't* happen, but previous code does not outright reject file paths that don't exist, so I didn't want to make that change here.

"reject file paths that don't exist" is definitely a worthwhile change, but can be handled in a follow up -- it's definitely a change of functionality.

brooksmtownsend (approved):

PR #1358, opened 2024-01-19
refactor(providers)!: unflatten provider errors & invocation errors

### LINKED ISSUE #719: [FEATURE] Unflatten provider errors from invocation errors

Right now we take errors from provider methods and send them as part of an invocation error. Instead, we should be serializing the `Result` type from the provider and then sending that message back.

Provider macro generation will need to be updated to send back serialized data and the `dispatch` method on the `MessageDispatch` trait in the provider SDK will need to be updated to only return an `InvocationError`.

This will also mean we need to update our adapters for the wasi-cloud interfaces to send back serialized `Result`s rather than flattening the error

vados-cosmonic:
## Feature or Problem

This commit refactors areas of provider code (SDK, in-tree providers)
that previously used `ProviderInvocationError`s which mixed
`InvocationError`s and a string-based catch-all for provider-internal
errors to return types that are true to the WIT contracts.

With this commit, provider developers must code to the interface that
matches the WIT contract (ex. `async fn operation() ->
T`), rather than having values that are wrapped in `ProviderInvocationResult`.

Contracts that were ported/not originally written with failure in mind
(i.e. not using `result<_,_>` in WIT) should be rewritten (in the
future) for operations that may fail, rather than relying on the
previously used `ProviderInvocation(Result|Error)` structures.

Resolves #719


connorsmith256 (approved):
Wow. 🚀

Issue #1091, opened 2023-11-25
Update event subscribers to use suffixed subjects


connorsmith256:
As of https://github.com/wasmCloud/wasmCloud/pull/1090, the host emits events on `wasmbus.evt.{latticeID}.{eventName}`. Some subscribers (tests in wash, the washboard, and wadm) still subscribe to the non-suffixed subject. 

These clients should be updated prior to wasmCloud 1.0

brooksmtownsend:
Hijacking this issue just to drop a nice-to-have, it would be really nice to have a defined list of all the events we publish and, optimally, some kind of JSON schema or other specifier for the expected fields on those events. Unsure if https://cloudevents.io/ has a good tool to do this

ahmedtadde:
hey @brooksmtownsend , do you have an idea/tip of how one would go about taking a survey of all events that are published? For the particular case of `wasmbus.evt.{latticeID}`, @connorsmith256 I haven't been able to located its use in `wadm` and `wash` tests (although, it seems like @vados-cosmonic already made such changes for tests some weeks ago). [ I only found one place](https://github.com/ahmedtadde/wasmCloud/blob/eb28511acc3d86a49f97616d39d9b6920824db07/washboard-ui/src/lattice/lattice-service.ts#L137) in `washboard-ui` where it is being used.

brooksmtownsend:
@ahmedtadde that's a good question. We do want to have a place where we officially note all of the events that are published, but for now it's best discovered by looking for calls to `publish_event`.

1. `host_heartbeat`
2. `host_started`
3. `host_stopped`
4. `actor_start_failed`
5. `actors_start_failed`
6. `provider_started`
7. `provider_start_failed`
8. `provider_stopped`
9. `config_set`
10. `config_deleted`
11. `linkdef_set`
12. `linkdef_deleted`
14. `actors_started`
16. `actors_stopped`

Soon to be added
1. `actor_scaled`
2. `actor_scale_failed`

There but soon to be gone 
13. `actor_started`
15. `actor_stopped`

connorsmith256:
- [ ] wadm https://github.com/wasmCloud/wadm/pull/241
- [ ] control interface and wash https://github.com/wasmCloud/wadm/pull/241
- [ ] washboard @lachieh

lachieh:
Clients are all done, good to close?

connorsmith256:
@lachieh yep!

PR #1381, opened 2024-01-24
Feat/update event subjects

connorsmith256:
## Feature or Problem
This updates the control interface and wash-lib to use the suffixed event subject. Since I modified a public function in the control interface, this is technically a breaking change. However, wash-lib is the only known client using this function

This also fixes an issue where wash would report `Timed out waiting for applicable event, operation may have failed`. The latest host emits the scale events, but wash was still looking for the start/stop events

## Related Issues
https://github.com/wasmCloud/wasmCloud/issues/1091

## Release Information
Next

### Manual Verification
Manually verified that `wash start actor` and `wash stop actor` received the correct events

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -10,6 +10,7 @@ mod broker;
 mod otel;
 mod types;
 
+use async_nats::Subscriber;
 pub use types::*;
 
 use core::fmt::{self, Debug};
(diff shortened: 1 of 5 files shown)


brooksmtownsend (approved):
Love the changes and the fix!

crates/control-interface/src/lib.rs
```diff
@@ -850,7 +826,7 @@ mod tests {
             .timeout(Duration::from_millis(1000))
             .auction_timeout(Duration::from_millis(1000))
             .build();
-        let mut receiver = client.events_receiver().await.unwrap();
+        let mut receiver = client.events_receiver("foobar".to_string()).await.unwrap();
```
brooksmtownsend, line 829: this just needs to be `vec!["foobar".to_string()]`

connorsmith256: Whoops! TY

Issue #1377, opened 2024-01-23
[FEATURE] Build required infra to enable E2E testing for washboard


lachieh:
### Is your feature request related to a problem? Please describe.

I'd like to get a base starting point to run some tests against, specifically to be able to run E2E tests against the washboard-ui. It would need to have at least the following up and running:
- 2+ wasmCloud hosts
- 2+ actors
- 2+ providers
- 2+ links

Looking at the [PetClinic](https://github.com/wasmCloud/examples/tree/main/petclinic) example, it seems that it may cover all these points if we add a second host to the cluster that starts.

### Describe the solution you'd like

Given that this is a specific use case, I'm not sure that updating the example is ideal, but I'd like to be able to run the example with minimal effort so that it can be run in CI. The example doesn't _have_ to be PetClinic, but it covers a lot of the requirements already, so it's probably a good starting point. ~The main thing here is that the NATS server needs to be configured with websockets enabled which requires a config file to be supplied.~ (this is the default configuration with `wash@0.26`).

vados-cosmonic:
Hey @lachieh for this, what do you think about setting up the e2e testing infra first? 

If we can run the E2E tests from `node` we can shell out and run `wash` along with the other deps (even via docker compose if we absolutely needed to) as set up for the E2E test suite. For configuring the workloads now that we have `wadm` that's also really easy to shell out and accomplish. 

Ends up looking very similar to the usual setup that we'd have to do anyway -- but I wasn't sure if the E2E machinery (namely the choice between `vitest`/`cypress`/etc plus `playwright`/`webdriver-io`/`puppeteer`/... for E2E tests in particular) is already made

lachieh:
@vados-cosmonic Playwright is already configured for E2E tests but there is only a basic test right now. I'd imagine starting the infra for the tests should be independent from the testing tools, so I'm down with doing that also. Ideally we should be able to trigger scripts/commands that do what is necessary for the different tests. For example:
- start washboard > assert empty
- start application from `wash` > assert components exist.

There isn't anything decided for unit/integration tests yet. I'd imagine vitest would be the easiest choice, but it's not a hard requirement and shouldn't really require any infra anyway.

lachieh:
I will add, I don't believe petclinic works on 1.0 yet, but the exact application is not the critical goal of the issue.

vados-cosmonic:
[...]

PR #1384, opened 2024-01-25
feat(docker): add websocket example

lachieh:
## Feature or Problem
Adds a docker example with a websocket enabled

## Related Issues
Semi-related to #1377


brooksmtownsend (approved):
Only request would be for us to update the README in the `examples/docker` folder soon to account for what compose files are for what 😄

connorsmith256 (approved):

PR #1386, opened 2024-01-26
ci: add conditions to rust build code

lachieh:
## Feature or Problem
After adding a path filter to the `wasmcloud` job, I noticed that PRs with only changes in the `washboard-ui` directory were stuck in a pending state because of [this issue](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore):
> If a workflow is skipped due to branch filtering, [path filtering](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), or a [commit message](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

## Solution
This PR adds a new `_file-changes.yml` workflow which categorizes changed files into groups, and then surfaces those groups as flags if the contents in those files changed.

Note that this is a temporary solution until a repo re-structure takes place and we can consolidate some of the pipeline steps and change the checks that are required for PRs to merge.

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -2,9 +2,8 @@ name: wasmCloud
 
 on:
   pull_request:
-    paths-ignore:
-      - washboard-ui/**
   merge_group:
+  workflow_dispatch:
   push:
     branches:
     - main
@@ -41,13 +40,15 @@ on:
     - 'wash-cli-v[0-9].[0-9]+.[0-9]+-*'
     - 'wash-lib-v[0-9].[0-9]+.[0-9]+'
     - 'wash-lib-v[0-9].[0-9]+.[0-9]+-*'
-  workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
+  check_for_changes:
+    uses: ./.github/workflows/_check-for-changes.yml
+    
   build-bin:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/wasmcloud.yml
```diff
@@ -102,25 +102,33 @@ jobs:
...
 
     name: wasmcloud-${{ matrix.config.target }}
+    needs: check_for_changes
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4.1.1
     - uses: ./.github/actions/install-nix
+      # need to run condition inside job steps so that job will pass if all steps are skipped
+      # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore
+      if: ${{ needs.check_for_changes.outputs.rust == 'true' }}
```
rvolosatovs, line 112: This should run if anything in Nix flake changes, Rust source changes and if GitHub action spec itself has changed.
Relevant for other occurrences as well.

.github/workflows/wasmcloud.yml
```diff
@@ -135,7 +143,10 @@ jobs:
 
   build-lipo:
     name: wasmcloud-universal-darwin
-    needs: build-bin
+    needs:
+      - build-bin
+      - check_for_changes
+    if: ${{ needs.check_for_changes.outputs.rust == 'true' }}
```
rvolosatovs, line 149: This will have to account for Nix and GitHub actions as well

.github/workflows/wasmcloud.yml
lachieh: ok, yeah good point, will update

lachieh: Ok, so I was already accounting for the flake changes in `.nix` but I've added github to that too. Combined them both into a `ci` check and added that into the rust check:
https://github.com/wasmCloud/wasmCloud/blob/5e57c5f7ca6d31c62f04fa9bda5683f0b45a801c/.github/workflows/_check-for-changes.yml#L74-L88

https://github.com/wasmCloud/wasmCloud/blob/5e57c5f7ca6d31c62f04fa9bda5683f0b45a801c/.github/workflows/_check-for-changes.yml#L44

https://github.com/wasmCloud/wasmCloud/blob/5e57c5f7ca6d31c62f04fa9bda5683f0b45a801c/.github/workflows/_check-for-changes.yml#L33-L35

connorsmith256 (approved):

PR #1407, opened 2024-01-29
fix(host): add subject to control interface logs

connorsmith256:
This adds the `subject` back to log messages related to processing control interface commands, which is helpful when debugging messages from incompatible clients. The `message` was on the span already, but only when trace level logs are enabled


connorsmith256:
FYI @thomastaylor312 since we discussed this at the time we removed `subject` from the inline logs: it turns out it's useful to be able to see the subject even when tracing on a given host is disabled/the log level is less verbose than `trace` 😅

brooksmtownsend (approved):

Issue #998, opened 2023-11-13
[FEATURE] log level toggle with a control command


protochron:
**Is your feature request related to a problem? Please describe.**
Right now the only way to change the log and tracing level in the wasmcloud host is to update the environment variable or flag and restart the process. This means that if you're trying to actively debug an issue as it is occurring, you will need to redeploy your host which may mask any issues or make it harder to reproduce a problem

**Describe the solution you'd like**
The wasmcloud host should support a control interface command to toggle the verbosity of the logging level at runtime. There is plenty of prior art in this space, ex. NATS now supports a command to dump a pprof with a SYS topic command.

**Describe alternatives you've considered**

**Additional context**

thomastaylor312:
Another possible alternative: using a feature flagging library of some kind. I recently learned about https://openfeature.dev/, but still need to evaluate it to see if it could work

protochron:
@thomastaylor312 I don't have a strong preference as long as it's something that can be changed at runtime. I proposed the control interface topic since it's what we have now, but if something like OpenFeature works better then I'm all for it

Issue #1108, opened 2023-11-28
[RFC] Formalize the wasmCloud Control Interface


brooksmtownsend:
# RFC: Formalize the wasmCloud Control Interface

## Introduction
The wasmCloud control interface is how any client can instruct a wasmCloud host to take an action over NATS. This can be anything from adding or removing labels, starting and stopping actors/providers, or even shutting down the host itself. In a sense, the control interface is the API server of wasmCloud, and as we approach 1.0 it should be scrutinized and stabilized in a way that protects us from future change.

## Motivation

Currently there are X aspects of the control interface that I think we should look at and consider before we call the control interface protocol "stable" for 1.0:
1. The topic format should be standard where possible, e.g. `wasmbus.ctl.<version>.<lattice>.<noun>.<verb>.<host_id>`
2. The topic format should be versioned by an API version (see `<version>` above) to protect from future changes that are over-the-wire incompatible
3. Our set of operations on the control interface should encompass the foreseeable list of operations that we want to have for interacting with the host

## Detailed Design

You can see a list of all control interface topics at https://wasmcloud.com/docs/hosts/lattice-protocols/control-interface, but for enumeration you can see them below. Similar to the redesign in `wash-cli`, I propose that we follow a noun -> verb ordering for control interface topics. Regarding 2 above, the version should be present after the standard topic prefix. Additionally, I'm proposing we add two additional control interface topics which are up for discussion.

| Current                                       | Proposed |
|-----------------------------------------------|----------|
| `wasmbus.ctl.{lattice_prefix}.auction.provider` |     `wasmbus.ctl.{version}.{lattice_prefix}.provider.auction`     |
| `wasmbus.ctl.{lattice_prefix}.auction.actor`    |      `wasmbus.ctl.{version}.{lattice_prefix}.actor.auction`    |
| `wasmbus.ctl.{lattice_prefix}.cmd.{host-id}.la` (deprecated) |     Remove in favor of scale    |
| `wasmbus.ctl.{lattice_prefix}.cmd.{host-id}.sa` |      Remove in favor of scale    |
| `wasmbus.ctl.{lattice_prefix}.cmd.{host-id}.scale` |     `wasmbus.ctl.{version}.{lattice_prefix}.actor.scale.{host_id}`     |
| `wasmbus.ctl.{lattice_prefix}.cmd.{host-id}.upd` |     `wasmbus.ctl.{version}.{lattice_prefix}.actor.update.{host_id}`     |
[...]

PR #1412, opened 2024-01-29
Feat/dynamic log level

connorsmith256:
## Feature or Problem
Today it's not possible to update the host's log level at runtime, which can make it impossible to debug issues which go away after restarts. This updates the host to listen on a subject and update its log level (and a companion subject for simply returning the current log level), which included a lot of plumbing. I also added `wash logs get/set` for good measure

**NOTE**: I tried to follow the subject convention outlined in https://github.com/wasmCloud/wasmCloud/issues/1108, so the new subjects might look out of place compared to the status quo.

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/998

## Release Information
Next, for `core`, `tracing`, `control-interface`, the host, `wash-lib`, and `wash-cli` 😅 

## Consumer Impact
N/A, new feature

### Manual Verification
Tested e2e with new builds:
```
./target/debug/wasmcloud
2024-01-29T23:21:50.310400Z  INFO async_nats::options: event: connected
2024-01-29T23:21:50.310410Z  INFO async_nats::options: event: connected
2024-01-29T23:21:50.311812Z  INFO wasmcloud_host::wasmbus: bucket already exists. Skipping creation. bucket=LATTICEDATA_default
2024-01-29T23:21:50.312093Z  INFO wasmcloud_host::wasmbus: bucket already exists. Skipping creation. bucket=CONFIGDATA_default
2024-01-29T23:21:50.315665Z  INFO wasmcloud_host::wasmbus: wasmCloud host started host_id="NAHSIKVUXY6THUZRHNG2KWCR725CRT3YEN2XWSDB62DDYOID6BYS4NLR"
2024-01-29T23:22:16.507724Z  INFO wasmcloud_host::wasmbus: updating log level level=Trace
2024-01-29T23:22:16.510618Z TRACE wasmcloud_host::wasmbus: handled control interface request
2024-01-29T23:22:20.281526Z TRACE heartbeat: wasmcloud_host::wasmbus: generating heartbeat
2024-01-29T23:22:20.281713Z TRACE heartbeat:inventory: wasmcloud_host::wasmbus: generating host inventory
2024-01-29T23:22:50.269039Z TRACE heartbeat: wasmcloud_host::wasmbus: generating heartbeat
2024-01-29T23:22:50.269222Z TRACE heartbeat:inventory: wasmcloud_host::wasmbus: generating host inventory
```

```
~/wasmcloud/wasmcloud/target/debug/wash logs get snowy

Host `snowy-dream-3908` configured with log level `info`

~/wasmcloud/wasmcloud/target/debug/wash logs set snowy -l trace

Host `snowy-dream-3908` configured with log level `trace`
```

### DIFF
--- a/crates/wash-lib/src/cli/logs.rs
+++ b/crates/wash-lib/src/cli/logs.rs
@@ -1,0 +1,103 @@
+use anyhow::{bail, Result};
+use clap::Parser;
+
+use crate::{
+    cli::{CliConnectionOpts, CommandOutput},
+    common::{boxed_err_to_anyhow, find_host_id},
+    config::WashConnectionOptions,
+};
+
+#[derive(Clone, Debug, Parser)]
+pub struct GetLoggingConfigCommand {
+    #[clap(flatten)]
+    pub opts: CliConnectionOpts,
+
+    /// ID of host to update the actor on. If a non-ID is provided, the host will be selected based
+    /// on matching the prefix of the ID or the friendly name and will return an error if more than
+    /// one host matches.
+    #[clap(name = "host-id")]
+    pub host_id: String,
+}
+
+#[derive(Clone, Debug, Parser)]
+pub struct SetLoggingConfigCommand {
+    #[clap(flatten)]
+    pub opts: CliConnectionOpts,
+
+    /// ID of host to update the actor on. If a non-ID is provided, the host will be selected based
+    /// on matching the prefix of the ID or the friendly name and will return an error if more than
+    /// one host matches.
+    #[clap(name = "host-id")]
+    pub host_id: String,
+
+    /// Logging level to set
+    #[clap(short = 'l', long = "level")]
+    pub level: String,
+}
+
+#[derive(Debug, Clone, Parser)]
+pub enum LogsCommand {
+    /// Get the logging config for a host
+    #[clap(name = "get")]
+    Get(GetLoggingConfigCommand),
+
+    /// Set the logging config for a host
+    #[clap(name = "set")]
+    Set(SetLoggingConfigCommand),
+}
+
+pub async fn handle_get_logging_config(cmd: GetLoggingConfigCommand) -> Result<CommandOutput> {
+    let wco: WashConnectionOptions = cmd.opts.try_into()?;
+    let client = wco.into_ctl_client(None).await?;
+
+    let (host_id, friendly_name) = find_host_id(&cmd.host_id, &client).await?;
+
+    let friendly_name = if friendly_name.is_empty() {
+        host_id.to_string()
+    } else {
+        friendly_name
+    };
+
+    let config = client
+        .get_logging_config(&host_id)
+        .await
+        .map_err(boxed_err_to_anyhow)?;
+
+    Ok(CommandOutput::from_key_and_text(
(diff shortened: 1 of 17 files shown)


crates/wash-lib/src/cli/logs.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct GetLoggingConfigCommand {
+    #[clap(flatten)]
+    pub opts: CliConnectionOpts,
+
+    /// ID of host to update the actor on. If a non-ID is provided, the host will be selected based
+    /// on matching the prefix of the ID or the friendly name and will return an error if more than
+    /// one host matches.
+    #[clap(name = "host-id")]
+    pub host_id: String,
+}
```
brooksmtownsend, line 20: When I read `wash logs get snowy`, as a user I think "I want to retrieve the logs for a specific host", not "I want to retrieve the log level for a specific host". Given the fact that `set` sets the log level and it doesn't make sense to "set" the logs for a host, I could reason about this, but did you consider any alternatives for the top level command that may disambiguate this?

My biggest issue is just that users will likely look at `wash logs` and think they can retrieve logs, but find out that they can't after inspecting it.

brooksmtownsend (commented):
Huge fan of the functionality, implementation, and topic format. I do have a concern around interpretation of the command on the wash side so leaving that comment for discussion

crates/wash-lib/src/cli/logs.rs
connorsmith256: I did think about this, and don't love where I ended up. You're right that `wash logs get` reads more like "get the logs for this host", which might be a useful feature, but is not what this is intended for.

I'm trying to avoid breaking changes shortly after 1.0. Today, the only use case this new feature enables is overriding the log level. But there might be other functionality we want, e.g. reconfiguring the OTEL endpoint, or updating sampling configuration, etc. We don't have any of that functionality today, but I wanted to make the command (and underlying payload) generic enough that we could simply add options to it, rather than scrapping it in favor of an entirely new command/subject/payload/etc.

In that light, can you help me think of a better name? Maybe `wash otel [get|set]`? I suppose since this is an advanced debug feature that won't be used all that often, we could even afford to be verbose: `wash otel-config get`? (`otel` is motivated by my journey into OTEL logging land today). E.g.:
- `wash otel-config get --host-id foo log-level`
- `wash otel-config set --host-id bar trace-sampling 0.1`
- `wash otel-config set --host-id baz log-format structured`

brooksmtownsend: This honestly sounds like a broader story around setting host configuration at runtime, like we've been setting up labels and log levels and perhaps we need a more general purpose set of config vars... I will noodle on this a bit but I don't have a great answer right now. Maybe the command we could use is "wash log level set|get"? That is intuitive but doesn't provide extensibility

thomastaylor312:
@brooksmtownsend I assume we want to revisit this post 1.0?

brooksmtownsend:
I think so, this feels like a good change to be able to make even with our idea of hosts being cattle not pets. The conflicts seem to be pretty huge at this point so it would be worth re-doing

PR #1414, opened 2024-01-29
refactor(host): move label parsing out of host library

connorsmith256:
## Feature or Problem
This is a simple refactor which moves the parsing logic for labels set via env vars to the main binary, rather than the host library.

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/757

## Release Information
Next

## Consumer Impact
v0.82.0. I also removed support for the (very old) `HOST_` prefix, so this is technically a breaking change

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,7 @@
 #![warn(clippy::pedantic)]
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::env;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -318,13 +319,28 @@ async fn main() -> anyhow::Result<()> {
...
+        } else {
+            return None;
+        };
+        if labels_from_args.contains(&key) {
+            warn!(
+                ?key,
+                "label provided via environment variable will override label set via args"
+            );
+        }
+        Some((key, value))
```
brooksmtownsend, line 342: Is this the common order of precedence that we normally take? e.g. 
```
WASMCLOUD_LABEL_foo=bar ./wasmcloud --label foo=baz
```
This code reads like `bar` would be the value of `foo`, but I would argue that we should take the arg value over the environment variable. It may be confusing to have something like 
```
# ~/.zshrc
export WASMCLOUD_LABEL_me=$(whoami)
# later
wash up --label me=test
```

I'm good either way, so long as we're consistent, and I lean towards the arg taking precedence as it's explicitly provided at runtime

connorsmith256: Nope, you're right, I got this backward. Fixed!

brooksmtownsend (approved):

Issue #857, opened 2023-10-02
[FEATURE] wash app status


aish-where-ya:
Currently when a wadm manifest is deployed, it displays the "deployed" message but if it fails, you can only see the status via `wash app list` . 
This does not display the specific error with the deployment and you need to print it into a json format to view the error `wash app list -o json` (or text).

This command however, displays all the apps and their statuses in a json format. To inspect a specific app, `wash app status` will help us to look into the details of only the desired app. Along with this, we can decide if we want to publish the logs related to this app when we run `wash app status`.

PR #1421, opened 2024-01-30
feat(wash-cli): add support for wash app status

connorsmith256:
## Feature or Problem
This adds support for `wash app status [model]`

Given the verbosity of the status message, I'm not sure I like the table view, so I'm open to suggestions

## Related Issues
Resolves https://github.com/wasmCloud/wasmCloud/issues/857

## Release Information
Next minor bump

## Consumer Impact
N/A

### Manual Verification
```
~/wasmcloud/wasmcloud/target/debug/wash app status typescript-http-hello-world -o json

{
  "status": {
    "message": "Successfully fetched status for model typescript-http-hello-world",
    "result": "ok",
    "status": {
      "components": [],
      "status": {
        "message": "Linkdef pending, waiting for file:///Users/connor/Documents/wasmCloud/wasmCloud/examples/typescript/actors/http-hello-world/build/index_s.wasm and wasmcloud.azurecr.io/httpserver:0.19.1 to start",
        "type": "reconciling"
      },
      "version": "v0.0.1"
    }
  },
  "success": true
}
```
```
~/wasmcloud/wasmcloud/target/debug/wash app status typescript-http-hello-world

                                                                                                                                                                                   
  Name                          Deployed Version   Deploy Status   Status Message                                                                                                  
  typescript-http-hello-world   v0.0.1             Reconciling     Linkdef pending, waiting for file:///Users/connor/Documents/wasmCloud/wasmCloud/examples/typescript/actors/http-hello-world/build/index_s.wasm and wasmcloud.azurecr.io/httpserver:0.19.1 to start
```

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -5,7 +5,7 @@ use clap::{Args, Subcommand};
 use serde_json::json;
 use wadm::server::{
     DeleteModelResponse, DeployModelResponse, GetModelResponse, GetResult, ModelSummary,
-    PutModelResponse, PutResult, VersionResponse,
+    PutModelResponse, PutResult, StatusResponse, VersionResponse,
 };
 use wash_lib::{
     app::{load_app_manifest, AppManifest},
@@ -25,6 +25,9 @@ pub enum AppCliCommand {
     /// Retrieve the details for a specific version of an app specification
     #[clap(name = "get")]
     Get(GetCommand),
+    /// Retrieve the status of a given model within the lattice
+    #[clap(name = "status")]
+    Status(StatusCommand),
     /// Retrieve the version history of a given model within the lattice
     #[clap(name = "history")]
     History(HistoryCommand),
@@ -118,6 +121,16 @@ pub struct GetCommand {
 }
 
 #[derive(Args, Debug, Clone)]
+pub struct StatusCommand {
+    /// The name of the app spec
+    #[clap(name = "name")]
+    model_name: String,
+
+    #[clap(flatten)]
+    opts: CliConnectionOpts,
+}
+
+#[derive(Args, Debug, Clone)]
 pub struct HistoryCommand {
     /// The name of the app spec
     #[clap(name = "name")]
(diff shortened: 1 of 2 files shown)


crates/wash-cli/src/app/mod.rs
```diff
@@ -341,3 +370,12 @@ fn show_model_history(results: VersionResponse) -> CommandOutput {
     map.insert("revisions".to_string(), json!(results));
     CommandOutput::new(output::list_revisions_table(results.versions), map)
 }
+
+fn show_model_status(model_name: String, results: StatusResponse) -> CommandOutput {
+    let mut map = HashMap::new();
+    map.insert("status".to_string(), json!(results));
```
vados-cosmonic, line 376: NIT: you could probably get away with `HashMap::from(...)` instead of making this temp var

crates/wash-cli/src/app/mod.rs
```diff
@@ -117,6 +120,16 @@ pub struct GetCommand {
     opts: CliConnectionOpts,
 }
 
+#[derive(Args, Debug, Clone)]
+pub struct StatusCommand {
+    /// The name of the app spec
+    #[clap(name = "name")]
+    model_name: String,
+
+    #[clap(flatten)]
```
vados-cosmonic, line 129: To address the overwide output problem, maybe we should introduce some sort of flattened options that represent output formatting? Similar to `CliConnectionOpts`. Unfortunately it'd have to flow through the command to the output function (`status_table`) but it might be reasonable to truncate by default (so people get a sense of the command and show the whole thing. 

Definitely can be done in a follow up though or if someone finds it really troublesome -- AFAIK we don't do precise screen width checking to ensure we don't overflow in the first place... IMO better to move fast and overflow for now then go back in and add by-default truncation with the option to print everything.

vados-cosmonic (approved):
LGTM :rocket: 

Some comments but none are blocking

crates/wash-cli/src/app/mod.rs
connorsmith256: Definitely could, but we do it this way alllll over the place in wash, and I think for good reason. The `map.insert` calls make it easier to visually see each entry we put in the map. So at the very least for consistency, I'm going to leave this as-is

crates/wash-cli/src/app/mod.rs
connorsmith256: Agreed, more smartness/optionality later is better. So long as the table columns make sense, I'm happy to ship this

PR #1423, opened 2024-01-30
feat: add TLS CA support to wash

protochron:
## Feature or Problem
NATS servers support TLS, which means sometimes you need to be able to provide certificates when instantiating a client.

This adds support for providing a CA certificate to `wash` with a command line flag. I opted not to add mTLS support by adding flags for a certificate and a key, but it would not be that hard now that the scaffolding is in place. I also chose not to add the same support to the wasmCloud host since you can simply add a CA to a leaf running as a sidecar, which is a understood deployment model. However that would also be a relatively straightforward change by adding the necessary paths to the control and RPC connections.

## Testing
Tested locally with a NATS server using a self-signed CA and certificate.


crates/wash-cli/src/up/config.rs
```diff
@@ -31,13 +31,15 @@ pub const WASMCLOUD_RPC_JWT: &str = "WASMCLOUD_RPC_JWT";
...
 pub const WASMCLOUD_RPC_TLS: &str = "WASMCLOUD_RPC_TLS";
+pub const WASMCLOUD_RPC_TLS_CA_FILE: &str = "WASMCLOUD_RPC_TLS_CA_FILE";
 // NATS CTL connection configuration
 pub const WASMCLOUD_CTL_HOST: &str = "WASMCLOUD_CTL_HOST";
 pub const WASMCLOUD_CTL_PORT: &str = "WASMCLOUD_CTL_PORT";
 pub const WASMCLOUD_CTL_SEED: &str = "WASMCLOUD_CTL_SEED";
 pub const WASMCLOUD_CTL_JWT: &str = "WASMCLOUD_CTL_JWT";
 pub const WASMCLOUD_CTL_CREDSFILE: &str = "WASMCLOUD_CTL_CREDSFILE";
 pub const WASMCLOUD_CTL_TLS: &str = "WASMCLOUD_CTL_TLS";
+pub const WASMCLOUD_CTL_TLS_CA_FILE: &str = "WASMCLOUD_CTL_TLS_CA_FILE";
```
connorsmith256, line 42: Nit: these env vars use "TLS_CA_FILE" but the args just use `-ca-file`. Can we have "tls" in both or neither?

crates/wash-lib/src/cli/mod.rs
```diff
@@ -160,6 +160,10 @@ pub struct CliConnectionOpts {
     #[clap(long = "ctl-credsfile", env = "WASH_CTL_CREDS", hide_env_values = true)]
     pub ctl_credsfile: Option<PathBuf>,
 
+    /// TLS CA file for CTL authentication. See https://docs.nats.io/using-nats/developer/connecting/tls for details.
+    #[clap(long = "ca-file", env = "WASH_CA_FILE", hide_env_values = true)]
+    pub ca_file: Option<PathBuf>,
```
connorsmith256, line 165: Nit: `ctl_ca_file` and `WASH_CTL_CA_FILE`? (related to other comment as well)

crates/wash-lib/src/config.rs
```diff
@@ -62,6 +62,9 @@ pub struct WashConnectionOptions {
     /// See https://docs.nats.io/using-nats/developer/connecting/creds for details.
     pub ctl_credsfile: Option<PathBuf>,
 
+    /// Path to a file containing a CA certificate to use for TLS connections
+    pub ca_file: Option<PathBuf>,
```
connorsmith256, line 66: `ctl_ca_file`?

connorsmith256 (commented):
Just some nits related to naming

It's _okay_ if we bump the versions in Cargo.toml in this PR, but unless we need to release this ASAP, I'd prefer to wait until we're ready to release, so we don't accidentally skip a version (we've done that before)

thomastaylor312:
@protochron this must have slipped though the cracks. Did you still want this in wash?

protochron:
Yes, we still need this in wash

PR #1426, opened 2024-01-30
Feat/otel logs

connorsmith256:
## Feature or Problem
This adds support for exporting OTEL logs over OTLP. In doing so, I refactored and simplified the "tracing" logic, which should make it easier to add metrics support as well (@joonas shared a POC of that work today)

The wasmCloud host now supports:
- exporting OTEL traces, when the `otlp` exporter is provided
- exporting OTEL logs, when the `otlp` exporter is provided
- logging to stderr in JSON, when the structured logging flag is enabled
- logging to stderr in plaintext, by default

(Note that there's no conflict between logging locally and exporting logs via OTEL)

"Tracing" is in quotes because it's ambiguous whether "tracing" refers to the `tracing` crate, which we use to emit logs _and_ events inside spans, or distributed traces/tracing.

Since wasmCloud is all-in on OTEL, I think we should rename this crate to `wasmcloud-otel`, but I didn't do that yet, in case others were opposed

## Release Information
Next. No breaking changes, although now it should be clear that the `OTEL_EXPORTER_OTLP_ENDPOINT` env var/arg should _not_ include a path like `/v1/traces`. It should just be the address of the server handling the given type of data. The underlying otel crates handle appending the appropriate path.

## Consumer Impact
N/A

### Manual Verification
So, so, so much manual testing. It's hard to display that here, but in summary I:
- set things up
  - started tempo using the [auxiliary](https://github.com/wasmCloud/wasmCloud/blob/main/examples/docker/docker-compose-auxiliary.yml) docker file
  - ran `otel/opentelemetry-collector:0.93.0` on another port
- confirmed that the host handles every combination of args and env vars related to OTEL
- confirmed that the host was able to log to stderr without impacting OTEL exports
- confirmed that the host could export logs and traces to separate servers

### DIFF
--- a/crates/tracing/src/lib.rs
+++ b/crates/tracing/src/lib.rs
@@ -6,9 +6,14 @@ pub mod context;
 use std::env;
 use std::io::{IsTerminal, StderrLock, Write};
 
+#[cfg(feature = "otel")]
+use anyhow::bail;
 use anyhow::Context;
+#[cfg(feature = "otel")]
 use heck::ToKebabCase;
 use once_cell::sync::OnceCell;
+#[cfg(feature = "otel")]
+use opentelemetry_otlp::WithExportConfig;
 use tracing::{Event, Subscriber};
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::fmt::format::{DefaultFields, Format, Full, Json, JsonFields, Writer};
@@ -53,7 +58,7 @@ impl<'a> Drop for LockedWriter<'a> {
 static STDERR: OnceCell<std::io::Stderr> = OnceCell::new();
 
 #[cfg(feature = "otel")]
-const DEFAULT_TRACING_ENDPOINT: &str = "http://localhost:4318/v1/traces";
+static LOG_PROVIDER: OnceCell<opentelemetry_sdk::logs::LoggerProvider> = OnceCell::new();
 
 /// A struct that allows us to dynamically choose JSON formatting without using dynamic dispatch.
 /// This is just so we avoid any sort of possible slow down in logging code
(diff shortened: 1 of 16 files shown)


brooksmtownsend:
Turns out we didn't have an OTEL logs issue for 1.0 so, marked this PR in the roadmap to account for it

examples/docker/docker-compose-auxiliary.yml
```diff
@@ -1,4 +1,4 @@
-# This docker-compose file starts supporting services for a wasmCloud ecosystem, including: 
+# This docker-compose file starts supporting services for a wasmCloud ecosystem, including:
```
brooksmtownsend, line 1: I think we should turn this `auxillary` docker compose into a `docker-compose-otel.yml` file for clarity, and add your collector container that you tested with. Thoughts? I'm fine with this being a follow up, I want it for metrics too

brooksmtownsend (approved):

crates/tracing/src/lib.rs
```diff
@@ -197,35 +211,88 @@ pub fn configure_tracing(
...
+        .with_batch_exporter(exporter, opentelemetry_sdk::runtime::Tokio)
+        .build();
+
+    // Prevent the exporter/provider from being dropped
+    LOG_PROVIDER
+        .set(log_provider)
+        .map_err(|_| anyhow::anyhow!("Logger provider already initialized"))?;
+
+    let log_layer = opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge::new(
+        LOG_PROVIDER.get().unwrap(),
```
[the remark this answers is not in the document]
connorsmith256: Always safe because the `set` happens 1 line up from here

examples/docker/docker-compose-auxiliary.yml
connorsmith256: Hmm, I'm not sure. This file also sets up the registry container, which doesn't relate to OTEL. And I didn't swap out/include the `otel/opentelemetry-collector` image because AFAIK there's no UI for it. It isn't really useful on its own, beyond local testing to confirm that OTEL data is being collected

brooksmtownsend: I think we can approach again once we have metrics in place, I'd be fine with having just a `docker-compose-otel.yml` file. Goal would just be to have some distinct examples to show how you can launch different services alongside wasmcloud. We should have a minimal example too with just NATS.

brooksmtownsend: https://github.com/wasmCloud/wasmCloud/pull/1445

thomastaylor312 (approved):

PR #1427, opened 2024-01-31
chore(brand): add logo files

lachieh:
## Feature or Problem
This PR adds all the most recent and correct versions of the logo files, moving them from https://github.com/wasmcloud/branding into the monorepo


connorsmith256:
@lachieh should we archive https://github.com/wasmcloud/branding after this merges then?

lachieh:
@connorsmith256 I totally missed that one. Yeah, let's make it monorepo.

brooksmtownsend (approved):

PR #1431, opened 2024-02-01
feat: Add initial support for metrics

joonas:
## Feature or Problem

In support of #664, this lays the initial foundation to be able to add support for various metrics outlined in the RFC for the wasmcloud-host.

It's worth mentioning that the goal of this PR is not to add comprehensive coverage of all the possible metrics outlined in #664, but instead lay the foundation to allow anyone to come in and add them as it makes sense.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -35,6 +35,7 @@ use tokio_stream::wrappers::IntervalStream;
 use tracing::{debug, error, info, instrument, trace, warn};
 use ulid::Ulid;
 use uuid::Uuid;
+use wasmcloud_tracing::{global, KeyValue};
 
 pub use config::Host as HostConfig;
 use wascap::{jwt, prelude::ClaimsBuilder};
@@ -59,7 +60,7 @@ use wasmcloud_runtime::Runtime;
 use wasmcloud_tracing::context::TraceContextInjector;
 
 use crate::{
-    fetch_actor, socket_pair, OciConfig, PolicyAction, PolicyHostInfo, PolicyManager,
+    fetch_actor, socket_pair, HostMetrics, OciConfig, PolicyAction, PolicyHostInfo, PolicyManager,
     PolicyRequestSource, PolicyRequestTarget, PolicyResponse, RegistryAuth, RegistryConfig,
     RegistryType,
 };
@@ -268,6 +269,7 @@ struct ActorInstance {
     actor_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::Actor>>>>,
     // TODO: use a single map once Claims is an enum
     provider_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::CapabilityProvider>>>>,
+    metrics: Arc<HostMetrics>,
 }
 
 impl Deref for ActorInstance {
@@ -1443,7 +1445,20 @@ impl ActorInstance {
                 let target = invocation.target.clone();
                 let operation = invocation.operation.clone();
 
+                let start_at = Instant::now();
                 let res = self.handle_call(invocation).await;
+                let elapsed = u64::try_from(start_at.elapsed().as_nanos()).unwrap_or_default();
+
+                self.metrics
+                    .wasmcloud_host_handle_rpc_message_duration_ns
+                    .record(
+                        elapsed,
+                        &[
+                            KeyValue::new("actor.ref", self.image_reference.clone()),
+                            KeyValue::new("operation", operation.clone()),
+                        ],
+                    );
+
                 match res {
                     Ok((msg, content_length)) => InvocationResponse {
                         msg,
@@ -1584,6 +1599,7 @@ pub struct Host {
     actor_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::Actor>>>>, // TODO: use a single map once Claims is an enum
     provider_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::CapabilityProvider>>>>,
     config_data_cache: Arc<RwLock<ConfigCache>>,
+    metrics: Arc<HostMetrics>,
 }
 
 #[allow(clippy::large_enum_variant)] // Without this clippy complains actor is at least 0 bytes while provider is at least 280 bytes. That doesn't make sense
@@ -2023,6 +2039,17 @@ impl Host {
         )
         .await?;
 
+        let meter = global::meter_with_version(
+            "wasmcloud-host",
+            Some(env!("CARGO_PKG_VERSION")),
+            None::<&str>,
+            Some(vec![
+                KeyValue::new("host.id", host_key.public_key()),
+                KeyValue::new("host.version", env!("CARGO_PKG_VERSION")),
+            ]),
+        );
+        let metrics = HostMetrics::new(&meter);
+
         let host = Host {
             actors: RwLock::default(),
             chunk_endpoint,
@@ -2054,6 +2081,7 @@ impl Host {
             actor_claims: Arc::default(),
             provider_claims: Arc::default(),
             config_data_cache: Arc::default(),
+            metrics: Arc::new(metrics),
         };
 
         let host = Arc::new(host);
@@ -2445,6 +2473,7 @@ impl Host {
                 image_reference: actor_ref.to_string(),
                 actor_claims: Arc::clone(&self.actor_claims),
                 provider_claims: Arc::clone(&self.provider_claims),
+                metrics: Arc::clone(&self.metrics),
             });
 
             let _calls = spawn({
--- a/crates/tracing/src/lib.rs
+++ b/crates/tracing/src/lib.rs
@@ -3,6 +3,17 @@
 #[cfg(feature = "otel")]
 pub mod context;
 
+pub mod metrics;
+
+pub use metrics::configure_metrics;
+
+#[cfg(feature = "otel")]
+pub use opentelemetry::{
+    global,
+    metrics::{Histogram, Meter},
+    KeyValue,
+};
+
 use std::env;
 use std::io::{IsTerminal, StderrLock, Write};
 
--- a/crates/tracing/src/metrics.rs
+++ b/crates/tracing/src/metrics.rs
@@ -1,0 +1,87 @@
+use anyhow::Context;
+use heck::ToKebabCase;
+
+#[cfg(not(feature = "otel"))]
+pub fn configure_metrics(_: String, _: &wasmcloud_core::OtelConfig) -> anyhow::Result<()> {
+    // no-op for when otel isn't enabled.
+    Ok(())
+}
+
+#[cfg(feature = "otel")]
+#[allow(clippy::missing_errors_doc)]
+#[allow(clippy::module_name_repetitions)]
+pub fn configure_metrics(
+    service_name: &str,
+    otel_config: &wasmcloud_core::OtelConfig,
+) -> anyhow::Result<()> {
+    use opentelemetry_otlp::WithExportConfig;
+    let normalized_service_name = service_name.to_kebab_case();
+
+    let mut exporter = opentelemetry_otlp::new_exporter()
+        .http()
+        .with_protocol(opentelemetry_otlp::Protocol::HttpBinary);
+
+    if let Some(ref endpoint) = otel_config.exporter_otlp_endpoint.clone() {
+        exporter = exporter.with_endpoint(endpoint);
+    }
+
+    opentelemetry_otlp::new_pipeline()
+        .metrics(opentelemetry_sdk::runtime::Tokio)
+        .with_exporter(exporter)
+        .with_resource(opentelemetry_sdk::Resource::new(vec![
+            opentelemetry::KeyValue::new("service.name", normalized_service_name),
+        ]))
+        .with_aggregation_selector(ExponentialHistogramAggregationSelector::new())
+        .with_temporality_selector(
+            opentelemetry_sdk::metrics::reader::DefaultTemporalitySelector::new(),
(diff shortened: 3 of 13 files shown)


brooksmtownsend (approved):
This looks great, and now that the plumbing is setup I can see a ton of potential for us instrumenting a bunch 😄 

The way you implemented the `handle_rpc_message` will capture pretty close to the amount of time the actor is actively handling the message, and it won't capture some of the de/serialization cost for the invocations. Is that the intent, to not include that as a part of the metric? I think that's the more useful piece to measure just want to make sure

src/main.rs
```diff
@@ -221,6 +221,10 @@ struct Args {
     #[clap(long = "otel-traces-exporter", env = "OTEL_TRACES_EXPORTER")]
     otel_traces_exporter: Option<String>,
 
+    /// Specifies which exporter to use for metrics. Only "otlp" is supported at this time
+    #[clap(long = "otel-metrics-exporter", env = "OTEL_METRICS_EXPORTER")]
+    otel_metrics_exporter: Option<String>,
```
connorsmith256, line 226: Hmm, I think we should use the `hide` clap arg for these two, and add a visible `--otel-exporter` arg that uses the same exporter for all three (metrics, logs, traces), like we have for `--otel-exporter-otlp-endpoint`. I know there's not a [standard env var](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#exporter-selection) corresponding to that, but speaking from experience, I think there's two ways users will want to set OTEL config:
- use the standard env vars (which we shouldn't get in the way of)
- just enable the simplest configuration, which is `otlp` with all 3 data types pointed at the same endpoint

crates/tracing/src/metrics.rs
```diff
@@ -0,0 +1,51 @@
...
+    use opentelemetry_otlp::WithExportConfig;
+    let normalized_service_name = service_name.to_kebab_case();
+
+    let endpoint = if let Some(ref endpoint) = otel_config.exporter_otlp_endpoint {
+        endpoint.clone()
+    } else {
+        eprintln!(
+            "OTEL exporter endpoint not set, defaulting metrics to '{DEFAULT_METRICS_ENDPOINT}'"
+        );
+        DEFAULT_METRICS_ENDPOINT.to_string()
```
connorsmith256, line 28: I don't think we should default to this ourselves. Instead, we should let the underlying library handle whether to set a custom endpoint. I accomplished this in my PR for traces/logs by using a mutable builder: https://github.com/wasmCloud/wasmCloud/pull/1426/files#diff-4416842cf20f21d591f893e02fe590b6dd14be4ecdc21c1fc9879caec85a040bR226

crates/tracing/src/lib.rs
```diff
@@ -3,6 +3,10 @@
 #[cfg(feature = "otel")]
 pub mod context;
 
+pub mod metrics;
+
+pub use metrics::configure_metrics;
```
connorsmith256, line 8: Nit: it feels weird to have metrics be a sub-module but re-exported here. Shouldn't tracing and metrics be "peers"?

crates/host/Cargo.toml
```diff
@@ -26,6 +26,7 @@ humantime = { workspace = true }
 oci-distribution = { workspace = true, features = ["rustls-tls"] }
 names = { workspace = true }
 nkeys = { workspace = true }
+opentelemetry = { workspace = true, features = ["metrics"] }
```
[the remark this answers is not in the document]
connorsmith256: Nit: how about re-exporting the needed `opentelemetry` types from the wasmCloud `tracing` crate, to avoid needing to add this dependency here

crates/core/src/lib.rs
```diff
@@ -92,8 +92,13 @@ pub type HostEnvValues = WitMap<String>;
 pub struct OtelConfig {
     /// OTEL_TRACES_EXPORTER https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/#otel_traces_exporter
     pub traces_exporter: Option<String>,
+    /// OTEL_METRICS_EXPORTER https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/#otel_metrics_exporter
+    pub metrics_exporter: Option<String>,
```
connorsmith256, line 96: Same comment as in `main`, but I don't think we need to add support for multiple exporters at this time

crates/tracing/src/metrics.rs
```diff
@@ -0,0 +1,51 @@
...
+        .with_resource(opentelemetry::sdk::Resource::new(vec![
+            opentelemetry::KeyValue::new("service.name", normalized_service_name),
+        ]))
+        .with_aggregation_selector(
+            opentelemetry::sdk::metrics::reader::DefaultAggregationSelector::new(),
+        )
+        .with_temporality_selector(
+            opentelemetry::sdk::metrics::reader::DefaultTemporalitySelector::new(),
+        )
+        .build();
```
connorsmith256, line 48: I think the otel crate errors can be converted to anyhow errors, so you should be able to do the following and drop the `map_err` from the next line
```suggestion
        .build().context("failed to create OTEL metrics provider")?;
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -1443,7 +1445,14 @@ impl ActorInstance {
                 let target = invocation.target.clone();
                 let operation = invocation.operation.clone();
 
+                let start_at = Instant::now();
                 let res = self.handle_call(invocation).await;
+                let elapsed = u64::try_from(start_at.elapsed().as_nanos()).unwrap_or_default();
+
+                self.metrics
+                    .wasmcloud_host_handle_rpc_message_duration_ns
+                    .record(elapsed, &[KeyValue::new("operation", operation.clone())]);
```
connorsmith256, line 1454: Can we also tag with the actor ID? I imagine being able to see duration by actor is more important than an aggregate of all workloads on the host

crates/host/src/wasmbus/mod.rs
```diff
@@ -1903,6 +1913,10 @@ impl Host {
             ("hostcore.os".into(), OS.into()),
             ("hostcore.osfamily".into(), FAMILY.into()),
         ]);
+
+        let meter = global::meter("wasmcloud-host");
+        let metrics = HostMetrics::new(&meter);
```
connorsmith256, line 1918: Nit: it's odd to see these stuck in between the logic for labels. Why not move them down to where they'll be used?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1903,6 +1913,10 @@ impl Host {
             ("hostcore.os".into(), OS.into()),
             ("hostcore.osfamily".into(), FAMILY.into()),
         ]);
+
+        let meter = global::meter("wasmcloud-host");
```
connorsmith256, line 1917: Should we make this a const we also reference in main when we call `configure_tracing`?

crates/host/src/wasmbus/mod.rs
```diff
@@ -2462,6 +2477,7 @@ impl Host {
                 image_reference: actor_ref.to_string(),
                 actor_claims: Arc::clone(&self.actor_claims),
                 provider_claims: Arc::clone(&self.provider_claims),
+                metrics: self.metrics.clone(),
```
connorsmith256, line 2480: should we wrap this in an Arc?

crates/host/src/wasmbus/mod.rs
```diff
@@ -3183,6 +3199,7 @@ impl Host {
             );
             let otel_config = OtelConfig {
                 traces_exporter: self.host_config.otel_config.traces_exporter.clone(),
+                metrics_exporter: self.host_config.otel_config.metrics_exporter.clone(),
```
connorsmith256, line 3202: Note that providers don't know how to handle this (yet?) so this is the same as `metrics_exporter: None`

src/main.rs
```diff
@@ -246,18 +250,25 @@ async fn main() -> anyhow::Result<()> {
...
-        "wasmcloud-host".to_string(),
+        service_name.clone(),
         &otel_config,
         args.enable_structured_logging,
         Some(&log_level),
     ) {
         eprintln!("Failed to configure tracing: {e}");
     };
 
+    if let Err(e) = configure_metrics(&service_name, &otel_config) {
```
connorsmith256, line 268: I think this should be called from the `configure_tracing` logic (which we can rename to `configure_observability`)

crates/host/Cargo.toml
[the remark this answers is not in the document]
joonas: Done, I've re-exported the types used in the `host` create under `wasmcloud_tracing`.

crates/tracing/src/metrics.rs
joonas: Agreed, there is plenty of defaulting to lean on from the library, using that is definitely the right move.

src/main.rs
joonas: So thinking about this some more, I wonder if a shared flag would actually provide any value here, or is it more just for the sake of documentation?

The reason I say that is because outside of OTLP, there currently is no other shared open source format option across all three signals which we would reasonably want to support, so the flag feels a little bit unnecessary.

joonas: To recap for posterity's sake, we decided not to add this new flag, and we will revisit how these flags are more generally defined in a future PR.

crates/tracing/src/metrics.rs
joonas: Good callout, thank you 👍

crates/host/src/wasmbus/mod.rs
joonas: Yep, I think that's probably the right move here. I've made it into one.

crates/host/src/wasmbus/mod.rs
joonas: I've nixed this.

crates/host/src/wasmbus/mod.rs
joonas: Yes, absolutely. Not entirely sure why I stuck them here to begin with, so I've gone ahead and moved them down next to the where the host instance gets initialized.

crates/host/src/wasmbus/mod.rs
joonas: We chatted offline, instead of logging the `actor_id`, we decided to log the `actor.ref` instead since the (at least, current) actor id is going away in 1.0.

We can revisit to add more things here before 1.0 ships if it makes sense, but for now we at least have a way to identify things by actor.ref for now.

crates/host/src/wasmbus/mod.rs
joonas: I've made this use `CARGO_PKG_NAME` inside of the host crate, we can decide to revisit this and turn it into a constant or leave as is if it doesn't feel that important.

crates/tracing/src/lib.rs
joonas: talked offline, opting to leave this for the future PR that reworks these under a more cohesive observability umbrella.

src/main.rs
joonas: talked offline, we'll leave this change for the future unifying observability PR.

examples/docker/docker-compose-full.yml
```diff
@@ -53,7 +79,7 @@ services:
       WASMCLOUD_ALLOW_FILE_LOAD: "true"
       WASMCLOUD_OCI_ALLOWED_INSECURE: registry:5000
       OTEL_TRACES_EXPORTER: otlp
-      OTEL_EXPORTER_OTLP_ENDPOINT: http://tempo:4318/v1/traces
+      OTEL_EXPORTER_OTLP_ENDPOINT: http://localhost:4318
```
connorsmith256, line 82: This is the default now, right? Do we need this anymore?

examples/docker/docker-compose-full.yml
```diff
@@ -29,21 +29,47 @@ services:
       - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
       - GF_AUTH_DISABLE_LOGIN_FORM=true
     depends_on:
+      - prometheus
       - tempo
+
+  otelcol:
```
connorsmith256, line 35: Should we add these to the auxiliary file too?

connorsmith256 (approved):
Just a couple non blocking questions about the docker example. We'll hold on merging this until after we release v0.82

examples/docker/docker-compose-full.yml
joonas: Probably yes, I'll add it there as well 👍

examples/docker/docker-compose-full.yml
joonas: Yep, I kept it there for no good reasons, I'll remove 👍

brooksmtownsend (approved):
Metrics tracking worked well, I could discover the actor NS measurements through the Grafana metrics dashboard (fast invocations ftw!)

As a future followup, I'd be in favor of a more detailed `OTEL.md` or similar in this folder and a `docker-compose-otel.yaml` specifically only launching OTEL services, providing an application manifest to use and walking through different steps, etc. The README steps worked great for me though so this looks good to me 😄

joonas:
> As a future followup, I'd be in favor of a more detailed `OTEL.md` or similar in this folder and a `docker-compose-otel.yaml` specifically only launching OTEL services, providing an application manifest to use and walking through different steps, etc. The README steps worked great for me though so this looks good to me 😄

That sounds like a great idea, I'd love to also include in that iteration the ability to view the logs, which we could do by including [an exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/lokiexporter) for [Loki](https://grafana.com/oss/loki/) for example.

PR #1439, opened 2024-02-04
fix(provider-sdk): use send_request to handle request timeout 

Iceber:
## Feature or Problem

`rpc_client.request` will always use the default timeout, and even if the user sets a larger timeout, we will still be limited by it.
https://github.com/wasmCloud/wasmCloud/blob/6219096dba216c4795c4c85f91a1e01012833065/crates/provider-sdk/src/rpc_client.rs#L214-L218
https://github.com/wasmCloud/wasmCloud/blob/6219096dba216c4795c4c85f91a1e01012833065/crates/provider-sdk/src/rpc_client.rs#L251-L255

use `send_request` to send the request

### DIFF
--- a/crates/provider-sdk/src/rpc_client.rs
+++ b/crates/provider-sdk/src/rpc_client.rs
@@ -6,7 +6,7 @@ use crate::{
 use std::{fmt, sync::Arc, time::Duration};
 
 use async_nats::{Client, Subject};
-use futures::{Future, TryFutureExt};
+use futures::TryFutureExt;
 use sha2::Digest;
 use tracing::{
     debug, error,
@@ -51,8 +51,8 @@ impl fmt::Debug for RpcClient {
 
 impl RpcClient {
     /// Constructs a new RpcClient with an async nats connection.
-    /// parameters: async nats client, rpc timeout
-    /// secret key for signing messages, host_id, and optional timeout.
+    /// parameters: async nats client, host_id, optional timeout,
+    /// secret key for signing messages,, and lattice id.
     pub fn new(
         nats: Client,
         host_id: String,
@@ -209,22 +209,13 @@ impl RpcClient {
             timeout
         };
 
-        // let this = self.clone();
-        // let topic_ = topic.clone();
-        let payload = if let Some(timeout) = timeout {
-            // match tokio::time::timeout(timeout, this.request(topic, nats_body)).await {
-            match tokio::time::timeout(timeout, self.request(topic, nats_body)).await {
-                Err(_) => Err(InvocationError::Timeout),
-                Ok(Ok(data)) => Ok(data),
-                Ok(Err(err)) => Err(err),
-            }
-        } else {
-            self.request(topic, nats_body).await
-        }
-        .map_err(|err| {
-            error!(%err, "sending request");
-            err
-        })?;
+        let payload = self
+            .request_timeout(topic, nats_body, timeout)
+            .await
+            .map_err(|err| {
+                error!(%err, "sending request");
+                err
+            })?;
 
         let mut inv_response = crate::deserialize::<InvocationResponse>(&payload)?;
         if inv_response.error.is_none() {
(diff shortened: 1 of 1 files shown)


Iceber:
It looks like `wasmbus-rpc` needs to be fixed as well, I'll submit a new pr over there after this one is approved.

connorsmith256:
Hey @Iceber, `request` is meant to be a low-level function used within the RPC SDK. Using a consistent timeout value is by design: providers receive this timeout value from the host as `default_rpc_timeout_ms` in the `HostData`. (This field's type is `Option<u64>`, but that is for historical reasons. The host always provides a value.)

`request` is not meant to be directly called by providers. I think it's a mistake that it is marked as public and the doc comment says it can be used for general NATS messages, because the client used by this function is configured to use an RPC-specific connection + credentials + timeout. CC @thomastaylor312). Providers should call `send` to send RPC messages.

For providers that need to send messages with a custom timeout, we have `send_timeout`, which takes a custom `timeout` argument. This should be used rarely. The only use case I know of is for the HTTP server, which allows setting a [custom timeout for handling requests](https://github.com/wasmCloud/wasmCloud/blob/main/crates/providers/http-server/src/lib.rs#L355).

I think you did find a bug, however: if a provider calls `send_timeout` with a timeout of 10s, but the RPC client's inner timeout was set to 2s, the inner future returned by `maybe_timeout` will resolve after 2s, not waiting the full timeout.

So I think the fix should be to remove the `tokio::time::timeout` wrappers from the provider SDK. The underlying NATS client is already sending requests with a timeout, so there's no way to _extend_ this timeout with tokio, which would be a more common use case than shortening the timeout.

To extend the timeout, we will need to refactor the call to `self.client.request` (note that the `maybe_timeout` around `self.client.publish` does nothing, since publishing should never take more than a few milliseconds). Instead of calling `async_nats::Client::request`, we will need to use `async_nats::Client::send_request` and pass a request with a timeout. Here is [an example](https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L384) where we did this in the host.

Iceber:
> request is not meant to be directly called by providers. I think it's a mistake that it is marked as public and the doc comment says it can be used for general NATS messages, because the client used by this function is configured to use an RPC-specific connection + credentials + timeout. CC @thomastaylor312). Providers should call send to send RPC messages.

Provide a `request_timeout` function wrapper in order to ensure that `request` calls and `request_timeout` calls have consistent call logs, I will remove the **pub**.

> So I think the fix should be to remove the tokio::time::timeout wrappers from the provider SDK. The underlying NATS client is already sending requests with a timeout, so there's no way to extend this timeout with tokio, which would be a more common use case than shortening the timeout.
> To extend the timeout, we will need to refactor the call to self.client.request (note that the maybe_timeout around self.client.publish does nothing, since publishing should never take more than a few milliseconds). Instead of calling async_nats::Client::request, we will need to use async_nats::Client::send_request and pass a request with a timeout. Here is [an example](https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L384) where we did this in the host.

Yes, it would be more appropriate to use send_request, and I'll update this pr

crates/provider-sdk/src/rpc_client.rs
```diff
@@ -249,17 +240,31 @@ impl RpcClient {
...
+    ) -> InvocationResult<Vec<u8>> {
+        let timeout = if timeout.is_none() {
+            self.timeout
+        } else {
+            timeout
+        };
+
+        let req = async_nats::Request::new()
+            .payload(payload.into())
+            .timeout(timeout);
```
Iceber, line 260: notes: `request` always sets the timeout, which will make the request unaffected by the timeout in the `nats client`

connorsmith256 (approved):
LGTM! Thanks for fixing this @Iceber. I'll wait to see if @thomastaylor312 has any other feedback before merging

EDIT: we're blocked on merging currently, as we debug an issue in CI. Stand by 🙏

connorsmith256:
@Iceber would you mind rebasing off main? I think we've resolved the issue in CI, but it's still happening on your branch

thomastaylor312 (approved):
This looks all good. Merge away @connorsmith256! Thanks for the fix @Iceber

PR #1440, opened 2024-02-04
fix(provider-sdk): set the default timeout of nats to None

Iceber:
## Feature or Problem

The request timeout in the configuration returned by `ConnectOptions::default()` is set to 10s. Even if the default_rpc_timeout is set to greater than 10s, the true timeout is still 10s.

### DIFF
--- a/crates/provider-sdk/src/provider_main.rs
+++ b/crates/provider-sdk/src/provider_main.rs
@@ -99,6 +99,7 @@ where
             }
         },
     )
+    .request_timeout(None)
     .connect(nats_server)
     .await?;
 


crates/provider-sdk/src/provider_main.rs
```diff
@@ -99,6 +99,7 @@ where
             }
         },
     )
+    .request_timeout(None)
```
connorsmith256, line 102: Shouldn't this be `Some(host_data.default_rpc_timeout_ms)`?

connorsmith256 (changes requested):

crates/provider-sdk/src/provider_main.rs
Iceber: Setting it to None is related to https://github.com/wasmCloud/wasmCloud/pull/1439. If we're using send_request, then it's more appropriate to set it to default_rpc_timeout_ms, and I'll change it there.

PR #1445, opened 2024-02-05
feat(example): added minimal docker compose example

brooksmtownsend:
## Feature or Problem
This PR just adds the minimal docker compose example to give onlookers an idea of the required set of infrastructure they need to run. This notably excludes Wadm as there is a bit of confusion as to whether it's _required_ to run wasmCloud, so I want to be explicit on the minimal requirements.

### Manual Verification
I did make sure this compose worked.

### DIFF
--- a/examples/docker/docker-compose-minimal.yml
+++ b/examples/docker/docker-compose-minimal.yml
@@ -1,0 +1,24 @@
+# This docker-compose file starts the required services for wasmCloud, which is just:
+#   a NATS server
+#   a wasmCloud host
+
+version: "3"
+services:
+  nats:
+    image: nats:2.10.7-alpine
+    ports:
+      - "4222:4222"
+      - "6222:6222"
+      - "8222:8222"
+    command: ["-js"]
+  wasmcloud:
+    depends_on:
+      - "nats"
+    image: wasmcloud/wasmcloud:latest
+    environment:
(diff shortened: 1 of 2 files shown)


examples/docker/docker-compose-minimal.yml
```diff
@@ -0,0 +1,24 @@
+# This docker-compose file starts the required services for wasmCloud, which is just:
+#   a NATS server
+#   a wasmCloud host
+
+version: "3"
+services:
+  nats:
+    image: nats:2.10.7-alpine
```
joonas, line 8: So that we don't need to keep up with all of the new releases of NATS, maybe we pin to the 2.10.x branch instead:
```suggestion
    image: nats:2.10-alpine
```

joonas: I know this is different than what we do in the other files, I think we should go back and update the other files to use the same version.

connorsmith256 (approved):

examples/docker/docker-compose-minimal.yml
connorsmith256: I'm okay with this change, but think it can be done in a sweeping chore. In fact, happy to do that now

connorsmith256: https://github.com/wasmCloud/wasmCloud/pull/1446

Issue #1389, opened 2024-01-26
[RFC] Componentizing Links with wRPC


brooksmtownsend:
### RFC Title

RFC: Componentizing Links with wRPC

### Introduction

*wRPC stands for wRPC Remote Procedure Calls*

Goal: Provide distributed component support via wasmCloud

Since the [unveiling of wasmCloud](https://kevinhoffman.medium.com/introducing-waxosuit-6ad754b48ed9) our RPC framework enabled server-side WebAssembly applications. The core of this framework is the idea that we send invocations over a distributed network, NATS, based on capability contracts that form the building blocks of applications. These capability contracts included things like HTTP servers in `wasmcloud:httpserver`, key-value stores in `wasmcloud:keyvalue`, pubsub messaging in `wasmcloud:messaging`, logging in `wasmcloud:builtin:logging`, etc. The business logic components that wish to use the abstract capability and the actual implementation are completely separate entities at compile time, and can communicate using link definitions that connect them at runtime.

With the WebAssembly component model, defining our interfaces in WIT, and the stabilization of WASI 0.2.0, everything changes with the way we handle RPC.

### Motivation

[wasmCloud aims to support as many Wasm standards as possible](https://wasmcloud.com/docs/roadmap/q3q4#goals) to enable the goals of the project. We currently support the Wasm component model by using the concept of polyfills, essentially translating the wasmCloud contracts into standard WASI interfaces and vice versa. For example, when you load a component that exports the `wasi:http/incoming-handler` function, this is what happens:

![Untitled](https://github.com/wasmCloud/wasmCloud/assets/12040685/426cc784-a595-4cb2-8c19-f41e41e5dd76)

We don’t want to keep up this mapping forever. Polyfilling wasmCloud contracts that were built in a time where standard interfaces did not exist doesn’t make sense to do when there are standard interfaces that exist. In this case, the ideal mapping would be something like:
![Untitled2](https://github.com/wasmCloud/wasmCloud/assets/12040685/e9a54be2-7f76-4417-b887-7066dacd62a6)

In this case, the implementation of the WASI interface (which could be a capability provider, a simple HTTP listener, or a test module) can directly call that exported function on the actor component. Note above that we are still going through wasmCloud as a sort of middleman, and that’s because we still want the host to be responsible for dealing with the complexity of a distributed use case. In their design, Wasm components are explicitly designed to be composed together where one Wasm component’s imports can satisfy the others exports and vice versa.

[...]

PR #1469, opened 2024-02-06
feat(provider-sdk): support pubsub on wRPC subjects

vados-cosmonic:
## Feature or Problem

Up until now, publishing and subscribing for RPC communcations on the NATS cluster happened on subjects that were related to the wasmbus protocol (i.e. 'wasmbus.rpc.*').

To support the WIT-native invocations,
i.e. wRPC (https://github.com/wasmCloud/wasmCloud/issues/1389), we must change the publication and subscription subjects to include also the subjects that are expected to be used by wprc.

This commit updates the provider-sdk to listen *additionally* to subjects that are required/used by wrpc, though we do not yet have an implementation for encode/deocde.


crates/provider-sdk/src/lib.rs
```diff
@@ -142,3 +150,107 @@ pub trait ProviderHandler: Sync {
...
+                    if let Some(iface) = resolve.interfaces.get(*iface_id) {
+                        let iface_name = iface.name.as_ref().ok_or_else(|| {
+                            ProviderInitError::Initialization(format!(
+                                "unexpectedly un-named iface ID [{iface_id:?}] in world [{world_name}]"
+                            ))
+                        })?;
+
+                        for (fn_name, _) in iface.functions.iter() {
+                            topics.push(format!(
+                                "{lattice_name}.wrpc.{wrpc_version}.{ns}:${pkg_name}/{iface_name}.{fn_name}"
```
thomastaylor312, line 201: Just a heads up that we are going to need to escape version identifiers with `!`. So `0.1.0` -> `0!1!0` since we use the `.` char for topic splitting and matching. Should have @rvolosatovs or @brooksmtownsend confirm this before changing though

crates/provider-sdk/src/lib.rs
```diff
@@ -142,3 +150,107 @@ pub trait ProviderHandler: Sync {
...
+#[async_trait]
+pub trait WitContract {
+    /// Retrieve all WIT related files from the provider
+    ///
+    /// This is a lower level API to enable processing of rebuilding of the
+    /// `wit_parser::Resolve` structure, when processing WIT-related information
+    ///
+    /// Normally, the provider only has to implement this method, returning contents
+    /// of files in the WIT folder.
+    async fn get_wit_files(&self) -> ProviderInitResult<Box<dyn FileLookup + Send>> {
```
thomastaylor312, line 242: I'd probably not provide a default implementation here since people really should be including the wit

crates/provider-sdk/src/provider.rs
```diff
@@ -196,109 +212,134 @@ impl ProviderConnection {
...
+
+        // For every subscription, spawn a task to listen to incoming RPCs received,
+        // keeping track of the handles generated
+        let mut handles = Vec::new();
+        for mut sub in subs {
+            let this = self.clone();
+            let mut quit = quit.resubscribe();
+            let lattice = lattice.clone();
+            let provider = provider.clone();
+            let handle = tokio::spawn(async move {
```
thomastaylor312, line 252: Nit: It might be high time to refactor this out into a function we can spawn instead

crates/provider-sdk/src/provider.rs
```diff
@@ -196,109 +212,134 @@ impl ProviderConnection {
...
+                        .queue_subscribe(topic, RPC_SUBSCRIPTION_QUEUE_GROUP.to_string())
+                        .await?,
+                )
+            }
+        }
+
+        // For every subscription, spawn a task to listen to incoming RPCs received,
+        // keeping track of the handles generated
+        let mut handles = Vec::new();
+        for mut sub in subs {
```
thomastaylor312, line 247: Super nit: Couldn't we just combine all the streams using `futures::stream::SelectAll` and then spawn a single task? This might not matter with wRPC since the logic there might be different (since we can get all the info about the operation name from the topic).

Definitely don't worry about this yet until you have a bit more info on the wRPC stuff, was just calling that out

crates/provider-sdk/src/rpc_client.rs
```diff
@@ -132,7 +139,12 @@ impl RpcClient {
         let subject = make_uuid();
         let issuer = self.key.public_key();
         let target_url = crate::url(&target, Some(&method));
-        let topic = rpc_topic(&target, &self.lattice);
+        let topic = rpc_topic(&target, &self.lattice, &self.wit_resolve);
+
+        // TODO: RPC topic is going to be different and method-dependent here
+        // QUESTION: what is the method here when we're dealing with WIT-native wRPC calls?
+        //   Some variation of the WIT method? `<package>:<namespace>/<interface>/<method>`?
```
thomastaylor312, line 146: Yep, you have it right here so far as I understand. So the method will be in the topic like this

crates/provider-sdk/src/lib.rs
vados-cosmonic: Yup, I think you're right on this -- it's [in here](https://github.com/wasmCloud/wasmCloud/issues/1389), so I'll change this

crates/provider-sdk/src/provider.rs
vados-cosmonic: Yeah let's do this in another ticket -- I'd like to focus on just the necessary changes without refactoring for now

crates/provider-sdk/src/provider.rs
vados-cosmonic: We definitely could -- depending on task placement we might not want this all concentrated in one thread though -- so this is probably an improvement to make in the future if needed IMO

crates/provider-sdk/src/rpc_client.rs
vados-cosmonic: Yeah I think the answer for this is actually in the code, and I haven't looked it up (since we already handle custom WIT contracts) - so I'll use what's there *for now*, until `wrpc` lib lands and things change possibly.

crates/provider-sdk/src/lib.rs
```diff
@@ -142,3 +150,107 @@ pub trait ProviderHandler: Sync {
...
+}
+
+/// Generate a list of incoming NATS RPC topics on which the provider should listen
+fn generate_incoming_rpcs_topics(
+    lattice_name: impl AsRef<str>,
+    wrpc_version: impl AsRef<str>,
+    resolve: &Resolve,
+) -> ProviderInitResult<Vec<String>> {
+    let lattice_name = lattice_name.as_ref();
+    let wrpc_version = wrpc_version.as_ref().replace(".", "!");
```
rvolosatovs, line 169: We're going to start with just using `.` everywhere, no need for this replacement

crates/provider-sdk/src/lib.rs
rvolosatovs: We're not doing that for 0.0.1

crates/provider-sdk/src/provider.rs
```diff
@@ -94,18 +103,23 @@ impl ProviderConnection {
...
         host_data: &HostData,
+        wit_resolve: Option<Resolve>,
     ) -> ProviderInitResult<ProviderConnection> {
         let key = Arc::new(
             KeyPair::from_seed(&host_data.invocation_seed)
                 .map_err(|e| ProviderInitError::Initialization(format!("key failure: {e}")))?,
         );
 
+        // Shared reference to the WIT that powers this provider
+        let wit_resolve = Arc::new(wit_resolve);
```
rvolosatovs, line 114: Just a note that this should not exist at runtime (in the future?). The macro will have to derive `wrpc::Type` values using resolve at compile-time, but only the function parameter and result tuple types should persist

crates/provider-sdk/src/lib.rs
vados-cosmonic: welp, [evidently not](https://github.com/wasmCloud/wasmCloud/pull/1469#discussion_r1481507649)

crates/provider-sdk/src/lib.rs
vados-cosmonic: Reverted!

crates/provider-sdk/src/provider.rs
vados-cosmonic: OK awesome, once `wrpc` exists, we can change this code here!

crates/provider-sdk/src/lib.rs
thomastaylor312: Oh ok that makes sense!

brooksmtownsend (commented):
Looks real solid as a start for pubsub 👍🏻 I think we've had a conversation offline about where the wRPC library will handle some parts of this though, so I expect this PR to change quite a bit. Will give another full review then 😄

PR #1475, opened 2024-02-07
fix(provider-sdk): use send_request to handle request timeout

connorsmith256:
This mirrors https://github.com/wasmCloud/wasmCloud/pull/1439. Just pushing to my fork to see if I get the same error in CI


brooksmtownsend:
After #1480 merges let's see if that helps here

brooksmtownsend (approved):

PR #1476, opened 2024-02-07
feat(ci): use larger runners

connorsmith256:
We've been having some unknown issues with running out of disk space during builds. This swaps to a runner with a bigger disk, and as a bonus, more cores


thomastaylor312 (approved):

brooksmtownsend (approved):
image

PR #1478, opened 2024-02-07
Feat/wasmtime 17

connorsmith256:
This updates the wasmCloud ecosystem to wasmtime 17. Leaving in draft since I left 37 `todo!()` because I didn't know what to do 🙃


brooksmtownsend:
@rvolosatovs do you know what to do for the `todo!()` blocks @connorsmith256 left?

ricochet:
We need to begin working with 0.2.0 components now while underlying pieces change. We have users that are already beginning to build with 0.2.0 and experiencing significant friction. Additionally it's critical that we begin parallelizing work which includes componentizing examples, sdk's, and other external to the runtime pieces of the ecosystem.

18 won't be released for at least two more weeks. That is too long to wait.

keyvalue and blobstore were updated to support wasi-io at 0.2.0. All interface level changes are with the goal of supporting 0.2.0 and were incompatible with the previous release candidate of wasi-io that they were pinned on. I don't see a way to avoid moving everything (today) to 0.2.0 of interfaces like wasi-io.

@rvolosatovs will you pick this up and help get us on 17? I'm sorry this came as a surprise and I know it's going to cause some rework once other aspects of 18 that you've added to wasmtime become available. But I believe this is the best path forward to make progress in multiple areas.

PR #1480, opened 2024-02-08
fix(tests): Adjusts vault startup timeout to a better value

thomastaylor312:


connorsmith256 (approved):

brooksmtownsend:
Added another commit here after doing some testing, for some ungodly reason the HTTP server starting _before_ the vault provider was failing the tests, but not if you reverse them. I'm not sure why, I just know that it's okay now

ricochet (approved):

brooksmtownsend (approved):
First commit is alright, but that second commit 😍

PR #1486, opened 2024-02-09
fix(wash): windows path to target

### LINKED ISSUE #1481: [BUG] wash build on windows appears to be broken
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

wash build failing on windows

### Steps to reproduce

Following the quickstart on windows:
```bash
cargo install wash-cli
wash new actor --git wasmcloud/wasmcloud --subfolder examples/rust/actors/http-hello-world hello
rustup target add wasm32-wasi
cd hello
wash build
```

`wash build` fails with:
`Could not find compiled wasm file, please ensure \\?\C:\Users\marku\wcloudtraining\hello\target/wasm32-wasi/release/http_hello_world.wasm exists`

### Expected behavior

No errors and a signed component is built.

### Environment

- OS: [Windows]
- Shell [powershell]
- wasmcloud Version [0.81.0]

Potentially related https://github.com/wasmCloud/wash/issues/341

ricochet:
Verbatim paths on Windows are not well supported,
e.g. "\\\\?\\C:\\Users..." while technically valid, causes some fs api's like `exists` to fail errantly.

The fix is to use a third party lib normpath to normalize the path to the wasm
binary.

Related issue: https://github.com/rust-lang/cargo/issues/9770

Fixes #1481

Signed-off-by: Bailey Hayes <behayes2@gmail.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -118,6 +118,7 @@ names = { version = "0.14", default-features = false }
 nix = { version = "0.27", default-features = false }
 nkeys = { version = "0.3", default-features = false }
 notify = { version = "6", default-features = false }
+normpath = { version = "1.1" , default-features = false }
 nuid = { version = "0.4", default-features = false }
 oci-distribution = { version = "0.9", default-features = false }
 once_cell = { version = "1", default-features = false }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1682,9 +1682,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.3.22"
+version = "0.3.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d6250322ef6e60f93f9a2162799302cd6f68f79f6e5d85c8c16f14d1d958178"
+checksum = "bb2c4422095b67ee78da96fbb51a4cc413b3b25883c7717ff7ca1ab31022c9c9"
 dependencies = [
  "bytes",
  "fnv",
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -118,6 +118,7 @@ names = { version = "0.14", default-features = false }
 nix = { version = "0.27", default-features = false }
 nkeys = { version = "0.3", default-features = false }
 notify = { version = "6", default-features = false }
+normpath = { version = "1.1.1" , default-features = false }
```
t3hmrman, line 121:
```suggestion
normpath = { version = "1.1" , default-features = false }
```

t3hmrman (approved):
LGTM :rocket: 

I'd love a test for this (especially if we could constraint it to only run on windows), but not going to hold up the fix getting out for that (could be a follow up ticket for later)

Cargo.toml
vados-cosmonic, line 121:
```suggestion
normpath = { version = "1.1" , default-features = false }
```

vados-cosmonic (approved):
LGTM :rocket: 

Would love a test for this so we can catch it in the future when we change/to make sure it keeps not being broken, but that can be a follow up thing!

brooksmtownsend (approved):
Just a note we'll need to make sure this makes it into the upcoming wash-lib release so we can release the fix in wash-cli

connorsmith256:
Hmm, it looks like a lot of the `wash build` integration tests are failing

PR #1492, opened 2024-02-09
feat(examples): Add standalone docker compose setup for OTEL

joonas:
## Feature or Problem

In #1431, @brooksmtownsend mentioned that it would be nice to have a [standalone observability setup for docker compose](https://github.com/wasmCloud/wasmCloud/pull/1431#pullrequestreview-1871005201), which I thought was a great idea, so this implements that.

This does duplicate some of the changes made in #1431, and while working through this I started wondering if we should consider using the [Docker Compose `include`](https://docs.docker.com/compose/multiple-compose-files/include/) directive to de-duplicate some of this configuration from the other compose files where we'd like to have a full observability setup.

The downside of doing so would be that we'd essentially pin the required Docker for Desktop version to [4.22.0 and newer, which was released August 2023](https://docs.docker.com/desktop/release-notes/#4220) (the required [Docker Compose version is from July 2023](https://github.com/docker/compose/releases/tag/v2.20.0)).

That does seem like a reasonable trade-off in my book, and I'm happy to make the change here or separately if we think it's worth doing, otherwise this should be fine to go in as-is standalone.

### DIFF
--- a/examples/docker/config/grafana-datasources.yaml
+++ b/examples/docker/config/grafana-datasources.yaml
@@ -1,0 +1,23 @@
+apiVersion: 1
+
+datasources:
+  - name: Loki
+    type: loki
+    access: proxy
+    orgId: 1
+    url: http://loki:3100
+    basicAuth: false
+    version: 1
+    editable: false
+  - name: Prometheus
+    type: prometheus
+    url: http://prometheus:9090
+    access: proxy
+    editable: true
+  - name: Tempo
+    type: tempo
+    access: proxy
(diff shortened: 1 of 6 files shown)


thomastaylor312 (approved):
I like this, and I think the include stuff would be nice to describe as a separate PR. Would like one more opinion here before I merge

brooksmtownsend (approved):
This is good to go in as is, and I'm totally fine with adding the include directive. Thanks @joonas !

Issue #1444, opened 2024-02-05
[FEATURE] `wash app undeploy` should accept files as the argument


wamli:
Documentation says: ["You can think of a wadm application like a Kubernetes Deployment." ](https://wasmcloud.com/docs/ecosystem/wadm/). Since people are familiar to Kubernetes, it would be advantageous if `wadm` behaved analogous to Kubernetes. At least, in terms of the **undeployment** of an app, this is not the case:

```
# deploy with kubernetes
kubectl apply -f deployment.yaml

# undeploy with kubernetes
kubectl delete -f deployment.yaml

with Wash/Wadm/Wasmcloud, I faced:
$ wash app deploy wadm.yaml
Successfully deployed model rust-http-hello-world v0.0.3

$ wash app undeploy wadm.yaml
Invalid subject: Found extra components of subject, ensure your manifest name consists of only alphanumeric characters, dashes, and underscores.

$ wash app list
                                                                                                                                                                                           
  Name                    Latest Version   Deployed Version   Deploy Status   Description                                                                                                  
  rust-http-hello-world   v0.0.3           v0.0.3              Compensating   HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)  
                                                                                                                                                                                           
$ wash app undeploy rust-http-hello-world
Successfully undeployed model rust-http-hello-world
```

**Therefore, the following feature request:** `wash app undeploy wadm.yaml` should undeploy the app which it previously deployed via `wash app deploy wadm.yaml`.

MarkusEicher:
Hello folks. After speaking with Brooks Townsend on the Slack channel, I would like to get this issue assigned. Thank you.

MarkusEicher:
Thank you, Brooks. I am now working on it and will open a WIP: PR asap.

MarkusEicher:
#1495

vados-cosmonic:
Hey @thomastaylor312 note that we have this supported in `wash delete` -- I just made a quick PR for this as well

PR #1495, opened 2024-02-11
WIP: [FEATURE] wash app undeploy should accept files as the argument - Closes Issue #1444

MarkusEicher:
## Feature or Problem

In Issue #1444 the author wants to change the behavior of `wash app undeploy` to accept files (in this case `wadm.yaml` to undeploy an app previously deployed with `wash app deploy wadm.yaml`.

## Related Issues
This PR is related to Issue #1444


brooksmtownsend:
No problem @MarkusEicher, I'm going to mark this as a draft then since it's still a WIP 🙂 Feel free to tag myself or others here when you're looking for a full review

MarkusEicher:
Just a short ping. I am working on this, but I didn't find more time this week. Just to keep record that I am still on it. It takes me more time than I expected to get deeper into the codebase. If you see that I am on a totally wrong path with my commits, please let me know. Thank you.

rimbi:
Hi @MarkusEicher, I want to share my observations. Apparently, this request requires changes in the wadm interface because
- Accepting yaml file means the name of the model has to be extracted from the file
- as far as I can see there is no existing functionality to extract it on the client side and wadm does not provide an api to get it.
- wadm returns the model name in response to a `put` call but calling `put` before undeploy means double `put`ting which ultimately fails.
- we might need to add a new call to `wadm` which will, given the file content, return the model info.

MarkusEicher:
Hello @rimbi. Thank you for your input. I guess that you might be right here. I am still finding out how this codebase works. My first step here was to do the same that the code for the deploy part does. If I am understanding it correctly, then the deploy task does read the name of the last entry of the wadm.yaml and then deploy it. My idea was to do the same with the undeploy code. I agree that the current code here does not do the trick. I will have a look at the wadm code asap. I am always happy to get input from people, many thanks for your suggestion. I appreciate your support. @brooksmtownsend and maintainers - It would be totally ok for me, if you like to assign this task to @rimbi (if he wants to take it) to get it done quicker, because I'm still digging into this codebase and therefore am not the fastest. Please do check what's best for you and your project. I have no problems with this. Project before personal feelings.

brooksmtownsend:
Hey folks, just wanted to note that there's no need for this to be done very quickly so I'm happy to assign it to anyone 😄 

It does seem like we need more information from the Wadm manifest on the client side. I still like the interaction where we send the full model to wadm for parsing, validation, and deployment. On the wash side, I think we could use a function where we can locally parse the manifest and pull information out of it.

What we're really seeing the need here for is a more robust wadm API that exposes more strongly typed functions for parsing and sending models to wadm. For the scope of this issue I don't think we _need_ it, as we could do some of the YAML parsing ourselves (e.g. only pulling out the name and version from metadata, that's fairly easy parsing.) but it's worth considering that for the future.

Pretty sure @thomastaylor312 has wanted to make a wadm API for quite a while, maybe this will nerd snipe him 😉

rimbi:
Thanks @MarkusEicher and @thomastaylor312 . Having previously looked into the matter, I wanted to offer my insights in the hopes that they might be of assistance to you, @MarkusEicher.

MarkusEicher:
Thanks @brooksmtownsend and @rimbi for your feedback. I notice that I have to consult the WADM repo to get familiar with what it does. If you are ok with me taking some more time to work on this, I will be glad to do so. In any case, have a good weekend.

thomastaylor312:
> Pretty sure @thomastaylor312 has wanted to make a wadm API for quite a while, maybe this will nerd snipe him

I assume you mean Client API in rust? Or were you referring to adding a new endpoint to the API

PR #1498, opened 2024-02-12
Update to wasmtime 17 and Preview 2 interfaces

rvolosatovs:
## Feature or Problem

Update to Wasmtime 17 and Preview 2

Examples may be broken. Maintainers must maintain 🤷 

## Related Issues

Closes #1478 
Closes #1509 
Refs https://github.com/WebAssembly/wasi-keyvalue/pull/39


crates/runtime/src/actor/component/mod.rs
```diff
@@ -37,16 +37,16 @@ mod messaging;
...
+type TableResult<T> = Result<T, ResourceTableError>;
 
 mod guest_bindings {
     wasmtime::component::bindgen!({
         world: "guest",
         async: true,
         with: {
-           "wasi:io/error@0.2.0-rc-2023-11-10": wasmtime_wasi::preview2::bindings::io::error,
-           "wasi:io/poll@0.2.0-rc-2023-11-10": wasmtime_wasi::preview2::bindings::io::poll,
-           "wasi:io/streams@0.2.0-rc-2023-11-10": wasmtime_wasi::preview2::bindings::io::streams,
```
brooksmtownsend, line 49: I saw this change in a few places, do we not need to version anymore because the bindings are under `wasmtime_wasi::preview2`?

I want to understand when there is a bump, e.g. 0.2.1, will we need to version that specifically or will there be a different binding

brooksmtownsend (commented):
FWIW, when building the actor with `wash build`, this worked out-of-the-box with our HTTP example actor in Rust. Other languages need to be tested too, those largely depend on updating dependent toolchains and adapters though.

One thing we do need to do is
1. Release the new `wash` version with build support for adapting these modules
2. Update documentation as some of these wit worlds changed. I don't mind doing that it's really only a few updates.

Not sure what we want to do here around coordinating this. I think I'd suggest updating our documentation to point to a git tag for v0.81 that way merging this doesn't break examples

brooksmtownsend:
Updated docs to point at 0.81 tag to avoid breaking when this merges https://github.com/wasmCloud/wasmcloud.com/pull/293

brooksmtownsend:
@rvolosatovs looks like the build is failing, but other than the question above this looks good to me since we can get around breaking examples w/ https://github.com/wasmCloud/wasmcloud.com/pull/293

crates/runtime/src/actor/component/mod.rs
rvolosatovs: We simply don't need the version - the host only ever implement one version anyway. That's how this macro is used in Wasmtime, honestly I don't know about the specifics.
We will most probably not need to change anything for 0.2.1, but if we do, we will just do like is done in Wasmtime

brooksmtownsend: I'm thinking forward to WASI p3 where we have components that work with and are built for 0.2.0 but we still want to support them, I assume most APIs may bump versions but remain the same, so I just wanted to see how that will work at the bindings level. Going to resolve this conversation as it's not something that should block this PR

brooksmtownsend (approved):
I'm working on updating example code and READMEs and will be putting in a PR to follow this one

PR #1513, opened 2024-02-13
fix(wash-lib): re-lock wit-parser

brooksmtownsend:
## Feature or Problem
Fixes an issue where `wash-lib` published properly but failed to build with all features checked.


thomastaylor312 (approved):
Good to go pending passing tests

PR #1527, opened 2024-02-15
feat(wrpc): change set-target to set-link-name

vados-cosmonic:
## Feature or Problem

Up until the relatively low-level `wasmcloud:bus/lattice` WIT interface has used a function called `set-target` to aim invocations that occurred in compliant actors and providers.

Since wRPC (https://github.com/wasmCloud/wasmCloud/issues/1389) enabled  wasmCloud 1.0 is going to be WIT-first going forward, all WIT-driven function executions have access to the relevant interface (WIT interfaces, rather than Smithy-derived ones) that they call, at call time.

Given that actor & provider side function executions have access to their WIT interfaces (ex. `wasi:keyvalue/readwrite.get`), what we need to do is differentiate between the case where *multiple targets* might be responding to the same WIT interface-backed invocations.

Unlike before, `set-target` only needs to really differentiate between *link names*.

This commit updates `set-target` to perform differentiate between link names, building on the work already done to introduce more opaque targeting via Component IDs.


crates/actor/src/compat/wasmcloud/bus/host.rs
```diff
@@ -60,30 +60,9 @@ fn host_call(
...
-                namespace,
-                operation,
-                payload,
-            )
-        }
-        Some(TargetEntity::Actor(
-            ActorIdentifier::PublicKey(namespace) | ActorIdentifier::Alias(namespace),
-        )) => host_call("", namespace, operation, payload),
-    }
+    panic!("not supported, only wrpc")
```
brooksmtownsend, line 67: 😱 

Real talk though, I think a `bail` would be nice to have here just in case

crates/actor/src/compat/wasmcloud/bus/lattice.rs
```diff
@@ -1,95 +1,27 @@
...
-/// Target entity
-#[derive(Clone, Eq, PartialEq)]
-pub enum TargetEntity {
-    /// Link target paired with an optional name
-    Link(Option<String>),
-    /// Actor target
-    Actor(ActorIdentifier),
-}
+/// The current link name in use
+pub(crate) const LINK_NAME: OnceLock<RwLock<LinkName>> = OnceLock::new();
```
brooksmtownsend, line 11: I'm a little worried about this (granted, this is the beginning of my review so I might be missing something.) Is this a global link name? As in, two different actors will modify the same link name here in the OnceLock?

I think we definitely want an "in use" link name per component instance

crates/core/src/lib.rs
```diff
@@ -25,6 +25,16 @@ pub type ActorLinks = Vec<LinkDefinition>;
 pub type ClusterIssuerKey = String;
 pub type ClusterIssuers = Vec<ClusterIssuerKey>;
 
+/// Any string that represents a link between two components on the lattice,
+/// possibly a `ComponentId` or or a target group
+pub type LinkName = String;
```
brooksmtownsend, line 30: The link name serves the same purpose as it did before, simply the "name" on the InterfaceLinkDefinition. The "target" will be the part that's possibly a component ID or target group

crates/host/src/wasmbus/mod.rs
```diff
@@ -318,79 +336,39 @@ struct Handler {
     origin: WasmCloudEntity,
     // package -> target -> entity
     links: Arc<RwLock<HashMap<String, HashMap<String, WasmCloudEntity>>>>,
-    /// The current link name to use for interface targets, overridable in actor code via set_target()
+    /// The current link name to use for interface targets, overridable in actor code via set_link_name()
     interface_link_name: Arc<RwLock<String>>,
-    // link name -> package -> interface -> target
+    // link name -> package -> interface -> component ID
```
brooksmtownsend, line 341: I think this should still just be `target`, but I'm fine with brainstorming other ways to specify it. Reason being is that the target here is not always a component ID, it could be a routing group or an actor nkey etc etc

crates/host/src/wasmbus/mod.rs
```diff
@@ -318,79 +336,39 @@ struct Handler {
...
     async fn call_operation_with_payload(
         &self,
-        target: Option<TargetEntity>,
+        target: Option<WrpcTarget>,
         operation: impl Into<String>,
         request: Vec<u8>,
     ) -> anyhow::Result<Result<Vec<u8>, String>> {
-        let links = self.links.read().await;
-        let aliases = self.aliases.read().await;
+        // todo(vadossi-cosmonic): handle aliases...?
```
brooksmtownsend, line 357: Good news, we don't need em anymore! Call aliases were useful because we didn't have actor links, but now we do. Instead of looking up the call alias, we'll look up the target actor based on link name the same way we do providers.

crates/host/src/wasmbus/mod.rs
```diff
@@ -1426,23 +1355,11 @@ impl ActorInstance {
...
-                .cloned()
-                .context("failed to look up claims for target")?;
-            let mut target = PolicyRequestTarget::from(claims);
-            target.link_name = Some(invocation.target.link_name.clone());
-            target
+        let target = PolicyRequestTarget {
+            public_key: None, // todo(vadossi-cosmonic): get this from somewhere (pubkey of actor/provider)
+            issuer: None, // todo(vadossi-cosmonic): get this from somewhere (target claim issuer)
+            contract_id: None, // todo(vadossi-cosmonic): get this from somewhere (contract ID)
+            link_name: Some(invocation.target.link_name.clone()),
```
brooksmtownsend, line 1362: FWIW, we're gonna have to reshape the policy request format to deal with not having claims information everywhere necessarily

crates/runtime/src/actor/component/bus.rs
```diff
@@ -92,15 +97,23 @@ impl host::Host for Ctx {
...
+        target: Option<lattice::WrpcTarget>,
         operation: String,
         payload: Vec<u8>,
     ) -> anyhow::Result<Result<Vec<u8>, String>> {
-        let target = target
-            .map(TryInto::try_into)
-            .transpose()
-            .context("failed to parse target")?;
-        match self.handler.call_sync(target, operation, payload).await {
+        let interface = TargetInterface::from_operation(&operation)?;
```
brooksmtownsend, line 104: I'm really curious to see how the `wrpc` library is handling the target here, since both this PR and my PR are manually crafting the target interface but we should know exactly what we're calling based on the component import and we shouldn't have to do all the string parsing

crates/actor/src/compat/wasmcloud/bus/lattice.rs
vados-cosmonic: See earlier note about `compat` -- shouldn't it be gone for 1.0?

crates/core/src/lib.rs
vados-cosmonic: Ah OK, I'll change the comment here to be something like:

```
/// Name of an established link ([`InterfaceLinkDefinition`]) 
```

crates/host/src/wasmbus/mod.rs
vados-cosmonic: Great! Will remove this TODO

- [ ] Remove TODO around actor aliases

crates/host/src/wasmbus/mod.rs
vados-cosmonic: Ah, so can I remove those TODOs? Are we going to get none of that information from there?

crates/runtime/src/actor/component/bus.rs
vados-cosmonic: Yeah `operation` here being a `String` isn't great -- we should come back and fix this to make it a better type (it's technically very similar to `TargetInterface`)

crates/actor/src/compat/wasmcloud/bus/host.rs
vados-cosmonic: I went and removed `compat` (and the feature flags for wasm modules)...

crates/actor/src/compat/wasmcloud/bus/lattice.rs
vados-cosmonic: Removed `compat` -- probably going to increase the change set quite a bit here to remove the support for all non-components, but we'd need to do this anyway

crates/runtime/src/actor/component/bus.rs
vados-cosmonic: OK so this should be tracked in https://github.com/wasmCloud/wasmCloud/issues/1548 now, we should be able to get to and change this later

PR #1528, opened 2024-02-15
docs(adr): record ADR 0017

### LINKED ISSUE #336: [RFC] Define Interfaces Using WIT
## Summary
This request for comment aims to collect feedback and commentary on the idea to transition wasmCloud from using [Smithy](https://aws.amazon.com/blogs/developer/introducing-smithy-idl-2-0/) to define all interfaces to using [wit](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).

## Rationale
In the current wasmCloud ecosystem with the current state of tooling, creating actors and providers requires the use of **Smithy** to declare the interface used by wasm component actors and capability provider implementations. While Smithy may not be a closed system (it was created by Amazon), it is far from ubiquitous. Smithy doesn't enjoy the same kind of recognition and tooling ecosystem as [protocol buffers](https://github.com/protocolbuffers/protobuf) (aka **protobufs**), [Avro](https://avro.apache.org), or optimizations like [BSON](https://bsonspec.org) or [cap'n proto](https://capnproto.org). 

With the status quo, if I want to work with an existing interface like `wasmcloud:httpserver`, my actor code must use both guest and host generated code that comes from a Smithy definition. Our current code generation tools only support Rust and (for the most part) Go. It also requires that if we want to support a new target language for either side of the interface, we have to write our own code generation templates for that target language and wasmCloud developers must then use that generation. This kind of ecosystem management does not scale, nor is it remotely practical.

Another source of pain is that if I want to create a brand new interface, something like `custom:iot` for IoT devices, I have to write the interface in Smithy, and I am forced to implement my provider in one of the two languages supported by our Smithy code generation, and all actors using this interface must also be either Rust or Go. In every case, the code a developer creates must either rely directly on generated code during the build phase (e.g. `build.rs`), or it must rely on a library/crate that encapsulates the generated code (like [this](https://crates.io/crates/wasmcloud-interface-blobstore) or [this](https://crates.io/crates/wasmcloud-interface-httpserver)).

Last but not least, Smithy and its related tooling have no support at all for WebAssembly components, which locks the current IDL and tools out of this roadmap.

Summed up, this situation includes the following warning signs:
* Vendor lock-in (possible-but far from practical-to escape this lock-in)
* Somewhat niche community around Smithy
* Limited language choice and slow or non-existent addition of new language support
  * Requires tremendous effort from the maintainers to add language support
* No support for WebAssembly components

## Implementation Plan
As stated in the initial summary, the plan is to transition to **wit** for all interface definitions. The trick, as usual, is in how we implement this. It will require a progressive adoption that will follow the adoption and stability curves for wit and wasm components. This plan also has impact on the host, capability providers, and guests (actor components).

### Guest Wit with Wasifills
As mentioned at the beginning of this RFC, wasifills are the key piece of the implementation that makes all the puzzle pieces fit together. Remember that with WebAssembly components, the component can be written in any component-supporting language. This means that we can write Rust code that translates between discrete interface functions like `keyvalue`'s `get` and opaque binaries that the host never inspects. This translation happens in our (wasmCloud) code in a portable `.wasm` file, allowing developers to write their own code and have **_no tightly coupled dependencies on any wasmCloud libraries_**.

In both of the phases below, we should be able to make it so that `wash build` selects the appropriate wasifill `.wasm` file and links it, producing a module that "just works" with wasmCloud.

In an ideal world, since we can control the serialization of payloads, we should be able to use automatic code generation to produce wasifills for any wit contract on demand. 🤞🏻

#### Phase 1
The first phase will be to define the existing wasmCloud interfaces in terms of **wit**. This will be a breaking change in terms of the names and data types of the functions being used, but the API surface area should be about the same. Every function supported on today's Smithy interfaces will be available in the phase 1 interfaces. 

Using these new interfaces would involve the use of [wit-bindgen](https://github.com/bytecodealliance/wit-bindgen), and so guest language support would automatically be the set of all languages supported by **wit-bindgen**. The reason for this phase is that the specifications for the various wasi interfaces like `wasi-messaging` and `wasi-keyvalue` are still highly volatile.

```rust
wit_bindgen::generate!({
    world: "key-value",
    path: "../wit",
});

...

match keyvalue::get("something") {
 ...
}
```
This would reach out to the wasmCloud-specific **wit** and generate the appropriate code. There would no longer be any proprietary or vendor-locked code in any of the guest components. During this phase, the interfaces will appear similar but slightly different than the officially accepted wasi cloud interfaces. 

During this phase a wasifill would be linked to developer Wasm components to provide the right interface to wasmCloud hosts. Remember that the developer's code has no wasmCloud-specific code, the wasifill provides that translation layer.

#### Phase 2
During phase 2, we support the standard specifications for wasi cloud interfaces that are stable enough for use in real applications. Note that not all interfaces will be in the same phase at the same time: we can and will support both simultaneously. For example, during the `wash build` phase, it could select `keyvalue_wasifill_p1.wasm` or `keyvalue_wasifill_p2.wasm`. These wasifills could even be sitting in a registry or repository somewhere so that they could be pulled in by `wash` and not stored with the developer's project.

All it would take to change from phase 1 to phase 2 is changing the source of the wit world definition, and changing any code that needs to change as a result of the different interface. If we're lucky, most of the interfaces should migrate with minor impact. For example, the following phase 2 sample code might be almost indistinguishable from phase 1:

```rust
wit_bindgen::generate!({
    world: "wasi-keyvalue",
    path: "../wit",
});

...

match keyvalue::get("something") {
 ...
}
```

### Provider Implementations
Migrating capability providers to work with `wit` interfaces is actually a bit more difficult and requires more planning and forethought than with the guest modules. Today's capability providers use Smithy-generated serialization and de-serialization code (uses [messagepack](https://msgpack.org/index.html) right now). To create a provider for a given interface, the current process involves declaring a Rust dependency on the shared crate containing the various translation code.

Capability providers all handle a single wasmCloud function that accepts an invocation. Anything more strongly typed than that is generated on top of the lowest level primitive. To enable providers to continue to support bi-directional RPC with hosts, they will need to continue to support this mode of operation. The difference will be that instead of a provider relying on a shared library for its translation functionality, phase 1 and 2 providers will rely on wit-bindgen generated code and translation shims that extract payloads from invocations and invoke wit-style functions.

Instead of using a Smithy/weld shared crate, a wit-bindgen-based provider might look something like this:

```rust
wit_bindgen::generate!({
    path: "../wit",
    world: "pingpong",
});

struct Pingpong;
impl handler::Handler for Pingpong {
    fn on_receive(e: Event) -> Result<(), u32> {
     ...
    }
}
```
We should be able to generate the necessary "lifting" code to extract a payload from an invocation on the wire and invoke the appropriate strongly-typed function. It's unclear at this time, but there may be some things related to wasifill shims that we can do on the provider side, even though they aren't WASI-based (yet).

## Risks
This RFC makes some assumptions about the maturity and ease of use of certain tools. It assumes that we'll be able to make capability providers as loosely coupled from wasmCloud as guest components. In a worst case scenario, we might have to rely on shared code between the wasifill and the provider (which would still leave the guests totally uncoupled).

In other words, there's pretty low risk on the guest/actor side of the work in this RFC. There are some unknowns surrounding the provider side of this, but even that worst case scenario of manual translation layers isn't all that bad, and is still better than the status quo.

vados-cosmonic:
## Feature or Problem

This commit adds ADR 0017, which records the decision to define interfaces using WIT instead of Smithy going forward in the wasmCloud host, providers and actors.

Resolves #336

### DIFF
--- a/adr/0016-rename-lattice-prefix.md
+++ b/adr/0016-rename-lattice-prefix.md
@@ -1,8 +1,8 @@
 # Rename Lattice Prefix to Lattice
 
-| Status   | Deciders                                                                  | Date        |
-| -------- | ------------------------------------------------------------------------- | ----------- |
-| Accepted | Commenters on [#1221](https://github.com/wasmCloud/wasmCloud/issues/1221) | 18 Jan 2024 |
+| Status   | Deciders                                                                                        | Date        |
+|----------|-------------------------------------------------------------------------------------------------|-------------|
+| Accepted | Commenters & tagged contributors in [#1221](https://github.com/wasmCloud/wasmCloud/issues/1221) | 18 Jan 2024 |
 
 ## Context and Problem Statement
 
--- a/adr/0017-define-interfaces-using-WIT.md
+++ b/adr/0017-define-interfaces-using-WIT.md
@@ -1,0 +1,156 @@
+# Define Interfaces using WIT
+
+| Status   | Deciders                                                                | Date        |
+|----------|-------------------------------------------------------------------------|-------------|
+| Accepted | Commenters on [#336](https://github.com/wasmCloud/wasmCloud/issues/336) | 16 Feb 2024 |
+
+## Context and Problem Statement
+
+To enable distributed communication and RPC behavior on wasmCloud lattices, [AWS Smithy][smithy] was chosen as a [Interface Definition Language (IDL)][wiki-idl].
+
+As Smithy is protocol-agnostic and does not define *how* to serialize or deserialize the data, the combination of Smithy and [MessagePack][msgpack] is what makes up [`wasmbus`][wasmbus], the primary means of communication between hosts, providers, and actors on wasmCloud lattices.
+
+<details>
+<summary><h3><code>wasmcloud:keyvalue</code> Smithy contract excerpt</h3></summary>
+
+```smithy
+// key-value.smithy
+// Definition of a key-value store and the 'wasmcloud:keyvalue' capability contract
+//
+
+// Tell the code generator how to reference symbols defined in this namespace
+metadata package = [{
+    namespace: "org.wasmcloud.interface.keyvalue",
+    crate: "wasmcloud_interface_keyvalue",
+    py_module: "wasmcloud_interface_keyvalue",
+    doc: "Keyvalue: wasmcloud capability contract for key-value store",
+}]
+
+namespace org.wasmcloud.interface.keyvalue
+
+use org.wasmcloud.model#wasmbus
+use org.wasmcloud.model#rename
+use org.wasmcloud.model#n
+use org.wasmcloud.model#U32
+use org.wasmcloud.model#I32
+
+@wasmbus(
+    contractId: "wasmcloud:keyvalue",
+    providerReceive: true )
+service KeyValue {
+  version: "0.1.1",
+  operations: [
+    Increment, Contains, Del, Get,
+    ListAdd, ListClear, ListDel, ListRange,
+    Set, , SetAdd, SetDel, SetIntersection, SetQuery, SetUnion, SetClear,
+  ]
+}
+
+/// Gets a value for a specified key. If the key exists,
+/// the return structure contains exists: true and the value,
+/// otherwise the return structure contains exists == false.
+@readonly
+operation Get {
+  input: String,
+  output: GetResponse,
+}
+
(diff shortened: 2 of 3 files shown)


brooksmtownsend (approved):
🚀

PR #1531, opened 2024-02-16
docs(adr): record ADR 0018

### LINKED ISSUE #664: [RFC] Host metrics

vados-cosmonic:
## Feature or Problem

This commit adds ADR 0018, which records the decision to implement OTEL-compliant metrics for the wasmCloud host.

Resolves #664

### DIFF
--- a/adr/0018-otel-metrics.md
+++ b/adr/0018-otel-metrics.md
@@ -1,0 +1,59 @@
+# Add OTEL-compliant metrics to wasmCloud
+
+| Status   | Deciders                                                                | Date        |
+|----------|-------------------------------------------------------------------------|-------------|
+| Accepted | Commenters on [#664](https://github.com/wasmCloud/wasmCloud/issues/664) | 17 Feb 2024 |
+
+## Context
+
+After much development and refinement, [OpenTelemetry][otel] has become *the* standard for [Observability][o11y] in the Cloud Native ecosystem, and the three primary "pillars of observability" are key:
+
+- Logs
+- Tracing
+- Metrics
+
+While wasmCloud has supported logs and tracing for a long time with the integration of the [`tracing`][crates-tracing], metrics has not been suported natively by the wasmCloud host, and thus not supported by wasmCloud as a whole.
+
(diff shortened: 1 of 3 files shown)


brooksmtownsend (approved):
Looks wonderful! @joonas do you have any thoughts or changes you'd like to make here?

joonas (approved):
Looks great, thanks for writing this up @vados-cosmonic!

vados-cosmonic:
No problem, thanks for doing all the hard work, @joonas !

PR #1550, opened 2024-02-20
feat: add (most of) custom interface plumbing

rvolosatovs:
cc @brooksmtownsend 

Left the host part for you to take care of. 

This will let you call arbitrary functions on components using `wrpc::Value` vectors and get vectors of the same type back.
It also removes all the custom `call` functionality. We can support modules assuming simply a function of `(list<u8>) -> (list<u8>)` signature for now

I did not bother fixing tests


brooksmtownsend:
>I did not bother fixing tests

That's what the branch is for 😄 we'll get to tests a little down the road. Thanks @rvolosatovs

brooksmtownsend (approved):

PR #1552, opened 2024-02-20
ci(messaging-nats): create pipeline

brooksmtownsend:
## Feature or Problem
This PR is the first of many to move our capability provider release actions from their vendor folder to the main repository. We'll likely want to write this as a matrix of actions instead of one-per-provider, but the first step is ensuring they run well and then we can optimize.

## Release Information
messaging-nats v0.18.0

### Manual Verification
https://github.com/wasmCloud/wasmCloud/actions/runs/7979855296


brooksmtownsend:
we don't really need this now 😄

PR #1554, opened 2024-02-20
feat(ctl)!: Changes config topic to be for named config

thomastaylor312:
This is the first in a set of changes to move over to named config. It is not technically complete as you essentially have to name your config the same as the actor ID. I did this purposefully so as to not have a PR of doom with all the changes. The next PR will be adding named config to the scale command, then support for named config and providers in another PR after that

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -361,40 +361,31 @@ impl Client {
         }
     }
 
-    /// Puts a config item for the given entity (actor or provider) ID at the given key, replacing
-    /// any data that is already present. Data is stored as a Vec of bytes, so anything that can be
-    /// turned into a vec is accepted as a parameter for the value.
+    /// Puts a named config, replacing any data that is already present.
     ///
-    /// Key names must be valid NATS subject strings and not contain any `.` or `>` characters.
-    ///
-    /// NOTE: This operation is currently experimental and may change or be removed at any time
+    /// Config names must be valid NATS subject strings and not contain any `.` or `>` characters.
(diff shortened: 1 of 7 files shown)


crates/control-interface/src/lib.rs
```diff
@@ -361,40 +361,31 @@ impl Client {
...
 
-    /// Puts a config item for the given entity (actor or provider) ID at the given key, replacing
-    /// any data that is already present. Data is stored as a Vec of bytes, so anything that can be
-    /// turned into a vec is accepted as a parameter for the value.
+    /// Puts a named config, replacing any data that is already present.
     ///
-    /// Key names must be valid NATS subject strings and not contain any `.` or `>` characters.
-    ///
-    /// NOTE: This operation is currently experimental and may change or be removed at any time
+    /// Config names must be valid NATS subject strings and not contain any `.` or `>` characters.
```
brooksmtownsend, line 366: What could we catch if we validated that it was a valid subject string? I believe the error case is getting a malformed sunject error when we try to request, but I'm unsure exactly what the user would see

brooksmtownsend (approved):
Looks great 👍🏻 Note that enforcing the string -> string hashmap simplifies quite a few possible edge cases, especially around encoding assumptions. We can always extend that in the future if needed.

Only comment here is probably a future improvement + good first issue to make the control interface library better at catching when a subject is invalid 😄

crates/control-interface/src/lib.rs
thomastaylor312: Yeah, I think rust errors are bit better now, but definitely a future improvement we should do here

PR #1565, opened 2024-02-21
feat(provider-sdk): support wrpc pubsub reloaded

vados-cosmonic:
## Feature or Problem

Up until now, publishing and subscribing for RPC communcations on the NATS cluster happened on subjects that were related to the wasmbus protocol (i.e. 'wasmbus.rpc.*').

To support the WIT-native invocations, i.e. wRPC (#1389), we must change the publication and subscription subjects to include also the subjects that are expected to be used by wprc.

This commit updates the provider-sdk to listen *additionally* to subjects that are required/used by wrpc, though we do not yet have an implementation for encode/deocde.


crates/provider-sdk/src/provider.rs
```diff
@@ -196,133 +224,166 @@ impl ProviderConnection {
...
-            .map_err(InvocationError::from)?;
-        self.validate_provider_invocation(&inv, &claims)
-            .await
-            .map_err(InvocationError::from)?;
-        let span = tracing::debug_span!("dispatch", public_key = %inv.origin.public_key, method = %inv.operation);
+        let operation = operation.as_ref();
+        let source_id = source_id.as_ref();
+
+        // todo(vadossi-cosmonic): do we need to do the stuff that used to happen here? (see handle_rpc())
+        // - dechunking (we still need to do this if the wrpc_transport::Value is big enough, right??) NATS is still gonna take chunks
```
rvolosatovs, line 370: wrpc-transport handles chunking, if a message of size S is to be sent on topic A and it exceeds the byte size limit N, then S/N messages will be sent on topic A.
E.g. with a hypothetical limit of 1 byte and a payload of `abc`, exactly 3 messages will be sent - `a`, `b` and `c`

Clients are free to arbitrarily chunk data as they see fit

crates/provider-sdk/src/provider.rs
```diff
@@ -182,147 +224,173 @@ impl ProviderConnection {
...
-            loop {
+        let mut handles = Vec::new();
+        let provider_key = provider_key.as_ref();
+
+        // Build a wrpc client that we can use to listen for incoming invocations
+        let wrpc_client = WrpcNatsClient::new(
+            self.rpc_client.client(),
+            format!("{lattice}.{}", provider_key),
+        );
+        let link_name = self.host_data.link_name.clone();
```
brooksmtownsend, line 247: nit: We should be able to get rid of all mentions to link name, since we're always passing "default" now from the host

crates/provider-sdk/src/provider.rs
```diff
@@ -182,147 +224,173 @@ impl ProviderConnection {
...
     /// This method starts a separate async task and returns immediately.
     /// It will exit if the nats client disconnects, or if a signal is received on the quit channel.
     pub async fn subscribe_rpc<P>(
         &self,
         provider: P,
-        mut quit: QuitSignal,
+        quit: QuitSignal,
         lattice: String,
-    ) -> ProviderInitResult<JoinHandle<()>>
+        provider_key: impl AsRef<str>,
```
brooksmtownsend, line 234: nit: we should rename to "provider_id"

crates/wash-lib/src/cli/output.rs
```diff
@@ -39,7 +39,7 @@ pub struct StopCommandOutput {
 /// JSON output representation of the `wash link query` command
 #[derive(Debug, Deserialize)]
 pub struct LinkQueryCommandOutput {
-    pub links: Vec<HashMap<String, ActorLinks>>,
+    pub links: Vec<HashMap<LinkName, Vec<InterfaceLinkDefinition>>>,
```
brooksmtownsend, line 42: Did this work for you? I _think_ that now this should just be a Vec<InterfaceLinkDefinition>, but we can approach that when testing wash

crates/provider-wit-bindgen-macro/src/wrpc.rs
```diff
@@ -0,0 +1,87 @@
...
+pub(crate) fn generate_wrpc_nats_subject_to_fn_mapping(
+    resolve: &Resolve,
+) -> Result<Vec<WrpcExport>> {
+    let mut subjects = Vec::new();
+
+    for (_, world) in resolve.worlds.iter() {
+        // Generate a lookup of all dynamic functions for this world export
+        let mut wrpc_type_lookup = wrpc_types::function_exports(&resolve, world.exports.iter());
+
+        // We need to also go through the exports in order to generate the
```
brooksmtownsend, line 36: nit: generate the what 😄

crates/providers/blobstore-fs/src/lib.rs
```diff
@@ -241,7 +242,7 @@ impl FsProvider {
 
         let cr = receiver.receive_chunk(chunk).await
             .with_context(|| format!(
-                "sending chunk error: Container({container_id}) Object({object_id}) to Actor({actor_id})",
+                "sending chunk error: Container({container_id}) Object({object_id}) to Actor({source_id})",
```
brooksmtownsend, line 245: For the blobstore provider, where the provider is the think linked _to_ a component, I think we should be sending chunks to the `target` here on the linkdef.

Since the blobstore provider is one that links to components and has components link to it (with interface linkdefs) I think this is a good example for us to find out how to properly branch the code based on that. Could be as simple as matching the source_id to see if it's the provider ID or if it's something else.

brooksmtownsend (commented):
I think that this PR supports capability providers generating types from WIT and handlers that work with wRPC, with the caveat that linking is still in the middleground between "old" and "new" links. I left some nits around link names and a comment around the blobstore provider but I believe it applies to all of them

crates/provider-sdk/src/provider.rs
vados-cosmonic: ah but that's not always going to be the case though, right -- this code shouldn't break when we use actual link names, either.

crates/provider-sdk/src/provider.rs
vados-cosmonic: Done!

crates/provider-wit-bindgen-macro/src/wrpc.rs
vados-cosmonic: Updated this!

crates/providers/blobstore-fs/src/lib.rs
vados-cosmonic: I thought it would still be the source -- because this is used via the look up the link that *came in*, right?

The link that is saved is the one that is made when the actor first connects, and the source=actor, target=provider.

crates/provider-sdk/src/provider.rs
brooksmtownsend: Ah yeah I mean that we can remove link name from the host data, but you're right that we need to keep them around when we use interface link definitions with link names

crates/providers/blobstore-fs/src/lib.rs
brooksmtownsend: Well, it's tricky, but for this situation the actor will need to call the provider and vice versa, so there will need to be a link with the provider as the target and the provider as the source

brooksmtownsend: Oh wait I'm misunderstanding, you mean that in this case the actor calls the provider, and so the provider is just sending chunks back to the original caller. That makes sense!

It does bring up a question on if we want to validate that there is in fact a link from the provider to the actor when calling `wasmcloud:blobstore/chunk-receiver.receive-chunk`, in a pure world we would require this and so we could either
1. Enforce it at the host level (deny the invocation since the component didn't have a link to it)
2. Enforce it at the provider level (subject to implementation details

I think this is a little bit weird and I need to think about it a little bit more, but it doesn't affect this PR merging

brooksmtownsend (approved):

crates/providers/blobstore-fs/src/lib.rs
vados-cosmonic: > Oh wait I'm misunderstanding, you mean that in this case the actor calls the provider, and so the provider is just sending chunks back to the original caller. That makes sense!

Yup!

Yeah the links no longer being bi-directional actually does bring up a rather interesting issue. We basically need a `put_link` on Actors as well... Should hosts be intermediating links that are made to actors, or should actors do that themselves?

Feels like the distinction between actor and provider continues to dwindle!

PR #1568, opened 2024-02-21
wRPC component export support

brooksmtownsend:
## Features
This PR adds support for components serving dynamic functions based on their exports on wRPC topics. I've tested this using a small binary that invokes a component exporting a function `foo`, as shown below in the manual verification section. The vast majority of this PR is removing code that's no longer exercised for handling and sending invocations over wasmbus, instead replaced with serving dynamic functions on wRPC.

I also implemented importing functions and, using the polyfilled interfaces, invoking other components on the lattice via wRPC. This will work great for our custom interfaces, but HTTP, keyvalue, blobstore, and messaging all need to be reworked to send wRPC messages instead (see TODO under `call_operation`).

I've left a few `TODO` comments regarding policy checking and invocation signing/response signing so we can keep an eye on where those should be implemented.

### Manual Verification

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -48,7 +48,7 @@ use wasmcloud_control_interface::{
     HostLabel, InterfaceLinkDefinition, LinkDefinition, LinkDefinitionList, ProviderAuctionAck,
     ProviderAuctionRequest, ProviderDescription, RegistryCredential, RegistryCredentialMap,
     RemoveLinkDefinitionRequest, ScaleActorCommand, StartProviderCommand, StopHostCommand,
-    StopProviderCommand, UpdateActorCommand, WitInterface, WitNamespace, WitPackage,
+    StopProviderCommand, UpdateActorCommand, WitInterface,
 };
 use wasmcloud_core::{
     HealthCheckResponse, HostData, Invocation, InvocationResponse, LatticeTargetId, LinkName,
@@ -58,7 +58,7 @@ use wasmcloud_runtime::capability::logging::logging;
 use wasmcloud_runtime::capability::{
     blobstore, guest_config, messaging, ActorIdentifier, Blobstore, Bus, CallTargetInterface,
     KeyValueAtomic, KeyValueEventual, Logging, Messaging, OutgoingHttp, OutgoingHttpRequest,
-    TargetEntity,
+    TargetEntity, WrpcInterfaceTarget,
 };
 use wasmcloud_runtime::Runtime;
 use wasmcloud_tracing::context::TraceContextInjector;
@@ -282,12 +282,8 @@ struct Handler {
     /// - A routing group
     /// - Some other opaque string
     #[allow(clippy::type_complexity)]
-    interface_links: Arc<
-        RwLock<HashMap<LinkName, HashMap<WitNsAndPackage, HashMap<WitInterface, LatticeTargetId>>>>,
-    >,
-
-    // NOTE(brooksmtownsend): needs deduplication of responsibility with interface_links
-    wrpc_targets: Arc<RwLock<HashMap<TargetInterface, TargetEntity>>>,
+    interface_links:
+        Arc<RwLock<HashMap<LinkName, HashMap<String, HashMap<WitInterface, LatticeTargetId>>>>>,
     aliases: Arc<RwLock<HashMap<String, WasmCloudEntity>>>,
     // interface -> function -> result types
     polyfilled_imports: HashMap<String, HashMap<String, Arc<[wrpc_types::Type]>>>,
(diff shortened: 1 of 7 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -774,156 +802,28 @@ impl Bus for Handler {
...
-        Box<dyn AsyncWrite + Sync + Send + Unpin>,
-        Box<dyn AsyncRead + Sync + Send + Unpin>,
-    )> {
-        let (mut req_r, req_w) = socket_pair()?;
-        let (res_r, mut res_w) = socket_pair()?;
-
-        let links = Arc::clone(&self.links);
-        let aliases = Arc::clone(&self.aliases);
-        let nats = self.nats.clone();
-        let chunk_endpoint = self.chunk_endpoint.clone();
```
brooksmtownsend, line 793: @rvolosatovs assumption here that wRPC handles chunking or a similar functionality if the message payload is above the NATS limit, is that correct?

crates/host/src/wasmbus/mod.rs
```diff
@@ -774,156 +802,28 @@ impl Bus for Handler {
...
+        params: Vec<wrpc_transport::Value>,
+    ) -> anyhow::Result<Vec<wrpc_transport::Value>> {
+        if let Some(TargetEntity::Wrpc(interface_target)) = target {
+            let prefix = format!("{}.{}", self.lattice, interface_target.id);
+            let wrpc_client =
+                wrpc_transport_nats::Client::new(self.nats.clone(), prefix.to_string());
+            let (result, _tx) = wrpc_client
+                // TODO: get types
+                .invoke_dynamic(instance, name, params, &[])
+                .await?;
```
brooksmtownsend, line 820: Also for @rvolosatovs ... I've been trying to consider how we actually get the types here that we expect from the function, I'm not sure how we would do it unless we had access to the component imports. I'm starting to lose some brain power, can you just confirm that I'm not thinking of this function completely wrong? (e.g., this function is called when we need to invoke a target over wRPC, not when the component is invoked)

rvolosatovs: Yeah, we'll need the imports here, I'll plumb it through.

All of this is a side-effect of wRPC library not having "native" wasmtime support, which I decided to pause for now. Once that's done, we'll remove this functionality

rvolosatovs: https://github.com/wasmCloud/wasmCloud/pull/1570

crates/host/src/wasmbus/mod.rs
```diff
@@ -798,163 +842,35 @@ impl Bus for Handler {
...
+        params: Vec<wrpc_transport::Value>,
+    ) -> anyhow::Result<Vec<wrpc_transport::Value>> {
+        if let Some(TargetEntity::Wrpc(interface_target)) = target {
+            let result_types = match self
+                .polyfilled_imports
+                .get(instance)
+                .and_then(|functions| functions.get(name))
+            {
+                Some(results) => results.as_ref(),
+                None => bail!("polyfilled import not found, could not determine result types"),
```
vados-cosmonic, line 860: NIT: can we get more specific about *which* polyfilled imports were not found?

crates/host/src/wasmbus/mod.rs
```diff
@@ -798,163 +842,35 @@ impl Bus for Handler {
...
-        operation: String,
-        request: Vec<u8>,
-    ) -> anyhow::Result<Vec<u8>> {
-        self.call_operation_with_payload(target, operation, request)
-            .await
-            .context("failed to call linked provider")?
-            .map_err(|e| anyhow!(e).context("provider call failed"))
+            Ok(result)
+        } else {
+            error!("invalid target");
```
vados-cosmonic, line 871: NIT: better error message please!

crates/host/src/wasmbus/mod.rs
```diff
@@ -313,19 +337,17 @@ impl Handler {
         request: Vec<u8>,
     ) -> anyhow::Result<Result<Vec<u8>, String>> {
         let operation = operation.into();
+        let links = self.links.read().await;
+        let aliases = self.aliases.read().await;
```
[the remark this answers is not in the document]
brooksmtownsend: We are! I just put it back in to compile for this PR. My next PR will remove it properly :)

crates/host/src/wasmbus/mod.rs
```diff
@@ -196,26 +195,30 @@ struct ActorInstance {
...
 impl ActorInstance {
-    pub(crate) async fn instantiate(&self) -> anyhow::Result<wasmcloud_runtime::actor::Instance> {
-        self.actor
-            .instantiate()
-            .await
-            .context("failed to instantiate actor")
-    }
-
-    pub(crate) async fn add_handlers<'a>(
+    /// Handle an incoming wRpc request to invoke an export on this actor instance.
```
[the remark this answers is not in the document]
brooksmtownsend: I think wRPC is the right thing, I could've sworn clippy got annoyed with me putting that capitalization will check again

PR #1571, opened 2024-02-22
feat(wrpc): change set-target to set-link-name

vados-cosmonic:
## Feature or Problem

Up until the relatively low-level `wasmcloud:bus/lattice` WIT interface has used a function called `set-target` to aim invocations that occurred in compliant actors and providers.

Since wRPC (#1389)
enabled  wasmCloud 1.0 is going to be WIT-first going forward, all WIT-driven function executions have access to the relevant interface (WIT interfaces, rather than Smithy-derived ones) that they call, at call time.

Given that actor & provider side function executions have access to their WIT interfaces (ex. `wasi:keyvalue/readwrite.get`), what we need to do is differentiate between the case where *multiple targets* might be responding to the same WIT interface-backed invocations.

Unlike before, `set-target` only needs to really differentiate between *link names*.

This commit updates `set-target` to perform differentiate between link names, building on the work already done to introduce more opaque targeting via Component IDs.


crates/control-interface/src/types.rs
```diff
@@ -433,29 +433,53 @@ pub struct LinkDefinition {
     pub values: LinkSettings,
 }
 
+/// Identifier of one or more entities on the lattice used for addressing. May take many forms, such as:
+/// - actor public key
+/// - provider public key
+/// - opaque string
+pub type LatticeTargetId = String;
```
brooksmtownsend, line 440: I want to noodle on this name a bit, it's a little strange to provide a target ID to the `source_id` struct member. I've been calling these component IDs, but maybe that needs to be more clear.

We can address this w/ conversation and a follow up, since it's just String aliasing 😄

crates/host/src/wasmbus/mod.rs
```diff
@@ -240,11 +241,19 @@ struct Handler {
...
+
     /// The current link name to use for interface targets, overridable in actor code via set_target()
-    interface_link_name: Arc<RwLock<String>>,
-    // link name -> package -> interface -> target
+    interface_link_name: Arc<RwLock<LinkName>>,
+
+    /// Map of link names -> WIT ns & package -> WIT interface -> Target
+    ///
+    /// While a target may often be a component ID, it is not guaranteed to be one, and could be
+    /// some other identifier of where to send invocations, representing one or more lattice entities.
```
brooksmtownsend, line 251: This is a great callout to argue for some lattice target, as it may be a component ID, a routing group, or even just an opaque string 🤔

crates/provider-sdk/src/lib.rs
```diff
@@ -53,28 +53,19 @@ pub fn rpc_topic(entity: &WasmCloudEntity, lattice: &str) -> String {
 }
 
 /// Generates a fully qualified wasmbus URL for use in wascap claims. The optional method parameter is used for generating URLs for targets being invoked
+// todo(vados-cosmonic): we can remove this entire function once claim signing is removed
+// see: https://github.com/wasmCloud/wasmCloud/issues/1219
```
brooksmtownsend, line 57: 👍🏻 yep a good bit of this will get yanked out soon

crates/providers/http-server/wit/provider-httpserver.wit
```diff
@@ -1,5 +1,7 @@
 package wasmcloud:provider-http-server;
 
 world provider-http-server {
+    /// todo(vados-cosmonic): bus/host no longer exists, this provider must be rewritten
+    /// to only use wasi:http
```
brooksmtownsend, line 5: nitty nit nit: we did talk about it previously, due to the complexity of wasi:http this will be a _wasi:http-like_ contract

crates/runtime/src/actor/component/mod.rs
```diff
@@ -1,5 +1,7 @@
 use crate::actor::claims;
-use crate::capability::{builtin, Bus, Interfaces, TargetInterface};
+use crate::capability::{builtin, Bus, Interfaces};
+//use crate::capability::bus::lattice::CallTargetInterface;
```
brooksmtownsend, line 3: nit: we can just remove the comment

crates/runtime/src/capability/builtin.rs
```diff
@@ -188,94 +188,73 @@ impl Eq for ActorIdentifier {}
...
+/// can identify an interface being called and optionally a specific function on that interface.
+pub struct CallTargetInterface {
+    /// WIT namespace (ex. `wasi` in `wasi:keyvalue/readwrite.get`)
+    pub namespace: String,
+    /// WIT package name (ex. `keyvalue` in `wasi:keyvalue/readwrite.get`)
+    pub package: String,
+    /// WIT interface (ex. `readwrite` in `wasi:keyvalue/readwrite.get`)
+    pub interface: String,
+    /// WIT package name (ex. `readwrite` in `wasi:keyvalue/readwrite.get`)
+    pub function: Option<String>,
```
brooksmtownsend, line 233: nit: I think this should be `get` right?

We definitely need to rethink the target interface things. I might change some of this up in #1568 depending on what I can see is obsolete

brooksmtownsend (approved):
All nits I believe that are nonblocking, I'm definitely seeing the places where we have some vestigial code when it comes to TargetInterface so I'll try to address in #1568 where I can

crates/control-interface/src/types.rs
vados-cosmonic: Yeah I agree -- honestly we should have a proper `enum` here at some point since we can somewhat identify *some* of them -- but yeah maybe this should be like `LatticeRouteName`.  Probably `Name` and not `Id` because we might end up giving every route an individual ID (like ULID/UUID)..

I'm gonna change it to that actually

- [ ] `LatticeTargetId` -> `LatticeRouteName`

vados-cosmonic: Just chatted about this, landed on `ComponentId` for source, and `LatticeTarget` for target

crates/runtime/src/capability/builtin.rs
vados-cosmonic: Yup, will fix!

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: Removed!

PR #1572, opened 2024-02-22
feat(host)!: fully support component interface links, remove aliases

brooksmtownsend:
## Feature or Problem
This PR updates the plumbing around link definitions to completely support interface links and remove call aliases as they are no longer needed. Leaving in draft as I'd rather merge #1568 first since this branch contains a few commits from there to minimize conflicts. This also isn't completely finished.

Outstanding work to be done
- [ ] Publish linkdef set/deleted events
- [ ] Publish linkdef set/deleted events to providers (assuming they need this information, might be replaced with put/delete configuration)
- [ ] Improve the matching of `resolve_target`, since we are now only targeting wRPC entities. 
- [ ] Depending on the link, we _might_ be able to attempt to turn a previously stored LinkDefinition into an InterfaceLinkDefinition. Consider the usefulness of this migration.

### DIFF
--- a/crates/control-interface/src/types.rs
+++ b/crates/control-interface/src/types.rs
@@ -180,6 +180,12 @@ pub struct LinkDefinitionList {
     pub links: Vec<LinkDefinition>,
 }
 
+/// A list of interface link definitions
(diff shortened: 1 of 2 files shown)


brooksmtownsend:
Going to open a different PR

PR #1574, opened 2024-02-23
fix: port provider bindgen fixes

vados-cosmonic:
## Feature or Problem

This commit fixes the provider bindgen issues for non http-server builds (ex. kv-redis)


brooksmtownsend:
LGTM, just needs a rebase it looks like 😄 Should just be the types import line in runtime

brooksmtownsend (approved):
Good to me as long as @rvolosatovs / @thomastaylor312 are cool with commenting out tests for now. I know that the tests are going to need a wider sweeping update to get them to properly test wRPC things, so in my opinion this is fine for iterating on a feature branch and we'll add tests back in before the PR to main.

crates/host/src/wasmbus/mod.rs
```diff
@@ -177,6 +177,7 @@ impl Queue {
 #[derive(Debug)]
 struct ActorInstance {
     actor: wasmcloud_runtime::Component,
+    #[allow(unused)]
```
thomastaylor312, line 180: Should all of these just be removed? Or are we just keeping them around for now? If we're going to remove in a follow up, let's leave a TODO comment so these are easier to find later

thomastaylor312 (approved):
I'm good with this. One minor comment. Also, I would think that adding the `#[ignore]` attribute to the tests is probably better than commenting the whole thing out and makes it easier to review when we reenable, but that is more of a nit

vados-cosmonic:
@thomastaylor312 with regards to the ignoring the tests -- the tests don't compile at all because HTTP server is broken and no longer builable by the `build.rs` code -- that's why they're commented out.

crates/host/src/wasmbus/mod.rs
vados-cosmonic: Will remove in a follow up!

PR #1575, opened 2024-02-23
Control interface component ID support, flatten actor/provider instances

### LINKED ISSUE #1092: [FEATURE] Replace actor start/stop events with actor_scaled event

brooksmtownsend:
## Feature or Problem
This PR includes quite a few changes in the control interface and host, but in sum:
1. Updates actor and provider commands on the control interface to support pulling in component IDs, which is a unique identifier for the component itself and used in routing
2. Flattens both the `ActorInstance` and `ProviderInstance` into their parent `Actor` / `Provider` struct, now that component IDs uniquely identify the component we don't need to index based on annotations
3. Drop the `link_name` and `contract_id` from providers, since we don't need them anymore. I attempted to keep the `link_name` as default in many places for some amount of backwards compat with provider things.
4. Fully support interface links through the `get` / `put` / `del` control interface topics
5. Remove the concept of aliases, no longer needed!

## Related Issues
Original PR that I closed: #1572 
Fully fixes #1092 by removing old events

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -542,12 +542,7 @@ impl Client {
     /// _will not_ supply a discrete confirmation that a provider has terminated. For that kind of
     /// information, the client must also monitor the control event stream
     #[instrument(level = "debug", skip_all)]
-    pub async fn stop_provider(
-        &self,
-        host_id: &str,
-        provider_id: &str,
-        annotations: Option<HashMap<String, String>>,
-    ) -> Result<CtlOperationAck> {
+    pub async fn stop_provider(&self, host_id: &str, provider_id: &str) -> Result<CtlOperationAck> {
         let host_id = parse_identifier(&IdentifierKind::HostId, host_id)?;
 
         let subject =
@@ -556,7 +551,6 @@ impl Client {
         let bytes = json_serialize(StopProviderCommand {
             host_id,
             provider_id: parse_identifier(&IdentifierKind::ComponentId, provider_id)?,
-            annotations,
         })?;
 
         match self.request_timeout(subject, bytes, self.timeout).await {
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -313,50 +313,34 @@ impl Client {
 
     /// Puts a link into the lattice. Returns an error if it was unable to put the link
     #[instrument(level = "debug", skip_all)]
-    pub async fn advertise_link(
-        &self,
-        source_id: &str,
-        target: &str,
-        link_name: &str,
-        wit_namespace: &str,
-        wit_package: &str,
-        interfaces: Vec<String>,
-        source_config: Vec<String>,
-        target_config: Vec<String>,
-    ) -> Result<CtlOperationAck> {
-        let ld = InterfaceLinkDefinition {
-            source_id: parse_identifier(&IdentifierKind::ComponentId, source_id)?,
-            target: parse_identifier(&IdentifierKind::ComponentId, target)?,
-            name: parse_identifier(&IdentifierKind::LinkName, link_name)?,
-            wit_namespace: wit_namespace.to_string(),
-            wit_package: wit_package.to_string(),
-            interfaces,
-            source_config,
-            target_config,
-        };
+    pub async fn put_link(&self, link: InterfaceLinkDefinition) -> Result<CtlOperationAck> {
+        // Validate link parameters
+        parse_identifier(&IdentifierKind::ComponentId, &link.source_id)?;
+        parse_identifier(&IdentifierKind::ComponentId, &link.target)?;
+        parse_identifier(&IdentifierKind::LinkName, &link.name)?;
 
-        let subject = broker::advertise_link(&self.topic_prefix, &self.lattice);
-        debug!("advertise_link:request {}", &subject);
+        let subject = broker::put_link(&self.topic_prefix, &self.lattice);
+        debug!("put_link:request {}", &subject);
 
-        let bytes = crate::json_serialize(&ld)?;
+        let bytes = crate::json_serialize(&link)?;
         match self.request_timeout(subject, bytes, self.timeout).await {
             Ok(msg) => Ok(json_deserialize(&msg.payload)?),
-            Err(e) => Err(format!("Did not receive advertise link acknowledgement: {e}").into()),
+            Err(e) => Err(format!("Did not receive put link acknowledgement: {e}").into()),
         }
     }
 
-    /// Removes a link from the lattice metadata keyvalue bucket. Returns an error if it was unable
+    /// Deletes a link from the lattice metadata keyvalue bucket. Returns an error if it was unable
     /// to delete. This is an idempotent operation.
     #[instrument(level = "debug", skip_all)]
-    pub async fn remove_link(
+    pub async fn delete_link(
         &self,
         source_id: &str,
         link_name: &str,
         wit_namespace: &str,
         wit_package: &str,
     ) -> Result<CtlOperationAck> {
-        let subject = broker::remove_link(&self.topic_prefix, &self.lattice);
-        let ld = RemoveInterfaceLinkDefinitionRequest {
+        let subject = broker::delete_link(&self.topic_prefix, &self.lattice);
+        let ld = DeleteInterfaceLinkDefinitionRequest {
             source_id: parse_identifier(&IdentifierKind::ComponentId, source_id)?,
             name: parse_identifier(&IdentifierKind::LinkName, link_name)?,
             wit_namespace: wit_namespace.to_string(),
@@ -365,7 +349,7 @@ impl Client {
         let bytes = crate::json_serialize(&ld)?;
         match self.request_timeout(subject, bytes, self.timeout).await {
             Ok(msg) => Ok(json_deserialize(&msg.payload)?),
-            Err(e) => Err(format!("Did not receive remove link acknowledgement: {e}").into()),
+            Err(e) => Err(format!("Did not receive delete link acknowledgement: {e}").into()),
         }
     }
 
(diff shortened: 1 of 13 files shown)


crates/control-interface/src/types.rs
```diff
@@ -3,7 +3,14 @@ use std::collections::HashMap;
 use anyhow::bail;
 use serde::{Deserialize, Serialize};
 
-/// One of a potential list of responses to an actor auction
+pub type ConstraintMap = std::collections::HashMap<String, String>;
+pub type AnnotationMap = std::collections::HashMap<String, String>;
+pub type KeyValueMap = std::collections::HashMap<String, String>;
+pub type LabelsMap = std::collections::HashMap<String, String>;
+pub type ProviderDescriptions = Vec<ProviderDescription>;
```
thomastaylor312, line 10: Nit: My other comment around type aliases I could be convinced away from, but I _despise_ aliased Vecs like this 😆

crates/control-interface/src/types.rs
```diff
@@ -198,19 +192,18 @@ pub struct ProviderAuctionAck {
 }
 
 /// A request to locate a suitable host for a capability provider. The
-/// provider's unique identity (reference + link name) is used to rule
-/// out sites on which the provider is already running.
+/// provider's unique identity is used to rule out sites on which the
```
thomastaylor312, line 195: Nit: Change "sites" to "hosts?"

crates/control-interface/src/types.rs
```diff
@@ -220,18 +213,12 @@ pub struct ProviderDescription {
     /// this provider instance
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub annotations: Option<AnnotationMap>,
-    /// Provider's unique 56-character ID
+    /// Provider's unique identifier
     #[serde(default, alias = "public_key")]
```
thomastaylor312, line 217: Probably time to remove the alias here

crates/control-interface/src/types.rs
```diff
@@ -299,23 +284,25 @@ pub type RegistryCredentialMap = std::collections::HashMap<String, RegistryCrede
...
+pub struct RemoveInterfaceLinkDefinitionRequest {
+    /// The source component's identifier.
+    pub source_id: ComponentId,
+    /// Name of the link. Not providing this is equivalent to specifying Some("default")
+    #[serde(default = "default_link_name")]
+    pub name: LinkName,
+    /// WIT namespace of the link, e.g. `wasi` in `wasi:keyvalue/readwrite.get`
+    pub wit_namespace: WitNamespace,
+    /// WIT package of the link, e.g. `keyvalue` in `wasi:keyvalue/readwrite.get`
+    pub wit_package: WitPackage,
```
thomastaylor312, line 296: Do we need the interface name as well? What if I want to unlink `wasi:keyvalue/readwrite` but not `wasi:keyvalue/atomic`?

crates/host/src/wasmbus/event.rs
```diff
@@ -36,65 +36,11 @@ fn format_actor_claims(claims: &jwt::Claims<jwt::Actor>) -> serde_json::Value {
     }
 }
 
-// TODO(#1092): Remove this event in favor of `actor_scaled`
-pub fn actors_started(
```
thomastaylor312, line 40: tenor-117690317

crates/host/src/wasmbus/mod.rs
```diff
@@ -284,26 +196,20 @@ struct Handler {
     #[allow(clippy::type_complexity)]
     interface_links:
         Arc<RwLock<HashMap<LinkName, HashMap<String, HashMap<WitInterface, LatticeTargetId>>>>>,
-    aliases: Arc<RwLock<HashMap<String, WasmCloudEntity>>>,
-    // interface -> function -> result types
+    /// Map of interface -> function name -> function result types
+    ///
+    /// When invoking a function that the cmoponent imports, this map is consulted to determine the
```
thomastaylor312, line 201: Nit(spelling): component

crates/host/src/wasmbus/mod.rs
```diff
@@ -2374,108 +2215,93 @@ impl Host {
...
+        let constraints_satisfied = constraints
+            .iter()
+            .all(|(k, v)| host_labels.get(k).map(|hv| hv == v).unwrap_or(false));
+        let actor_id_running = self.actors.read().await.contains_key(&actor_id);
+
+        // This host can run the actor if all constraints are satisfied and the actor is not already running
+        if constraints_satisfied && !actor_id_running {
+            let buf = serde_json::to_vec(&ActorAuctionAck {
+                actor_ref,
+                constraints,
```
thomastaylor312, line 2297: Should we return the actor ID back here as well? Feels like we should because this is a host saying "Yes, I can run actor with $REF and $ID"

crates/host/src/wasmbus/mod.rs
```diff
@@ -2485,38 +2311,35 @@ impl Host {
...
-                 ..
-             }| { *image_ref == provider_ref && instances.contains_key(&link_name) },
-        ) {
-            // Do not reply if the provider is already running
-            return Ok(None);
+        let provider_running = providers.contains_key(&provider_id);
+        if constraints_satisfied && !provider_running {
+            let buf = serde_json::to_vec(&ProviderAuctionAck {
+                provider_ref,
+                constraints,
```
thomastaylor312, line 2335: Same note here around ID

crates/host/src/wasmbus/mod.rs
```diff
@@ -2374,108 +2215,93 @@ impl Host {
...
-        entry: hash_map::OccupiedEntry<'a, String, Arc<Actor>>,
-        annotations: &BTreeMap<String, String>,
-        host_id: &str,
-    ) -> anyhow::Result<()> {
-        trace!(actor_id = %entry.key(), "stopping actor");
-
-        let actor = entry.remove();
-        let claims = actor.claims().context("claims missing")?;
-        let mut instances = actor.instances.write().await;
+    async fn stop_actor(&self, actor: Arc<Actor>, host_id: &str) -> anyhow::Result<()> {
```
thomastaylor312, line 2247: Note: I think this can take a borrowed actor (as long as `actor.calls.abort()` is `&self`, which I think it is). Feels a little odd to clone the arc in order to stop the actor

crates/host/src/wasmbus/mod.rs
```diff
@@ -3374,14 +3071,13 @@ impl Host {
...
+    #[instrument(level = "debug", skip_all)]
     async fn handle_links(&self) -> anyhow::Result<Bytes> {
-        trace!("handling links"); // FIXME: set back to debug when instrumentation is re-enabled
+        debug!("handling links");
 
         let links = self.links.read().await;
-        let links: Vec<LinkDefinition> = links.values().cloned().collect();
-        let res = serde_json::to_vec(&LinkDefinitionList { links })
-            .context("failed to serialize response")?;
+        let links: Vec<InterfaceLinkDefinition> = links.values().flatten().cloned().collect();
```
thomastaylor312, line 3079: Given that this is a fairly common call for wadm, I think we should avoid the clone here and just get a vec of borrowed linkdefs. Serde only needs to borrow the data here anyway

crates/host/src/wasmbus/mod.rs
```diff
@@ -3514,62 +3213,119 @@ impl Host {
...
+                wit_namespace,
+                wit_package,
+                &interfaces,
+                &source_config,
+                &target_config,
+            ),
+        )
+        .await?;
+        // TODO(#1548): When providers can handle interface links, tell them to set the link.
+        // Alternatively, send them configuration cc @thomastaylor312
```
thomastaylor312, line 3247: Yep, I plan on sending config right after linkdef put

thomastaylor312 (changes requested):
Overall this looks mostly ready. Just some questions about API design and handling to work through

crates/control-interface/src/lib.rs
```diff
@@ -349,13 +359,13 @@ impl Client {
...
     #[instrument(level = "debug", skip_all)]
-    pub async fn query_links(&self) -> Result<Vec<LinkDefinition>> {
+    pub async fn query_links(&self) -> Result<Vec<InterfaceLinkDefinition>> {
         let subject = broker::queries::link_definitions(&self.topic_prefix, &self.lattice);
         debug!("query_links:request {}", &subject);
         match self.request_timeout(subject, vec![], self.timeout).await {
             Ok(msg) => {
-                let list: LinkDefinitionList = json_deserialize(&msg.payload)?;
-                Ok(list.links)
+                let links: Vec<InterfaceLinkDefinition> = json_deserialize(&msg.payload)?;
```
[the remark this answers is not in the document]
brooksmtownsend: I do want to do a wrapper type! I will add in another commit with that change so it's easy to review

crates/control-interface/src/types.rs
brooksmtownsend: That's a good question. It's a little ambiguous and probably up to our own design, but what I was thinking was that if you put a link with interfaces then in the future you can either put the link with different interfaces, or remove it.

E.g. if you wanted to unlink `readwrite`, you'd re-put the link with just `atomic`.

brooksmtownsend: If we did want to support something like this, it would be nice to not have to specify every interface for the unlinking operation, at least at the CLI level. This command is getting pretty big:
```bash
wash link del mysource --namespace wasi --package keyvalue --interface atomic --interface eventual --name default
```

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Yeah that's a good idea, no reason not to

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Maybe we can just move it 🤔 if we're stopping it we shouldn't use it afterwards

crates/control-interface/src/types.rs
thomastaylor312: I’m ok if we shortcut and use the package name to unlink all interfaces, but I would want a way to not unlink everything as well. I think this is good enough for now and something like this could be added later without breaking

crates/control-interface/src/lib.rs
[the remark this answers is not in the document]
brooksmtownsend: #1588

thomastaylor312 (approved):

PR #1576, opened 2024-02-23
feat(*): Glues in named config to actors

thomastaylor312:
This introduces a new config bundle that can watch for config changes. There is probably a way to reduce the number of allocations here, but it is good enough for now.

Also, sorry for the new file. I renamed `config.rs` to `host_config.rs` so I could reuse the `config.rs` file, but I forgot to git mv. So that file hasn't changed


thomastaylor312:
Keeping this as a draft until some of the latest changes land from @brooksmtownsend. When those land, I'll rebase and add another test update for this

crates/host/src/wasmbus/config.rs
```diff
@@ -1,107 +1,423 @@
...
+
+async fn update_merge(
+    merged_config: &RwLock<HashMap<String, String>>,
+    changed_notifier: &Sender<()>,
+    ordered_receivers: &[Receiver<HashMap<String, String>>],
+) {
+    // We get a write lock to start so nothing else can update the merged config while we merge
+    // in the other configs (e.g. when one of the ordered configs is write locked)
+    let mut lock = merged_config.write().await;
+    lock.clear();
```
brooksmtownsend, line 325: I can't find this function in the tokio `RwLock` docs, what is this for?

brooksmtownsend (approved):
Looks great, huge fan of inline docs and additional tests. We should probably rename it to something other than `pinger` and `ponger` 😅 

I had a question about a function called but LGTM

crates/host/src/wasmbus/config.rs
thomastaylor312: This is `clear` from HashMap

PR #1579, opened 2024-02-26
feat(tests): add wasmcloud-test-util crate

vados-cosmonic:
## Feature or Problem

This commit adds a `wasmcloud-test-util` crate, which contains utilities for testing wasmCloud hosts, providers, and actors locally

### DIFF
--- a/crates/test-util/src/actor.rs
+++ b/crates/test-util/src/actor.rs
@@ -1,0 +1,108 @@
+//! Actor management utilities for use during testing
+
+use std::{
+    collections::HashMap,
+    num::{NonZeroU32, NonZeroUsize},
+    path::Path,
+};
+
+use anyhow::{anyhow, bail, ensure, Context as _, Result};
+use nkeys::KeyPair;
+use serde::Deserialize;
+use std::time::Duration;
+use tokio::fs;
+
+use wascap::{jwt, wasm::extract_claims};
+use wasmcloud_control_interface::{Client as WasmCloudCtlClient, CtlOperationAck};
+
+/// This is a *partial* struct for the ActorScaled event, which normally consists of more fields
+#[derive(Deserialize)]
+struct ActorScaledEvent {
+    pub max_instances: NonZeroUsize,
+}
+
+/// Given a path to an actor on disks, extract claims
+pub async fn extract_actor_claims(
+    wasm_binary_path: impl AsRef<Path>,
+) -> Result<jwt::Claims<jwt::Actor>> {
+    let wasm_binary_path = wasm_binary_path.as_ref();
+    let jwt::Token { claims, .. } = extract_claims(fs::read(wasm_binary_path).await?)
+        .context("failed to extract kv http smithy actor claims")?
+        .context("component actor claims missing")?;
+    Ok(claims)
+}
+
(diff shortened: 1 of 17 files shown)


crates/test-util/src/host.rs
```diff
@@ -0,0 +1,121 @@
...
+        let lattice_name = lattice_name.as_ref();
+        let cluster_key = Arc::new(cluster_key.unwrap_or(KeyPair::new_cluster()));
+        let host_key = Arc::new(host_key.unwrap_or(KeyPair::new_server()));
+
+        let (host, shutdown_hook) = Host::new(HostConfig {
+            ctl_nats_url: nats_url.clone(),
+            rpc_nats_url: nats_url.clone(),
+            lattice: lattice_name.into(),
+            cluster_key: Some(Arc::clone(&cluster_key)),
+            cluster_issuers: Some(vec![cluster_key.public_key(), cluster_key.public_key()]),
```
brooksmtownsend, line 79: Nit: double reference to cluster public key here

crates/test-util/src/actor.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut sub_started = nats_client
+        .subscribe(format!("wasmbus.evt.{lattice_name}.actors_started"))
+        .await?;
+    let mut sub_stopped = nats_client
+        .subscribe(format!("wasmbus.evt.{lattice_name}.actors_stopped"))
+        .await?;
+    let CtlOperationAck { accepted, error } = ctl_client
+        .scale_actor(&host_key.public_key(), url.as_ref(), count, annotations)
+        .await
+        .map_err(|e| anyhow!(e).context("failed to start actor"))?;
```
brooksmtownsend, line 86: Now that we have `actor_scaled`, you can actually just assert that we receive that event and that it's the count we asked for!

brooksmtownsend (approved):
One nit and one improvement suggestion, but neither are blocking the utility of moving the common test functionality into a reusable crate that we can eventually publish.

crates/test-util/src/host.rs
vados-cosmonic: thanks, deduped!

crates/test-util/src/actor.rs
vados-cosmonic: Does the comment below not apply? Waiting for stopped/started instead of scale because of scale being early-ACK? Has how it worked changed?

brooksmtownsend: Oh! Yes that still applies. I was just saying that instead of waiting on actors_started or actors_stopped, you can simply wait for actor_scaled and that will contain the count that it was scaled to

We added that new event to cut down on the number of events, how weird it was to calculate new vs old running instances, and now the event is more idempotent

vados-cosmonic: Right, but why the code waited for started/stopped was because scale ACKs early... Are you saying we don't need to do that, and I can wait for just `actor_scaled`? If I can, then I'll do that!

vados-cosmonic: Are start actor/scale actor broken? start provider works just fine as the previous code, but `event_receiver` utility doesn't seem to work with started or scaled actors.

PR #1582, opened 2024-02-26
fix macro generation wit functions

LUK3ARK:
## Feature or Problem
When generating macros for my custom wit providers i would get error messages saying 'not a valid identifier'
e.g. 
   = help: message: `"org.package.methodnameInvocation"` is not a valid identifier

# Proposed solution
It seems that the method names separated by periods may be used for lattice linking and for generating the rust structs, where this seems it needs to be sanitised and does not correctly line up with the doc comments above.

# Potential issue
Ensure that this lines up with the lattice method mapping inside Wasmcloud.

### DIFF
--- a/crates/provider-wit-bindgen-macro/src/wit.rs
+++ b/crates/provider-wit-bindgen-macro/src/wit.rs
@@ -520,7 +520,7 @@ impl WitFunctionLatticeTranslationStrategy {
         // (ex. MessagingConsumerRequestMultiInvocation)
         let struct_name = format_ident!(
             "{}{}Invocation",
-            wit_iface_name.to_upper_camel_case(),
+            wit_iface_name.replace(".", "").to_upper_camel_case(),
             trait_method.sig.ident.to_string().to_upper_camel_case()
         );
 


vados-cosmonic:
Hey @LUK3ARK could you post a snippet your intended WIT contract? 

AFAIK `.` isn't valid inside WIT identifiers and you shouldn't have it in your interface name. 

See the following links:

- https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-identifiers
- https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions

vados-cosmonic (changes requested):
See my earlier comment regarding naming

LUK3ARK:
package myorg:my-custom-messaging;

interface handler {
    use myorg:dependency/tx.{transaction-with-meta};
    use account.{account-state-update};

    handle-tx-message: func(msg: transaction-with-meta) -> result<_, string>;

    handle-account-message: func(msg: account-state-update) -> result<_, string>;
}

here is an example of one of my wit definitions. I don't think i have any wit interfaces with . in any of my naming.

LUK3ARK:
I assumed the . came from the wasmcloud side and you use it in mapping function signatures to the lattice runtime.

vados-cosmonic:
Thanks for noting your wit -- I've reproduced it below (GH now supports highlighting WIT):

```wit
package myorg:my-custom-messaging;

interface handler {
  use myorg:dependency/tx.{transaction-with-meta};
  use account.{account-state-update};

  handle-tx-message: func(msg: transaction-with-meta) -> result<_, string>;

  handle-account-message: func(msg: account-state-update) -> result<_, string>;
}
```

Could I see your `world` as well? I'd like to completely reproduce your case and to do so I'd need to see how at least one of them is being exported (the error is happening when translating an exported function).

That said, I *think* I have [a fix for the bug](https://github.com/wasmCloud/wasmCloud/pull/1583), so you could also try my work on this branch, by using my version temporarily in your `Cargo.toml` (or making the changes locally):

```toml
wasmcloud-provider-wit-bindgen = { git = "https://github.com/vados-cosmonic/wasmCloud.git", branch = "fix(provider-wit-bindgen)=exported-bundle-args-invocation-naming" }
```

LUK3ARK:
Thank you Victor.

`package myorg:my-custom-messaging-consumer;

world provider-message-consumer {
    export myorg:my-custom-messaging/handler;
    import myorg:messaging/tx;
    import myorg:messaging/account;
}
`

here is my world definition.

LUK3ARK:
I also have seen this issue occur for the member name of the struct as well not just for functions in the past, however not entirely sure if that was my fault. Just something to be aware of.

vados-cosmonic:
Appreciate it @LUK3ARK ! Let me know if my branch works for you if it's not too much trouble -- I just made a tiny reproduction of your stuff, and it works -- `MyorgMyCustomMessagingHandler` etc gets generated.

vados-cosmonic:
Ah I think that's actually separate -- it looks like it's having a problem not with `Invocation`s anymore, but with another bit, will take another look

vados-cosmonic:
Hey @LUK3ARK it seems like my reproduction is not matching what you have, but I've done the replacement in more places. Did the arguments to the exports, or the import/export mixture change? I can't figure out how that handler is ending up there. 

That said, please try the branch again

LUK3ARK:
Still no luck, I reached out to you on Slack whenever you get chance

vados-cosmonic:
Closing this in favor of https://github.com/wasmCloud/wasmCloud/pull/1583 since we just confirmed that it's working! Hopefully we can get that out and released in a minor bugfix version very quickly.

vados-cosmonic:
Thanks again for your help @LUK3ARK -- the fixes have made their way to main, and we have a [version bump](https://github.com/wasmCloud/wasmCloud/pull/1584) incoming as well

PR #1583, opened 2024-02-26
fix(provider-wit-bindgen): exported bundled arg invocation name

vados-cosmonic:
## Feature or Problem

This commit fixes a bug in determining the struct name for bundled exports. Code that was using expected a WIT interface name was receiving a WIT inteface path (which is `.`-delimited), and attempting to use it directly, rather that converting to a valid Rust identifier.


LUK3ARK:
I also have seen this issue occur for the member name of the struct as well not just for functions in the past, however not entirely sure if that was my fault. Just something to be aware of.

thomastaylor312 (approved):

PR #1587, opened 2024-02-27
fix(provider-wit-bindgen): argument bundling & type resolution

vados-cosmonic:
## Feature or Problem

This commit fixes two issues with provider bindgen:

- Types replaced as part of structs were using their full qualified paths, depsite types being lifted to the top level (with structs), not needing it.

- Invocation arguments for `LatticeMethod`s that were recorded were not being saved when processing structs. This comes up when exporting a function that needed it's arguments bundled into a struct


ricochet (approved):
I wish there was a test to see the difference in output but I understand how that is likely difficult to add right now.

PR #1588, opened 2024-02-27
feat(control-interface)!: wrap all operations in CtlResponse

brooksmtownsend:
## Feature or Problem
This PR tackles a large refactor to better wrap errors and responses for the control interface. After #1108 our topics are much more predictable, but it still remained difficult to differentiate between protocol errors, deserialization errors, and control interface handling errors. This PR makes this _much_ easier to do by more strongly typing handlers in the host and the control interface, so as a user you can tell the difference between these issues.

Note that a good bit of the handlers use helpers like `CtlResponse::success()` or return `CtlResponse<()>`, which is because of the generic type. Even if we return `None` as a response, we still need to "type" it just in case it was a Some. This works pretty well though.

## Related Issues
This is based off of #1575 and moved into its own PR because it's sufficiently large.

### Unit Test(s)
A big part of this PR is the `ctl_response_comprehensive` unit test which is ignored right now for general logistics, but passes perfectly well with a running NATS server and host. I wanted to ensure that each and every one of our control interface operations is working properly, albeit with one or two error cases that I'm taking advantage of. This is a large enough change that ensuring that it's working end-to-end is important so we know the control interface is sound.

### Acceptance or Integration
We'll fully exercise this functionality with wRPC integration tests, but it will be important to handle error cases well on the client side depending on the operation's outer `Result` and inner `success` member.

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -3,6 +3,12 @@
 //! This library provides a client API for consuming the wasmCloud control interface over a
 //! NATS connection. This library can be used by multiple types of tools, and is also used
 //! by the control interface capability provider and the wash CLI
+//!
+//! ## Usage
+//! All of the [Client] functions are handled by a wasmCloud host running in the specified lattice.
+//! Each function returns a Result<CtlResponse<T>> wrapper around the actual response type. The outer
+//! result should be handled for protocol (timeouts, no hosts available) and deserialization errors (invalid response payload).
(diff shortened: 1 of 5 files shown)


thomastaylor312 (approved):

brooksmtownsend:
Will merge once I can confirm it's not going to disrupt @vados-cosmonic too hard

vados-cosmonic:
Hey go ahead and merge this @brooksmtownsend ! I'm ready to rebase once you do.

brooksmtownsend:
@vados-cosmonic Done!

PR #1592, opened 2024-02-28
fix(wasmbus): correctly invoke custom functions

rvolosatovs:
fix wRPC usage:
- don't spoof function call responses (this potentially may cause logical errors in the application, which is very dangerous)
- ensure parameters are transmitted fully before returning

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -740,30 +740,21 @@ impl Bus for Handler {
         name: &str,
         params: Vec<wrpc_transport::Value>,
     ) -> anyhow::Result<Vec<wrpc_transport::Value>> {
-        if let Some(TargetEntity::Wrpc(interface_target)) = target {
-            let result_types = match self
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -740,30 +740,21 @@ impl Bus for Handler {
...
-
-            Ok(result)
+                .and_then(|functions| functions.get(name)).with_context(|| format!("polyfilled import {instance}/{name} not found, could not determine result types"))?;
+            let (results, tx) = wrpc_transport_nats::Client::new(
+                self.nats.clone(),
+                format!("{}.{id}", self.lattice),
+            )
+            .invoke_dynamic(instance, name, params, results)
+            .await?;
+            tx.await.context("failed to transmit parameters")?;
```
brooksmtownsend, line 754: Ah didn't know we needed to await transmission, makes total sense though.

brooksmtownsend (approved):

PR #1596, opened 2024-02-28
`wrpc:http/outgoing-handler` implementation

rvolosatovs:
Tested with an ad-hoc hand-written "provider" in the `wrpc` test

### DIFF
--- a/crates/runtime/src/actor/component/http.rs
+++ b/crates/runtime/src/actor/component/http.rs
@@ -1,7 +1,7 @@
 use super::{Ctx, Instance, InterfaceInstance};
 
 use crate::capability::http::types;
-use crate::capability::{IncomingHttp, OutgoingHttp, OutgoingHttpRequest};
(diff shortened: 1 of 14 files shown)


crates/runtime/src/actor/component/http.rs
```diff
@@ -43,44 +43,21 @@ impl WasiHttpView for Ctx {
...
-                OutgoingHttpRequest {
-                    use_tls,
-                    authority,
-                    request,
-                    connect_timeout,
-                    first_byte_timeout,
+            match OutgoingHttp::handle(&handler, request).await {
+                Ok(Ok(resp)) => Ok(Ok(IncomingResponseInternal {
+                    resp,
+                    worker: Arc::new(preview2::spawn(async {})),
```
rvolosatovs, line 57: note to self: we should use this task for I/O eventually - will have to propagate a future here through the interface

PR #1598, opened 2024-02-28
feat(runtime,host): Updates to request throughput

thomastaylor312:
This PR consists of two commits that help improve the ability of the system to manage higher numbers of requests per second. The first commit updates the components to be pre-instantiated rather than doing a full instantiation each time. The second commit fixes an issue where we were write locking on the policy decision cache for every request. Now we only write lock when necessary. 

In my load testing of this locally, these led to a 25-28% increase in throughput. Please note I only did the preinstantiation for components as we're removing module support for 1.0

### DIFF
--- a/crates/host/src/policy.rs
+++ b/crates/host/src/policy.rs
@@ -1,6 +1,6 @@
 use core::time::Duration;
 
-use std::collections::{hash_map, HashMap};
+use std::collections::HashMap;
 use std::hash::Hash;
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


thomastaylor312:
Please note that if we think this is important enough, we can bring in this, and the "utilize all cores" fix from #1376, in for a 0.82.1, but I honestly think we can just wait until 1.0

thomastaylor312:
Yeah I can change it

rvolosatovs (approved):
Nice!

PR #1602, opened 2024-02-29
feat(metrics) add invocation and error counts for actor invocations

protochron:
## Feature or Problem
Add two new metrics for actors:
* the count of the number of invocations (`wasmcloud_host.actor.invocations`)
* the count of errors (`wasmcloud_host.actor.errors`)

This also adds a bunch of new attributes to the existing actor metrics so that they make sense in an environment with multiple hosts. Specifically this adds:
* the lattice ID
* the host ID
* provider information if a provider invoked the actor: ** the contract ID
  * the provider ID
  * the name of the linkdef

For actor to actor calls, instead of having the provider metadata it instead has the public key of the invoking actor.

An example of what this looks like as an exported Prometheus metric:

```
wasmcloud_host_actor_invocations_total{actor_ref="wasmcloud.azurecr.io/echo:0.3.8", caller_provider_contract_id="wasmcloud:httpserver", caller_provider_id="VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M", caller_provider_link_name="default", host="ND7L3RZ6NLYJGN25E6DKYS665ITWXAPXZXGZXLCUQEDDU65RB5TVUHEN", job="wasmcloud-host", lattice="default", operation="HttpServer.HandleRequest"}
```

Provider metrics will likely need to wait until the wRPC work is finished.

## Related Issues
A start on adding some of the additional metrics highlighted in #664.

## Testing
Tested locally with an OTEL collector relaying metrics to Prometheus.

### DIFF
--- a/crates/host/src/metrics.rs
+++ b/crates/host/src/metrics.rs
@@ -1,24 +1,53 @@
-use wasmcloud_tracing::{Histogram, Meter};
+use wasmcloud_tracing::{Counter, Histogram, Meter, Unit};
 
 /// `HostMetrics` encapsulates the set of metrics emitted by the wasmcloud host
 #[derive(Clone, Debug)]
 #[allow(clippy::module_name_repetitions)]
 pub struct HostMetrics {
     /// Represents the time it took for each handle_rpc_message invocation in nanoseconds.
-    pub wasmcloud_host_handle_rpc_message_duration_ns: Histogram<u64>,
+    pub handle_rpc_message_duration_ns: Histogram<u64>,
+    /// The count of the number of times an actor was invoked.
+    pub actor_invocations: Counter<u64>,
+    /// The count of the number of times an actor invocation resulted in an error.
+    pub actor_errors: Counter<u64>,
+
+    /// The host's ID.
+    // TODO this is actually configured as an InstrumentationScope attribute on the global meter,
+    // but we don't really hve a way of getting at those. We should figure out a way to get at that
+    // information so we don't have to duplicate it here.
+    pub host_id: String,
+    /// The host's lattice ID.
+    // Eventually a host will be able to support multiple lattices, so this will need to either be
+    // removed or metrics will need to be scoped per-lattice.
+    pub lattice_id: String,
 }
 
 impl HostMetrics {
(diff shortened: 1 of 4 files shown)


crates/host/src/metrics.rs
```diff
@@ -1,24 +1,53 @@
...
+            .init();
+
+        let actor_invocation_count = meter
+            .u64_counter("wasmcloud_host.actor.invocations")
+            .with_description("Number of actor invocations")
+            .init();
+
+        let actor_error_count = meter
+            .u64_counter("wasmcloud_host.actor.errors")
+            .with_description("Number of actor errors")
```
vados-cosmonic, line 42: Does it make more sense to call this `actor_invocation_error_count`, since we're tracking errors of actor invocations in particular? "actor error" is a bit vague

vados-cosmonic (approved):
LGTM :rocket:

crates/host/src/metrics.rs
vados-cosmonic: Also, "actor invocations" seems like it might be confusable -- does it include errored invocations? maybe a more specific name might be helpful, if it's only counting *successful* invocations then maybe that's worth calling out -- unless the expectation (and we should be consistent going forward) is that if "successful" is not stated in the metric, it's implied.

protochron: > Does it make more sense to call this `actor_invocation_error_count`, since we're tracking errors of actor invocations in particular? "actor error" is a bit vague

It actually ends up being 'actor_error_count' when it makes it to the exporter and Prometheus because the unit of a counter is implicitly a count. The example in the PR description shows that I think

vados-cosmonic: ah yeah but the fact that it's an error for an actor *invocation* isn't clear right -- actors could fail to start up as well. But if we think that the only kind of error actors will have is invocation errors, then no need to be more specific.

protochron: Updated to be `wasmcloud_host.actor.invocation.errors`

vados-cosmonic:
Tried to merge but it looks like we need an approval from @thomastaylor312

brooksmtownsend:
@thomastaylor312 bump

thomastaylor312 (approved):

PR #1604, opened 2024-02-29
feat(host): switch to `wrpc:{keyvalue,blobstore}`

rvolosatovs:
No tests yet, let's get providers working first and then use the existing test bench for this

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -8,7 +8,6 @@ use std::collections::hash_map::{self, Entry};
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::env;
 use std::env::consts::{ARCH, FAMILY, OS};
-use std::io::Cursor;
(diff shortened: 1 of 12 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -335,93 +334,106 @@ fn decode_empty_provider_response(buf: impl AsRef<[u8]>) -> anyhow::Result<()> {
...
-            target,
-            "wasmcloud:blobstore/Blobstore.CreateContainer",
-            &name,
-        )
-        .await
-        .and_then(decode_empty_provider_response)
+            .await?
+            .context("unknown target")?;
+        let wrpc =
+            wrpc_transport_nats::Client::new(self.nats.clone(), format!("{}.{id}", self.lattice));
```
brooksmtownsend, line 349: Note to myself, need to ensure we are propagating traces for each interface here too.

brooksmtownsend (approved):

PR #1606, opened 2024-02-29
feat(host): wasmcloud_transport wrappers to include headers on invocations

brooksmtownsend:
## Feature or Problem
This PR wraps a few traits from [wrpc-transport](https://crates.io/crates/wrpc-transport) in order to pass and send back headers when handling invocations. We already receive headers from incoming invocations, this PR just ensures that we are sending traceparent headers _back_ in response to an invocation and to send them when creating an outbound invocation.

### Manual Verification
This works and propagates headers all the way through! I tested it by invoking what we do in the wrpc integration test but invoking manually with a NATS header. For the curious:
```bash
> nats req "default.header.wrpc.0.0.1.wasmcloud:testing/pingpong.ping" "{}" -H source:myterminalteehee
[#15] Received on "default.header.wrpc.0.0.1.wasmcloud:testing/pingpong.ping" with reply "_INBOX.fnOVvdeTxGZoQazDtpSXAY.p0ShVKPK"
source: myterminalteehee

{}


[#16] Received on "_INBOX.fnOVvdeTxGZoQazDtpSXAY.p0ShVKPK" with reply "_INBOX.9ygF5OmJmpbGXiyXjq4LN4"
nil body


[#17] Received on "_INBOX.fnOVvdeTxGZoQazDtpSXAY.p0ShVKPK.results"
source: myterminalteehee

pong
```

Just a note that I tested this, and we _don't_ implicitly propagate all headers all the way through every single invocation, and that's actually intended. We want to be able to attach specific headers to invocations, and with tracing, we want to be able to propagate the traceparent header which is done using our injector traits. That's the only remaining TODO, to ensure that works well.

More verification, we can carry trace state and a trace parent all the way through!
```bash
[#557] Received on "_INBOX.6njCcm7AE4nmZIQCVPSHLD.results"
source-id: ponger
tracestate:
traceparent: 00-49089ae55b5d79d934114fe3032292d5-08fe6951080b0567-01

[#558] Received on "_INBOX.wN5l96SWWAfxigPUsPhlZK.cUMj2b66.results"
firstinvoker: myterminal

mPing pong, meaning of universe is: 42, split: ["hi", "there", "friend"], is_same: true, archie good boy: true
```

image
image

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -8,7 +8,6 @@ use std::collections::hash_map::{self, Entry};
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::env;
 use std::env::consts::{ARCH, FAMILY, OS};
-use std::f32::consts::E;
 use std::process::Stdio;
 use std::str::FromStr;
 use std::sync::Arc;
@@ -615,7 +614,7 @@ impl Bus for Handler {
             .unwrap_or_default()))
     }
 
-    #[instrument(level = "info", skip(self, target, params, instance, name), fields(interface = instance, function = name))]
+    #[instrument(level = "info", skip(self, params, instance, name), fields(interface = instance, function = name))]
     async fn call(
         &self,
         target: Option<TargetEntity>,
@@ -1142,7 +1141,7 @@ impl Actor {
 
     /// Handle an incoming wRPC request to invoke an export on this actor instance.
     #[instrument(
-        level = "debug",
+        level = "info",
         skip(self, context, result_subject, transmitter),
         fields(
             component_id = self.id.as_str(),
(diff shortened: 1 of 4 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -622,9 +626,14 @@ impl Bus for Handler {
                 .polyfilled_imports
                 .get(instance)
                 .and_then(|functions| functions.get(name)).with_context(|| format!("polyfilled import {instance}/{name} not found, could not determine result types"))?;
-            let (results, tx) = wrpc_transport_nats::Client::new(
+
+            let injector = TraceContextInjector::default_with_span();
+            let mut headers = injector_to_headers(&injector);
+            headers.insert("source-id", self.component_id.as_str());
```
brooksmtownsend, line 632: Note here this isn't set-in-stone, I just wanted to show an example to demonstrate how we can show the source of an invocation. Realistically I think what we want to pass here is the named target configuration for this specific link, but I'm leaving it as a placeholder for @thomastaylor312 and I to 🍐 on implementing

brooksmtownsend:
If this looks good to folks feel free to merge 👍🏻

vados-cosmonic (approved):
LGTM :rocket:

PR #1612, opened 2024-03-01
fix(host): improve error handling for wrpc links

brooksmtownsend:
## Feature or Problem
This PR updates a few of our internal structs for target entity discovery and routing and removes some outdated enum members. In addition, this PR includes some updates to error messages when a component isn't linked on a given interface.

If a component attempts to invoke something it's not linked to, here are the logs and error message:
```
2024-03-01T15:45:38.757607Z DEBUG handle_invocation{params=Custom { interface: "wasmcloud:testing/invoke", function: "call" } component_id="ponger" component_ref="file:///Users/brooks/github.com/wasmcloud/wasmcloud/tests/actors/rust/wrpc-pinger-component/build/wrpc_pinger_component_s.wasm"}:call{interface="wasmcloud:testing/invoke" function="call"}: wasmcloud_host::wasmbus: component is not linked to a lattice target for the given interface namespace="wasmcloud" package="testing" interface="pingpong" __self.component_id="ponger"
2024-03-01T15:45:38.758697Z ERROR wasmcloud_host::wasmbus: failed to handle invocation err=failed to call actor

Caused by:
    0: failed to call function
    1: error while executing at wasm backtrace:
           0: 0x19f264 - wit-component:shim!indirect-wasmcloud:testing/pingpong-ping
           1:  0xfce - <unknown>!wasmcloud:testing/invoke#call
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information
    2: failed to identify interface target
```

The goal here is to help developers identify the problem, here being that the component isn't linked to a target for that interface. 

## Consumer Impact
@vados-cosmonic this might touch some of the same code you're working on for provider invocation, so if it does I'll rebase this on top of your work.

### DIFF
--- a/crates/runtime/src/capability/builtin.rs
+++ b/crates/runtime/src/capability/builtin.rs
@@ -9,12 +9,12 @@ use core::time::Duration;
 use std::ops::RangeInclusive;
 use std::sync::Arc;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
 use async_trait::async_trait;
 use futures::Stream;
 use nkeys::{KeyPair, KeyPairType};
 use tokio::io::AsyncRead;
-use tracing::{instrument, trace};
+use tracing::{error, instrument, trace};
 use wasmtime_wasi_http::body::{HyperIncomingBody, HyperOutgoingBody};
 use wrpc_transport::IncomingInputStream;
 
@@ -189,8 +189,8 @@ impl PartialEq for ActorIdentifier {
 impl Eq for ActorIdentifier {}
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-/// Interface target to be invoked over `wRPC`
-pub struct WrpcInterfaceTarget {
+/// Interface target to be invoked over the lattice using `wRPC`
+pub struct LatticeInterfaceTarget {
     /// wRPC component routing identifier
     pub id: String,
     /// wRPC component interface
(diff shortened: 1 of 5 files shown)


vados-cosmonic (approved):
LGTM :rocket:

thomastaylor312 (approved):

crates/runtime/src/capability/builtin.rs
```diff
@@ -211,23 +211,10 @@ impl std::fmt::Display for WrpcInterfaceTarget {
...
-        if let Self::Wrpc(target) = self {
-            Some(target)
-        } else {
-            None
-        }
-    }
+    /// Component to invoke over the lattice using wRPC
+    Lattice(LatticeInterfaceTarget),
+    /// (unimplemented) Component to invoke locally
+    Local(LatticeInterfaceTarget),
```
rvolosatovs, line 217: Is there any work planned regarding supporting "local" invocations and links? If not I'd say we should remove it and, in fact, remove enum altogether and use `LatticeInterfaceTarget` directly

rvolosatovs (approved):

crates/runtime/src/capability/builtin.rs
brooksmtownsend: We do have work planned for the future ™️ but I will remove the unimplemented member and tag it as non_exhaustive so we can add it without breaking things in the future

thomastaylor312 (approved):

crates/runtime/src/capability/builtin.rs
rvolosatovs: If we do indeed implement this feature, I would not expect users to have to explicitly select a "local" target - wasmCloud runtime should be able to perform the optimization transparently from the user's perspective.
In other words I would not expect to see any linking feature changes to enable "local" links - wasmCloud should be able to infer if it's running something on the prefix-instance-function tuple and if so, directly call that component without using NATS - this functionality would hook into `wrpc::Client` implementation and would not require any linking/targeting changes

PR #1613, opened 2024-03-01
[FEAT] Downgrade claims to metadata, formalize policy service

### LINKED ISSUE #1104: Drop deprecated wasmbus.evt.{latticeID} subject
As of https://github.com/wasmCloud/wasmCloud/pull/1090, the host now emits events on `wasmbus.evt.{latticeID}` as well as `wasmbus.evt.{latticeID}.{eventName}`

Prior to 1.0, we should drop the former, as it is deprecated. Clients which wish to continue subscribing to all events can do so with a `wasmbus.evt.{latticeID}.>` subject filter

### LINKED ISSUE #1441: [FEAT] Security examples
These examples should cover a few scenarios to illustrate defense-in-depth security in wasmCloud. In our documentation we have a few key pages talking about the ways we secure wasmCloud applications but we don't have exact code examples for what happens when you try and violate one of these principles.

In order to run these, Docker compose, bash/justfile/README scripts, and documentation is the key here. Diagrams definitely welcome.

1. Launching two wasmCloud hosts that don't share a cluster issuer to show denying invocations
1. Launching a "bad" actor that is not signed
1. Launching a "bad" actor whos hash does not match the module bytes
1. Launching a "bad" actor that tries to use a capability that's not allowed
1. Bonus: Policy service

brooksmtownsend:
## Feature or Problem
This PR:
- [ ] Downgrades embedded claims in actors from required to optional metadata
- [ ] Downgrades embedded claims in providers from required to optional metadata
- [ ] Reshapes the policy service request format to a more maintainable format with explicit policy request kinds and payloads. 
- [ ] Changed a few of the `par` internals to cache providers based on Host ID and provider ID, rather than link name and contract ID.

EDIT: I came back and added an OPA example 😄 

## Related Issues
Fixes #1441
Fixes #1104

## Consumer Impact
Implementations of policy servers will need to adjust their deserialization logic to receive the new requests.

### Manual Verification
Using a CLI policy service (`nats reply 'wasmcloud.policy' '{"permitted": true, "requestId": "abc123"}'`) I was able to see policy requests succeed and get cached. In the case where the operation was not permitted, error logs in the host report:
```
2024-03-01T18:34:56.058886Z ERROR wasmcloud_host::wasmbus: failed to scale actor actor_ref=file:///Users/brooks/github.com/wasmcloud/wasmcloud/tests/actors/rust/wrpc-ponger-component/build/wrpc_ponger_component_s.wasm actor_id=header err=Policy denied request to scale actor `abc123`: `None`
2024-03-01T18:36:18.559730Z ERROR wasmcloud_host::wasmbus: failed to handle invocation err=policy denied request to invoke component `foobar`: `None`
```


crates/host/src/lib.rs
```diff
@@ -130,20 +128,21 @@ pub async fn fetch_actor(
 }
 
 /// Fetch a provider from a reference.
-#[instrument(skip(provider_ref, link_name))]
+#[instrument(skip(registry_config, provider_id, host_id), fields(provider_ref = %provider_ref.as_ref()))]
```
thomastaylor312, line 131: Nit: Might be worth having the provider ID as a field for easier log searching later

crates/host/src/policy.rs
```diff
@@ -10,110 +10,145 @@ use futures::{
...
-    pub cluster_issuers: Vec<String>,
 }
 
 /// The action being requested
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Hash)]
-pub enum Action {
-    /// The host is checking whether it may invoke the target actor
-    #[serde(rename = "perform_invocation")]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Hash)]
+pub enum RequestKind {
```
thomastaylor312, line 102: Because a policy service can be written by anyone, should this be `#[serde(untagged)]` so it has some cleaner JSON?

crates/host/src/policy.rs
```diff
@@ -148,58 +172,30 @@ fn is_expired(expires: u64) -> bool {
...
-impl From<jwt::Claims<jwt::CapabilityProvider>> for RequestSource {
-    fn from(claims: jwt::Claims<jwt::CapabilityProvider>) -> Self {
-        RequestSource {
-            public_key: Some(claims.subject),
-            contract_id: claims.metadata.map(|m| m.capid),
-            link_name: None, // Unfortunately, since claims don't include a link name, we can't populate this
-            capabilities: vec![],
-            issuer: Some(claims.issuer),
-            issued_on: Some(claims.issued_at.to_string()),
+impl From<&jwt::Claims<jwt::CapabilityProvider>> for PolicyClaims {
```
thomastaylor312, line 187: Nit: Probably still want the owned version of these too

crates/host/src/policy.rs
```diff
@@ -260,18 +256,97 @@ impl Manager {
...
+    #[instrument(level = "trace", skip_all)]
+    pub async fn evaluate_perform_invocation(
         &self,
-        source: Option<RequestSource>,
-        target: RequestTarget,
-        action: Action,
+        component_id: impl AsRef<str>,
+        image_ref: impl AsRef<str>,
+        annotations: &BTreeMap<String, String>,
+        claims: &Option<&jwt::Claims<jwt::Actor>>,
```
thomastaylor312, line 306: This (and all the other claims params in these functions) should just be `Option<&jwt::Claims<jwt::Actor>>`. You can take a borrowed option and use `as_ref()` or `as_deref` to get a borrowed inner value

crates/host/src/policy.rs
```diff
@@ -260,18 +256,97 @@ impl Manager {
...
+        let kind = match request {
+            RequestBody::StartComponent(_) => RequestKind::StartComponent,
+            RequestBody::StartProvider(_) => RequestKind::StartProvider,
+            RequestBody::PerformInvocation(_) => RequestKind::PerformInvocation,
+            RequestBody::Unknown => RequestKind::Unknown,
+        };
+        let cache_key = match request {
+            RequestBody::StartComponent(ref req) => RequestKey {
+                kind,
+                request: RequestBody::StartComponent(req.clone()),
```
thomastaylor312, line 337: This is a lot of data to clone on every invocation. Anyway we can make RequestKey more lightweight by borrowing (or using Cows) instead?

crates/host/src/wasmbus/event.rs
```diff
@@ -37,38 +37,57 @@ fn format_actor_claims(claims: &jwt::Claims<jwt::Actor>) -> serde_json::Value {
 }
 
 pub fn actor_scaled(
-    claims: &jwt::Claims<jwt::Actor>,
+    claims: &Option<&jwt::Claims<jwt::Actor>>,
```
thomastaylor312, line 40: Same thing here with the double borrow in these functions

crates/host/src/par.rs
```diff
@@ -44,45 +44,42 @@ fn native_target() -> String {
...
-            .unwrap_or_default(),
-    );
-    let link_name = normalize_for_filename(link_name.as_ref());
+/// Returns the path to the cache file for a provider
+///
+/// # Arguments
+/// * `host_id` - The host ID this provider is starting on. Required in order to isolate provider caches
+///            for different hosts
+/// * `provider_id` - The unique provider identifier
+pub fn cache_path(host_id: impl AsRef<str>, provider_id: impl AsRef<str>) -> PathBuf {
```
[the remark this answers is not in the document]
brooksmtownsend: I think I'll change this to be per-provider reference per-host ID then. I finally remembered why host ID is important, it's to protect looking for possibly securely fetched providers on a multitenant system and then starting them from cache

crates/host/src/policy.rs
brooksmtownsend: I don't know if it needs to be, the request format looks like this:
```json
{
  "requestId": "018dfb36-c639-ab3b-caa3-f7823e8d04c9",
  "kind": "startComponent",
  "version": "v1",
...
}
```

I know untagging the request body prevented a weird nesting, but here since it's just a String rename I think it's okay

crates/host/src/policy.rs
brooksmtownsend: I figured I'd implement this once we actually need it instead of having the impl without the need, if that makes sense. Will add if it helps address remaining PR comments

crates/host/src/policy.rs
brooksmtownsend: Let me see if we can drop it down to more required information

brooksmtownsend: Changed to, instead of cloning the full set of data, synthesizing a cache key based on a few of the strings. I also considered a Cow, or a use of lifetimes to reference a few strings for the cache key, but this is pretty simple and easy to understand. Let me know your thoughts or if I'm missing where `Cow` can make this a bit better

crates/host/src/policy.rs
thomastaylor312: Ok, that should be fine then

crates/host/src/policy.rs
thomastaylor312: Nah you're good

thomastaylor312 (approved):

PR #1614, opened 2024-03-01
test: invoke `wasi:http/incoming-handler` using `hyper`

rvolosatovs:
This moves a bunch of stuff from `builtins-component-reactor` to `pinger` for testing.
Inside the test we create a little Hyper "provider", which listens on HTTP and invokes the `wasi:http/incoming-handler.handle` exported by the component via wRPC.


crates/runtime/src/actor/component/mod.rs
```diff
@@ -270,6 +270,7 @@ where
             | "wasi:io/error@0.2.0"
             | "wasi:io/poll@0.2.0"
             | "wasi:io/streams@0.2.0"
+            | "wasi:logging/logging"
```
brooksmtownsend, line 273: I was wondering what that error was about 🙏🏻

tests/wrpc.rs
```diff
@@ -107,15 +119,131 @@ async fn wrpc() -> anyhow::Result<()> {
...
+            try_join!(
+                async {
+                    info!("await connection");
+                    let (stream, addr) = listener
+                        .accept()
+                        .await
+                        .context("failed to accept connection")?;
+                    info!("accepted connection from {addr}");
+                    let wrpc_client = Arc::clone(&wrpc_client);
+                    hyper_util::server::conn::auto::Builder::new(TokioExecutor::new())
```
brooksmtownsend, line 137: Good inspiration for how the provider will work 👍🏻

brooksmtownsend (approved):
This is awesome. Thank you for bringing a huge part of the builtins-component-reactor over to this test so we can validate that from now on

brooksmtownsend (approved):

PR #1615, opened 2024-03-01
feat: improve byte stream efficiency

rvolosatovs:
See https://github.com/rvolosatovs/wrpc/commit/7fd68ab80bf1f962f8cae9637ed70a527dff0ed2

### DIFF
--- a/tests/wrpc.rs
+++ b/tests/wrpc.rs
@@ -1,4 +1,4 @@
-use core::str::FromStr as _;
+use core::str::{self, FromStr as _};
 use core::time::Duration;
 
 use std::net::Ipv4Addr;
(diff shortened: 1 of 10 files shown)


tests/wrpc.rs
```diff
@@ -253,7 +253,7 @@ async fn assert_incoming_http(
             ensure!(config_value.is_none());
             ensure!(all_config == []);
             ensure!(ping == "pong");
-            ensure!(long_value == "1234567890".repeat(1000));
+            ensure!(long_value == "1234567890".repeat(5000));
```
rvolosatovs, line 256: I think we can go much further, but don't want to make the tests too slow here

brooksmtownsend (approved):

PR #1616, opened 2024-03-01
feat(runtime): Switches runtime config to use a pooling allocator

thomastaylor312:
This leads to significant performance increases on Linux and brings it up to the same performance levels as other operating systems. Please note that I explicitly did not expose any of these knobs directly in the host yet as it is really finnicky to tweak properly without requesting too much memory. I'd really like to sit down with some of the wasmtime folks before exposing this to users and see if we can dynamically figure out what these numbers should be. For now, these seem to be fairly sensible defaults

### DIFF
--- a/crates/runtime/src/runtime.rs
+++ b/crates/runtime/src/runtime.rs
@@ -10,11 +10,18 @@ use core::fmt::Debug;
 use std::sync::Arc;
 
 use anyhow::Context;
+use wasmtime::{InstanceAllocationStrategy, PoolingAllocationConfig};
+
+const KB: u64 = 1024;
+const MB: u64 = KB * 1024;
+const GB: u64 = MB * 1024;
 
 /// [`RuntimeBuilder`] used to configure and build a [Runtime]
 #[derive(Clone, Default)]
 pub struct RuntimeBuilder {
     engine_config: wasmtime::Config,
+    max_components: u32,
+    max_component_size: u64,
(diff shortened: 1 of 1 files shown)


crates/runtime/src/runtime.rs
```diff
@@ -26,8 +33,12 @@ impl RuntimeBuilder {
         let mut engine_config = wasmtime::Config::default();
         engine_config.async_support(true);
         engine_config.wasm_component_model(true);
+        engine_config.memory_init_cow(false);
+
         Self {
             engine_config,
+            max_components: 10000,
+            max_component_size: 15 * MB,
```
brooksmtownsend, line 41: I think we might want to make this larger, the out-of-the-box hello world component for Python is 35mb
```bash
➜ ll build
Permissions Size User   Date Modified Name
.rw-r--r--     0 brooks  4 Mar 10:14  .gitkeep
.rw-r--r--   35M brooks  4 Mar 10:30  http_hello_world.wasm
.rw-r--r--   35M brooks  4 Mar 10:30  http_hello_world_s.wasm
```

I'm not sure what would balloon this up to a larger number, and since our TS/JS component is only 9mb Python is the only one that exceeds this. Thoughts on at least upping this to 40MB, or maybe 64MB for a computer number?

brooksmtownsend (changes requested):
Only request is to accommodate large components like Python

PR #1623, opened 2024-03-05
refactor(core,control-interface): InterfaceLinkDefinition -> core

vados-cosmonic:
## Feature or Problem

This commit refators the types defined in both `wasmcloud-core` and `wasmcloud-control-interface` to make it easier to distinguish what types belong where and what they're related to.

Ultimately the goal here was was to move `InterfaceLinkDefinition` into `wasmcloud-core` so it can be used in other places, but it was a good chance to reorganize.


crates/core/src/lattice.rs
```diff
@@ -0,0 +1,282 @@
...
+
+use crate::otel::TraceContext;
+use crate::wit::{deserialize_wit_map, serialize_wit_map};
+
+/// Key of an issuer (nkey)
+pub type ClusterIssuerKey = String;
+
+/// List of (allowed) cluster issuers
+pub type ClusterIssuers = Vec<ClusterIssuerKey>;
+
```
brooksmtownsend, line 22: After comments in https://github.com/wasmCloud/wasmCloud/pull/1575, we generally had consensus that this type of type aliasing was a higher burden to mentally parse, like when the alias is just for a String or just for a Vec / HashMap. What are your thoughts here?

vados-cosmonic: I think this kind of aliasing is definitely unnecessary -- `Vec<ClusterIssuerKey>` is definitely fine with me -- I didn't add this particular one.

vados-cosmonic: removed!

brooksmtownsend (approved):

PR #1625, opened 2024-03-06
fix(host): re-tag request type enum policy

brooksmtownsend:
## Feature or Problem
This PR re-tags the enum for `RequestType` since omitting it actually sets that field to `null`, breaking the policy service.


vados-cosmonic (approved):
LGTM :rocket:

Do we have a policy service test? I can write one up real quick if we don't.

rvolosatovs (approved):

brooksmtownsend:
> LGTM :rocket:
> 
> 
> 
> Do we have a policy service test? I can write one up real quick if we don't.

I don't think so right now, maybe commented out? I'd have to look, but a test would be great! My apologies for not putting it in the OG pr

PR #1626, opened 2024-03-07
refactor: move wasmcloud wrpc transport client to core

vados-cosmonic:
## Feature or Problem

This commit moves the wasmcloud-specific wrpc transport client to the `wasmcloud-control-interface` crate. From here, it can be used by both the host (`wasmbus`) and other places like tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5403,6 +5403,7 @@ version = "0.33.0"
 dependencies = [
  "anyhow",
  "async-nats",
+ "async-trait",
  "bytes",
  "cloudevents-sdk",
  "futures",
@@ -5443,6 +5444,8 @@ version = "0.2.0"
 dependencies = [
  "anyhow",
  "async-nats",
+ "async-trait",
+ "bytes",
  "futures",
  "hex",
  "nkeys",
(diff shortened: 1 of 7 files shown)


brooksmtownsend:
I’m a little concerned with the refactor for two reasons 
blends the responsibility of the control interface crate and ability to interact over rpcconfusing to have different clients, especially if the host uses different connections for RPC and CTL
I do know that this needs to be available to the test util and wash and others, though, and I’m hesitant to create more crates to deal with. Curious to hear your thoughts here and I’ll try to think of a good alternative too

rvolosatovs:
> I’m a little concerned with the refactor for two reasons 
> - blends the responsibility of the control interface crate and ability to interact over rpc
> - confusing to have different clients, especially if the host uses different connections for RPC and CTL
> I do know that this needs to be available to the test util and wash and others, though, and I’m hesitant to create more crates to deal with. 
> 
> Curious to hear your thoughts here and I’ll try to think of a good alternative too 

How about moving it to `wasmcloud-core`? It seems like it's shared functionaity that the host, providers and arbitrary wasmCloud plugins would use

vados-cosmonic:
Moved the client to `core`

brooksmtownsend (approved):
Thanks @vados-cosmonic !

PR #1627, opened 2024-03-07
refactor(providers): get values from new link def constistently

vados-cosmonic:
## Feature or Problem

This commit updates all providers that were marked with TODOs related to using named config to use `InterfaceLinkDefinition.target_config` temporarily.

The idea is to stuff configuration into the "names"s of configs that were supposed to be in use (i.e. `["NAME=VALUE", "NAME=VALUE"]` rather than `["config-1", "config-2"]`), ahead of named config being ready.

### DIFF
--- a/crates/core/src/link.rs
+++ b/crates/core/src/link.rs
@@ -1,5 +1,7 @@
 //! Core reusable types related to links on wasmCloud lattices
 
+use std::collections::HashMap;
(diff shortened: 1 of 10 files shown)


brooksmtownsend (approved):
I'm a big fan of this temporary measure 😌

Issue #1624, opened 2024-03-05
[BUG] `wash build` always generates keys in the local folder instead of `$HOME/.wash/keys


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

I've noticed recently `wash` stopped generating and sourcing keys from `$HOME/.wash/keys`, and instead opts to generate keys local to the component folder. I could see this being accepted behavior for the module key itself, but for my account key I don't want to regenerate that for each project.

### Steps to reproduce

```bash
wash new actor --template-name hello-world-rust hello
cd hello
wash build
ll ./keys
```

### Expected behavior

I expect for new keys to be generated and stored at `$HOME/.wash/keys` instead of relative in the local actor folder unless I specify to do that explicitly in `wasmcloud.toml`. 

### Environment

- OS: MacOS
- Shell: zsh
- wasmcloud Version: 0.82.0
- wash Version: 0.26.0

brooksmtownsend:
For folks looking to take this on, see the TODOs added in #1628 for where we should be defaulting to `~/.wash/keys` 😄

siddharthkhonde:
hi @brooksmtownsend , is this issue still open? I am planning to take this up as my first issue

brooksmtownsend:
Closed by #1837

PR #1628, opened 2024-03-07
feat(wash)!: implement wash building provider for host machine

brooksmtownsend:
## Feature or Problem
This PR implements wash building a capability provider archive just for the local arch/os target. In the future I'd love to add support for multiple other architectures as well with predefined paths, but `wash build` won't ever be responsible for cross compiling (something something sure would be nice if we had a platform agnostic binary target to compile to)

## Related Issues
I added two TODOs re: #1624 because I found out why we're generating keys to `./keys`

## Release Information
next minor wash-cli
next minor wash-lib

### Manual Verification
I verified this worked for a provider that will soon be an example 😄

### DIFF
--- a/crates/wash-lib/src/build/component.rs
+++ b/crates/wash-lib/src/build/component.rs
@@ -1,5 +1,3 @@
-//! Build (and sign) a wasmCloud actor, provider, or interface. Depends on the "cli" feature
-
 use std::{
     borrow::Cow,
     fs,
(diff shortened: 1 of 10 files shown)


crates/wash-lib/src/build/mod.rs
```diff
@@ -0,0 +1,88 @@
+//! Build (and sign) a wasmCloud actor, or provider. Depends on the "cli" feature
```
vados-cosmonic, line 1: do we want to make this real?

```
#![cfg(feature = "cli")]
```

crates/wash-lib/src/build/provider.rs
```diff
@@ -0,0 +1,132 @@
...
+    common_config: &CommonConfig,
+    signing_config: Option<&SignConfig>,
+) -> Result<PathBuf> {
+    let LanguageConfig::Rust(rust_config) = language_config else {
+        bail!("Unsupported language for provider: {:?}", language_config)
+    };
+    let mut command = match rust_config.cargo_path.as_ref() {
+        Some(path) => process::Command::new(path),
+        None => process::Command::new("cargo"),
+    };
```
vados-cosmonic, line 28:
```suggestion
    let mut command = process::Command::new(rust_config.cargo_path.as_ref().unwrap_or("cargo"));
```

crates/wash-lib/src/build/provider.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    // Change directory into the project directory
+    std::env::set_current_dir(&common_config.path)?;
+
+    // Build for a specified target if provided, or the default rust target
+    let build_args = if let Some(override_target) = &provider_config.target {
+        vec!["build", "--release", "--target", override_target]
+    } else {
+        vec!["build", "--release"]
```
vados-cosmonic, line 37: could probably factor this out and make it mutable to add target if `override_target` is present

crates/wash-lib/src/build/provider.rs
```diff
@@ -0,0 +1,132 @@
...
+            .iter()
+            .find_map(|p| {
+                p.targets.iter().find_map(|t| {
+                    if t.kind.iter().any(|k| k == "bin") {
+                        Some(t.name.clone())
+                    } else {
+                        None
+                    }
+                })
+            }).context("Could not infer provider binary name in metadata, please specify under provider.bin_name")?
```
vados-cosmonic, line 67: NIT: feels like we should factor this out

crates/wash-lib/src/build/provider.rs
```diff
@@ -0,0 +1,132 @@
...
+            capid: provider_config.package.clone().unwrap_or_default(),
+            vendor: provider_config.vendor.to_string(),
+            revision: Some(common_config.revision),
+            version: Some(common_config.version.to_string()),
+            schema: None,
+            name: common_config.name.to_string(),
+            arch: detect_arch(),
+        },
+        &provider_bytes,
+    )?;
```
vados-cosmonic, line 92: `.context` would be nice here -- IIRC `create_provider_archive` doesn't have very descriptive errors

crates/wash-lib/src/parser/mod.rs
```diff
@@ -133,58 +133,59 @@ impl TryFrom<RawActorConfig> for ActorConfig {
...
 
 #[derive(Deserialize, Debug, PartialEq)]
 struct RawProviderConfig {
-    /// The capability ID of the provider.
-    pub capability_id: String,
-    /// The vendor name of the provider. Optional, defaults to 'NoVendor'.
+    /// The vendor name of the provider.
     pub vendor: Option<String>,
+    /// Optional WIT world package for the provider
+    pub package: Option<String>,
```
vados-cosmonic, line 157:
```suggestion
    pub wit_package: Option<String>,
```

crates/wash-lib/src/parser/mod.rs
```diff
@@ -133,58 +133,59 @@ impl TryFrom<RawActorConfig> for ActorConfig {
...
+    /// The target operating system of the provider archive. Defaults to the current OS.
+    pub os: Option<String>,
+    /// The target architecture of the provider archive. Defaults to the current architecture.
+    pub arch: Option<String>,
+    /// The Rust target triple to build for. Defaults to the default rust toolchain.
+    pub target: Option<String>,
```
vados-cosmonic, line 163:
```suggestion
    pub rust_target: Option<String>,
```

crates/wash-lib/src/parser/mod.rs
vados-cosmonic: is this really only the package? not the namespace + package?

vados-cosmonic (approved):
Nothin but nits, LGTM :rocket:

crates/wash-lib/src/build/component.rs
```diff
@@ -97,7 +35,7 @@ pub fn build_actor(
     actor_config: &ActorConfig,
     language_config: &LanguageConfig,
     common_config: &CommonConfig,
-    signing_config: Option<SignConfig>,
+    signing_config: Option<&SignConfig>,
```
thomastaylor312, line 38: What was the reason behind the changes for borrowing `SignConfig` in the function signatures now? Doesn't really matter since this is the CLI, but I was curious

crates/wash-lib/src/build/provider.rs
thomastaylor312: I also endorse this nit

crates/wash-lib/src/build/provider.rs
```diff
@@ -0,0 +1,132 @@
...
+    let result = command.args(build_args).status().map_err(|e| {
+        if e.kind() == ErrorKind::NotFound {
+            anyhow!("{:?} command is not found", command.get_program())
+        } else {
+            anyhow!(e)
+        }
+    })?;
+
+    if !result.success() {
+        bail!("Compiling provider failed: {}", result.to_string())
```
thomastaylor312, line 48: Super nit: I don't think you need to `to_string()` it here as it should implement `Display`

thomastaylor312 (approved):
Also good here, mostly nits and a question around why there is some more borrowing

crates/wash-lib/src/build/mod.rs
brooksmtownsend: It's real! We just gate it in `lib.rs` 😄

crates/wash-lib/src/build/provider.rs
brooksmtownsend: Oh we have it like this because `cargo_path` is a PathBuf but "cargo" is a &str. Will see if I can make type checker happy

crates/wash-lib/src/build/provider.rs
brooksmtownsend:
```
    let mut build_args = Vec::with_capacity(4);
    build_args.extend_from_slice(&["build", "--release"]);
    if let Some(override_target) = &provider_config.target {
        build_args.extend_from_slice(&["--target", override_target]);
    };
```

I think it looks a little weirder but I don't mind, will be better if we need more build args down the line

crates/wash-lib/src/parser/mod.rs
brooksmtownsend: Yeah it's both, not like we use it anyways but it's just a standin for capability contract ID

crates/wash-lib/src/build/provider.rs
brooksmtownsend: I cannot without making it uglier 😄

crates/wash-lib/src/build/component.rs
brooksmtownsend: Primarily for consistency, this is in the library so I figured that depending on the usage it would be useful to be able to pass a reference instead of an owned config, just like the other functions

crates/wash-lib/src/build/provider.rs
brooksmtownsend: I went back and forth on this and figured it was acceptable for now to just leave as-is, let me know if you feel strongly I should factor it out

vados-cosmonic: definitely don't feel strongly on it! LGM

vados-cosmonic (approved):
LGTM :rocket:

Finally, wash builds providers!

PR #1630, opened 2024-03-07
feat(wash-cli): basic wasi:cli/run style wrpc invocation

brooksmtownsend:
Draft PR as it branches off of #1628 

## Feature or Problem
This PR implements a very basic `wash call` functionality that allows developers to invoke a component that exports an interface function that takes no arguments and returns a String. It's very basic, but it will allow manual invocation of a component over wRPC and we can extend this in the future by passing in WIT files 😄 

For example, a component WIT world like:
```wit
package wasmcloud:hello;

interface easy {
  call: func() -> string;
}
```

Can be invoked like:
```bash
wash call <component_id> wasmcloud:hello/easy.call
```

## Release Information
next wash-cli

### Manual Verification
This works great! I attempted to include basic error handling around timeouts and no responders errors.

### DIFF
--- a/crates/core/src/wrpc.rs
+++ b/crates/core/src/wrpc.rs
@@ -132,7 +132,7 @@ impl Client {
     ///
     /// ## Arguments
     /// * `nats` - The NATS client to use for communication.
-    /// * `lattice` - The lattice to use for communication.
+    /// * `prefix` - The lattice and component ID to use for communication, e.g. "default.echo".
     /// * `headers` - The headers to include with each outbound invocation.
     pub fn new(
         nats: impl Into<Arc<async_nats::Client>>,
--- a/crates/core/src/wrpc.rs
+++ b/crates/core/src/wrpc.rs
@@ -128,19 +128,24 @@ pub struct Client {
 }
 
 impl Client {
-    /// Create a new wRPC [Client] with the given NATS client, lattice, and headers.
+    /// Create a new wRPC [Client] with the given NATS client, lattice, component ID, and headers.
     ///
     /// ## Arguments
     /// * `nats` - The NATS client to use for communication.
-    /// * `prefix` - The lattice and component ID to use for communication, e.g. "default.echo".
+    /// * `lattice` - The lattice ID to use for communication.
+    /// * `prefix` - The component ID to use for communication.
     /// * `headers` - The headers to include with each outbound invocation.
     pub fn new(
         nats: impl Into<Arc<async_nats::Client>>,
-        lattice: String,
+        lattice: impl AsRef<str>,
+        component_id: impl AsRef<str>,
         headers: HeaderMap,
     ) -> Self {
         Self {
-            inner: wrpc_transport_nats::Client::new(nats, lattice),
+            inner: wrpc_transport_nats::Client::new(
+                nats,
+                format!("{}.{}", lattice.as_ref(), component_id.as_ref()),
+            ),
             headers,
         }
     }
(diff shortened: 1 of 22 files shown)


crates/core/src/wrpc.rs
```diff
@@ -132,7 +132,7 @@ impl Client {
     ///
     /// ## Arguments
     /// * `nats` - The NATS client to use for communication.
-    /// * `lattice` - The lattice to use for communication.
+    /// * `prefix` - The lattice and component ID to use for communication, e.g. "default.echo".
```
vados-cosmonic, line 135: can we make these separate args? let's not force people to concat strings before they call this.

crates/wash-cli/src/call.rs
```diff
@@ -166,100 +173,63 @@ pub struct CallCommand {
...
-        "you can use either -d/--data or the payload args, but not both."
-    );
-    ensure!(!actor_id.is_empty(), "actor ID may not be empty");
-
-    let payload = if let Some(fname) = data {
-        std::fs::read_to_string(fname)?
-    } else {
-        payload.join("")
+) -> Result<String> {
+    ensure!(!component_id.is_empty(), "component ID may not be empty");
```
vados-cosmonic, line 193: we can set this as required via `clap` right?

crates/wash-cli/src/call.rs
```diff
@@ -166,100 +173,63 @@ pub struct CallCommand {
...
-    /// Operation to invoke on actor
-    #[clap(name = "operation")]
-    pub operation: String,
-
-    /// Payload to send with operation (in the form of '{"field": "value"}' )
-    #[clap(name = "payload")]
-    pub payload: Vec<String>,
+    /// Fully qualified function to invoke on the actor, e.g. `wasi:cli/run.run`
+    #[clap(name = "function")]
+    pub function: String,
```
vados-cosmonic, line 182: Maybe we should call this a "fully qualified WIT export", rather than function?

crates/wash-cli/src/call.rs
```diff
@@ -166,100 +173,63 @@ pub struct CallCommand {
...
+        opts.rpc_credsfile,
+    )
+    .await?;
+    let mut headers = async_nats::HeaderMap::new();
+    headers.insert("source-id", "wash");
+    let lattice = opts.lattice.unwrap_or_else(|| DEFAULT_LATTICE.to_string());
+    let wrpc_client =
+        wasmcloud_core::wrpc::Client::new(nc, format!("{lattice}.{component_id}"), headers);
+
+    let Some((instance, name)) = function.rsplit_once('.') else {
```
vados-cosmonic, line 210: Don't we have a function that deals in this? I think this is precisely where we can use that interface target "function" param.

thomastaylor312 (commented):
This looks much cleaner now

crates/wash-cli/src/call.rs
brooksmtownsend: We do have it as required, but that doesn't stop someone from supplying an empty string:
```bash
➜ cargo run -p wash-cli -- call "" "wasi:cli/run.run"
    Finished dev [unoptimized + debuginfo] target(s) in 0.44s
     Running `target/debug/wash call '' 'wasi:cli/run.run'`

component ID may not be empty
```

crates/core/src/wrpc.rs
brooksmtownsend: Yes I like this idea, since that's the way we're going to use it always

crates/wash-cli/src/call.rs
vados-cosmonic: Ah so it looks like what you can do is `#[value_parser = "clap::builder::NonEmptyStringValueParser" ]` on the argument and it will check that for you... 

Probably not worth though so resolving this

PR #1631, opened 2024-03-07
chore: prefix NATS_PORT and NATS_URL with WASMCLOUD

protochron:
## Feature or Problem
This adds a `WASMCLOUD` prefix to both the `NATS_PORT` and `NATS_URL` env vars, since there are potentially conficts with those variables in other systems. Specifically this would play badly deploying in Kubernetes with a service named `nats` running in the same namespace, since a `NATS_PORT` env var is automatically injected in that case which stops the wasmcloud binary from starting. It is also a good practice to have all of our environment variables using the same prefix.

This is a breaking change to wasmCloud CLI args for the host.


brooksmtownsend (changes requested):
Pedantic request to just amend the commit message to show that it's a breaking change please 😌

brooksmtownsend (approved):

PR #1633, opened 2024-03-08
fix(tests): add messaging triggered test actor

vados-cosmonic:
## Feature or Problem

This commit is the culmination of a few things that were required for getting our flavor of E2E tests (in the top level `tests/` dir) working for a Provider & Actor.

This commit is quite large because it does many things:

- Adds missing implementation to bindgen for provider -> actor invocations
- Uncomments implementation from the host for wasmcloud:messaging
- Adds an invoker component that reacts to messaging rather than HTTP
- Uses messaging & keyvalue providers plus the actor in a single test

With this, we have an easy to understand way to test every provider that we have in the repository.


crates/provider-wit-bindgen/src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub mod deps {
     pub use serde;
     pub use serde_bytes;
     pub use serde_json;
+    pub use tracing;
```
vados-cosmonic, line 9: I backed out the code that's using this, but I'm going to come back to it -- bindgen needs to have tracing all over it (OTEL too!)

crates/runtime/src/actor/component/mod.rs
```diff
@@ -273,9 +273,7 @@ where
             | "wasi:logging/logging"
             | "wasi:sockets/tcp@0.2.0"
             | "wasmcloud:bus/lattice"
-            | "wasmcloud:bus/guest-config"
-            | "wasmcloud:messaging/messaging"
-            | "wasmcloud:messaging/message-subscriber" => continue,
```
brooksmtownsend, line 278: With this, we can remove the trait impl for `Messaging` for `Handler` in wasmbus I think

tests/blobstore-s3.rs
brooksmtownsend, line 1: Are we removing these test files just to clear out the old so we can usher in new integration tests?

crates/provider-sdk/src/provider.rs
```diff
@@ -153,6 +153,20 @@ impl ProviderConnection {
         self.rpc_client.clone()
     }
 
+    /// Used for fetching the RPC client in order to make RPC calls
+    pub fn get_wrpc_client(&self, target: impl AsRef<str>) -> WrpcNatsClient {
```
brooksmtownsend, line 157: Once we get to working on tracing (as you mentioned) this function should  take trace context into account to set up headers

brooksmtownsend (approved):
All comments for the future and a nit, let's consider if any of these should be tracked as new issues outside of the wRPC MVP issue

crates/provider-wit-bindgen/src/lib.rs
brooksmtownsend: I left a comment below just to point out where I see this need as well but I assume you're on it too 😄

tests/blobstore-s3.rs
vados-cosmonic: Yep, that's the idea!

crates/provider-sdk/src/provider.rs
vados-cosmonic: Ah good point -- will make sure to add this

vados-cosmonic: made a note on the integeration issue!

crates/runtime/src/actor/component/mod.rs
vados-cosmonic: Yeah it's on clear why these were left out, because the custom trait impl wasn't working either (and it couldn't even be targeted because of this being left out). Will investigate removing the trait impl!

crates/provider-sdk/src/provider.rs
```diff
@@ -153,6 +153,20 @@ impl ProviderConnection {
...
+        let target = target.as_ref();
+        let mut headers = HeaderMap::new();
+        headers.insert("lattice-id", self.lattice.as_ref());
```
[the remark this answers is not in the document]
vados-cosmonic: removed this!

crates/provider-wit-bindgen/src/lib.rs
vados-cosmonic: Yup, thanks! definitely need to pass that trace context around

PR #1634, opened 2024-03-08
FEAT: Local component invocation routing

brooksmtownsend:
## Feature or Problem
This PR isn't meant to be merged (yet.) It's meant to show how simple it is to enable local component routing once we can iron out a few questions with that feature, like
1. How do we enable this to respect the max instances concurrency limit
2. Is this worth doing if using it in a leaf node deployment
3. How can we do this for Providers too
4. Should this be on by default or feature toggled

Likely going to close this PR but, in case folks were curious


vados-cosmonic:
heads up @brooksmtownsend I'm not sure this was supposed to get closed automatically.

PR #1640, opened 2024-03-10
feat: Standardize flags for enabling and configuring observability 

joonas:
## Feature or Problem

This standardizes the flags used for enabling and configuring observability, with this change observability can be:
* Enabled either for all signals at once, or selectively enabled/disable on a per-signal basis.
* Configured to emit all signals to a single endpoint or override the individual signal endpoints.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -215,28 +215,41 @@ struct Args {
     #[clap(long = "enable-observability", env = "WASMCLOUD_OBSERVABILITY_ENABLED")]
     enable_observability: bool,
 
-    /// Determines whether tracing should be enabled.
-    #[clap(long = "enable-tracing", env = "WASMCLOUD_TRACING_ENABLED")]
+    /// Determines whether tracing should be enabled, overriding any value set with `--enable-observability`.
+    #[clap(
+        long = "enable-tracing",
+        env = "WASMCLOUD_TRACING_ENABLED",
+        hide = true
+    )]
     enable_tracing: Option<bool>,
 
-    /// Determines whether metrics should be enabled.
-    #[clap(long = "enable-metrics", env = "WASMCLOUD_METRICS_ENABLED")]
+    /// Determines whether metrics should be enabled, overriding any value set with `--enable-observability`.
+    #[clap(
+        long = "enable-metrics",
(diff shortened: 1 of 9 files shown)


src/main.rs
```diff
@@ -211,16 +211,45 @@ struct Args {
...
+        hide = true
     )]
-    otel_exporter_otlp_endpoint: Option<String>,
+    enable_metrics: Option<bool>,
+
+    /// Determines whether logs should be enabled, overriding any value set with `--enable-observability`.
+    #[clap(long = "enable-logs", env = "WASMCLOUD_LOGS_ENABLED", hide = true)]
+    enable_logs: Option<bool>,
+
+    /// Overrides the OpenTelemetry endpoint used for emitting traces, metrics and logs. This can also be set with `OTEL_EXPORTER_OTLP_ENDPOINT`.
```
brooksmtownsend, line 238: 👨🏻‍🍳 💋

brooksmtownsend (changes requested):
Just one requested change to catch folks from making weird mistakes with flags at runtime 😄

src/main.rs
```diff
@@ -211,16 +211,45 @@ struct Args {
     )]
     oci_password: Option<String>,
 
-    /// Specifies which exporter to use for traces. Only "otlp" is supported at this time
-    #[clap(long = "otel-traces-exporter", env = "OTEL_TRACES_EXPORTER")]
-    otel_traces_exporter: Option<String>,
+    /// Determines whether observability should be enabled.
+    #[clap(long = "enable-observability", env = "WASMCLOUD_OBSERVABILITY_ENABLED")]
+    enable_observability: bool,
```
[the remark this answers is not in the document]
joonas: Addressed this in 2f9252cd9d9089b8a22ce6001da0701fb7922f9b

src/main.rs
joonas: As much as I'd love to keep this comment around, it looks like I need to resolve it to appease the GitHub machinery 😆

joonas:
@brooksmtownsend I think we should be all set here, I've addressed the feedback and rebased this following #1636

brooksmtownsend (approved):

thomastaylor312 (approved):

Issue #1644, opened 2024-03-11
[FEATURE] Update `wash call` integration test to use a published wRPC component


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

I had to disable the `integration_call` test because we don't have a published component.

### Describe the solution you'd like

I would like the `wash-cli` `integration_call` test to be commented back in and work with a published component that we can call.

PR #1645, opened 2024-03-11
fix(wash-cli): disable wash call integration test

brooksmtownsend:
## Feature or Problem
This PR disables the call integration test as we don't have a published component to use with it at this time. I definitely plan to add this back in shortly, just when we publish an example to use with it 😄 

## Related Issues
https://github.com/wasmCloud/wasmCloud/issues/1644


vados-cosmonic (approved):
LGTM :see_no_evil: :rocket:

PR #1647, opened 2024-03-11
feat(control-interface)!: support static named config for providers

brooksmtownsend:
## Feature or Problem
This PR implements the change to the start provider logic in the control interface and host to, instead of accepting a single config string, accept a list of named configurations in order to build the provider configuration. At runtime now we will pass a `HashMap<String, String>` to the provider which will be its configuration (built from all of the named configs).

In a followup issue, which I'll create shortly, we should support reloading this config.

## Related Issues
https://github.com/wasmCloud/wasmCloud/issues/1648

### Manual Verification
I verified this worked by creating configuration and having NATS communicate on a different URI than supplied 😄


vados-cosmonic (approved):
LGTM :rocket:

rvolosatovs (approved):

thomastaylor312 (approved):

Issue #1594, opened 2024-02-28
[RFC] `wash app dev`


lachieh:
### RFC Title

RFC: Introduce `dev` command to wash app management

### Introduction

This RFC proposes that a new `dev` command be added to the `wash app` subcommand which will make iterating on components as part of applications easier.

### Motivation

As we gain more experience building applications with `wadm`, the wasmCloud team and a number of community members have experienced some friction relating to the the dev cycle experience of writing, iterating and deploying components.

### Detailed Design

- `wash app dev` will:
  - look for `*.wadm.yaml`[^1] in the current directory and load it if there is only one present. If there are multiple, it will list all applications.
  - for each `component` in the loaded manifest, it will look for a `build` key[^2] which should be a path to a `wasmcloud.toml` file. If one exists, it will attempt to run the build command inside the `wasmcloud.toml`.
    - once all components are built, it will put and deploy the application through `wadm`
    - start watching the parent folders of each `wasmcloud.toml` file found for changes.
    - when a change is detected in a component folder, run the build command, then stop all the components with the same ID so that wadm will reload them from disk
  - for each component that does not have a `build` key, start watching the file in the `image` path
    - when a change is detected in a component folder, stop all the components with the same ID so that wadm will reload them from disk
  
[^1]: adding the `wadm` suffix before the filetype helps with identifying the files, but also allows for dev tooling and IDEs to assume schema.
[^2]: This is a new field in the the wadm oam schema. Inspiration from the [Docker Compose Build](https://docs.docker.com/compose/compose-file/build) configuration.

### Backwards Compatibility

New command. Shouldn't need too much consideration in backwards compatibility.

### Unresolved Questions

- Is there a problem adding a new key to the oam spec?
- Are there safeguards to put in place around production workloads?

### Conclusion

Adding `wash app dev` will make for a really smooth developer experience.

brooksmtownsend:
@lachieh thank you for formalizing this, I think we need exactly this for the developer iteration loop for wasmCloud apps. The inspiration from `docker compose` makes a lot of sense for how wash can work here.

> Is there a problem adding a new key to the oam spec?
Nope! OAM is extensible, and we define our own information for those components. We make the most use of the `Application` specification, and information about where we would add `build:` is at https://github.com/oam-dev/spec/blob/master/7.application.md#properties.

> Are there safeguards to put in place around production workloads?
[...]

PR #1649, opened 2024-03-11
feat(wash)!: update wash to 1.0 control interface

brooksmtownsend:
This PR updates wash to the latest control interface, including changes in `wash-lib`, the `wash-cli` code and CLI, and the unit/integration tests.

As I work through validating the commands work well, I renamed `actor` to `component` where I can 😄 for every command, with backwards compatible `actor` aliases where possible.

## Feature or Problem
This PR updates `wash-lib` and `wash-cli` to update to the latest control interface client that's present in this repository but not-yet published to `crates.io`. I expect for tests to fail as `wash` integration tests work by using a published version of the wasmCloud host.

This PR grew in scope to accommodate the above changes, and I've attempted to consolidate commits where possible. In addition to updating the control interface, this PR:
- Renames `actor` to `component` where possible
- Removes deprecated commands like `reg push` and `ctl start`
- Updates `spy` and `capture` to work with wrpc invocations
- Fixes a bug with putting new links with differing interfaces
- Fixes a bug with component update where we would deadlock 100% of the time
- Changes component update to be a "fast ack" operation as it involves OCI downloads.

## Related Issues
~It would be nice for #1636 to merge before this comes out of draft so we can validate tests work well.~

### Unit Test(s)
CLI tests were modified as necessary to either remove outdated commands or update with the latest command line arguments.

### Acceptance or Integration
I disabled two integration tests `get_claims` and `update_actor` in wash for now. Once we cut the next alpha version of wasmcloud these can be reenabled (I verified both work with the unreleased version in this PR).

Additionally, integration tests now depend on fetching a component from my GHCR, because we don't have a published one yet 😬 

### Manual Verification
I made sure that each command at the very least works with the 1.0 alpha 😄 We've got many planned improvements to dev in #1594 and we need to add a config command, but I'll do that as a followup to this PR.


brooksmtownsend:
All the integration tests pass locally for me so I'll have to take another look at what's going on

crates/wash-cli/src/ctl/mod.rs
```diff
@@ -64,83 +53,15 @@ pub enum CtlGetCommand {
     Claims(GetClaimsCommand),
 }
 
-pub async fn handle_command(
```
thomastaylor312, line 67: tenor-78384544

crates/wash-cli/src/dev.rs
```diff
@@ -270,36 +283,26 @@ pub async fn handle_command(
...
-            skip_wait: false,
-            timeout_ms: None,
-        })
-        .await?
-        .actor_id
-        .ok_or_else(|| anyhow!("failed to do thing"))?;
-    }
+    // Since the only restriction on actor_id is that it must be unique, we can just use the artifact path as the actor_id
+    // to ensure uniqueness
+    let actor_id = artifact_path.display().to_string();
```
thomastaylor312, line 288: Doesn't the actor_id need to be NATS subject compliant?

crates/wash-lib/src/cli/capture.rs
```diff
@@ -51,17 +50,10 @@ pub enum CaptureSubcommand {
...
-
-    /// A provider ID to filter captured invocations by. This will filter anywhere the provider is
-    /// the source or the target of the invocation. If provided with an actor ID, it will filter
-    /// down to interactions only between the actor and provider
-    #[clap(name = "provider_id", long = "provider-id", value_parser)]
-    pub provider_id: Option<ServiceId>,
+    /// A component ID to filter captured invocations by. This will filter anywhere the component is the
+    /// source or the target of the invocation.
+    #[clap(name = "component_id", long = "component-id", value_parser)]
+    pub component_id: Option<String>,
```
thomastaylor312, line 56: Should we have source and target ID so you can filter for either direction? I know the destination would be a little harder as you'd have to parse from the topic, but it doesn't feel too bad

crates/wash-lib/src/cli/capture.rs
```diff
@@ -76,36 +68,42 @@ pub async fn handle_replay_command(cmd: CaptureReplayCommand) -> Result<CommandO
...
     let filtered = capture.messages.into_iter().filter_map(|msg| {
-        let mut inv: Invocation = rmp_serde::from_slice(&msg.payload).ok()?;
-
-        if let Some(actor_id) = &cmd.actor_id {
-            if (inv.origin.is_actor() && inv.origin.public_key != actor_id.as_ref())
-                || (inv.target.is_actor() && inv.target.public_key != actor_id.as_ref())
-            {
-                return None;
-            }
+        // lattice.component.wrpc.0.0.1.operation.function
```
thomastaylor312, line 71: So how do we handle these extra dots in the version exactly? A valid version could be something like `0.0.1-alpha.1` right?

thomastaylor312 (changes requested):
Mostly some comments to discuss here. Nothing super big

crates/wash-cli/src/common/link_cmd.rs
```diff
@@ -3,31 +3,29 @@ use std::collections::HashMap;
...
 
 use crate::appearance::spinner::Spinner;
 use crate::ctl::{link_del_output, links_table};
 
 /// Generate output for link put command
 pub fn link_put_output(
-    actor_id: impl AsRef<str>,
-    provider_id: impl AsRef<str>,
+    source_id: impl AsRef<str>,
+    target: impl AsRef<str>,
```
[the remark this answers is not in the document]
brooksmtownsend: Well, no not necessarily. So right now it's always a target ID, but in the future this might just be a group so it doesn't need to be a target.

There's some nuance there so, I just wanted to keep it `target` to be neutral for later when we can change it

crates/wash-lib/src/cli/link.rs
```diff
@@ -15,47 +11,60 @@ pub struct LinkDelCommand {
...
+    /// The WIT package of the link, e.g. "http" in "wasi:http/incoming-handler"
+    #[clap(name = "wit-package")]
+    pub wit_package: String,
+
+    /// The interface of the link, e.g. "incoming-handler" in "wasi:http/incoming-handler"
+    #[clap(long = "interface", required = true)]
+    pub interfaces: Vec<String>,
+
+    /// List of named configuration to make available to the source
+    #[clap(long = "source_config")]
```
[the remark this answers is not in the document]
brooksmtownsend: Ugh yes, I forgot about this. Thank you

crates/wash-cli/src/dev.rs
brooksmtownsend: It does, good point. Worth a sanitization pass

crates/wash-lib/src/cli/capture.rs
brooksmtownsend: I don't think we considered non-numeric versions for the protocol, but I'm unsure. @rvolosatovs may have thoughts here, but afaik it was only planned to me Major.Minor.Patch

crates/wash-cli/src/dev.rs
brooksmtownsend: I tried to validate component IDs all over the CLI with a value parser

crates/wash-lib/src/cli/mod.rs
```diff
@@ -345,6 +345,30 @@ pub fn labels_vec_to_hashmap(constraints: Vec<String>) -> Result<HashMap<String,
     Ok(hm)
 }
 
+/// This function is a simple helper to ensure that a component ID is a valid
+/// string containing only alphanumeric characters and underscores
+pub fn validate_component_id(id: &str) -> anyhow::Result<String> {
+    if id.chars().all(|c| c.is_alphanumeric() || c == '_') {
```
thomastaylor312, line 351: This can also have dashes

crates/wash-lib/src/cli/mod.rs
```diff
@@ -345,6 +345,30 @@ pub fn labels_vec_to_hashmap(constraints: Vec<String>) -> Result<HashMap<String,
...
+        bail!("Component ID must contain only alphanumeric characters and underscores")
+    }
+}
+
+/// This function is a simple helper to ensure that a component ID is a valid
+/// by transforming any non-alphanumeric characters to underscores
+pub fn sanitize_component_id(id: &str) -> String {
+    id.chars()
+        .map(|c| {
+            if c.is_alphanumeric() || c == '_' {
```
thomastaylor312, line 363: This also can have dashes

thomastaylor312: Super nit: consider making this char check a function so you can use it in the multiple places here

thomastaylor312 (changes requested):
Just a few minor things

thomastaylor312 (approved):

PR #1652, opened 2024-03-12
feat(provider-sdk): add `run_provider_handler`

rvolosatovs:
## Feature or Problem

- Simplify and remove unused code
- Add `run_provider_handler` for building providers using just `ProviderHandler` trait implementations, which manually serve using returned `wrpc_transport::Client` (e.g. all handwritten providers)


brooksmtownsend (commented):
This all looks fine to me, @vados-cosmonic deferring to you on review for anything in the bindgen macro etc etc

vados-cosmonic:
@brooksmtownsend will do. This is going to break my other PRs but let's get this merged in first since it's a good refactor of `provider-sdk`

crates/provider-sdk/src/lib.rs
```diff
@@ -7,25 +7,35 @@ use async_trait::async_trait;
...
+    format!("wasmbus.rpc.{lattice}.{provider_key}.linkdefs.del")
+}
+
+pub fn health_subject(lattice: &str, provider_key: &str) -> String {
+    format!("wasmbus.rpc.{lattice}.{provider_key}.health")
+}
+
+pub fn shutdown_subject(lattice: &str, provider_key: &str, link_name: &str) -> String {
+    format!("wasmbus.rpc.{lattice}.{provider_key}.{link_name}.shutdown")
+}
```
vados-cosmonic, line 38: this should probably be in `core::rpc`

crates/provider-sdk/src/provider.rs
```diff
@@ -81,6 +146,543 @@ macro_rules! process_until_quit {
...
+                    }
+                }
+            });
+        }
+        .instrument(span),
+    );
+    Ok(link_del_rx)
+}
+
+pub struct ProviderCommands {
```
vados-cosmonic, line 326: This is a bit of a weird name, maybe `ProviderCommandReceivers`?

crates/provider-sdk/src/provider.rs
```diff
@@ -81,6 +146,543 @@ macro_rules! process_until_quit {
...
+    pub host_id: String,
+    pub lattice_rpc_prefix: String,
+    pub link_name: String,
+    pub provider_key: String,
+    pub link_definitions: Vec<InterfaceLinkDefinition>,
+    pub commands: ProviderCommands,
+}
+
+#[instrument]
+pub async fn init_provider(name: &str) -> ProviderInitResult<ProviderInitState> {
```
vados-cosmonic, line 347: Does this need to be public? If so, we should document it -- if not, then let's remove `pub`

crates/provider-sdk/src/lib.rs
```diff
@@ -7,25 +7,35 @@ use async_trait::async_trait;
...
-pub use rpc_client::RpcClient;
-
-pub use wasmcloud_tracing;
 
 /// Re-export of types from [`wasmcloud_core`]
 pub use core::{
     HealthCheckRequest, HealthCheckResponse, InterfaceLinkDefinition, WasmCloudEntity, WitFunction,
     WitInterface, WitNamespace, WitPackage,
 };
+pub use provider::*;
```
vados-cosmonic, line 20: NIT: it's better to have our exports be explicit here so we can be very sure what we expose/don't expose to the SDK.

crates/provider-sdk/src/provider.rs
```diff
@@ -90,81 +692,76 @@ pub struct ProviderConnection {
...
+    nats: Arc<async_nats::Client>,
+    lattice: &str,
+    provider_key: &str,
+    target: &str,
+) -> wasmcloud_core::wrpc::Client {
+    let mut headers = HeaderMap::new();
+    headers.insert("source-id", provider_key);
+    headers.insert("target-id", target);
+    wasmcloud_core::wrpc::Client::new(nats, lattice, target, headers)
+}
```
vados-cosmonic, line 730: NIT: Should this be public?  maybe `pub(crate)`?

The vast majority of the time, people should be getting their client from an active connection via the SDK, not building their own, since you generally don't the `provider_key` without reading host data, and that information goes into the `ProviderConnection`.

vados-cosmonic (approved):
LGTM, nothing but nits :rocket:

crates/provider-sdk/src/lib.rs
rvolosatovs: I've moved it closer to usage, let's address this one later - I don't see similar functionality in core now

PR #1653, opened 2024-03-12
fix(provider-bindgen): remove argument bundling

vados-cosmonic:
## Feature or Problem

Up until now, for wasmbus (i.e. msgpack) compatibility, bindgen has had to bundle arguments to functions into structs that were dynamically generated by bindgen.

Given that wRPC can/must individually transmit arguments, we should no longer need this code, and can just deal with arguments and no bundling.

This commit removes bundling and argument handling strategies options and functionality from bindgen.


brooksmtownsend (approved):
🚀 Love PRs with more removals than additions

PR #1656, opened 2024-03-12
feat(provider-sdk): rename put_link to receive_link_config_as_*

vados-cosmonic:
## Feature or Problem

This commit renames `put_link` which was a part of the `ProviderHandler` trait to `receive_link_config_as_target` and `receive_link_config_as_source` depending on the position of the provider when the link is put.

With both of these explicit methods, users should be able to configure their providers appropriately depending on how the link has been put to them.


crates/host/src/wasmbus/mod.rs
```diff
@@ -2868,7 +2868,10 @@ impl Host {
                 // TODO(#1548): Providers should not receive [wasmcloud_control_interface::InterfaceLinkDefinition]s
                 // they should receive a list of named configurations to use
                 link_definitions: vec![],
-                config: config.get_config().await.clone(),
+                configs: HashMap::from([(
+                    "default".to_string(),
+                    config.get_config().await.clone(),
+                )]),
```
brooksmtownsend, line 2874: This should be different, the configuration as passed to the provider isn't based on a link name, it's the set of component configuration for that entire provider instance. The way I see it is that this parameter, `config`, should be that piece of static config (for example, a default connection address for every link), but we should have another parameter called `link_configs` which entirely replaces `link_definitions`. Does that make sense?

crates/core/src/host.rs
```diff
@@ -43,9 +44,9 @@ pub struct HostData {
     pub link_definitions: Vec<InterfaceLinkDefinition>,
     /// list of cluster issuers
     pub cluster_issuers: Vec<ClusterIssuerKey>,
-    /// Merged named configuration set for this provider at runtime
+    /// Merged named configuration set for this provider at runtime, indexed by link name
     #[serde(default)]
-    pub config: HashMap<String, String>,
+    pub configs: HashMap<LinkName, HashMap<String, String>>,
```
brooksmtownsend, line 49: Re: above comment, I think the config should stay the same, but this hashmap of link name -> config should replace `link_definitions`

crates/provider-sdk/src/error.rs
```diff
@@ -3,6 +3,9 @@
 pub type InvocationResult<T> = Result<T, InvocationError>;
 pub type ProviderInitResult<T> = Result<T, ProviderInitError>;
 
+/// Result form for [`ProviderHandler`] trait methods
+pub type ProviderOperationResult<T> = Result<T, Box<dyn std::error::Error + Send + Sync>>;
```
brooksmtownsend, line 7: 👨🏻‍🍳 💋

crates/provider-sdk/src/lib.rs
```diff
@@ -148,31 +150,125 @@ pub struct Context {
...
 pub trait Provider: ProviderHandler + WrpcNats + WrpcDispatch + Send + Sync + 'static {}
 
+/// Configuration of a link that is passed to a provider
+pub trait LinkConfig: Send + Sync {
+    /// Given that the link was established with the source as this provider,
+    /// get the target ID which should be a component
+    fn get_target_id(&self) -> &ComponentId;
+
+    /// Given that the link was established with the source as this provider,
+    /// get the target ID which should be a component
```
brooksmtownsend, line 160:
```suggestion
    /// get the source ID which should be a component
```

crates/provider-sdk/src/lib.rs
```diff
@@ -148,31 +150,125 @@ pub struct Context {
 /// The super trait containing all necessary traits for a provider
 pub trait Provider: ProviderHandler + WrpcNats + WrpcDispatch + Send + Sync + 'static {}
 
+/// Configuration of a link that is passed to a provider
+pub trait LinkConfig: Send + Sync {
+    /// Given that the link was established with the source as this provider,
+    /// get the target ID which should be a component
+    fn get_target_id(&self) -> &ComponentId;
+
+    /// Given that the link was established with the source as this provider,
```
brooksmtownsend, line 159:
```suggestion
    /// Given that the link was established with the target as this provider,
```

brooksmtownsend: It's confusing naming, unfortunately. But folks should call this function when `put_target_config` is called

crates/provider-sdk/src/lib.rs
```diff
@@ -148,31 +150,125 @@ pub struct Context {
...
+    /// get the target ID which should be a component
+    fn get_source_id(&self) -> &ComponentId;
+
+    /// Get the name of the link that was provided
+    fn get_link_name(&self) -> &ComponentId;
```
brooksmtownsend, line 164: This should just be a `String`, not a component ID

crates/provider-sdk/src/lib.rs
```diff
@@ -148,31 +150,125 @@ pub struct Context {
 /// The super trait containing all necessary traits for a provider
 pub trait Provider: ProviderHandler + WrpcNats + WrpcDispatch + Send + Sync + 'static {}
 
+/// Configuration of a link that is passed to a provider
+pub trait LinkConfig: Send + Sync {
+    /// Given that the link was established with the source as this provider,
+    /// get the target ID which should be a component
+    fn get_target_id(&self) -> &ComponentId;
```
brooksmtownsend, line 157: This may or may not be a component ID _technically_, but under the hood it's just a string so it's fine.

crates/provider-sdk/src/lib.rs
```diff
@@ -148,31 +150,125 @@ pub struct Context {
...
+        HashMap<String, String>,
+    )
+{
+    fn get_source_id(&self) -> &ComponentId {
+        &self.0
+    }
+
+    fn get_target_id(&self) -> &ComponentId {
+        &self.1
+    }
```
brooksmtownsend, line 184: This should be a `LatticeTarget` based on the impl, right

crates/provider-sdk/src/lib.rs
```diff
@@ -148,31 +150,125 @@ pub struct Context {
...
+    )
+{
+    fn get_source_id(&self) -> &ComponentId {
+        self.0
+    }
+
+    fn get_target_id(&self) -> &ComponentId {
+        self.1
+    }
+
```
brooksmtownsend, line 210: Same comment re: impl

crates/provider-sdk/src/provider.rs
```diff
@@ -499,33 +506,47 @@ impl ProviderConnection {
...
+            .unwrap_or_default();
+
+        info!("Linking actor with provider");
+        match provider
+            .receive_link_config_as_target((&ld.source_id, &ld.target, &ld.name, &config))
+            .await
+        {
+            Ok(()) => self.put_link(ld).await,
+            Err(e) => {
+                warn!(error = %e, "receive_link_config_as_target() failed");
```
brooksmtownsend, line 549: I think here we may need to actually check to see if the `source_id` == `provider_key` in order to tell if this is a source or target config

crates/provider-sdk/src/provider_main.rs
```diff
@@ -119,17 +119,39 @@ where
...
-                link_definition = ?ld,
-                "Failed to initialize link during provider startup",
-            );
-        } else {
-            connection.put_link(ld).await;
+        // Get config for the link name
+        //
+        // todo(vados-cosmonic): this doesn't make sense -- are named configs included in the link def
+        // supposed to be looked up by the provider when they're received? Why do we get an all combined one
+        // that is keyed by link name? (and if it wasn't keyed by link name it'd be all smushed together)
```
brooksmtownsend, line 130: We shouldn't be giving an all combined one, what we should be giving the provider at initialization time is a list of link configurations that includes the link name, the source ID and the target and based on the source we should deliver them to the provider for handling. It's up to the provider to store the config as it needs to

crates/provider-wit-bindgen-macro/src/lib.rs
```diff
@@ -554,19 +589,19 @@ pub fn generate(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
-            pub fn new(ld: &'a ::wasmcloud_provider_wit_bindgen::deps::wasmcloud_provider_sdk::core::InterfaceLinkDefinition) -> Self {
+        impl InvocationHandler {
+            pub fn new(target: ::wasmcloud_provider_wit_bindgen::deps::wasmcloud_provider_sdk::core::ComponentId) -> Self {
                 let connection = ::wasmcloud_provider_wit_bindgen::deps::wasmcloud_provider_sdk::provider_main::get_connection();
                 // NOTE: The link definition that is used here is likely (source_id=?, target=<provider>)
                 //
                 // todo(vados-cosmonic): This invocation handler should arguably create a uni-directional
                 // "return" link to anything it wants to call, since links are uni-directional now.
-                Self { ld, wrpc_client: connection.get_wrpc_client(&ld.source_id) }
+                Self { target: target.clone(), wrpc_client: connection.get_wrpc_client(target) }
```
brooksmtownsend, line 604: Sanity check me here, this client already includes the lattice name in there as well correct?

brooksmtownsend (changes requested):
Big fan of the new descriptive handlers and `LinkConfig`, I think despite a possibly initial set of confusion your doc comments clear it up real well and we can document this properly.

Requested changes around some types, and primarily the method of delivering config to the provider. The initial config bundle shouldn't be keyed by link name as it's global provider instance configuration

crates/provider-sdk/src/lib.rs
vados-cosmonic: *`LinkName`

crates/provider-sdk/src/lib.rs
vados-cosmonic: Ah no, I'll change this to `LatticeTarget` since we have a proper name for it!

crates/provider-sdk/src/lib.rs
vados-cosmonic: Fixed!

crates/provider-sdk/src/lib.rs
vados-cosmonic: Fixed!

crates/provider-sdk/src/provider.rs
vados-cosmonic: Yeah I did this in the other spot, will do here as well

crates/provider-wit-bindgen-macro/src/lib.rs
vados-cosmonic: Yup, it does! the `target` probably doesn't have to be saved though, so I might be able to remove that

crates/core/src/host.rs
vados-cosmonic: OK I've reverted this for now -- just the `config` that comes in gets used

crates/host/src/wasmbus/mod.rs
vados-cosmonic: Sure -- I'm a bit confused on how the static config works for multiple link names that come through, but I get that it's not supposed to be stored here -- I've reverted this

crates/provider-sdk/src/provider_main.rs
vados-cosmonic: Ah see the note earlier, got it! I've reverted the `configs` change

crates/core/src/host.rs
```diff
@@ -43,7 +43,7 @@ pub struct HostData {
     pub link_definitions: Vec<InterfaceLinkDefinition>,
     /// list of cluster issuers
     pub cluster_issuers: Vec<ClusterIssuerKey>,
-    /// Merged named configuration set for this provider at runtime
+    /// Merged named configuration set for this provider at runtime, indexed by link name
```
brooksmtownsend, line 46: This still isn't quite right, the `config` is actually just one single HashMap that's materialized from the named configurations given at runtime.

So for example, I'd start a provider by doing something like this:
```bash
wash config put http_reasonable_default PORT=8080
wash start provider wasmcloud.azurecr.io/httpserver:v1 my_httpserver_id --config http_reasonable_default
```

At runtime, the idea is that the HTTP server would startup, and in it's initialization config it would get 
```json
{
    "PORT": "8080"
}
```

And the provider can use that info to set a global default port, for example. Does that make sense?

vados-cosmonic: Ah forgot to take this line out -- will remove, this is from back when it was `configs`!

The set up make sense to me, it's the merging them together, and how a single provider can serve multiple clients with multiple configs that's a bit fuzzy for me. But this actual comment is just old

vados-cosmonic: So this is fixed now -- I removed the addition here. I can't merge this though until https://github.com/wasmCloud/wasmCloud/pull/1647 lands.

PR #1661, opened 2024-03-12
feat(host)!: refactor component spec, deliver links to providers

brooksmtownsend:

## Feature or Problem
This PR refactors the component specification `links` parameter from a nested hashmap designed for component import lookups to a `Vec<InterfaceLinkDefinition>` which allows for storing the entirety of the link and (somewhat) easier lookups. It's less efficient than a map/set structure at lookups, but the only way to use a hashset like before would be to provide a very strange `Hash,Eq` impl that asserts links are equal when they are truly not.Instead, I opt to translate the list of links in a component specification into the efficient lookup structure that a component needs when the component starts or the spec updates. 

This PR also ensures we're storing component specifications for providers, supporting putting links before a component actually starts, and delivers links to providers at startup and anytime they are changed afterwards.

**Very important nuance of this PR**: I move all of the actual modification of inmemory structures for actors and host links into the `handle_component_spec_put` function, which is the function every host in the lattice will be notified of when there's a change to the component spec. This ensures that the host that handles the link put / link del makes the proper modification to jetstream state, and each host (including that host) simply catches up on that state in order to handle links. This ensures each host on startup also populates link definitions.

The only thing I'm not super happy with is the fact that when we receive a link, I need to publish to two topics for providers depending on if it's the source or the target of the invocation. I think we can be a bit smarter here, and I think we'll need to change this code once we support routing groups, but I can't think of a way to do it better now.

I left a few TODOs which actually encapsulate delivering the real configuration to providers instead of named config. I'll need to address that once #1647 and #1656 merge. (@rvolosatovs by the way here, we're passing configuration as a string inside of the link configuration. So to test your HTTPserver provider, set `source_config` to something like `["PORT=8080"]` and split on the `=` to get the value. It's hacky, but will be replaced as soon as the aformentioned PRs are in.

## Related Issues
Enables #1659 

## Consumer Impact
If you've been testing wasmCloud since feat/wrpc merged, please make sure to delete the contents of your LATTICEDATA bucket before using this PR. I changed the shape of the component spec entirely from a hashmap to a vec in order to support actually storing the proper information.

### Unit Test(s)
Added one unit test so you all can verify my wild function to transform a Vec<InterfaceLinkDefinition> -> HashMap<String, HashMap<String, HashMap<String, String>>> works as expected 😉

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1,5 +1,5 @@
 use std::collections::hash_map::{self, Entry};
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap};
 use std::env;
 use std::env::consts::{ARCH, FAMILY, OS};
 use std::future::Future;
@@ -1133,13 +1133,6 @@ impl std::fmt::Debug for InvocationParams {
 }
 
 impl Actor {
-    /// Returns the component specification for this unique actor
-    pub(crate) async fn component_specification(&self) -> ComponentSpecification {
-        let mut spec = ComponentSpecification::new(&self.image_reference);
-        spec.links = self.handler.interface_links.read().await.clone();
-        spec
-    }
-
     /// Handle an incoming wRPC request to invoke an export on this actor instance.
     #[instrument(
         level = "info",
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1323,14 +1316,14 @@ pub struct Host {
...
-    links: RwLock<HashMap<String, HashSet<InterfaceLinkDefinition>>>,
+    links: RwLock<HashMap<String, Vec<InterfaceLinkDefinition>>>,
     actor_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::Actor>>>>, // TODO: use a single map once Claims is an enum
     provider_claims: Arc<RwLock<HashMap<String, jwt::Claims<jwt::CapabilityProvider>>>>,
     config_data_cache: Arc<RwLock<ConfigCache>>,
     metrics: Arc<HostMetrics>,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Default)]
```
rvolosatovs, line 1326: I don't feel strongly about it, but I would not expect this struct to implement `Default` - I feel like every component specification requires a URL

rvolosatovs (approved):

rvolosatovs:
This was supposed to be auto-merged, doing it myself!

PR #1668, opened 2024-03-13
chore!: update wasi-cloud interfaces

ricochet:


examples/rust/actors/http-hello-world/wit/deps/lattice/lattice-control.wit
brooksmtownsend, line 1: I think we should keep this one out of the example, only because it's likely to change

brooksmtownsend (changes requested):

crates/runtime/wit/deps/wasmcloud/host.wit
```diff
@@ -0,0 +1,14 @@
+package wasmcloud:host;
```
rvolosatovs, line 1: this does not look right, I'd suggest `rm -rf wit/deps` and trying again from latest `main`

crates/runtime/wit/deps/wasmcloud/bus.wit
```diff
@@ -1,35 +0,0 @@
-package wasmcloud:bus;
```
rvolosatovs, line 1: this should not have been removed

crates/runtime/wit/deps.toml
```diff
@@ -1,7 +1,14 @@
 blobstore = "https://github.com/WebAssembly/wasi-blobstore/archive/main.tar.gz"
 http = "https://github.com/WebAssembly/wasi-http/archive/v0.2.0.tar.gz"
-io = "../../../wit/deps/io"
+io = "https://github.com/WebAssembly/wasi-io/archive/v0.2.0.tar.gz"
 keyvalue = "https://github.com/WebAssembly/wasi-keyvalue/archive/main.tar.gz"
 logging = "https://github.com/WebAssembly/wasi-logging/archive/main.tar.gz"
 messaging = "https://github.com/wasmCloud/messaging/archive/main.tar.gz"
-wasmcloud = "../../../wit"
+wasmcloud = "../wit"
```
rvolosatovs, line 7: Why is this changed?

crates/runtime/wit/deps.toml
```diff
@@ -1,7 +1,14 @@
...
+
+# resolution for transitive dependency conflict for `cli`, `clocks`, `sockets` and `random`
+cli = "https://github.com/WebAssembly/wasi-cli/archive/main.tar.gz"
```
rvolosatovs, line 10: I think we should use tagged versions for all of these.

crates/runtime/wit/deps.toml
```diff
@@ -1,7 +1,14 @@
 blobstore = "https://github.com/WebAssembly/wasi-blobstore/archive/main.tar.gz"
 http = "https://github.com/WebAssembly/wasi-http/archive/v0.2.0.tar.gz"
-io = "../../../wit/deps/io"
+io = "https://github.com/WebAssembly/wasi-io/archive/v0.2.0.tar.gz"
```
rvolosatovs, line 3: I'd expect this to not be required if all other imports depend on the same `wasi-io` version

crates/runtime/wit/deps.toml
ricochet: I likely should have deleted my deps folder

Issue #846, opened 2023-06-06
[FEATURE] `wash up` should be idempotent


connorsmith256:
Running `wash up` multiple times on the same machine makes for cool demos, but can also lead to a confusing user experience, or worse, in the case where `wash up` is programmatically being invoked

`wash up` should be idempotent by default. If a wasmCloud host is already running, a friendly message should be displayed indicating the state

For users who do want to run multiple hosts, a flag (defaults to false) should be added to allow a second host to connect to the NATS server. Naming is hard and I haven't thought much about what the name should be, so I'll throw out a strawman and hope someone else has a better idea: `wash up --increment` 

This might be resolved incidentally by https://github.com/wasmCloud/wash/pull/556, though I'm not sure what message will be displayed

cdmurph32:
In my PR, I went with the most basic approach, which is just to allow more hosts to be created with `--scale`. Things would get a more complicated if we wanted to specify how many hosts to create, ie `--scale 2`, so it seemed like a good place to start.
If we go with the more complicated version:
- `--scale` might be a bit confusing. Let's say I have 5 hosts running and I run `wash up --scale 2`. Would the use expect a result of 7 hosts or 2 hosts? If the desired behavior is the latter, then we are now merging the functionality of `wash up` and `wash down`.
- Would we want to add `--scale` to `wash down` as well?

brooksmtownsend:
@cdmurph32 In the interest of keeping this simple... do we need to support a `--scale` function? What if we just supported launching one locally? IMO there's only a few scenarios where you really want to launch multiple hosts locally, do we want to save `--scale` for a future enhancement?

brooksmtownsend:
To actually answer your question though, I think the desired behavior here would be for `--scale N` to ensure there are `N` running hosts, whether that means scaling up or down. `wash down` could just be an alias for `--scale 0`

cdmurph32:
Sounds. Good. Sorry I just saw this. Gmail is awful.

cdmurph32:
Hey @brooksmtownsend. I'm looking into it not. 

- Would that message only be printed if `-d` is passed? Do we just treat it as if `-d` was passed if a host is already running?
- It looks like the code allows you to pass a wadm manifest, but `wash up --help` doesn't seem to show those options. Should a wadm manifest still be applied if there is already a host running?

brooksmtownsend:
> Would that message only be printed if -d is passed? Do we just treat it as if -d was passed if a host is already running?

[...]

PR #1674, opened 2024-03-14
Wash up should be idempotent.

cdmurph32:
If a wasmcloud host is runnning, `wash up` will not start an additional host unless `--scale` is passed.

When starting multiple hosts, NATS should be run independently of wasmcloud, otherwise killing the original `wash up` will also stop NATS.

## Feature or Problem
Wash up should be idempotent https://github.com/wasmCloud/wasmCloud/issues/846

## Consumer Impact
This will change how users start multiple wasmCloud hosts on the same machine

## Testing
Integration test added.

### DIFF
--- a/crates/wash-cli/src/up/mod.rs
+++ b/crates/wash-cli/src/up/mod.rs
@@ -1,15 +1,14 @@
 use std::collections::HashMap;
 use std::fmt::Write;
 use std::io::ErrorKind;
-use std::path::Path;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
     Arc,
 };
 
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{bail, Context, Result};
 use async_nats::Client;
 use clap::Parser;
 use serde_json::json;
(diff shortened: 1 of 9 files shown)


crates/wash-cli/src/up/mod.rs
```diff
@@ -386,6 +384,15 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
     // If this fails, we should return early since wasmCloud wouldn't be able to connect either
     nats_client_from_wasmcloud_opts(&wasmcloud_opts).await?;
 
+    if tokio::fs::try_exists(install_dir.join(WASMCLOUD_PID_FILE))
```
thomastaylor312, line 387: Nit: `!cmd.wasmcloud_opts.multi_local` should probably go first so it short circuits if `multi_local` is set. Otherwise it will always try to check for file existence when it might not need to

crates/wash-cli/tests/common/mod.rs
```diff
@@ -490,6 +491,15 @@ impl TestWashInstance {
             .output()
             .await
             .context("failed to stop host")?;
+        let install_dir = downloads_dir()?;
+        let lock_file_exists = tokio::fs::try_exists(install_dir.join(WASMCLOUD_PID_FILE)).await?;
+        if lock_file_exists {
```
thomastaylor312, line 496: Nit: Might be worth adding a comment here. It took me a second to catch that you were deleting it as a cleanup step if it still existed

thomastaylor312 (approved):
Sorry for the delay @cdmurph32, got sidetracked with KubeCon stuff. I'm good with this pending passing tests. Might be worth rebasing to make sure you picked up any of the other test fixes we added in

crates/wash-cli/src/up/mod.rs
cdmurph32: Nice catch. I've been writing too much javascript.

cdmurph32:
I'm still trying to get the integration tests working. I should be able to get it done tomorrow.

cdmurph32:
All of the tests are working locally for me now.

thomastaylor312 (approved):

PR #1676, opened 2024-03-14
feat(provider-sdk): add support for init()

vados-cosmonic:
## Feature or Problem

This commit adds support for a implementer-provided provider initialization hook that `provider-sdk` can call during initialization.

### DIFF
--- a/crates/provider-sdk/src/lib.rs
+++ b/crates/provider-sdk/src/lib.rs
@@ -5,6 +5,7 @@ use anyhow::Context as _;
 use async_nats::{ConnectOptions, Event};
 use async_trait::async_trait;
 use core::{ComponentId, LatticeTarget, LinkName};
+use provider::ProviderInitState;
 use tracing::{error, info, warn};
 
 pub mod error;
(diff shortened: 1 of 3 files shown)


crates/provider-sdk/src/lib.rs
```diff
@@ -203,9 +204,45 @@ impl LinkConfig
...
+    /// Get host-configured provider ID.
+    ///
+    /// This value may not be knowable to the provider at build time but must be known by runtime.
+    fn get_provider_id(&self) -> &str;
+
+    /// Retrieve the static configuration for the provider
+    ///
+    /// This normally consists of named configuration that were set for the provider,
+    /// merged, and received from the host *before* the provider has started initialization.
+    fn get_static_config(&self) -> &HashMap<String, String>;
```
brooksmtownsend, line 221: Do you think it's a bad name calling it "static" config if we might support updating it in the future? Or is that acceptable? I know this is what I've been calling it, but now is a good time for a gut check 😄  cc @thomastaylor312

brooksmtownsend (commented):
Looks good to me 👍🏻 I'd like to see us hook into this function in the future rather than loading host data in our providers to show when you'd use it

crates/provider-sdk/src/lib.rs
vados-cosmonic: I can definitely rename this `config` but I'd note that this `trait` in particular is named very specifically for *init* -- I don't expect it to be called again. 

Definitely happy to change this to just `get_config()` though, so will do that!

brooksmtownsend: Perfect 😄 We do eventually want to support updating this config in the future, so perhaps we can add another trait method here like "update" where providers can opt into updates to config

vados-cosmonic: Hey sorry a bit late getting back to this, but can we add that when the functionality exists? This was really more about `init()` only

thomastaylor312 (approved):

crates/provider-sdk/src/lib.rs
vados-cosmonic: Filing this so we don't forget it:

https://github.com/wasmCloud/wasmCloud/issues/1747

PR #1677, opened 2024-03-14
feat(wash-cli)!: wash config put/get/del implementation

brooksmtownsend:
## Feature or Problem
This PR implements the `wash config` subcommand, including `get`, `put`, and `del`.

I had to make a change in the host to actually watch the config data bucket to get updates on config, so tests for this feature in wash will have to wait for the next alpha wasmcloud release.

## Release Information
wasmcloud v1.0.0-alpha.2
next wash

### Manual Verification
I manually made sure all of this worked! I also included a special case to ensure we aren't inserting `success:true` when people fetch JSON config, might conflict with their keys and/or cause confusion.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1289,8 +1289,6 @@ struct Provider {
     claims: Option<jwt::Claims<jwt::CapabilityProvider>>,
 }
 
-type ConfigCache = HashMap<String, HashMap<String, String>>;
-
 /// wasmCloud Host
 pub struct Host {
     // TODO: Clean up actors after stop
@@ -1310,6 +1308,7 @@ pub struct Host {
     /// NATS client to use for RPC calls
     rpc_nats: Arc<async_nats::Client>,
     data: Store,
+    /// Task to watch for changes in the LATTICEDATA store
     data_watch: AbortHandle,
     config_data: Store,
     config_generator: BundleGenerator,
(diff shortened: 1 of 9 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1882,6 +1887,43 @@ impl Host {
             }
         });
 
+        let config_watch: JoinHandle<anyhow::Result<_>> = spawn({
```
thomastaylor312, line 1890: So I actually got rid of this because the config bundles do the caching/watching for us now. I figured the ctl interface could do a passthrough to the KV bucket. Is the concern here around how often wadm might be querying the config? I actually think we are ok there too.

If we do want to go back to this, we should probably change out how the config bundle works. Otherwise we end up with possibly _two_ whole copies of all config in memory

crates/wash-lib/src/cli/mod.rs
```diff
@@ -329,16 +329,16 @@ pub fn extract_keypair(
     }
 }
 
-/// Transforms a list of labels in the form of (label=value) to a hashmap
-pub fn labels_vec_to_hashmap(constraints: Vec<String>) -> Result<HashMap<String, String>> {
+/// Transforms a list of key in the form of (key=value) to a hashmap
+pub fn input_vec_to_hashmap(values: Vec<String>) -> Result<HashMap<String, String>> {
```
thomastaylor312, line 333: Is this going to cause an issue for `wash label`?

thomastaylor312 (changes requested):
Mostly there, just two big discussion items

crates/host/src/wasmbus/mod.rs
brooksmtownsend: @thomastaylor312 gotcha, then I think we may not have the config bundle fully implemented since I wasn't getting config updates before I re-added this. Let me take a second look and see if I just missed something simple with the config generator

crates/wash-lib/src/cli/mod.rs
brooksmtownsend: Looks like `wash label` does a custom parse instead of using this function 🤔

thomastaylor312 (approved):

PR #1678, opened 2024-03-14
feat: implement `wasi:http/outgoing-handler` provider

rvolosatovs:
## Feature or Problem

HTTP - now with streams! :)


crates/providers/http-client/src/lib.rs
```diff
@@ -1,177 +1,117 @@
...
-
-wasmcloud_provider_wit_bindgen::generate!({
-    impl_struct: HttpClientProvider,
-    contract: "wasmcloud:httpclient",
-    replace_witified_maps: true,
-    wit_bindgen_cfg: "provider-http-client"
-});
+#[instrument(level = "trace", skip_all)]
+pub async fn serve_handle<Ctx, Tx, C>(
+    client: hyper_util::client::legacy::Client<
```
thomastaylor312, line 15: Do we have to worry about using the legacy client here? Just want to make sure it isn't going to get ripped out soon

thomastaylor312 (approved):

crates/providers/http-client/src/lib.rs
rvolosatovs: not really - there is no other client currently provided by hyper and is what's suggested to use in the docs of https://docs.rs/hyper-rustls/latest/hyper_rustls/index.html

brooksmtownsend (approved):

PR #1680, opened 2024-03-15
Implement `kvredis` in terms of `wrpc:keyvalue`

rvolosatovs:
## Feature or Problem

Switch `kvredis` provider to `wrpc:keyvalue`. 
The test is adapted from wasmbus.rs (RIP) and builtins-component-reactor


crates/host/src/wasmbus/mod.rs
```diff
@@ -169,7 +169,7 @@ impl Queue {
 
 #[derive(Clone, Debug)]
 struct Handler {
-    nats: async_nats::Client,
+    nats: Arc<async_nats::Client>,
```
thomastaylor312, line 172: Any reason you're wrapping the NATS client in an Arc? A NATS client is cheaply clonable as most of the inner data members are in an Arc already (or are meant to be cloned): https://docs.rs/async-nats/latest/src/async_nats/client.rs.html#66

rvolosatovs: My main concern is the prefix `String` stored within the client - since this is a value we're cloning quite often, it seems that cloning a single `Arc` is much more efficient than cloning multiple (contained within the client) along with the mutable heap-allocated string buffer.
I really would have expected the prefix field to be `Arc<str>`

thomastaylor312: Yeah, we should probably just contribute this back to async nats. I can do that in a bit, but we can keep this arc for now

thomastaylor312 (approved):

crates/host/src/wasmbus/mod.rs
thomastaylor312: See https://github.com/nats-io/nats.rs/pull/1236

PR #1681, opened 2024-03-15
feat(providers): implement `wrpc:blobstore/blobstore` for FS

rvolosatovs:
## Feature or Problem

migrate `blobstore-fs` to `wrpc:blobstore/blobstore`

Moved the test from `builtins-component-reactor`, we'll have to expand it a bit going forward, but this should be good enough for now

## Related Issues

https://github.com/WebAssembly/wasi-blobstore/issues/7


tests/actors/rust/wrpc-pinger-component/wit/deps/blobstore/world.wit
```diff
@@ -0,0 +1,5 @@
+package wasi:blobstore@0.2.0-draft;
+
+world imports {
+	import blobstore;
```
ricochet, line 4: NIT: We should have a formatter for WIT. Noticed this has tabs whereas WASI landed on a convention of spaces.

thomastaylor312 (approved):

PR #1685, opened 2024-03-17
WIP: [issue#1624-modified the default key_directory for wash build]

siddharthkhonde:
## Feature or Problem
`wash build` always generates keys in the local folder instead of `$HOME/.wash/keys

## Testing
this is a draft PR with some work still in progress.

### Manual Verification
modified the default key_directory for `wash build`. tried to run wasmCloud locally to see if the changes are being resonated while creating a new actor.

### DIFF
--- a/crates/wash-cli/src/build.rs
+++ b/crates/wash-cli/src/build.rs
@@ -58,7 +58,6 @@ pub async fn handle_command(command: BuildCommand) -> Result<CommandOutput> {
 
     match config.project_type {
         TypeConfig::Actor(ref actor_config) => {
-            print!("Inside wash build of an actor");
             let sign_config = if command.build_only {
                 None
             } else {
--- a/crates/wash-cli/src/build.rs
+++ b/crates/wash-cli/src/build.rs
@@ -58,6 +58,7 @@ pub async fn handle_command(command: BuildCommand) -> Result<CommandOutput> {
 
     match config.project_type {
         TypeConfig::Actor(ref actor_config) => {
+            print!("Inside wash build of an actor");
             let sign_config = if command.build_only {
                 None
             } else {
--- a/crates/wash-lib/src/parser/mod.rs
+++ b/crates/wash-lib/src/parser/mod.rs
@@ -116,7 +116,7 @@ impl TryFrom<RawActorConfig> for ActorConfig {
             // TODO(#1624): Default to ~/.wash/keys
             key_directory: raw_config
                 .key_directory
-                .unwrap_or_else(|| PathBuf::from("./keys")),
+                .unwrap_or_else(|| PathBuf::from("~/.wash/keys")),
             wasm_target: raw_config
                 .wasm_target
                 .map(WasmTarget::from)
@@ -186,7 +186,7 @@ impl TryFrom<RawProviderConfig> for ProviderConfig {
             // TODO(#1624): Default to ~/.wash/keys
             key_directory: raw_config
                 .key_directory
-                .unwrap_or_else(|| PathBuf::from("./keys")),
+                .unwrap_or_else(|| PathBuf::from("~/.wash/keys")),
         })
     }
 }
--- a/crates/wash-lib/src/parser/mod.rs
+++ b/crates/wash-lib/src/parser/mod.rs
@@ -110,7 +110,7 @@ impl TryFrom<RawActorConfig> for ActorConfig {
     type Error = anyhow::Error;
 
     fn try_from(raw_config: RawActorConfig) -> Result<Self, Self::Error> {
-        let key_dir = if let Some(key_directory) = raw_config.key_directory {
+        let key_directory = if let Some(key_directory) = raw_config.key_directory {
             key_directory
         } else {
             let home_dir = dirs::home_dir()
@@ -120,7 +120,7 @@ impl TryFrom<RawActorConfig> for ActorConfig {
         Ok(Self {
             claims: raw_config.claims.unwrap_or_default(),
             push_insecure: raw_config.push_insecure.unwrap_or(false),
-            key_directory: key_dir,
+            key_directory,
             wasm_target: raw_config
                 .wasm_target
                 .map(WasmTarget::from)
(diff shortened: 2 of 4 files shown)


crates/wash-lib/src/parser/mod.rs
```diff
@@ -116,7 +116,7 @@ impl TryFrom<RawActorConfig> for ActorConfig {
             // TODO(#1624): Default to ~/.wash/keys
             key_directory: raw_config
                 .key_directory
-                .unwrap_or_else(|| PathBuf::from("./keys")),
+                .unwrap_or_else(|| PathBuf::from("~/.wash/keys")),
```
brooksmtownsend, line 119: Just a note for later that `~` will probably work well on all unix systems but _might_ be strange on Windows. Last I checked it does work with an alias on Windows 11, but we may want to consider a crate like https://crates.io/crates/homedir (perhaps we already have one in `Cargo.toml` that I'm forgetting about?) for this

brooksmtownsend:
Hey @siddharthkhonde , would you mind enumerating what kinds of tests you've run, and if you see any difference in behavior after your change?

siddharthkhonde:
> Hey @siddharthkhonde , would you mind enumerating what kinds of tests you've run, and if you see any difference in behavior after your change?

Hi @brooksmtownsend , sure. I performed the following steps to test:
- I ran the wasmCloud host locally after making the above changes
- created a new actor and ran wash build inside the actor directory
- it seems like the changes are not getting recognized by the actor as it is still creating the keys inside the /.keys directory in the actor

crates/wash-lib/src/parser/mod.rs
siddharthkhonde: my bad. I completely forgot to think of other OS. really appreciate the Rust crate link. let me research a little on that.

siddharthkhonde: hi @brooksmtownsend , agreed. I was thinking of identifying the issue on unix systems first and getting the keys generated just in the user's $HOME directory. I can then take up the task task for `wash build` to detect the user's OS. Please let me know your thoughts on this :)

thomastaylor312: Heads up that we have the `dirs` crate available that can help with the home directory: https://docs.rs/dirs

crates/wash-cli/src/build.rs
```diff
@@ -58,6 +58,7 @@ pub async fn handle_command(command: BuildCommand) -> Result<CommandOutput> {
 
     match config.project_type {
         TypeConfig::Actor(ref actor_config) => {
+            print!("Inside wash build of an actor");
```
thomastaylor312, line 61: Left over debug?

crates/wash-lib/src/parser/mod.rs
```diff
@@ -109,14 +109,19 @@ struct RawActorConfig {
...
-    fn try_from(raw_config: RawActorConfig) -> Result<Self> {
+    fn try_from(raw_config: RawActorConfig) -> Result<Self, Self::Error> {
+        let key_dir = if let Some(key_directory) = raw_config.key_directory {
+            key_directory
+        } else {
+            let home_dir = dirs::home_dir()
+                .ok_or_else(|| anyhow::anyhow!("Unable to determine the user's home directory"))?;
+            home_dir.join(".wash/keys")
+        };
+        print!("User's key directory: {}", key_dir.display());
```
thomastaylor312, line 120: Left over debug?

thomastaylor312 (changes requested):
Looking real close! Thanks for adding in the test changes as well. Should just be some debug code to remove

crates/wash-lib/src/parser/mod.rs
siddharthkhonde: ah, apologies. rookie miss :)

crates/wash-cli/src/build.rs
siddharthkhonde: ah, apologies. rookie miss :)

crates/wash-lib/src/parser/mod.rs
```diff
@@ -109,14 +109,18 @@ struct RawActorConfig {
...
+            home_dir.join(".wash/keys")
+        };
         Ok(Self {
             claims: raw_config.claims.unwrap_or_default(),
             push_insecure: raw_config.push_insecure.unwrap_or(false),
-            // TODO(#1624): Default to ~/.wash/keys
-            key_directory: raw_config
-                .key_directory
-                .unwrap_or_else(|| PathBuf::from("./keys")),
+            key_directory: key_dir,
```
brooksmtownsend, line 123: nit: if you call this variable `key_directory` the last line here is simpler with the syntactic sugar.
```rust
let key_directory = ...
...
        Ok(Self {
            claims: raw_config.claims.unwrap_or_default(),
            push_insecure: raw_config.push_insecure.unwrap_or(false),
            key_directory,
```

crates/wash-lib/src/parser/mod.rs
```diff
@@ -171,7 +175,14 @@ struct RawProviderConfig {
 impl TryFrom<RawProviderConfig> for ProviderConfig {
     type Error = anyhow::Error;
 
-    fn try_from(raw_config: RawProviderConfig) -> Result<Self> {
+    fn try_from(raw_config: RawProviderConfig) -> Result<Self, Self::Error> {
+        let key_dir = if let Some(key_directory) = raw_config.key_directory {
```
brooksmtownsend, line 179: Same nit re: key_directory

brooksmtownsend (changes requested):
This looks like a nice, elegant fix! Only request to change here really is just to rename the PR and the commits to remove the "WIP" prefix 😄. If you'd like, I can do that as well, I just don't want to push over your commits.

I left two nits that are optional to fix.

crates/wash-lib/src/parser/mod.rs
siddharthkhonde: great catch. updated to a more cleaner code :)

crates/wash-lib/src/parser/mod.rs
siddharthkhonde: great catch. updated to a more cleaner code :)

thomastaylor312 (approved):

PR #1693, opened 2024-03-18
Switch to `wasmcloud:messaging` in providers

rvolosatovs:
## Feature or Problem

Switch to `wasmcloud:messaging` in providers (https://github.com/wasmCloud/messaging/)

That's the interface, which was implemented in the host for a long time - compared to in-tree interface it also provides error handling among other things.

## Related Issues

Removed `request-multi` from the interface in https://github.com/wasmCloud/messaging/pull/4 for simplicity and minimizing API surface - I don't think we ever really used it

Blocked on #1690 

### Unit Test(s)

Existing tests cover NATS provider, kafka was untested, and still is


crates/providers/messaging-nats/wit/deps/messaging/messaging.wit
```diff
@@ -1,50 +1,34 @@
...
-    record message {
-        /// Subject (or topic) of the message
+// Types common to message broker interactions
+interface types {
+    // A message sent to or received from a broker
+    record broker-message {
         subject: string,
-
-        /// Topic on which the reply should be sent
+        body: option<list<u8>>,
```
thomastaylor312, line 15: Why did we change this to an option? It seems much cleaner to just have an empty body rather than an optional one, especially since it is a list

rvolosatovs: That's what was implemented in the host for the last ~year. I can change it to a `list<u8>`, but the intention of this PR was just getting to use the same contract in the providers as the host

rvolosatovs: https://github.com/wasmCloud/messaging/pull/4/commits/09f3b13faacb37f05ca31dc061cc1dd22a4a2821

will integrate this now

thomastaylor312 (approved):

PR #1697, opened 2024-03-18
feat(examples): Adds blobby component

thomastaylor312:
This adds blobby as a component example that others can use. I'll be the first to admit it is a bit hacky, but it does maintain most of its functionality


brooksmtownsend (approved):

examples/rust/components/blobby/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+    }
+    Ok(())
+}
+
+fn send_response_error(response_out: ResponseOutparam, error: Error) {
+    let response = OutgoingResponse::new(Fields::new());
+    response
+        .set_status_code(error.status_code.as_u16())
+        .expect("Unable to set status code");
+    let response_body = response.body().unwrap();
```
rvolosatovs, line 67: nit; `expect` explaining that this can only be called once would be more informational for an example (https://github.com/WebAssembly/wasi-http/blob/13dad387dc2cd5ba2849f531b1848d6c63a683d9/wit/types.wit#L277-L279)

examples/rust/components/blobby/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+    response_body
+        .write()
+        .unwrap()
+        .blocking_write_and_flush(error.message.as_bytes())
+        .unwrap();
```
rvolosatovs, line 72: apart from changing `unwrap`s to `expect` with clear panic messages, looking at this example could potentially signal to developers that this message is guaranteed to be fully sent (it's not) https://github.com/WebAssembly/wasi-io/blob/324be895965666805cb1c622ed4f071b9e3cbd65/wit/streams.wit#L134-L135

This operation should always be performed in a loop for byte lists of unknown size

examples/rust/components/blobby/src/wrapper.rs
```diff
@@ -0,0 +1,112 @@
+// This is adapted from `wasmcloud-actor` but instead is impl'd directly on the type. This is
```
rvolosatovs, line 1: Why not use `wasmcloud_actor` adapters directly?

That's the only necessary change to implementation to make it work: https://github.com/wasmCloud/wasmCloud/blob/ffad11e8caf1fe39aead4ed9843bfd21b1c248ba/tests/actors/rust/builtins-component-reactor/src/lib.rs#L6-L8

examples/rust/components/blobby/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+        log(
+            Level::Error,
+            "put_object",
+            &format!("Failed to flush data to blobstore: {}", e),
+        );
+        return Err(Error {
+            status_code: StatusCode::INTERNAL_SERVER_ERROR,
+            message: format!("Failed to flush data to blobstore: {}", e),
+        });
+    }
```
rvolosatovs, line 376: why not use `blocking-splice` instead? I don't think `forward` will be added to streams, since `splice` already exists

examples/rust/components/blobby/src/lib.rs
thomastaylor312: I'll just change this to use `write_all`

examples/rust/components/blobby/src/wrapper.rs
thomastaylor312: Wasmcloud actor isn't published and I expect people to work from this as a starting point from their own applications

examples/rust/components/blobby/src/lib.rs
thomastaylor312: Blocking splice wasn't working for me and it required me to write a bunch of looping stuff and know the content length to read. This was simple enough to get started and then we can go back and optimized

thomastaylor312: Oh and I did check with @ricochet and she says she's pretty sure forward is coming in. We can circle back on this and improve in a follow up

PR #1701, opened 2024-03-19
ci: build providers

rvolosatovs:
## Feature or Problem

Make providers part of the workspace and build them alongside `wasmcloud` binary. CI for actually turning these into PARs will follow.

I'm hoping we can get this PR merged fast to avoid painful merge conflicts later

- remove the need for custom build scripts for providers
- ensure that providers are built for all targets supported by wasmcloud host (note, how `blobstore-fs` being broken on Windows was uncovered by this PR)
- use consistent provider naming scheme
- simplify dependency management (which should, in fact, improve our Nix caching performance)
- update `nixify` to pull in fix for https://github.com/rvolosatovs/nixify/pull/220
- run provider unit tests (which were ignored up to this point) - the broken ones are ignored for now

It's still possible to build wasmcloud with `--no-default-features` to avoid building the providers and only build the host binary.


thomastaylor312 (approved):
Way nicer. Great job here

PR #1703, opened 2024-03-19
chore: update `test-actors` dependencies

rvolosatovs:


rvolosatovs:
Integrated in #1736

brooksmtownsend (approved):
I'm fine with all these changes especially since these are test components, and I do want to move away from all usages of "actor" in the future where possible. Do you think that using "Component" as a struct name is too confusing? I think we can help people understand what the best name is to use for their structs

PR #1705, opened 2024-03-19
ci: release providers

rvolosatovs:
## Feature or Problem

build and release PARs in CI. Publish `main` and tag builds to GHCR and AzureCR.

Example tag build available at https://github.com/wasmCloud/wasmCloud/actions/runs/8362687944 (`provider-blobstore-fs-v0.6.0-rc.1`)

https://github.com/wasmCloud/wasmCloud/pkgs/container/blobstore-fs/193605679?tag=0.6.0-rc.1


.github/workflows/wasmcloud.yml
```diff
@@ -239,6 +259,58 @@ jobs:
...
+      - build-windows
+      - test-linux
+    permissions:
+      packages: write
+    uses: ./.github/workflows/provider.yml
+    with:
+      name: ${{ matrix.name }}
+      capability: ${{ matrix.capability }}
+    secrets:
+      issuer: ${{ secrets.WASMCLOUD_ISSUER }}
```
thomastaylor312, line 309: This should be `WASMCLOUD_ACCOUNT_OFFICIAL` to pick up on the org level secret that contains the issuer

.github/workflows/provider.yml
```diff
@@ -0,0 +1,168 @@
...
+    - run: chmod +x "./artifacts/wasmcloud-aarch64-unknown-linux-musl/bin/${{ inputs.name }}"
+    - run: chmod +x "./artifacts/wasmcloud-x86_64-apple-darwin/bin/${{ inputs.name }}"
+    - run: chmod +x "./artifacts/wasmcloud-x86_64-pc-windows-msvc/bin/${{ inputs.name }}.exe"
+    - run: chmod +x "./artifacts/wasmcloud-x86_64-unknown-linux-musl/bin/${{ inputs.name }}"
+
+    - run: mv "./artifacts/wasmcloud-x86_64-unknown-linux-musl/bin/wash" wash
+    - run: chmod +x wash
+
+    - run: |
+        if [ "${{ secrets.issuer }}" != '' ]; then
```
thomastaylor312, line 66: Nit: Might just be easier to have these be set as env variables for this step rather than exporting in the script

thomastaylor312 (changes requested):

thomastaylor312:
I've added all the necessary secrets I believe. So this should be good to go once we merge

.github/workflows/provider.yml
rvolosatovs: That was my initial approach as well, however having these set to empty string breaks `wash` unfortunately, so the only way to support key autogeneration here is to check for the value explicitly before setting the env

thomastaylor312: We wouldn’t want it to autogenerate keys though if they’re not set, right?

thomastaylor312 (approved):
Good to go but one more clarifying question I left in a previous thread

.github/workflows/provider.yml
rvolosatovs: That would prevent this from running on forks, meaning that this would never run on PRs from forks. I do think that we want to build PARs always, at least for the convenience of testing proposed changes, but also to ensure a PR does not break this process

thomastaylor312: Yeah I can definitely get behind that

PR #1723, opened 2024-03-20
feat(blobstore-s3)!: convert to `wrpc:blobstore`

rvolosatovs:
## Feature or Problem

- Convert `blobstore-s3` to `wrpc:blobstore`
- Implement `move-object` and `copy-object`
- Significantly increase blobstore test coverage

### Manual Verification

Tested this with a real AWS s3 bucket


rvolosatovs:
integrated in #1736

crates/provider-blobstore-fs/src/lib.rs
```diff
@@ -924,7 +925,9 @@ impl FsProvider {
                         .context("failed to resolve destination container path")?;
                     let dest = resolve_subpath(&dest_container, dest.object)
                         .context("failed to resolve destination object path")?;
+                    debug!("copy `{}` to `{}`", src.display(), dest.display());
```
thomastaylor312, line 928: Nit: These should probably be named fields instead of formatted in the message

crates/provider-blobstore-fs/src/lib.rs
```diff
@@ -924,7 +925,9 @@ impl FsProvider {
...
                     fs::copy(&src, dest).await.context("failed to copy")?;
+                    debug!("remove `{}`", src.display());
```
thomastaylor312, line 930: Nit: Same here with named fields

thomastaylor312 (approved):

Issue #1658, opened 2024-03-12
[FEATURE] Unify `wash start actor|component` and `wash start provider` into a single `wash start`


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Right now we have a nesting of `wash start` commands, but it would be awesome to be able to simply `wash start ./my_component.wasm` or `wash start my.oci.ref/repo:v1` instead of having to specify the type of component. We have enough information based off of the reference (see if it's a local file or not) and we can inspect it in order to discover the artifact type.

### Describe the solution you'd like

I would like `wash start` to intelligently discover a local file, or inspect an OCI reference's type (actor component or provider) in order to auction the resource and then subsequently start the resource. Because components must be scaled, we can simply scale with a `max-instances` of 1.

ahmedtadde:
hey @brooksmtownsend, I am eyeing this and I have some questions:
- can we deterministically distinguish between an actor component and provider component by probing a `DecodedWasm::resolve()` or `DecodeWasm::package()`. If yes, what's the differentiation prop(s) to look for. 
- If the answer to the previous question is no... then, would `ProviderArchive::try_load(artifact_bytes)` suffice to check that we are dealing with a provider component? if that fails, we check for token claims as way to assert that we are dealing with an actor component?


I mainly scanned the code for the `InspectCommand` thinking it would have some tidbits that help with differentiation between the two kinds of components. If there is somewhere else that you believe/intuit that I should look for such an insight, please do let me know.

thomastaylor312:
As discussed in #1725, this feature probably won't work out, so closing this issue

PR #1725, opened 2024-03-20
[FEATURE] Unify wash start actor|component and wash start provider into a single wash start cmd

ahmedtadde:
## Feature or Problem
Right now we have a nesting of wash start commands, but it would be awesome to be able to simply wash start ./my_component.wasm or wash start my.oci.ref/repo:v1 instead of having to specify the type of component. We have enough information based off of the reference (see if it's a local file or not) and we can inspect it in order to discover the artifact type.

## Related Issues
#1658 

## Consumer Impact
- quality of life improvement for the `wash` cli

### Unit Test(s)
- added [fn test_start_comprehensive() -> Result<()>](https://github.com/ahmedtadde/wasmCloud/blob/2f986829508907d224b7e1af941aa504e212fe3e/crates/wash-cli/src/common/start_cmd.rs#L66)
- updated [fn test_ctl_comprehensive() -> anyhow::Result<()>](https://github.com/ahmedtadde/wasmCloud/blob/2f986829508907d224b7e1af941aa504e212fe3e/crates/wash-cli/src/ctl/mod.rs#L89)

### Manual Verification
- updated [async fn integration_start_stop_actor_serial() -> Result<()>](https://github.com/ahmedtadde/wasmCloud/blob/2f986829508907d224b7e1af941aa504e212fe3e/crates/wash-cli/tests/wash_start.rs#L13) and [async fn integration_start_stop_provider_serial() -> Result<()>](https://github.com/ahmedtadde/wasmCloud/blob/2f986829508907d224b7e1af941aa504e212fe3e/crates/wash-cli/tests/wash_start.rs#L34)


thomastaylor312:
Hey @ahmedtadde, looks like there might have been some confusion as to what we should do here from the issue. So I think the goal was to just have something that looks like `wash start <path to provider>` and `wash start <path to component>`. I think that is what you were getting at with reusing some of the inspect code as you said in your comment [here](https://github.com/wasmCloud/wasmCloud/issues/1658#issuecomment-2005107689). 

So I think the goal is to just inspect the file or to try and load it as a component (the thing formerly known as an actor) and then fall back to loading as a provider if that fails. Does that make more sense or should we go back and discuss some more on that issue?

ahmedtadde:
Thanks for the feedback @thomastaylor312 

So do we want to keep the existing commands as well? wash start actor/provider? If the answer is yes, I struggled to make that work via Clap... basically defining the actor, provider subcommands but also having be optional/superseded by the case when the user doesn't specify either. Admittedly, I believe this is almost definitely a skill issue.

thomastaylor312:
Nope, I don't think we need to keep the old commands. I think it is all just `wash start` when this is done. We're ok with breaking this here

crates/provider-lattice-controller/README.md
```diff
@@ -33,7 +33,7 @@ async fn start_component(ctx: &Context) -> RpcResult<CtlOperationAck> {
         user_seed: None
     }).await;
 
-    let cmd = StartComponentCommand {
+    let cmd = StartActorCommand {
```
thomastaylor312, line 36: This should be `StartComponentCommand` still

crates/wash-lib/src/cli/start.rs
```diff
@@ -72,9 +73,111 @@ pub struct StartComponentCommand {
...
+
+    let provider = ProviderArchive::try_load(&buf).await;
+
+    if provider.is_ok() {
+        return handle_start_provider(cmd)
+            .await
+            .with_context(|| format!("Failed to start provider component: {}", component_ref));
+    }
+
+    let actor = match wasmparser::Parser::new(0).parse_all(&buf).next() {
```
thomastaylor312, line 138: I think components are probably going to be the most common thing to start, so it might be better to try parsing as a component first and then a provider. Thoughts?

crates/wash-lib/src/cli/start.rs
```diff
@@ -72,9 +73,111 @@ pub struct StartComponentCommand {
...
+    #[clap(long = "insecure")]
+    pub insecure: bool,
+
+    /// skip the local OCI cache and pull the artifact from the registry to inspect
+    #[clap(long = "no-cache")]
+    pub no_cache: bool,
 }
 
-pub async fn handle_start_component(cmd: StartComponentCommand) -> Result<CommandOutput> {
+pub async fn handle_start_component(cmd: StartCommand) -> Result<CommandOutput> {
```
thomastaylor312, line 102: Let's call this `handle_start` and have `handle_start_component` still be the way to start what was used to be called an "actor." We've moved away from using the actor terminology and just call them "components" now, since that is what they are

crates/wash-lib/src/cli/start.rs
```diff
@@ -72,9 +73,111 @@ pub struct StartComponentCommand {
...
+
+    if provider.is_ok() {
+        return handle_start_provider(cmd)
+            .await
+            .with_context(|| format!("Failed to start provider component: {}", component_ref));
+    }
+
+    let actor = match wasmparser::Parser::new(0).parse_all(&buf).next() {
+        // Inspect claims inside of Wasm
+        Some(Ok(wasmparser::Payload::Version {
```
thomastaylor312, line 140: We don't need to do any of this claims inspection as it will be done by the host (plus claims are optional now). I would just check if parses like you do here and then continue on

crates/wash-cli/tests/common/mod.rs
```diff
@@ -314,7 +313,7 @@ impl TestWashInstance {
             .context("failed to parse output of `wash start actor`")
     }
 
-    /// Trigger the equivalent of `wash start provider` on a [`TestWashInstance`]
+    /// Trigger the equivalent of `wash start` on a [`TestWashInstance`]
     pub(crate) async fn start_provider(
```
thomastaylor312, line 317: We can just get rid of this now and use the new function you created

crates/wash-cli/tests/common/mod.rs
```diff
@@ -342,6 +340,33 @@ impl TestWashInstance {
             .context("failed to parse output of `wash start provider`")
     }
 
+    /// Trigger the equivalent of `wash startn` on a [`TestWashInstance`]
+    pub(crate) async fn start_component(
```
thomastaylor312, line 344: This should probably just be called `start`

crates/wash-cli/tests/common/mod.rs
```diff
@@ -287,7 +287,7 @@ impl TestWashInstance {
         })
     }
 
-    /// Trigger the equivalent of `wash start actor` on a [`TestWashInstance`]
+    /// Trigger the equivalent of `wash start` on a [`TestWashInstance`]
     pub(crate) async fn start_actor(
```
thomastaylor312, line 291: We can just get rid of this now and use the new function you created

thomastaylor312 (changes requested):
This is pretty close, just a few points around naming and one discussion item

crates/wash-lib/src/cli/start.rs
```diff
@@ -72,9 +71,93 @@ pub struct StartComponentCommand {
...
+        f.read_to_end(&mut buf)?;
+    } else {
+        let cache_file = (!cmd.no_cache).then(|| cached_oci_file(&cmd.component_ref.clone()));
+        buf = get_oci_artifact(
+            component_ref.clone(),
+            cache_file,
+            OciPullOptions {
+                digest: cmd.digest.clone(),
+                allow_latest: cmd.allow_latest,
+                user: cmd.user.clone(),
```
thomastaylor312, line 120: @brooksmtownsend I just realized this on my second review: this could be a little gnarly if you’re using a private registry. For the command to work you’d need to have creds. Should we have a helper flag that indicates what it is and skips the smart check?

thomastaylor312:
@ahmedtadde this is all good to go but I realized one thing on my rereview we should decide how we want to handle. @brooksmtownsend is out this week so we might wait a few days if that is ok. 

Sorry to delay this and thanks for all the hard work!

crates/wash-lib/src/cli/start.rs
brooksmtownsend: @thomastaylor312 maybe I don't understand, are you saying that the smart check wouldn't work if a private registry is used?

brooksmtownsend:
@ahmedtadde my apologies for the big delay here, just asking about a question Taylor asked but this all looks good!

brooksmtownsend:
Since it's been a few weeks @ahmedtadde it might be worth a rebase on `main` here, GitHub doesn't show any conflicts but just in case

crates/wash-lib/src/cli/start.rs
thomastaylor312: Imagine a normal deployment strategy in an enterprise. You have an internally accessible registry, generally something like ACR or ECR that your deployment target (kubernetes, bare VMs, etc). Those nodes/VMs are configured to be able to access those registries, but a normal developer might not have access to them, or would have to set credentials. If you don't set credentials, this command will fail.

Contrast that to how it works right now. `wash` doesn't attempt to pull anything down, it just sends the reference to the host, which then tries to pull it. Does that make any sense?

brooksmtownsend: Yeah that makes sense now. Basically there's a flaw in assuming that the resources a host has access to is the same set that the user or operator does with `wash` (similarly, it could be the opposite relationship where `wash` has access but wasmCloud doesn't but that's a different problem.)

It sounds like a helper flag that allows you to specify to skip the check and to tell `wash` that the resource is a component/provider would be the best solution, users would rarely need to use it but it's there just in case. Thoughts @thomastaylor312 ?

Strawman, something like `--type component|provider`? In the help text we can indicate that this skips the initial pull

thomastaylor312: I thought about that too, but my other concern is with bandwidth/download time. With providers in particular they can be quite large so there is a noticeable delay while it gets downloaded and you're essentially downloading twice (once locally and once on the host)

thomastaylor312:
@ahmedtadde curious if you have any ideas here. Right now as things stand, this feels like it would end up adding some rough DevEx edges to cut yourself on, even though we all thought it was a good idea to start with. It basically boils down to two things:

- If you use a private registry, you'd need to have creds set or pass a special flag to bypass the check (which would lead to a different command for someone doing local dev from someone doing it in prod)
- This could add significant times for downloading providers. For example, if it takes 30s to download a big provider, then it will take 30s to download locally + 30s to download on the host, so a total of 1 minute. Even if we bypass that with some tricks like inspecting the OCI manifest, we'd still have to have logic to handle the case from loading a file, which also feels gross

Can you think of any other ways we could solve this or should we abandon this for now?

ahmedtadde:
@thomastaylor312 

- so short answer is "this should be canned for now". those two reasons are pretty solid. Especially the second one. 
- long answer
   - can we have this new implementation fail for non-local component refs (ie, if component ref is not a `file://...` type)? we would require a flag `--type component|provider` like @brooksmtownsend suggested when the component ref is a URI to some remote repository.
   - can we somehow make the component auction agnostic... each suitable host can include some information on whether or not the slot it's offering is component/actor or provider specific?

thomastaylor312:
> can we have this new implementation fail for non-local component refs (ie, if component ref is not a file://... type)? we would require a flag --type component|provider like @brooksmtownsend suggested when the component ref is a URI to some remote repository.

I actually brainstormed a similar idea with @brooksmtownsend but it didn't get anywhere that sounded great. I think it was really around the command being entirely different from doing it in local dev. So I think we probably cancel this for now is what I'm leaning towards

ahmedtadde:
I am fully onboard with the decision to cancel this for now. I will go ahead and close the PR.

PR #1730, opened 2024-03-21
build: build Rust examples as part of test

rvolosatovs:
## Feature or Problem

Build Rust examples as part of normal testing procedure, while keeping the examples in their own workspaces

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -898,6 +898,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "blobby"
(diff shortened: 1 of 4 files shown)


thomastaylor312 (commented):
Love the idea of doing this and it is a neat Cargo trick if we can get it to work. Will approve once out of draft

rvolosatovs:
After thinking a bit about this, let's just update `test_actors` and write unit tests like in #1929

Issue #1672, opened 2024-03-14
[BUG] kubernetes wasmCloud host does not show up in local washboard


Fasibru:
### Affected project(s)

- [X] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [X] other / not sure

### Describe the bug

When deploying wasmCloud via helm chart you cannot right away open a local washboard and see the kubernetes host. The [documentation](https://github.com/wasmCloud/wasmCloud/tree/main/chart#running-the-chart) only mentions to forward port 4222 which is not enough.

### Steps to reproduce

1. Deploy helm chart on e.g. local k8s cluster
2. Follow documentation to forward port 4222
3. Start `wash ui`
4. Connection to NATS not working (and k8s wasmCloud host not showing up)

### Expected behavior

After deploying the helm chart and following the documentation the washboard should show the k8s wasmCloud host.

### Environment

- OS:  MacOS
- Shell: zsh
- wasmcloud Version 0.81.0
- helm Chart Version 0.7.2

thomastaylor312:
If I were to take a guess, you'll have to enable websockets on the NATS server, which we don't have an option for right now in the chart as far as I know. So we'll have to fix that or figure out a better way to do this by starting the washboard in the chart

Fasibru:
> If I were to take a guess, you'll have to enable websockets on the NATS server, which we don't have an option for right now in the chart as far as I know. So we'll have to fix that or figure out a better way to do this by starting the washboard in the chart

That's what I was tackling in the PR https://github.com/wasmCloud/wasmCloud/pull/1735. I have essentially added websockets support and wadm to the helm chart. But I wonder now if we should really put effort into the current helm chart or focus on a helm-chart for the operator and potentially wadm?

thomastaylor312:
Yeah I think focusing on support for the operator would be better. 

Before I close however, I think @lachieh had some ideas about doing something that doesn’t involve web sockets as well so it plays nicer with the operator. Any additional details to add here @lachieh

lachieh:
You're correct in that the washboard needs websocket support. We've been discussing moving the washboard NATS connection to the server side so that the client can just interact with the server and not have to worry about enabling websockets on the NATS server. That move would solve this, _and_ it would allow the washboard to run inside of k8s.

[...]

PR #1735, opened 2024-03-22
WIP: feat: improves helm charts to allow better integration with kubernetes

Fasibru:
## Feature or Problem

At the moment the documentation for the helm chart is misleading in terms of what to do to access the Kubernetes wasmCloud and NATS deployment from your local machine and local washboard. To make this experience smoother I have added the possibility to configure and expose websocket for NATS and also the possibility to deploy wadm with the helm chart.

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/1672

## Release Information

next

## Consumer Impact

The added configuration possibilities should only be used for testing, not production.

### Manual Verification

I have deployed the helm chart in a local Kubernetes cluster and made sure that applications can be successfully deployed to the Kubernetes wasmCloud host and that the washboard shows the expected host, components, providers and links.

### DIFF
--- a/chart/templates/configmap.yaml
+++ b/chart/templates/configmap.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.nats.leafnode.enabled -}}
+{{- if or .Values.nats.leafnode.enabled .Values.nats.insecureWebsocket.enabled -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -7,6 +7,7 @@ metadata:
     {{- include "wasmcloud_host.labels" . | nindent 4 }}
 data:
   nats.conf: |-
+    {{ if .Values.nats.leafnode.enabled -}}
     {{ if .Values.nats.jetstreamDomain -}}
     jetstream {
       domain: {{ .Values.nats.jetstreamDomain }}
@@ -20,14 +21,17 @@ data:
         },
       ]
     }
----
-apiVersion: v1
-kind: Secret
-metadata:
-  name: {{ include "wasmcloud_host.fullname" . }}-nats
-  labels:
-    {{- include "wasmcloud_host.labels" . | nindent 4 }}
-data:
-  nats.creds: |-
-    {{ required "Credentials are required for a leafnode" .Values.nats.leafnode.credentials | b64enc }}
+    {{- end }}
+    {{- if and (not .Values.nats.leafnode.enabled) .Values.nats.insecureWebsocket.enabled -}}
+    port: 4222
+    monitor_port: 8222
+    server_name: "wasmcloud"
+
+    jetstream {}
+
+    websocket {
+      port: 4001
+      no_tls: true
+    }
+    {{- end }}
 {{- end }}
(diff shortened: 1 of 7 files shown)


chart/templates/secret.yaml
```diff
@@ -21,4 +21,16 @@ data:
...
+{{- if .Values.nats.leafnode.enabled -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "wasmcloud_host.fullname" . }}-nats
+  labels:
+    {{- include "wasmcloud_host.labels" . | nindent 4 }}
+data:
+  nats.creds: |-
+    {{ required "Credentials are required for a leafnode" .Values.nats.leafnode.credentials | b64enc }}
```
Fasibru, line 35: Secret was moved from former `leafnode.yaml`.

chart/templates/configmap.yaml
Fasibru, line 1: I have adjusted the whole configmap logic in a way that if nats.leafnode.enabled: true then the websocket settings are ignored. The reason is simply that I don't know if these two settings (leafnode and websocket) work properly together. I would need some support to discuss how to maybe reduce the control loops but still have a proper setup for leafnode and websocket.

chart/templates/deployment.yaml
Fasibru, line 1: See the comment on the configmap. I implemented websocket support but it only has an effect if the leafnode is not enabled.

Fasibru (commented):
I marked this PR since I need some help to figure out how leafnode and websocket can work together. If they can work together then a few of the if-statements can be reduced.

Fasibru:
I also saw that https://github.com/wasmCloud/wasmCloud/blob/main/chart/templates/NOTES.txt is misleading and I'm not sure what the intention here is. Should it tell users how to access the washboard or how to access an application that is deployed to the Kubernetes wasmCloud host? Either way it needs improvements.

And I was wondering if it wouldn't make more sense to make wasmCloud and Nats individual deployments (like wadm in this PR)? This would help to better structure the `values.yaml` file and would allow to just scale wasmcloud hosts independently of nats servers.
I know that my focus is quite Kubernetes-heavy and I know that wasmCloud is absolutely independent of k8s. So I'm wondering if there is a need/wish from your side to actually make the integration with k8s easier (this would mean updates to the docs like quickstart and updates to the helm charts that allow a production usage in k8s) or if you would like to explicitly not do that?

thomastaylor312:
Thank you so much for adding this in @Fasibru! So to answer your question, we finally open sourced the wasmcloud operator: https://github.com/wasmcloud/wasmcloud-operator. This is probably going to be our standard way of integrating wasmcloud into Kubernetes. What that does need though is a Helm chart to install the CRDs and operators for people to use.

With that in mind, is that something you'd prefer to use or would you still like to have a chart like this one that installs all the different components and dependencies for you?

Fasibru:
Thanks for getting back to me @thomastaylor312. The operator looks cool and if that is the preferred way of deploying wasmCloud hosts to k8s then it just makes sense to invest time there to write some helm charts. This would mean that in the future there are helm charts for wasmcloud-operator and nats, but what about wadm?

thomastaylor312:
There is already an open PR for a wadm chart (https://github.com/wasmCloud/wadm/pull/248), and I think @protochron might have plans to make the wadm part easier as part of the operator? Either way there will be options

PR #1748, opened 2024-03-25
feat(*): remove capability signing

vados-cosmonic:
## Feature or Problem

This series of commits removes capability signing from various crates.

Since wasmCloud is now ideally WIT-driven, we do not need to specify or sign our providers or actors with capabilities and can deduce them from the related WIT contracts as necessary.


crates/host/src/wasmbus/mod.rs
```diff
@@ -1317,7 +1317,7 @@ impl ComponentSpecification {
 
 #[allow(clippy::large_enum_variant)] // Without this clippy complains actor is at least 0 bytes while provider is at least 280 bytes. That doesn't make sense
 enum Claims {
-    Actor(jwt::Claims<jwt::Actor>),
+    Actor(jwt::Claims<jwt::Component>),
```
thomastaylor312, line 1320: Should we rename this variant to Component while we're here?

crates/runtime/src/actor/mod.rs
```diff
@@ -6,23 +6,27 @@ pub use component::{
...
     pub require_signature: bool,
 }
 
-/// Extracts and validates claims contained within `WebAssembly` binary, if such are found
-fn claims(wasm: impl AsRef<[u8]>) -> anyhow::Result<Option<jwt::Claims<jwt::Actor>>> {
+/// Extracts and validates claims contained within a WebAssembly binary, if present
+///
+/// # Arguments
+///
+/// * `wasm` - Bytes that consitute a valid WebAssembly binary
```
thomastaylor312, line 24: Spelling: constitute

thomastaylor312 (commented):
Overall the code looks great. However, I do have a question around providers. Providers might have wit that they use to generate their code, but they don't carry wit with them like a component. So how is a provider supposed to indicate what it is allowed to do?

crates/host/src/wasmbus/mod.rs
vados-cosmonic: Ah yes, will rename!

- [ ] rename `Claims::Actor` variant

crates/runtime/src/actor/mod.rs
vados-cosmonic: - [ ] fix spelling in rustdoc

vados-cosmonic:
> However, I do have a question around providers. Providers might have wit that they use to generate their code, but they don't carry wit with them like a component.

So it's possible for a provider to carry around the WIT that it was generated with, but IIRC this was argued against for the initial implementation of wRPC. There's no reason we can't have this information if we want. 

> So how is a provider supposed to indicate what it is allowed to do?

Right now, the invocation lookup would be the only way -- it's parametrized with a lattice name, target ID and wRPC version, but you can find out exactly what world key names and functions are used.

That said, we're actually not restricting this in any way right now, so a plan on what/how to restrict probably makes sense before we add that in.

thomastaylor312 (approved):

rvolosatovs (approved):

Issue #424, opened 2023-08-02
Adopt `bors` or similar


rvolosatovs:
Using a merge bot like https://bors.tech/ would let us avoid accidentally breaking `main`

refs https://github.com/wasmCloud/wasmCloud/pull/423

rvolosatovs:
> Looks like Bors actually recommends using the GH merge queue feature on their main page: https://github.blog/2023-07-12-github-merge-queue-is-generally-available/

It's a new feature I have not had a chance to try yet, could you enable it?

thomastaylor312:
Went to turn it on and it informed me of this change we need to do to our triggers: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

Once we do that we can turn on the merge queue

vados-cosmonic:
Since we now have merge queues enabled, we probably don't need this anymore! Let's re-open if we ever decide to switch to a smarter bot for merging PRs/doing other tasks.

PR #1749, opened 2024-03-25
fix(blobstore-fs): Fixes missing import

thomastaylor312:
Not quite sure how the original PR passed tests, but we were missing an import. This should fix the issue


ricochet (approved):

rvolosatovs (approved):

rvolosatovs:
> Not quite sure how the original PR passed tests, but we were missing an import. This should fix the issue

That's a known drawback of the way GitHub handles merges and CI - since the PR was based on a version of `main`, where `trace` was imported - CI passed, but since then `main` changed and `trace` import was removed, however the CI was not run on that new `main` state.

The only way to avoid these kind of issues is to use a merge bot or the merge queue feature https://github.com/wasmCloud/wasmCloud/issues/424

PR #1757, opened 2024-03-26
refactor(provider-sdk): move wasmbus RPC topic generation to core

vados-cosmonic:
## Feature or Problem

This commit moves the topic generation functions that were used for wasmbus RPC topics from `provider-sdk` to `core` so that they can be used/referred to more widely.

### DIFF
--- a/crates/core/src/rpc.rs
+++ b/crates/core/src/rpc.rs
@@ -12,3 +12,34 @@ pub struct HealthCheckResponse {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub message: Option<String>,
 }
+
+/// Generate the wasmbus RPC subject for putting links on a NATS cluster
+///
+/// When messages are published on this subject, hosts set up and update (if necessary) link information,
+/// which may include calling `receive_link_config_*()` functions on relevant providers.
+pub fn link_put_subject(lattice: &str, provider_key: &str) -> String {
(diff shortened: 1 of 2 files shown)


crates/core/src/rpc.rs
```diff
@@ -12,3 +12,47 @@ pub struct HealthCheckResponse {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub message: Option<String>,
 }
+
+/// Generate the wasmbus RPC subject for putting links on a NATS cluster
+///
+/// When messages are published on this subject, hosts set up and update (if necessary) link information,
+/// which may include calling `receive_link_config_*()` functions on relevant providers.
+pub fn link_put_subject(lattice: impl AsRef<str>, provider_key: impl AsRef<str>) -> String {
```
thomastaylor312, line 20: @rvolosatovs Curious on your opinion here. I've generally been told that `&str` is the better way to go for public APIs (I think it is due to dereferencing being nice). However, I do know the `AsRef` gives you a lot of flexibility. It has been a situational choice for me before, so I just wanted a second opinion

thomastaylor312: Let's just keep it as `&str` since we have a pretty good idea that most types are going to be easily referenceable as a str

vados-cosmonic: Done!

thomastaylor312 (approved):

Issue #1721, opened 2024-03-20
[BUG] rust actor template does not work


fliepeltje:
### Affected project(s)

- [X] documentation
- [X] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

Using the quickstart, I cannot create a rust actor.
The output I get is

```bash

🔧   Cloning template from repo wasmCloud/wasmCloud subfolder examples/rust/actors/http-hello-world...

Failed to generate project

Caused by:
    ⛔   Invalid subfolder path: No such file or directory (os error 2)

```

Seems like the directory should point to `examples/rust/components/http-hello-world` instead of `examples/rust/actors/http-hello-world`

### Steps to reproduce

1. run `wash new actor hello --template-name hello-world-rust`

### Expected behavior

I expected a project directory to be created.

### Environment

- OS: Linux
- Shell zsh

t3hmrman:
Hey @fliepeltje would you mind confirming that `wash` works for you now? The fix has been merged to `main` so it should be usable now

vados-cosmonic:
Hey @fliepeltje would you mind confirming that `wash` works for you now? The fix has been merged to `main` so it should be usable now

fliepeltje:
> Hey @fliepeltje would you mind confirming that `wash` works for you now? The fix has been merged to `main` so it should be usable now

I have veriefied it and it indeed works as intended! :+1:

PR #1758, opened 2024-03-26
fix(wash-lib): restore rust examples folder

vados-cosmonic:
## Feature or Problem

Previous reshuffling of the `examples` directory has broken `wash new`, and while the newer (unreleased) code in `wash-lib` has the right paths, the currently released version of `wash` (and `wash-lib`, in particular `0.18.1`) is broken.

This commit adds a soft link to fix the problem temporarily before a new version is out.

Resolves https://github.com/wasmCloud/wasmCloud/issues/1721


thomastaylor312 (approved):
Looks fine to me. I can't remember how links pick up in git, but this looks like it added all of the files

vados-cosmonic:
@thomastaylor312 Ah a bit late but that comment in stale, will update it -- softlinks don't work with the way wash sets up the template, had to actually copy the real files

PR #1761, opened 2024-03-26
chore(runtime): reenable `clippy::pedantic`

rvolosatovs:
## Feature or Problem

Reenable `clippy::pedantic` in the runtime and address warnings introduced in the time it was missing

## Related Issues

fixup https://github.com/wasmCloud/wasmCloud/commit/42d069eee87d1b5befff1a95b49973064f1a1d1b, where this clippy lint was removed with no explanation

### DIFF
--- a/crates/runtime/src/actor/component/mod.rs
+++ b/crates/runtime/src/actor/component/mod.rs
@@ -470,16 +470,14 @@ fn instantiate(
             _ => {}
         }
 
-        let item = match item {
-            component::types::ComponentItem::ComponentInstance(item) => item,
-            _ => continue,
+        let component::types::ComponentItem::ComponentInstance(item) = item else {
+            continue;
         };
         let exports = item.exports(engine);
         let mut instance = HashMap::with_capacity(exports.len());
         for (func_name, item) in exports {
-            let ty = match item {
-                component::types::ComponentItem::ComponentFunc(ty) => ty,
-                _ => continue,
+            let component::types::ComponentItem::ComponentFunc(ty) = item else {
+                continue;
             };
             instance.insert(func_name.to_string(), ty.results().collect());
         }
@@ -604,6 +602,7 @@ impl Component {
     /// Returns a map of dynamic function export types.
     /// Top level map is keyed by the instance name.
     /// Inner map is keyed by exported function name.
+    #[must_use]
     pub fn exports(&self) -> &Arc<HashMap<String, HashMap<String, DynamicFunction>>> {
         &self.exports
     }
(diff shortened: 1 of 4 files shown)


thomastaylor312 (changes requested):
I did this change explicitly because pedantic is well...pedantic, which often discourages new contributors. Most current contributors found most of these lints annoying when contributing. I also did check some of the more popular Rust projects and none of them use pedantic clippy so I don't see a lot of previous examples of even bigger and more mature projects using it.

I'd be more in favor of this if we can talk about how it can help out the community and new contributors, but I just don't see similar efforts across other projects. So the annoyance to help ratio seems a bit off for me. Curious if @brooksmtownsend has thoughts here as well

thomastaylor312:
To add some additional context, I'd be ok with maybe adding some of the individual lints from pedantic if we think they are important for our codebase, but I don't know that I want to enable all of them

rvolosatovs:
> I did this change explicitly because pedantic is well...pedantic, which often discourages new contributors. Most current contributors found most of these lints annoying when contributing. I also did check some of the more popular Rust projects and none of them use pedantic clippy so I don't see a lot of previous examples of even bigger and more mature projects using it.
> 
> I'd be more in favor of this if we can talk about how it can help out the community and new contributors, but I just don't see similar efforts across other projects. So the annoyance to help ratio seems a bit off for me. Curious if @brooksmtownsend has thoughts here as well

A few examples of bigger projects using this lint group are:
- firecracker https://github.com/search?q=repo%3Afirecracker-microvm%2Ffirecracker+pedantic&type=code
- clap https://github.com/search?q=repo%3Aclap-rs%2Fclap%20pedantic&type=code
- ruby parser https://github.com/search?q=repo%3Aruby%2Fprism%20pedantic&type=code
- sqlx https://github.com/search?q=repo%3Alaunchbadge%2Fsqlx%20pedantic&type=code
- clippy itself

> To add some additional context, I'd be ok with maybe adding some of the individual lints from pedantic if we think they are important for our codebase, but I don't know that I want to enable all of them

In general sense I agree, but I think there is not one-size-fits-all solution here, rather I think there should be a tiered approach.
Runtime is a crucial part of the project, where IMO code quality should be very high and every little bit counts. I do think runtime should use this lint. Outside contributions to this crate are also extremely rare and generally unlikely.
Any Wash-related crate? We certainly should not use it there.

I think:
- `clippy::pedantic` should be enabled in the runtime crate
- Some of the lints from `clippy::pedantic` should be enabled in the host crate
- Other crates should probably just use the defaults

thomastaylor312 (approved):

Issue #1433, opened 2024-02-02
[BUG] "musl" based wasmcloud host fails to load providers and actors from remote due to dns resolution error


wamli:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

The wasmcloud host which I use, **wasmcloud-x86_64-unknown-linux-musl (v0.81.0)**, cannot locate entities in remote registries due to a dns resolution error:

```
$ wash up --nats-websocket-port 4001
🏃 Running in interactive mode.
🎛️  If you enabled --nats-websocket-port, start the dashboard by executing `wash ui`
🚪 Press `CTRL+c` at any time to exit
2024-02-01T09:51:45.826587Z  INFO async_nats::options: event: connected
2024-02-01T09:51:45.827320Z  INFO async_nats::options: event: connected
[..]
Caused by:
    0: failed to fetch provider under OCI reference `wasmcloud.azurecr.io/httpserver:0.19.1`
    1: failed to fetch OCI path
    2: failed to fetch OCI manifest
    3: error sending request for url (https://wasmcloud.azurecr.io/v2/): error trying to connect: dns error: failed to lookup address information: Name does not resolve
    4: error trying to connect: dns error: failed to lookup address information: Name does not resolve
    5: dns error: failed to lookup address information: Name does not resolve
    6: failed to lookup address information: Name does not resolve
```
**Interesting observation:** `wash pull wasmcloud.azurecr.io/httpserver:0.19.1` is totally relaxed and pulls it.

### Steps to reproduce

1. `wash up`
2. `wash app deploy wadm.yaml` where "wadm.yaml" can be any app which will try to load a provider and/or an actor from a remote registry, e.g. not localhost.

### Expected behavior

My expectation was that artifacts like **wasmcloud.azurecr.io/httpserver:0.19.1** should be downloaded without problem.

### Environment

- OS: **Linux (Ubuntu)**
- Shell: **bash**
- wasmcloud Version: **v0.81.0**
- deployment: **"local"**, e.g. NOT in a container

brooksmtownsend:
Thank you @wamli !
[...]

PR #1772, opened 2024-03-27
feat: use native TLS roots along webpki

rvolosatovs:
## Feature or Problem

Include both webpki and native OS root certificates by default for TLS. This ensures that outgoing connections work *everywhere* - be it a `FROM scratch` Docker container, consumer OS behind a corporate firewall, an embedded device or anything in-between.

Eventually we should standardize TLS setup in providers and have a way to switch between the CA pools, but for now we should focus on getting things working.

~The specified `reqwest` feature set ensures both bundles are included https://github.com/seanmonstar/reqwest/blob/14e46ff8cb7550473c950f3471d049ad2e139d3f/src/async_impl/client.rs#L501-L529~ this fails when the OS bundle is missing, so simply load the native certs ourselves and log if we failed to load the native bundle

## Related Issues

May fix - https://github.com/wasmCloud/wasmCloud/issues/1433

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2331,20 +2331,19 @@ dependencies = [
 
 [[package]]
 name = "hyper-rustls"
-version = "0.26.0"
+version = "0.27.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a0bea761b46ae2b24eb4aef630d8d1c398157b6fc29e6350ecf090a0b70c952c"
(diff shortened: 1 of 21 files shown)


thomastaylor312 (commented):
I think this approach is straightforward and unblocks any cert issue we're seeing, so once this is ready for final review I can go ahead and approve

thomastaylor312 (approved):

PR #1774, opened 2024-03-27
feat(components): add component-echo-http

vados-cosmonic:
## Feature or Problem

This commit adds a wasi:http compliant echo component to replace the echo
example (https://github.com/wasmCloud/examples/tree/main/actor/echo) that existed in previous versions of wasmCloud.

Adding this component should unblock other tasks that rely on a published component that can be used with wasmCloud 1.0.


crates/component-echo-http/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+        };
+
+        // Build the outgoing response
+        let resp = OutgoingResponse::new(Fields::new());
+        let body = resp.body().expect("failed to open outgoing response body");
+        let out = body
+            .write()
+            .expect("failed to start writing to outgoing response body");
+        out.blocking_write_and_flush(&resp_bytes)
+            .expect("failed to complete blocking write of request");
```
rvolosatovs, line 81: This has to happen in a loop, otherwise the full buffer is not guaranteed to be written https://github.com/WebAssembly/wasi-io/blob/324be895965666805cb1c622ed4f071b9e3cbd65/wit/streams.wit#L134-L160

crates/component-echo-http/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+                Ok(bytes) => body_buf.extend(bytes),
+                Err(StreamError::Closed) => {
+                    body_buf.shrink_to_fit();
+                    break;
+                }
+                Err(e) => {
+                    panic!("failed to read bytes: {e}");
+                }
+            }
+        }
```
rvolosatovs, line 54: Can we stream the body instead?

crates/component-echo-http/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+            .path_with_query()
+            .map(String::from)
+            .unwrap_or_else(|| "/".into());
+        let (path, query) = path_with_query.split_once('?').unwrap_or_else(|| ("/", ""));
+        let response_json = serde_json::json!({
+                    "method": method,
+                    "path": path,
+                    "query_string": query,
+                    "body": body_buf,
+        });
```
rvolosatovs, line 68: Is there a reason this had to be JSON?
Why not set the original request paths as response headers and stream the body back as-is?

crates/component-echo-http/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+        // Build the outgoing response
+        let resp = OutgoingResponse::new(Fields::new());
+        let body = resp.body().expect("failed to open outgoing response body");
+        let out = body
+            .write()
+            .expect("failed to start writing to outgoing response body");
+        out.blocking_write_and_flush(&resp_bytes)
+            .expect("failed to complete blocking write of request");
+        drop(out);
+        OutgoingBody::finish(body, None).unwrap();
```
rvolosatovs, line 83: I would expect the request trailers to be returned here as well

thomastaylor312: Yeah, it is probably worth showing how to use all the buttons for those who use this as a starting point

crates/component-echo-http/src/lib.rs
thomastaylor312: @rvolosatovs Should we just go ahead and publish the helpers we have in the repo that wrap these with `Write` and `Read` implementations? That would probably help out here

thomastaylor312: Ok, my point still stands, but now that I am thinking about it more, I'd rather not add a dependency to a basic example like this. So let's just show the whole loop until empty stuff here

crates/component-echo-http/src/lib.rs
vados-cosmonic: From what I understand, the point of this example (the original `echo` example) is primarily as demo ware, so you can *see* the request (and maybe pipe it to `jq`). 

A similar project is  https://github.com/traefik/whoami. 

I'm fine with changing it, but it's all up to what this component is intended to be used as. @thomastaylor312 @brooksmtownsend @ricochet any comments here, as people who very often do demos?

crates/component-echo-http/src/lib.rs
vados-cosmonic: yeah so we *could*, but the point of this component is to echo the things about the request as JSON as it's good for demoing/using in tests -- it's not really about making an efficient server that returns back what was received, at whatever size.

crates/component-echo-http/src/lib.rs
vados-cosmonic: So this isn't an `example` but is actually supposed to mirror the [`examples/echo`](https://github.com/wasmCloud/wasmCloud/blob/main/examples/echo) repo -- I can definitely add trailers here (maybe someone will do a HTTP/2 demo w/ trailers?), but I don't think it's actually crucial.

crates/component-echo-http/src/lib.rs
vados-cosmonic: - [ ] fix use of `blocking_write_and_flush()`

crates/component-echo-http/src/lib.rs
rvolosatovs: Why not just print text to stdout like `curl -v` or https://github.com/traefik/whoami you mentioned would do?
Then we can trivially just stream the request back.

rvolosatovs (changes requested):
I don't think this example belongs in `crates/`, I'd expect it to be located in `examples/`
I'm also concerned about not showing the expected `wasi:http` usage patterns here (not doing streaming, and ignoring trailers) potentially misleading developers but I suppose that's minor issue. I would expect `json` to appear somewhere in the name of this example, since it seems like serializing an http request to JSON is what it does.

How about `http-request-jsonify` or something like that?

I would expect an `echo` component to:
- Send data back as soon as it's received in the exact shape it was received (i.e. stream body bytes back as they come in using `wasi:io/streams.splice`)
- Send back all elements received (headers, trailers etc.)

I'd hope to have a single `echo` component to do the same for e.g. `wasmcloud:messaging`, so we would just have a single component echoing back whatever it receives on whichever interface

crates/component-echo-http/src/lib.rs
vados-cosmonic: Since this is no longer an `echo` component, I think we can skip adding the extra robustness yet. It's a bit more clear (and I spell out in the documentation) that this component is not meant to be a robust echo server.

crates/component-echo-http/src/lib.rs
vados-cosmonic: Since this is no longer an `echo` component, I think we can skip adding the extra robustness yet. It's a bit more clear (and I spell out in the documentation) that this component is not meant to be a robust echo server.

.github/workflows/examples-components.yml
```diff
@@ -0,0 +1,119 @@
...
+        if: ${{ matrix.project.lang == 'typescript' }}
+        with:
+          node-version: ${{ matrix.project.lang_version }}
+      - name: install node reqs
+        if: ${{ matrix.project.lang == 'typescript' }}
+        run: |
+          npm install -g @bytecodealliance/jco
+          npm install -g @bytecodealliance/componentize-js
+      # Build example project(s)
+      - name: build project
```
thomastaylor312, line 117: Didn't we need to be publishing these as artifacts for us to consume as well?

examples/rust/components/http-jsonify/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+            .expect("failed to start writing to outgoing response body");
+        let mut spool = Vec::<u8>::with_capacity(MAX_READ_BYTES as usize);
+        loop {
+            match buf.read(&mut spool) {
+                // If we read the max bytes then we're clear to read more
+                Ok(bytes_read) => {
+                    let _ = out.blocking_write_and_flush(spool.as_slice());
+                    spool.clear();
+                    // If we read *less* than the maximum amount of bytes, we're at the end
+                    if bytes_read < MAX_READ_BYTES as usize {
```
thomastaylor312, line 134: Generally you're supposed to read until you get a `bytes_read` of `0` or an `EOF`

thomastaylor312 (changes requested):
Mostly good, just a few points around publishing and then also how the component is reading data

examples/rust/components/http-jsonify/src/lib.rs
thomastaylor312: We could also just simplify this function and have it take the whole body as a param instead of an `impl Read`

vados-cosmonic: Hey @thomastaylor312 which body to you mean for replacing the `impl Read`? Are you thinking instead of the `impl Read` it should take a `Vec<u8>` (`resp_bytes`)? 

We talked about this previously but that what is returned is wrapped in JSON (w/ method, path, etc) -- would you like to return the request body directly instead?

thomastaylor312 (approved):

thomastaylor312:
I'd still like to make the write stuff cleaner, but this is good enough for people to get started with

Issue #1756, opened 2024-03-26
[FEATURE] Keyvalue, blobstore and HTTP specific provider libraries


rvolosatovs:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Currently implementing blobstore and keyvalue providers requires a lot of boilerplate, which is copy-pasted. HTTP suffers from the the issue, but custom HTTP providers are more rare and there is less boilerplate, so priority for that one is lower.

We should at least move the subscription setup and the processing loop of keyvalue and blobstore to separate crates or modules in `provider-sdk` crate.
We could have traits like:

- `provider_sdk::interfaces::KeyvalueAtomic`

- `provider_sdk::interfaces::BlobstoreBlobstore`

etc…

Users could then implement this trait and pass implementations directly to `run_provider_handler`

Some things to consider here are the response typing - for example, currently, provider implementations depend on the fact `Some(Vec<u8>)` or `Some(Bytes)` can be used in place of `stream<u8>` on the client's side. Implementations also depend on the fact that e.g. `anyhow::Result<T>` can be transmitted in place of `result<T, string>`.

These kinds of tricks would not be possible if we had a statically typed result type for user-provided handlers specified in the trait.

Perhaps, as a first step we could start with passing the `wrpc_transport::Client` into the handler and let the users figure out what they want to send. Another option is to have handler signatures all return `Result<impl Encode, impl Encode>`, however it's not clear how cases where the implementation may need to send an error in the middle of transmission would be handled - such a situation could arise when e.g. the provider could start the transmission of the response stream and an error occurs on the stream async - the implementation then must communicate the error to the server by transmitting the error string on the error subject.

### Describe the solution you'd like

I think as the first step we should start with something like:

```rust
// somewhere in `provider_sdk`
pub struct Context {
   pub wrpc: wasmcloud_core::wrpc::Client,
   pub result_subject: wasmcloud_core::wrpc::Subject,
   pub error_subject: wasmcloud_core::wrpc::Subject,
   pub source: Option<String>,
}

// probably in `provider_sdk::interfaces::keyvalue`
pub fn serve_atomic(provider: impl AtomicProvider, until: impl Future<Output = ()>)
[...]

PR #1775, opened 2024-03-27
feat: provider interface SDK

rvolosatovs:
## Feature or Problem

Replace boilerplate by nicer-to-use traits

## Related Issues

Closes https://github.com/wasmCloud/wasmCloud/issues/1756

### DIFF
--- a/crates/provider-sdk/src/interfaces/keyvalue.rs
+++ b/crates/provider-sdk/src/interfaces/keyvalue.rs
@@ -1,0 +1,202 @@
+use core::future::Future;
+use core::pin::pin;
+
+use anyhow::Context as _;
+use bytes::Bytes;
+use futures::{Stream, StreamExt as _};
+use tokio::{select, spawn};
+use tracing::{debug, error, instrument, warn};
+use wrpc_transport::{AcceptedInvocation, Transmitter};
+
+use crate::{get_connection, Context};
+
+use super::WrpcContextClient;
+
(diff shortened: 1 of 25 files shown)


crates/provider-sdk/src/interfaces/keyvalue.rs
```diff
@@ -0,0 +1,202 @@
...
+        invocation: AcceptedInvocation<Option<Context>, (String, String, u64, u64), Tx>,
+    ) -> impl Future<Output = ()> + Send;
+
+    fn serve_increment<Tx: Transmitter + Send>(
+        &self,
+        invocation: AcceptedInvocation<Option<Context>, (String, String, u64), Tx>,
+    ) -> impl Future<Output = ()> + Send;
+}
+
+// TODO: This should probably be split into two
```
thomastaylor312, line 57: I also was thinking about this so let’s create an issue so we follow up on this

src/bin/blobstore-s3/main.rs
```diff
@@ -8,7 +9,7 @@ async fn main() -> anyhow::Result<()> {
     let fut = run_provider_handler(provider.clone(), "blobstore-s3-provider")
```
thomastaylor312, line 9: One more improvement might be to make it so callers don’t have to do this step and the serve functions just set it up for you

thomastaylor312 (approved):
This is much cleaner. Great work here

PR #1782, opened 2024-03-29
feat(*)!: Adds version to control API

thomastaylor312:
This should be the final breaking change of the API and it will require a two phased rollout. I'll need to cut new core and host versions first and then update wash to use the new host for tests.

### DIFF
--- a/crates/wash-cli/Cargo.toml
+++ b/crates/wash-cli/Cargo.toml
@@ -63,7 +63,7 @@ wadm = { workspace = true }
 warp = { workspace = true }
 wascap = { workspace = true }
 wash-lib = { workspace = true, features = ["cli", "parser", "nats", "start"] }
-wasmcloud-control-interface = { workspace = true }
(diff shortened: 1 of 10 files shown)


crates/wash-cli/Cargo.toml
```diff
@@ -63,7 +63,7 @@ wadm = { workspace = true }
 warp = { workspace = true }
 wascap = { workspace = true }
 wash-lib = { workspace = true, features = ["cli", "parser", "nats", "start"] }
-wasmcloud-control-interface = { workspace = true }
+wasmcloud-control-interface = { version = "1.0.0-alpha.1" }
```
thomastaylor312, line 66: Note that this crate is pinned here and in wash-lib until I cut the new host release

rvolosatovs (approved):

ricochet (approved):

thomastaylor312:
Going to bypass the review from wash since that was a single version pin change

PR #1794, opened 2024-04-02
feat(providers): add trace context to wrpc client

vados-cosmonic:
## Feature or Problem

This commit adds tracing information (if provided) to the wRPC client when creating it manually from inside a provider (and when using an `InvocationHandler`).

### DIFF
--- a/crates/provider-sdk/src/provider.rs
+++ b/crates/provider-sdk/src/provider.rs
@@ -827,17 +827,43 @@ impl ProviderConnection {
         })
     }
 
-    /// Used for fetching the RPC client in order to make RPC calls
+    /// Retrieve a wRPC client that can be used based on the NATS client of this connection
+    ///
+    /// # Arguments
+    ///
+    /// * `target` - Target ID to which invocations will be sent
     pub fn get_wrpc_client(&self, target: &str) -> wasmcloud_core::wrpc::Client {
-        let mut headers = HeaderMap::new();
-        headers.insert("source-id", self.provider_key.as_str());
-        headers.insert("target-id", target);
+        self.get_wrpc_client_custom(target, None, None)
+    }
+
+    /// Retrieve a wRPC client that can be used based on the NATS client of this connection,
+    /// customized with headers and duration
+    ///
+    /// # Arguments
+    ///
+    /// * `target` - Target ID to which invocations will be sent
+    /// * `headers` - Additional headers (other than `source-id`, `target-id`) to be placed on the client
+    /// * `timeout` - Timeout to be set on the client (by default if this is unset it will be 10 seconds)
+    pub fn get_wrpc_client_custom(
+        &self,
+        target: &str,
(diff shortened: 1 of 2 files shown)


crates/provider-sdk/src/provider.rs
```diff
@@ -827,17 +827,27 @@ impl ProviderConnection {
         })
     }
 
-    /// Used for fetching the RPC client in order to make RPC calls
-    pub fn get_wrpc_client(&self, target: &str) -> wasmcloud_core::wrpc::Client {
-        let mut headers = HeaderMap::new();
-        headers.insert("source-id", self.provider_key.as_str());
-        headers.insert("target-id", target);
+    /// Retrieve a wRPC client that can be used based on the NATS client of this connection
+    pub fn get_wrpc_client(
```
thomastaylor312, line 831: This feels a little awkward to be calling all the time with 2 `None` params. Maybe we should have a second function that is called something like `get_wrpc_client_with_context` (or something better) that has this function signature and the normal `get_wrpc_client` just wraps that function with the single target call for the default use case

thomastaylor312 (commented):
Mostly good, just had a question around API design

brooksmtownsend (commented):
This should set the baseline for propagating the trace context through with the invocation, do we need to have anything additional that sets the parent of the incoming trace context, e.g. when a component calls a provider? (Forgive me if I missed that implementation in another PR

crates/provider-sdk/src/provider.rs
vados-cosmonic: Yeah I certainly agree with you there -- I thought it was appropriate [for `InvocationHandler` to have two](https://github.com/wasmCloud/wasmCloud/pull/1794/files#diff-105fc2ae91492fff7c55a63d749a4477d6094dff696f800b75fdb0a565063b81R544), but not really for this part. It feels awkward/excessive to have two functions for getting a wRPC client...

I was thinking that in the future people generally don't want to be calling this themselves actually (they should be using `InvocationHandler` after bindgen-ing), but the awkwardness is definitely there.

This choice is kind of forced by the upstream `Client` -- there's no way to iteratively modify/build the client either.

That said, if we're happy with having the two functions, I'll add a second one!

- [ ] Add `get_wrpc_client_with_context`

vados-cosmonic:
@brooksmtownsend I'm not quite sure but maybe you're referring to these pieces of code:

https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-sdk/src/provider.rs#L970
https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-sdk/src/provider.rs#L789

Context that comes in on the request are all added to the current span (if `otel` is enabled) and then the dispatch happens. The information is also on the context for the provider to pass along (when it calls anything else), but that's for the provider to do, at present.

brooksmtownsend (approved):

PR #1795, opened 2024-04-02
fix(examples): Add logging to TS deps

ericgregory:
The TypeScript http-hello-world template is missing WIT definitions for wasi:logging and fails on build. This PR fixes the issue on main.


thomastaylor312 (changes requested):
Can you add the logging import to the `deps.toml` and run `wit-deps update`?

ericgregory:
> Can you add the logging import to the `deps.toml` and run `wit-deps update`?

@thomastaylor312 Done! The TypeScript deps.toml was also missing keyvalue, so I added that while I was there.

PR #1796, opened 2024-04-02
chore/control interface actor rename

brooksmtownsend:
- **chore(control-interface)!: rename actor to component**
- **chore(host)!: rename ctl actor to component**

## Feature or Problem
This PR renamed the use of `actor` to `component` in the control interface library, and propagates those changes through the host and wash. Note that this doesn't change all code occurrences as those are purely internal, and this PR focuses instead on error messages, NATS topics, and logs.

The notable remaining thing to change from actor to component is the `actor_scaled` and `actor_scale_failed` events.

## Release Information
Next breaking alpha/rc of wasmcloud and wash

## Consumer Impact
Compatibility with previous alpha versions of wasmCloud and wash will be broken, but it already was with the topic change to include v1 so there shouldn't be too much friction here.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -35,11 +35,11 @@ use tracing::{debug, error, info, instrument, trace, warn};
 use uuid::Uuid;
 use wascap::{jwt, prelude::ClaimsBuilder};
 use wasmcloud_control_interface::{
-    ActorAuctionAck, ActorAuctionRequest, ActorDescription, CtlResponse,
+    ComponentAuctionAck, ComponentAuctionRequest, ComponentDescription, CtlResponse,
     DeleteInterfaceLinkDefinitionRequest, HostInventory, HostLabel, InterfaceLinkDefinition,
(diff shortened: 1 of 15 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -2642,7 +2655,7 @@ impl Host {
         Ok(())
     }
 
-    // TODO(#1548): With actor IDs, new actor references, configuration, etc, we're going to need to do some
+    // TODO(#1548): With component IDs, new actor references, configuration, etc, we're going to need to do some
```
thomastaylor312, line 2658: Heads up that I don't see this covered in #1548. We should probably create an issue to decide what we're going to do here so we know if it would be a breaking change or not

thomastaylor312 (changes requested):
Good to go, but might have to do the trick with wash where you pin to a specific version for a commit so we can release

brooksmtownsend:
@thomastaylor312 I ended up popping the commit off and putting it as a part of https://github.com/wasmCloud/wasmCloud/tree/chore/control-interface-actor-rename-wash so I can create another PR afterwards with this change in wash, so this basically just keeps this PR clean and for the host.

PR #1798, opened 2024-04-02
fix(examples): Add logging to TS deps on 0.82-examples

ericgregory:
The TypeScript http-hello-world template is missing WIT definitions for wasi:logging and fails on build. This PR fixes the issue on the 0.82-examples branch used by v0.82 docs.


thomastaylor312 (commented):
This should be good. As a note to ourselves, on main the `deps.toml` should be updated to have logging in it. I'm ok here because 0.82 is a couple of wasmtime versions behind

thomastaylor312 (approved):

PR #1809, opened 2024-04-03
feat(runtime)!: Removes old guest config and uses runtime config instead

thomastaylor312:
Most of the changes are related to wit updates, but this removes the guest config from `wasmcloud:bus` and pulls down `wasi:config` in its place


crates/runtime/src/actor/component/mod.rs
```diff
@@ -465,8 +467,7 @@ fn instantiate(
             | "wasi:io/poll@0.2.0"
             | "wasi:io/streams@0.2.0"
             | "wasi:sockets/tcp@0.2.0"
-            | "wasmcloud:bus/lattice"
-            | "wasmcloud:bus/guest-config" => continue,
```
brooksmtownsend, line 469: :excellent:

crates/runtime/src/actor/component/config.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+
+#[async_trait]
+impl runtime::Host for Ctx {
+    #[instrument(skip(self))]
+    async fn get(
+        &mut self,
+        key: String,
+    ) -> anyhow::Result<Result<Option<String>, config::runtime::ConfigError>> {
```
brooksmtownsend, line 25: I know the runtime interface is implemented as a `Result<Option<String>>`, is there a reason why we double wrap this here with another anyhow Result?

brooksmtownsend (approved):
LGTM and approved even though I had a question because I see we use that pattern elsewhere

crates/runtime/src/actor/component/config.rs
thomastaylor312: The result is something defined by the host bindings insofar as I understand. I think everything is wrapped by a result type

PR #1810, opened 2024-04-03
feat(*): Bumps our wasmcloud wit to 1.0

thomastaylor312:
This bumps our main `wasmcloud:bus` package and the package in the `runtime` crate to 1.0.0 in preparation for release. Also removes the use of the wasmcloud interface in tests where it wasn't needed


thomastaylor312:
Please note that this is stacked on top of #1809 to avoid merge conflicts. I will rebase once the other is merged

wit/bus.wit
brooksmtownsend, line 1: I see a pretty significant TODO in this WIT, do we want to make sure this interface is ready for stability?

What I think of as the (pseudocode) usage of this interface is:
```rust
wasi::keyvalue::set("foo", "bar"); // sets a key in whatever is linked on default
wasmcloud::bus::lattice::set_link_name("secret", "wasi:keyvalue/set");
wasi::keyvalue::set("foo", "p4ssw0rd"); // sets a key in whatever is linked on secret
```

If we removed the interfaces, then we could do the same as above with a little less code at the cost of setting the link name for every interface call, which makes me think the above is still necessary.

brooksmtownsend: A conversation for another PR is whether or not this implementation in the host works with different component instances (cc @vados-cosmonic, since we spoke about this)

wit/README.md
brooksmtownsend, line 1: I think it would be great to include a small README here just noting that this WIT interface is used for the advanced use case of multiple links on the same interface (you could leave it at that.) Optional pointer to where the full list of WIT things we support (runtime?)

brooksmtownsend (changes requested):
Version bumps look great to me. Requesting changes since I think we should figure out these issues before we call the wits 1.0, but I don't think we have to hold this specific PR back as long as we're OK with it changing in `main`

wit/README.md
thomastaylor312: I’d rather document it in the interface itself since docs can be generated from that

wit/bus.wit
thomastaylor312: We should actually figure that out here before we bump to 1.0

thomastaylor312: So I think we should keep interfaces just so it can be more focused. If you have one interface that only uses a custom link name and the others are default, it would really suck to have to unset the link name with every call. I think we can make some nice wrappers for this that could go into `wasmcloud_actor` eventually

wit/README.md
thomastaylor312: Documentation added!

thomastaylor312:
This will be ready to go once #1815 is merged. I'll rebase on that just to be sure before merging this

brooksmtownsend (approved):

PR #1817, opened 2024-04-05
chore: update `messaging` to 0.2

rvolosatovs:
## Feature or Problem

This began as just a WIT package update, but I quickly found out that versions were not supported by existing `provider-wit-bindgen`. Adding support for that proved to be too difficult due to NATS topic string-matching etc.
I've adapted upstream https://github.com/bytecodealliance/wit-bindgen in wRPC over the course of last week and switched to that instead to enable versioned WIT interfaces along with other use cases (anonymous interfaces, functions outside interfaces, import of multiple interfaces with equal names to name a few)
The UX of this bindgen is exactly the same as upstream `wit-bindgen` used for components and for most use cases simple `wit_bindgen_wrpc::generate!()` call is enough with no extra arguments.

Everything except for flags (will be added shortly), exported resources, `future` and `stream` types is supported by `wit-bindgen-wrpc`

Did a few changes and clean-ups, that helped with simplifying migration:
- Remove `ProviderHandler` trait, there's just `Provider` now
- Remove wrpc-related traits from `provider_sdk` - these are now redundant, since we use the `wrpc_transport::Client` directly to serve and invoke functionality as opposed to reimplementing parts of wRPC in provider-related crates as was done before

I think we should figure out a way to generate a `run` function, akin to one currently defined as a method on messaging providers for custom interfaces in a follow-up.

`0.2.0` version was chosen to match current wasi-cloud interface versions

## Related Issues

https://github.com/wasmCloud/messaging/pull/5
https://github.com/rvolosatovs/wrpc/commit/65e830e5d29a4a7b834cb26bcaca179c9d10145b

### Manual Verification

`cargo test` and https://github.com/rvolosatovs/wrpc/commit/65e830e5d29a4a7b834cb26bcaca179c9d10145b


src/bin/messaging-nats/main.rs
```diff
@@ -1,15 +1,10 @@
...
 
-fn main() -> anyhow::Result<()> {
-    let host_data = load_host_data().context("failed to load host data")?;
-    start_provider(
-        NatsMessagingProvider::from_host_data(host_data),
-        "nats-messaging-provider",
-    )?;
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    NatsMessagingProvider::run().await?;
```
brooksmtownsend, line 7: 😍

crates/core/src/chunking.rs
brooksmtownsend, line 1: We may be able to remove this file entirely now that we have wRPC

crates/provider-sdk/src/lib.rs
```diff
@@ -135,74 +132,21 @@ pub struct Context {
...
-pub trait Provider: ProviderHandler + WrpcNats + WrpcDispatch + Send + Sync + 'static {}
-
 /// Configuration of a link that is passed to a provider
-pub trait LinkConfig: Send + Sync {
+pub struct LinkConfig<'a> {
     /// Given that the link was established with the source as this provider,
-    /// get the target ID which should be a component
-    fn get_target_id(&self) -> &LatticeTarget;
+    /// this is the target ID which should be a component
+    pub target_id: &'a str,
```
brooksmtownsend, line 139: We actually originally made LinkConfig a trait so that we could add or remove members to it without breaking the public API, do you think that wont be necessary?

crates/runtime/src/actor/component/mod.rs
```diff
@@ -445,7 +445,9 @@ fn instantiate(
     let mut custom_result_types = HashMap::with_capacity(imports.len());
     for (instance_name, item) in imports {
         match instance_name {
-            "wasi:config/runtime@0.2.0-draft"
+            "wasi:blobstore/blobstore@0.2.0-draft"
+            | "wasi:blobstore/container@0.2.0-draft"
+            | "wasi:blobstore/types@0.2.0-draft"
```
brooksmtownsend, line 450: It could be worth adding a comment here to indicate what interfaces should be ignored here

crates/provider-sdk/src/error.rs
```diff
@@ -3,9 +3,6 @@
 pub type InvocationResult<T> = Result<T, InvocationError>;
 pub type ProviderInitResult<T> = Result<T, ProviderInitError>;
 
-/// Result form for [`ProviderHandler`] trait methods
-pub type ProviderOperationResult<T> = Result<T, Box<dyn std::error::Error + Send + Sync>>;
-
```
brooksmtownsend, line 8: For this change from basically the standard Result -> an anyhow::Result, what's the motivation? I don't actually feel dedicated to either route, I originally remember talking with Victor about using this form to be a little more flexible

thomastaylor312 (commented):
I think this is looking pretty good. Leaving this as a comment so as to not block approval once @brooksmtownsend re-reviews

crates/provider-sdk/src/error.rs
thomastaylor312: I actually prefer anyhow to boxed dynner errors. If anything, I'd love for us to have a generic `E` if we can, but I can live without it

crates/provider-sdk/src/lib.rs
thomastaylor312: I actually think we're ok here as we will definitely need at least these fields. Anything else is additive, which we could get away with...probably. It would break the SDK, but not compatibility with wasmcloud hosts. Let's just keep it simple for now as if we have to add something to the trait it would break as well. tl;dr; I think I'm ok with this

brooksmtownsend: Agree

crates/provider-sdk/src/error.rs
brooksmtownsend: Agree

brooksmtownsend (approved):

crates/provider-sdk/src/error.rs
rvolosatovs: @brooksmtownsend the effect here is actually the opposite, the error is generic `E`, just like @thomastaylor312 described in https://github.com/wasmCloud/wasmCloud/pull/1817#discussion_r1554275395
`anyhow::Error` is merely the default - developers are free to choose any error type they wish

crates/provider-sdk/src/lib.rs
rvolosatovs: I've marked the struct `non_exhaustive` to address the compatibility concerns here - for all other intents and purposes I struggle to see the difference in terms of compatibility compared to a trait here.

In terms of motivation for this change - like @thomastaylor312 outlined above, mostly simplicity and making developer experience smoother/easier.
Do note that this change actually fixed a bug https://github.com/wasmCloud/wasmCloud/blob/10ea04b4c36c31a16b20a3cfdffc56da5a39c948/crates/provider-keyvalue-vault/src/lib.rs#L368, which IMO is caused exactly by the trait abstraction

Apart from that the big reason for the change is maintainability - implementing this trait for opaque tuples is very error-prone and adding fields to those tuples would only make it worse

PR #1824, opened 2024-04-08
fix(provider-wit-bindgen): handle TypeDef::Enum(_)

vados-cosmonic:
## Feature or Problem

In the handling of TypeDefKinds done in `convert_wit_typedef()` Enum was excluded but should have been handled. Since Enums are very similar to Records, the same handling should work for both.


rvolosatovs (commented):
Thank you for contribution! I'm not sure this still would still be relevant, since we've migrated to https://docs.rs/wit-bindgen-wrpc/latest/wit_bindgen_wrpc/index.html in #1817

vados-cosmonic:
Ah thanks for the pointer -- yes you're right, with the bindgen crates gone this certainly isn't needed any more

PR #1829, opened 2024-04-08
fix(host): warn scaling with different imageref

brooksmtownsend:
## Feature or Problem
This PR fixes a hole in one of our common operations, scale component. There are two things that this PR addresses:
1. When scaling a component that you've previously stopped on all hosts to a new image reference, the component will update to the new image reference
1. When scaling a component that's currently running to a new level of scale, the host will scale the currently running component to the specified scale WITHOUT updating the image reference.

This primarily fixes a particularly annoying issue where you're trying to just update a component that you ran previously. As a secondary aspect we now properly enforce the scale operation being something that's meant to declaratively change the scale of a component, not something that supports upgrades.

## Release Information
wasmCloud 1.0.0-rc.1

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2255,24 +2255,10 @@ impl Host {
                 .context("failed to store claims")?;
         }
 
-        let component_spec =
-            if let Ok(Some(mut spec)) = self.get_component_spec(&component_id).await {
-                // If the component didn't start yet, the URL will be empty but the spec may contain links.
-                // Populate the URL and store the updated spec.
-                if spec.url.is_empty() {
-                    spec.url = component_ref.to_string();
-                } else if spec.url != component_ref {
-                    // Ensure another actor isn't already running with the same ID
-                    bail!(
-                        "component spec URL does not match component reference: {} != {}",
-                        spec.url,
-                        component_ref
-                    );
-                }
-                spec
-            } else {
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -2621,6 +2607,14 @@ impl Host {
                 let actor = entry.get_mut();
                 let config_changed =
                     &config != actor.handler.config_data.read().await.config_names();
+
+                if actor.image_reference != component_ref {
+                    warn!(
+                        "Requested to scale existing component to a different image reference: {} != {}. The component will be scaled but the image reference will not be updated. If you meant to update this component to a new image ref, use the update command.",
```
thomastaylor312, line 2613: Should we send this message back in the response to the user or are we too far down a spawned process?

thomastaylor312 (changes requested):

crates/host/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-host"
-version = "0.82.0"
+version = "1.0.0-alpha.1"
```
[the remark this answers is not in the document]
brooksmtownsend: I just think it's weird to have it at the same `0.82`. I can revert

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Pushing this

PR #1839, opened 2024-04-08
fix(wash-cli): use config version in error output

lachieh:
## Feature or Problem
Error output was using the static version for wadm even when `--wadm-version` is set.

## Related Issues
None

## Release Information
`next`

## Consumer Impact
Better DX

### DIFF
--- a/crates/wash-cli/src/up/mod.rs
+++ b/crates/wash-cli/src/up/mod.rs
@@ -430,7 +430,8 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
                 }
             }
             Err(e) => {
-                eprintln!("🟨 Couldn't download wadm {WADM_VERSION}: {e}");
+                let wadm_version: String = cmd.wadm_opts.wadm_version.clone();
+                eprintln!("🟨 Couldn't download wadm {wadm_version}: {e}");
                 None
             }
         }


crates/wash-cli/src/up/mod.rs
```diff
@@ -430,7 +430,8 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
                 }
             }
             Err(e) => {
-                eprintln!("🟨 Couldn't download wadm {WADM_VERSION}: {e}");
+                let wadm_version: String = cmd.wadm_opts.wadm_version.clone();
```
lachieh, line 433: is this necessary? Is there a different/better way to do this?

crates/wash-cli/src/up/mod.rs
```diff
@@ -430,7 +430,8 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
...
+                eprintln!("🟨 Couldn't download wadm {wadm_version}: {e}");
```
rvolosatovs, line 434:
```suggestion
                eprintln!("🟨 Couldn't download wadm {}: {e}", cmd.wadm_opts.wadm_version);
```

(Response to https://github.com/wasmCloud/wasmCloud/pull/1839#discussion_r1556439390)

Not that it really matters here

rvolosatovs (approved):

thomastaylor312 (approved):

PR #1849, opened 2024-04-09
feat(examples): publish artifacts for select examples

vados-cosmonic:
## Feature or Problem

This commit adds machinery to publish artifacts for select examples which should be externally available.

To publish artifacts, an tag named `component-<name>-vX.X.X` should be pushed to the repository, where `<name>` is an existing example configuration in `.github/workflows/examples-components.yml`.

### DIFF
--- a/.github/workflows/examples-components.yml
+++ b/.github/workflows/examples-components.yml
@@ -2,7 +2,10 @@ name: wasmCloud example components
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+    tags:
+      - component-http-jsonify-v[0-9]+.[0-9]+.[0-9]+
   pull_request:
     branches: [main]
     paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/examples-components.yml
```diff
@@ -117,3 +132,86 @@ jobs:
...
+            name: "http-jsonify"
+            wasm-bin: "wasmcloud_component_http_jsonify_s.wasm"
+    steps:
+      # Determine tag version (if this is a release tag), without the 'v'
+      - name: Determine version
+        id: meta
+        shell: bash
+        env:
+          TAG: ${{ github.ref_name }}
+          BRANCH_PREFIX: ${{ format('refs/tags/component-{0}-v', matrix.project.name) }}
```
brooksmtownsend, line 164: I'm having a hard time narrowing it down, but looking at the matrix here am I right in thinking that this will release the http-jsonify component every time this action is triggered when the tag starts with `component-`?

brooksmtownsend (commented):
The action looks good to release this specific component, I think I'm missing the filter that _only_ releases this component when we add more examples to trigger this matrix

.github/workflows/examples-components.yml
vados-cosmonic: Yup, and for every step after that which matters it should narrow it down to the actual branch for the single component that matches in the matrix -- see the `if`s that skip afterwards, depending on `branch-prefix`

brooksmtownsend: Ah so that's the part. Okay cool 👍🏻 Branch prefix is what I needed

brooksmtownsend (approved):

.github/workflows/examples-components.yml
vados-cosmonic: Yeah I will agree it's a bit hard to understand...

One thing I could do is change this to break apart/parse the tag and do it that way? We just need to be very specific about tag names and then it would be good?

PR #1850, opened 2024-04-09
chore(host): imrpove wording for spec/provider ref mismatch

vados-cosmonic:
## Feature or Problem

This commit slightly improves the wording when a provider ID and component specification URL mismatch occurs, along with specifying a possible solution.

This error is thrown by `wash` and it's a bit difficult to figure out what to resolve it otherwise.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2895,24 +2895,26 @@ impl Host {
             "policy denied request to start provider `{request_id}`: `{message:?}`",
         );
 
-        let component_specification =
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
Ah shoot, we should probably remove this check anyways to match the change in #1829 . No harm in inserting this, but you could just remove the check entirely in favor of checking to see if the provider is already running.

Since there's no update provider command, I think we definitely have to allow people to overwrite it

vados-cosmonic:
Yeah I'm a bit wary as people need to be careful not to depend on the values in there without checking for updates, but will follow up with a removal of the check!

PR #1862, opened 2024-04-10
fix(ci): use wash push in a single line

vados-cosmonic:

### DIFF
--- a/.github/workflows/examples-components.yml
+++ b/.github/workflows/examples-components.yml
@@ -164,7 +164,7 @@ jobs:
           TAG: ${{ github.ref_name }}
           BRANCH_PREFIX: ${{ format('refs/tags/component-{0}-v', matrix.project.name) }}
         run: |
-          export VERSION=${TAG#$PREFIX};
+          export VERSION=${TAG#$BRANCH_PREFIX};
           echo -e "version=${VERSION}" >> $GITHUB_OUTPUT;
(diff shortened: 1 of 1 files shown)


.github/workflows/examples-components.yml
```diff
@@ -185,39 +185,27 @@ jobs:
           WASH_REG_USER: ${{ github.repository_owner }}
           WASH_REG_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          wash push \
-            --config=${{ steps.meta.outputs.wasmcloud-toml-path }} \
-            ghcr.io/${{ github.repository_owner }}/component-${{ matrix.project.name }}:${{ github.sha }} \
-            artifacts/${{ steps.meta.outputs.bin-name }}
+          wash push ghcr.io/${{ github.repository_owner }}/component-${{ matrix.project.name }}:${{ github.sha }} artifacts/${{ steps.meta.outputs.bin-name }}
```
rvolosatovs, line 188: no more config?

vados-cosmonic: yep, removed this -- the problem didn't turn out to be there -- `registry` is coming through as empty somehow, which is what made me think it was the `\`s. 

See [this failed run](https://github.com/wasmCloud/wasmCloud/actions/runs/8632918526/job/23665073246#step:6:14) for the error message. Clearly something is wrong if `wash push` is not reading the two readily provided arguments.

PR #1864, opened 2024-04-10
chore: address clippy warnings

rvolosatovs:
## Feature or Problem

I mostly just let `cargo clippy --fix --allow-dirty --allow-staged --all-targets --workspace -- -Wclippy::pedantic` do its job a few times and ran `cargo fmt --all` a few times as well


thomastaylor312 (approved):
Approving so as to not block on the nits. They are just nits and we can fix later if needed

PR #1869, opened 2024-04-10
test(wash): re-enable wash call test

vados-cosmonic:
## Feature or Problem

Resolves https://github.com/wasmCloud/wasmCloud/issues/1644

## Release Information

This PR modifies `wash call` to be able to support sending HTTP requests to actors that implement `wasi:http/incoming-handler.handle`

### DIFF
--- a/crates/wash-cli/src/call.rs
+++ b/crates/wash-cli/src/call.rs
@@ -1,19 +1,30 @@
 use std::collections::HashMap;
 use std::io::Write;
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::time::Duration;
 
 use anyhow::{bail, ensure, Context, Result};
+use bytes::{Bytes, BytesMut};
 use clap::Args;
-use serde::Deserialize;
+use futures::StreamExt as _;
+use serde::{Deserialize, Serialize};
+use serde_json::json;
 use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
 use tracing::debug;
+
 use wash_lib::cli::{validate_component_id, CommandOutput};
 use wash_lib::config::{create_nats_client_from_opts, DEFAULT_LATTICE};
 use wasmcloud_core::parse_wit_meta_from_operation;
+use wrpc_interface_http::IncomingHandler;
 use wrpc_transport::Client;
 
 use crate::util::{default_timeout_ms, msgpack_to_json_val};
+
+const DEFAULT_HTTP_SCHEME: &str = "http";
+const DEFAULT_HTTP_HOST: &str = "localhost";
+/// Default port used by wasmCloud HTTP server provider
+const DEFAULT_HTTP_PORT: u16 = 8080;
 
 #[derive(Deserialize)]
 struct TestResult {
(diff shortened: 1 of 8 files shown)


vados-cosmonic:
Looks like some upstream changes have fixed `wash call` at least locally for me -- call succeeds now!

crates/wash-cli/tests/wash_call.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    // Call the component
+    let cmd_output = instance
+        .call_component(
+            &component_id,
+            "wasi:http/incoming-handler.handle",
+            // TODO: special case for incoming-handler operations
+            //
+            // We need to build a HTTP request out of whatever we get as JSON input, then
+            // push it over the wire...?
```
brooksmtownsend, line 46: Is this still a TODO?

crates/wash-cli/tests/wash_call.rs
```diff
@@ -0,0 +1,55 @@
...
+        "method": "GET",
+        "path": "/",
+        "body": "",
+        "queryString": "",
+        "header": {},
+    });
+
+    // TODO: add test when no input data is specified as well
+    // (should probably turn into a general GET rquest")
+
```
brooksmtownsend, line 37: This is the default right? Or is this just a note for the jsonify component

crates/wash-cli/src/call.rs
```diff
@@ -222,27 +234,135 @@ pub async fn handle_call(
...
+                )]),
+            ))
+        }
+        // All other responses we're not
+        _ => bail!("unexpected response after HTTP wRPC invocation"),
+    }
+}
+
+/// Invoke a wRPC endpoint that takes nothing and returns a string
+async fn wrpc_invoke_string_echo(
```
brooksmtownsend, line 340: nit: `echo` feels a little misleading here since it's calling a no-arg function but I don't have a better suggestion than like `wrpc_invoke_simple`

crates/wash-cli/src/call.rs
```diff
@@ -182,6 +180,10 @@ pub struct CallCommand {
     /// Fully qualified WIT export to invoke on the component, e.g. `wasi:cli/run.run`
     #[clap(name = "function")]
     pub function: String,
+
+    /// Input to the function, in a format that is hopefully readable by the function
+    #[clap(long = "input")]
+    pub input: Option<String>,
```
brooksmtownsend, line 186: I don't see this used anywhere, can we remove it?

brooksmtownsend (changes requested):

crates/wash-cli/src/call.rs
vados-cosmonic: Ah sorry forgot to update, but this is fixed -- I expanded this to be a bit more inline with what I wanted, more customization for people using `wash call`

crates/wash-cli/src/call.rs
vados-cosmonic: Yup, renamed!

crates/wash-cli/tests/wash_call.rs
vados-cosmonic: Yup this is the default if nothing is specified, removed the TODO (and greatly expanded customization)

crates/wash-cli/tests/wash_call.rs
vados-cosmonic: Nope! removed

vados-cosmonic:
Hey @brooksmtownsend I thought that we should add a bit more customization here, especially if we're going to show off `wash call`/use it with HTTP handlers. The code's a bit different (I expanded customization options) but the core is the same, probably reasonable to look at it with semi-fresh eyes now.

brooksmtownsend (approved):

PR #1877, opened 2024-04-11
feat(examples): add NATS messaging provider example

brooksmtownsend:
## Feature or Problem
This PR adapts our NATS messaging provider to simplify the options and configuration and set it up for our developer guide for building a provider. The action item here for developers is to generate this capability provider and then implement the functionality for the `wasmcloud:messaging` interface (instructions will be detailed on wasmcloud.com).

The TODOs and `todo!`s in the code are intentional and meant for developers to implement it after generating the project from the template. For a full implementation the README references our capability provider.

- [x] We should bump the provider SDK as well, I will create a followup PR in order to do that, release the provider SDK 0.4.0, and rename `actor` to `component` in the context.

## Release Information
N/A


brooksmtownsend:
Integration test probs failing because I changed the provider template 😄 Yay tests!

examples/rust/providers/messaging-nats/Cargo.lock
thomastaylor312, line 1: I don't think we should have the lock file committed if this is going to be a template

examples/rust/providers/messaging-nats/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[dependencies]
+anyhow = "1.0.82"
+async-nats = "0.33.0"
+bytes = "1.6.0"
+futures = "0.3.30"
+serde = { version = "1.0.197" , features = ["derive"] }
+serde_json = "1.0.115"
+tokio = { version = "1.37.0", features = [ "full" ] }
+tracing = "0.1"
+wasmcloud-provider-sdk = { version = "0.3.0", path = "../../../../crates/provider-sdk" }
```
thomastaylor312, line 23: We should be pointing at the released crate if this is going to be used as a template

examples/rust/providers/messaging-nats/src/connection.rs
```diff
@@ -0,0 +1,67 @@
...
+        ConnectionConfig {
+            subscriptions: vec![],
+            uri: DEFAULT_NATS_URI.to_string(),
+        }
+    }
+}
+
+impl ConnectionConfig {
+    /// Construct configuration Struct from the passed config values
+    pub fn from_map(values: &HashMap<String, String>) -> Result<ConnectionConfig> {
```
thomastaylor312, line 53: Nit: This should probably just be a `From` implementation for `ConnectionConfig` (also would prefer that it takes owned data instead of borrowed, but that is more of a super nit)

thomastaylor312: Oh, and this definitely doesn't need to be a `Result` as there are no error cases

examples/rust/providers/messaging-nats/src/nats.rs
```diff
@@ -0,0 +1,258 @@
...
+        serve(
+            &connection.get_wrpc_client(connection.provider_key()),
+            provider,
+            shutdown,
+        )
+        .await
+    }
+
+    /// Build a [`NatsMessagingProvider`] from [`HostData`]
+    pub fn from_host_data(host_data: &HostData) -> NatsMessagingProvider {
```
thomastaylor312, line 71: Super nit (honestly more of a follow up for later): We have this function in most of our providers, but we don't have a function like `run_with_host_data` that does all the same things as `run` without fetching the host data

thomastaylor312 (changes requested):
Mostly nits with a few minor changes

examples/rust/providers/messaging-nats/src/connection.rs
brooksmtownsend: Yep good call. Leftover from the "real" impl

examples/rust/providers/messaging-nats/Cargo.toml
brooksmtownsend: Yes we actually have to release 0.4.0 😢 Captured as a bullet in the PR but will fix since it should be done now

brooksmtownsend: @thomastaylor312 bumped to 0.4.0, will remove the path in a quick followup PR.

brooksmtownsend:
In order to bump the provider SDK to 0.4.0 and release it, I also had to bump wasmcloud-tracing to 0.3.0 in order to bring in the wasmcloud-core bump to 0.5.0. Wheee crates!

(no breaking changes in wasmcloud-tracing, it would just break projects that depend on both tracing and core, like the provider SDK)

thomastaylor312 (approved):

PR #1879, opened 2024-04-11
fix(ci): use ref prefix rather than ref name

vados-cosmonic:
## Feature or Problem

This commit modifies the release logic to use the full ref name rather than just tag name. This both is more consistent with other code, is clearer ("branch prefix" wasn't right), and avoids collisions with refs that are non-tags.

### DIFF
--- a/.github/workflows/examples-components.yml
+++ b/.github/workflows/examples-components.yml
@@ -163,13 +163,13 @@ jobs:
         id: meta
         shell: bash
         env:
-          TAG: ${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
Thanks for the name change. I reviewed the last one thinking it was just supposed to be the branch, not the full ref due to the name

vados-cosmonic:
Yeah I realized that ONLY after I had merged it and ran the pipeline (which happily skipped everything)

PR #1881, opened 2024-04-11
feat: add echo messaging example

brooksmtownsend:
## Feature or Problem
This PR creates an example component called `echo-messaging` that will respond to any message received on the `wasmcloud:messaging` interface that includes a `reply_to` field with the same payload it received. The most basic example is exercised in the README, which is sending a `nats req` and receiving back the same payload.

## Related Issues
This will be great to test https://github.com/wasmCloud/wasmCloud/pull/1877

### Manual Verification
I did make sure this worked


examples/rust/components/echo-messaging/Cargo.lock
thomastaylor312, line 1: Probably don't need the lock file committed for the example

thomastaylor312 (changes requested):
One small comment to address here

examples/rust/components/echo-messaging/Cargo.lock
brooksmtownsend: We have it committed for all other components so if we remove it from here I'd rather remove it from all

thomastaylor312 (approved):

PR #1909, opened 2024-04-16
fix(host): allow overwriting provider reference

brooksmtownsend:
## Feature or Problem
This PR allows overwriting the provider reference in the component specification for a given ID if that provider is not running on this host. This primarily solves a developer experience issue where a provider is no longer running and it cannot be updated without deleting the component spec manually.

## Related Issues
Implemented for components in https://github.com/wasmCloud/wasmCloud/pull/1829

## Release Information
1.0.0


vados-cosmonic (approved):
LGTM :rocket:

ricochet (approved):

PR #1913, opened 2024-04-16
fix(host): differentiate no config and config error

### LINKED ISSUE #1906: [FEATURE] Return `None` instead of an error for missing configuration
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

There is no way to discern if the host failed to retrieve configuration because of a network issue and if the configuration isn't present.

When I retrieve configuration from a host, if the configuration doesn't exist I expect to get a CtlResponse with `success: true` and `response: None`, with a message indicating the config is missing. Right now, instead, I get an unsuccessful response with `None`. 

_No response_

### Additional context

Here's where we should do additional logic to return the proper payload https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L2386

brooksmtownsend:
## Feature or Problem
This PR updates the handling of the `get_config` control interface function to return a `None` when a configuration isn't found, instead of an error making it difficult to tell if the configuration was absent or if there was a legitimate error.

## Related Issues
Fixes #1906

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2364,15 +2364,18 @@ impl Host {
     #[instrument(level = "trace", skip(self))]
     async fn handle_config_get(&self, config_name: &str) -> anyhow::Result<Vec<u8>> {
         trace!(%config_name, "handling get config");
-        let config_bytes = self
-            .config_data
-            .get(config_name)
-            .await?
-            .context("config not found")?;
-        let config_map: HashMap<String, String> = serde_json::from_slice(&config_bytes)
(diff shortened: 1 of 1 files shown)


thomastaylor312 (changes requested):
This code is good to go, but we should also include an update to the `get_config` function in the ctl client

thomastaylor312:
Yeah, figured it worked fine, but was just more concerned about the API. I think a comment would be fine

PR #1924, opened 2024-04-17
fix(providers): ensure trace context is attached for messaging-nats

vados-cosmonic:

### DIFF
--- a/crates/provider-messaging-nats/src/lib.rs
+++ b/crates/provider-messaging-nats/src/lib.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 
 use anyhow::{anyhow, bail, Context as _};
 use async_nats::subject::ToSubject;
+use async_nats::HeaderMap;
 use futures::StreamExt;
 use opentelemetry_nats::{attach_span_context, NatsHeaderInjector};
 use tokio::fs;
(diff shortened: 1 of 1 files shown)


crates/provider-messaging-nats/src/lib.rs
```diff
@@ -200,7 +216,7 @@ impl NatsMessagingProvider {
     }
 }
 
-#[instrument(level = "debug", skip_all, fields(component_id = %component_id, subject = %nats_msg.subject, reply_to = ?nats_msg.reply))]
+#[instrument(skip_all, fields(component_id = %component_id, subject = %nats_msg.subject, reply_to = ?nats_msg.reply))]
```
rvolosatovs, line 219: This will make providers very noisy by default and I don't think that's nice for developer experience.
There are also many other instrumented calls in our dependency graph (including wRPC, but also e.g. wasmtime), that instrument only at trace or debug levels.

How about we just capture all logs instead (including trace level even) for OTEL?

vados-cosmonic: Sorry I'm not sure I understand -- by noisy do you mean that there will be many more spans emitted that ideal?
I'm not sure that providers should be at the level of wrpc or wasmtime -- those are much lower levels and deserve to be at trace/debug. 

These changes are based off of consistency with HTTP server provider -- do you by any chance remember why we felt the need to use the default log level there, but not in the other providers?

Could you lay out what you mean by "just capture all logs instead"? I think that would add a LOT of noise to OTEL traces that developers wouldn't want to get... People mostly want to know the levels *above* wRPC and wasmtime are working, not the levels below, I think.

vados-cosmonic: Removed this from the PR

rvolosatovs (approved):

brooksmtownsend (approved):
🚀

PR #1929, opened 2024-04-17
Test `http-keyvalue-counter` example

rvolosatovs:
## Feature or Problem

Add a unit test for `http-keyvalue-counter`

Would be nice to use Wadm as a library here instead of manually setting up entities

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ dump.rdb
 host_config.json
 node_modules
 target/
+examples/**/Cargo.lock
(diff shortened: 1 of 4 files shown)


rvolosatovs:
Currently blocked on #1931

brooksmtownsend (approved):
This looks great @rvolosatovs 🚀 

Are you interested in wadm being embeddable, so that we could use the library here to just deploy the manifest and then test from that? Or just a library to deploy the manifest and we can run wadm as a test binary?

rvolosatovs:
> Are you interested in wadm being embeddable, so that we could use the library here to just deploy the manifest and then test from that? Or just a library to deploy the manifest and we can run wadm as a test binary?

What I'd like to be able to do is something like `wadm::deploy(parsed_yaml)` where `parsed_yaml` is of `wadm::Spec` type (or whatever the thing stored in the manifest is called)

PR #1930, opened 2024-04-17
chore(examples): remove redundant world spec

rvolosatovs:
## Feature or Problem

Remove the redundant `world` specification for `wit-bindgen` invocations. It's not necessary when there's only one world in the package

### DIFF
--- a/examples/rust/components/blobby/src/lib.rs
+++ b/examples/rust/components/blobby/src/lib.rs
@@ -1,9 +1,7 @@
 #![allow(clippy::missing_safety_doc)]
 // ^^^ This is for clippy complaining about something that wit bindgen is generating
 
-wit_bindgen::generate!({
-    world: "blobby",
-});
+wit_bindgen::generate!();
 
 use std::io::{Read, Write};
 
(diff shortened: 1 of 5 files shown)


vados-cosmonic (dismissed):
Are we sure we want to do this? This is going to be worse for anyone looking at these to learn.

The fact that the world is auto-selected is not immediately obvious, but `world: "..."` makes it more explicit. We shouldn't depend on implicit functionality when someone completely new to the ecosystem might be showing up.

Maybe adding a comment to explain the mechanism would make sense -- `wit_bindgen` options and functionality are not very well documented at all, upstream.

rvolosatovs:
> Are we sure we want to do this? This is going to be worse for anyone looking at these to learn.
> 
> The fact that the world is auto-selected is not immediately obvious, but `world: "..."` makes it more explicit. We shouldn't depend on implicit functionality when someone completely new to the ecosystem might be showing up.

People completely new to the ecosystem are very unlikely to have more than one world defined in their WIT packages - it's an advanced feature and if anything, we should be steering developers away from it to avoid confusion.

In any case, these examples should not be using `wit-bindgen` at all - https://github.com/bytecodealliance/cargo-component is the official tool to build Rust components, especially for newcomers

vados-cosmonic:
In the interest of not holding up 1.0, we'll get this in and hopefully no one is confused by it

PR #1936, opened 2024-04-17
chore(examples): remove config put calls

brooksmtownsend:
## Feature or Problem
As of wadm v0.11.0, we no longer need to put config before you deploy the application.


vados-cosmonic (approved):
LGTM :rocket:

Issue #1888, opened 2024-04-12
[BUG] `component_scaled` events are published twice when a component stops


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

When a component is scaled to a max instances value of `0`, we emit two `component_scaled` events that assert that it was scaled to zero.

### Steps to reproduce

```bash
wash start component wasmcloud.azurecr.io/http-hello-world:0.1.0 asd
nats sub "wasmbus.evt.>"
wash stop component asd
```

### Expected behavior

I expect to see one event.

### Environment

- OS: [e.g. Linux, MacOS]
- Shell [e.g. bash, zsh, powershell]
- wasmcloud Version [e.g. 0.81.0]

yerzham:
Related to this error, when running the following command for the running component:
```bash
wash scale component <host-id> wasmcloud.azurecr.io/http-hello-world:0.1.0 asd
```
There are also two `component_scaled` events, but one asserts that it is unbounded (max_instances: 4294967295), and the consequent event asserts that it is scaled to 0.

Similarly, when using
```bash
wash update component asd wasmcloud.azurecr.io/http-hello-world:0.1.0
```
It emits `component_scaled` about the updated component, but emits two `component_scaled` events that an old component is scaled to 0.

I have identified the root issue and will create a pull request soon.

PR #1968, opened 2024-04-21
fix(host): remove extra component_scaled events

yerzham:
Signed-off-by: Yerzhan Zhamashev <yerzhan.contribs@novel.systems>

Similarly, an extra `component_scaled` event with `max_instances: 0` is emitted for `wash scale` and `wash update`.

## Related Issues
#1888 

## Release Information
`wasmCloud v1.0`

## Consumer Impact
Extra `component_scaled` event with `max_instances: 0` is no longer emitted when a component is updated or scaled (and scaled to 0 i.e stopped)

## Testing
All tests should work as expected

### Manual Verification
Listening to `wasmbus.evt.>` on NATS while performing `wash start`, `wash scale`, `wash update`, and `wash stop` on a component.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1438,7 +1438,7 @@ impl Host {
     }
 
     #[instrument(level = "debug", skip_all)]
-    async fn stop_actor(&self, actor: &Actor, host_id: &str) -> anyhow::Result<()> {
+    async fn stop_actor(&self, actor: &Actor, _host_id: &str) -> anyhow::Result<()> {
         trace!(component_id = %actor.id, "stopping component");
 
         // TODO: How to know if abort was successful?
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1441,7 +1441,6 @@ impl Host {
     async fn stop_actor(&self, actor: &Actor, _host_id: &str) -> anyhow::Result<()> {
         trace!(component_id = %actor.id, "stopping component");
 
-        // TODO: How to know if abort was successful?
         actor.calls.abort();
 
         Ok(())
@@ -1895,7 +1894,6 @@ impl Host {
             self.stop_actor(actor, host_id)
                 .await
                 .context("failed to stop old component")?;
-            // TODO: How to differentiate between old and new component if image reference is the same?
             self.publish_event(
                 "component_scaled",
                 event::component_scaled(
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1438,24 +1438,12 @@ impl Host {
     }
 
     #[instrument(level = "debug", skip_all)]
-    async fn stop_actor(&self, actor: &Actor, host_id: &str) -> anyhow::Result<()> {
+    async fn stop_actor(&self, actor: &Actor, _host_id: &str) -> anyhow::Result<()> {
         trace!(component_id = %actor.id, "stopping component");
 
+        // TODO: How to know if abort was successful?
```
brooksmtownsend, line 1444: This abort happens immediately and doesn't return a Result (and can't panic) so there's no additional feedback to gain here 😄 

If you mean in terms of knowing that a component finishes handling its current invocations, that might be worth an additional issue since that's something we could address

brooksmtownsend: I would prefer we remove this comment though unless it's tied to an active issue

crates/host/src/wasmbus/mod.rs
```diff
@@ -1907,6 +1895,7 @@ impl Host {
             self.stop_actor(actor, host_id)
                 .await
                 .context("failed to stop old component")?;
+            // TODO: How to differentiate between old and new component if image reference is the same?
```
brooksmtownsend, line 1898: An update should _only_ happen with separate image references. We should reject updates with the same image reference though, and that's a good issue to file if you're interested 😄 

Just like the above, I would recommend removing this comment unless it's tied to an issue

brooksmtownsend (changes requested):
image

With some testing, looks great! I left two comments + requests to remove the TODO comments if they aren't tied to an open issue so we don't accumulate questions around the codebase without action items 😄

crates/host/src/wasmbus/mod.rs
yerzham: I can take the initiative to check image references 👍

crates/host/src/wasmbus/mod.rs
yerzham: Removed the line

crates/host/src/wasmbus/mod.rs
yerzham: Removed the line

brooksmtownsend:
Thank you @yerzham ! It looks like the only other thing you need here is a DCO signoff, which you had in the first commit. The DCO failed action will give you the easy fix for this https://github.com/wasmCloud/wasmCloud/pull/1968/checks?check_run_id=24113309713

yerzham:
oops, will fix that

brooksmtownsend (approved):

brooksmtownsend:
Should be good to cut this as a patch release 🚀

PR #1977, opened 2024-04-23
chore: remove sample-manifest.yaml

vados-cosmonic:
## Feature or Problem

This commit removes `sample-manifest.yaml` which is now somewhat stale.


brooksmtownsend (approved):

brooksmtownsend:
For any curious, we removed support for this a while back in `wash` in favor of just committing to the wadm file 😄

PR #1989, opened 2024-04-26
feat(wash): Wash plugins

### LINKED ISSUE #841: [RFC] Support WebAssembly Components for CLI Plugins
## Summary
This RFC proposes a new feature for the `wash` CLI that would allow the discovery and usage of plugins at runtime. This would allow wash functionality to be extended both by wasmCloud maintainers and third parties wishing to build extensions and platforms. This RFC also recommends using WebAssembly to illustrate yet another use for Wasm in production and "in the wild".

## Rationale
The `wash` command line tool has been growing in power and complexity since it was first created. It can control lattices, build actors, create provider archives, and now it can even start a fully functioning wasmCloud host and an accompanying lattice all with just the `up` command. 

Adding new features and functionality to wash now involves working with a very large codebase and navigating through a number of intricate systems involved in obtaining NATS connections, reading contexts, and more. If someone just wanted to add a small piece of isolated functionality to `wash`, it might present so much of a barrier to entry that people may be discouraged from doing so.

We also want to avoid creating a _sprawl_ of multiple CLIs that all operate within the wasmCloud ecosystem. Keeping `wash` as the single source of CLI functionality for wasmCloud and higher-level platforms makes sense and keeps the developer experience as simple as possible.

## Use Cases
As mentioned, the wasmCloud maintainers would like to have an easier way of providing extensible functionality at the command line. The following is an incomplete list of other potential use cases.

* **[Concordance](https://github.com/cosmonic/concordance)** - Cosmonic's event sourcing system will soon be getting a number of additional features, including things like generating code, documentation, flow diagrams, and more. Rather than create an entirely new CLI for people to use, it would be extremely beneficial for Concordance users to be able to simply use a plugin that adds a new sub-command to wash, e.g. `wash es generate model.ttl` would generate appropriate output from the specified model.
* Custom App - If you've built an application on top of wasmCloud that has features that you'd like to be able to control from a CLI, this would be an ideal use.
* Custom Builds - If your application or components have special build requirements that regular `wash` doesn't know about, adding a subcommand for your builds would be useful
* Internal Corporate - If your enterprise or team within an enterprise is working with wasmCloud and needs utility functionality that can't be made public (and so therefore can't be part of core `wash`), then creating an internal enterprise plugin is the perfect solution

## Implementation
Plugins for `wash` would be implemented as Wasm components. Wash would act as a trimmed-down, bespoke wasmCloud host that exposed a small number of capabilities for these components. Because Wasm components can't do anything on their own, `wash` will give them a limited set of capabilities to work with.

The following is a list of what each of these components will be able to do:
* **File I/O** - WASI file I/O will be enabled for the component, `wash` will convert command input into I/O parameters allowing the component to read and write specific files/locations. Whether this comes from regular WASI or from a blob store interface is an implementation detail to be sorted out.
* **Messaging** - If a component is _signed_ with the messaging claim, wash will establish a NATS connection based on input parameters and make it available to the plugin
* **HTTP Client** - plugins will be able to perform as many HTTP requests as they need

### The Plugin Interface
The following is a hypothetical interface definition that would be _exported_ by wash plugins. This may not even compile or pass a validation check and is only here to give you an idea of how easy it could be to implement a plugin.

```wit
default interface wash-plugin {
    record command-identifier {
        name: string,
        description: string,
        version: string,
        sub-commands: list<string>
    }

    record command-options {
        input-files: option<list<string>>,
        output-files: option<list<string>>
    }

    get-command: func() -> result<command-identifier, string>
    get-help: func(cmdpath: list<string>) -> result<string, string>
    execute-command: func(cmdpath: list<string>, args: list<string>, opts: command-options) -> result<string, string>
}
```
The main reason we've chosen to provide `cmdpath` and `args` is so that the plugin developer doesn't have to deal with whitespace splitting and escaping, etc. Wash will take care of that and provide sanitized input. 

For example, if I wanted to use the Concordance plugin to generate an SVG from my model file, and `wash` had sanitized the input for me, the parameters to `execute-command` might be:

* cmdpath: [`es`, `generate`, `svg`]
* args: []
* opts: `{ input-files: ["./model.ttl"], output-files: ["./model.svg"] }`

Where the command I typed to wash might be:

```
$ wash es generate svg --input ./model.ttl --output ./model.svg
```

### Sample Help
This is what you might see when you get the main `wash` help with a few plugins installed (`es` and `mycorp` commands added by plugins).

```

Runtime environments:
  up           Bootstrap a wasmCloud environment
  down         Tear down a wasmCloud environment launched with wash up

Other:
  es           Interact with Concordance event sourcing
  mycorp       Access internal enterprise functionality from MyCorp, LLC

Options:
  -o, --output <OUTPUT>  Specify output format (text or json) [default: text]
  -h, --help             Print help
  -V, --version          Print version
```

## Bonus Features 🎉
The following is a list of some of the cool things we get "for free" as a bonus by embracing Wasm components as the plugin model.

* `wash plugin add https://cosmonic.azurecr.io/concordance-plugin:0.1.1` - Developers can easily add a plugin to wash simply by referencing an OCI registry
* Just copy any `.wasm` plugin into `~/.wash/plugins` and `wash` immediately becomes aware of the new functionality
* Can update an existing plugin with a new version specified with a new OCI or Bindle URL
* Security could be set up so that `wash` would prompt you to trust a plugin issuer the first time it is accessed - `You are about to run a plugin from XXXX. Do you trust this vendor?` ❓Since plugins are zero trust and capability-limited anyway, does this extra user trust just add unnecessary steps? 
* `wash plugin build` would be able to easily perform the necessary component linking to convert your Rust project into a ready-to-use wash plugin.

Regardless of developer-granted trust, `wash` would _always_ validate the embedded JWT for the plugin and reject any that have been tampered with, expired, or forged.

thomastaylor312:
This PR introduces wash plugins, built on top of the component model! There is now a wit interface that supports adding custom subcommands to `wash` via wasm components. Basically it is a thin wrapper around the `wasi:cli` world. This also adds a new `wash plugin` subcommand that lets you install (via local file, HTTP, or OCI) plugins, uninstall them, and list them. Please note that I did not get around to updating the ctl interface wit, so that has not been added yet. But as soon as we do, we will add that as an available import for plugins. See the individual commit messages for additional information.

I plan on adding documentation for this feature in a follow up PR to the docs repo.

Closes #841


crates/wash-lib/src/plugin/subcommand.rs
```diff
@@ -0,0 +1,188 @@
+wasmtime::component::bindgen!({
```
rvolosatovs, line 1: nit; it's usually a good idea to put these in a `mod bindings`

crates/wash-cli/src/util.rs
```diff
@@ -32,6 +39,52 @@ pub fn json_str_to_msgpack_bytes(payload: &str) -> Result<Vec<u8>> {
...
+        .context("Unable to read plugin directory")?;
+
+    let mut plugins = SubcommandRunner::new().context("Could not inititalize plugin runner")?;
+
+    // We load each plugin separately so we only warn if a plugin fails to load
+    while let Some(entry) = readdir.next_entry().await.transpose() {
+        let entry = match entry {
+            Ok(entry) => entry,
+            Err(e) => {
+                eprintln!("WARN: Could not read plugin directory entry. Skipping: {e:?}");
```
rvolosatovs, line 68: nit; shouldn't these be `tracing` calls?

crates/wash-cli/src/plugin.rs
```diff
@@ -0,0 +1,328 @@
...
+    if let Some(expected_digest) = compute_digest {
+        spinner.update_spinner_message(" Computing digest");
+        let mut digest = Sha256::new();
+        let data = tokio::fs::read(&temp_location)
+            .await
+            .context("Unable to read plugin data for digest computation")?;
+        digest.update(data);
+        let hash = format!("{:x}", digest.finalize());
+        let sanitized = expected_digest.trim().to_lowercase();
+        if hash != sanitized {
```
rvolosatovs, line 208: nit; this could be replaced by `ensure!`

crates/wash-cli/src/bin/wash.rs
```diff
@@ -355,3 +442,40 @@ async fn main() {
...
+
+/// Helper for loading plugins. This function does not return an error because it will exit the
+/// process in case of a fatal error. As such, this shouldn't be called after the CLI is parsed.
+async fn load_plugins() -> (SubcommandRunner, PathBuf) {
+    // We need to use env vars here because the plugin loading needs to be initialized before
+    // the CLI is parsed
+    let plugin_dir = match ensure_plugin_dir(std::env::var("WASH_PLUGIN_DIR").ok()).await {
+        Ok(dir) => dir,
+        Err(e) => {
+            eprintln!("Could not load wash plugin directory: {}", e);
```
rvolosatovs, line 454: why not `tracing::error`?

crates/wash-cli/src/bin/wash.rs
```diff
@@ -355,3 +442,40 @@ async fn main() {
...
+            std::process::exit(1);
```
rvolosatovs, line 455: what do you think about not exiting here and just letting `wash` run without plugins?
I would imagine to be able to run e.g. `wash up` even if plugin directory failed to get initialized (because my file system is read-only for example)

crates/wash-cli/src/bin/wash.rs
```diff
@@ -213,7 +220,86 @@ async fn main() {
...
+            // NOTE(thomastaylor312): This is a hack to get the raw args to pass to the plugin. I
+            // don't really love this, but we can't add nice help text and structured arguments to
+            // the CLI if we just get the raw args with `allow_external_subcommands(true)`. We can
+            // revisit this later with something if we need to. I did do some basic testing that
+            // even if you wrap wash in a shell script, it still works.
+            let args: Vec<String> = std::env::args().skip(1).collect();
+            let dir = match ensure_plugin_scratch_dir_exists(plugin_dir, id).await {
+                Ok(dir) => dir,
+                Err(e) => {
+                    eprintln!("Error creating plugin scratch directory: {}", e);
```
rvolosatovs, line 288: why not `tracing`?

rvolosatovs (approved):
Nice, this looks great!

crates/wash-cli/src/plugin.rs
```diff
@@ -0,0 +1,328 @@
...
+        .write(true)
+        .read(true)
+        .open(&temp_location)
+        .await
+        .context("Unable to create temp file for plugin download")?;
+
+    let (scheme, rest) = cmd
+        .url
+        .split_once("://")
+        .context("Invalid URL. It should contain a scheme (e.g. file://)")?;
```
brooksmtownsend, line 116: nit: does this mean `wash plugin install ./myplugin.wasm` is invalid and should be `wash plugin install file://./myplugin.wasm`? Could we just have the fallback here be an `unwrap_or_else(|_| (None, cmd.url))` and allow no schemes to be files?

As I'm writing this, I realize the difficulty in figuring out oci/http/file from just a plain string, so it makes sense. The help text addresses this well IMO so I'm just going to resolve this

crates/wash-cli/src/bin/wash.rs
```diff
@@ -213,7 +220,86 @@ async fn main() {
         .with_env_filter(EnvFilter::from_default_env())
         .init();
 
-    let cli: Cli = Parser::parse();
+    let mut command = Cli::command();
+    // Load plugins if they are not disabled
+    let plugins = if std::env::var("WASH_DISABLE_PLUGINS").is_err() {
```
brooksmtownsend, line 225: nit: I suppose specifying `WASH_DISABLE_PLUGINS=false` would still disable the plugins

brooksmtownsend: It would be great to have this environment variable documented at least, just a note on how defining it at all will disable plugins addresses both comments 👍🏻

crates/wash-lib/src/plugin/subcommand.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /// Run a subcommand with the given name and args. The plugin will inherit all
+    /// stdout/stderr/stdin/env. The given plugin_dir is used to grant the plugin access to the
+    /// filesystem in a specific directory, and should already exist. An error will only be returned
+    /// if there was a problem with the plugin (such as the plugin_dir not existing) or the
+    /// subcommand itself.
+    ///
+    /// All plugins will be passed environment variables starting with
+    /// `WASH_PLUGIN_${plugin_id.to_upper()}_` from the current process. Other vars will be ignored
```
brooksmtownsend, line 152: This will be good to add to documentation since it's likely folks will just expect the regular variable. I think this is a good measure to not give all plugins all-access

crates/wash-lib/tests/plugins/hello_plugin/src/lib.rs
brooksmtownsend, line 1: As a followup, it would be _awesome_ to add an example plugin to `examples/rust/components/` and reference it as a template, so people could do something like `wash new component --template-name wash-plugin-rust`

brooksmtownsend (commented):
I have mostly requests for surfacing docs from internal comments to either wasmcloud.com or in the CLI help text. This is awesome 🔥

crates/wash-cli/src/util.rs
thomastaylor312: I was doing them as warnings but I realized by default the logger is set to error level. These aren't fatal but I wanted them to be warnings the user would see so they knew why something wasn't working. @brooksmtownsend thoughts here?

crates/wash-cli/src/bin/wash.rs
thomastaylor312: This one was explicitly because I would expect a CLI to write an error to stderr if it was exiting with 1. But now that I think about it again, so does `tracing::error` so I'll just change it

crates/wash-cli/src/bin/wash.rs
thomastaylor312: Good point, I'll change it out to just write an error

crates/wash-lib/src/plugin/subcommand.rs
thomastaylor312: Yeah all of this will be added to documentation

crates/wash-lib/tests/plugins/hello_plugin/src/lib.rs
thomastaylor312: I can copy the test one to the examples directory as well.

brooksmtownsend (approved):

PR #2000, opened 2024-04-29
feat: update to Wasmtime 20

rvolosatovs:
## Feature or Problem

Update to Wasmtime 20

### DIFF
--- a/crates/runtime/src/actor/component/messaging.rs
+++ b/crates/runtime/src/actor/component/messaging.rs
@@ -16,7 +16,7 @@ pub mod messaging_handler_bindings {
         world: "messaging-handler",
         async: true,
         with: {
-           "wasmcloud:messaging/types@0.2.0": crate::capability::messaging::types,
+           "wasmcloud:messaging/types": crate::capability::messaging::types,
         },
     });
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -752,7 +752,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "miniz_oxide",
- "object",
+ "object 0.32.2",
  "rustc-demangle",
 ]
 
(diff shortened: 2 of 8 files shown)


crates/runtime/src/actor/component/messaging.rs
```diff
@@ -16,7 +16,7 @@ pub mod messaging_handler_bindings {
         world: "messaging-handler",
         async: true,
         with: {
-           "wasmcloud:messaging/types@0.2.0": crate::capability::messaging::types,
+           "wasmcloud:messaging/types": crate::capability::messaging::types,
```
brooksmtownsend, line 19: As we move forward versions, e.g. if we were to release a `wasmcloud:messaging/types@0.3.0`, how does this change affect that?

I assume both
- No breaking changes: It lets us support 0.3.0 without upgrading the host
- Breaking changes: We will need to upgrade the host in order to support 0.2.0 + 0.3.0 separately

This is just for one interface but I'm more interested in our broader strategy in order to support 0.2.0-versioned interfaces in the long term (like wasi-http when we start experimenting with the 0.3.0 draft versions)

brooksmtownsend (approved):
Upgrade looks good, this is a non-breaking change and can go out as a 1.0.1 right?

crates/runtime/src/actor/component/messaging.rs
brooksmtownsend: (this is a broader discussion and doesn't hold up this PR)

rvolosatovs: This change is unrelated to this process and concerns purely the shape of the host side bindings (structs, etc.)

In general, any missing import will be polyfilled.
For optimization of handling outdated interfaces, for which static bindings exist (e.g. support for `wasi:http 0.2.0`, when `wasi:http 0.3.0` is the latest) - it depends...
First of all we need to see what Wasmtime will do. I do expect for `wasi:http 0.2.0` support to be there in `0.3.0`, if not - I think we'll just add it ourselves by exporting `wasi:http 0.2.0` alongside `wasi:http 0.3.0`

rvolosatovs:
> Upgrade looks good, this is a non-breaking change and can go out as a 1.0.1 right?

yes, as long as our crates (like `wasmcloud-runtime`) are not included in 1.0 compat guarantee

brooksmtownsend:
> yes, as long as our crate API (like wasmcloud-runtime) are not included in 1.0 compat guarantee

Correct, the `wasmcloud-runtime` crate API is something that can technically have breaking changes as long as the wasmCloud binary isn't broken 😄

PR #2002, opened 2024-04-29
feat: limit max execution time to 10 minutes

rvolosatovs:
## Feature or Problem

limit max execution time to 10 minutes. This should eventually be configurable on binary level via CLI flags on env vars, perhaps even per-instance or per-component via control interface. This is just the first step


brooksmtownsend (approved):
This feels good as a general max, I was going to suggest 15 minutes to match the expectation for a service like AWS lambda but I don't see a reason to change it without evidence.

Agreed that in general we should be limiting this per-component or per-function for example based on timeouts that are more specific so a general max here is good

Issue #1996, opened 2024-04-29
[BUG] Blobby example, unknown "wasi:blobstore/blobstore" target


brianw-0:
### Affected project(s)

- [ ] documentation
- [X] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

This error occurs when running the Blobby example under wasmCloud/examples/rust/components. When the blobby component receives a curl command, it attempts to check if the blobstore container exists, and instead returns the error "unknown `wasi:blobstore/blobstore` target"

Nothing from the example code has been modified, so the wadm.yaml, wit, and source files remain the same as in the public repo.

### Steps to reproduce

1. wash up
2. wash app deploy wadm.yaml (from the blobby example folder)
3. curl [http://localhost:8080](http://localhost:8080)

### Expected behavior

I expected the example code to continue executing and create the corresponding container with name "default".

### Environment

* OS: Ubuntu 22.04.1
* Shell: bash
* wash: 0.27.0
* wasmcloud: 1.0.0

### Screenshots / Logs / Additional context

Running the commands from the wasmCloud/examples/rust/components/blobby directory:

```
> wash app deploy wadm.yaml
Successfully deployed model blobby v0.0.1

> curl localhost:8080
Error when communicating with blobstore: unknown `wasi:blobstore/blobstore` target
```

brooksmtownsend:
I didn't consider that merging the PR would mark this as "completed", I was able to make it work with this fix but please let me know if it works for you! If it doesn't we can reopen

brianw-0:
Thank you for the fast response!  I applied the fix but I still get a blobstore error (although slightly different this time): 
```
Error when communicating with blobstore: function failed: failed to lookup creation date: creation time is not available on this platform currently
```

brooksmtownsend:
Woah, that's surprising. That error is coming back from the standard Rust library, so it must be an issue we need to compensate for in our blobstore provider. I suppose we could just default at 0, or the UNIX epoch, if the OS doesn't support the creation time.

Thoughts @thomastaylor312 ? Not sure if you've seen this handled in a different way

thomastaylor312:
Honestly I think that fix will work just fine

brooksmtownsend:
[...]

PR #2005, opened 2024-04-29
fix/blobstore fs creation time

brooksmtownsend:
## Feature or Problem
This PR fixes an issue where creation time may not be a supported API on a platform, so we support a reasonable fallback.

## Related Issues
Addresses the second issue in #1996

## Release Information
blobstore-fs v0.6.1

### DIFF
--- a/crates/provider-blobstore-fs/src/lib.rs
+++ b/crates/provider-blobstore-fs/src/lib.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use std::io::SeekFrom;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::time::SystemTime;
+use std::time::{Duration, SystemTime};
 
 use anyhow::{anyhow, bail, Context as _};
 use bytes::{Bytes, BytesMut};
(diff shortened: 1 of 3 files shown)


crates/provider-blobstore-fs/src/lib.rs
```diff
@@ -261,10 +261,15 @@ impl Blobstore for FsProvider {
...
-                        .duration_since(SystemTime::UNIX_EPOCH)
-                        .context("creation time before Unix epoch")?;
+
+                    let created_at = if let Ok(created_time) = md.created() {
+                        created_time
+                            .duration_since(SystemTime::UNIX_EPOCH)
+                            .context("creation time before Unix epoch")?
+                    } else {
+                        // NOTE: Some platforms don't have support for creation time, so we default to the unix epoch
+                        std::time::Duration::from_secs(0)
```
rvolosatovs, line 271:
```suggestion
                        Duration::from_secs(0)
```

crates/provider-blobstore-fs/src/lib.rs
```diff
@@ -508,10 +513,15 @@ impl Blobstore for FsProvider {
...
```
rvolosatovs, line 523:
```suggestion
                        Duration::from_secs(0)
```

rvolosatovs (approved):
I would expect the error from `md.created()` to be logged as well (at least as debug)

brooksmtownsend:
@rvolosatovs left it at debug for now just because it could get annoying if at a `warn`, but we can level it up if needed

crates/provider-blobstore-fs/src/lib.rs
```diff
@@ -258,13 +258,22 @@ impl Blobstore for FsProvider {
...
-                        .duration_since(SystemTime::UNIX_EPOCH)
-                        .context("creation time before Unix epoch")?;
+
+                    let created_at = if let Ok(created_time) = md.created() {
+                        created_time
+                            .duration_since(SystemTime::UNIX_EPOCH)
+                            .context("creation time before Unix epoch")?
+                    } else {
+                        // NOTE: Some platforms don't have support for creation time, so we default to the unix epoch
+                        debug!(
```
rvolosatovs, line 271: nit; can we get the error printed here as well, otherwise this is not too helpful? (probably want to use a `match`)

rvolosatovs (approved):

crates/provider-blobstore-fs/src/lib.rs
brooksmtownsend: Oh, yes

PR #2006, opened 2024-04-29
feat(providers): allow missing default connection for redis

vados-cosmonic:
## Feature or Problem

Without a default connection the redis KV provider would normally fail to start -- this can be quite confusing considering usually a connetion is not expected to be made yet.

This commit refactors the conenction logic to allow running a provider without a pre-made default connection, and creates one upon the first connection of an actor that relies on the default
connection (i.e. doesn't have a connection specified via link config).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5974,7 +5974,7 @@ dependencies = [
 
 [[package]]
 name = "wasmcloud-provider-keyvalue-redis"
-version = "0.24.0"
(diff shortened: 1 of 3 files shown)


brooksmtownsend (approved):
Makes sense, good to let the provider run and only connect on behalf of a component 👍🏻

rvolosatovs (approved):

PR #2007, opened 2024-04-29
fix: http-server background task handle

brooksmtownsend:
- **fix(http-server): handle incoming requests in task**
- **chore(http-server): bump to 0.20.1**

## Feature or Problem
This PR changes the implementation of instantiating a new http-server incoming requests handler, such that the axum server is handling requests inside of a task. This was an issue as links from the HTTP server to a component would establish this server, then `await` on the server forever. Put another way, the HTTP server could only handle one server at a time because it was waiting for incoming requests.

## Release Information
http-server v0.20.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5959,7 +5959,7 @@ dependencies = [
 
 [[package]]
 name = "wasmcloud-provider-http-server"
-version = "0.20.0"
+version = "0.20.1"
 dependencies = [
  "anyhow",
  "async-nats",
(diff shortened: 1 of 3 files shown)


rvolosatovs (approved):

brooksmtownsend:
Going to push an additional commit to error log in the task in case it fails to bind

ricochet (approved):

thomastaylor312 (approved):

PR #2010, opened 2024-04-29
fix(providers): ignore URL config key vase for kv redis provider

vados-cosmonic:

### DIFF
--- a/crates/provider-keyvalue-redis/src/lib.rs
+++ b/crates/provider-keyvalue-redis/src/lib.rs
@@ -301,7 +301,11 @@ impl Provider for KvRedisProvider {
             source_id, config, ..
         }: LinkConfig<'_>,
     ) -> anyhow::Result<()> {
-        let conn = if let Some(url) = config.get(CONFIG_REDIS_URL_KEY) {
+        let conn = if let Some(url) = config
+            .keys()
+            .find(|k| k.eq_ignore_ascii_case(CONFIG_REDIS_URL_KEY))
+            .and_then(|url_key| config.get(url_key))
+        {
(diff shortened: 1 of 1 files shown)


joonas (approved):
awesome, thank you!

vados-cosmonic:
We did it in one place, and not the other @joonas... so the idea it should be in the provider SDK is probably worth visiting.

joonas:
@vados-cosmonic would it be reasonable to also updated the impacted examples to use a newer version (assuming we want to publish this change as a patch bump)?

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -301,7 +301,11 @@ impl Provider for KvRedisProvider {
             source_id, config, ..
         }: LinkConfig<'_>,
     ) -> anyhow::Result<()> {
-        let conn = if let Some(url) = config.get(CONFIG_REDIS_URL_KEY) {
+        let conn = if let Some(url) = config
+            .keys()
+            .find(|k| k.eq_ignore_ascii_case(CONFIG_REDIS_URL_KEY))
```
brooksmtownsend, line 306: but but this is O(n) not O(1)! What if I have millions of keys!?!?!?

brooksmtownsend (approved):
Nice quality of life improvement 👍🏻

crates/provider-keyvalue-redis/src/lib.rs
vados-cosmonic: I honestly did/am considering doing a check for both the original key and it lowercased... O(2) so much better than O(n), but we had this code "laying around" so to speak...

Issue #1882, opened 2024-04-11
[BUG] The directionality of the NATS messaging provider link is inverted


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

In #1881, I had to define my link from the echo component to NATS like so:
```
    - name: echo
      type: component
      properties:
        image: file://./build/echo_messaging_s.wasm
        id: echo
      traits:
        # Govern the spread/scheduling of the component
        - type: spreadscaler
          properties:
            replicas: 1
        - type: link
          properties:
            target: nats
            namespace: wasmcloud
            package: messaging
            interfaces: [consumer]
            target_config:
              - name: simple-subscription
                properties:
                  subscriptions: wasmcloud.echo
```

This is correct for the link on the `consumer` interface, but I should not set up subscriptions using this link. This should be a link that configures all of the publish and request operations that the component will do.

The NATS messaging provider is meant to invoke a component based on a subscription. The way you _should_ specify a link from the NATS provider to a component is like this (in wadm):
```yaml
    - name: nats
      type: capability
      properties:
        image: ghcr.io/wasmcloud/messaging-nats:canary
      traits:
        - type: link
          properties:
            target: echo
            namespace: wasmcloud
            package: messaging
            interfaces: [handler]
            source_config:
              - name: simple-subscription
                properties:
                  subscriptions: wasmcloud.echo
```

The configuration here should be the source configuration, since you're configuring NATS to subscribe and then invoke the component's `handler.handle-message` function. This should be fairly easily changed inside of the provider, we should be receiving link configuration both as a source and as a target to do the different setups.

### Steps to reproduce

The fact that you can use the manifest in https://github.com/wasmCloud/wasmCloud/pull/1881 to create a subscription is just slightly wrong, so deploying that is the issue.

### Expected behavior

All in all, the correct manifest should look like this (component -> NATS on `consumer`, NATS -> component on `handler` with subscription information):
```yaml
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: rust-echo-messaging
  annotations:
    version: v0.0.1
    description: "Echo demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)"
[...]

PR #2011, opened 2024-04-29
feat(provider-sdk)!: support directional link storage

brooksmtownsend:
- **feat(provider-sdk)!: support storing directional links**
- **refactor(provider-sdk)!: remove link_name**

## Feature or Problem
This PR adds support to the provider SDK to store links in 2 different hashmaps, one for links where the provider is the source and one for where the provider is the target. This directionality is necessary in order to properly store different configuration, WASI interfaces, and different links where the provider is the source.

I don't believe this is a breaking change technically but I did modify the logic for link storage so providers might unknowingly depend on that, so a minor bump is appropriate.

## Release Information
provider-sdk v0.5.0

### DIFF
--- a/crates/provider-sdk/src/provider.rs
+++ b/crates/provider-sdk/src/provider.rs
@@ -20,7 +20,9 @@ use tokio::{select, spawn, try_join};
 use tracing::{debug, error, info, instrument, trace, warn, Instrument as _};
 use wasmcloud_core::nats::convert_header_map_to_hashmap;
 use wasmcloud_core::rpc::{health_subject, link_del_subject, link_put_subject, shutdown_subject};
-use wasmcloud_core::{HealthCheckRequest, HealthCheckResponse, HostData, InterfaceLinkDefinition};
+use wasmcloud_core::{
+    HealthCheckRequest, HealthCheckResponse, HostData, InterfaceLinkDefinition, LatticeTarget,
+};
 
 #[cfg(feature = "otel")]
 use wasmcloud_core::TraceContext;
@@ -486,15 +488,14 @@ where
 {
     if ld.source_id == connection.provider_key {
         if let Err(e) = provider.delete_link_as_source(&ld.target).await {
-            error!(error = %e, "failed to delete link");
+            error!(error = %e, target = &ld.target, "failed to delete link to component");
         }
     } else if ld.target == connection.provider_key {
-        connection.delete_link(&ld.source_id).await;
         if let Err(e) = provider.delete_link_as_target(&ld.source_id).await {
-            error!(error = %e, "failed to delete link");
+            error!(error = %e, source = &ld.source_id, "failed to delete link from component");
         }
     }
-    Ok(())
+    Ok(connection.delete_link(&ld.source_id, &ld.target).await)
 }
 
 /// Handle provider commands in a loop.
@@ -563,8 +564,8 @@ async fn handle_provider_commands(
             req = link_put.recv() => {
                 if let Some((ld, tx)) = req {
                     // If the link has already been put, return early
-                    if connection.is_linked(&ld.source_id).await {
-                        warn!("Ignoring duplicate link put");
+                    if connection.is_linked(&ld.source_id, &ld.target).await {
+                        warn!(source = &ld.source_id, target = &ld.target, "Ignoring duplicate link put");
                     } else {
                         info!("Linking component with provider");
                         if let Err(e) = receive_link_for_provider(&provider, connection, ld).await {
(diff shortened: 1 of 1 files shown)


crates/provider-sdk/src/provider.rs
```diff
@@ -182,11 +184,10 @@ async fn subscribe_shutdown(
     quit: broadcast::Sender<()>,
     lattice: &str,
     provider_key: &str,
-    link_name: &str,
     host_id: &'static str,
 ) -> ProviderInitResult<mpsc::Receiver<oneshot::Sender<()>>> {
     let mut sub = nats
-        .subscribe(shutdown_subject(lattice, provider_key, link_name))
```
brooksmtownsend, line 189: note: as of 1.0 we've _always_ passed `"default"` as the link name for backwards compatibility. https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L2132

brooksmtownsend:
This should also assist with #1882 and solving this problem for all providers

crates/provider-sdk/src/provider.rs
vados-cosmonic: it might be nice to put that as a comment somewhere in the code, if you think it'd confuse someone in this area

crates/provider-sdk/src/provider.rs
```diff
@@ -351,6 +350,7 @@ async fn init_provider(name: &str) -> ProviderInitResult<ProviderInitState> {
         structured_logging,
         log_level,
         otel_config,
+        link_name: _link_name,
```
vados-cosmonic, line 353: was this named `_link_name` because we weren't using it before?

crates/provider-sdk/src/provider.rs
```diff
@@ -484,16 +482,16 @@ async fn delete_link_for_provider<P>(
...
         if let Err(e) = provider.delete_link_as_source(&ld.target).await {
-            error!(error = %e, "failed to delete link");
+            error!(error = %e, target = &ld.target, "failed to delete link to component");
         }
-    } else if ld.target == connection.provider_key {
-        connection.delete_link(&ld.source_id).await;
+    } else if ld.target == connection.provider_id {
         if let Err(e) = provider.delete_link_as_target(&ld.source_id).await {
-            error!(error = %e, "failed to delete link");
+            error!(error = %e, source = &ld.source_id, "failed to delete link from component");
```
vados-cosmonic, line 491: It's a bit weird since `source`s are almost always components (for now), but I've been calling this `component_id = source_id` to match the content of the error... 

Maybe someday we'll go through and be more consistently generic about what "source" means, but IMO if we're calling it a "component" then we should have `component_id` there

crates/provider-sdk/src/provider.rs
```diff
@@ -780,28 +778,50 @@ impl ProviderConnection {
...
-    /// Get the provider key that was assigned to this host @ startup
+    /// Get the provider key that was assigned to this host at startup
     #[must_use]
     pub fn provider_key(&self) -> &str {
-        &self.provider_key
+        &self.provider_id
     }
 
-    /// Stores component with link definition
+    /// Stores link in the [ProviderConnection], either as a source link or target link
```
vados-cosmonic, line 787:
```suggestion
    /// Stores link in the [`ProviderConnection`], either as a source link or target link
```

crates/provider-sdk/src/provider.rs
```diff
@@ -780,28 +778,50 @@ impl ProviderConnection {
...
+    /// depending on if the provider is the source or target of the link
     pub async fn put_link(&self, ld: InterfaceLinkDefinition) {
-        let mut update = self.links.write().await;
-        update.insert(ld.source_id.to_string(), ld);
+        if ld.source_id == self.provider_id {
+            self.source_links
+                .write()
+                .await
+                .insert(ld.target.to_string(), ld);
+        } else {
```
vados-cosmonic, line 795: The assumption here is that the provider ID is the target right? but what... if it isn't? Do we want to put that case in here, similar to `delete_link`?

crates/provider-sdk/src/provider.rs
```diff
@@ -780,28 +778,50 @@ impl ProviderConnection {
...
-    /// Deletes link
-    pub async fn delete_link(&self, component_id: &str) {
-        let mut update = self.links.write().await;
-        update.remove(component_id);
+    /// Deletes link from the [ProviderConnection], either a source link or target link
+    /// based on if the provider is the source or target of the link
+    pub async fn delete_link(&self, source_id: &str, target: &str) {
+        if source_id == self.provider_id {
+            self.source_links.write().await.remove(source_id);
+        } else if target == self.provider_id {
```
vados-cosmonic, line 808: That final `else`, maybe we print a warning? Seems extremely unlikely but

vados-cosmonic (approved):
LGTM 🚀 nothin' but nits

[EDIT] - NITS were resolved so we could get this merged, but feel free to look at/address them later if you want!

crates/provider-sdk/src/provider.rs
vados-cosmonic: resolving for 🚀

PR #2017, opened 2024-04-30
feat(host): handle invocations in tasks

rvolosatovs:

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1353,7 +1353,19 @@ impl Host {
                         };
                         if let Err(err) = timeout(
                             max_execution_time,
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1353,7 +1353,19 @@ impl Host {
...
+                                    actor
+                                        .handle_invocation(
+                                            context,
+                                            params,
+                                            result_subject,
+                                            &transmitter,
+                                        )
+                                        .await
+                                }
+                            }),
```
brooksmtownsend, line 1368: This fixes the issue that we had where an infinitely looping component would prevent other instances of that component from executing 👍🏻

brooksmtownsend (approved):

brooksmtownsend (approved):

PR #2022, opened 2024-04-30
fix(examples): Fixes issue with wit-deps pull for plugin

thomastaylor312:
This was follow up from making sure it worked after I cut the plugin wit release

### DIFF
--- a/examples/rust/components/wash-plugin-rust/wit/deps.lock
+++ b/examples/rust/components/wash-plugin-rust/wit/deps.lock
@@ -31,7 +31,7 @@ sha256 = "622bd28bbeb43736375dc02bd003fd3a016ff8ee91e14bab488325c6b38bf966"
 sha512 = "5a63c1f36de0c4548e1d2297bdbededb28721cbad94ef7825c469eae29d7451c97e00b4c1d6730ee1ec0c4a5aac922961a2795762d4a0c3bb54e30a391a84bae"
 
 [subcommand]
-url = "http://localhost:8000/test.tar.gz"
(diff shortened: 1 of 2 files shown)


examples/rust/components/wash-plugin-rust/wit/deps.lock
```diff
@@ -31,7 +31,7 @@ sha256 = "622bd28bbeb43736375dc02bd003fd3a016ff8ee91e14bab488325c6b38bf966"
 sha512 = "5a63c1f36de0c4548e1d2297bdbededb28721cbad94ef7825c469eae29d7451c97e00b4c1d6730ee1ec0c4a5aac922961a2795762d4a0c3bb54e30a391a84bae"
 
 [subcommand]
-url = "http://localhost:8000/test.tar.gz"
+url = "https://github.com/wasmCloud/wasmCloud/releases/download/wash-plugin-v0.1.0/wash-plugins-0.1.0.tar.gz"
```
rvolosatovs, line 34: Should we crate a separate repo for this WIT package?

rvolosatovs (approved):

examples/rust/components/wash-plugin-rust/wit/deps.lock
thomastaylor312: I honestly wasn't sure. That was my first thought but I was trying to keep things in the monorepo. Ideally I'd like to get to the point where we can just push a bare component with the wit and use that (with tools like `cargo component`)

brooksmtownsend (approved):

PR #2042, opened 2024-05-01
test(washboard): add a basic wash up test with wash board

vados-cosmonic:
## Feature or Problem

While this doesn't completely close out the related issues, it provides the first step to being able to easily write tests that test washboard functionality. The next steps should be:

- Adding ergonomic manipulation of the spawned wasmCloud instance(s)
- Adding a more substantial test with a `wadm.yaml` that is more complex (at the very least the HTTP provider and either `echo` or `http-jsonify`)

## Related Issues

#1377


lachieh:
I love the instance class. Being able to control wash from the fixture will be great. Is it possible to have the fixutre enabled by default?

Looks like the test fails because of a connection refused error. If you didn't see this in testing then I'm guessing that the wash web server isn't quite fast enough to start before the test loads. Small issue there is that `wash ui` will download fresh assets from the latest `washboard-v*` tag so it's not going test the branch code.

The project is already set up to use the dev server from vite to host the files so instead of using the `wash ui` command, so let's disable the ui on the `WasmCloudInstance` (see `playwirght.config.ts` > `webServer`).

Aside, do you have eslint in your editor? linting isn't part of the ci (yet) but there's a few flags. I'm pretty sure all of them are style related.

vados-cosmonic:
> I love the instance class. Being able to control wash from the fixture will be great. Is it possible to have the fixutre enabled by default?

Not that I can see looking at [the docs](https://playwright.dev/docs/test-fixtures#automatic-fixtures)...

> Looks like the test fails because of a connection refused error. If you didn't see this in testing then I'm guessing that the wash web server isn't quite fast enough to start before the test loads. Small issue there is that wash ui will download fresh assets from the latest washboard-v* tag so it's not going test the branch code.
>
> The project is already set up to use the dev server from vite to host the files so instead of using the wash ui command, so let's disable the ui on the WasmCloudInstance (see playwirght.config.ts > webServer).

Ahh thank you -- I'll do this, and not start the UI by default, and use the vite-managed server.

Maybe this deserves wider discussion later but I'm not sure `playwright` can do randomized ports for webservers... Which makes me wonder how it can run in parallel effectively with the webserver. Right now there's a default `PORT` (5173), but I would expect tests to spin up their own webservers on demand so we can parallelize...

If we're up for possibly de-coupling our test runner from our headless browser runner, I'd love to use `vitest`.

> Aside, do you have eslint in your editor? linting isn't part of the ci (yet) but there's a few flags. I'm pretty sure all of them are style related.

Ahh I was only running `prettier` from my editor, will run eslint as well!

lachieh:
Looks like the eslint settings weren't working anyway, so I've added [this PR](https://github.com/wasmCloud/wasmCloud/pull/2054) which allows it to eslint to run on the test files, and brings the noise down a little. A lot of them should be auto-fixable with IDE fix on save or `yarn lint:fix`.

I'm down to use vitest. I didn't realize they had a browser mode! If you want to do that in this PR, I'm all for it. I'd want to add the appropriate [Testing Library](https://testing-library.com/docs/) packages and eslint plugin too.

washboard-ui/packages/washboard-ui/tests/common/wasmcloud-instance.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  constructor(opts?: WasmCloudInstanceOpts) {
+    this.opts = opts ?? WasmCloudInstanceOpts.default();
+    this._uuid = uuidv1();
+  }
+
+  /** Return the UUID for this instance */
+  uuid(): string {
+    return this._uuid.slice();
+  }
```
lachieh, line 128: What's the need to do this?

vados-cosmonic: Ah thanks for pointing this out, I believe this was from when I was using the `Uint8Array` that comes out of the `uuid` package natively -- it's a string so I don't need this.

Will fix this and rebase!

vados-cosmonic:
hey @lachieh looks like `tests` was still not covered in `lint:*` scripts somehow -- I added it right inline there with `src`, hopefully that's fine

washboard-ui/package.json
```diff
@@ -24,6 +24,7 @@
     "@wasmcloud/prettier-config": "workspace:*",
     "eslint": "^8.57.0",
     "eslint-config-turbo": "^1.13.3",
+    "pino": "^9.0.0",
```
lachieh, line 27: Move this one into `washboard-ui/packages/washboard-ui/package.json` and then run `yarn` so that the lockfile is updated and then commit that also.

lachieh: That should resolve the build issue 👍🏻

vados-cosmonic: Ah thank you -- just fixed this!

lachieh (approved):

thomastaylor312 (approved):

PR #2052, opened 2024-05-03
fix(runtime): Fixes issue with running runtime on smaller hosts

thomastaylor312:
We found that on hosts with less then 2GB of memory (give or take a little) the host would fail to start when the pooling allocator started allocating virtual memory. You could get around this by setting `vm_overcommit` to 1 or 2, but this is not the best experience for those running in production. This fixes the issue by falling back to the dynamic memory allocator if setup fails.

We already have a plan to add some more of these tunables as options for the host, but those will be added in future PRs as a feature add

### DIFF
--- a/crates/runtime/src/runtime.rs
+++ b/crates/runtime/src/runtime.rs
@@ -29,6 +29,7 @@ pub struct RuntimeBuilder {
     max_execution_time: Duration,
     handler: builtin::HandlerBuilder,
     actor_config: ComponentConfig,
+    force_pooling_allocator: bool,
 }
 
 impl RuntimeBuilder {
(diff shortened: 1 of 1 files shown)


crates/runtime/src/runtime.rs
```diff
@@ -155,7 +157,7 @@ impl RuntimeBuilder {
         }
     }
 
-    /// Sets the maximum number of components that can be run simultaneously. Defaults to 20000
+    /// Sets the maximum number of components that can be run simultaneously. Defaults to 10000
```
brooksmtownsend, line 160: We should document this somewhere 🤔

brooksmtownsend (approved):

PR #2065, opened 2024-05-06
ref(wash): Change plugins to support arbitrary path access

thomastaylor312:
This allows plugins to mark arguments as paths so that they can be preopened and allowed in the component. This tries to walk a path between security and flexibility. If an argument is marked as a path, wash will allow full access to it if it is a directory and then limited access to a directory and full access to the file if it is a path. It isn't perfect due to the limited nature of preopens, but it does mean that the plugin will not get access to anything outside of its scratch dir without the user explicitly passing the path.

Once this is merged there will be two follow ups: one is a PR to this repo updating the example code and the other will be to the docs repo to update documentation on the security around paths


crates/wash-lib/tests/plugins/hello_plugin/src/lib.rs
```diff
@@ -89,25 +136,117 @@ impl RunGuest for HelloPlugin {
...
+    get_directories()
+        .into_iter()
+        .find_map(|(dir, dir_path)| {
+            println!(
+                "Looking to see if {dir_path} matches {}",
+                path.as_ref().to_str().unwrap()
+            );
+            path.as_ref().starts_with(dir_path).then_some(dir)
+            // (<std::string::String as std::convert::AsRef<Path>>::as_ref(&dir_path) == path.as_ref())
+            //     .then_some(dir)
```
brooksmtownsend, line 213: nit: leftover code

crates/wash-lib/src/plugin/subcommand.rs
```diff
@@ -171,18 +186,79 @@ impl SubcommandRunner {
...
+                        parent.to_str().unwrap().to_string(),
+                        DIRECTORY_DENY,
+                    )
+                }
+            };
+            println!(
+                "Host path: {}, guest_path: {}",
+                host_path.display(),
+                guest_path
+            );
```
brooksmtownsend, line 235: Should this be a log to avoid interfering with output? Or at least to stderr?

brooksmtownsend (changes requested):
Largely looks great, just one request for a println to be a log (and a nit in the test that's not really required)

crates/wash-lib/src/plugin/subcommand.rs
thomastaylor312: This was just here for debugging purposes. Will be gone shortly

crates/wash-lib/tests/plugins/hello_plugin/src/lib.rs
thomastaylor312: Also here for debugging

brooksmtownsend (approved):

Issue #1994, opened 2024-04-27
WIP: [RFC] Provide a secure runtime for Development Tools


raskyld:
### RFC Title

RFC: Provide a secure runtime for *Development Tools*

### Introduction

This RFC study whether wasmCloud could provide developers with a secure runtime for *Development Tools*.

*Development Tools* are programs used by developers throughout the software development lifecycle, it includes things like:

- Linters,
- CLIs,
- Compilers,
- Documentation auto-generators,
- Automations,
- Small local servers to serve their apps..

### Motivation

The main motivations are:

- 🔒 To help users securing their [Software Supply Chain][slsa] at the beginning of the chain (the *Producers*).
- 🌳 To foster an ecosystem of tools that leverage the platform-agnosticism provided by Webassembly and WASI.
- 🚀 To make wasmCloud a good candidate for a distributed and wasm-based build system (e.g. a distributed CI/CD platform where every step is a WASI Component), the tools used for local development could easily be run remotely!

In today's world, there are really few mechanism that prevents *Development Tools* to do nasty things. They tend to have too much capabilities on the developers' workstation. Yet, stories of legitimate package getting compromised are frequent, and their potential impact severe.

[The Deno Runtime's Security Model][deno-security] is an example of what should have been done from the beginning by our industry as a counter-measure to this attack vector. Sadly, Deno is not a language-agnostic runtime which may slow down its adoption.

*Wasmtime* and the Component Model are really good candidates for a runtime that brings security, interoperability and platform-agnosticism to *Development Tools*.

wasmCloud can be seen as a way to "clusterize" *Wasmtime*. As such, it has the potential to provide a platform for developers to build and run their [*DevOps Pipeline*][devops-pipeline]. There seems to be a demand for solutions allowing to use the same tools whether running locally or remotely in a CI/CD system, for instance:

- https://earthly.dev
- https://dagger.io

[slsa]: https://slsa.dev/spec/v1.0/terminology#software-supply-chain
[deno-security]: https://docs.deno.com/runtime/manual/basics/permissions
[devops-pipeline]: https://resources.github.com/devops/pipeline/

### Detailed Design

TODO

### Backwards Compatibility

TODO

### Alternatives Considered

TODO

### Unresolved Questions

TODO

### Conclusion

TODO

raskyld:
[...]

PR #2075, opened 2024-05-07
WIP: `wash run`

raskyld:
## Feature or Problem

We want to provide developers with a runtime specialized in creating sandboxes to run components capable of accessing their local workstation in a controlled way.

- [x] Implement a first version of the runtime in `wash-lib` capable of running `wasi:cli` component.
- [x] Start the CLI part in `wash-cli`, it should be a sub-command named `run`.
- [x] Implement a small `wasi:cli` example component and run it with `wash run <ref>`.
- [x] Start to map CLI flags to `CtxBuilder` invocations.
- [x] Enrich the example to test host access policies enforcement.
- [ ] Start to study how to support `wasi:http` components aswell.

## Related Issues

#1994 

## Release Information

WIP.

## Consumer Impact

None, it's a new feature.

## Testing

WIP.

### Unit Test(s)

WIP.

### Acceptance or Integration

WIP.

### Manual Verification

WIP.


raskyld:
I started to implement a bit but I still have to familiarize myself with the ecosystem and I am not a seasoned Rust developer so I open a WIP PR so I may improve and get feedbacks from anyone passing by! 🙏 Thanks everyone!

raskyld:
I already have some questions regarding the permissions set.

Namely:

- Is there any reason to limit `stdin/stdout/stderr` access?
- Should we put control over `random`? I was thinking about the case of hosts with low entropy but Idk how `random` is implemented by wasmtime for now. 
- Should we put control over monotonic and wall clock? I was thinking about timing attacks but I cannot figure out a real world scenario that would justify disabling access to the host clock by default.

crates/wash-cli/src/run.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+#[derive(Parser, Debug, Clone)]
+pub struct DirOptions {
+    /// Allow mutable access to a directory tree.
+    ///
+    /// Mutability means files can be deleted and created.
+    /// The format of this argument is:
+    ///
+    /// ```shell
```
raskyld, line 51: Those docs are not really pretty when rendered in the help message of the CLI 😢 

If anyone has idea to make those more user-friendly 🙏

crates/wash-cli/src/run.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+pub async fn handle_command(cmd: RunCommand, _output_kind: OutputKind) -> Result<CommandOutput> {
+    let runtime = LocalRuntime::new()?;
+    let mut ctx_builder = CtxBuilder::new();
+
+    ctx_builder.set_reference(cmd.reference.clone());
+    ctx_builder.wasi_ctx().args(&cmd.args);
+
+    ctx_builder.wasi_ctx().inherit_stdio();
```
raskyld, line 209: ❓ Should we give control over the stdin, stderr and stdout?

I don't see any reason to block those fd but maybe some of you have idea?
Maybe the most immediate reason is to not let components return https://en.wikipedia.org/wiki/ANSI_escape_code to avoid it controlling the user terminal?

crates/wash-cli/src/run.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    handle_env(&mut ctx_builder, &cmd.env);
+    handle_dir(&mut ctx_builder, &cmd.dir)?;
+
+    runtime.run(ctx_builder.build()?).await?;
+
+    // Q(raskyld): Is there a way to avoid producing outputs and let the run component handle stdout, err, in?
+    return anyhow::Ok(CommandOutput::from(""));
```
raskyld, line 217: ❓ Is there a way to avoid the default behavior of printing the `CommandOutput`?

The problem is that currently, it prints two new lines which may be a problem if you want to pipe things together 😬

crates/wash-lib/src/run/ctx.rs
```diff
@@ -0,0 +1,113 @@
...
+            wasi_ctx: self.wasi.build(),
+            oci_pull_options: self.oci_pull_options.take().unwrap_or_default(),
+            oci_cache_file: self.oci_cache_file.take(),
+            reference: self.reference.take().unwrap(),
+        })
+    }
+
+    /// Returns a [`WasiCtxBuilder`] that can be used to control access to most WASI
+    /// native interfaces.
+    pub fn wasi_ctx(&mut self) -> &mut WasiCtxBuilder {
```
raskyld, line 70: 💡 I am not sure whether we should expose the `WasiCtxBuilder` like that or decorates it with our own functions instead.

This approach is easy and simple but it forces me, for example, to re-publish parts of the wasmtime crate like the `wasmtime_wasi::{DirPerms, FilePerms}` to avoid `wash-cli` a direct dependency on wasmtime.

crates/wash-lib/src/run/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    pub async fn run(&self, context: Ctx) -> Result<()> {
+        let artifact = get_oci_artifact(
+            context.reference.clone(),
+            context.oci_cache_file.clone(),
+            context.oci_pull_options.clone(),
+        )
+        .await
+        .context("failed to pull the component")?;
```
raskyld, line 46: ❓ Is there a plan to harmonize how we resolve artifacts from strings?

I saw @thomastaylor312 re-implemented that in wash plugin so I am not sure which approach I should take.

raskyld (commented):
Small update on the current state of this PR 😄

For now, we have a really simple local runtime implemented in `wash-lib`.
It allows `wash-cli` to translate the *permission flags* into a `Ctx` which holds all the necessary information to run a component (its reference, OCI Pull options, permissions).

At the moment, only the `NetworkOptions` permissions are not implemented yet.

There is an example rust component `wash-run` whose goal is to test every WASI interfaces to see if the permissions are indeed enforced correctly. A JSON report of the available access is then produced and printed on stdout.

At the moment, only the arguments passed and the preopened dirs (and their content) is reported.

It should report:

- [x] Available arguments,
- [x] Available environment variables and their content (should it be controlled by a flag?),
- [x] Available preopened dirs and their direct children,
- [ ] Name resolution available?
- [ ] Test binding socket on a given range with a given protocol,
- [ ] Test connecting to given peers.

I think the structures that are serialized into the JSON report should be put in a crate to be re-used, by the component (indeed) but also, by testing code so we can test scenarios and compare the observed access with expected access.

Do not hesitate to give me early feedback 🙏

crates/wash-cli/src/run.rs
brooksmtownsend: In this case, it might be best to take the more verbose docs and publish them on https://wasmcloud.com under the command line reference for more information, leaving this to simply be (for example):
```
    /// Allow mutable access to a directory tree. https://wasmcloud.com/docs/cli/run#mutable
    pub mutable: Vec<String>,
```

crates/wash-lib/src/run/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+            wasmtime_wasi::bindings::Command::instantiate_async(&mut store, &component, &linker)
+                .await
+                .context("failed to instantiate the compiled component")?;
+
+        instance
+            .wasi_cli_run()
+            .call_run(&mut store)
+            .await
+            .context("failed to run the component")?
+            .map_err(|_| anyhow::anyhow!("the component has been run but returned an error"))
```
brooksmtownsend, line 69: Could we reuse any functionality from `wash-lib/src/plugin/subcommand.rs` when it comes to wasmtime Ctx / executing the run function? @thomastaylor312 might have some thoughts. I know linking is different but it might help here since in both cases we're calling `wasi:cli/run`

crates/wash-cli/src/run.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// Allow the component to bind to any interface and port using the UDP protocol.
+    #[clap(long = "allow-net-bind-udp-all")]
+    pub allow_bind_udp_all: bool,
+
+    /// Allow the component to bind to a specific interface and specific port ranges using the UDP protocol.
+    ///
+    /// The format expected is the same as expected of `--allow-net-connect`.
+    #[clap(long = "allow-net-bind-udp")]
+    pub allow_bind_udp: Vec<String>,
+}
```
brooksmtownsend, line 149: My general feedback is that this looks like a very robust set of options, and that might be a little overwhelming on the command line and in the code. What do you think about cutting some of these down to an initial set of flags that we can prove out the MVP, and then adding from there?

For context, I think about the use case of being able to run `wash run ./component` to be pretty constrained to components that are running with the `wasi:cli` world to start but then it can grow over time, but I still want to reconcile that with #1594 🤔 Do you have any thoughts on that?

brooksmtownsend:
Regarding your permissions questions:

> * Is there any reason to limit `stdin/stdout/stderr` access?
> * Should we put control over `random`? I was thinking about the case of hosts with low entropy but Idk how `random` is implemented by wasmtime for now.
> * Should we put control over monotonic and wall clock? I was thinking about timing attacks but I cannot figure out a real world scenario that would justify disabling access to the host clock by default.

Honestly for all of these I think that we can keep the access here, at least the purpose I have in my head is that `wash run` is a developer tool and not a real way you should run the component in a sensitive environment. Giving access to specific files is reasonable, and same with the network. Perhaps in the future we could deny these resources as a testing grounds for making sure applications don't fall over if they are restricted from accessing specific APIs but that can easily be a followup 🙂

crates/wash-lib/src/run/mod.rs
raskyld: There may be overlaps indeed, especially since a `wasi:cli/run` runtime could be used to run `subcommand` (as long as it gives control over the linking step).

When I looked at the plugin system, the main problem that motivated me to re-develop the abstractions is that I needed my API to give control over host permissions and ATM, the permissions are specifically made for the case of running sub commands.

Maybe we should write a `cli` + `http` runner whose API let the caller customise the permissions and extend the linking process and then use it in `Subcommand`? 🤔

crates/wash-cli/src/run.rs
raskyld: We may cut it down to:
-  `--allow-connect`: Allow the component to create client socket to any peers.
- `--allow-bind-all`: Allow the component to bind to any port with any protocol.
- `--allow-bind <range>`: Allow the component to bind to a range of port.

WDYT?

About reconciling it with `wash dev`, I don't have enough experience with the linking process of wasmCloud but would it means implementing every interface as host funcs or running providers locally?

raskyld:
Thanks a lot for your time reviewing my PR @brooksmtownsend 🙏

crates/wash-lib/src/run/mod.rs
brooksmtownsend: That makes sense. I know we wanted plugins to possibly have HTTP access, and eventually have NATS access in order to do things like query the state of the lattice and/or manipulate lattice things like starting components. I'll see if Taylor has thoughts here since the plugins were meant to coursely allow permissions.

I don't want to make the code more complicated re: re-use, but I want to make sure we aren't maintaining two separate wasi:cli/run implementations 😄

crates/wash-cli/src/run.rs
brooksmtownsend: For your use case are you using `wasi-sockets` directly, or `wasi-http`? I wanted to check since, with wasmCloud, we mostly use the HTTP implementation for networking, so my thought around `wash dev` was more to get to something that after you `wash run ./component.wasm` it could also run in wasmCloud, if that makes sense 🙂

crates/wash-lib/src/run/ctx.rs
thomastaylor312: Yeah wasmtime updates enough I'm not sure I'd want to expose this as part of our public API. However, I also don't mind bringing the wasmtime dep to wash-cli since we already bring it in transitively via wash-lib. Let me do a full review and see if I have any other ideas

crates/wash-lib/src/run/mod.rs
thomastaylor312: I really like using the URI approach like I did in wash plugins so it can be really clear (e.g. `file://` and `oci://`)

crates/wash-lib/src/run/mod.rs
thomastaylor312: Yeah I don't really know if there would be a lot we could reuse. Generally you can reuse the `Engine` but since this is a cli tool, you can't simultaneously be running plugin commands and the run command, so it doesn't really gain us anything. I'll take a look in just a second though

crates/wash-lib/src/run/mod.rs
raskyld: I am 100% aligned with you, what we lose in "simplicity" (because we now need to add an URI), we gain in determinism: the resolution of your artifact is now more "explicit" there is less "magic" involved.

crates/wash-cli/src/run.rs
raskyld: So my idea is to allow tools that are more "POSIX-y" to use sockets and modern tools that don't need to do low level stuff should use HTTP indeed.

raskyld:
I may try to make a more re-usable embedding of wasmtime which would be usable for the plugin system and wash dev.

brooksmtownsend:
@raskyld one thing I'm concerned about this PR is, at the end of the implementation, we end up with essentially [wasmtime run](https://docs.wasmtime.dev/cli-options.html#run). I totally see the value of having a secure runtime for a CLI that's interacting with the filesystem, sockets, and `wasi-cli`, and right now I think `wasmtime run` fills that bubble for the most part. I'm concerned with the current implementation here because `wash run` and deploying a wasmCloud app on a wasmCloud host feel like two totally different goals at the moment.

What I would be really interested in is to collaborate here, or in the RFC itself, on how we can make `wash run` into something that really improves the development of wasmCloud applications. Being a part of the wasmCloud shell (wash, hehe) should all roll up to the end goal of working with wasmCloud apps. To be fair, we don't really have a story for running a distributed `wasi-cli` component quite yet, since it's difficult to conceptualize remotely running a command line.

Does that make sense at all? I really appreciate the enthusiasm and I think that `wash run` can be a feature for rapid development and debugging of a component, and I think that using it for the goal of sandboxing a CLI is more inline with the goals of `wasmtime run`.

PR #2092, opened 2024-05-08
fix(wash-cli): reg test failure again

brooksmtownsend:
## Feature or Problem
Fixing a few more failures that I just assumed were good since I fixed the others.


thomastaylor312 (approved):

joonas:
Given https://github.com/wasmCloud/wasmCloud/pull/2104, this is probably no longer needed I think?

brooksmtownsend:
Ah I might have to change the test output here

PR #2101, opened 2024-05-08
feat(washboard): update lattice-client and related packages

lachieh:
### ✨ It's time! ✨ 

This PR adds a rework of the `@wasmcloud/lattice-client-core` and `@wasmcloud/lattice-client-react` libraries that are ready for release as packages on npm.

Notable changes are:
- **Parity with _most_ Lattice Control API features**
   The aim here is to allow the JS lattice client to be able to perform as many actions that can be performed with NATS requests.
- **Minimal external dependencies**
   Previously we were using RxJS to handle events and make state updates that came across the lattice. This has been removed and the only remaining dependency is the NATS Websocket library.
- **No state management**
   Initially, the client was keeping a copy of the lattice state in memory. This has been removed and the client is now solely focused on interactions, rather than state management. This responsibility should be handled by consumers of the client. We aim to have an implementation example included in washboard and may decide to release this as a separate library.
- **Built in documentation**
   I've tried to include as much documentation on the individual methods as possible. If you're experimenting with the client and have questions, please let us know in an issue or on slack.

 
### Reviewer notes
There are a few changes here so I've broken the major pieces of work into a few commits. They should be descriptive enough but please let me know if this needs some 'splainin'.


brooksmtownsend (approved):

washboard-ui/packages/lattice-client-react/src/context/use-lattice-client.tsx
```diff
@@ -0,0 +1,14 @@
+import {type LatticeClient} from '@wasmcloud/lattice-client-core';
+import * as React from 'react';
+import {LatticeClientContext} from '@/context/lattice-client-context';
+
+export function useLatticeClient(LatticeClient?: LatticeClient): LatticeClient {
```
yordis, line 5:
```suggestion
export function useLatticeClientContext(LatticeClient?: LatticeClient): LatticeClient {
```

yordis (approved):
🚀 💜 lets go

PR #2104, opened 2024-05-09
fix: Make wash push returned digest based on the pushed manifest

joonas:
## Feature or Problem

This addresses a couple of problems:
1. `wash push` was returning the manifest from the manifest-attached config, not the manifest itself.
2. `wash push` was passing in an empty hashmap for `annotations` even if there were none, which resulted in an empty `"annotations": {}` being stored on the registry, when it really should've just been omitted altogether

In addition, it adds logic to calculate the manifest digest locally based on an alphabetically sorted set of bytes to match the behavior of the most commonly available/widely deployed registry, [distribution](https://github.com/distribution/distribution).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5827,6 +5827,7 @@ dependencies = [
  "serde_json",
  "serde_with",
  "serde_yaml",
+ "sha2",
  "tempfile",
  "term-table",
  "test-case",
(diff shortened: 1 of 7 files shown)


thomastaylor312 (approved):
Good to go pending passing tests

PR #2122, opened 2024-05-13
feat(examples): Adds an http-client example

thomastaylor312:
We've been missing an example of the http-client (outgoing-response) interface for a while. This adds one that fetches you a random picture of a dog


crates/wash-lib/src/generate/favorites.toml
```diff
@@ -36,6 +36,12 @@ description = "a hello-world component (in Python) that responds over an HTTP co
 git = "wasmCloud/wasmCloud"
 subfolder = "examples/python/components/http-hello-world"
 
+[[component]]
+name = "rust-http-client"
```
vados-cosmonic, line 40: hey since this is a dog picture renderer... maybe we should call it a name that reflects that?

examples/rust/components/http-client/wadm.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: core.oam.dev/v1beta1
+kind: Application
+metadata:
+  name: rust-http-client
+  annotations:
+    version: v0.0.1
+    description: "HTTP hello world demo in Rust, showing use of the server and client providers"
```
vados-cosmonic, line 7: Maybe include some information about the dog pictures and/or websites it uses?

examples/rust/components/http-client/README.md
```diff
@@ -0,0 +1,38 @@
+# HTTP Hello World
+
+This is a simple Rust Wasm example that responds with a "Hello World" message for each request.
+
+## Prerequisites
+
+- `cargo` 1.75
+- [`wash`](https://wasmcloud.com/docs/installation) 0.27.0
+- `wasmtime` 19.0.0 (if running with wasmtime)
```
vados-cosmonic, line 9: Does the example not work with `20.x`?

examples/rust/components/http-client/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+                    };
+                    let _trailers = wasi::http::types::IncomingBody::finish(response_body);
+                    let dog_response: DogResponse = serde_json::from_slice(&body).unwrap();
+                    dog_response.message
+                } else {
+                    panic!("HTTP request failed with status code {}", response.status());
+                }
+            }
+            Ok(_) => {
+                panic!("Got response, but it wasn't ready");
```
rvolosatovs, line 52: I don't think that's the right behavior for an example - we should be showing best practices here, why not block on the pollable until it's ready instead?

examples/rust/components/http-client/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        let response = OutgoingResponse::new(Fields::new());
+        response.set_status_code(200).unwrap();
+        let response_body = response.body().unwrap();
+        response_body
+            .write()
+            .unwrap()
+            .blocking_write_and_flush(dog_picture_url.as_bytes())
+            .unwrap();
+        OutgoingBody::finish(response_body, None).expect("failed to finish response body");
+        ResponseOutparam::set(response_out, Ok(response));
```
rvolosatovs, line 67: This is normally done *before* the body is written, at least that's the expected use case

examples/rust/components/http-client/README.md
thomastaylor312: This is copy/pasta from http-hello. I can change

examples/rust/components/http-client/src/lib.rs
thomastaylor312: I have no idea how to do that and this is copied from what the plugin does (it just logs an error) which is copied from what our test does. Is blocking just looping until it is ready or is there some other mechanism I can use?

examples/rust/components/http-client/src/lib.rs
thomastaylor312: We should probably switch this in the http server example then as well

examples/rust/components/http-client/wadm.yaml
thomastaylor312: I don't know if that info needs to be here. I think of it more as a happy surprise when they look at the code

crates/wash-lib/src/generate/favorites.toml
thomastaylor312: Trying to go boring here so it is easier for people to understand what it does

examples/rust/components/http-client/src/lib.rs
thomastaylor312: Just checking here: Is there any weird behavior if we call `response.body()` and then pass ownership of `response` to `ResponseOutparam`?

examples/rust/components/http-client/src/lib.rs
thomastaylor312: Figured it out. `Pollable` just has `block`. Updating in all examples

examples/rust/components/http-client/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+            Err(e) => {
+                format!("Got error when trying to fetch dog: {}", e)
+            }
+        };
+        let response = OutgoingResponse::new(Fields::new());
+        response.set_status_code(200).unwrap();
+        let response_body = response.body().unwrap();
+        ResponseOutparam::set(response_out, Ok(response));
+        response_body
+            .write()
```
rvolosatovs, line 61: This should be called before `ResponseOutparam::set` - first you get a write handle and then use it after HTTP status code and headers are written

crates/wash-lib/src/generate/favorites.toml
```diff
@@ -36,6 +36,12 @@ description = "a hello-world component (in Python) that responds over an HTTP co
...
+description = "a hello-world component (in Rust) that responds over an HTTP connection with some data fetched from an upstream API"
```
rvolosatovs, line 41: What makes this a "hello-world" component?

crates/wash-lib/src/generate/favorites.toml
rvolosatovs: I don't feel too strongly about this, but I agree with @vados-cosmonic here.
The fact that something called "http-client" is actually an HTTP server, which sends a request to a hard-coded URL is certainly surprising to me.
This looks much more like an API proxy, so how about `http-proxy`? As a matter of fact, this component actually implements the `wasi:http/proxy` world

examples/rust/components/http-client/wit/world.wit
```diff
@@ -0,0 +1,7 @@
+package wasmcloud:hello;
+
+world hello {
+  import wasi:http/outgoing-handler@0.2.0;
```
rvolosatovs, line 4: Why not `include wasi:http/proxy@0.2.0;` instead?

crates/wash-lib/src/generate/favorites.toml
thomastaylor312: I know it is technically a proxy world, but it is hard to show the http client in isolation. Normally it is used in conjunction with something else. The whole intent of this example is to show how http client works since it is not super easy to use. I am not trying to show off that it is proxying a request – that is just a means to an end. For example, I could also do this inside of a messaging example, but I feel like that would confuse the situation

examples/rust/components/http-client/wit/world.wit
thomastaylor312: Basically my reason is the same as https://github.com/wasmCloud/wasmCloud/pull/2122#discussion_r1600250528. I am not trying to show how to implement the proxy world, I'm trying to show how you would use a client

crates/wash-lib/src/generate/favorites.toml
thomastaylor312: Reviewing our examples, I think we prefer describing what it does. I think @vados-cosmonic’s suggestion of something like dog-fetcher is more accurate (and funny)

vados-cosmonic: Resolving this so it doesn't gate!

vados-cosmonic (approved):
LGTM 🚀

.github/workflows/examples-components.yml
```diff
@@ -10,6 +10,7 @@ on:
       - component-http-hello-world-v[0-9]+.[0-9]+.[0-9]+*
       - component-http-keyvalue-counter-v[0-9]+.[0-9]+.[0-9]+*
       - component-http-jsonify-v[0-9]+.[0-9]+.[0-9]+*
+      - component-dog-fetcher-v[0-9]+.[0-9]+.[0-9]+*
```
rvolosatovs, line 13: nit; could we keep this list lexicographically sorted please?

examples/rust/components/dog-fetcher/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+        };
+        let response = OutgoingResponse::new(Fields::new());
+        response.set_status_code(200).unwrap();
+        let response_body = response.body().unwrap();
+        let write_stream = response_body.write().unwrap();
+
+        // Write the headers and then write the body
+        ResponseOutparam::set(response_out, Ok(response));
+        write_stream
+            .blocking_write_and_flush(dog_picture_url.as_bytes())
```
rvolosatovs, line 64: This will only write up to 4096 bytes. Probably want to call this in a loop or at least `assert` that the buffer is smaller than 4096 bytes

rvolosatovs (approved):

examples/rust/components/dog-fetcher/src/lib.rs
thomastaylor312: Yeah, I'll add in a comment. I didn't want to introduce more complexity about writing in these examples

Issue #1914, opened 2024-04-16
[RFC] Implement Postgres Interface


ricochet:
### RFC Title

RFC: Create a Postgres specific interface

### Introduction

Prior to 1.0, wasmCloud supported a `wasmCloud-sqldb` contract. Very similar to `wasmCloud-sqldb`, there is a proposal for `wasi:sql` world in the WASI space. Both API's have the goal of being generic across SQL providers, e.g. postgres and MySQL may use the same interface. The `wasi:sql` proposal has stalled mostly due to a lack of a desire by key stakeholders to implement a SQL interface that does not support semantic types for a given DB.

### Motivation

We need a solution for our users to connect to a relational database.

Postgres is one of the most popular DB's (according to StackOverflow surveys, the most popular). Instead of proposing to build interfaces and design those for each major SQL flavor, e.g. postgres, sqllite, and MySQL, I am proposing that we build and implement one SQL flavor (postgres) fully before approaching others. This pattern should be readily replicatable with key differences being the WIT representation of types.

### Detailed Design

Instead of using a `resource` to model a connection, instead use an interface that can later be "resourcified" once necessary CM features like WIT templates are available for declarative and statically verifiable multi-db connections. The types are elided in this example, but this package will define `postgres:types` that are used in the `query` and `execute` functions, e.g. a `Result` type.

```wit
package wasmcloud:postgres {
  interface connection {
    query: func(...) -> ...;
    execute: func(...) -> ...;
  }
```

### Backwards Compatibility

New API so no backwards compatibility

### Alternatives Considered

- Use `wasi:sql`. This proposal is in the earliest stage in WASI. There are major open questions about its viability concerning the generic approach used in wasi-sql. While many of the "wasi-cloud" API's are targeted for the 80% use-case, SQL is that one tricky area where if the solution is to "just use a string", the client implementations suffer from lossy casting to loosing type semantics.
[...]

PR #2129, opened 2024-05-14
feat(providers): add sqldb-postgres provider

vados-cosmonic:
## Feature or Problem

This commit adds a SQL database provider that supports the widely used, featureful Postgres database system.

One of Postgres's main benefits is a robust type system, and part of the work to make Postges possible to use from the wasmCloud ecosystem involves representing that robust type system in WIT -- see `bindings.rs` for more details.

This provider introduces multiple interfaces for making queries to Postgres cluster -- managed, named profile driven, and completely component-driven connections.

## To try it out
[Read `examples/rust/components/sqldb-postgres-managed/README.md`](https://github.com/wasmCloud/wasmCloud/blob/70fe4de4a20fbc436b0bea1957cb60b225cd6902/examples/rust/components/sqldb-postgres-managed/README.md), and follow the instructions (please raise issues if something doesn't work for you)


vados-cosmonic:
NOTE: I merged the example commit into this one -- while it's one *massive* PR, reviews should be manageable if we go by commit.

crates/provider-sqldb-postgres/README.md
```diff
@@ -0,0 +1,180 @@
+# 🐘 SQL Database Postgres Provider
+
+This capability provider implements the [wasmcloud:sqldb-postgres](https://github.com/wasmCloud/interfaces/tree/main/keyvalue) WIT package, which enables SQL-driven database interaction with a [Postgres][postgres] database cluster.
```
charypar, line 3: Link seems to point at a different interface?

charypar:
This is very exciting!

crates/provider-sqldb-postgres/README.md
vados-cosmonic: Ah thanks for noting this, I need to update that link -- we'll likely be publishing the WIT definition to another repo similar to how wasi-cloud does it, but we still need to discuss it a bit!

I'll leave myself a TODO so I don't forget your comment here

- [ ] fix the WIT interface link after WIT contract is iterated & published.

crates/provider-sqldb-postgres/README.md
```diff
@@ -0,0 +1,181 @@
...
+| `MANAGED_TLS_REQUIRED`        | `false`                     | Whether TLS should be required for al managed connections                  |
+| `PROFILE_<name>_URL`          | `postgres://some-db:5432`   | URL to use for all named profile connection `<name>`                       |
+| `PROFILE_<name>_HOST`         | `localhost`                 | Hostname to use for named profile connection `<name>`                      |
+| `PROFILE_<name>_PORT`         | `5432`                      | Port to use for named profile connection `<name>`                          |
+| `PROFILE_<name>_USERNAME`     | `postgres`                  | Username to use for named profile connection `<name>`                      |
+| `PROFILE_<name>_PASSWORD`     | `postgres`                  | Password to use for named profile connection `<name>`                      |
+| `PROFILE_<name>_TLS_REQUIRED` | `false`                     | Whether to require TLS for named profile connection `<name>`               |
+
+> ![NOTE]
+> As multiple components will connect to the same provider, the primary method of configuring connection "profiles" is via ENV variables
```
thomastaylor312, line 131: Especially for future compatibility when this could be a component, this should be done through provider config and not environment variables

crates/provider-sqldb-postgres/src/bindings.rs
```diff
@@ -0,0 +1,1114 @@
...
+                *octet0, *octet1, *octet2, *octet3, *octet4, *octet5, *octet6, *octet7,
+            )))
+        }
+        _ => bail!("unexpected number of bytes in received macaddress"),
+    }
+}
+
+/// Convert a 6byte MacAddress to a tuple of bytes
+#[allow(clippy::type_complexity)]
+fn macaddr8_to_bytes(m: MacAddress) -> anyhow::Result<(u8, u8, u8, u8, u8, u8, u8, u8)> {
```
thomastaylor312, line 159: Nit: Should this be an aliased type instead of a big tuple?

crates/provider-sqldb-postgres/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+            StatementPrepareError::Unexpected(format!("failed to build client from pool: {e}"))
+        })?;
+
+        let statement = client.prepare(query).await.map_err(|e| {
+            StatementPrepareError::Unexpected(format!("failed to prepare query: {e}"))
+        })?;
+
+        let statement_token = format!("prepared-statement-{}", Ulid::new().to_string());
+
+        let mut prepared_statements = self.prepared_statements.write().await;
```
thomastaylor312, line 144: Do we need to have something that cleans up prepared statements or should they continuously grow?

crates/provider-sqldb-postgres/README.md
thomastaylor312: However, I can't see in code where it is loading anything from the environment, so I am unsure if this is accurate or not?

crates/provider-sqldb-postgres/wit/connection.wit
```diff
@@ -0,0 +1,61 @@
+package wasmcloud:postgres@0.1.0-draft;
+
+/// Manual connection management
```
thomastaylor312, line 3: See my note in the README for my thoughts on manual connection

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
```diff
@@ -0,0 +1,102 @@
+package wasmcloud:postgres@0.1.0-draft;
+
+/// Interface for querying a Postgres database
+interface query {
```
thomastaylor312, line 4: We should probably just have this interface that uses the connection handle/token as that was described as the main idea here: https://github.com/wasmCloud/wasmCloud/issues/1914#issuecomment-2086514118. Mixing the two interfaces seems a bit odd

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
```diff
@@ -0,0 +1,102 @@
...
+  query: func(token: connection-token, query: string, params: list<pg-value>) -> result<list<result-row>, query-error>;
+}
+
+/// Managed queries are similar to regular queries, except that clients trust the
+/// provider/implementer to completely manage all connection/session semantics and information.
+///
+/// Connections must be set up and managed out-of-band, completely unknown to the calling component.
+interface managed-query {
+  use types.{pg-value, result-row};
+  use query.{query-error};
```
thomastaylor312, line 37: If we do keep the `managed-query` interface around, we should probably break out the `query-error` type into the types package. Otherwise if an implementor only wants to implement the `managed-query` interface, they'll still need to have a dependency on/implementation of the `query` interface

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
```diff
@@ -0,0 +1,102 @@
...
+    /// An otherwise known query execution error
+    query-error(query-error),
+    /// A completely unexpected error, specific to prepared statements
+    unexpected(string),
+  }
+
+  /// Prepare a statement, given a connection token (which can represent a connection *or* session),
+  /// to a Postgres database.
+  ///
+  /// Queries *must* be parametrized, with named arguments in the form of `$<integer>`, for example:
```
thomastaylor312, line 84: Nit: Spelling: parameterized

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
```diff
@@ -0,0 +1,102 @@
...
+///
+/// Connections must be set up and managed out-of-band, completely unknown to the calling component.
+interface managed-query {
+  use types.{pg-value, result-row};
+  use query.{query-error};
+
+  /// Query a Postgres database without a connection token, leaving connection/session management
+  /// to the callee/implementer of this interface
+  ///
+  /// Queries *must* be parametrized, with named arguments in the form of `$<integer>`, for example:
```
thomastaylor312, line 42: Nit: Spelling: parameterized

crates/provider-sqldb-postgres/wit/deps/postgres/types.wit
```diff
@@ -0,0 +1,262 @@
...
+  ///
+  /// ex. "America/New York", which is UTC-4 can be expressed as western-hemisphere-secs(4 * 3600)
+  variant offset {
+    eastern-hemisphere-secs(s32),
+    western-hemisphere-secs(s32),
+  }
+
+  /// Dates are represented similarly to tokio-postgres implementation
+  /// see: https://docs.rs/postgres-types/0.2.6/postgres_types/enum.Date.html#variant.Value
+  variant date {
```
thomastaylor312, line 65: Question: Would it be better to use wasi-clocks' [datetime type](https://github.com/WebAssembly/wasi-clocks/blob/6cfb0f8fc33000f29652657efe144953846f80ee/wit/wall-clock.wit#L18) and then convert to postgres times from there?

crates/provider-sqldb-postgres/wit/deps/postgres/types.wit
```diff
@@ -0,0 +1,262 @@
...
+    //
+    // fractional precision is determined by the database's `lc_monetary` setting.
+    //
+    // NOTE: if you are storing currency amounts, consider
+    // using integer (whole number) counts of smallest indivisible pieces of currency
+    // (ex. cent amounts to represent United States Dollars; 100 cents = 1 USD)
+    money(numeric),
+    money-array(list<numeric>),
+
+    // Postgres-internal
```
thomastaylor312, line 226: Can we get away with eliding these types for now or are we worried about full compatibility?

examples/rust/components/sqldb-postgres-managed/README.md
thomastaylor312, line 1: We already talked about this offline, but noting here: All of these demos could probably be combined into one with some comments

crates/provider-sqldb-postgres/README.md
vados-cosmonic: Ah sorry yeah this isn't accurate, I'm referring to the variables above but that's provider config -- they just *look* like ENV variables, will remove!

crates/provider-sqldb-postgres/src/lib.rs
vados-cosmonic: Prepared statements live for the lifetime of the connection that they were made on -- which is only when a link is dropped or shutdown, what were you thinking  here in terms of clean up?

crates/provider-sqldb-postgres/wit/connection.wit
vados-cosmonic: Saw this! Will resolve in favor of the other comment for discussion!

crates/provider-sqldb-postgres/wit/deps/postgres/types.wit
vados-cosmonic: These are useful for people wanting to do replication log things -- we cant actually support them in the *implementation* properly yet, but they should definitely be at the WIT level written down.

crates/provider-sqldb-postgres/wit/deps/postgres/types.wit
vados-cosmonic: I don't think so -- there's quite a difference between this and that type, also this type is much easier to construct, and more accurate/reflecting of postgres's understanding of dates.

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
vados-cosmonic: Sure! moving all errors to `types`

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
vados-cosmonic: Sorry could you explain what you mean exactly? Are you saying to remove `managed-query`?

crates/provider-sqldb-postgres/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+This capability provider implements the [wasmcloud:sqldb-postgres][wasmcloud-sqldb-postgres-wit] WIT package, which enables SQL-driven database interaction with a [Postgres][postgres] database cluster.
+
+This provider handles concurrent component connections, and components can choose how they manage connections:
+
+| Management type    | Description                                                       |
+|--------------------|-------------------------------------------------------------------|
+| Completely Managed | Defined at provider startup via host-supplied named configuration |
+| Link Driven        | Defined via config for each component-provider link               |
+| Component Driven   | Controlled completely by connected components                     |
```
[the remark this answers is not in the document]
thomastaylor312: I feel like those details can be modified at the link level for a component, correct?

crates/provider-sqldb-postgres/src/bindings.rs
thomastaylor312: Yeah, simple type alias

crates/provider-sqldb-postgres/src/lib.rs
thomastaylor312: The thing I'm trying to avoid is let's say a component makes a bunch of prepared statements but then only uses them once. If there are 200 components connected to this provider, all with several hundred prepared statements that are only used once, is that going to take up a bunch of memory? This could absolutely not be a problem, but just wanted to call it out

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
thomastaylor312: Yeah, I think we'll always have a "handle" to a connection managed by a provider/host and that is what is passed in the `query` interface (just like the `bucket` argument in keyvalue). So I don't think we need `managed-query`

crates/provider-sqldb-postgres/README.md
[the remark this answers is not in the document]
vados-cosmonic: So right now what you can do is *pick* the pre-existing profile -- specifying overrides is a new feature that hasn't really been considered up until now, AFAIK (or at least I haven't seen it!)

Then it's a question of *what* we want to override -- only DB name is probably pretty safe, so I can add that if that satisfies your concern here, and allows for people to at least implement some of that multi-tenancy pattern I noted earlier

crates/provider-sqldb-postgres/src/lib.rs
vados-cosmonic: So it's definitely not going to take up much memory... compared to the actual data being queried -- and also note that the prepared statement is more importantly *on the DB side*.

I don't think this is a problem we need to worry about first -- connection pooling params would be more important to start

crates/provider-sqldb-postgres/src/bindings.rs
vados-cosmonic: Done! Since we had to remove `eui48` (it's got a vuln in a dep), this actually turned into it's own struct so we could hang implementations off of it.

crates/provider-sqldb-postgres/README.md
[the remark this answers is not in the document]
LUK3ARK: For what it is worth - I prefer the way it is currently. The way I see it is a north star should be a WIT world should be able to manage use cases for each level of 'convention over composition'. For example:

- level 1 - with this interface, it seems to allow components to manage everything, which allows computations to be completely decoupled from the state the computation is executed on.
- level 2 - with this interface, it allows for components to explicitly couple a computation with the state it needs to execute on
- level 3 - with this interface, it allows for components to explicitly couple a computation with the state it needs to execute on, and allows to specify the expected output of the transform (result-rows)

Very happy with this interface!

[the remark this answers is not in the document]
thomastaylor312: @LUK3ARK So I definitely understand and agree, but only for custom interfaces (or product specific interfaces). What I don't want us to be in the business of is encouraging the component to do the same thing they are doing now: coupling specific connection string types or details in their actual code. Since this is going to be a generic interface that multiple things can implement, I am strongly against putting that style of management in a more generic, community supported contract, but fully supportive of a cloud provider or other third party/service provider having this kind of functionality if they so choose.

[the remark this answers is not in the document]
vados-cosmonic: Well the point that this might tempt people to put credentials in their WASM binaries (which are certainly not obfuscated in any way) is certainly quite bad (at the very least until [a good answer for moving secrets](https://github.com/wasmCloud/wasmCloud/issues/2190) arrives)...

The medium here is probably level 1 and 2 + overrides for profile driven connections (most importantly the DB name) -- we can leave the completely manual connection management unimplemented by this particular provider

crates/provider-sqldb-postgres/wit/deps/postgres/connection.wit
```diff
@@ -0,0 +1,50 @@
...
+    ///
+    /// NOTE: components probably won't need this, but more dynamic setups might, we can also disable this kind by default at the provider level
+    manual(connection-create-options),
+  }
+
+  /// A token that represents a connection of some sort,
+  /// either a raw database connection or a database session.
+  ///
+  /// This token can be expected to be somewhat opaque to users.
+  type connection-token = string;
```
esoterra, line 46: Why isn't this modeled as a `resource` if it is meant to be an opaque reference to something?

vados-cosmonic: Because IIRC resources aren't quite yet supported by the Rust tooling! Our Golang tooling actually leapfrogged our Rust and resources are supported but require some extra work right now -- one of the reasons this design is `0.1.0-draft`!

There are two things that we really should have that will come later:

- Resources (`resource connection`/`resource session`)
- Async (for `stream<result-row>`)

vados-cosmonic:
Chatted w/ @ricochet and the move here is that we want the `managed-query` interface to be the *only* interface. 

Together with @thomastaylor312 's feedback, the contract & provider is going to undergo the following changes:

- [ ] Remove `managed` and `manual` connection modes (and the connection creation machinery in general)
- [ ] Remove the pseudo-resource (`connection-token`, etc)

Connections will be configured via the combination of named config (`PROFILE_<name>_*`) and link configuration (selecting which profile a component should use).

crates/provider-sqldb-postgres/wit/deps/postgres/connection.wit
vados-cosmonic: Hey so note also that for the first version of this we're going to [simplify the interface *even* further](https://github.com/wasmCloud/wasmCloud/pull/2129#issuecomment-2138041188) -- we're going to remove the manual connection stuff as a possibility. 

In the future we *will* still want the `stream` stuff, but for this version you won't even need the pseudo-resource `connection-token`.

vados-cosmonic: I'm slightly tempted to implement a manual cursor interface as the return (rather than `list<result-row>`), but for now let's keep it simple and just wait for `stream<result-row>` to be possible

crates/provider-sqldb-postgres/wit/deps.toml
```diff
@@ -0,0 +1 @@
+postgres = "https://github.com/vados-cosmonic/wit-wasmcloud-postgres/archive/main.tar.gz"
```
ricochet, line 1: I think we'd want this interface in the wasmCloud repo right?

vados-cosmonic: Definitely! Once it's stabilized I'll be happy to move it over -- I'm thinking we can move it to `wasmcloud/wit-wasmcloud-postgres`?

crates/provider-sqldb-postgres/README.md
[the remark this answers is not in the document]
vados-cosmonic: Just want to note that as always I really appreciate you kicking the tires @LUK3ARK , helping us find bugs and think through corner cases and general design!

crates/provider-sqldb-postgres/wit/deps.toml
brooksmtownsend: I think matching it with the other wasi proposals would just be `wasmcloud/wasmcloud-postgres` right? Let me know if you need repos created/to accept a transfer

vados-cosmonic: Sure -- I personally like the prefixing of repos with what they contain but `wasmcloud/wasmcloud-postgres` works as well!

crates/provider-sqldb-postgres/wit/deps/postgres/types.wit
```diff
@@ -0,0 +1,299 @@
...
+interface types {
+
+  /// Errors that occur while executing queries
+  variant query-error {
+    /// Unknown/invalid query parameters
+    invalid-params(string),
+    /// Invalid/malformed query
+    invalid-query(string),
+    /// A completely unexpected error, specific to executing queries
+    unexpected(string),
```
thomastaylor312, line 13: Nit: Should we have a "No rows" error like what `database/sql` has in Go?

examples/rust/components/sqldb-postgres-query/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+> [!NOTE]
+> `wash up` will run as long as the host is running (you can cancel it with `Ctrl-C`)
+
+
+## Set up configuration for the provider
+
+Since configuration for Database clusters is usually sensitive information, we must pre-establish configuration for the provider using the [named configuration feature of wasmCloud][named-config].
+
+### Fully Managed connections
```
thomastaylor312, line 64: I think this documentation needs to be updated as well

examples/rust/components/sqldb-postgres-query/wadm.yaml
```diff
@@ -0,0 +1,42 @@
...
+            package: postgres
+            interfaces: [query]
+            target_config:
+              - name: default-postgres
+
+    # Add a capability provider that interacts with the filesystem
+    - name: sqldb-postgres
+      type: capability
+      properties:
+        #image: ghcr.io/wasmcloud/provider-sqldb-postgres:0.1.0
```
thomastaylor312, line 35: Let's create an issue to have this updated to be the default once we publish the provider

thomastaylor312 (changes requested):
Everything is pretty much ready to go, just some doc updates are needed

crates/provider-sqldb-postgres/src/config.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub username: String,
+    /// Password used when accessing the Postgres cluster
+    pub password: String,
+    /// Database to connect to
+    pub database: String,
+    /// Whether TLS is required for the connection
+    pub tls_required: bool,
+}
+
+impl TryFrom<ConnectionCreateOptions> for deadpool_postgres::Config {
```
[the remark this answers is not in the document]
vados-cosmonic: If there was I've forgotten -- fixed!

crates/provider-sqldb-postgres/wit/deps/postgres/types.wit
vados-cosmonic: I don't think so -- no rows being returned is not necessarily a bug...

examples/rust/components/sqldb-postgres-query/README.md
vados-cosmonic: You're right, done!

examples/rust/components/sqldb-postgres-query/wadm.yaml
vados-cosmonic: Sure -- I'm going to change this right now, and make a ticket to release the provider

thomastaylor312 (approved):

examples/rust/components/sqldb-postgres-query/wadm.yaml
vados-cosmonic: https://github.com/wasmCloud/wasmCloud/issues/2198

crates/provider-sqldb-postgres/wit/deps.toml
vados-cosmonic: Done! we're now using the following artifact:

https://github.com/wasmCloud/wasmCloud/releases/download/wit-wasmcloud-postgres-v0.1.0-draft/wit-wasmcloud-postgres-0.1.0-draft.tar.gz

ricochet (approved):

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
```diff
@@ -0,0 +1,48 @@
...
+  ///
+  /// Queries *must* be parameterized, with named arguments in the form of `$<integer>`, for example:
+  ///
+  /// ```
+  /// SELECT email,username FROM users WHERE uuid=$1;
+  /// ```
+  ///
+  /// NOTE: To see how to obtain a `connection-token`, see `connection.wit`.
+  ///
+  prepare: func(
```
esoterra, line 39: as in the question about connection tokens, why is this represented by a token and not a resource?

crates/provider-sqldb-postgres/wit/deps/postgres/connection.wit
esoterra: I think keeping it simple with a list and aiming for a stream later is fine.
That said, I have the same question for the prepared statements and filed a separate comment for that.

crates/provider-sqldb-postgres/wit/deps/postgres/query.wit
vados-cosmonic: Hey @esoterra the reason here is the same -- this is a token because right now arbitrary resources aren't quite covered by our host yet! It requires a little extra integration which IMO is a large price to pay for achieving some notion of idiomatic WIT use!

That said -- likely before official `0.1.0` or *maybe* a `0.2.0` we'll address *both* async and resource usage, but the utility of the interface doesn't diminish by much in the meantime.

PR #2131, opened 2024-05-15
chore(washboard): update execa dep

vados-cosmonic:


lachieh (approved):
@vados-cosmonic thank you! was going to ping you on this one. Looks like execa has a few other breaking changes that are killing the tests but once it passes you're good to merge 👍🏻

PR #2132, opened 2024-05-15
fix(wash-lib): Setup extra_root_certificates for OCI push client

joonas:
## Feature or Problem

Looking through the OCI push and pull code, I noticed that the client we set up for [pull has `extra_root_certificates` configured](https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-lib/src/registry.rs#L133-L141), but the one for push did not, so I felt like that would be worth remedying.

Presumably we'll want the same behavior in both cases.


vados-cosmonic (approved):
LGTM 🚀

PR #2135, opened 2024-05-15
Add option to skip certificate validation for the OCI registry connection

joonas:
## Feature or Problem

Sometimes it can be easier to simply ignore HTTPS certificate than try to ensure that it's correctly set up correctly in the OS  trust store (such as when you have self-signed certificates), this should make those scenarios easier to use any wash commands that interact with OCI registries.


ricochet (approved):

vados-cosmonic (approved):
LGTM 🚀

PR #2140, opened 2024-05-16
feat(wash-lib): add support for specifying multiple labels

vados-cosmonic:
## Feature or Problem

This commit adds support for specifying multiple labels to `wash label`.

Users can use `wash label <host-id> key1=value1,key2=value2` to set multiple labels on the host at the same time, in a best-effort manner


brooksmtownsend (approved):
looks nice! Wasn't there an issue we can close out from this?

PR #2141, opened 2024-05-16
feat(*): Updates wash and host to support new oci wasm manifests

### LINKED ISSUE #2038: [FEATURE] Add support for new wasm OCI artifact type
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

_No response_

With the release of the new OCI spec guidance, we want to make sure we implement that specification: https://github.com/cncf/tag-runtime/pull/135. This will need to be implemented in the host and in wash

_No response_

### Additional context

_No response_

thomastaylor312:
This PR contains two commits: one of them updates wash to only push the new type of wasm manifest, but supports pulling both types of manifest. This is technically a small breaking change for wash, but see the commit message for more details. The second commit updates the host to support the new wasm type as well as the old manifest type. This backwards compatibility will be in place until at least wasmCloud 2.0. 

Closes #2038

### DIFF
--- a/crates/host/src/oci.rs
+++ b/crates/host/src/oci.rs
@@ -12,6 +12,7 @@ use anyhow::{bail, Context as _};
 use oci_distribution::client::{ClientConfig, ClientProtocol, ImageData};
 use oci_distribution::secrets::RegistryAuth;
 use oci_distribution::{Client, Reference};
+use oci_wasm::{WASM_LAYER_MEDIA_TYPE, WASM_MANIFEST_MEDIA_TYPE};
 use serde::{Deserialize, Serialize};
 use tokio::fs;
 use tokio::io::AsyncWriteExt;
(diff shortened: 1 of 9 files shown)


crates/host/src/oci.rs
```diff
@@ -214,7 +228,10 @@ impl Fetcher {
     /// Returns an error if either fetching fails or reading the fetched OCI path fails
     pub async fn fetch_component(&self, oci_ref: impl AsRef<str>) -> anyhow::Result<Vec<u8>> {
         let path = self
-            .fetch_path(oci_ref, vec![WASM_MEDIA_TYPE, OCI_MEDIA_TYPE])
+            .fetch_path(
+                oci_ref,
+                vec![WASM_MEDIA_TYPE, OCI_MEDIA_TYPE, WASM_LAYER_MEDIA_TYPE],
```
mattwilkinsonn, line 233: Should these types be in `oci-wasm`?

mattwilkinsonn (approved):

crates/host/src/oci.rs
thomastaylor312: `WASM_MEDIA_TYPE` is a soon-to-be deprecated old type that exists in the oci-distribution crate (which it honestly shouldn't) that we have for backwards compatibility. I honestly don't know if we need `OCI_MEDIA_TYPE` any more since I don't think we pushed artifacts with that type, but I also kept it just in case

brooksmtownsend (approved):

PR #2142, opened 2024-05-16
chore: Bump oci-distribution to 0.11.0

joonas:
## Feature or Problem

We're currently using 0.9.4 from 2022

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3349,6 +3349,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "oci-distribution"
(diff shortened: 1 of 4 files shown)


thomastaylor312:
@joonas #2141 bumps it to 0.11. Should probably just bump to that instead

thomastaylor312 (changes requested):
Forgot to block on that last comment

thomastaylor312:
Only change is that you don't need a mutable client any more

thomastaylor312 (approved):

PR #2144, opened 2024-05-17
feat(provider-sdk): allow providers to access wit metadata for links

vados-cosmonic:
## Feature or Problem

This commit enables providers to easily access WIT meatadata related to a link via the `LinkConfig` struct available at link time.


crates/provider-sdk/src/lib.rs
```diff
@@ -120,6 +120,9 @@ pub struct LinkConfig<'a> {
 
     /// Configuration provided to the provider (either as the target or the source)
     pub config: &'a HashMap<String, String>,
+
+    /// WIT metadata for the link
+    pub wit_meta: (&'a WitNamespace, &'a WitPackage, &'a Vec<WitInterface>),
```
brooksmtownsend, line 125: Feels like a nit but I think it's valid, could we just call it `wit_metadata`? I think the longer form helps with parsing it in my brain. I would love to come up with a good name for what you call the ns, package, and interfaces but I can't think of one

vados-cosmonic: Will do -- yeah the longer form is also clearer

thomastaylor312 (approved):

PR #2155, opened 2024-05-20
ref(composition): Updates the composition README

thomastaylor312:
This adds a more simple way of composing using `wac plug` and keeps the old instructions around for advanced usage


brooksmtownsend (approved):
🔨 🔌

ericgregory (approved):

PR #2164, opened 2024-05-22
Configure reqwest with user-agent

joonas:
## Feature or Problem

Certain proxies will require authentication to be provided if the User-Agent header is missing when establishing a tunnel, so we should set a reasonable default here (`wasmcloud-core/<crate-version>` for now, we can adjust it later if need be) to enable usage in networks that are behind such proxies.

### DIFF
--- a/crates/core/src/tls.rs
+++ b/crates/core/src/tls.rs
@@ -64,10 +64,14 @@ pub static DEFAULT_HYPER_CONNECTOR: Lazy<
 #[cfg(all(feature = "reqwest", feature = "rustls-native-certs"))]
 pub static DEFAULT_REQWEST_CLIENT: Lazy<reqwest::Client> = Lazy::new(|| {
     reqwest::ClientBuilder::default()
+        .user_agent(REQWEST_USER_AGENT)
         .with_native_certificates()
         .build()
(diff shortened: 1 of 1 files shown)


rvolosatovs (approved):
LGTM, we should probably check that all codebase actually uses this one

`http-client` provider comes to mind, for example, I think it uses Hyper directly - not `reqwest`

rvolosatovs:
We probably want to set `User-Agent` header here if it's not set already https://github.com/wasmCloud/wasmCloud/blob/790cb148e088301656e8482a5a9b44fcf774f1c6/crates/provider-http-client/src/lib.rs#L57-L61

brooksmtownsend (approved):
I think in addition to the HTTP client, we should make sure that `wash` uses this when downloading artifacts and also `oci-distribution` internally can supply this for downloading components/providers.

Can be done as a followup of course

joonas:
Given that https://github.com/wasmCloud/wasmCloud/pull/2175 shipped with a separate client (and it's own User-Agent), I'm inclined to merge this as-is and do the User-Agent work for the http-client provider as a separate follow-up.

@rvolosatovs @brooksmtownsend any objections to taking that road?

Issue #1991, opened 2024-04-26
[FEATURE] Provide aliases for commonly shortened commands


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

`wash link delete` is a perfectly reasonable alternative to `wash link del`, and it stinks to error out at that point.

### Describe the solution you'd like

For the common commands that have shortened names, we should provide these common aliases

Aditya1404Sal:
I believe I have a solution for this, can I have this issue assigned please ? Thank you!

brooksmtownsend:
@Aditya1404Sal go for it! Feel free to ask questions here or just open up the PR when you're ready 😄

Aditya1404Sal:
https://github.com/wasmCloud/wasmCloud/blob/790cb148e088301656e8482a5a9b44fcf774f1c6/crates/wash-lib/src/cli/link.rs#L77-L90

```#[clap(name = "del", alias = "delete")]```
 
This should solve it, However I haven't written a test for it.

Sorry for the inconvenience but i have a question,
If I've made changes to the codebase locally and did the `make build` , is there any way to test out the cli changes in my terminal itself without writing tests? I looked through multiple READme's but couldn't find any clue to running a local copy of wash (if that's even possible) for development purposes.

**Comment by brooksmtownsend** [comment]
@Aditya1404Sal yep! This is built right into the Rust tool `cargo`. Inside of the `crates/wasi-cli` directory you can run `cargo run` to build and run wash (the binary will be under `target/debug/wash`). Just put arguments after a `--` in the CLI.

For example, to test the above you could try:
```bash
cargo run -- link del source ns package
cargo run -- link delete source ns package
```

Aditya1404Sal:
It works! Thanks a lot!
Should I add a section to the README.me explaining `cargo run` for new contributors, so that they won't face the same problem as I did ?

thomastaylor312:
@Aditya1404Sal That would be a great addition to the README!

Aditya1404Sal:
> @Aditya1404Sal That would be a great addition to the README! 

I'll get started on it ASAP !

brooksmtownsend:
[...]

PR #2165, opened 2024-05-22
Provides Alias for "wash link del <args>" as "wash link delete <args>"

Aditya1404Sal:
## Feature
Solves #1991

## Testing
Ran `cargo build` and  `cargo test` in .crates/wash-cli

### Manual Verification
ran `cargo run up` & `cargo run -- link delete source ns package` , got executed in the same way as `cargo run --link del source ns package`


brooksmtownsend (approved):
Looks great! I'd love to keep the issue open for now since del/delete is just one of the aliases to add 😄

PR #2173, opened 2024-05-23
feat(example): add blank slate rust provider

brooksmtownsend:


examples/rust/providers/blank-slate/src/provider.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        // If your provider has exports, uncomment the following connection and serve() call instead of
+        // `shutdown.await`. This will set up RPC topics for your provider's exports and await invocations.
+        // let connection = wasmcloud_provider_sdk::get_connection();
+        // serve(
+        //     &connection.get_wrpc_client(connection.provider_key()),
+        //     provider,
+        //     shutdown,
+        // )
+        // .await
```
vados-cosmonic, line 53: This is *probably* the more common case? Maybe this one should be uncommented and the other one should be commented?

examples/rust/providers/blank-slate/src/provider.rs
```diff
@@ -0,0 +1,154 @@
...
+        let config = ProviderConfig::from(&host_data.config);
+        BlankSlateProvider {
+            config,
+            ..Default::default()
+        }
+    }
+}
+
+impl Provider for BlankSlateProvider {
+    /// When your provider is linked to a component, this method will be called with the [`LinkConfig`] that
```
vados-cosmonic, line 67: I'm biased but I'd love an example/little intro method for `init()` here

examples/rust/providers/blank-slate/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Running
+
+Prerequisites:
+
+1. [Rust toolchain](https://www.rust-lang.org/tools/install)
+1. [nats-server](https://github.com/nats-io/nats-server)
+1. [nats-cli](https://github.com/nats-io/natscli)
+
+You can run this capability provider as a binary by passing a simple base64 encoded [HostData](https://docs.rs/wasmcloud-core/0.6.0/wasmcloud_core/host/struct.HostData.html) struct, in order to do basic testing. For example:
```
vados-cosmonic, line 22: This is excellent for quick testing, but what do you think about including a `wadm` file? Maybe with a commented out component that's attached to the provider?

examples/rust/providers/blank-slate/src/provider.rs
brooksmtownsend: _probably_, yeah, I just can't leave it uncommented if there's no exports in the wit 😄 

I may just add a sample export/import for that reason and bundle in a component that can make use of it

examples/rust/providers/blank-slate/README.md
brooksmtownsend: Love that idea!

examples/rust/providers/blank-slate/src/provider.rs
vados-cosmonic: Oh yeah IMO that is a much better idea -- vast majority of providers surely are going to... provide something, right?

examples/rust/providers/blank-slate/README.md
vados-cosmonic: Also one thing I think that took me a surprisingly long time to understand was how to specify named config from a WADM file versus source/target config (and maybe a note about how you shouldn't put secrets in the file) -- it's just... in a different spot, but for some reason I couldn't figure it out for a bit.

examples/rust/providers/blank-slate/src/provider.rs
brooksmtownsend: Oh! Good idea

vados-cosmonic:
Hey @brooksmtownsend what do you think about the name? `blank-state` is OK but maybe... `template` or `example` might be nice?

brooksmtownsend:
@vados-cosmonic Honestly I'm not too dedicated to blank slate 😄 Template is good, I think just something that provides the message that this is a sample but meant to be used as a "base" for a custom provider is perfect

brooksmtownsend:
> Oh also another thing I was thinking is that this is will be an input to wash new right? so a more generic name like that might be very useful/good for docs

Good point. Perhaps it could be `custom-template`, is that too generic?

PR #2174, opened 2024-05-24
Handle disconnected NATS connection better

lachieh:
## Feature or Problem
Found that the nats connection wasn't reconnecting if it failed to connect initially, or was disconnected for some reason. This makes the connection attempt to reconnect if a new request is made while it is disconnected/errored.


vados-cosmonic (approved):
LGTM :rocket:

brooksmtownsend (approved):

PR #2175, opened 2024-05-24
feat(wash-lib): Support configuring proxy credentials for HTTP(S)_PROXY when downloading artifacts

joonas:
## Feature or Problem

This makes it possible to specify basic authentication credentials to be used with the configured `HTTP(S)_PROXY` (or `http(s)_proxy`) environment variables by way of specifying `WASH_PROXY_USERNAME` and `WASH_PROXY_PASSWORD` environment variables.

If we think this is a reasonable approach, I'm also happy to extend this to the [`wash ui` command](https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-cli/src/ui/mod.rs#L66-L76)

### DIFF
--- a/crates/wash-lib/src/start/github.rs
+++ b/crates/wash-lib/src/start/github.rs
@@ -9,7 +9,10 @@ use std::{ffi::OsStr, io::Cursor};
 use tokio::fs::{create_dir_all, metadata, File};
 use tokio_stream::StreamExt;
 use tokio_tar::Archive;
-use wasmcloud_core::tls;
+use wasmcloud_core::tls::NativeRootsExt;
+
+const DOWNLOAD_CLIENT_USER_AGENT: &str =
+    concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
 
 /// Reusable function to download a release tarball from GitHub and extract an embedded binary to a specified directory
 ///
@@ -35,7 +38,7 @@ where
 {
     let bin_path = dir.as_ref().join(bin_name);
     // Download release tarball
-    let body = match tls::DEFAULT_REQWEST_CLIENT.get(url).send().await {
+    let body = match get_download_client()?.get(url).send().await {
         Ok(resp) => resp.bytes().await?,
         Err(e) => bail!("Failed to request release tarball: {:?}", e),
     };
@@ -86,4 +89,33 @@ where
     metadata(dir.as_ref().join(bin_name))
         .await
         .map_or(false, |m| m.is_file())
+}
+
+/// Helper function to set up a reqwest client for performing the download
+pub(crate) fn get_download_client() -> Result<reqwest::Client> {
+    let proxy_username = std::env::var("WASH_PROXY_USERNAME").unwrap_or_default();
+    let proxy_password = std::env::var("WASH_PROXY_PASSWORD").unwrap_or_default();
+
+    let mut builder = reqwest::ClientBuilder::default()
+        .user_agent(DOWNLOAD_CLIENT_USER_AGENT)
+        .with_native_certificates();
+
+    if let Ok(http_proxy) = std::env::var("HTTP_PROXY").or_else(|_| std::env::var("http_proxy")) {
+        let mut proxy = reqwest::Proxy::http(http_proxy)?.no_proxy(reqwest::NoProxy::from_env());
+        if !proxy_username.is_empty() && !proxy_password.is_empty() {
+            proxy = proxy.basic_auth(&proxy_username, &proxy_password);
+        }
+        builder = builder.proxy(proxy);
+    }
+
+    if let Ok(https_proxy) = std::env::var("HTTPS_PROXY").or_else(|_| std::env::var("https_proxy"))
+    {
+        let mut proxy = reqwest::Proxy::https(https_proxy)?.no_proxy(reqwest::NoProxy::from_env());
+        if !proxy_username.is_empty() && !proxy_password.is_empty() {
+            proxy = proxy.basic_auth(&proxy_username, &proxy_password);
+        }
+        builder = builder.proxy(proxy);
+    }
+
+    Ok(builder.build()?)
 }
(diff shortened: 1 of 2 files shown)


crates/wash-lib/src/start/github.rs
```diff
@@ -87,3 +90,41 @@ where
...
+    let mut builder = reqwest::ClientBuilder::default()
+        .user_agent(DOWNLOAD_CLIENT_USER_AGENT)
+        .with_native_certificates();
+
+    if !http_proxy_addr.is_empty() {
+        let mut proxy = reqwest::Proxy::http(http_proxy_addr).expect("failed to create http proxy");
+        if !proxy_username.is_empty() && !proxy_password.is_empty() {
+            proxy = proxy.basic_auth(&proxy_username, &proxy_password);
+        }
+        builder = builder.proxy(proxy);
```
vados-cosmonic, line 117: NIT: I think you could *probably* pull these lines out?

crates/wash-lib/src/start/github.rs
```diff
@@ -87,3 +90,41 @@ where
         .await
         .map_or(false, |m| m.is_file())
 }
+
+/// Helper function to set up a reqwest client for performing the download
+fn get_download_client() -> reqwest::Client {
+    let proxy_username = std::env::var("WASH_PROXY_USERNAME").unwrap_or_default();
+    let proxy_password = std::env::var("WASH_PROXY_PASSWORD").unwrap_or_default();
+    // TODO(joonas): should we consider system proxy configuration here, or just explicit config?
```
vados-cosmonic, line 98: Is this still relevant? specifying `http_proxy` is pretty common for specifying configuration of a proxy, but not sure which system proxy configuration you're referring to here

crates/wash-lib/src/start/github.rs
```diff
@@ -87,3 +90,41 @@ where
...
+    if !https_proxy_addr.is_empty() {
+        let mut proxy =
+            reqwest::Proxy::https(https_proxy_addr).expect("failed to create https proxy");
+        if !proxy_username.is_empty() && !proxy_password.is_empty() {
+            proxy = proxy.basic_auth(&proxy_username, &proxy_password);
+        }
+        builder = builder.proxy(proxy);
+    }
+
+    builder.build().expect("failed to build HTTP client")
```
vados-cosmonic, line 129: would be nice to bubble up the error here and not turn it into a panic!

vados-cosmonic (changes requested):
Almost there, only one serious concern which is the panic!

crates/wash-lib/src/start/github.rs
joonas: it's probably not relevant here for this particular use case, for context [reqwest](https://github.com/seanmonstar/reqwest/blob/master/src/proxy.rs#L284-L294) has this way of detecting what proxies might be set up at the system level, but I think given the intent of this feature, we probably don't actually care about that since we're expecting that an explicit `HTTP(S)_PROXY` is configured here.

I'll go ahead and remove the TODO, that's a good callout.

crates/wash-lib/src/start/github.rs
joonas: that's fair, and I agree that would probably be the right thing to do here, my intent was to preserve the [current/existing implementation's handling](https://github.com/wasmCloud/wasmCloud/blob/main/crates/core/src/tls.rs#L65-L70), but I agree that bubbling up the error would be better.

crates/wash-lib/src/start/github.rs
joonas: So the intention is to create multiple proxies (`HTTP_PROXY` vs `HTTPS_PROXY`) while also preserving the `NO_PROXY` based on the environment variables, not just one or the other.

So the [Config inside of the ClientBuilder](https://github.com/seanmonstar/reqwest/blob/v0.12.4/src/async_impl/client.rs#L187) keeps [a `Vec<reqwest::Proxy>` of proxies](https://github.com/seanmonstar/reqwest/blob/v0.12.4/src/async_impl/client.rs#L107) that then gets passed around as necessary, which is why I am setting up a separate Proxy for each of the different schemes, because each Proxy instance can only intercept traffic for the ProxyScheme it was created for. It's also important to note that each configured Proxy should get the basic auth credentials configured for it if it's set, because we won't know which Proxy those credentials should apply to (it seems reasonable to assume that they would apply to both if set).

crates/wash-lib/src/start/github.rs
joonas: I've made these into `Result<...>` instead, thanks for the callout!

crates/wash-lib/src/start/github.rs
```diff
@@ -87,3 +90,37 @@ where
...
+/// Helper function to set up a reqwest client for performing the download
+pub(crate) fn get_download_client() -> Result<reqwest::Client> {
+    let proxy_username = std::env::var("WASH_PROXY_USERNAME").unwrap_or_default();
+    let proxy_password = std::env::var("WASH_PROXY_PASSWORD").unwrap_or_default();
+    let http_proxy_addr = std::env::var("HTTP_PROXY")
+        .or_else(|_| std::env::var("http_proxy"))
+        .unwrap_or_default();
+    let https_proxy_addr = std::env::var("HTTPS_PROXY")
+        .or_else(|_| std::env::var("https_proxy"))
+        .unwrap_or_default();
```
brooksmtownsend, line 103: nit: I think keeping these as `Option<String>`'s will keep the code a little cleaner, and then later using an `if let` would work well.
```rust
if let Some(http_proxy) = http_proxy_addr {}
```

brooksmtownsend: You could even fetch the env var in the same statement if you don't need it after this block
```rust
if let Some(http_proxy_addr) = std::env::var("HTTP_PROXY")
        .or_else(|_| std::env::var("http_proxy")) {}
```

brooksmtownsend (commented):
One question about `NO_PROXY`, and a code nit 😄

joonas:
@vados-cosmonic if you wouldn't mind another look to make sure your feedback was adequately addressed, I'd love to move this forward :)

vados-cosmonic (approved):
LGTM :rocket:

PR #2176, opened 2024-05-24
fix(providers): Correct http-client provider README, reference hyper instead of reqwest

joonas:
## Feature or Problem

The README is out of date with regards to the underlying implementation, so let's fix that.

### DIFF
--- a/crates/provider-http-client/README.md
+++ b/crates/provider-http-client/README.md
@@ -1,6 +1,6 @@
 # HTTP Client Capability Provider
 
-This capability provider implements the `wasmcloud:httpclient` capability contract, and enables an component to make outgoing HTTP(s) requests. It is implemented in Rust using the [hyper](https://hyper.rs/) library.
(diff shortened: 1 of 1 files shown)


crates/provider-http-client/README.md
```diff
@@ -1,6 +1,6 @@
 # HTTP Client Capability Provider
 
-This capability provider implements the `wasmcloud:httpclient` capability contract, and enables an component to make outgoing HTTP(s) requests. It is implemented in Rust using the [reqwest](https://docs.rs/reqwest) library.
+This capability provider implements the `wasmcloud:httpclient` capability contract, and enables an component to make outgoing HTTP(s) requests. It is implemented in Rust using the [hyper](https://hyper.rs/) library.
```
brooksmtownsend, line 3: While we're here 😄 
```suggestion
This capability provider implements the `wasi:http/outgoing-handler` interface, and enables an component to make outgoing HTTP(s) requests. It is implemented in Rust using the [hyper](https://hyper.rs/) library.
```

brooksmtownsend (changes requested):

PR #2178, opened 2024-05-24
feat(wash-cli): `wash up --wadm-manifest`

vados-cosmonic:
## Feature or Problem

This commit introduces the functionality for `wash up --wadm-manifest <path to manifest>`, which starts a host and immediately deploys an application defined in a WADM manifest, cleaning it up when the `Ctrl+C` is hit, along with normal host termination.

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -220,10 +220,8 @@ async fn deploy_model(cmd: DeployCommand) -> Result<DeployModelResponse> {
 
     // If --replace was specified, we should attempt to replace the resources by deleting them beforehand
     if cmd.replace {
-        if let (Ok(name), version) = (
-            get_model_name(&app_manifest),
-            get_model_version(&app_manifest).map(Into::into),
-        ) {
+        if let (Some(name), version) = (app_manifest.name(), app_manifest.version().map(Into::into))
+        {
             if let Err(e) =
                 wash_lib::app::delete_model_version(&client, lattice.clone(), name, version, false)
                     .await
@@ -233,10 +231,19 @@ async fn deploy_model(cmd: DeployCommand) -> Result<DeployModelResponse> {
         }
     }
 
-    match app_manifest {
+    deploy_model_from_manifest(&client, lattice, app_manifest, cmd.version).await
+}
+
+pub(crate) async fn deploy_model_from_manifest(
+    client: &async_nats::Client,
+    lattice: Option<String>,
+    manifest: AppManifest,
+    version: Option<String>,
+) -> Result<DeployModelResponse> {
+    match manifest {
         AppManifest::SerializedModel(manifest) => {
             let put_res = wash_lib::app::put_model(
-                &client,
+                client,
                 lattice.clone(),
                 serde_yaml::to_string(&manifest)
                     .context("failed to convert manifest to string")?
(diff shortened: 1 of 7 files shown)


crates/wash-cli/src/up/mod.rs
```diff
@@ -494,17 +501,65 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
...
+        if let Some(ref manifest_path) = cmd.wadm_opts.wadm_application {
+            spinner.update_spinner_message(
+                // wadm and NATS both exit immediately when sent SIGINT
+                "CTRL+c received, undeploying specified WADM manifest...".to_string(),
+            );
+            let manifest = load_app_manifest(AppManifestSource::File(manifest_path.clone()))
+                .await
+                .with_context(|| {
+                    format!(
+                        "failed to load manifest from path [{}]",
```
thomastaylor312, line 551: Nit: Might want to note that this happened on cleanup and manual cleanup will be required

crates/wash-cli/src/app/mod.rs
```diff
@@ -216,10 +216,35 @@ async fn deploy_model(cmd: DeployCommand) -> Result<DeployModelResponse> {
         None => load_app_manifest("-".parse()?).await?,
     };
 
-    match app_manifest {
+    deploy_model_from_manifest(&client, lattice, app_manifest, cmd.version).await
+}
+
+pub(crate) fn get_model_name(manifest: &AppManifest) -> Result<&str> {
```
thomastaylor312, line 222: Nit: Should this just be a method on `AppManifest`?

crates/wash-cli/src/up/mod.rs
```diff
@@ -494,17 +501,65 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
...
+                        manifest_path.display()
+                    )
+                })?;
+            let model_name = get_model_name(&manifest)?;
+            wash_lib::app::undeploy_model(&client, Some(lattice), model_name, false).await?;
```
thomastaylor312, line 556: This shouldn't fail if the API returns a noop as it no longer exists so cleanup was successful

thomastaylor312 (changes requested):
A few nits and then one thing around error handling

crates/wash-cli/src/app/mod.rs
vados-cosmonic: Sure, a follow up PR to `wadm` to add it would be great -- tracking [that here](https://github.com/wasmCloud/wadm/issues/293)

crates/wash-cli/src/up/mod.rs
vados-cosmonic: Sorry I'm not sure what you mean here -- does a no-op return an `Err`? Is a no-op not an `Ok` with a different `DeployModelResponse`? I thought the result would be `DeployResult::Acknowledged` in that case.

thomastaylor312: I thought that the undeploy code we wrote did a delete under the hood. It doesn't so this comment isn't valid

thomastaylor312 (approved):

vados-cosmonic:
After discussion with Taylor, there's one more thing left to do:

- [ ] Check for `NoOp`/`Error` response in the `Ok` result from `undeploy_model_*` and `delete_model_*` calls

PR #2180, opened 2024-05-25
chore(wash-lib): Test HTTP(S)_PROXY and WASH_PROXY_* configurations

joonas:
## Feature or Problem

This builds on https://github.com/wasmCloud/wasmCloud/pull/2175 by adding a set of tests to try out the various configurations.

I chose not to push this into #2175 in case we didn't want to go down this route for testing. The goal is mostly to prove out that #2175 works as intended.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -954,6 +954,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "bollard"
+version = "0.16.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0aed08d3adb6ebe0eff737115056652670ae290f177759aac19c30456135f94c"
+dependencies = [
+ "base64 0.22.1",
+ "bollard-stubs",
+ "bytes",
+ "futures-core",
+ "futures-util",
+ "hex",
+ "home",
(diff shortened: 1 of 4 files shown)


joonas:
Note: the tests on the `windows-latest` and `macos-11` runners are failing because:
1. the [underlying library testcontainers uses for interacting with Docker](https://github.com/fussybeaver/bollard) isn't for some reason able to find Docker on Windows.
2. MacOS runner images do not contain Docker at all.

Both of these can probably be addressed with upstream work / downstream/inline workarounds, or alternatively we could limit these tests to Linux only for the moment.

brooksmtownsend (approved):

brooksmtownsend:
I suppose I'm good with either this or #2175 merging (or both!)

joonas:
My goal was to merge #2175 first and then rebase this on top of it, mostly because I wanted to unblock #2175 before we had a chance to look at the testing here, but given that #2175 is blocked on requested changes, this one would've ended up being able to merge first.

I'll also note that this ends up reimplementing #2175 only because you can't stack PRs from forks, at least not that I could find.

PR #2185, opened 2024-05-28
feat(wash-cli): enable deleting manifests by file name

vados-cosmonic:
## Feature or Problem

This commit enables `wash app delete` to delete applications by manifest file, rather than strictly by model name inside the manifest.

We do this simply by checking whether the filename is a valid file path, and attempting to read that file, with the assumption that it is very unlikely someone would call their model `wadm.yaml` (or any variation of a normally named YAML WADM manifest.


brooksmtownsend (approved):
Good to note too that you can't call an app `wadm.yaml`, so no edge cases! 😄

PR #2186, opened 2024-05-28
feat(wash-cli): add --replace option to `wash app deploy`

vados-cosmonic:
## Feature or Problem

This commit adds an option `--replace` to `wash app deploy` which helps run repeated deploys of the *same* manifest by undeploying the specified model right before deploying.

While this isn't quite as robust as we want an option like `wash app dev` to be, it does enable easy scripting.

### Manual Verification

I tested this locally -- you can use the following from `crates/wash-cli`:

```
cargo run -- app deploy --replace <path to wadm.yaml>
```

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -73,6 +73,10 @@ pub struct DeployCommand {
     #[clap(name = "version")]
     version: Option<String>,
 
+    /// Whether or not wash should attempt to replace the resources by performing an optimistic delete shortly before applying resources.
+    #[clap(long = "replace")]
+    replace: bool,
+
     #[clap(flatten)]
     opts: CliConnectionOpts,
 }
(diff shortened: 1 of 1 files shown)


thomastaylor312:
Is this something that is handled by the new optional version stuff in wadm or is the purpose here more to force undeploy resources before deploying again?

vados-cosmonic:
@thomastaylor312 this is more a precursor to [`wash app dev`](https://github.com/wasmCloud/wasmCloud/issues/1594) than to do with the version-less stuff, this enables re-deploying in one command without changing the manifest, which is a bit easier than removing the version and putting it back in later just before saving changes. This works *with* a version-less manifest and turns two commands to one still

crates/wash-cli/src/app/mod.rs
```diff
@@ -216,6 +220,18 @@ async fn deploy_model(cmd: DeployCommand) -> Result<DeployModelResponse> {
         None => load_app_manifest("-".parse()?).await?,
     };
 
+    // If --replace was specified, we should attempt to replace the resources by deleting them beforehand
+    if cmd.replace {
+        if let (Ok(name), version) = (
+            get_model_name(&app_manifest),
+            get_model_version(&app_manifest).map(Into::into),
+        ) {
+            let _ =
```
thomastaylor312, line 229: Nit: Might be worth logging a warning if this fails

crates/wash-cli/src/app/mod.rs
```diff
@@ -268,6 +284,36 @@ async fn put_model(cmd: PutCommand) -> Result<PutModelResponse> {
...
+        }
+    }
+}
+
+/// Retrieve the version of a given [`AppManifest`], returning None if the manifest
+/// does not contain a version (or is not the type to contain a version)
+pub(crate) fn get_model_version(manifest: &AppManifest) -> Option<&str> {
+    match manifest {
+        AppManifest::ModelName(_) => None,
+        AppManifest::SerializedModel(manifest) => manifest
```
thomastaylor312, line 309: Nit: I don't know if this returns the wadm `Manifest` type, but if it does, there is a [`version`](https://docs.rs/wadm-types/latest/wadm_types/struct.Manifest.html#method.version) method

thomastaylor312 (approved):

crates/wash-cli/src/app/mod.rs
vados-cosmonic: It does not! That's a plain YAML object (`serde_yaml::Value`) in there, unfortunately.

crates/wash-cli/src/app/mod.rs
vados-cosmonic: Ah this is a great suggestion, added!

PR #2192, opened 2024-05-28
chore(tracing): Update opentelemetry to use reqwest with rustls

joonas:
## Feature or Problem

Given that we're using rustls pretty much everywhere else, it seems like the right thing would be to use rustls with the OpenTelemetry ecosystem as well.

### DIFF
--- a/crates/tracing/Cargo.toml
+++ b/crates/tracing/Cargo.toml
@@ -37,7 +37,7 @@ opentelemetry-otlp = { workspace = true, features = [
     "logs",
     "trace",
     "metrics",
-    "reqwest-client",
+    "reqwest-rustls",
 ], optional = true }
 serde = { workspace = true, features = ["derive"] }
 tracing = { workspace = true, features = ["log"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4513,6 +4513,7 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "rustls 0.21.12",
+ "rustls-native-certs 0.6.3",
  "rustls-pemfile 1.0.4",
  "serde",
  "serde_json",


rvolosatovs (approved):

rvolosatovs:
Looks like there's some config necessary to specify the CA roots

joonas:
> Looks like there's some config necessary to specify the CA roots

I didn't think there was, I'm pretty sure this should be just enabling different feature flags in opentelemetry-otlp's use of reqwest, per their [feature flags docs](https://docs.rs/opentelemetry-otlp/0.16.0/opentelemetry_otlp/index.html#feature-flags) (I realize that's 0.16.0, and we're on 0.14.0, but the flag behavior hasn't changed, only the docs were added since).

Are you by chance referring to the failing wasmcloud::interfaces test that's not able to start Vault, or something else? If something else, could you point me at what you were looking at?

rvolosatovs:
> > Looks like there's some config necessary to specify the CA roots
> 
> I didn't think there was, I'm pretty sure this should be just enabling different feature flags in opentelemetry-otlp's use of reqwest, per their [feature flags docs](https://docs.rs/opentelemetry-otlp/0.16.0/opentelemetry_otlp/index.html#feature-flags) (I realize that's 0.16.0, and we're on 0.14.0, but the flag behavior hasn't changed, only the docs were added since).
> 
> Are you by chance referring to the failing wasmcloud::interfaces test that's not able to start Vault, or something else? If something else, could you point me at what you were looking at?

I'm referring to the failing test indeed. Simply enabling this feature is not enough, because it uses native certs only.

I think what we need here is:
- update `tracing` crate to use `core` HTTP client, we shouldn't need any reqwest features at all then
- if otlp crate does not support passing a custom client, then let's start with `reqwest-rustls-webpki-roots` feature flag and fix this upstream

crates/tracing/Cargo.toml
```diff
@@ -37,7 +37,7 @@ opentelemetry-otlp = { workspace = true, features = [
     "logs",
     "trace",
     "metrics",
-    "reqwest-client",
+    "reqwest-rustls",
```
rvolosatovs, line 40:
```suggestion
```

Looks like should be able to use https://docs.rs/opentelemetry-otlp/0.16.0/opentelemetry_otlp/struct.HttpExporterBuilder.html#method.with_http_client to configure our HTTP client from `core`

PR #2193, opened 2024-05-29
fix(http-client)!: use platform native system certs, enable tls 1.2

brooksmtownsend:
## Feature or Problem
This PR enables the HTTP client provider to use the platform's native certificate store at runtime, as specified in https://crates.io/crates/hyper-rustls via rustls-native-certs. As far as I know this doesn't introduce any platform specific dependencies, but it can fail in platform specific ways.

Worth noting this would also enable loading a set of certs using the `SSL_CERT_FILE` environment variable, however we don't pass environment variables to providers by default. We may need to consider adding that as a separate feature.

## Release Information
http-client v0.10.0

### Manual Verification
Draft pending verification this fixes an issue on machines with additional certificates

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2698,6 +2698,7 @@ dependencies = [
  "hyper 1.3.1",
  "hyper-util",
  "rustls 0.23.5",
+ "rustls-native-certs 0.7.0",
  "rustls-pki-types",
  "tokio",
  "tokio-rustls 0.26.0",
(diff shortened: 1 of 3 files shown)


thomastaylor312:
We could load from a config value instead for the path location. We should definitely decide how to configure local system access things in a standard way

thomastaylor312 (approved):

PR #2201, opened 2024-05-30
feat: Add support for configuring grpc protocol with OpenTelemetry

joonas:
## Feature or Problem

We need to support OpenTelemetry's gRPC protocol in the exporter configuration as [there can be benefits](https://github.com/open-telemetry/oteps/blob/main/text/0035-opentelemetry-protocol.md#otlp-over-grpc) to using it over the (default) HTTP-based binary protocol.

This makes it possible to pass in `--observability-protocol=grpc` or `WASMCLOUD_OBSERVABILITY_PROTOCOL=grpc` to enable it.

The default is still kept with the HTTP-based binary protocol.

### DIFF
--- a/crates/core/src/otel.rs
+++ b/crates/core/src/otel.rs
@@ -1,8 +1,16 @@
 //! Core reusable logic around [OpenTelemetry ("OTEL")](https://opentelemetry.io/) support
 
+use std::str::FromStr;
+
+use anyhow::bail;
 use serde::{Deserialize, Serialize};
 
 use crate::wit::WitMap;
+
+// We redefine the upstream variables here since they are not exported by the opentelemetry-otlp crate:
+// https://github.com/open-telemetry/opentelemetry-rust/blob/opentelemetry-0.23.0/opentelemetry-otlp/src/exporter/mod.rs#L57-L60
+const OTEL_EXPORTER_OTLP_GRPC_ENDPOINT_DEFAULT: &str = "http://localhost:4317";
+const OTEL_EXPORTER_OTLP_HTTP_ENDPOINT_DEFAULT: &str = "http://localhost:4318";
 
 /// Configuration values for OpenTelemetry
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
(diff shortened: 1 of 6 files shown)


crates/core/src/otel.rs
```diff
@@ -23,6 +30,86 @@ pub struct OtelConfig {
     pub metrics_endpoint: Option<String>,
     /// Overrides the OpenTelemetry endpoint for logs.
     pub logs_endpoint: Option<String>,
+    /// Determines whether grpc or http will be used for exporting the telemetry.
+    pub protocol: Option<String>,
```
brooksmtownsend, line 34: If you make this an `Option<OtelProtocol>` it would definitely simplify other parts of the code 😄 What I would want to double check though is that it serializes cleanly to JSON as the lowercase variant

crates/core/src/otel.rs
```diff
@@ -23,6 +30,86 @@ pub struct OtelConfig {
...
+        };
+        Ok(self)
+    }
+
+    pub fn otel_protocol(&self) -> OtelProtocol {
+        match self
+            .protocol
+            .clone()
+            .unwrap_or(DEFAULT_PROTOCOL.to_string())
+            .as_str()
```
brooksmtownsend, line 51: Clippy is probably gonna get you here with the `unwrap_or` function call instead of `unwrap_or_else`, but if you're able to change the data type of the protocol this function can be a bit simpler

src/main.rs
```diff
@@ -241,6 +241,14 @@ struct Args {
...
     #[clap(long = "override-logs-endpoint", hide = true)]
     logs_endpoint: Option<String>,
+
+    /// Configures whether grpc or http will be used for exporting the enabled telemetry. This defaults to 'http'.
+    #[clap(
+        long = "observability-protocol",
+        env = "WASMCLOUD_OBSERVABILITY_PROTOCOL",
+        hide = true
+    )]
+    observability_protocol: Option<String>,
```
brooksmtownsend, line 251: Here I think this should definitely be an `Option<OtelProtocol>`, clap should be able to handle the parsing for you well only accepting `grpc` or `http` and failing at the command line instead of later on

crates/core/src/otel.rs
```diff
@@ -23,6 +30,86 @@ pub struct OtelConfig {
...
+}
+
+impl OtelConfig {
+    pub fn parse(self) -> anyhow::Result<Self> {
+        match self.protocol.clone().unwrap_or_default().as_ref() {
+            "http" | "grpc" => {}
+            protocol => bail!("unsupported protocol: {:?}", protocol),
+        };
+        Ok(self)
+    }
```
brooksmtownsend, line 44: I find this function a little confusing since it reads more like a `validate()` since it acts on existing "valid" variables, but I think that changing the stored variable to an `OtelProtocol` would remove the need for this

brooksmtownsend (changes requested):
Mostly requests around more strongly enforcing the protocol at the `OtelConfig` struct level, since it should be able to make the rest of the code simpler 😃 In general I'm a big fan of the refactors to just passing around `otel_config`

crates/core/src/otel.rs
brooksmtownsend: That being said, you may as well just make it non-optional, since you _have_ to have a protocol right?

crates/core/src/otel.rs
thomastaylor312: Yeah this really confused me as well. But I also agree with @brooksmtownsend's suggested change

src/main.rs
thomastaylor312: Yeah, you can probably move your `parse` function to be an implementation of `FromStr` for OtelProtocol and I'm pretty sure clap will just use that

crates/core/src/otel.rs
joonas: I made this non-optional and added a `impl Default for OtelProtocol` which takes care of the case where a None would be passed via the command-line.

crates/core/src/otel.rs
joonas: I was struggling with this, turns out what I really wanted was the `impl FromStr for OtelProtocol` which essentially solved this the way I wanted in the first place.

src/main.rs
joonas: This is exactly what I wanted, thanks for the pointer, I had implemented `From<Option<String>>` which didn't give me the option to return an Error in the case that the supplied value couldn't be parsed to something.

crates/core/src/otel.rs
joonas: You're exactly right, I already had `Default` impl in place for OtelProtocol, all I needed was to add `FromStr` and this could be handled with the combination of clap arguments and `unwrap_or_default()` on the arguments as they get passed in to the `OtelConfig` struct, thanks!

joonas:
@brooksmtownsend @thomastaylor312 thanks for your feedback, `FromStr` was exactly the piece I was missing!

brooksmtownsend (approved):
Looks great!

Issue #1827, opened 2024-04-08
[FEATURE] Rename `Actor` to `Component`


brooksmtownsend:
### Affected project(s)

- [X] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

In the leadup to wasmCloud 1.0, we discussed and decided to rename "actor" to "component" as there is nothing special about a component that runs in wasmCloud, it's just a WebAssembly component 😄 In all of our protocols, events, and public APIs we've attempted to do this already to avoid breaking changes, but we have not updated all internal naming from actor -> component.

### Describe the solution you'd like

This is okay as a slow-burn issue as it's not critical to do immediately, but I wanted to file this issue for two reasons:
1. If there are new features that touch code that mention actors, go ahead and rename it to components while you're in there
2. If any community members see "actor" and that drives confusion, I want to be clear that it's just a WebAssembly component 🚀 

### Describe alternatives you've considered

We could do this as a part of a big refactor, which is fine, I just want to be conscious about doing that while it might cause larger merge conflicts amidst other work

### Additional context

We discussed this rename in the 2024-03-13 community meeting: https://wasmcloud.com/community/2024/03/13/community-meeting

nitame:
Hello, as this issue is labeled good first issue, I would like to try to do this issue or at least one of the tasks included in it. Is it ok? Or does anyone want to do it?
Thanks,

thomastaylor312:
@nitame you can go ahead and work on this! Should just be rename of internal variables, function parameters, etc. Let us know if you have any questions!

brooksmtownsend:
Pinning good here, but @nitame is knocking this out quickly so we should be able to wrap this up!

brooksmtownsend:
@nitame You've done a ton of work here, I think we can close this up! If there are any other dangling references we can certainly take care of them, but I believe the vast majority are taken care of now. Any other last ones we should fix?

nitame:
@brooksmtownsend Yes you can close this issue. As you said, I think there are a few remaining references to actors, but it could surely be handled on the go with BSR.

PR #2206, opened 2024-05-31
chore(runtime-crate): refactor runtime/src structure to remove 'actor…

nitame:
…' references

## Feature or Problem
Replace `actor` with `component` inside the runtime crate.

 
## Related Issues
Linked to #1827


thomastaylor312:
@nitame sorry for the delay in response here. So I think `component/component` is fine. There is some nesting involved because we're making sure we have the same structure as some of the wasmtime bindings crates.

crates/runtime/src/lib.rs
```diff
@@ -17,12 +17,7 @@ pub mod runtime;
 /// wasmCloud I/O functionality
 pub mod io;
 
-#[deprecated(
-    since = "0.3.1",
-    note = "ActorConfig has been renamed to ComponentConfig"
-)]
-pub use actor::Config as ActorConfig;
-pub use actor::{Component, ComponentInstance, Config as ComponentConfig};
+pub use component::{Component, ComponentInstance, Config as ComponentConfig};
```
nitame, line 20: Here I removed the import of Config as ActorConfig because it was marked deprecated and I didn't find a reference to ActorConfig. Does it seem legit to you?

thomastaylor312: Yeah I think you're good here. The runtime crate hasn't actually been released, so if it isn't anywhere in the repo, we should be good

nitame:
Hello, I have some trouble with cargo clippy which enforces to avoid having nested mods with the same names. A proper fix could be to have different names for `component/component` but I don't know what naming makes more sense.

vados-cosmonic:
Hey @nitame I think you can suppress that lint -- if I'm understanding correctly it's the `component::Component` thing, right? I think we do want that hierarchy (at least for now it's outside of the scope of this PR), so feel free to suppress it! 

I'm sure you saw it in the error message but you can use:

```
#[allow(clippy::module_inception)]
```

thomastaylor312 (approved):
This is good to go @nitame! If you can rebase real quick, we'll go ahead and merge

thomastaylor312:
Oh wait, nvm on the rebase, looks like we can merge

PR #2207, opened 2024-05-31
fix(*): Clarifies additional licenses based off of CNCF license list

thomastaylor312:
This involved removing a dependency that we weren't commonly using or wasn't even documented in some cases (file loading a .env file in 2 providers). Things should be good to go now

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5220,12 +5220,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "simple_env_load"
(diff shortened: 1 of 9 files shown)


ricochet (approved):

thomastaylor312:
I am going to override the automatic tag of host-maintainer reviewer since this most definitely doesn't touch host code. It was tagged because I removed a dep in the top level Cargo.toml, which was only used in providers

PR #2208, opened 2024-05-31
feat(*): Allows for pushing binary wit packages with wash

thomastaylor312:
This rounds out a feature I didn't think we'd need for a while. This should allow us to push any type of component

### DIFF
--- a/crates/wash-cli/tests/wash_claims.rs
+++ b/crates/wash-cli/tests/wash_claims.rs
@@ -12,11 +12,13 @@ use std::{
 use assert_json_diff::assert_json_include;
 use serde_json::json;
 
+const ECHO_OCI: &str = "ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0";
+const ECHO_ACC: &str = "ADVIWF6Z3BFZNWUXJYT5NEAZZ2YX4T6NRKI3YOR3HKOSQQN7IVDGWSNO";
+const ECHO_MOD: &str = "MBFFVNGFK3IA2ZXXG5DQXQNYM6TNG45PHJMJIJFVFI6YKS3XTXL3DRRK";
+const ECHO_SHA: &str = "sha256:079275a324c0fcd0c201878f0c158120c4984472215ec3f64eb91ba9ee139f72";
+
 #[test]
-#[cfg_attr(
-    not(can_reach_wasmcloud_azurecr_io),
-    ignore = "wasmcloud.azurecr.io is not reachable"
-)]
+#[cfg_attr(not(can_reach_ghcr_io), ignore = "ghcr.io is not reachable")]
 fn integration_claims_sign() {
     const SUBFOLDER: &str = "claims_sign";
     let sign_dir = test_dir_with_subfolder(SUBFOLDER);
@@ -29,12 +31,7 @@ fn integration_claims_sign() {
     // as signing an unsigned wasm
     let echo = test_dir_file(SUBFOLDER, "echo.wasm");
     let get_hello_wasm = wash()
-        .args([
-            "pull",
-            "wasmcloud.azurecr.io/echo:0.2.1",
-            "--destination",
-            echo.to_str().unwrap(),
-        ])
+        .args(["pull", ECHO_OCI, "--destination", echo.to_str().unwrap()])
         .output()
         .expect("failed to pull echo for claims sign test");
     assert!(get_hello_wasm.status.success());
@@ -150,9 +147,6 @@ fn integration_claims_sign() {
 )]
 fn integration_claims_inspect() {
     const SUBFOLDER: &str = "claims_inspect";
-    const ECHO_OCI: &str = "wasmcloud.azurecr.io/echo:0.2.1";
-    const ECHO_ACC: &str = "ACOJJN6WUP4ODD75XEBKKTCCUJJCY5ZKQ56XVKYK4BEJWGVAOOQHZMCW";
-    const ECHO_MOD: &str = "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5";
     let inspect_dir = test_dir_with_subfolder(SUBFOLDER);
     let echo_claims = &format!("{LOCAL_REGISTRY}/echo:claimsinspect");
 
@@ -189,8 +183,7 @@ fn integration_claims_inspect() {
         "component": ECHO_MOD,
         "can_be_used": "immediately",
         "expires": "never",
-        "tags": "None",
-        "version": "0.2.1"
+        "version": "0.1.0"
     });
 
     assert_json_include!(
@@ -212,13 +205,7 @@ fn integration_claims_inspect() {
 
     let remote_inspect = wash()
         .args([
-            "claims",
-            "inspect",
-            ECHO_OCI,
-            "--digest",
-            "sha256:55689502d1bc9c48f22b278c54efeee206a839b8e8eedd4ea6b19e6861f66b3c",
-            "-o",
-            "json",
+            "claims", "inspect", ECHO_OCI, "--digest", ECHO_SHA, "-o", "json",
         ])
         .output()
         .expect("failed to inspect local registry wasm");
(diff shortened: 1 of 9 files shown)


crates/wash-cli/tests/wash_claims.rs
```diff
@@ -12,9 +12,14 @@ use std::{
...
+const ECHO_OCI: &str = "ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0";
+const ECHO_ACC: &str = "ADVIWF6Z3BFZNWUXJYT5NEAZZ2YX4T6NRKI3YOR3HKOSQQN7IVDGWSNO";
+const ECHO_MOD: &str = "MBFFVNGFK3IA2ZXXG5DQXQNYM6TNG45PHJMJIJFVFI6YKS3XTXL3DRRK";
+const ECHO_SHA: &str = "sha256:079275a324c0fcd0c201878f0c158120c4984472215ec3f64eb91ba9ee139f72";
+
 #[test]
 #[cfg_attr(
-    not(can_reach_wasmcloud_azurecr_io),
+    not(can_reach_ghcr_io),
     ignore = "wasmcloud.azurecr.io is not reachable"
```
vados-cosmonic, line 23: Ah can we update this message too?

crates/wash-cli/tests/wash_inspect.rs
```diff
@@ -13,9 +13,11 @@ use serde_json::json;
...
 fn integration_inspect_component() {
     const SUBFOLDER: &str = "inspect";
-    const ECHO_OCI: &str = "wasmcloud.azurecr.io/echo:0.2.1";
-    const ECHO_ACC: &str = "ACOJJN6WUP4ODD75XEBKKTCCUJJCY5ZKQ56XVKYK4BEJWGVAOOQHZMCW";
-    const ECHO_MOD: &str = "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5";
+    const ECHO_OCI: &str = "ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0";
+    const ECHO_ACC: &str = "ADVIWF6Z3BFZNWUXJYT5NEAZZ2YX4T6NRKI3YOR3HKOSQQN7IVDGWSNO";
+    const ECHO_MOD: &str = "MBFFVNGFK3IA2ZXXG5DQXQNYM6TNG45PHJMJIJFVFI6YKS3XTXL3DRRK";
+    const ECHO_SHA: &str =
+        "sha256:079275a324c0fcd0c201878f0c158120c4984472215ec3f64eb91ba9ee139f72";
```
vados-cosmonic, line 20: Ah this could benefit from the aforementioned move to a common spot...

crates/wash-cli/tests/wash_reg.rs
```diff
@@ -69,7 +63,7 @@ fn integration_reg_pull_comprehensive() {
             "--destination",
             comprehensive_echo.to_str().unwrap(),
             "--digest",
-            "sha256:a17a163afa8447622055deb049587641a9e23243a6cc4411eb33bd4267214cf3",
+            "sha256:079275a324c0fcd0c201878f0c158120c4984472215ec3f64eb91ba9ee139f72",
```
vados-cosmonic, line 66: Did we want to move this SHA out too? Also, I wonder if they also make sense in `common`

crates/wash-cli/tests/wash_reg.rs
```diff
@@ -110,7 +104,7 @@ fn integration_reg_pull_comprehensive() {
 // NOTE: This test will fail without a local docker registry running
 #[test]
 #[cfg_attr(
-    not(can_reach_wasmcloud_azurecr_io),
+    not(can_reach_ghcr_io),
     ignore = "wasmcloud.azurecr.io is not reachable"
```
vados-cosmonic, line 108: ignore msg update?

crates/wash-lib/src/cli/registry.rs
```diff
@@ -93,6 +93,11 @@ pub struct RegistryPushCommand {
     #[clap(short = 'a', long = "annotation", name = "annotations")]
     pub annotations: Option<Vec<String>>,
 
+    /// World name of the component you are pushing. This is only used if you are pushing a binary
+    /// wit component that may contain multiple worlds.
+    #[clap(long = "world", env = "WASH_REG_WORLD")]
```
vados-cosmonic, line 98:
```suggestion
    #[clap(long = "wit-world", env = "WASH_REG_WIT_WORLD")]
```
What do you think about `wit-world` versus `world`? excessive?

crates/wash-lib/src/registry.rs
```diff
@@ -69,6 +71,12 @@ pub enum SupportedArtifacts {
     Wasm(Config, ImageLayer),
 }
 
+/// An enum indicating the type of artifact that was pulled
+pub enum ArtifactType {
```
vados-cosmonic, line 75: Does it make sense for this to be `OciArtifactType`? Are we going to have any other kind of registry/`pull`able area?

crates/wash-lib/src/registry.rs
```diff
@@ -255,11 +264,14 @@ pub async fn push_oci_artifact(
 
 /// Helper function to determine artifact type and parse it into a config and layer ready for use in
 /// pushing to OCI
-pub async fn parse_and_validate_artifact(artifact: Vec<u8>) -> Result<SupportedArtifacts> {
+pub async fn parse_and_validate_artifact(
+    artifact: Vec<u8>,
```
vados-cosmonic, line 268: Do you need to take ownership of the vector here?

crates/wash-lib/src/registry.rs
```diff
@@ -268,11 +280,21 @@ pub async fn parse_and_validate_artifact(artifact: Vec<u8>) -> Result<SupportedA
     }
 }
 
+/// Function that identifies whether the artifact is a component or a provider archive. Returns an
+/// error if it isn't a known type
+// NOTE: This exists because we don't care about parsing the proper world when pulling
+pub async fn identify_artifact(artifact: Vec<u8>) -> Result<ArtifactType> {
```
vados-cosmonic, line 286: Might be nice to avoid having to own it, if we could avoid it

crates/wash-cli/tests/wash_claims.rs
```diff
@@ -12,9 +12,14 @@ use std::{
 use assert_json_diff::assert_json_include;
 use serde_json::json;
 
+const ECHO_OCI: &str = "ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0";
+const ECHO_ACC: &str = "ADVIWF6Z3BFZNWUXJYT5NEAZZ2YX4T6NRKI3YOR3HKOSQQN7IVDGWSNO";
+const ECHO_MOD: &str = "MBFFVNGFK3IA2ZXXG5DQXQNYM6TNG45PHJMJIJFVFI6YKS3XTXL3DRRK";
+const ECHO_SHA: &str = "sha256:079275a324c0fcd0c201878f0c158120c4984472215ec3f64eb91ba9ee139f72";
```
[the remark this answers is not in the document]
thomastaylor312: This is already better than it was before. They were redefined in each function. So I kept the change small for now, but we should probably refactor some more

crates/wash-lib/src/registry.rs
thomastaylor312: I was following similar naming to `SupportedArtifacts`, but since this is in the registry module, I don't think we need the additional context

crates/wash-lib/src/cli/registry.rs
thomastaylor312: I think in the context we're working with here, world should be pretty well known. But I'm not super married to it either

crates/wash-lib/src/registry.rs
thomastaylor312: Well, I did need it before but then I needed to clone anyway. I'll change that

crates/wash-lib/src/registry.rs
thomastaylor312: Yep, copy paste

brooksmtownsend (approved):

PR #2223, opened 2024-06-03
feat(wascap): add Host entity type

protochron:
## Feature or Problem
Add a `Host` entity type to wascap. This allows us to generate JWTs for hosts that contains assertions about the metadata they were started with. For now this only includes host labels, but this could change in the future.

### DIFF
--- a/crates/wascap/src/jwt.rs
+++ b/crates/wascap/src/jwt.rs
@@ -138,6 +138,15 @@ pub struct Invocation {
     pub invocation_hash: String,
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
+pub struct Host {
+    /// Optional friendly descriptive name for the host
+    pub name: Option<String>,
+    /// Optional labels for the host
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub labels: Option<HashMap<String, String>>,
+}
+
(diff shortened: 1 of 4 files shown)


crates/wascap/src/jwt.rs
```diff
@@ -138,6 +138,14 @@ pub struct Invocation {
     pub invocation_hash: String,
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
+pub struct Host {
+    /// Optional friendly descriptive name for the host
+    pub name: Option<String>,
+    /// Optional labels for the host
+    pub labels: Option<HashMap<String, String>>,
```
brooksmtownsend, line 146: Do you feel there's a meaningful difference between a `None` here and an empty hashmap? 

If we do change to an empty hashmap, it should be a `#[serde(default)]`, but this is probably a bikeshed

brooksmtownsend (approved):

crates/wascap/src/jwt.rs
protochron: I don't think there is, though I'd go a step further and ensure that it has a `#serde(skip_serializing_if = Option::is_none)]` which I should have done originally. If the host doesn't have labels then we skip serializing the key to begin with

protochron: Updated to skip serializing if it's empty, so I think it should stay an `Option`

brooksmtownsend:
@protochron if you can sign off your commit this looks good to go

PR #2243, opened 2024-06-06
feat(example): set outgoing response before writing

brooksmtownsend:
## Feature or Problem
This PR changes the ordering in our example HTTP components to ensure we are setting the response outparam _before_ writing bytes to the body.


rvolosatovs (approved):

vados-cosmonic (approved):
LGTM :rocket: 

A bit surprised it works like this but :shrug:

brooksmtownsend:
> A bit surprised it works like this but 🤷

Me too! But it does kind of make sense. I ran into some issues when writing larger payloads using splice or read/write loops because without setting the ResponseOutparam I can write a certain amount to an internal buffer, but that isn't actually streamed to the requester. This way we will immediately write/flush the bytes back to the requester and then we can "finish" the body at the end.

PR #2244, opened 2024-06-06
feat(host): add option for normalizing config

vados-cosmonic:
## Feature or Problem

This commit adds an option for normalizing configuration keys that are used/provided by hosts, with a new option `--normalize-config-keys`.

When this option is enabled, config keys that are retrieved and used by the host will all be normalized (for now, lowercased), before use.

### DIFF
--- a/crates/host/src/wasmbus/config.rs
+++ b/crates/host/src/wasmbus/config.rs
@@ -179,16 +179,19 @@ pub struct BundleGenerator {
     store: Store,
     watch_cache: WatchCache,
     watch_handles: Arc<RwLock<AbortHandles>>,
+    /// Whether to normalize configurations retrieved
+    pub normalize_keys: bool,
 }
 
 impl BundleGenerator {
     /// Create a new bundle generator
     #[must_use]
-    pub fn new(store: Store) -> Self {
+    pub fn from_store(store: Store) -> Self {
         Self {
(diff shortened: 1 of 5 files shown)


brooksmtownsend (commented):
I don't feel strongly for or against this PR, though in general I feel like this isn't really necessary to add in at the wasmCloud level. I think it would be frustrating to depend on this only to use a provider that uses uppercase keys by convention (e.g. like environment variables) and needing to turn it off

vados-cosmonic:
Ahh sure -- some people might be expecting the *opposite* functionality, that's certainly true. Well I guess what we can do is wait for this to actually get reported as an issue that people commonly run into before we add some sort of normalization step?

Talking with @thomastaylor312 the reason it's in the host as opposed to somewhere like `provider-sdk` was so that this same normalization could be used by components as well. 

That said, components *and* providers would both fall under the logic you were mentioning -- if they know what it *should* be, then there's no need for normalization.

thomastaylor312:
Yeah @brooksmtownsend I definitely understand here. If we normalize though, it does have to be done either individually by each provider/component or it has to be done by the host. I am honestly ok to wait until people ask for this so it can be more clearly defined

vados-cosmonic:
OK great, let's wait till people ask for it then!

PR #2254, opened 2024-06-07
feat(#1740): support purging NATS during down

### LINKED ISSUE #1740: [FEATURE] Provide simple way to reset/wipe `wadm` store/s
### Affected project(s)

- [x] documentation
- [X] wasmCloud CLI (wash)

During a debugging session with a community member, I noticed that even after deleting everything in `~/.wash`, the wadm store stayed in tact. This didn't turn out to be a problem, but I would have thought that the JetStream storage location would have been contained in the `~/.wash` folder also.

I'd love a way to be able to delete everything nats/wadm/wash related with a `wash` subcommand, or by deleting a single folder.

### Proposed Action

After discussion (see thread), the proposed action to take is to:
- Add a `--purge` option to `wash down` which will delete everything in the `~/.wash/` folder.
- Move the JetStream storage location inside `~/.wash`

brooksmtownsend:
## Feature or Problem
This PR allows for purging NATS jetstream persistent streams and kv buckets that are needed by wadm, wasmcloud, or both. For example, to completely reset local wasmCLoud, run:
```
wash down --purge all
```

While I was in there I downgraded some of the errors in `wash down` to just be warning logs, like trying to shutdown the local NATS/wadm if the wasmCloud pidfile was missing. I want `down` to feel like something you can run over and over, and a command that supports removing incompletely cleaned up resources.

## Related Issues
Fixes #1740

## Release Information
wash 0.29 or 0.30 depending on time

### DIFF
--- a/crates/wash-cli/src/down/mod.rs
+++ b/crates/wash-cli/src/down/mod.rs
@@ -78,7 +78,11 @@ pub struct DownCommand {
     pub all: bool,
 
     /// Purge NATS Jetstream storage and streams that persist when wasmCloud is stopped
-    #[clap(long = "purge", alias = "flush")]
+    #[clap(
+        long = "purge-jetstream",
+        alias = "flush-jetstream",
+        default_value = "none"
+    )]
     pub purge: PurgeJetstream,
 }
 
@@ -128,16 +132,19 @@ pub async fn handle_down(cmd: DownCommand, output_kind: OutputKind) -> Result<Co
             );
             return Ok(CommandOutput::new(out_text, out_json));
         } else {
-            // TODO: don't die on this
             let wasmcloud_pid_file_path = install_dir.join(WASMCLOUD_PID_FILE);
-            tokio::fs::remove_file(&wasmcloud_pid_file_path)
+            // Failing to find the pid file is not an error that should prevent stopping other resources
+            if let Err(e) = tokio::fs::remove_file(&wasmcloud_pid_file_path)
                 .await
                 .with_context(|| {
                     format!(
                         "failed to find wasmcloud pid file [{}]",
                         wasmcloud_pid_file_path.display()
                     )
-                })?;
+                })
+            {
+                warn!("{}", e.to_string());
+            }
         }
     } else {
         warn!("Couldn't connect to NATS, unable to stop running hosts")
@@ -146,12 +153,16 @@ pub async fn handle_down(cmd: DownCommand, output_kind: OutputKind) -> Result<Co
     match stop_wadm(&install_dir).await {
         Ok(_) => {
             let pid_file_path = &install_dir.join(WADM_PID);
-            // TODO: don't die on this
-            tokio::fs::remove_file(&pid_file_path)
+            // Failing to find the pid file is not an error that should prevent stopping other resources
+            if let Err(e) = tokio::fs::remove_file(&pid_file_path)
                 .await
                 .with_context(|| {
                     format!("failed to find WADM pid file [{}]", pid_file_path.display())
-                })?;
+                })
+            {
+                warn!("{}", e.to_string());
+            }
+
             out_json.insert("wadm_stopped".to_string(), json!(true));
             out_text.push_str("✅ wadm stopped successfully\n");
         }
--- a/crates/wash-cli/src/down/mod.rs
+++ b/crates/wash-cli/src/down/mod.rs
@@ -4,6 +4,7 @@ use std::process::{Output, Stdio};
 
 use anyhow::{bail, Context as _, Result};
 use clap::Parser;
+use futures::future::join_all;
 use serde_json::json;
 use tokio::process::Command;
 use tracing::warn;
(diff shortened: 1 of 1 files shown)


crates/wash-cli/src/down/mod.rs
```diff
@@ -59,8 +73,13 @@ pub struct DownCommand {
     #[clap(long = "host-id")]
     pub host_id: Option<ServerId>,
 
+    /// Shutdown all hosts running locally if launched with --multi-local
     #[clap(long = "all")]
     pub all: bool,
+
+    /// Purge NATS Jetstream storage and streams that persist when wasmCloud is stopped
+    #[clap(long = "purge", alias = "flush")]
+    pub purge: PurgeJetstream,
```
vados-cosmonic, line 82: Can we name this in a way that reflects it's purging NATS/Jetstream in particular?

crates/wash-cli/src/down/mod.rs
```diff
@@ -109,14 +129,18 @@ pub async fn handle_down(cmd: DownCommand, output_kind: OutputKind) -> Result<Co
...
                 .await
                 .with_context(|| {
                     format!(
                         "failed to find wasmcloud pid file [{}]",
                         wasmcloud_pid_file_path.display()
                     )
-                })?;
+                })
+            {
+                warn!("{}", e.to_string());
```
vados-cosmonic, line 142:
```suggestion
                warn!("failed to remove wasmcloud PID file [{}]: {}", wasmcloud_pid_file_path.display(), e.to_string());
```

crates/wash-cli/src/down/mod.rs
```diff
@@ -125,11 +149,16 @@ pub async fn handle_down(cmd: DownCommand, output_kind: OutputKind) -> Result<Co
...
-            tokio::fs::remove_file(&pid_file_path)
+            // Failing to find the pid file is not an error that should prevent stopping other resources
+            if let Err(e) = tokio::fs::remove_file(&pid_file_path)
                 .await
                 .with_context(|| {
                     format!("failed to find WADM pid file [{}]", pid_file_path.display())
-                })?;
+                })
+            {
+                warn!("{}", e.to_string());
```
vados-cosmonic, line 159:
```suggestion
                warn!("failed to remove wasmcloud PID file [{}]: {}", wasmcloud_pid_file_path.display(), e.to_string());
```

vados-cosmonic (approved):
LGTM :rocket:

Only important thing was the naming of the option

crates/wash-cli/src/down/mod.rs
brooksmtownsend: The warn does include this information! It's added with the `with_context` call above. For example:
```
➜ RUST_LOG=warn cargo run -- down --purge all
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.28s
     Running `/Users/brooks/github.com/wasmcloud/wasmcloud/target/debug/wash down --purge all`
>>> ⡀⠀  Stopping wasmCloud ...                                                                                                   2024-06-07T13:56:40.989987Z  WARN wash_cli::down: failed to find wasmcloud pid file [/Users/brooks/.wash/downloads/wasmcloud.pid]

✅ wasmCloud hosts stopped successfully
❌ Could not stop wadm: No pidfile found at [/Users/brooks/.wash/downloads/wadm.pid]
✅ NATS Jetstream purged successfully
❌ NATS server did not stop successfully: No pidfile found for nats-server, assuming it's managed externally
🛁 wash down completed successfully
```

crates/wash-cli/src/down/mod.rs
brooksmtownsend: Do you think like, `--purge-nats` or `--purge-jetstream`? From those I kinda do prefer `--purge` but I'm not that hard-convinced

vados-cosmonic: `--purge-jetstream` is probably pretty reasonable? Or maybe something like `--purge-state`, because that's what you're really purging right?

crates/wash-cli/src/down/mod.rs
vados-cosmonic: Ahhh sorry that's right -- I was distracted by the existing message... it's a little wrong, it should say it failed to *delete* the file (there could be other reasons)

crates/wash-cli/src/down/mod.rs
vados-cosmonic: This must be the same as that other case -- will resolve!

crates/wash-cli/src/down/mod.rs
vados-cosmonic: Maybe `--purge-nats-state`? Because you're really purging state, but only the stuff in NATS -- not like you're cleaning out the filesystem

brooksmtownsend: I think `purge-jetstream` feels fine here, I know in general we could have a `--purge` if we also added support for some of the other drain parameters like cache or removing the `~/.wash` directory

vados-cosmonic: Perfect! Resolving for :rocket:

vados-cosmonic (approved):
LGTM :rocket:

PR #2255, opened 2024-06-07
fix(provider-blobstore-fs): make ROOT configuration case-insensitive

ericliu0000:
This PR adjusts the ROOT path value matching behavior to convert all keys to uppercase. This makes the relevant value case-insensitive.

## Feature or Problem
When configuring the value of ROOT for blobstore-fs, the only allowable spelling of the key to configure the value is "ROOT" — case sensitive.


brooksmtownsend (approved):
Looks great! This should be a lot nicer not worrying about the cases here.

In general, just as a note for new providers / contributors, we should prefer lowercasing if we are not making things case insensitive.

PR #2257, opened 2024-06-07
chore(examples): add missing manifests, use official images

vados-cosmonic:
## Feature or Problem

This commit adds WADM manifests for examples that didn't have them, and also updates all manifests to have a line about how to use the local code versus the officially released components

This change *should* make it much easier to pull and run any WADM manifest in an example.


brooksmtownsend (approved):
This is great!

PR #2259, opened 2024-06-07
feat(wash) use wadm-client crate, update CLI

brooksmtownsend:
- **feat(wash-lib)!: wrap new wadm-client**
- **feat(wash-cli)!: use new wadm-client**

## Feature or Problem
This PR updates `wash-lib` and `wash-cli` to use the new wadm-client crate which provides better strongly typed methods for managing applications. It also gives us much nicer error formatting, so I was able to do a good bit of refactoring on the wash-cli side. I kept the wash-lib API as similar as possible to reduce breaking changes with dependents, but in general using the wadm-client crate directly is recommended.

While I was in there, I noticed something that has bugged me for a bit around our use of "app spec" and "model" when it could be a whole lot simpler to talk about. I updated the help text and some variable names to standardize on CRUDdy operations and the management of "applications" rather than dealing with specs and models and versions and apps. Feedback welcome.

## Release Information
next minor wash-lib and wash-cli.

### Manual Verification
I took a look at the help texts but I need to manually verify these work as expected.

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -1,13 +1,11 @@
 use std::collections::HashMap;
 use std::path::PathBuf;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, Context};
 use clap::{Args, Subcommand};
 use serde_json::json;
-use wadm_types::api::{
-    DeleteModelResponse, DeployModelResponse, GetModelResponse, GetResult, ModelSummary,
-    PutModelResponse, PutResult, StatusResponse, VersionResponse,
-};
+use wadm_client::Result;
+use wadm_types::api::ModelSummary;
 use wadm_types::validation::{validate_manifest_file, ValidationFailure, ValidationOutput};
 use wash_lib::app::{load_app_manifest, AppManifest};
 use wash_lib::cli::{CliConnectionOpts, CommandOutput, OutputKind};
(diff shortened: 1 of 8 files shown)


crates/wash-cli/src/app/mod.rs
```diff
@@ -98,7 +96,7 @@ pub struct DeleteCommand {
 
 #[derive(Args, Debug, Clone)]
 pub struct PutCommand {
-    /// Possible sources: file from fs, remote file http url,  or stdin. if no source is provided (or arg marches '-'), stdin is used.
+    /// The source of the application manifest, either a file path, remote file http url, or stdin. If no source is provided (or arg marches '-'), stdin is used.
```
thomastaylor312, line 99: marches -> matches?

crates/wash-cli/src/app/mod.rs
```diff
@@ -285,54 +284,81 @@ async fn put_model(cmd: PutCommand) -> Result<PutModelResponse> {
...
+        wash_lib::app::get_model_details(&client, lattice, &cmd.app_name, cmd.version).await?;
+
+    let mut map = HashMap::new();
+    map.insert("application".to_string(), json!(manifest));
+    let yaml = serde_yaml::to_string(&manifest).unwrap();
+    Ok(CommandOutput::new(yaml, map))
 }
 
-async fn delete_model_version(cmd: DeleteCommand) -> Result<DeleteModelResponse> {
+async fn delete_application_version(cmd: DeleteCommand) -> Result<CommandOutput> {
```
thomastaylor312, line 361: Nit: Maybe just call `delete_application`

thomastaylor312 (approved):
Nothin' but nits

PR #2262, opened 2024-06-07
feat(port): Specify port mapping for postgres example

ritesh089:
## Feature or Problem
Without the mapping , i get a connection error in the provider. I added a port mapping to make it work.


vados-cosmonic (approved):
LGTM :rocket: 

Thanks for the correction @ritesh089!

vados-cosmonic:
Hey @ritesh089 would you mind changing the commit message? `feat(port) should be something like `docs(examples)` I think!

ritesh089:
> Hey @ritesh089 would you mind changing the commit message? `feat(port)` should be something like `docs(examples)` I think -- I was able to rebase with GitHub but can't change the commit message!

I have updated the commit message.

Issue #2265, opened 2024-06-07
[BUG] OTEL traces/logs get constant 404s after otlp upgrade


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

In the console when running the wasmCloud host from `main`, or any providers built from `main`, I get the following output constantly as direct console output (not formatted with our logs):
```
OpenTelemetry trace error occurred. HTTP status client error (404 Not Found) for url (http://localhost:4318/)
OpenTelemetry log error occurred. HTTP status client error (404 Not Found) for url (http://localhost:4318/)
OpenTelemetry trace error occurred. HTTP status client error (404 Not Found) for url (http://localhost:4318/)
```

### Steps to reproduce

1. checkout main
2. Launch the OTEL docker compose under examples/docker
3. `cargo run -- --enable-observability`
4. See errors

### Expected behavior

I expected for the default tracing/logging endpoints to keep working as before.

### Environment

- OS: [e.g. Linux, MacOS] MacOS
- Shell [e.g. bash, zsh, powershell] zsh
- wasmcloud Version [e.g. 0.81.0] `main`

### Screenshots / Logs / Additional context

I don't think this is _technically_ a bug in the upstream otlp crate, it was just a breaking change that wasn't documented well enough for us to catch it from the changelog. https://github.com/open-telemetry/opentelemetry-rust/issues/1866

Introduced in https://github.com/wasmCloud/wasmCloud/pull/2197 so this hasn't been released in any providers/the host

brooksmtownsend:
As a workaround I've been able to test using our override endpoints.
```
./wasmcloud --override-traces-endpoint http://localhost:4318/v1/traces --override-logs-endpoint http://localhost:4318/v1/logs
```

peter-jerry-ye:
I'm somehow still experiencing this using the `docker-compose-full.yml` in the repository. Shouldn't it been fixed?

peter-jerry-ye:
@vados-cosmonic It's different in that I used the docker file directly or with endpoint `http://otelcol:4318/` and they both gave me 404, which is exactly as described in this issue.

[...]

PR #2197, opened 2024-05-29
chore: Bump opentelemetry-* crates and tracing-opentelemetry to latest version

joonas:
## Feature or Problem

The version of [opentelemetry](https://github.com/open-telemetry/opentelemetry-rust/) we're depending on is [a bit out of date](https://github.com/open-telemetry/opentelemetry-rust/releases/tag/v0.21.0) by now, so this brings us up to date on the release as of [2 weeks ago](https://github.com/open-telemetry/opentelemetry-rust/releases/tag/opentelemetry-0.23.0).

I tried my best to look across the various [CHANGELOG](https://github.com/open-telemetry/opentelemetry-rust/compare/v0.21.0...opentelemetry-0.23.0) entries to see if there were any breaking changes we would be impacted, it didn't seem like we should be impacted by anything other than the introduction of the new Gauge instrument kind.

PR #2266, opened 2024-06-07
fix: Add signal-specific path components to OtelConfig's default endpoints

joonas:
## Feature or Problem

@brooksmtownsend pointed out that the change made in https://github.com/open-telemetry/opentelemetry-rust/pull/1706 that was brought in as part of #2197 changed the default endpoint behavior when you're passing in an explicit endpoint, which we are even if one wasn't provided.

We were relying on the prior behavior where the `/v1/<signal>` path component would be concatenated to the endpoint if it didn't already have one, but that behavior has since been changed to default to `/` instead, which does not work when you're attempting to use the OpenTelemetry collector configured at the default endpoints (which are `/v1/<signal>`).

Fixes https://github.com/wasmCloud/wasmCloud/issues/2265

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6797,6 +6797,7 @@ dependencies = [
  "tower",
  "tracing",
  "ulid",
+ "url",
  "uuid 1.8.0",
  "wascap 0.15.0",
  "webpki-roots 0.26.1",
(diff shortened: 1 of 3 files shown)


brooksmtownsend (approved):
Nice @joonas !

brooksmtownsend:
@joonas adding no-merge label since we had an offline discussion and want to add tests + improvements here 😄

PR #2271, opened 2024-06-09
Feat(wash): Adds flag to [wash up] to allow reading custom nats config

### LINKED ISSUE #1405: [FEATURE] `wash up` should have a flag to specify a NATS server config
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

Right now it is impossible to specify a NATS server config since wash does the right thing and creates one for you when you run `wash up`. Normally this is fine, but if you are working with a particular NATS setup or want to set specific values in the NATS server configuration you have no way to do that

A `--nats-config-file` flag (or something like it!) that allows you to provide a custom NATS server config file that `wash` will use to start the NATS server.

_No response_

### Additional context

_No response_

Aditya1404Sal:
## Related Issues
Resolves #1405 
## Release Information
Next

### Manual Verification
ran `cargo run -- up --nats-config-file ~/Desktop/Wasm/special.conf --nats-host="127.0.0.1" --nats-port=4959` which executed as expected and also exited normally by `cargo run -- down -- all` stopping all components successfully

verified all possible failure cases , such as 
1) Providing directory path instead of file : resulted in `The provided NATS config File ["/home/aditya/Desktop/Wasm/"] is not a valid File`
2) Ommiting `--nats-host` and `--nats-port` : resulted in `the following required arguments were not provided:
  --nats-host <NATS_HOST>`

### DIFF
--- a/crates/wash-cli/src/up/mod.rs
+++ b/crates/wash-cli/src/up/mod.rs
@@ -66,6 +66,15 @@ pub struct NatsOpts {
         requires = "nats_remote_url"
     )]
     pub nats_credsfile: Option<PathBuf>,
+    /// Optional path to a NATS config file
+    /// NOTE: If your configuration changes the address or port to listen on from 0.0.0.0:4222, ensure you set --nats-host and --nats-port
+    #[clap(
+        long = "nats-config-file",
+        env = "NATS_CONFIG",
+        requires = "nats_host",
+        requires = "nats_port"
+    )]
+    pub nats_configfile: Option<PathBuf>,
 
     /// Optional remote URL of existing NATS infrastructure to extend.
     #[clap(long = "nats-remote-url", env = "NATS_REMOTE_URL")]
@@ -122,6 +131,7 @@ impl From<NatsOpts> for NatsConfig {
             remote_url: other.nats_remote_url,
             credentials: other.nats_credsfile,
             websocket_port: other.nats_websocket_port,
+            config_path: other.nats_configfile,
         }
     }
 }
@@ -395,6 +405,7 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
             remote_url: cmd.nats_opts.nats_remote_url,
             credentials: cmd.nats_opts.nats_credsfile.clone(),
             websocket_port: cmd.nats_opts.nats_websocket_port,
+            config_path: cmd.nats_opts.nats_configfile,
         };
         start_nats(&install_dir, &nats_binary, nats_config).await?;
         Some(nats_binary)
--- a/crates/wash-lib/src/start/nats.rs
+++ b/crates/wash-lib/src/start/nats.rs
@@ -1,7 +1,6 @@
+use anyhow::{bail, Result};
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
-
-use anyhow::{bail, Result};
 use tokio::fs::{metadata, write};
 use tokio::process::{Child, Command};
 use tracing::warn;
@@ -125,6 +124,7 @@ pub struct NatsConfig {
     pub remote_url: Option<String>,
     pub credentials: Option<PathBuf>,
     pub websocket_port: u16,
+    pub config_path: Option<PathBuf>,
 }
 
 /// Returns a standalone NATS config with the following values:
@@ -144,6 +144,7 @@ impl Default for NatsConfig {
             remote_url: None,
             credentials: None,
             websocket_port: 4223,
+            config_path: None,
         }
     }
 }
@@ -168,6 +169,7 @@ impl NatsConfig {
         remote_url: String,
         credentials: PathBuf,
         websocket_port: u16,
+        config_path: Option<PathBuf>,
     ) -> Self {
         NatsConfig {
             host: host.to_owned(),
@@ -177,6 +179,7 @@ impl NatsConfig {
             remote_url: Some(remote_url),
             credentials: Some(credentials),
             websocket_port,
+            config_path,
         }
     }
     /// Instantiates config for a standalone NATS server. Unless you're looking to extend
@@ -263,6 +266,7 @@ where
     T: Into<Stdio>,
 {
     let host_addr = format!("{}:{}", config.host, config.port);
+
     // If we can connect to the local port, NATS won't be able to listen on that port
     if tokio::net::TcpStream::connect(&host_addr).await.is_ok() {
         bail!(
@@ -271,31 +275,54 @@ where
             config.port
         );
     }
-    let child = if let Some(parent_path) = bin_path.as_ref().parent() {
+
+    let bin_path_ref = bin_path.as_ref();
+
+    if let Some(parent_path) = bin_path_ref.parent() {
         let config_path = parent_path.join(NATS_SERVER_CONF);
         let host = config.host.clone();
         let port = config.port;
-        config.write_to_path(&config_path).await?;
-        Command::new(bin_path.as_ref())
-            .stderr(stderr)
+
+        let mut cmd_args = vec![
+            "-js".to_string(),
+            "--addr".to_string(),
+            host,
+            "--port".to_string(),
+            port.to_string(),
+            "--pid".to_string(),
+            parent_path
+                .join(NATS_SERVER_PID)
+                .to_string_lossy()
+                .to_string(),
+            "--config".to_string(),
+        ];
+
+        if let Some(nats_cfg_path) = &config.config_path {
+            anyhow::ensure!(
+                nats_cfg_path.is_file(),
+                "The provided NATS config File [{:?}] is not a valid File",
+                nats_cfg_path
+            );
+
+            cmd_args.push(nats_cfg_path.to_string_lossy().to_string());
+        } else {
+            config.write_to_path(&config_path).await?;
+            cmd_args.push(config_path.to_string_lossy().to_string());
+        }
+
+        let child = Command::new(bin_path_ref)
+            .stderr(stderr.into())
             .stdin(Stdio::null())
-            .arg("-js")
-            .arg("--config")
-            .arg(config_path)
-            .arg("--addr")
-            .arg(host)
-            .arg("--port")
-            .arg(port.to_string())
-            .arg("--pid")
-            .arg(parent_path.join(NATS_SERVER_PID))
+            .args(&cmd_args)
             .spawn()
-            .map_err(anyhow::Error::from)
+            .map_err(anyhow::Error::from)?;
+
+        wait_for_server(&host_addr, "NATS server")
+            .await
+            .map(|()| child)
     } else {
         bail!("could not write config to disk, couldn't find download directory")
-    }?;
-    wait_for_server(&host_addr, "NATS server")
-        .await
-        .map(|()| child)
+    }
 }
 
 /// Helper function to get the path to the NATS server pid file
@@ -444,6 +471,7 @@ mod test {
             "connect.ngs.global".to_string(),
             creds.clone(),
             4204,
+            None,
         );
 
         config.write_to_path(creds.clone()).await?;


Aditya1404Sal:
I'm running into a problem here,
so a really basic config file (nats.conf) looks like this:
```bash
jetstream {
    domain=core
    store_dir="/tmp/wash-jetstream-4222"
}


websocket {
    port: 4223
    no_tls: true
}
```
 `nats.conf` being a low profile file format does not have compatibility with any parsing libraries in rust , even nats-server uses a golang implementation of the parser for starting the server, so my current action plan is to build a function that populates the NatsConfig object and pass it onto `start_nats` , moreover I thought it would be better if `start_nats_server` did not save the config file alongside starting the server. 
My first thought for the parser was to first convert the `.conf` to a `toml` file and then work with it (which was wrong) , however there are other options I am looking into which are 
1) using `config` crate and defining a custom parse manifest

Am I going in the right direction ?

brooksmtownsend (changes requested):
Hey @Aditya1404Sal ! This scaffolding is looking great so far. I wanted to save you a bit of time just to let you know that in this case, where a NATS config is passed as a flag, we should actually use that configuration directly instead of trying to parse it and write an additional config. 

Since the NATS configuration file supports _many_ more options than we can handle individually, we will just want to pass in the file path as the NATS config file that `wash` will supply.

So, for example, running `wash up --nats-config ./my-nats-server.conf` should actually end up with us running:
```
nats-server -js --config /path/to/my-nats-server.conf
```

Does that make sense? It should make it a lot easier on you regarding parsing! If you still want to continue parsing to infer some of the variables we use to connect that's a nice addition, but we shouldn't have to write out another file 😄

Aditya1404Sal:
> Hey @Aditya1404Sal ! This scaffolding is looking great so far. I wanted to save you a bit of time just to let you know that in this case, where a NATS config is passed as a flag, we should actually use that configuration directly instead of trying to parse it and write an additional config.
> 
> Since the NATS configuration file supports _many_ more options than we can handle individually, we will just want to pass in the file path as the NATS config file that `wash` will supply.
> 
> So, for example, running `wash up --nats-config ./my-nats-server.conf` should actually end up with us running:
> 
> ```
> nats-server -js --config /path/to/my-nats-server.conf
> ```
> 
> Does that make sense? It should make it a lot easier on you regarding parsing! If you still want to continue parsing to infer some of the variables we use to connect that's a nice addition, but we shouldn't have to write out another file 😄

I just realised that right now ,the `start_nats_server` actually uses the start nats binary so the `nats-server -c /path_to_config` would work fine.  i'll revert the changes and work out a simple function `start_nats_config` or something that will go inside the `start_nats` function and trigger if wash UpCommand has the --nats_config file. 

Thank you so much for taking time and guiding me, your explanation makes a lot of sense

I feel dumb about not knowing this before, but I'm confident I can work further on it.

brooksmtownsend:
@Aditya1404Sal This is why I love draft PRs! It's great to get assumptions checked and start with an initial implementation. Feel free to just open up a PR once you get something working with the `-c` flag, and feel welcome to do another draft/WIP PR to ask questions

Aditya1404Sal:
> @Aditya1404Sal This is why I love draft PRs! It's great to get assumptions checked and start with an initial implementation. Feel free to just open up a PR once you get something working with the `-c` flag, and feel welcome to do another draft/WIP PR to ask questions

Sounds great to me! I'll get right to it :smile:

Aditya1404Sal:
There is a small problem i'm encountering, Normally when `wash up` is carried out with the host and port flags (ie: `--nats-host` and `nats-port`) the host and port variables are known and by that logic, the function `wait_for_server` can be carried out to know if the respective server has been provisioned or not , However in the case of the forementioned new feature `wash up --nats-config ./path` , there isn't a way to check if the host and port have been provisioned since nats.conf can't be accessed. Which ends up ending the `wash up` workflow but the `nats-server` command still executes in the background. Currently  `wash up --nats-config ./path` only works if the config file does not have port and host mentioned.
So to fix this there is a need to validate whether or not server has spun up by introducing a wait time or
I was thinking if nats.log and  [debug & trace flag](https://docs.nats.io/running-a-nats-service/configuration/logging) could be useful for checking the server status 
what do you think @brooksmtownsend ?

crates/wash-lib/src/start/nats.rs
```diff
@@ -262,40 +266,62 @@ where
...
-    }
-    let child = if let Some(parent_path) = bin_path.as_ref().parent() {
-        let config_path = parent_path.join(NATS_SERVER_CONF);
-        let host = config.host.clone();
-        let port = config.port;
-        config.write_to_path(&config_path).await?;
-        Command::new(bin_path.as_ref())
+    if let Some(natscfg_path) = config.config_path {
+        // Use the existing config path
+        let child = Command::new(bin_path.as_ref())
```
vados-cosmonic, line 271: I think we can get away with checking for this later -- what do you think? 

This is a NIT but I think we can start building the `Command` mutably (i.e. `let mut child = Command::new(...)`), add the shared stuff, *then* check for whether a NATS config path was provided and return early succinctly!

Something like:

```rust
let mut child = Command::new(bin_path.as_ref())
.stderr(stderr)
.stdin(Stdio::null())
.arg("-js")
.arg("--config");

if let Some(nats_cfg_path) = config.config_path {
child = child.arg(natscfg_path.join(NATS_SERVER_CONF))
.arg("--pid")
.arg(nats_cfg_path.join("NATS_SERVER_PID")
.spawn()
......
}

// add the rest

```

This also makes me think -- if we're expecting the NATS config path to be a directory, can we make this explicit in the name of the argument? We should also maybe check that the path is a directory.

I know that when I see `path` I think file -- so I might think that `--nats-config` should be the path to a NATS config file, not the directory it's in. 

Something like `--nats-config-dir`/`--nats-dir`? Because we actually *write* the PID file into that directory too. It's not like we can pass a readonly directory mount in.

Aditya1404Sal: So I'll refactor `--nats-config` to `--nats-config-dir` and add a `is_dir()` check as you suggested.
I'll make `child` a mutable too

Aditya1404Sal:
```rust
let mut child = Command::new(bin_path.as_ref())
.stderr(stderr)
.stdin(Stdio::null())
.arg("-js")
.arg("--config");
```
The temporary value of child gets freed at the end of the statement, so the Command::new(bin_path.as_ref()) does not persist. It needs `.spawn()` attached to the command call.
I initially planned on taking this approach :point_down:  , however `wait_for_server` has a dependency on `host_addr` which will have conflicts with the host and port specified in the nats.conf file, so I think it would be best if we checked for `nats_cfg_path` in advance

```rust
async fn start_nats_server_2<P, T>(bin_path: P,config: NatsConfig,stderr: T,) -> Result<tokio::process::Child>
...
....
//rest of the code
.....
.......
    let bin_path_ref = bin_path.as_ref();

    if let Some(parent_path) = bin_path_ref.parent() {
        let config_path = parent_path.join(NATS_SERVER_CONF);
        let host = config.host.clone();
        let port = config.port;

        let mut cmd_args = vec!["-js".to_string(), "--config".to_string()];

        if let Some(nats_cfg_path) = &config.config_path {
            if nats_cfg_path.is_dir(){
                cmd_args.push(nats_cfg_path.join(NATS_SERVER_CONF).to_string_lossy().to_string());
                cmd_args.push("--pid".to_string());
                cmd_args.push(nats_cfg_path.join(NATS_SERVER_PID).to_string_lossy().to_string());
            }else{
                bail!("couldn't find directory path");
            }
        }  else {
            config.write_to_path(&config_path).await?;
            cmd_args.push(config_path.to_string_lossy().to_string());
            cmd_args.push("--addr".to_string());
            cmd_args.push(host);
            cmd_args.push("--port".to_string());
            cmd_args.push(port.to_string());
        }

        let child = Command::new(bin_path_ref)
            .stderr(stderr.into())
            .stdin(Stdio::null())
            .args(&cmd_args) //additional arguements get added here
            .spawn()
            .map_err(anyhow::Error::from)?;

        wait_for_server(&host_addr, "NATS server")
            .await
            .map(|()| child)
    }
....
....
    }
}
```

crates/wash-cli/src/up/mod.rs
```diff
@@ -66,6 +66,14 @@ pub struct NatsOpts {
...
     )]
     pub nats_credsfile: Option<PathBuf>,
+    /// Optional path to a NATS configuration file | NOTE: Works only if accompanied by nats-host and nats-port
+    #[clap(
+        long = "nats-config-dir",
+        env = "NATS_CONFIG",
+        requires = "nats_host",
+        requires = "nats_port"
+    )]
+    pub nats_configfile_dir: Option<PathBuf>,
```
vados-cosmonic, line 76:
```suggestion
    pub nats_config_dir: Option<PathBuf>,
```

This change will break other stuff, but we don't need "configfile" right? 

Also, please note in the `///` comment that it is the path to a NATS config folder (i.e. one that contains a NATS config file -- named <file name>)

crates/wash-lib/src/start/nats.rs
```diff
@@ -271,31 +275,69 @@ where
...
+                cmd_args.push("--pid".to_string());
+                cmd_args.push(
+                    nats_cfg_path
+                        .join(NATS_SERVER_PID)
+                        .to_string_lossy()
+                        .to_string(),
+                );
+            } else {
+                bail!(
+                    "The provided config path is not a directory: {:?}",
```
vados-cosmonic, line 312:
```suggestion
                    "The provided NATS config directory [{nats_cfg_path}] is not a valid directory",
```

crates/wash-lib/src/start/nats.rs
```diff
@@ -271,31 +275,69 @@ where
...
+            "-js".to_string(),
+            "--addr".to_string(),
+            host,
+            "--port".to_string(),
+            port.to_string(),
+            "--config".to_string(),
+        ];
+
+        if let Some(nats_cfg_path) = &config.config_path {
+            if nats_cfg_path.is_dir() {
```
vados-cosmonic, line 296: Consider using [`anyhow::ensure`](https://docs.rs/anyhow/latest/anyhow/macro.ensure.html) here!

Something like:

```rust
anyhow::ensure!(nats_cfg_path.is_dir(), "The provided config...")
```

With that, we can remove one `if` branch here and make it easier to read

vados-cosmonic (commented):
Just a *few* more changes and I think we're good to go! Thanks for the refactor @Aditya1404Sal !

crates/wash-cli/src/up/mod.rs
```diff
@@ -66,6 +66,15 @@ pub struct NatsOpts {
...
     pub nats_credsfile: Option<PathBuf>,
+    /// Optional path to a NATS config Folder (i.e. the folder that contains the file named nats.conf)
+    /// NOTE: Works only if accompanied by nats-host and nats-port
+    #[clap(
+        long = "nats-config-dir",
+        env = "NATS_CONFIG",
+        requires = "nats_host",
+        requires = "nats_port"
+    )]
+    pub nats_config_dir: Option<PathBuf>,
```
brooksmtownsend, line 77: I think it would be a more clear experience to have this configuration flag to be the path directly to the NATS configuration file, that way you can point directly at a file that may be named anything and use that as the NATS configuration. Was this an approach you took before that had some downsides?

crates/wash-cli/src/up/mod.rs
```diff
@@ -66,6 +66,15 @@ pub struct NatsOpts {
         requires = "nats_remote_url"
     )]
     pub nats_credsfile: Option<PathBuf>,
+    /// Optional path to a NATS config Folder (i.e. the folder that contains the file named nats.conf)
+    /// NOTE: Works only if accompanied by nats-host and nats-port
```
brooksmtownsend, line 70: nit: 
```suggestion
    /// NOTE: If your configuration changes the address or port to listen on from 0.0.0.0:4222, ensure you set --nats-host and --nats-port
```

brooksmtownsend (changes requested):
I think all the code changes look great here, I just have one request to change the command line flag from specifying a dir where a `nats.conf` file has to exist to specifying a path directly to a configuration file (which may be named something other than `nats.conf`). Let me know if this makes sense / if you have questions!

brooksmtownsend (approved):
Looks great! @Aditya1404Sal are you able to take a look at the failing test? Is that a flake?

Aditya1404Sal:
> Looks great! @Aditya1404Sal are you able to take a look at the failing test? Is that a flake?
> 
> https://github.com/wasmCloud/wasmCloud/actions/runs/9666490749/job/26666108227?pr=2271

The failing test are 
`integration_help_subcommand_check()`
and
`test_subcommand()`

I use `ubuntu 22.04` for development, so I'm not sure why the test is failing for windows 8 cores
There are 2 tests that are flaky which are `can_download_and_start_nats` and `can_download_and_start_wasmcloud`

vados-cosmonic:
Thanks again @Aditya1404Sal ! It looks like some fixes merged into main fixed those tests :) -- a rebase was all we needed

Aditya1404Sal:
> Thanks again @Aditya1404Sal ! It looks like some fixes merged into main fixed those tests :) -- a rebase was all we needed

Thank you @vados-cosmonic and @brooksmtownsend , for being patient with this PR. Being a beginner in Opensource Wasmcloud community has been very supportive from the start especially in panic inducing moments,  I'll keep learning Wasm and work my way up to more complex issues

Issue #2190, opened 2024-05-28
[RFC] Secrets in wasmCloud


protochron:
### RFC Title

RFC: Secrets in wasmCloud

### Introduction

The goal of this RFC is to propose a new API specification for *secrets backends*, which are programs that will listen on NATS for requests for secrets and be able to satisfy them. This RFC also outlines changes to the wasmCloud host opt in to the feature and to connect to these secret service backends in order to provide secrets to components and providers. This new API is designed so that many secrets backends can be connected to a wasmCloud cluster so that end applications can use multiple secrets providers at runtime.

### Motivation

Production applications often require secrets at runtime to access internal or external systems. Up until now our only option for providing secrets was to link a component to a capability provider such as the Vault KV provider that could access secrets. This approach has several drawbacks:

- You must run a dedicated provider for ex. Vault that need their own secrets in order to connect to upstream providers.
- Secrets need to be retrieved every time a component or provider needs them which can be inefficient.
- Secrets management is tightly bound to a single provider and interface type, which means that the component's implementation is irrevocably tied to it. This violates the spirit of what wasmCloud is trying to solve for adopters and makes it difficult to develop applications that consume secrets.

## Objectives

- Outline a new NATS-based API for secrets backends. A wasmCloud cluster *must* have a running secrets backend in order to provide secrets to components and providers. A wasmCloud cluster *must* support multiple secrets backends running at the same time.
- Define the changes to the host required to implement support for securely retrieving secrets from a secrets backend.
- Define a WIT interface used for retrieving secrets from a host.
- Outline the necessary changes to a wadm manifest to allow for secrets support in target and source configurations.

The implementation of this RFC will result in a reference implementation based on a NATS Key Value secrets service backend. In addition, we will also work on additional secret backends, including Vault, OpenBAO and others. We may opt to include the NATS KV-backend as part of Wadm, enabled by passing in a flag and a xkey used for encrypting the underlying NATS KV bucket.

## Design
Secrets are intended to be very similar to config in wasmCloud, except that they are explicit *references* to data that lives outside of the wasmCloud cluster. It is the host's responsibility to retrieve secrets from a secrets backend and provide them to components and providers that request them. Secrets backends will be reachable by hosts via NATS using known NATS subject prefixes. Each secrets backend will satisfy a common API so that anyone can write a backend and add an implementation to their cluster.

[...]

PR #2294, opened 2024-06-11
feat: add wasmcloud secrets crate and initial implementation of a NATS KV backend

protochron:
## Feature or Problem
This adds an example implementation of a _secrets backend_ for wasmCloud, backed by NATS KV. The API is defined in #2190, but the idea is that we can have many secrets backends running in a wasmCloud cluster, each of which is listening on a specific prefix. Hosts will communicate with these backends to retrieve secrets on behalf of applications before they are instantiated for the first time, similar to what we do with config.

Common types and errors are stored in a separate crate so that other Rust implementations can reference them. As a follow up we should add OpenAPI specifications for these types to make it easier to develop backends written in other languages.

The NATS KV backend implements the secrets API in addition to several unique endpoints for managing secrets and mappings between components/providers and keys. These endpoints are _not_ expected to be duplicated in any other implementation, but are only available to define an interface for managing secrets in NATS KV. Other backends will likely have their own process for managing secrets out of band and only be concerned with retrieving them.

Secret payloads are encrypted using a key provided by the host on each request and a dedicated encryption key generated by the backend. For more information on why there are two keys, [see this link](https://cr.yp.to/ecdh.html) that decribes how Curve25519 works. The host is expected to provide two forms of _context_: the _runtime_ context, which is information about the host itself including its labels, and the _build_ context of the component or provider, which is the signed JWT that it was built with. For now, both are expected to be present so that a secrets backend can pass that information to upstream APIs in order to make a policy decision on whether or not the component or provider is allowed to access that secret. For now, the host JWT will need to be self signed. This will change in future releases of wasmCloud where we will have better ways of providing signing keys to new hosts.

In this particular implementation, secrets are stored encrypted in the KV bucket with a xkey, similar to how other secrets managers store secrets at rest.

## Related Issues
#2190 

## Release Information
next

## Testing
See the integration tests in the backend implementation


crates/secrets-nats-kv/src/api.rs
```diff
@@ -0,0 +1,660 @@
...
+                        None => {
+                            let _ = self
+                                .client
+                                .publish(reply, "no entity provided".into())
+                                .await;
+                            continue;
+                        }
+                    };
+
+                    let payload = msg.payload;
```
thomastaylor312, line 281: Super duper nit: I don't know if it would technically make it any more efficient or not since JSON allocates from borrowed data, but if you import the `bytes::Buf` trait, you can shorten all these to be `serde_json::from_reader(msg.payload.reader())` while consuming the payload

crates/secrets-nats-kv/src/api.rs
```diff
@@ -0,0 +1,660 @@
...
+            let str = serde_json::to_string(&values)?;
+            state.put(entity.clone(), str.into()).await?;
+        }
+        let s = c.get_stream(&self.lock_stream_name()).await?;
+        s.delete_message(seq).await?;
+        Ok(())
+    }
+
+    async fn remove_mapping(&self, entity: String, values: HashSet<String>) -> anyhow::Result<()> {
+        let c = jetstream::new(self.client.clone());
```
thomastaylor312, line 489: Nit: The logic around getting the lock and the stream could probably be abstracted out of this function and `add_mapping` to reduce duplication. Probably could implement the guard thing you mention in the TODO below

crates/secrets-nats-kv/src/api.rs
```diff
@@ -0,0 +1,660 @@
...
+        state.put(entity.clone(), new_vals.into()).await?;
+
+        // TODO all this locking logic should probably be wrapped up into some sort of a wrapper
+        // struct that does the right thing when it goes out of scope
+        let s = c.get_stream(&self.lock_stream_name()).await?;
+        s.delete_message(seq).await?;
+        Ok(())
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
thomastaylor312, line 518: Nit: Consider a config struct or a builder here

crates/secrets-types/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "wasmcloud-secrets-types"
```
thomastaylor312, line 2: I didn't see a workflow for publishing this crate. Should we add that here or is it ok to wait until after this merges?

thomastaylor312 (commented):
None of these are blocking, but some of them aren't quite nits either. Leaving this as just a comment so we have some time to discuss and then I can come back and give the ✅

crates/secrets-types/Cargo.toml
protochron: I was planning on waiting until after this merges

thomastaylor312 (approved):
Good to go here. Feel free to mark the other nits as resolved when you're ready

crates/secrets-types/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+/// The key of a NATS header containing the wasmCloud host's public xkey used to encrypt a secret request.
+/// It is also used to encrypt the response so that only the requestor can decrypt it.
+pub const WASMCLOUD_HOST_XKEY: &str = "WasmCloud-Host-Xkey";
+pub const RESPONSE_XKEY: &str = "Server-Response-Xkey";
+
+/// The request context for retrieving a secret
+#[derive(Serialize, Deserialize, Default)]
+pub struct Context {
+    /// The component or provider's signed JWT.
+    pub entity_jwt: String,
```
joonas, line 19: Perhaps this is covered somewhere, but after taking another look at this, I'm wondering what's the reasoning to prefer `entity_jwt: String` over `component_jwt: Option<String>` and `provider_jwt: Option<String>`?

thomastaylor312: I prefer not having separate jwt fields for components and providers. You should always have a JWT and it doesn't matter much which thing it is coming from

protochron: I agree with @thomastaylor312 which is why I implemented it this way. The JWT should be present regardless of the type, so distinguishing between them doesn't matter.

PR #2295, opened 2024-06-11
debug: replicate trace with dog fetcher

vados-cosmonic:
This commit contains experimental code used to debug/replicate the o11y traces for making a call with http-client & http-provider.

Running this requires the following hackery:

- running the docker compose for o11y
- (re) building dog-fetcher
- modifying the WADM w/ dog fetcher (done by this commit)
- build & create PAR for http-client
- build & create PAR for http-server
- set WASMCLOUD_OVERRIDE_TRACES_ENDPOINT before `wash up`
- replacing existing wasmcloud host (in `~/.wash/downloads/v1.0.2`)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -227,19 +227,34 @@ struct Args {
     enable_logs: Option<bool>,
 
     /// Overrides the OpenTelemetry endpoint used for emitting traces, metrics and logs. This can also be set with `OTEL_EXPORTER_OTLP_ENDPOINT`.
-    #[clap(long = "override-observability-endpoint")]
+    #[clap(
+        long = "override-observability-endpoint",
+        env = "WASMCLOUD_OVERRIDE_OBSERVABILITY_ENDPOINT",
+    )]
     observability_endpoint: Option<String>,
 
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -227,19 +227,34 @@ struct Args {
...
+    )]
     traces_endpoint: Option<String>,
 
     /// Overrides the OpenTelemetry endpoint used for emitting metrics. This can also be set with `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`.
-    #[clap(long = "override-metrics-endpoint", hide = true)]
+    #[clap(
+        long = "override-metrics-endpoint",
+        env = "WASMCLOUD_OVERRIDE_METRICS_ENDPOINT",
+        hide = true,
+    )]
```
brooksmtownsend, line 249: Just a note I'm going to remove these since we decided back in the day to override the metrics/traces/logs endpoints using the OTEL specific environment variables instead

vados-cosmonic: Hey so I saw the comment above these lines, and tried those other ENV vars and they didn't work well -- I was still seeing it trying to send traces the wrong place. :man_shrugging:

PR #2300, opened 2024-06-11
chore(wash)!: Replace actor references by component in wash-lib crate

nitame:
## Related Issues
#1827


crates/wash-lib/src/wait.rs
```diff
@@ -126,7 +126,7 @@ pub async fn wait_for_component_scaled_event(
         }
 
         match cloud_event.event_type.as_str() {
-            "com.wasmcloud.lattice.component_scaled" | "com.wasmcloud.lattice.actor_scaled" => {
+            "com.wasmcloud.lattice.component_scaled" | "com.wasmcloud.lattice.actor_scaled" => { // NOTE(nitame): Is actor still need to be part of match case?
```
nitame, line 129: If actor_scaled and actor_scale_failed do not need to be matched any more. I can remove them and the handling of both actor_id and component_id.

thomastaylor312: @brooksmtownsend I think we can drop the actor_scale* events now, right? I'm not remembering us using them anywhere anymore

crates/wash-lib/src/app.rs
```diff
@@ -451,7 +451,7 @@ mod test {
 
         // test url
         let url = AppManifestSource::from_str(
-            "https://raw.githubusercontent.com/wasmCloud/examples/main/actor/hello/wadm.yaml",
+            "https://raw.githubusercontent.com/wasmCloud/wasmCloud/main/examples/rust/components/http-hello-world/wadm.yaml",
```
thomastaylor312, line 454: Thanks for updating these to the new URL as well!

crates/wash-lib/src/component.rs
```diff
@@ -62,12 +62,12 @@ pub async fn scale_component(
...
 ) -> Result<ComponentScaledInfo> {
     // If timeout isn't supplied, override with a longer timeout for starting component
-    let timeout_ms = timeout_ms.unwrap_or(DEFAULT_START_ACTOR_TIMEOUT_MS);
+    let timeout_ms = timeout_ms.unwrap_or(DEFAULT_START_COMPONENT_TIMEOUT_MS);
 
     // Create a receiver to use with the client
     let mut receiver = client
         .events_receiver(vec![
-            "actor_scaled".to_string(),
+            "actor_scaled".to_string(), // NOTE(nitame): Is these events_types still relevant?
```
thomastaylor312, line 70: Just calling this out here if @brooksmtownsend confirms that we can remove these now

thomastaylor312 (approved):
This looks great! Thanks @nitame! 

Waiting to merge to @brooksmtownsend can get back on the actor_scale events

crates/wash-lib/src/component.rs
brooksmtownsend: Yep confirmed, we can remove the `actor_scaled` event since it's not emitted by post-1.0 hosts. All `component_scaled` now 🚀

crates/wash-lib/src/wait.rs
brooksmtownsend: Yep! We can remove `actor_scaled`. Nothing wrong with it being in there but it'll be nice to get rid of the extra case + handling of actor_id

nitame:
Hello, thanks for the feedback. I removed the cloud events actor references and applied cargo fmt.

PR #2313, opened 2024-06-12
fix(wash-lib,wash-cli): do not dowmload wadm on every invocation of wasm up

### LINKED ISSUE #2308: [BUG] wadm binary is fetched on every invocation of wash up
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

It is related to change of `wadm --version` output. recent versions of wadm returns something like `wadm-cli 0.12.0`. Old version returned `wadm 0.10.0`

Check for version string was not updated in wash sources :-(

Actually, I found the bug trying to solve the #2134 digging into source code

### Steps to reproduce

1. check wadm creation time with `ls -l ~/.wash/downloads/wadm`
2. run `wash up`
3. check creation time again, you have a fresh copy

### Expected behavior

Should not download `wadm` for every invocation

### Environment

- OS: All
- Shell: All
- wasmcloud Version 1.0.3

_No response_

deem0n:
## Feature or Problem

Using regexp to detect wadm version instead of constant prefix. Also added some messages to the user, so he will know wadm version and download URL for wadm if download occurs.

## Related Issues

Resolves #2308 

## Release Information

next

## Consumer Impact

I think all users are affected by the bug. So please include the fix to the next release.

## Testing

run `wash up` and check messages on console.

### Unit Test(s)

nope

### Manual Verification

run `cargo run -- up` in the folder `wasmCloud/crates/wash-cli`

### DIFF
--- a/crates/wash-lib/src/start/wadm.rs
+++ b/crates/wash-lib/src/start/wadm.rs
@@ -74,14 +74,27 @@ where
         // Check version to see if we need to download new one
         if let Ok(output) = Command::new(&wadm_bin_path).arg("--version").output().await {
             let stdout = String::from_utf8_lossy(&output.stdout).to_string();
-            if stdout.replace("wadm", "").trim() == version.trim_start_matches('v') {
+            println!("👀 Found wadm version on the disk: {}", stdout.trim_end());
+            let re = regex::Regex::new(r"^wadm[^\s]*").unwrap();
+            if re.replace(&stdout, "").to_string().trim() == version.trim_start_matches('v') {
                 // wadm already exists, return early
+                println!("✅ Using wadm version [{}]", &version);
                 return Ok(wadm_bin_path);
             }
         }
     }
     // Download wadm tarball
-    download_binary_from_github(&wadm_url(os, arch, version), dir, WADM_BINARY).await
+    println!(
+        "🎣 Downloading new wadm from {}",
+        &wadm_url(os, arch, version)
+    );
+
+    let res = download_binary_from_github(&wadm_url(os, arch, version), dir, WADM_BINARY).await;
+    if let Ok(ref path) = res {
+        println!("🎯 Saved wadm to {}", path.display());
+    }
+
+    res
 }
 
 /// Downloads the wadm binary for the architecture and operating system of the current host machine.
--- a/crates/wash-cli/src/up/mod.rs
+++ b/crates/wash-cli/src/up/mod.rs
@@ -846,7 +846,7 @@ async fn run_wasmcloud_interactive(
     });
 
     if output_kind != OutputKind::Json {
-        println!("🏃 Running in interactive mode.",);
+        println!("🏃 Running in interactive mode.");
         if let Some(ref manifest_path) = wadm_manifest {
             println!(
                 "🚀 Deploying WADM manifest at [{}]",


crates/wash-lib/src/start/wadm.rs
```diff
@@ -74,14 +74,25 @@ where
...
         if let Ok(output) = Command::new(&wadm_bin_path).arg("--version").output().await {
             let stdout = String::from_utf8_lossy(&output.stdout).to_string();
-            if stdout.replace("wadm", "").trim() == version.trim_start_matches('v') {
+            print!("👀 Found wadm version: {}", stdout.trim_end());
+            let re = regex::Regex::new(r"^wadm[^\s]*").unwrap();     
+            if re.replace(&stdout, "").to_string().trim() == version.trim_start_matches('v') {
                 // wadm already exists, return early
+                println!(" Using it.✅");
                 return Ok(wadm_bin_path);
             }
```
thomastaylor312, line 83: This should probably have a `print!` newline in an else block in case it doesn't find the right version

vados-cosmonic:
Hey @deem0n would you mind using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) to structure your commit message? Something like:

```
fix(wash-lib,wash-cli): avoid repeated downloads of wadm binary
```

If you want you could also split up the commits to `fix(wash-lib): ...` and `fix(wash-cli): ...` if you want!

crates/wash-lib/src/start/wadm.rs
```diff
@@ -74,14 +74,25 @@ where
         // Check version to see if we need to download new one
         if let Ok(output) = Command::new(&wadm_bin_path).arg("--version").output().await {
             let stdout = String::from_utf8_lossy(&output.stdout).to_string();
-            if stdout.replace("wadm", "").trim() == version.trim_start_matches('v') {
+            print!("👀 Found wadm version: {}", stdout.trim_end());
+            let re = regex::Regex::new(r"^wadm[^\s]*").unwrap();     
+            if re.replace(&stdout, "").to_string().trim() == version.trim_start_matches('v') {
                 // wadm already exists, return early
+                println!(" Using it.✅");
```
vados-cosmonic, line 81: It might be nice to make this a bit more explicit -- `"using wadm version [{}]"` or something like that?

And could we keep the emoji location consistent here -- emoji at the front?

deem0n:
Ok, will do today

crates/wash-lib/src/start/wadm.rs
deem0n: Refactored!

crates/wash-lib/src/start/wadm.rs
deem0n: Done with emoji, also separated work into 2 commits

vados-cosmonic: Thanks!

crates/wash-lib/src/start/wadm.rs
deem0n: @thomastaylor312: I slightly adjusted the code.

thomastaylor312 (approved):

vados-cosmonic (approved):
LGTM :rocket:

Looks like a test was flaky and `cargo fmt` needed to be run.

PR #2329, opened 2024-06-13
ci(wasmcloud): enable rust cache for windows

brooksmtownsend:
## Feature or Problem
This PR enables Rust cache for Windows CI when merging into main. This should help speed up the windows builds

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -160,7 +160,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           shared-key: windows-latest-8-cores-shared-cache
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
Yeah not sure why we were excluding main here... but LGTM

brooksmtownsend:
@vados-cosmonic This really just came down to making sure that clean builds would work, but since the windows builds are already running in the PR and we will do clean builds for releases so... if that bites us later I'll take the blame in order to speed up our merge queue by ~10 minutes avg 😄

vados-cosmonic:
Yeah, I don't know why we made that decision in the first place -- releases are a great time to do clean builds, check artifacts and what not. There is mostly no need to avoid cache or even generate artifacts on random commits to main. We didn't always have such a nice release PR flow though, so glad we have that now :rocket:

thomastaylor312 (approved):

PR #2330, opened 2024-06-13
feat(tracing): add support for flame graph generation

rvolosatovs:
## Feature or Problem

Generate flame graphs using https://docs.rs/tracing-flame/latest/tracing_flame/index.html if configured to do so, also refactor the observability function to do that in a sane way

### DIFF
--- a/crates/tracing/src/lib.rs
+++ b/crates/tracing/src/lib.rs
@@ -6,6 +6,7 @@ pub use opentelemetry::{
     metrics::{Counter, Histogram, Meter, Unit},
     KeyValue,
 };
+use std::path::Path;
 use wasmcloud_core::logging::Level;
 use wasmcloud_core::OtelConfig;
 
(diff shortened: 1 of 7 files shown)


crates/tracing/src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub use opentelemetry::{
     metrics::{Counter, Histogram, Meter, Unit},
     KeyValue,
 };
+use std::path::Path;
```
thomastaylor312, line 9: Super nit: Should this import go up top since it is `std`?

thomastaylor312 (approved):

brooksmtownsend:
Going to give this a shot tomorrow with our tracing infra 😄

crates/tracing/src/lib.rs
thomastaylor312: Resolving this so it can merge since this is just a super nit

rvolosatovs: I really was not sure due to all the feature-gated imports. I think normally you'd define a feature-gated module instead, where all imports are just imports (no features)

PR #2334, opened 2024-06-13
feat(host): add link name to wRPC invocations

vados-cosmonic:
## Feature or Problem

This commit adds the "link-name" header to invocations performed by the host using wRPC.

Resolves https://github.com/wasmCloud/wasmCloud/issues/2332

PR #2335, opened 2024-06-13
fix(providers,examples): use link name for kvredis provider

vados-cosmonic:
## Feature or Problem

The kvredis provider not using link names to distinguish connected components has been a problem for a while. This PR fixes that *and* updates the http-keyvalue-counter example to be able to switch between different links, with a query parameter called `link_name`.

Resolves


thomastaylor312 (commented):
This looks good, just wanted to check if this should be rebased on top of #2334

thomastaylor312 (approved):

PR #2341, opened 2024-06-14
fix(provider-sdk)!: change inputs to Provider::delete_link_*()

vados-cosmonic:
This commit introduces new methods to the `Provider` trait for deleting links -- `delete_configured_link_as_source()` and `delete_configured_link_as_target()`.

Thse new methods are meant to address an issue with deleting links in that the name of the link was unavailable -- only the source of the link was provided.

In order to phase out old uses as fast as possible, along with these new methods, the following methods are deprecated:

- delete_link()
- delete_link_as_target()
- delete_link_as_source()

### DIFF
--- a/crates/provider-sdk/src/lib.rs
+++ b/crates/provider-sdk/src/lib.rs
@@ -195,6 +195,45 @@ impl ProviderConfigUpdate for &HashMap<String, String> {
     }
 }
 
+/// Present information related to a link delete, normally used as part of the [`Provider`] interface,
+/// for providers that must process a link deletion in some way.
+pub trait LinkDeleteInfo: Send + Sync {
+    /// Retrieve the source of the link
+    ///
+    /// If the provider receiving this LinkDeleteInfo is the target, then this is
+    /// the workload that was invoking the provider (most often a component)
+    ///
+    /// If the provider receiving this LinkDeleteInfo is the source, then this is
+    /// the ID of the provider itself.
+    fn get_source_id(&self) -> &str;
+
+    /// Retrieve the target of the link
+    ///
+    /// If the provider receiving this LinkDeleteInfo is the target, then this is the ID of the provider itself.
+    ///
+    /// If the provider receiving this LinkDeleteInfo is the source (ex. a HTTP server provider which
+    /// must invoke other components/providers), then the target in this case is the thing *being invoked*,
+    /// likely a component.
+    fn get_target_id(&self) -> &str;
+
+    /// Retrieve the link name
+    fn get_link_name(&self) -> &str;
+}
+
+impl LinkDeleteInfo for &InterfaceLinkDefinition {
+    fn get_source_id(&self) -> &str {
+        &self.source_id
+    }
+
+    fn get_target_id(&self) -> &str {
+        &self.target
+    }
+
+    fn get_link_name(&self) -> &str {
+        &self.name
+    }
+}
+
 /// Capability Provider handling of messages from host
 pub trait Provider<E = anyhow::Error>: Sync {
     /// Initialize the provider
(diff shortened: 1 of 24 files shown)


crates/provider-sdk/src/provider.rs
```diff
@@ -554,15 +554,16 @@ async fn delete_link_for_provider<P>(
...
             error!(error = %e, target = &ld.target, "failed to delete link to component");
         }
-    } else if ld.target == *connection.provider_id {
-        if let Err(e) = provider.delete_link_as_target(&ld.source_id).await {
+    } else if *ld.target == *connection.provider_id {
+        if let Err(e) = provider.delete_configured_link_as_target(&ld).await {
             error!(error = %e, source = &ld.source_id, "failed to delete link from component");
         }
     }
+    warn!("delete_link() is deprecated, consider using delete_configured_link_*() variants. In a future release delete_link() will be removed.");
```
brooksmtownsend, line 566: Won't we still need to call the `delete_link_as_*` methods in order to not make this a breaking change?

crates/provider-sdk/src/lib.rs
```diff
@@ -229,26 +268,56 @@ pub trait Provider<E = anyhow::Error>: Sync {
...
+    ) -> impl Future<Output = Result<(), E>> + Send {
+        async { Ok(()) }
+    }
+
+    /// Notify the provider that the link is dropped where the provider is the source
+    ///
+    /// Unlike [`delete_link_as_source`], metadata like link name are available to be inspected.
+    fn delete_configured_link_as_source(
+        &self,
+        _info: impl LinkDeleteInfo,
```
brooksmtownsend, line 319: Alternatively to the above, could we simply proxy this function to `self.delete_link_as_source` so that if the developer doesn't implement the function, we just call the "old" method for now?

brooksmtownsend (changes requested):
Just want to make sure I'm not missing anything regarding this change

crates/provider-sdk/src/lib.rs
vados-cosmonic: Ah yes! That's what I should have done in the first one -- we can switch to using the new stuff in internal code if we use the existing calls when not specified

crates/provider-sdk/src/provider.rs
vados-cosmonic: So I think when I made this, I wanted the change to be *breaking* in Provider SDK -- and get it out as fast as possible. Since we're not at 1.x.x, `0.6.0` would have been a breaking change (now it needs to be like `0.8.0` I think

crates/provider-sdk/src/lib.rs
vados-cosmonic: On second thought, I think we should just make this a breaking change -- it doesn't affect the host but actually only affects `provider-sdk` -- what do you think?

crates/provider-sdk/src/provider.rs
vados-cosmonic: The warning should probably just not be there actually, I can remove it

crates/provider-sdk/src/lib.rs
brooksmtownsend: I think we can just make this a breaking change, let's just keep the reasonable names that we had before & we can add release notes with the new version! It's just better

vados-cosmonic: Yup I agree! definitely should just be a breaking change -- I updated all the providers

brooksmtownsend (approved):
LGTM 😄 Consumers of the provider SDK will just need the heads up in the release notes and that's fairly easy to take care of. I may just resolve this conflict to get this PR moving

PR #2347, opened 2024-06-15
feat(wit): add WIT for email provider

vados-cosmonic:
## Feature or Problem

A while back we did a hackathon where I developed this provider for sending external emails... Would be nice to upstream this and get some support in main for this prototype interface!

This commit adds the WIT interface for the wasmcloud:email provider, particularly the outgoing email sending functionality

### DIFF
--- a/wit/email/README.md
+++ b/wit/email/README.md
@@ -1,0 +1,6 @@
+# 📧 Email WIT
+
+This folder contains a [WIT][wit] interface for sending email via [WebAssembly][wasm].
+
+[wasm]: https://webassembly.org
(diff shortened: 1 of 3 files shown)


thomastaylor312:
@vados-cosmonic Love this, but it is definitely making me think we need a standard pattern for creating and sharing things that don't quite feel like they belong in the monorepo. Let's do some brainstorming as a community and figure out what we want to do!

vados-cosmonic:
Hey @thomastaylor312 thanks! Yeah we should probably figure out where interfaces that aren't necessary in the monorepo go!

PR #2349, opened 2024-06-16
fix(providers,examples): update Go provider custom template

ericgregory:
Corrects provider archive name in provided wadm.yaml and adds prebuilt component in `/component/build` per readme. (Also removes that build folder from the gitignore.)


t3hmrman (approved):

vados-cosmonic (approved):
LGTM :rocket

Issue #1927, opened 2024-04-17
[FEATURE] Improve error output for incorrectly configured links.


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

When linking on an incorrect interface, or when an interface link is not present, the errors can be opaque and appear like a stacktrace.

### Describe the solution you'd like

Though we eventually do need to return errors and we might not be able to prevent a panic inside of the component, it would be great to output proper logs to help narrow down issues.

vados-cosmonic:
Another reason that this can happen is that there isn't a link with the right name. It's possible that *a* link is configured, but not one that matches the expected link name.

PR #2362, opened 2024-06-20
chore(*): improve error messages for missing links

vados-cosmonic:
## Feature or Problem

This commit improves error messaging when we fail to call a target for an interface that has no link specified.

Resolves https://github.com/wasmCloud/wasmCloud/issues/1927

### DIFF
--- a/crates/host/src/wasmbus/handler.rs
+++ b/crates/host/src/wasmbus/handler.rs
@@ -8,7 +8,7 @@ use async_trait::async_trait;
 use futures::{stream, Stream};
 use tokio::io::{AsyncRead, AsyncReadExt};
 use tokio::spawn;
-use tokio::sync::RwLock;
+use tokio::sync::{RwLock, TryLockError};
 use tracing::{debug, error, instrument};
 use wasmcloud_core::LatticeTarget;
 use wasmcloud_runtime::capability::config::runtime::ConfigError;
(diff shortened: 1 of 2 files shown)


crates/host/src/wasmbus/handler.rs
```diff
@@ -107,65 +107,113 @@ impl Handler {
...
-                "http",
-                "outgoing-handler",
+                ns, pkg, iface
             )))
             .await
-            .context("unknown `wasi:http/outgoing-handler` target")?;
+            .with_context(|| {
+                let mut msg = format!("failed to call interface `{ns}:{pkg}/{iface}`");
+                if let Ok(false) = self.try_link_exists_for_interface(ns, pkg, iface) {
+                    msg.push_str(&format!(" (failed to find a link from this component [{}], please check your configuration)", self.component_id));
```
brooksmtownsend, line 137:
```suggestion
                    msg.push_str(&format!(" (failed to find a configured link from component [{}], please check your configuration)", self.component_id));
```

consistency with other messages suggestion

crates/host/src/wasmbus/handler.rs
```diff
@@ -545,7 +593,13 @@ impl Bus for Handler {
             tx.await.context("failed to transmit parameters")?;
             Ok(results)
         } else {
-            bail!("component attempted to invoke a function on an unknown target")
+            bail!(
+                "component [{}] attempted to invoke a function [{}/{}] on an unknown target [{}]",
+                self.component_id,
+                instance,
+                name,
+                target.lattice_id().unwrap_or("<unknown>"),
```
brooksmtownsend, line 601: Shouldn't this target be the target "id", rather than the lattice ID?

brooksmtownsend: Or, after reading the implementation of the method, this is not the _lattice name_ this is the invocation target on the lattice, right? If I understand correctly this LGTM

brooksmtownsend (approved):
One request for consistency one ask for my own clarification

crates/host/src/wasmbus/handler.rs
vados-cosmonic: Ah thank you!

crates/host/src/wasmbus/handler.rs
vados-cosmonic: Ah this is not the lattice name -- it's the "reference by which the entity can be addressed on the lattice" becuase that is off the `TargetEntity`. 

This is confusing though, so I'll change this!

Issue #2183, opened 2024-05-28
[BUG] label order in `wash get inventory` is not sorted


protochron:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

The the ordering of host labels when running `wash get inventory` is unpredictable, and seemingly random when run multiple times. I'm guessing we're using a `HashMap<String,String>` to store them and we should probably use a `BTreeMap` instead.

### Steps to reproduce

* add a number of labels to a host and start it
* run `wash get inventory` multiple times and observe that the order on the labels is not sorted

### Expected behavior

The label ordering should be consistent every time I run the command

### Environment

- OS: OSX
- Shell: zsh
- wash 0.28.0

PR #2363, opened 2024-06-20
chore(*): order labels in wash get inventory

vados-cosmonic:
## Feature or Problem

Resolves https://github.com/wasmCloud/wasmCloud/issues/2183

### DIFF
--- a/crates/control-interface/src/types/host.rs
+++ b/crates/control-interface/src/types/host.rs
@@ -1,8 +1,8 @@
 //! Data types used for managing hosts on a wasmCloud lattice
 
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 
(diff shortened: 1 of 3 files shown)


crates/control-interface/src/types/host.rs
```diff
@@ -59,6 +59,7 @@ pub struct HostInventory {
     pub friendly_name: String,
     /// The host's labels
     #[serde(default)]
+    #[serde(serialize_with = "serialize_as_btreemap")]
```
brooksmtownsend, line 62: Dumb question, do we need to have a deserialize_with here too? Looks like the test works at least 😄

brooksmtownsend (approved):

crates/control-interface/src/types/host.rs
vados-cosmonic: So to avoid a breaking change here I did not actually change the type -- it doesn't make sens to try and re-serialize as the `BTreeMap` because the `HashMap` it's turning into is unordered, so we can't count on that anyway.

PR #2367, opened 2024-06-20
fix(tests): test component builds

vados-cosmonic:
## Feature or Problem

After switching to Rust 1.79 we had a problem building test components used in the repository -- this PR fixes that by updating the `build.rs` script we were using, and refactoring it a bit.


tests/components/build.rs
```diff
@@ -3,20 +3,70 @@ use std::env;
...
+}
+
+/// Convert a hard-coded artifact name (see [`WASI_WASM32_PACKAGES`])
+///
+/// Post-Rust 1.79, kebab cased project names turn into snake_case cased artifacts
+#[rustversion::since(1.79)]
+fn generate_expected_artifact_name(project_name: &str) -> impl AsRef<str> {
+    use heck::ToSnakeCase;
+    project_name.to_snake_case()
+}
```
brooksmtownsend, line 63: Oh boy, any changes that we need to address on the `wash build` side here?

tests/components/build.rs
```diff
@@ -97,129 +168,68 @@ async fn build_artifacts(
...
+    for (manifest_path, package) in WASI_WASM32_PACKAGES {
+        artifacts.spawn(build_artifacts(
+            [
+                format!(
+                    "--manifest-path={}/{manifest_path}",
+                    env!("CARGO_MANIFEST_DIR")
+                ),
+                "--target=wasm32-wasi".to_string(),
+                // We use dashes for packages, since they're still always
+                // format!("-p={}", package.replace("_", "-")),
```
brooksmtownsend, line 194: Extra comment?

tests/components/build.rs
```diff
@@ -247,22 +257,15 @@ async fn main() -> anyhow::Result<()> {
...
-        "builtins-component-reactor",
-        "example-http-keyvalue-counter",
-        "interfaces-handler-reactor",
-        "interfaces-reactor",
-        "pinger-config-component",
-        "ponger-config-component",
-    ] {
-        let path = out_dir.join(format!("rust-{name}.wasm"));
+    for (_manifest_path, package_name) in WASI_WASM32_PACKAGES {
+        // let path = out_dir.join(format!("rust-{package_name}.wasm"));
```
brooksmtownsend, line 261: extra comment?

brooksmtownsend (changes requested):

tests/components/build.rs
vados-cosmonic: Nope I don't think so! `wash build` works just fine for me -- we don't really mess with the artifact names this directly/specifically.

Plus we also have this tested via CI so we'll know when this breaks

tests/components/build.rs
vados-cosmonic: Yep! good catch, will remove

tests/components/build.rs
vados-cosmonic: Yup, will remove!

thomastaylor312 (approved):

vados-cosmonic:
Hey @brooksmtownsend dismissing the review here so we can get these changes merged -- I think the only big thing was the possible implications on `wash` due to changed output of `cargo` but considering I'm running `1.79.0` (and many people probably are), I think it's probably fine -- if there's an issue we can definitely open a follow up

PR #2369, opened 2024-06-20
chore(examples): Remove experimental: true annotation from example

joonas:
## Feature or Problem

This is no longer needed, so let's get rid of it.


vados-cosmonic (approved):
LGTM :rocket:

PR #2370, opened 2024-06-21
Wrpc update and preparation for resource support

LUK3ARK:
## Feature or Problem
- Updated to the latest version of NATS.
- Updated to the latest version of Wasmtime.
- Updated to the latest version of wRPC.
- Updated outgoing HTTP function to accept request options parameter.
- Updated bus call function to work with the new wRPC multiplexed streams and sessions error handling.
- Added get_func_paths to the bus to share path indexing from host to runtime for wRPC.
- Created an internal IncomingInputStream type for compatibility as wRPC no longer exports it.
- Added trappable imports setting, which is not default in the new version.
- Added paths variable to components to avoid carrying the resolve on objects intended to be cheaply cloneable.
- Updated the method of managing functions among instances.
- Updated to use wRPC transport NATS client. A client builder was made but it may be better to revert to a thinner wrapper or remove it entirely.

## Related Issues
1. Added TODO note to double-check wasi mapping -  I think these started raising an error now that unmapped bindings generate errors...

2. Temporarily commented out some trait implementations for capability providers to allow for compilation. These need to be implemented back in before the end of this pull request to support first-hand interfaces in wasmCloud @ricochet . Need to implement with newest WRPC API, so these will likely change in form anyway.
    To resolve this, we need to determine the best way to map different invocation types. Options include:
    - Finding a way to fulfill these requirements now that wRPC does not generate the functions for serving directly.
    - Continuing to use wrpc_interface_http, wrpc_interface_blobstore, etc., crate dependencies in parallel.

3. Update provider sdk after wrpc implementation has been signed off

4. Implement resources on top of new wrpc integration, which can move forward once @rvolosatovs is able to finalize resource support natively in wrpc.

## Consumer Impact
Make sure we have support for wrpc and new bindings for providers or port to new style. This may have some breaking changes as @brooksmtownsend 

Other improvements
- Reassess the client builder for wRPC transport NATS client to determine if a thinner wrapper or removal is better


Cargo.toml
```diff
@@ -175,7 +174,7 @@ members = ["crates/*"]
 anyhow = { version = "1", default-features = false }
 assert-json-diff = { version = "2", default-features = false }
 async-compression = { version = "0.3", default-features = false }
-async-nats = { version = "0.33", default-features = false }
+async-nats = { version = "0.35", git = "https://github.com/rvolosatovs/nats.rs", branch = "feat/command-sender", default-features = false }
```
rvolosatovs, line 177: Refs https://github.com/nats-io/nats.rs/pull/1267

Looks like there's still no response on the PR, so I'll go ahead and publish `async-nats-wrpc` crate with this patch applied

Cargo.toml
```diff
@@ -326,23 +325,24 @@ wasmcloud-secrets-types = { version = "*", path = "./crates/secrets-types", defa
...
-wit-parser = { version = "0.202", default-features = false }
+wit-component = { version = "0.208", default-features = false }
+wit-parser = { version = "0.208", default-features = false }
 wrpc-interface-blobstore = { version = "0.16", default-features = false }
 wrpc-interface-http = { version = "0.22", default-features = false }
-wrpc-runtime-wasmtime = { version = "0.16", default-features = false }
-wrpc-transport = { version = "0.24.3", default-features = false }
-wrpc-transport-nats = { version = "0.21", default-features = false }
-wrpc-types = { version = "0.6", default-features = false }
+wrpc-runtime-wasmtime = { git = "https://github.com/rvolosatovs/wrpc.git", branch = "fix/transport-traits", default-features = false }
```
rvolosatovs, line 344: looks like https://github.com/bytecodealliance/wrpc/pull/143 is good to go?

crates/host/src/wasmbus/handler.rs
```diff
@@ -1,27 +1,30 @@
...
-
-use crate::bindings::{wasmcloud, wrpc};
+use wasmtime::component::{Type, Val};
+use wasmtime_wasi_http::body::{HyperIncomingBody, HyperOutgoingBody};
+use wasmtime_wasi_http::types::OutgoingRequestConfig;
+use wit_parser::Function;
+use wrpc_runtime_wasmtime::read_value;
+use wrpc_transport::Index;
+use wrpc_transport::Session;
+use wrpc_transport::{Invocation, Invoke};
```
rvolosatovs, line 26: nit; could we get these merged into a single line?

crates/host/src/wasmbus/handler.rs
```diff
@@ -60,9 +63,15 @@ pub struct Handler {
...
-    pub polyfills: Arc<HashMap<String, HashMap<String, wrpc_types::DynamicFunction>>>,
+    pub polyfills: Arc<HashMap<String, HashMap<String, Function>>>,
+
+    pub exports: Arc<HashMap<String, HashMap<String, Function>>>,
 
+    /// Reference to store for collection of instances
+    // pub store:,
     pub invocation_timeout: Duration,
+
+    pub func_paths_map: Arc<HashMap<(Arc<String>, Arc<String>), Arc<Vec<Arc<[Option<usize>]>>>>>,
```
rvolosatovs, line 74:
```suggestion
    pub func_paths_map: Arc<HashMap<(Arc<str>, Arc<str>), Arc<[Arc<[Option<usize>]>]>>>,
```

crates/host/src/wasmbus/handler.rs
```diff
@@ -128,33 +127,32 @@ impl Handler {
...
             .identify_wrpc_target(&CallTargetInterface::from_parts((
                 "wasi", "keyvalue", "store",
             )))
             .await
             .context("unknown `wasi:keyvalue/store` target")?;
-        Ok(self.wrpc_client(&lit))
+        Ok(self.wrpc_builder(&lit))
     }
 
-    #[instrument(level = "trace", skip(self))]
```
rvolosatovs, line 156: is there a reason to remove this?

crates/host/src/wasmbus/mod.rs
```diff
@@ -327,108 +313,99 @@ impl Component {
...
-                attributes.push(KeyValue::new(
-                    "operation",
-                    "wasmcloud:messaging/handler.handle-message",
-                ));
-                self.metrics
-                    .record_component_invocation(elapsed, &attributes, res.is_err());
-                let res = res?;
-                transmitter.transmit_static(result_subject, res).await
-            }
+            // TODO - How are these specific cases going to be served
```
rvolosatovs, line 336: I don't think we should need any changes in the way the routing is set up here

crates/host/src/wasmbus/handler.rs
```diff
@@ -163,293 +161,356 @@ impl Handler {
...
-            )
-            .await
-            .context("failed to invoke `wrpc:blobstore/blobstore.move-object`")?;
-        // TODO: return a result directly
-        res.map_err(|err| anyhow!(err).context("function failed"))?;
-        tx.await.context("failed to transmit parameters")?;
-        Ok(())
-    }
-}
+// #[async_trait]
```
rvolosatovs, line 168: You'll need to adapt these trait implementations using the new transport API.
We can either do it here directly, most probably using `invoke_values` https://bytecodealliance.github.io/wrpc/wrpc/transport/trait.Invoke.html#method.invoke_values (currently unstable, and will change soon, likely today)

or work on updating https://github.com/wrpc/blobstore and https://github.com/wrpc/http and integrating here

crates/runtime/src/actor/component/mod.rs
```diff
@@ -262,15 +274,18 @@ pub struct Component {
...
     handler: builtin::HandlerBuilder,
-    polyfills: Arc<HashMap<String, HashMap<String, DynamicFunction>>>,
-    exports: Arc<HashMap<String, HashMap<String, DynamicFunction>>>,
+    polyfills: Arc<HashMap<String, HashMap<String, Function>>>,
+    exports: Arc<HashMap<String, HashMap<String, Function>>>,
     ty: types::Component,
     instance_pre: wasmtime::component::InstancePre<Ctx>,
     max_execution_time: Duration,
+    // A map of function names to their respective paths for params
+    paths: Arc<HashMap<String, Vec<Vec<Option<usize>>>>>,
```
rvolosatovs, line 283:
```suggestion
    paths: Arc<HashMap<String, Arc<[Arc<[Option<usize>]>]>>>,
```

crates/runtime/src/actor/component/mod.rs
```diff
@@ -377,65 +392,152 @@ where
...
+                let func_paths = match handler.get_func_paths(&instance_name, &func_name) {
+                    Some(func_paths) => func_paths,
+                    None => {
+                        error!(
+                            "Function paths not found for instance: {} and name: {}",
+                            instance_name, func_name
+                        );
+                        return Box::new(async { Ok::<(), anyhow::Error>(()) });
+                    }
+                };
```
rvolosatovs, line 441: Probably want to do this outside `func_new_async`.

crates/runtime/src/actor/component/mod.rs
```diff
@@ -377,65 +392,152 @@ where
...
+                        instance_name, func_name
+                    );
+                    continue;
+                }
+            };
+
+            let result_types_vec: Vec<_> = result_types.collect();
+            let result_types_arc = Arc::new(result_types_vec);
+
+            if let Err(err) = linker.func_new_async(rpc_name, move |mut store, params, results| {
```
rvolosatovs, line 424: I think we should be using https://bytecodealliance.github.io/wrpc/wrpc_runtime_wasmtime/fn.polyfill.html here
or, at least, `polyfill_function`, which I've extracted in https://github.com/bytecodealliance/wasmtime/pull/8737.
I'll work on pulling in changes from that PR back to `wrpc-runtime-wasmtime`

crates/runtime/src/actor/component/mod.rs
```diff
@@ -730,18 +858,20 @@ impl Instance {
...
-    pub async fn call(
+    #[instrument(skip(self, instance, name, incoming, outgoing), fields(interface = instance, function = name))]
+    pub async fn call<C>(
         &mut self,
         instance: &str,
         name: &str,
-        params: Vec<wrpc_transport::Value>,
-    ) -> anyhow::Result<Vec<wrpc_transport::Value>> {
+        incoming: Reader,
+        outgoing: SubjectWriter,
```
rvolosatovs, line 867:
```suggestion
        incoming: C::Incoming,
        outgoing: C::Outgoing,
```

And you'll need to bind `C: wrpc_transport::Invoke`

crates/runtime/src/runtime.rs
```diff
@@ -103,30 +91,6 @@ impl RuntimeBuilder {
         }
     }
 
-    /// Set a [`KeyValueAtomics`] handler to use for all component instances unless overriden for the instance
```
rvolosatovs, line 106: Why are these removed?

crates/runtime/src/actor/component/http.rs
```diff
@@ -37,30 +38,27 @@ impl WasiHttpView for Ctx {
...
-        let res = HostFutureIncomingResponse::new(wasmtime_wasi::runtime::spawn(async move {
-            match OutgoingHttp::handle(&handler, request).await {
-                Ok(Ok(resp)) => Ok(Ok(IncomingResponseInternal {
+        let between_bytes_timeout = config.between_bytes_timeout.clone();
+        let handle = wasmtime_wasi::runtime::spawn(async move {
+            match OutgoingHttp::handle(&handler, request, config).await {
+                Ok(Ok(resp)) => Ok(Ok(IncomingResponse {
                     resp,
-                    worker: Arc::new(wasmtime_wasi::runtime::spawn(async {})),
+                    worker: Some(wasmtime_wasi::runtime::spawn(async {})),
```
rvolosatovs, line 53: I haven't checked the docs, but looks like this could just be `None`?

crates/runtime/src/runtime.rs
LUK3ARK: I commented out while getting to compile - I just need to think about the best way to architect this with parameters being passed as bytes... as encoding and decoding depend on store context, I need to find a way to somehow ove the context through to the handler, to then generate as these, all pass the bytes to these handlers.

I added some comments here:
Temporarily commented out some trait implementations for capability providers to allow for compilation. These need to be implemented back in before the end of this pull request to support first-hand interfaces in wasmCloud @ricochet . Need to implement with newest WRPC API, so these will likely change in form anyway.
To resolve this, we need to determine the best way to map different invocation types. Options include:

Finding a way to fulfill these requirements now that wRPC does not generate the functions for serving directly.
Continuing to use wrpc_interface_http, wrpc_interface_blobstore, etc., crate dependencies in parallel.

We can carry on directly supporting the crates from before I believe?

rvolosatovs: I don't think we should need to change anything, the "bytes" approach should only be used for the dynamic use cases (e.g. `call`), for static bindings, `wasmcloud_host` should decode and encode typed values, like it does now

crates/host/src/wasmbus/handler.rs
rvolosatovs: FYI, almost ready, will have to finish next week https://github.com/bytecodealliance/wrpc/pull/147
(note the `invoke_values` and `serve_values` APIs)

crates/runtime/src/actor/component/mod.rs
rvolosatovs: See https://github.com/bytecodealliance/wrpc/pull/149

LUK3ARK:
Second go round, big thanks to @rvolosatovs for the help:

- There is so much friction moving over, therefore I took a different approach and kept usage of both the previous wrpc client (now referred to as LegacyClient) and the new WrpcClient. The provider sdk still relies on the legacy client, and any messages that need to be directly handled for http and wasmcloud messaging custom use cases will be handled by the legacy client, ensuring the unresourced api remains used.

-I have removed the call function from the bus, as this was only for controlling middleware routing, now this is completely encapsulated by wrpc. All custom components will fallback to using the new wrpc client which allows the usage of resources and streaming more effectively. The only invocations that should be delegated to the runtime are the unresourced definitions for blobstore, keyvalue, etc. Everything else is binded by wrpc directly.

- We might need to revisit the provider-sdk for any resources exported by providers.

- to iterate... any instances that need to be explicitly served by the legacy client for any component will do so

- I had to move around some of the logic and i don't like the fact that the instantiate function now required a lattice string. it works though, config structs have to be remade for these processes anyway

- Because we now store a lattice property on the handler for each component, can someone confirm the expected behaviour for 'copy_for_new' ? we can even change lattice to an option if need be :)

rvolosatovs:
I believe this is now addressed given that #2459 is merged?

brooksmtownsend:
Closing because of #2459 , but @LUK3ARK please re-open if there's a piece that hasn't been addressed!

PR #2372, opened 2024-06-21
docs(examples): add comments and annotations to WADM

vados-cosmonic:
## Feature or Problem

This commit adds comments explaining linking and custom annotations to WADM manifests in examples.


brooksmtownsend (approved):
I'm a fan 😄 We use `wasmcloud.dev` for some of our other annotations, thoughts on either being the preferred prefix?

vados-cosmonic:
Hmnnn @brooksmtownsend I think this came up when @lachieh and I talked about it -- `wasmcloud.com` are both equally fine to me -- we control both domains.. if we're using `wasmcloud.dev` in other places I'll switch to it here!

Will go ahead and add this to all the example WADMs

PR #2373, opened 2024-06-21
feat(*): add utility method for accessing link name

### LINKED ISSUE #2371: [FEATURE] Utility function for getting link_name from Context for providers
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [X] capability providers
- [ ] provider bindgen
- [ ] other / not sure

The `provider-sdk` should provide an easier way to get at the link name on a given `Context`.

Currently we require people to [dig around themselves](https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-keyvalue-redis/src/lib.rs#L118), but the DX could be better.

A utility method on `Context` that returns the link name.

In general the link name should always be present, so it not being in the OTEL headers is likely a bug/issue, but we probably don't want to panic there, so `Option` might be a good idea for now?

N/A

### Additional context

_No response_

vados-cosmonic:
## Feature or Problem

This PR:
- Adds `Context::link_name()` utility method to easily retrieve link names
- Uses the new method from the KV redis provider

Resolves #2371

### DIFF
--- a/crates/provider-keyvalue-redis/src/lib.rs
+++ b/crates/provider-keyvalue-redis/src/lib.rs
@@ -104,23 +104,17 @@ impl KvRedisProvider {
 
     #[instrument(level = "debug", skip(self))]
     async fn invocation_conn(&self, context: Option<Context>) -> anyhow::Result<ConnectionManager> {
-        let Some(Context {
-            component: Some(source_id),
-            tracing,
-        }) = context
-        else {
+        let ctx = context.context("unexpectedly missing context")?;
+
+        let Some(ref source_id) = ctx.component else {
             return self.get_default_connection().await.map_err(|err| {
                 error!(error = ?err, "failed to get default connection for invocation");
                 err
             });
         };
 
-        let Some(link_name) = tracing.get("link-name") else {
-            bail!("unexpectedly missing link name on context for invocation");
-        };
-
         let sources = self.sources.read().await;
-        let Some(conn) = sources.get(&(source_id.to_string(), link_name.to_string())) else {
+        let Some(conn) = sources.get(&(source_id.into(), ctx.link_name().into())) else {
             error!(source_id, "no Redis connection found for component");
             bail!("No Redis connection found for component [{source_id}]. Please ensure the URL supplied in the link definition is a valid Redis URL")
         };
--- a/crates/provider-sdk/src/lib.rs
+++ b/crates/provider-sdk/src/lib.rs
@@ -104,6 +104,24 @@ pub struct Context {
     pub tracing: HashMap<String, String>,
 }
 
+impl Context {
+    /// Get link name from the request.
+    ///
+    /// While link name should in theory *always* be present, it is not natively included in [`Context`] yet,
+    /// so we must retrieve it from headers on the request.
+    ///
+    /// Note that in certain (older) versions of wasmCloud it is possible for the link name to be missing
+    /// though incredibly unlikely (basically, due to a bug). In the event that the link name was *not*
+    /// properly stored on the context 'default' (the default link name) is returned as the link name.
+    #[must_use]
+    pub fn link_name(&self) -> &str {
+        self.tracing
+            .get("link-name")
+            .map(String::as_str)
+            .unwrap_or("default")
+    }
+}
+
 /// Configuration of a link that is passed to a provider
 #[non_exhaustive]
 pub struct LinkConfig<'a> {


crates/provider-sdk/src/lib.rs
```diff
@@ -104,6 +104,17 @@ pub struct Context {
     pub tracing: HashMap<String, String>,
 }
 
+impl Context {
+    /// Get link name from the request.
+    ///
+    /// While link name should in theory *always* be present, it is not natively included in [`Context`] yet,
```
thomastaylor312, line 110: Should we have it be `default` if it returns `None` or is that not correct behavior?

crates/provider-sdk/src/lib.rs
```diff
@@ -104,6 +104,17 @@ pub struct Context {
...
+    /// so we must retrieve it from headers on the request.
+    #[must_use]
+    pub fn link_name(&self) -> Option<&String> {
```
thomastaylor312, line 113: I think it is more idiomatic to return an `&str` instead

crates/provider-sdk/src/lib.rs
vados-cosmonic: Here the failure to get the `link-name` header does not mean that there isn't a link name (or that it's `default`), which is why I left it at `None`.

It's unfortunate because we *know* that a link name should be present, but with how it's implemented right now, we can't *ensure* that the value is present, thus we have to look for it (and possibly not find it, no matter what the actual value is).

crates/provider-sdk/src/lib.rs
vados-cosmonic: Reasonable! Will change this

- [ ] Return `&str` for `Context::link_name`

crates/provider-sdk/src/lib.rs
brooksmtownsend: I think the big reason to have it return `default`, potentially with a debug level log when this occurs, is so that newer provider versions can run on 1.0 hosts and still work 🤔 

The other downside there I think is that for all hosts post 1.1 that header should always be there, so it feels like an extra case to handle when it's unlikely.

Bottom line I can't really think of what we would do in our providers other than `link_name.unwrap_or("default")` so I figure we can just go ahead and set that, but if you have a case in mind where we wouldn't handle it that way I think we can also just choose to treat None as default

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -104,18 +104,16 @@ impl KvRedisProvider {
 
     #[instrument(level = "debug", skip(self))]
     async fn invocation_conn(&self, context: Option<Context>) -> anyhow::Result<ConnectionManager> {
-        let Some(Context {
-            component: Some(source_id),
-            tracing,
-        }) = context
-        else {
+        let ctx = context.context("unexpectedly missing context")?;
+
```
brooksmtownsend, line 108: When would this happen, could we add any extra information here?

crates/provider-sdk/src/lib.rs
vados-cosmonic: > I think the big reason to have it return default, potentially with a debug level log when this occurs, is so that newer provider versions can run on 1.0 hosts and still work 🤔

I'm not sure I understand -- newer providers would be using a version of the SDK that makes it part of `Context` right -- so it *should* be impossible for them to run into this.

If in the future we properly make it required, then this function can stay and will *always* return `Some(..)`

> The other downside there I think is that for all hosts post 1.1 that header should always be there, so it feels like an extra case to handle when it's unlikely.

Yeah definitely agree on this, but until that header is actually properly part of the contract, it feels like we shouldn't promise it *is* there, or worse give a wrong value... for it to be guaranteed to be there, `Context` has to change, and that requires breaking changes/future versions depending on how hard we semver.

> Bottom line I can't really think of what we would do in our providers other than link_name.unwrap_or("default") so I figure we can just go ahead and set that, but if you have a case in mind where we wouldn't handle it that way I think we can also just choose to treat None as default in our impls

IMO there's a difference between the SDK making the wrong choice for you and leaving you to do it -- you generally don't want the SDK to do the hopeful but possibly wrong thing. 

That said, I'm not against adding this case to the documentation and doing an `unwrap_or`!

crates/provider-keyvalue-redis/src/lib.rs
vados-cosmonic: Well this is just a result of the contract as it's written -- the custom interface libraries define the fucntions with an optional context:

```rust
impl keyvalue::store::Handler<Option<Context>> for KvRedisProvider {
    #[instrument(level = "debug", skip(self))]
    async fn delete(
        &self,
        context: Option<Context>,
        bucket: String,
        key: String,
    ) -> anyhow::Result<Result<()>> {
        propagate_trace_for_ctx!(context);
        check_bucket_name(&bucket);
        Ok(self.exec_cmd(context, &mut Cmd::del(key)).await)
    }
```

So this bleeds through to `exec_cmd` and eventually `invocation_conn`... There's no good reason Context would be missing (and it shouldn't ever be), so there's not a ton more to say other than that...

crates/provider-sdk/src/lib.rs
vados-cosmonic: Added this at the SDK level!

thomastaylor312 (approved):

Issue #2354, opened 2024-06-17
[BUG] Washboard UI retryCount cannot be changed 


markkovari:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [X] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

When changing the retryCount on the washboard UI settings, the value change results to error of the form submission. There is an error message, but disappears pretty fast and not blocking the drawer being closed.

### Steps to reproduce

1. issue `wash ui` command
2. go to `localhost:3030`
3. click on the settings icon/cog
4. change the retryCount value
5. click on the update

### Expected behavior

The settings' retry count should be changed with the new value.

### Environment

- OS: MacOs Sonoma 14.5 (23F79)
- wash 0.29.1
- mozilla firefox 127.0 
- safari Version 17.5 (19618.2.12.11.6)
- arc Version 1.47.1 (50893)

### Screenshots / Logs / Additional context

<img width="568" alt="Screenshot 2024-06-17 at 16 42 36" src="https://github.com/wasmCloud/wasmCloud/assets/33259514/040a95da-66f0-42b6-99b0-0ef6fa4705a1">

https://github.com/wasmCloud/wasmCloud/assets/33259514/f228a4bf-606b-4f3f-a39b-d635cfa86889

I pretty much gave up after 30 minutes of react debugging, but my best guess is that the Input component is not happy with the type of the value that it receives.
<img width="884" alt="Screenshot 2024-06-17 at 18 05 15" src="https://github.com/wasmCloud/wasmCloud/assets/33259514/9326dfba-796e-48ad-b327-49d982d364ad">

Side-note: probably the 10 is a pretty good number (I changed to 29 to test it with the default), I did not even wanted to change it.

On the other hand, the sheet (or drawer) should not be closed if error in my opinion.

markkovari:
For some reason referencing this is a pr did not close this. 🤔 
Great work @emattiza 🙏

PR #2375, opened 2024-06-23
fix(wash-ui): type mismatch number input and schema

emattiza:
## Feature or Problem

For events dispatched by an HTMLInputElement, regardless of the type "number" attribute used, the target.value type is string. The html5 input element though will only ever send strings that are trivially parsable as Number(event.target.value). This uses a zod coercion to preprocess the data by passing it through a Number constructor and filling the form field data on change properly.
## Related Issues

closes [BUG] Washboard UI retryCount cannot be changed  #2354
### Manual Verification

ran the washboard ui locally and was able to update the retryCount and have form success before drawer close

### DIFF
--- a/washboard-ui/packages/washboard-ui/src/app/components/lattice-settings.tsx
+++ b/washboard-ui/packages/washboard-ui/src/app/components/lattice-settings.tsx
@@ -22,7 +22,7 @@ const formSchema = z.object({
     ),
   latticeId: z.string(),
   ctlTopicPrefix: z.string(),
-  retryCount: z.number().min(0),
+  retryCount: z.number().or(z.string()).pipe(z.coerce.number().min(0)),
 });
 
 type LatticeFormInput = z.input<typeof formSchema>;


vados-cosmonic (approved):
LGTM :rocket: 

`<input type="number">` strikes again.

vados-cosmonic:
Oh also, you might want to check this out:

```
z.number().or(z.string()).pipe(z.coerce.number())
```

Right out of the zod docs -- it's a little wordy, but a tiny bit more robust

emattiza:
> Oh also, you might want to check this out:
> 
> ```
> z.number().or(z.string()).pipe(z.coerce.number())
> ```
> 
> Right out of the [zod docs](https://github.com/colinhacks/zod?tab=readme-ov-file#pipe) -- it's a little wordy, but a tiny bit more robust

i put this in place, but have a type error when trying to enforce the min(0) constraint on top. might need to fiddle a bit more to make that work.

FWIW, i think it will only ever produce `"[-]?\d+"` as a string output, so the Number coercion should be acceptable. the number to number coercion is just an identity function really (for the case of default values coming from a configured lattice).

emattiza:
got it with `z.number().or(z.string()).pipe(z.coerce.number().min(0))`, which should constrain it to the acceptable input types possible

vados-cosmonic:
Awesome, thank you, I didn't realize it would take the extra fiddling -- this looks great @emattiza

Issue #2368, opened 2024-06-20
[BUG] Logging not being recorded for Providers


vados-cosmonic:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

Thanks to [Christoph on the wasmCloud Slack](https://wasmcloud.slack.com/archives/CS38R7N9Y/p1718608586847639) for noticing this -- it looks like at some point our provider binaries stopped being able to properly log to the log stream of the host.

After doing some investigating, it seems that this might has to do with our use of [`tracing::subscriber::set_global_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.set_global_default.html) -- we're calling this from inside a library crate (`provider-sdk`), but technically the binary that should be calling it is the host itself.

### Steps to reproduce

1. Start wasmCloud (`wash up`)
2. Start a provider (`wash start provider <...>` or with WADM via `wash app deploy path/to/some/wadm.yaml`)
3. Observe that provider logs don't show up, no matter the level (by default `warn` and `error` should show at least)

### Expected behavior

Provider logs should be showing up in host logs.

### Environment

- OS:  Linux
- Shell zsh
- wasmCLoud Version 1.0.2

PR #2378, opened 2024-06-24
fix(providers): broken logging

vados-cosmonic:
## Feature or Problem

This commit fixes the issue with logging in providers which was caused by our use of `set_global_default`[0] from within a library (in this case `wasmcloud-tracing`).

`set_global_default` should not be called from libraries, so the `configure_observability()` function needed to be modified to ensure that configured subscribers were passed *back out* (as `tracing::Dispatch`s), and usable from the calling code in a binary context.

[0]: https://docs.rs/tracing-core/0.1.32/tracing_core/dispatcher/fn.set_global_default.html

Resolves https://github.com/wasmCloud/wasmCloud/issues/2368


vados-cosmonic:
NOTE: PR is failing tests due to #2379 -- once that merges and this PR is rebased we should be good to go -- please feel free to review the code that is here (not all providers are updated anyway, but the changes will look identical).

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -56,10 +58,38 @@ pub async fn run() -> anyhow::Result<()> {
...
+    pub fn name() -> &'static str {
+        "keyvalue-redis-provider"
+    }
+
     pub async fn run() -> anyhow::Result<()> {
-        let HostData { config, .. } = load_host_data().context("failed to load host data")?;
+        let HostData {
+            config,
+            otel_config,
+            structured_logging,
```
brooksmtownsend, line 69: Do these changes need to be made on each provider that wants to initialize tracing/logging?

brooksmtownsend (commented):
The change to the SDK looks good, just wondering if we need to modify _all_ providers or if we could make this a simpler change in the provider SDK (or if this was just specific to the kvredis provider)

crates/provider-keyvalue-redis/src/lib.rs
vados-cosmonic: Yup, they do -- See this bit in the PR/commit:

> set_global_default should not be called from libraries, so the configure_observability() function needed to be modified to ensure that configured subscribers were passed back out (as tracing::Dispatchs), and usable from the calling code in a binary context.

Everything seems to be pointing to the fact that we should have never been calling `set_global_default` from inside the SDK/lib side to begin with -- they were interfering with each other.

All the other providers need to do this as well

vados-cosmonic: As I [noted here](https://github.com/wasmCloud/wasmCloud/pull/2378#issuecomment-2186777538):

> NOTE: PR is failing tests due to https://github.com/wasmCloud/wasmCloud/pull/2379 -- once that merges and this PR is rebased we should be good to go -- please feel free to review the code that is here (not all providers are updated anyway, but the changes will look identical).

If the direction looks good, I can make this change to all the providers fairly quickly!

thomastaylor312: Can we have a helper function in the sdk that handles init of logging?

vados-cosmonic: Ah so we can't have it be a function but we *could* have a macro (that must get called from binary code somewhere). The problem here is that we can't... call the function that actually completes the setup from the SDK (as it's gonna be in multiple places/that's the problem)!

Maybe something like `wasmcloud_provider_sdk::init_tracing!("name-of-your-provider")` ?

vados-cosmonic: basically, the `dispatch.try_init()` needs to *not* be called from library code. Everything else is fair game -- but the provider needs to hold on to the guard anyway as well for the duration of when it's running.

brooksmtownsend: Oh yeah I'd be cool with a macro 👍🏻 Just something that simplifies the couple of extra steps, as long as the functions you have to call are documented that'll work fine

vados-cosmonic: Done! Macro has been added, and all providers updated

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -56,10 +57,20 @@ pub async fn run() -> anyhow::Result<()> {
...
 
 impl KvRedisProvider {
+    pub fn name() -> &'static str {
+        "keyvalue-redis-provider"
+    }
+
     pub async fn run() -> anyhow::Result<()> {
+        initialize_observability!(
+            KvRedisProvider::name(),
+            std::env::var_os("PROVIDER_KV_REDIS_FLAMEGRAPH_PATH")
```
brooksmtownsend, line 67: Do we actually pass this environment variable down to the provider?

brooksmtownsend (approved):
This looks _really_ nice with the macro 😄 I think it makes a lot of sense for our docs to note how to create the `init` function and then call this macro. I left a note for us to consider on the wasmCloud side but it doesn't necessitate blocking broken logging

crates/provider-keyvalue-redis/src/lib.rs
vados-cosmonic: Yeah this is for the case someone is running the provider locally (and trying to debug it), less so for when it's running in the host just normally. 

That said, we should probably have a way to pass more than a few env vars down to the child processes for providers!

vados-cosmonic: Hmnnn maybe it's better to get that value from `HostData` *and* ENV (and use whichever is specified). Having an example of how to do that or a utility function might be quite nice.

Issue #1818, opened 2024-04-05
[FEATURE] Add `update` field to scale component command


thomastaylor312:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the solution you'd like

During our preparation for 1.0, @brooksmtownsend and I were chatting about the ergonomics of `scale` vs `update` in the control API. Essentially, once you start a component, its image reference we store in lattice data can only change if you `update` the reference, even if you scale to 0. This is a safety measure that prevents a user from unintentionally upgrading/downgrading a component when scaling or unintentionally running two different versions on two different hosts. In order to streamline API usage when the actual intent _is to update the image ref_, we think we should add an `update` boolean field to the `ScaleComponentCommand` that would indicate to a host that it should simultaneously perform the update operation along with anything else from the scale command. `wash` can then expose this as a flag for those users who may consume the imperative ctl API

### Additional context

One thing that we also need to capture in our docs is that the ctl API is a "dumb" API that doesn't have any smarts around handling everything that is running. This is intentional so anyone can build their own extensions on top of it, just like how we build wadm. This should be made extra clear in explaining why the API is designed the way it is. We purposefully don't build anything in around coordinating simultaneous updates unless using a higher level tool like wadm.

We're not opposed to adding something like coordinating an update, but we would like to see some actual user production use cases before we build that API

thomastaylor312:
To clarify, this could be added post-1.0 as it is an addition to the API and not a breaking change

ahmedtadde:
hey @thomastaylor312 , looking at  [the code here) {](https://github.com/ahmedtadde/wasmCloud/blob/8f89ac1f09d1e1ab91e7629f6879034b06141241/crates/host/src/wasmbus/mod.rs#L1609-L1610) and then [here](https://github.com/ahmedtadde/wasmCloud/blob/8f89ac1f09d1e1ab91e7629f6879034b06141241/crates/host/src/wasmbus/mod.rs#L1624-L1625)... it appears that one can do both a scale & update simultaneously with some warning message about the component refs not being the same... the scale task still proceeds with the provided `component_ref` input which may not be the same as the existing known ref (ie, `entry.image_reference`).

[...]

PR #2385, opened 2024-06-25
feat(*): add update to scale component command

vados-cosmonic:
## Feature or Problem

Resolves https://github.com/wasmCloud/wasmCloud/issues/1818


brooksmtownsend (approved):
This looks really nice! 

If not too big of a deal, could we augment one of our existing tests to try this? I think it would work well as a part of the wash integration test suite since in wasmCloud integration we use built components, so if that needs to be a followup once this is released that works for me

vados-cosmonic:
Merging this for now, and we'll figure out the control iface changes in a follow up issue!

Issue #2374, opened 2024-06-21
[BUG] Segmentation error when writing logs on link while running a custom go provider


ritesh089:
### Affected project(s)

- [ ] documentation
- [X] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

I get the following error when i try to build the custom capability provider in go:
```
wasmcloud_host::wasmbus: provider started provider_ref="file:///Users/****/git/wasm/wasmcloud/wasmCloud/examples/golang/providers/custom-template/build/custom-template.par.gz" provider_id="custom-template"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x13a35dd]

goroutine 1 [running]:
main.handleNewSourceLink(0xc0000124c8, {{0xc000014d40, 0xf}, {0xc00012c300, 0x2e}, {0xc000014d50, 0x7}, {0xc000014d57, 0x9}, {0xc000014d60, ...}, ...})
	/Users/****/git/wasm/wasmcloud/wasmCloud/examples/golang/providers/custom-template/main.go:85 +0x1d
main.run.func1({{0xc000014d40, 0xf}, {0xc00012c300, 0x2e}, {0xc000014d50, 0x7}, {0xc000014d57, 0x9}, {0xc000014d60, 0x7}, ...})
```

### Steps to reproduce

1. wash up
2. navigate to ['https://github.com/wasmCloud/wasmCloud/tree/main/examples/golang/providers/custom-template/component'](https://github.com/wasmCloud/wasmCloud/tree/main/examples/golang/providers/custom-template/component)

3. wash build 
4. cd .. && wash build
5. wash app deploy wadm.yaml

### Expected behavior

I expect the log line to be written on target link :
```
func handleNewTargetLink(handler *Handler, link provider.InterfaceLinkDefinition) error {
	handler.provider.Logger.Info("Handling new target link", "link", link)
	handler.linkedFrom[link.Target] = link.TargetConfig
	return nil
}
```

### Environment

- OS: MacOS
- Shell :zsh
- wasmcloud Version : 0.29.2

brooksmtownsend:
After this closes we can revert https://github.com/wasmCloud/wasmCloud/pull/2386

[...]

PR #2386, opened 2024-06-25
fix(custom-template-go): print instead of log

brooksmtownsend:
## Feature or Problem
This PR is a temporary workaround for a segfault that occurs when accessing the `provider` struct during the handler initialization functions.

## Related Issues
#2374


vados-cosmonic (approved):
LGTM :rocket:

PR #2388, opened 2024-06-25
fix(custom-template-golang): initialize maps

brooksmtownsend:
## Feature or Problem
This PR ensures the link maps are initialized in the custom template.


vados-cosmonic:
Note @brooksmtownsend it might be a bit more idiomatic to create a `NewHandler()` constructor so people can't get it wrong

PR #2395, opened 2024-06-26
Add wasmCloud secrets example

brooksmtownsend:

## Feature or Problem
This PR adds an example of a modified keyvalue-counter component that requires authentication and a modified keyvalue Redis provider that connects to a Redis database with a password requirement. Both the component and the provider use the wasmCloud secrets implementation in order to fetch these passwords securely at runtime without transmitting or storing any secret in plaintext.

See the run.sh script for how this is all set up, starting with setting up infrastructure and starting the NATS KV secrets backend to deploying the application.

~Marked as WIP waiting on #2344 to be complete, once finished I will be able to simplify a few steps in the scripts~

## Related Issues
#2411 notes a few improvements to be made to this example


examples/security/secrets/README.md
ossfellow, line 1: There's a tiny miss, where `in` is missing from `builtin`.

examples/security/secrets/README.md
```diff
@@ -0,0 +1,62 @@
...
+- [redis-server](https://redis.io/docs/latest/operate/oss_and_stack/install/install-redis/)
+- [nats-server](https://docs.nats.io/running-a-nats-service/introduction)
+- [jq](https://jqlang.github.io/jq/download/)
+
+## Running this example
+
+Build wasmCloud, the NATS KV secrets backend, the keyvalue counter auth component, and the keyvalue redis password provider:
+
+```bash
+./build.sh
```
vados-cosmonic, line 25: What do you think about doing this without `bash`? If we were going to do bash scripts I'd prefer `Makefile`s to start with... and if we were going to do `Makefile`s, I'd prefer `Justfile`s...

Since no extra tooling would be best, this might be a great place to actually introduce https://github.com/matklad/cargo-xtask

examples/security/secrets/component-keyvalue-counter-auth/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "component-keyvalue-counter-auth"
+edition = "2021"
+version = "0.1.0"
```
vados-cosmonic, line 4: Could we get a description?

examples/security/secrets/component-keyvalue-counter-auth/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            true
+        }
+        _ => false,
+    }
+}
+
+fn get_header(request: &IncomingRequest, header: String) -> Option<String> {
+    let raw_header = request.headers().get(&header);
+    if raw_header.first().map(|val| val.is_empty()).unwrap_or(true) {
```
vados-cosmonic, line 83: Can we just do this all with `Option` chaining? Would something like the below work?

```rust
request
    .headers()
    .get(&header)
    .first()
    .filter(|v| !val.is_empty())
    .map(std::str::from_utf8)
    .map(String::to_string)
}
```

Also it might be nice to name it something like `get_first_header`, and add a comment to note what it does 

```rust
/// Gets first non-empty value for HTTP header named `header` (case-sensitive)
```

examples/security/secrets/provider-keyvalue-redis-password/src/provider.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// RPC topics via `wrpc::serve`.
+    ///
+    /// This step is essentially the same for every provider, and you shouldn't need to modify this function.
+    pub async fn run() -> anyhow::Result<()> {
+        let HostData {
+            secrets, config, ..
+        } = load_host_data().context("failed to load host data")?;
+        let SecretValue::String(password) = secrets
+            .get("redis_password")
+            .expect(&format!("redis_password secret not found: {:?}", secrets))
```
vados-cosmonic, line 39:
```suggestion
            .expect(&format!("redis_password secret not found: {:?}", secrets))
```

Ah I think this should be `.context(..)`?

examples/security/secrets/provider-keyvalue-redis-password/src/provider.rs
```diff
@@ -0,0 +1,106 @@
...
+        else {
+            bail!("password secret not a string")
+        };
+        let url = config
+            .get("url")
+            .cloned()
+            .unwrap_or_else(|| "127.0.0.1:6379".to_string());
```
vados-cosmonic, line 46: Do you need to clone the value here? Also might make sense to hoist this value out to a `DEFAULT_REDIS_URL` or something

examples/security/secrets/provider-keyvalue-redis-password/src/provider.rs
```diff
@@ -0,0 +1,106 @@
...
+        let conn_manager = match redis::Client::open(format!("redis://:{password}@{url}")) {
+            Ok(client) => match client.get_connection_manager().await {
```
vados-cosmonic, line 48: This might be easier to read as separate `let else` statements

examples/security/secrets/provider-keyvalue-redis-password/src/provider.rs
```diff
@@ -0,0 +1,106 @@
...
+        delta: u64,
+    ) -> anyhow::Result<Result<u64, exports::wrpc::keyvalue::atomics::Error>> {
+        Ok(self
+            .default_connection
+            .write()
+            .await
+            .incr(key, delta)
+            .await
+            .map_err(|e| {
+                exports::wrpc::keyvalue::atomics::Error::Other(format!("redis error {:?}", e))
```
vados-cosmonic, line 101:
```suggestion
                exports::wrpc::keyvalue::atomics::Error::Other(format!("redis error: {:?}", e))
```

examples/security/secrets/secret-setup/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+    let component_secret = wasmcloud_secrets_types::Secret {
+        name: "api_password".to_string(),
+        string_secret: Some("opensesame".to_string()),
+        ..Default::default()
+    };
+    let provider_secret = wasmcloud_secrets_types::Secret {
+        name: "redis_password".to_string(),
+        string_secret: Some("sup3rS3cr3tP4ssw0rd".to_string()),
+        ..Default::default()
+    };
```
vados-cosmonic, line 13: In lieu of some top level docs describing this binary, what do you think about moving these to where they're used? 

Then this `main` would be like... 4 lines, which are really clearly putting secrets.

Also, if you made the function `put_secrets` take an iterator of secrets we could do multiple at the same time!

examples/security/secrets/secret-setup/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    let nats_client = async_nats::connect("127.0.0.1:4222").await?;
+
+    put_secret(&nats_client, component_secret).await?;
+    put_secret(&nats_client, provider_secret).await?;
+
+    Ok(())
+}
+
+async fn put_secret(
```
vados-cosmonic, line 23: NIT: This feels like it could go on a trait in `wasmcloud_secrets_types` and be implemented for `async_nats::Client` under a feature flag or something... Sure wouldn't want to write this again

examples/security/secrets/README.md
brooksmtownsend: It's hard since both bash and make have difficulties running on windows, so there's not a great option there. I would personally love to condense this down to either a cargo example or a much smaller set of runnable commands in the README, but for now this example is just too complicated to setup (lack of wadm support for secrets, running binaries from main, etc.)

I think this example has quite a long way to go before it is polished but I would ideally like to split it out into different README steps once it's complete. I was thinking the current scripts are just a little too long to do that now though

examples/security/secrets/provider-keyvalue-redis-password/src/provider.rs
brooksmtownsend: I do just because I can't return a `&String` from the unwrap_or_else 😢 But I'll bring it up to a const

examples/security/secrets/provider-keyvalue-redis-password/src/provider.rs
brooksmtownsend: I did some refactoring of this method in #2505 so I will take a look there since it's abstracted to a function

examples/security/secrets/secret-setup/src/main.rs
brooksmtownsend: > what do you think about moving these to where they're used?

What do you mean by this, like moving the secret out to another place?

examples/security/secrets/secret-setup/src/main.rs
brooksmtownsend: Yeah I think this should be a part of a library from the nats-kv implementation of the secrets backend, since this is specifically built for putting a secret there

brooksmtownsend: Addressed by moving the function to the NATS KV library

examples/security/secrets/README.md
vados-cosmonic: Yeah I can see that -- let's punt on this

examples/security/secrets/secret-setup/src/main.rs
vados-cosmonic: Ah this was a NIT about the secret creation itself -- they could have been defined inline. Definitel ya NIT though!

brooksmtownsend: Ah okay. Yeah I think it looks a little cleaner now with the moved function to secrets-nats-kv, let me know what you think. We can also add an `add_mapping` function now that I'm thinking about it to make it more clear

vados-cosmonic (approved):
LGTM :rocket:

examples/security/secrets/secret-setup/src/main.rs
vados-cosmonic: Oh there are a couple more things we could do here -- we could make the fields `wasmcloud_secrets_types::Secret` *not* `pub` so that we can enforce the string/binary secret difference (or use an enum), and some constructors would be nice... but honestly that's all icing.

thomastaylor312 (approved):

PR #2400, opened 2024-06-27
feat(custom): Fixed the exception on link Events in custom component

ritesh089:
## Feature or Problem
Fixed the exception on link Events in custom component by using the latest provider sdk for go.


vados-cosmonic:
Hey @ritesh089 this looks awesome -- would you mind changing the `scope` in your commit to `examples`? Otherwise this looks perfect -- though I'll defer to @brooksmtownsend since he's a bit more familiar to the Golang provider (and might be able to better tell if we *shouldn't* be updating just yet or something).

examples/golang/providers/custom-template/.gitignore
```diff
@@ -1,2 +1,3 @@
 keyvalue-inmemory
-keyvalue-inmemory.par.gz
\ No newline at end of file
+keyvalue-inmemory.par.gz
+build/*
```
brooksmtownsend, line 3: I actually chose to commit the component artifact so that users of the custom provider can just build the provider and automatically have a component to test with. I'm not strongly held to this methodology, but I do like the simplicity of it

brooksmtownsend (changes requested):
If this works, should we revert #2386 in this PR?

examples/golang/providers/custom-template/.gitignore
ritesh089: isnt it better to add an instruction to perform a wash build for the component ? Instead of placing the build results in git?

ritesh089:
> If this works, should we revert #2386 in this PR?

Done

examples/golang/providers/custom-template/.gitignore
ritesh089: I updated the instructions in Readme to build the component.

ritesh089:
> Hey @ritesh089 this looks awesome -- would you mind changing the `scope` in your commit to `examples`? Otherwise this looks perfect -- though I'll defer to @brooksmtownsend since he's a bit more familiar to the Golang provider (and might be able to better tell if we _shouldn't_ be updating just yet or something).

I changed the scope to examples.

vados-cosmonic (approved):
LGTM :rocket:

PR #2401, opened 2024-06-27
fix(example): direction of links

brooksmtownsend:
## Feature or Problem
I even get these confused sometimes :(

Also updated the `close()` function based on Roman's recommendation


vados-cosmonic (approved):
LGTM :rocket:

PR #2418, opened 2024-06-30
feat: Allow per-signal enabling to override enable-observability

joonas:
## Feature or Problem

Being able to support overriding individual signal whether observability as a whole is enabled or not makes it easier to write code that wraps the host (such as the `wasmcloud-operator`).

I don't see there being really any downsides to this, nor does it seem like breaking any existing contracts, since any previous configurations that might've tried controlling individual signals while also enabling observability as a whole wouldn't have worked.


joonas:
Never mind.

PR #2426, opened 2024-07-01
fix(tests): wash tests

vados-cosmonic:


vados-cosmonic:
Closing this for now -- on the 8 core windows machines we get the same error as smaller machines (stack overflow). Finding the root cause can be done off of `wasmcloud/wasmcloud`

PR #2431, opened 2024-07-02
refactor(providers): enable consumer groups with kafka provider

### LINKED ISSUE #1442: [FEAT] Ensure Kafka provider can run multiple copies without duplicate handling
The current implementation of the [messaging-kafka](https://github.com/wasmCloud/capability-providers/tree/main/messaging-kafka) provider is fairly experimental and created as an example. I would love it if the Kafka provider was closer to being ready for a real workload by ensuring that, while running multiple Kafka providers, it can consume/produce from/to a Kafka stream without duplicate handling of messages. This is a great issue to take on if you're familiar with Kafka and want to take a stab at writing a wasmCloud provider 😄

vados-cosmonic:
This commit enables consumer groups via link time configuration in the kafka provider.

The native Rust ecosystem for Kafka is somewhat immature, so to do this we must switch to a different crate
`kafka-rs` (https://crates.io/crates/kafka), which implements consumer groups but is otherwise missing some features.

With this branch, one can specify CONSUMER_GROUP in link configuration to control the consumer group that is used for Kafka consumers (whose incoming messages will trigger downstream components).

Resolves #1442

### DIFF
--- a/crates/provider-messaging-kafka/README.md
+++ b/crates/provider-messaging-kafka/README.md
@@ -12,10 +12,16 @@ It exposes publish and subscribe functionality to components to operate on Kafka
 
 ## Named Config Settings
 
-| Property | Description                                                                                                                                                                                                                                                                |
-| -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `hosts`  | A comma-separated list of bootstrap server hosts. For example, `HOSTS=127.0.0.1:9092,127.0.0.1:9093`. A single value is accepted as well, and the default value is the Kafka default of `127.0.0.1:9092`. This will be used for both the consumer and producer connections |
-| `topic`  | The Kafka topic you wish to consume. Any messages on this topic will be forwarded to this component for processing                                                                                                                                                         |
+| Property              | Description                                                                                                                                                                                                                                                                |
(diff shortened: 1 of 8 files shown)


crates/provider-messaging-kafka/README.md
```diff
@@ -12,10 +12,16 @@ It exposes publish and subscribe functionality to components to operate on Kafka
...
+| `topic`               | The Kafka topic you wish to consume. Any messages on this topic will be forwarded to this component for processing                                                                                                                                                         |
+| `consumer-group`      | Consumer group to use when consuming messages                                                                                                                                                                                                                              |
+| `consumer-partitions` | Comma delimited list of partitions to use when subscribing to the topic specified by the link.                                                                                                                                                                             |
+| `producer-partitions` | Comma delimited list of partitions to use when handling `publish` calls from components (unrelated to the subscription topic)                                                                                                                                              |
```
brooksmtownsend, line 21: I know we use config with underscores in other places, should we do that here too?

brooksmtownsend (approved):
Looks great! Nit around the link config key names

crates/provider-messaging-kafka/README.md
vados-cosmonic: Sure, I can use underscores here -- thanks for pointing this out, will change!

vados-cosmonic:
This PR needs to be rebased to `main` once it has https://github.com/wasmCloud/wasmCloud/pull/2489

PR #2432, opened 2024-07-02
added support for new and old wrpc transport versions

LUK3ARK:
This is in preparation for next changes. no changes in business logic, just dependency resolving


crates/core/src/wrpc.rs
```diff
@@ -11,6 +11,9 @@
 //!
 //! [wrpc-transport]: https://docs.rs/wrpc-transport
 
+extern crate wrpc_transport_legacy as wrpc_transport;
```
rvolosatovs, line 14:
```suggestion
pub use wrpc_transport_legacy as wrpc_transport;
```

crates/runtime/src/io.rs
```diff
@@ -7,10 +7,8 @@ use std::sync::{Arc, MutexGuard};
 use bytes::Bytes;
 use futures::{Stream, StreamExt as _};
 use tokio::io::{AsyncRead, AsyncSeek, AsyncWrite, ReadBuf};
-use wrpc_transport::IncomingInputStream;
-
+use wrpc_transport_legacy::IncomingInputStream;
 /// wasmCloud I/O functionality
-
```
rvolosatovs, line 13: Could you bring this line back?

Cargo.toml
```diff
@@ -345,6 +345,8 @@ wit-parser = { version = "0.202", default-features = false }
 wrpc-interface-blobstore = { version = "0.16", default-features = false }
 wrpc-interface-http = { version = "0.22", default-features = false }
 wrpc-runtime-wasmtime = { version = "0.16", default-features = false }
-wrpc-transport = { version = "0.24.3", default-features = false }
-wrpc-transport-nats = { version = "0.21", default-features = false }
+wrpc-transport = { git = "https://github.com/bytecodealliance/wrpc.git", default-features = false }
```
rvolosatovs, line 348: We should set a version here

PR #2433, opened 2024-07-03
Update Wrpc 

LUK3ARK:
Another attempt


LUK3ARK:
So there are even more blockers than the previous now due to updates in versions again.

- Host crate has heavy dependency on two versions of async nats, I cannot sort this out
- Depend on wasmtime 0.22 is forced, but required a solution to passing in a parameter that uses both wasiImpl and WasiHttpImpl. 
- wrpc interface needs updating to new wasmtime interface, unless you just want to ignore the config and create a mapping to a hyper body from previous versions

handler logic is done, this depends on wrpc client being updated to use latest wrpc

LUK3ARK:
this is basically a dependency grid lock. I dont see a way solving this without updating some upstream crates 🤷‍♂️

Issue #2398, opened 2024-06-27
[FEATURE] `wash get inventory` output is unreadable


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Looking at the inventory of hosts is one of the most common operations for `wash`. However, when using file references or long generated component IDs, it becomes pretty unreadable.
![image](https://github.com/wasmCloud/wasmCloud/assets/12040685/be75a3ac-05f7-4b2d-a3b5-7da861ae5c05)
![image](https://github.com/wasmCloud/wasmCloud/assets/12040685/ae0d7f2b-1342-48b7-afe4-9157a4ebcff8)

Taking a look at other infrastructure tools, like kubectl and nomad, none of them quite overflow the terminal like wash. I think this is largely due to the fidelity of information, and with wash displaying the full image reference really just isn't quite necessary.

### Describe the solution you'd like

I would love for the host inventory to display just the right amount of information for the human readable format. Just like kubernetes, we can hide all of the other information behind the `json` flag for when it's necessary.
![image](https://github.com/wasmCloud/wasmCloud/assets/12040685/121bb53b-5c89-471c-b15a-d6aa4ba81fd7)

I propose that the host inventory display the following information:
- Host ID
- Host friendly name
- Host Labels
- Components list
  - Component ID
  - Name
  - Max count
- providers list
  - Provider ID
  - Name

I could even be convinced to drop the friendly name / host labels for the inventory operation, but it can be useful at a glance to verify that, for example, your application is spread properly across availability zones.

### Describe alternatives you've considered

- We could push host information into the `wash get hosts` command to keep the terminal more clean
- We could be smarter with our table library to fit to the width of the terminal and truncate the image reference. This could work, but a truncated image reference or ID is pretty unhelpful

ossfellow:
I can take this; please assign it to me.

ossfellow:
Hello @brooksmtownsend,
Does this look okay? If it does, then, I'll create the PR; otherwise let me know of any other improvements you'd like to see:

![image](https://github.com/wasmCloud/wasmCloud/assets/50174679/78ac1a7a-d801-4254-8b2c-8f89132a3331)

[...]

PR #2435, opened 2024-07-03
feat: Improved the output of wash get inventory command, as outlined in issue #2398

ossfellow:
## Feature or Problem
This PR improves the output of `wash get inventory` command.

## Related Issues
[Issue #2398](https://github.com/wasmCloud/wasmCloud/issues/2398)

## Testing
The new output format has been tested locally.

### DIFF
--- a/crates/wash-cli/src/ctl/output.rs
+++ b/crates/wash-cli/src/ctl/output.rs
@@ -110,16 +110,24 @@ pub fn host_inventories_table(invs: Vec<HostInventory>) -> String {
     crate::util::configure_table_style(&mut table);
 
     invs.into_iter().for_each(|inv| {
-        table.add_row(Row::new(vec![TableCell::new_with_alignment(
-            format!("Host Inventory ({})", inv.host_id),
-            4,
-            Alignment::Left,
-        )]));
+        table.add_row(Row::new(vec![
+            TableCell::new_with_alignment("Host ID", 1, Alignment::Left),
+            TableCell::new_with_alignment("Friendly name", 1, Alignment::Left),
+        ]));
+        table.add_row(Row::new(vec![
+            TableCell::new_with_alignment(inv.host_id.clone(), 1, Alignment::Left),
+            TableCell::new_with_alignment(inv.friendly_name.clone(), 1, Alignment::Left),
+        ]));
 
         if !inv.labels.is_empty() {
             table.add_row(Row::new(vec![TableCell::new_with_alignment(
                 "",
                 2,
+                Alignment::Left,
+            )]));
+            table.add_row(Row::new(vec![TableCell::new_with_alignment(
+                "Host labels",
+                1,
                 Alignment::Left,
             )]));
             inv.labels.iter().for_each(|(k, v)| {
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Hey @ossfellow sorry for the delay! Would you mind adding a test to this PR?

I think one of the things @brooksmtownsend was really trying to get at was the fact that when we had a long component ref (ex. `ghcr.io/some/repository/with/a/component/in/it:0.3.0`), it would make the regular non-JSON output unreadable due to line wrapping. 

It would be awesome if we had a test that was *exactly* that case -- just checking the lines in stdout would be enough. You could mostly copy https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-cli/tests/wash_get.rs#L75 (that's probably also where the test should go), and we could get a nice regression test for this!

vados-cosmonic:
Oh and BTW the fix to the integration tests should be merged soon -- so if you wouldn't mind clicking the "allow edits by maintainers" checkbox, we can rebase in one click from this side before merging!

ossfellow:
@vados-cosmonic, I can certainly add the test and let maintainers update the PR (which I thought is the default); I'll be back at my desk in a couple of hours and will do what you've suggested.
Thanks for the test sample btw.

ossfellow:
> Hey @ossfellow sorry for the delay! Would you mind adding a test to this PR?
> 
> I think one of the things @brooksmtownsend was really trying to get at was the fact that when we had a long component ref (ex. `ghcr.io/some/repository/with/a/component/in/it:0.3.0`), it would make the regular non-JSON output unreadable due to line wrapping.
> 
> It would be awesome if we had a test that was _exactly_ that case -- just checking the lines in stdout would be enough. You could mostly copy https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-cli/tests/wash_get.rs#L75 (that's probably also where the test should go), and we could get a nice regression test for this!

Hello @vados-cosmonic, based on the requirement, I have removed the `component ref` from the non-JSON output; so, could you please clarify if you want me to add a test for the JSON output, or any of the attributes of the non-JSON output?

Also, the option to allow maintainers edit the PR doesn't show under the list of PR participants; if this isn't due to a project rule, maybe converting the PR to draft will make it work, although I'm not sure to be honest.

Once you clarify, I'll add the test and update the PR.

vados-cosmonic:
Hey @ossfellow that's a good point -- so I think we might want to solve that a slightly different way... Rather than *remove* the `ref`, we might want to shorten/length constrain it (ellipses?)... And we definitely don't want it out of the JSON output, because that is probably used for machines to read and we want to provide as much information as possible (might even break some tests we have now). 

Oh and if it isn't there then don't worry about it I guess! Not sure why it wouldn't show up!

ossfellow:
@vados-cosmonic, there seems to be a difference of opinions over what should be in the table formatted output of the `wash get inventory` command.

While I'm not against adding the `image_ref` back in the output, may I ask you to speak to @brooksmtownsend and update the issue this PR is referencing, for the sake of consistency?

If it's decided to add it back in, I'll only add the file name and drop the path.

Also, please note that the json output is left intact, as the ask was for the table output.

vados-cosmonic:
Hey @ossfellow I see what you meant!

Oh you're right @brooksmtownsend wanted to remove the image reference completely, as it says in the issue:

>  with wash displaying the full image reference really just isn't quite necessary.

I think based on what's in the issue you've absolutely solved the problem, but it would be nice to have some way to make sure that the inventory command doesn't overflow again, given possibly long out put for any of those other fields... 

For example I'm thinking what if I make a really long friendly name -- do things go off the rails then?

What do you think?

> Also, please note that the json output is left intact, as the ask was for the table output.

Great, thank you!

ossfellow:
Hey @vados-cosmonic, it's possible that _name_, or _label_ fields could potentially get longer that the width of output window; but all the variables (i.e. what name to use, what labels to use, and output window size) are user selected, and I'm not if, on the _wash cli_ side, we could, or even should, force anything on the user.
I think the intent of the issue Brooks created has been to minimize the chance of output line wrapping, without being prescriptive about user choices.

Nonetheless, I'm totally open to any suggestions, which could improve the user experience.

brooksmtownsend (approved):
This looks great to me! Likely just needs a rebase to bring in some of the latest changes to fix tests. @ossfellow I can take care of this if you'd like, just let me know, otherwise I will wait for your rebase 😄 

I know that there's probably some smart things we can do, especially around names, in order to wrap longer names, but I think that can be an optimization we make when we want to be a bit smarter with the table. This should really help 95% of the host inventories look better

vados-cosmonic:
Hey @ossfellow would you mind rebasing this PR to `main` if it isn't already? not sure why CI is stuck...

I'll mark it "merge when ready" so that it can merge as soon as checks pass.

ossfellow:
I did rebase it last night and the PR is showing up to date on GitHub (which as you know will show if it isn’t).
As a lost resort, if the CI remains stuck, I can close the PR and issue a new one; but hopefully we won’t get there.

brooksmtownsend:
Trying to get GHA to trigger.

Issue #2016, opened 2024-04-30
[FEATURE] `max_execution_time` config


rvolosatovs:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [X] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

In #2002 and #2015 support for `max_execution_time` was added to the host and runtime, it should be configurable per-host and also per-component

### Describe the solution you'd like

Host-level config via flags and env vars
Component-level config via control interface

brooksmtownsend:
This is a great first issue to get into 1.1!

Aditya1404Sal:
I'd love to work on this feature, @vados-cosmonic  can I pick this up ?

vados-cosmonic:
Hey I think so @Aditya1404Sal -- thanks for hopping on this!

Aditya1404Sal:
> Hey I think so @Aditya1404Sal -- thanks for hopping on this! IMO you could do it in two commits -- first the host configuration and then the component-level config separately!

Yes that sounds great -- that way it'll be systematic and easily trackable :+1:

Aditya1404Sal:
> Hey I think so @Aditya1404Sal -- thanks for hopping on this! IMO you could do it in two commits -- first the host configuration and then the component-level config separately!

also, could I have this ticket assigned to me? to avoid any conflicts

vados-cosmonic:
Sure!

vados-cosmonic:
Hey @Aditya1404Sal it's probably a good idea to break this up into two distinct PRs -- let's update the host first and control interface/component-level configuration can be handled later!

Aditya1404Sal:
I understand :+1: -- I'll make a separate PR for the Component-level configuration after host is updated :)

Aditya1404Sal:
> Hey @rvolosatovs  -- The Issue ticket suggests adding Component-level configurability for max_execution_time via control interface
[...]

PR #2439, opened 2024-07-03
feat(host,cli): max_execution_time config flag support for host runtime

Aditya1404Sal:
## Feature or Problem
Adds a `--max-execution-time` flag to `wash up` for configuring the host's max execution time implemented in #2002 and #2015 

## Related Issues
(1/2) of  #2016 
## Release Information
next
### Manual Verification
Ran `cargo run -- up --max-execution-time <humantime>` to which `wash -up` ran as usual, value propagation and time duration formatting works.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -769,6 +769,7 @@ impl Host {
 
         // TODO: Configure
         let (runtime, epoch, epoch_end) = Runtime::builder()
+            .max_execution_time(config.max_execution_time)
             .build()
             .context("failed to build runtime")?;
         let event_builder = EventBuilderV10::new().source(host_key.public_key());
@@ -840,6 +841,8 @@ impl Host {
 
         let config_generator = BundleGenerator::new(config_data.clone());
 
+        let max_execution_time_ms = config.max_execution_time;
+
         let host = Host {
             components: RwLock::default(),
             event_builder,
@@ -869,7 +872,7 @@ impl Host {
             component_claims: Arc::default(),
             provider_claims: Arc::default(),
             metrics: Arc::new(metrics),
-            max_execution_time: Duration::from_secs(10 * 60),
+            max_execution_time: max_execution_time_ms,
         };
 
         let host = Arc::new(host);
--- a/crates/wash-cli/src/up/config.rs
+++ b/crates/wash-cli/src/up/config.rs
@@ -21,6 +21,8 @@ pub const WASMCLOUD_JS_DOMAIN: &str = "WASMCLOUD_JS_DOMAIN";
 pub const WASMCLOUD_CLUSTER_ISSUERS: &str = "WASMCLOUD_CLUSTER_ISSUERS";
 pub const WASMCLOUD_CLUSTER_SEED: &str = "WASMCLOUD_CLUSTER_SEED";
 pub const WASMCLOUD_HOST_SEED: &str = "WASMCLOUD_HOST_SEED";
+pub const WASMCLOUD_MAX_EXECUTION_TIME_MS: &str = "WASMCLOUD_MAX_EXECUTION_TIME_MS";
+pub const DEFAULT_MAX_EXECUTION_TIME_MS: &str = "600000";
 // NATS RPC connection configuration
 pub const WASMCLOUD_RPC_HOST: &str = "WASMCLOUD_RPC_HOST";
 pub const WASMCLOUD_RPC_PORT: &str = "WASMCLOUD_RPC_PORT";
@@ -81,6 +83,10 @@ pub async fn configure_host_env(wasmcloud_opts: WasmcloudOpts) -> Result<HashMap
             cluster_issuers.join(","),
         );
     }
+    host_config.insert(
+        WASMCLOUD_MAX_EXECUTION_TIME_MS.to_string(),
+        wasmcloud_opts.max_execution_time.to_string(),
+    );
 
     if wasmcloud_opts.allow_latest {
         host_config.insert(WASMCLOUD_OCI_ALLOW_LATEST.to_string(), "true".to_string());
--- a/crates/wash-cli/src/up/mod.rs
+++ b/crates/wash-cli/src/up/mod.rs
@@ -1,3 +1,7 @@
+use anyhow::{anyhow, bail, Context, Result};
+use async_nats::Client;
+use clap::Parser;
+use serde_json::{json, Value};
 use std::collections::HashMap;
 use std::fmt::Write;
 use std::io::ErrorKind;
@@ -7,13 +11,7 @@ use std::sync::{
     atomic::{AtomicBool, Ordering},
     Arc,
 };
-
-use anyhow::{anyhow, bail, Context, Result};
-use async_nats::Client;
-use clap::Parser;
-use serde_json::{json, Value};
 use sysinfo::{System, SystemExt};
-
 use tokio::fs::create_dir_all;
 use tokio::{
     io::{AsyncBufReadExt, BufReader},
@@ -272,6 +270,10 @@ pub struct WasmcloudOpts {
     /// If enabled, allows starting additional wasmCloud hosts on this machine
     #[clap(long = "multi-local")]
     pub multi_local: bool,
+
+    /// Defines the Max Execution time (in ms) that the host runtime will execute for
+    #[clap(long = "max-execution-time-ms", alias = "max-time-ms", env = WASMCLOUD_MAX_EXECUTION_TIME_MS, default_value = DEFAULT_MAX_EXECUTION_TIME_MS)]
+    pub max_execution_time: u64,
 }
 
 impl WasmcloudOpts {
@@ -355,11 +357,9 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
     let ctx = ContextDir::new()?
         .load_default_context()
         .context("failed to load context")?;
-
     // falling back to the context's ctl_ connection won't always be right, but we have to pick one, since the context values are not optional
     let nats_host = cmd.nats_opts.nats_host.clone().unwrap_or(ctx.ctl_host);
     let nats_port = cmd.nats_opts.nats_port.unwrap_or(ctx.ctl_port);
-
     let wasmcloud_opts = WasmcloudOpts {
         lattice: Some(cmd.wasmcloud_opts.lattice.unwrap_or(ctx.lattice)),
         ctl_host: Some(cmd.wasmcloud_opts.ctl_host.unwrap_or(nats_host.clone())),
@@ -373,6 +373,7 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
         rpc_jwt: cmd.wasmcloud_opts.rpc_jwt.or(ctx.rpc_jwt),
         rpc_seed: cmd.wasmcloud_opts.rpc_seed.or(ctx.rpc_seed),
         rpc_credsfile: cmd.wasmcloud_opts.rpc_credsfile.or(ctx.rpc_credsfile),
+        max_execution_time: cmd.wasmcloud_opts.max_execution_time,
         cluster_seed: cmd
             .wasmcloud_opts
             .cluster_seed
--- a/src/main.rs
+++ b/src/main.rs
@@ -173,6 +173,9 @@ struct Args {
         requires = "policy_topic"
     )]
     policy_changes_topic: Option<String>,
+    /// If provided, allows to set a custom Max Execution time for the Host in ms.
+    #[clap(long = "max-execution-time-ms", default_value = "600000", env = "WASMCLOUD_MAX_EXECUTION_TIME_MS", value_parser = parse_duration)]
+    max_execution_time: Duration,
     /// If provided, allows setting a custom timeout for requesting policy decisions. Defaults to one second. Requires `policy_topic` to be set.
     #[clap(
         long = "policy-timeout-ms",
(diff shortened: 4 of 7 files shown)


crates/runtime/src/runtime.rs
```diff
@@ -54,6 +54,26 @@ impl RuntimeBuilder {
             force_pooling_allocator: false,
         }
     }
+    /// Returns a custom Max Execution Time [`RuntimeBuilder`]
+    pub fn custom(max_exec_time: Duration) -> Self {
```
rvolosatovs, line 58: how about we add a `max_execution_time` method on `RuntimeBuilder` instead?

rvolosatovs: Actually, I see we already have it, why would we need the `custom` here?

crates/host/src/wasmbus/mod.rs
```diff
@@ -864,7 +864,7 @@ impl Host {
         let (stop_tx, stop_rx) = watch::channel(None);
 
         // TODO: Configure
-        let (runtime, epoch, epoch_end) = Runtime::builder()
+        let (runtime, epoch, epoch_end) = Runtime::custom_builder(config.max_execution_time.clone())
```
rvolosatovs, line 867:
```suggestion
        let (runtime, epoch, epoch_end) = Runtime::builder().max_execution_time(config.max_execution_time)
```

crates/wash-cli/src/up/mod.rs
```diff
@@ -272,6 +272,10 @@ pub struct WasmcloudOpts {
     /// If enabled, allows starting additional wasmCloud hosts on this machine
     #[clap(long = "multi-local")]
     pub multi_local: bool,
+
+    /// Defines the Max Execution time (in minutes) that the host runtime will execute for
+    #[clap(long = "max-execution-time", alias = "max-time", env = WASMCLOUD_MAX_EXECUTION_TIME)]
+    pub max_execution_time: Option<u64>,
```
rvolosatovs, line 278: I'm not sure that minutes are granular enough. Should we make this seconds?
Or, even better, how about we allow a string here like `2s` for 2 seconds or `5m` for 5 minutes?

crates/runtime/src/runtime.rs
Aditya1404Sal: > Actually, I see we already have it, why would we need the `custom` here?

That makes so much more sense, yep we don't need a custom RunBuilder generator _yet_ (till there's a need to make things more granular, which is still farfetched)

crates/wash-cli/src/up/mod.rs
Aditya1404Sal: Okay, cmd input type needs more granularity -- about allowing strings as `2s` and `5m`, 
this function would allow us to do exactly that -- any thoughts or suggestions ?
```rust
fn parse_duration(s: &str) -> Result<Duration, String> {
    if let Some(secs) = s.strip_suffix('s') {
        secs.parse::<u64>()
            .map(Duration::from_secs)
            .map_err(|e| format!("Invalid duration: {}", e))
    } else if let Some(mins) = s.strip_suffix('m') {
        mins.parse::<u64>()
            .map(|m| Duration::from_secs(m * 60))
            .map_err(|e| format!("Invalid duration: {}", e))
    } else {
        Err("Invalid duration format, use 's' for seconds or 'm' for minutes".to_string())
    }
}
```

crates/wash-cli/src/up/mod.rs
```diff
@@ -711,6 +725,21 @@ fn is_process_running(pid: &str) -> bool {
     }
 }
 
+/// Parses max_execution_time input string to return a duration string.
```
rvolosatovs, line 728: perhaps it makes sense to adopt something like https://docs.rs/clap-duration/latest/clap_duration/macro.duration_range_value_parse.html instead of doing manual parsing here?

crates/wash-cli/src/up/mod.rs
```diff
@@ -272,6 +272,10 @@ pub struct WasmcloudOpts {
     /// If enabled, allows starting additional wasmCloud hosts on this machine
     #[clap(long = "multi-local")]
     pub multi_local: bool,
+
+    /// Defines the Max Execution time (in minutes or seconds) that the host runtime will execute for
+    #[clap(long = "max-execution-time", alias = "max-time", env = WASMCLOUD_MAX_EXECUTION_TIME)]
+    pub max_execution_time: Option<String>,
```
rvolosatovs, line 278: Let's set the default here and remove the `Option`
```suggestion
    /// Defines the Max Execution time (in minutes or seconds) that the host runtime will execute for
    #[clap(long = "max-execution-time", alias = "max-time", env = WASMCLOUD_MAX_EXECUTION_TIME, default_value = DEFAULT_MAX_EXECUTION_TIME)]
    pub max_execution_time: String,
```

ideally, though, this would be a duration type like the one provided by https://docs.rs/clap-duration/0.1.11/clap_duration/macro.duration_range_value_parse.html

crates/wash-cli/src/up/mod.rs
```diff
@@ -355,11 +359,20 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
...
+    let max_exec_time = match max_exec_time_str {
+        Some(ref time_str) => match parse_duration(time_str) {
+            Ok(duration) => Some(duration),
+            Err(e) => {
+                eprintln!("Error parsing max-execution-time: {}", e);
+                None
+            }
+        },
+        None => None,
+    };
```
rvolosatovs, line 372: once the value is non-optional:
```suggestion
    let max_execution_time = parse_duration(&cmd.wasmcloud_opts.max_execution_time).context("failed to parse `max-execution-time`")?;
```

if the execution time cannot be parsed, I think we should immediately

crates/wash-cli/src/up/mod.rs
```diff
@@ -373,6 +386,7 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
         rpc_jwt: cmd.wasmcloud_opts.rpc_jwt.or(ctx.rpc_jwt),
         rpc_seed: cmd.wasmcloud_opts.rpc_seed.or(ctx.rpc_seed),
         rpc_credsfile: cmd.wasmcloud_opts.rpc_credsfile.or(ctx.rpc_credsfile),
+        max_execution_time: max_exec_time,
```
rvolosatovs, line 389:
```suggestion
        max_execution_time,
```

crates/wash-cli/src/up/mod.rs
```diff
@@ -711,6 +725,21 @@ fn is_process_running(pid: &str) -> bool {
     }
 }
 
+/// Parses max_execution_time input string to return a duration string.
+fn parse_duration(s: &str) -> Result<String, String> {
```
rvolosatovs, line 729: If we do manual parsing, let's return `anyhow::Result<Duration>`

crates/wash-cli/src/up/mod.rs
rvolosatovs: actually, I just found out we can just use https://docs.rs/humantime/latest/humantime/struct.Duration.html directly

crates/wash-cli/src/up/mod.rs
rvolosatovs, line 278: better alternative:
```suggestion
    pub max_execution_time: humantime::Duration,
```

crates/wash-cli/src/up/mod.rs
brooksmtownsend: Based on our own baselines, we currently measure other times in terms of milliseconds which gives a pretty good level of granularity. What are your thoughts on just continuing that pattern?

So, by default, the execution timeout would be `10 * 60 * 1000` aka `600000` milliseconds. Then the environment variable could be `WASMCLOUD_MAX_EXECUTION_TIME_MS`

src/main.rs
```diff
@@ -169,6 +169,9 @@ struct Args {
         requires = "policy_topic"
     )]
     policy_changes_topic: Option<String>,
+    /// If provided, allows to set a custom Max Execution time for the Host in seconds.
+    #[clap(long = "max-execution-time", env = "WASMCLOUD_MAX_EXECUTION_TIME")]
+    max_execution_time: Option<u64>,
```
Aditya1404Sal, line 174: Since default time is already present ,there's no need for `Option<>` right ?

crates/wash-cli/src/up/mod.rs
Aditya1404Sal: I don't think we should continue the manual parsing, humantime::Duration is much more flexible 
input like `2h45min22s` gets auto-parsed into `Duration` which we can directly push into `host_config` as either milliseconds or seconds depending on which one is more granular.

Aditya1404Sal:
Hey @brooksmtownsend ,Just as a heads up I'll proceed with refactoring the changes in compliance with milli-second time Format.

Aditya1404Sal:
https://github.com/wasmCloud/wasmCloud/blob/b2ccdeb6ddfcf2dfc7d92041a877e99741e9d103/crates/runtime/src/runtime.rs#L77-L82
@rvolosatovs 
It's a little out of scope but this is a valid refactor right ? 

```rust
max_execution_time.max(Duration::from_millis(1000)
```

Aditya1404Sal:
> > https://github.com/wasmCloud/wasmCloud/blob/b2ccdeb6ddfcf2dfc7d92041a877e99741e9d103/crates/runtime/src/runtime.rs#L77-L82
> > 
> > @rvolosatovs
> > It's a little out of scope but this is a valid refactor right ?
> > ```rust
> > max_execution_time.max(Duration::from_millis(1000)
> > ```
> 
> I don't think it's wrong, but could you help me understand why we'd want to switch from `Duration::from_secs(1)` to `Duration::from_millis(1000)`? Current approach seems easier to understand for me

I assumed it was because the max() would compare the max_execution_time in milliseconds to 1 sec (1000 > 1) due to the refactor, , so it would never get set to 1 second. Because host_config.max_ex..... is of type Duration::from_millis (the `new` constructor)  
but I was wrong , Compiler will automatically compare it in the same standard of duration
sorry for the flase call!

brooksmtownsend:
@Aditya1404Sal we had some fixes go into `main` for the CI staying in Expected, if you rebase off of main that should resolve that!

Aditya1404Sal:
> @Aditya1404Sal we had some fixes go into `main` for the CI staying in Expected, if you rebase off of main that should resolve that!
Done!

Aditya1404Sal:
Hey @brooksmtownsend, This PR is good to go for a final review, take a look whenever you're free :smile: .

src/main.rs
```diff
@@ -169,6 +169,9 @@ struct Args {
         requires = "policy_topic"
     )]
     policy_changes_topic: Option<String>,
+    /// If provided, allows to set a custom Max Execution time for the Host in ms.
+    #[clap(long = "max-execution-time", env = "WASMCLOUD_MAX_EXECUTION_TIME_MS")]
+    max_execution_time: u64,
```
brooksmtownsend, line 174: We can use existing parsers to make this simpler
```suggestion
    /// If provided, allows to set a custom Max Execution time for the Host in ms.
        #[clap(long = "max-execution-time-ms", default_value = "2000", env = "WASMCLOUD_MAX_EXECUTION_TIME_MS", value_parser = parse_duration)]
    max_execution_time: Duration,
```

src/main.rs
```diff
@@ -395,6 +397,7 @@ async fn main() -> anyhow::Result<()> {
         otel_config,
         policy_service_config,
         version: env!("CARGO_PKG_VERSION").to_string(),
+        max_execution_time: Duration::from_millis(args.max_execution_time),
```
brooksmtownsend, line 400: Once you change the above you can just pass the duration directly

crates/wash-cli/src/up/mod.rs
```diff
@@ -272,6 +270,10 @@ pub struct WasmcloudOpts {
     /// If enabled, allows starting additional wasmCloud hosts on this machine
     #[clap(long = "multi-local")]
     pub multi_local: bool,
+
+    /// Defines the Max Execution time (in humantime eg : 2h45min) that the host runtime will execute for
+    #[clap(long = "max-execution-time", alias = "max-time", env = WASMCLOUD_MAX_EXECUTION_TIME_MS, default_value = DEFAULT_MAX_EXECUTION_TIME)]
+    pub max_execution_time: humantime::Duration,
```
brooksmtownsend, line 276: I'm a little worried about the inconsistency about using humantime here but standard Durations for the other args, since we've aimed to keep the environment variables the same w/ `wash` and `wasmCloud`. It's less fun, but I think a `u64` would suffice here for the execution time.

crates/wash-cli/src/up/config.rs
```diff
@@ -81,6 +83,10 @@ pub async fn configure_host_env(wasmcloud_opts: WasmcloudOpts) -> Result<HashMap
             cluster_issuers.join(","),
         );
     }
+    host_config.insert(
+        WASMCLOUD_MAX_EXECUTION_TIME_MS.to_string(),
+        wasmcloud_opts.max_execution_time.as_millis().to_string(),
+    );
```
brooksmtownsend, line 89: Once you make the above change, you will just need to .to_string() the u64

crates/host/src/wasmbus/mod.rs
```diff
@@ -810,7 +811,7 @@ impl Host {
             labels: RwLock::new(labels),
             ctl_nats,
             rpc_nats: Arc::new(rpc_nats),
-            host_config: config,
+            host_config: config.clone(),
```
brooksmtownsend, line 814: optional nit: if you assign `config.max_execution_time` to a variable above this `let host = Host {}` block, you won't need to clone here

brooksmtownsend (changes requested):
Hey @Aditya1404Sal I appreciate the back and forth here, my last request is just to standardize on `std::time::Duration` across this feature so we can stay as consistent as possible with the CLI and wasmCloud host. I could be convinced about the humantime argument for each flag in wash-cli, but I think it would be better to avoid that much of a change now

brooksmtownsend (approved):

PR #2440, opened 2024-07-04
bump rustls version

LUK3ARK:


LUK3ARK:
Trying to solve zeroize dependency issues, its actually tonic 😅

PR #2462, opened 2024-07-10
fix: remove wasm binaries from custom provider examples

ericgregory:
Removes wasm binaries from component subfolder of custom provider examples/templates and updates readmes accordingly.


rvolosatovs (approved):

vados-cosmonic (approved):
LGTM :rocket: 

I'm not sure we're calling out the need to build the individual component enough, but let's fix that in a follow up if people find it a bit too painful (or let's actually publish the component)

PR #2468, opened 2024-07-11
feat: Add support for supplying additional CA certificates to OCI and OpenTelemetry clients

### LINKED ISSUE #2460: [FEATURE] Support configuring additional CA certificates for OCI and OTEL clients
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

In environments where you may have non-WebPKI CA (i.e. self-hosted CA), you might need the ability to configure the wasmCloud Host to trust your custom CA without necessarily including it in the wasmCloud Host OS trust store (for example, if you're using/mirroring the upstream Docker images and don't wish to modify them to include your local CA).

It would be handy to have the ability to pass in a file path to load additional CA certificates from, possibly as a repeated option in case you needed multiple different CAs.

Ideally this would take the form of something like `--tls-ca-path` that could be repeated in case you needed to have multiple of these files loaded (as opposed to having them all be concatenated together in to a single file).

Optionally, this could also support passing in a path to directory instead of an individual file(s), which would then iterate all of the files in the directory (or perhaps filtering files based on a specific extension, `.crt` for example?) and attempt to parse them to certificates.

Parsing the provided argument to the set of files should be done as part of the host start process to ensure that the file(s) / directory of files is valid and failing fast with clear feedback if it's not.

_No response_

### Additional context

_No response_

joonas:
## Feature or Problem

This adds the support for providing additional CA certificates to be loaded at host initialization time, which are then configured in the OCI and OpenTelemetry clients.

It's intended for use inside of environments where there may be non-WebPKI certificates in use, such as an in-house CA.

## Related Issues

Fixes #2460

### DIFF
--- a/crates/core/src/otel.rs
+++ b/crates/core/src/otel.rs
@@ -2,7 +2,7 @@
 //!
 //! [otel]: https://opentelemetry.io
 
-use std::str::FromStr;
+use std::{path::PathBuf, str::FromStr};
 
 use anyhow::bail;
 use serde::{Deserialize, Serialize};
@@ -40,6 +40,9 @@ pub struct OtelConfig {
     /// Determines whether http or grpc will be used for exporting the telemetry.
     #[serde(default)]
     pub protocol: OtelProtocol,
+    /// Additional CAs to include in the OpenTelemetry client configuration
+    #[serde(default)]
+    pub additional_ca_paths: Vec<PathBuf>,
 }
 
 impl OtelConfig {
--- a/crates/core/src/tls.rs
+++ b/crates/core/src/tls.rs
@@ -20,8 +20,9 @@
 //! [webpki-roots]: https://crates.io/crates/webpki-roots
 //! [rustls-native-certs]: https://crates.io/crates/rustls-native-certs
 
-use std::sync::Arc;
+use std::{path::Path, sync::Arc};
 
+use anyhow::Context as _;
 use once_cell::sync::Lazy;
 
 #[cfg(feature = "rustls-native-certs")]
@@ -107,5 +108,37 @@ impl NativeRootsExt for reqwest::ClientBuilder {
             .iter()
             .cloned()
             .fold(self, reqwest::ClientBuilder::add_root_certificate)
+    }
+}
+
+/// Attempt to load certificates from a given array of paths
+pub fn load_certs_from_paths(
+    paths: &[impl AsRef<Path>],
+) -> anyhow::Result<Vec<rustls::pki_types::CertificateDer<'static>>> {
+    paths
+        .iter()
+        .map(read_certs_from_path)
+        .flat_map(|result| match result {
+            Ok(vec) => vec.into_iter().map(Ok).collect(),
+            Err(er) => vec![Err(er)],
+        })
+        .collect::<Result<Vec<_>, _>>()
+}
+
+/// Read certificates from a given path
+///
+/// At present this function only supports files -- directories will return an empty list
+pub fn read_certs_from_path(
+    path: impl AsRef<Path>,
+) -> anyhow::Result<Vec<rustls::pki_types::CertificateDer<'static>>> {
+    // TODO(joonas): Support directories
+    let path = path.as_ref();
+    if path.is_file() {
+        let f = std::fs::File::open(path)
+            .with_context(|| format!("failed to open file at provided path: {}", path.display()))?;
+        let mut reader = std::io::BufReader::new(f);
+        Ok(rustls_pemfile::certs(&mut reader).collect::<Result<Vec<_>, _>>()?)
+    } else {
+        Ok(vec![])
     }
 }
--- a/crates/host/src/oci.rs
+++ b/crates/host/src/oci.rs
@@ -26,6 +26,8 @@ const OCI_MEDIA_TYPE: &str = "application/vnd.oci.image.layer.v1.tar";
 /// Configuration options for OCI operations.
 #[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 pub struct Config {
+    /// Additional CAs to include in the OCI client configuration
+    pub additional_ca_paths: Vec<PathBuf>,
     /// Whether or not to allow downloading OCI artifacts with the tag `latest`
     pub allow_latest: bool,
     /// A list of OCI registries that are allowed to be accessed over HTTP
@@ -107,6 +109,7 @@ async fn cache_oci_image(
 /// OCI artifact fetcher
 #[derive(Clone, Debug)]
 pub struct Fetcher {
+    additional_ca_paths: Vec<PathBuf>,
     allow_latest: bool,
     allow_insecure: bool,
     auth: RegistryAuth,
@@ -115,6 +118,7 @@ pub struct Fetcher {
 impl Default for Fetcher {
     fn default() -> Self {
         Self {
+            additional_ca_paths: Vec::default(),
             allow_latest: false,
             allow_insecure: false,
             auth: RegistryAuth::Anonymous,
@@ -128,6 +132,7 @@ impl From<&RegistryConfig> for Fetcher {
             auth,
             allow_latest,
             allow_insecure,
+            additional_ca_paths,
             ..
         }: &RegistryConfig,
     ) -> Self {
@@ -135,6 +140,7 @@ impl From<&RegistryConfig> for Fetcher {
             auth: auth.into(),
             allow_latest: *allow_latest,
             allow_insecure: *allow_insecure,
+            additional_ca_paths: additional_ca_paths.clone(),
         }
     }
 }
@@ -145,6 +151,7 @@ impl From<RegistryConfig> for Fetcher {
             auth,
             allow_latest,
             allow_insecure,
+            additional_ca_paths,
             ..
         }: RegistryConfig,
     ) -> Self {
@@ -152,6 +159,7 @@ impl From<RegistryConfig> for Fetcher {
             auth: auth.into(),
             allow_latest,
             allow_insecure,
+            additional_ca_paths,
         }
     }
 }
@@ -179,9 +187,20 @@ impl Fetcher {
         } else {
             ClientProtocol::Https
         };
+        let mut certs = tls::NATIVE_ROOTS_OCI.to_vec();
+        if !self.additional_ca_paths.is_empty() {
+            certs.extend(
+                tls::load_certs_from_paths(&self.additional_ca_paths)?
+                    .iter()
+                    .map(|cert| oci_distribution::client::Certificate {
+                        encoding: oci_distribution::client::CertificateEncoding::Der,
+                        data: cert.to_vec(),
+                    }),
+            );
+        }
         let c = Client::new(ClientConfig {
             protocol,
-            extra_root_certificates: tls::NATIVE_ROOTS_OCI.to_vec(),
+            extra_root_certificates: certs,
             ..Default::default()
         });
 
@@ -259,5 +278,11 @@ impl Fetcher {
         par::read(&path, host_id, oci_ref)
             .await
             .with_context(|| format!("failed to read `{}`", path.display()))
+    }
+
+    /// Used to set additional CA paths that will be used as part of fetching components and providers
+    pub fn with_additional_ca_paths(mut self, additional_ca_paths: &Vec<PathBuf>) -> Self {
+        self.additional_ca_paths.clone_from(additional_ca_paths);
+        self
     }
 }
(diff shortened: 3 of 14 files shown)


crates/host/src/oci.rs
```diff
@@ -107,6 +109,7 @@ async fn cache_oci_image(
 /// OCI artifact fetcher
 #[derive(Clone, Debug)]
 pub struct Fetcher {
+    additional_ca_paths: Vec<std::path::PathBuf>,
```
protochron, line 112: Nit: can just be `PathBuf` here, right?

crates/core/src/tls.rs
```diff
@@ -109,3 +110,30 @@ impl NativeRootsExt for reqwest::ClientBuilder {
...
+        .map(read_certs_from_path)
+        .flat_map(|result| match result {
+            Ok(vec) => vec.into_iter().map(Ok).collect(),
+            Err(er) => vec![Err(er)],
+        })
+        .collect::<Result<Vec<_>, _>>()
+}
+
+pub fn read_certs_from_path(
+    path: &std::path::PathBuf,
```
vados-cosmonic, line 128:
```suggestion
    path: impl AsRef<Path>,
```

This requires some other changes but should be slightly more flexible this way

crates/core/src/tls.rs
```diff
@@ -109,3 +110,30 @@ impl NativeRootsExt for reqwest::ClientBuilder {
...
+        .iter()
+        .map(read_certs_from_path)
+        .flat_map(|result| match result {
+            Ok(vec) => vec.into_iter().map(Ok).collect(),
+            Err(er) => vec![Err(er)],
+        })
+        .collect::<Result<Vec<_>, _>>()
+}
+
+pub fn read_certs_from_path(
```
vados-cosmonic, line 127:
```suggestion
/// Read certificates from a given path
///
/// At present this function only supports files -- directories will return an empty list
pub fn read_certs_from_path(
```

crates/core/src/tls.rs
```diff
@@ -109,3 +110,30 @@ impl NativeRootsExt for reqwest::ClientBuilder {
...
+) -> anyhow::Result<Vec<rustls::pki_types::CertificateDer<'static>>> {
+    // TODO(joonas): Support directories
+    if path.is_file() {
+        let f = std::fs::File::open(path)
+            .with_context(|| format!("failed to open file at provided path: {}", path.display()))?;
+        let mut reader = std::io::BufReader::new(f);
+        Ok(rustls_pemfile::certs(&mut reader).collect::<Result<Vec<_>, _>>()?)
+    } else {
+        Ok(vec![])
+    }
```
vados-cosmonic, line 138:
```suggestion
    let path = path.as_ref();
    // TODO(joonas): Support directories
    if !path.is_file() {
        return Ok(Vec::with_capacity(0));
    }
    let f = std::fs::File::open(&path)
        .with_context(|| format!("failed to open file at provided path: {}", path.display()))?;
    let mut reader = std::io::BufReader::new(f);
    Ok(rustls_pemfile::certs(&mut reader).collect::<Result<Vec<_>, _>>()?)
```

*if* you want to pull in that `AsRef<Path>` change... 

Take it or leave it :)

crates/host/src/oci.rs
```diff
@@ -179,9 +187,20 @@ impl Fetcher {
         } else {
             ClientProtocol::Https
         };
+        let mut certs = tls::NATIVE_ROOTS_OCI.to_vec();
+        if !self.additional_ca_paths.is_empty() {
+            certs.extend(
+                tls::load_certs_from_paths(&self.additional_ca_paths)?
```
vados-cosmonic, line 193:
```suggestion
                tls::load_certs_from_paths(&self.additional_ca_paths).context("failed to load TLS certs from provided paths")?
```

crates/host/src/oci.rs
```diff
@@ -260,4 +279,10 @@ impl Fetcher {
             .await
             .with_context(|| format!("failed to read `{}`", path.display()))
     }
+
+    /// Used to set additional CA paths that will be used as part of fetching components and providers
+    pub fn with_additional_ca_paths(mut self, additional_ca_paths: &Vec<PathBuf>) -> Self {
```
vados-cosmonic, line 284:
```suggestion
    pub fn with_additional_ca_paths(mut self, additional_ca_paths: &[impl AsRef<Path>]) -> Self {
```

NIT: Requires some more changes but, might be nice to have the flexibility

crates/tracing/src/lib.rs
```diff
@@ -63,3 +63,27 @@ pub fn configure_observability(
         log_level_override,
     )
 }
+
+pub(crate) fn get_http_client(otel_config: &OtelConfig) -> anyhow::Result<reqwest_0_11::Client> {
```
vados-cosmonic, line 67: A comment here would be great on why we use the very specific version of reqwest

crates/tracing/src/metrics.rs
```diff
@@ -12,15 +12,22 @@ pub fn configure_metrics(
...
-            .http()
-            .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
-            .with_endpoint(otel_config.metrics_endpoint())
-            .into(),
-        OtelProtocol::Grpc => opentelemetry_otlp::new_exporter()
-            .tonic()
-            .with_endpoint(otel_config.metrics_endpoint())
-            .into(),
+        OtelProtocol::Http => {
+            let client = crate::get_http_client(otel_config)?;
```
vados-cosmonic, line 16: `context` might be nice here

src/main.rs
```diff
@@ -254,6 +255,10 @@ struct Args {
     /// Path to generate flame graph at
     #[clap(long = "flame-graph", env = "WASMCLOUD_FLAME_GRAPH")]
     flame_graph: Option<String>,
+
+    /// Configures the set of certificate authorities as repeatable set of file paths to load into the OCI and OpenTelemetry clients
+    #[arg(long = "tls-ca-path")]
```
vados-cosmonic, line 260: I can't remember, is this set of args prefixed when they're used from the outside? Since these CAs are specifically for telemetry it might be nice to prefix them to indicate that

src/main.rs
```diff
@@ -442,3 +453,13 @@ fn parse_label(labelpair: &str) -> anyhow::Result<(String, String)> {
         _ => bail!("invalid label format `{labelpair}`. Expected `key=value`"),
     }
 }
+
+fn parse_certs_for_paths(paths: Vec<PathBuf>) -> anyhow::Result<()> {
```
vados-cosmonic, line 457: Do we load certs just to make sure that they *can* be loaded and that there are some?

crates/host/src/oci.rs
joonas: I'm going to leave this one for someone else who has the patience with the borrow checker, I can't figure out how to make this work nicely. I'm happy to open up a follow-up issue for tracking it if this needs to be done.

src/main.rs
joonas: The thing is, these certs are used across OCI and OTEL clients, which is why I made the argument more generic.

Alternatively I could make two separate arguments, one for `--oci-tls-ca-path` and another for `–-(observability|otel)-tls-ca-path`, but I settled on this approach instead and made the argument repeatable in case you needed to pass in different set of CAs to load for the two different sets of clients.

src/main.rs
joonas: Correct, the idea is to provide feedback as early as possible when the user might be providing this flag to a non-existent certificate, or potentially badly formatted one.

Otherwise these certificates wouldn't get loaded until they try to pull down an OCI image and/or report OTEL telemetry, which could be potentially minutes to hours to days later, which is far from ideal, so rather than failing late and in an obscure way, I took the approach of failing fast and loudly on initialization.

src/main.rs
vados-cosmonic: Yeah whenI see `--tls-ca-path` I assume it basically will be used for *everything*, and I imagine a world where the OCI CA path is separate from the OTEL one.

Maybe this isn't a real problem though -- we can try to fix it/introduce more specific ones if someone finds this a problem? I error on the side of starting specific -- slight inconvenience for easier legibility, but maybe this is premature.

src/main.rs
vados-cosmonic: What do you think about renaming this to reflect that? `assert_...`  or `ensure_...` might be nice?

Take it or leave it -- otherwise I'm good on the clarifying question!

crates/host/src/oci.rs
vados-cosmonic: All good thank you for trying -- I'll do this!

src/main.rs
vados-cosmonic: resolving for :rocket: , take it or leave it!

crates/host/src/oci.rs
vados-cosmonic: resolving for :rocket: , this can be trivially added later

crates/host/src/oci.rs
joonas: Good call, yes it can and should be, thanks!

crates/core/src/otel.rs
```diff
@@ -32,6 +32,8 @@ pub struct OtelConfig {
     /// Determines whether http or grpc will be used for exporting the telemetry.
     #[serde(default)]
     pub protocol: OtelProtocol,
+    /// Additional CAs to include in the OpenTelemetry client configuration
+    pub additional_ca_paths: Vec<PathBuf>,
```
brooksmtownsend, line 36: Please mark this as `#[serde(default)]` for backwards compatibility in providers

crates/tracing/src/metrics.rs
```diff
@@ -12,15 +12,23 @@ pub fn configure_metrics(
...
+                .context("failed to get an http client for otel metrics exporter")?;
+            opentelemetry_otlp::new_exporter()
+                .http()
+                .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
+                .with_http_client(client)
+                .with_endpoint(otel_config.metrics_endpoint())
+                .into()
+        }
+        OtelProtocol::Grpc => {
+            // TODO(joonas): Configure tonic::transport::ClientTlsConfig via .with_tls_config(...), passing in additional certificates.
```
brooksmtownsend, line 26: Does this need to be done as a part of this PR to support additional CA certs for Grpc?

crates/tracing/src/traces.rs
```diff
@@ -189,15 +189,23 @@ where
...
+                .context("failed to get an http client for otel tracing exporter")?;
+            opentelemetry_otlp::new_exporter()
+                .http()
+                .with_endpoint(otel_config.traces_endpoint())
+                .with_http_client(client)
+                .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
+                .into()
+        }
+        OtelProtocol::Grpc => {
+            // TODO(joonas): Configure tonic::transport::ClientTlsConfig via .with_tls_config(...), passing in additional certificates.
```
brooksmtownsend, line 203: Same comment re: Grpc

crates/tracing/src/traces.rs
```diff
@@ -230,15 +238,23 @@ where
...
+                .context("failed to get an http client for otel logging exporter")?;
+            opentelemetry_otlp::new_exporter()
+                .http()
+                .with_endpoint(otel_config.logs_endpoint())
+                .with_http_client(client)
+                .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
+                .into()
+        }
+        OtelProtocol::Grpc => {
+            // TODO(joonas): Configure tonic::transport::ClientTlsConfig via .with_tls_config(...), passing in additional certificates.
```
brooksmtownsend, line 252: Same comment re: Grpc

brooksmtownsend (changes requested):
One request, and one set of questions that aren't blockers if they aren't needed

crates/core/src/otel.rs
joonas: Good catch, thanks!

crates/tracing/src/metrics.rs
joonas: I've opted to leave this for a fast follow-up PR that focuses on the tonic story specifically since I'm not aware of any usage of this functionality

brooksmtownsend (approved):

crates/tracing/src/traces.rs
joonas: TL;DR from previous comment: leaving this as a fast follow-up PR to address Tonic specifically

crates/tracing/src/traces.rs
joonas: TL;DR from previous comment: leaving this as a fast follow-up PR to address Tonic specifically

PR #2459, opened 2024-07-09
feat: upgrade `wrpc`, `async-nats`, `wasmtime`

rvolosatovs:
## Feature or Problem

This PR is *massive* - but unfortunately, I don't really see a simple way forward here, which would minimize the change set due a few technical debt issues in existing design that are now fixed and the fact that pre-MVP wRPC version was adopted in various areas of wasmCloud.

- Update to latest wRPC version
   - update `wasmtime` to enable that (wasmtime 22)
   - update `async-nats` to enable that (async-nats 0.35.1)
       - switch to `async-nats-wrpc`, until async functionality is exposed by the crate https://github.com/nats-io/nats.rs/pull/1267#issuecomment-2209422981 fork at https://github.com/wrpc/nats.rs
- Migrate blobstore providers to use `wit-bindgen-wrpc`, which now supports async
- Add support for custom resources by virtue of upgrading wRPC (disabled in the runtime currently, but it's a quick additive change once this massive PR is merged)
- Change link and target storage to associate *instances* (be it an interface, root, nested interface etc) to targets
- Minimize cloning and heap allocations of long-lived strings (like lattice and component IDs) on the hot path by using `Arc<str>` instead of `String`
- Remove `Handler` abstraction and use wRPC functionality directly in the runtime
   - Pass `ReplacedInstanceTarget` through the context to the host to deduplicate native wRPC instance call from a replaced one
- Respect `max_instances` for components by using a semaphore and a task set.
- Avoid compiling components on fetch, only do so if required for fresh start (otherwise, clone the existing pre-compiled one)
- Remove the wasmcloud core wRPC client - the `wrpc_transport_nats` client is now used directly, with support for passing headers as context per invocation and native `.timeout` available in `InvokeExt` trait https://bytecodealliance.github.io/wrpc/wrpc/transport/trait.InvokeExt.html#method.timeout

## Related Issues

closes #2433 
closes https://github.com/wasmCloud/wasmCloud/pull/2432

## Consumer Impact

There are breaking changes in wRPC protocol! Namely:
- encoding of non-ASCII `char` values (strings are fine, only concerns "raw" char types)
- encoding of flag values for types containing over 7 cases, encoding the bits 8 and further

This should be transparent for most users

### Manual Verification

`cargo test` mostly and the CI bingo

PR #2478, opened 2024-07-12
chore: update wRPC

rvolosatovs:
## Feature or Problem

small follow-up on #2459 , I did not want to touch that PR after being reviewed

- instrument the component export handlers
- do not fail if component exports something we do not support (like, another component or a resource), just log a warning
- check if a component exports resources, and if so, warn that those will be ignored
- propagate the invocation timeout into the polyfill

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,6 +405,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c59bdb34bc650a32731b31bd8f0829cc15d24a708ee31559e0bb34f2bc320cba"
+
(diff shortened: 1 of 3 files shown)


rvolosatovs:
FWIW, if we ever wanted to support components exporting resources, here's how we could do it https://github.com/bytecodealliance/wrpc/blob/b55c7f2c41eb78bc6560cd8889d22c1590555f42/crates/wasmtime-nats-cli/src/lib.rs#L291-L606

Basically, do the normal, stateless operation if a component does not export resources or use a single shared store and instance if it does

brooksmtownsend (approved):
Looks good to me 👍🏻 I was on the fence about allowing components to instantiate even if the exports are going to be ignored, but I think bailing on unsupported imports is the important piece and in this scenario a component export just won't be called.

PR #2498, opened 2024-07-15
ci: add explicit content read allowance

brooksmtownsend:

## Feature or Problem
This PR fixes an issue where even if you specify top-level permissions for a workflow, you must specify all appropriate permissions for a workflow where you add a `permissions` block. It doesn't seem to make sense that it works this way, but the wasmCloud workflow was not running when only `packages: write` was specified.

### DIFF
--- a/.github/workflows/comment-cc.yml
+++ b/.github/workflows/comment-cc.yml
@@ -21,6 +21,7 @@ jobs:
   comment:
     runs-on: ubuntu-latest
     permissions:
+      contents: read
       pull-requests: write
     steps:
       - name: Explain conventional commits in comment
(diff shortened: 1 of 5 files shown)


brooksmtownsend:
In fact https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions specifies 
>You can use permissions either as a top-level key, to apply to all jobs in the workflow, or within specific jobs. When you add the permissions key within a specific job, all actions and run commands within that job that use the GITHUB_TOKEN gain the access rights you specify.

I assumed this would be additive, not overwriting, but I suppose if not specifying `contents` == specifying `contents: none` it makes sense.

thomastaylor312 (approved):

PR #2511, opened 2024-07-17
feat(ci): add action for mirroring images

vados-cosmonic:
## Feature or Problem

This commit adds a CI action that can be used to mirror images that are used in tests or common workflows to our local GHCR.

This can be use to avoid rate-limits when accessing docker.io for a wide variety of images.


.github/workflows/mirror-docker-image.yml
```diff
@@ -0,0 +1,37 @@
...
+      - uses: oras-project/setup-oras@ca28077386065e263c03428f4ae0c09024817c93
+      - name: Docker login
+        shell: bash
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | oras login ghcr.io --username ${{ github.repository_owner }} --password-stdin
+      - name: Copy image
+        shell: bash
+        run: |
+          oras cp \
+          ${{ inputs.image }}${{ inputs.version }} \
```
joonas, line 36: should this be:
```suggestion
          ${{ inputs.image }}:${{ inputs.version }} \
```

.github/workflows/mirror-docker-image.yml
```diff
@@ -0,0 +1,37 @@
...
+          ghcr.io/${{ github.repository_owner }}/mirrors/${{ inputs.output-image }}${{ inputs.version }}
```
joonas, line 37: You will need the lowercased version of `github.repository_owner`: https://github.com/wasmCloud/wasmCloud/blob/7e26ac135ff6e6f9678f21e44e9631734311c264/.github/workflows/wasmcloud-chart.yml#L81-L83

OCI doesn't support uppercase letters in the identifier.

.github/workflows/mirror-docker-image.yml
joonas: Oh I see, the expectation is for folks to remember to add `:`

I think it would be worth adding a check to validate that the `input.version` starts with `:` or `@` and error the action out if it doesn't to make it obvious when you might've have accidentally it out.

.github/workflows/mirror-docker-image.yml
```diff
@@ -0,0 +1,37 @@
...
+  mirror-docker-image:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - uses: oras-project/setup-oras@ca28077386065e263c03428f4ae0c09024817c93
+      - name: Docker login
+        shell: bash
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | oras login ghcr.io --username ${{ github.repository_owner }} --password-stdin
```
joonas, line 31: We should have a step to also run `oras logout` at the end of the workflow since this isn't a pre-built action that handles post-action clean up.

.github/workflows/mirror-docker-image.yml
vados-cosmonic: Yeah that's pretty reasonable, I can add a step to validate that they're using the right prefix there!

- [ ] Add check of prefix for version

.github/workflows/mirror-docker-image.yml
vados-cosmonic: Ahh thanks -- will add this in!

- [ ] Add in lowercasing of repo owner

.github/workflows/mirror-docker-image.yml
```diff
@@ -0,0 +1,37 @@
...
+
+permissions: {}
+
+jobs:
+  mirror-docker-image:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - uses: oras-project/setup-oras@ca28077386065e263c03428f4ae0c09024817c93
```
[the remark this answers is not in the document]
vados-cosmonic: Sure -- I don't think we need to actually update this manually (we can leave it to dependabot), but will add!

joonas (approved):

brooksmtownsend (approved):

PR #2518, opened 2024-07-18
refactor(ci): only run provider builds on tags

vados-cosmonic:
## Feature or Problem

This PR ensures providers are built and published only when a tag is pushed that intentionally releases a given provider.

We should still catch breakage of provider crates during normal build, but now we can avoid building them and trying to publish them (or not) on every CI run.

This should prevent some spurrious failures we see on CI on `main` and in dependabot PRs (`download-artifact` is surprisingly unreliable)

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -291,6 +291,7 @@ jobs:
           path: doc
 
   providers:
+    if: startswith(github.ref, 'refs/tags/provider-')
     strategy:
       matrix:
         include:


.github/workflows/wasmcloud.yml
```diff
@@ -291,6 +291,7 @@ jobs:
           path: doc
 
   providers:
+    if: startswith(github.ref, 'refs/tags/provider-')
```
markkovari, line 294: Hey @vados-cosmonic, I have seen the `startsWith` expression (or function) with lowercase `startswith`. The documentation is using `startsWith` pretty much everywhere. Since this is fairly not testable thing (other than eye-test). Should we rather go with? Did not found any sign that this would be case insensitive in the [docs](https://docs.github.com/en/actions/learn-github-actions/expressions#startswith).

```suggestion
    if: startsWith(github.ref, 'refs/tags/provider-')
```

markkovari (commented):
This *might* save us from a re-release. Unfortunately I have bad relationship with GitHub Actions especially their testability.

.github/workflows/wasmcloud.yml
markkovari: Yikes, never mind it is all over the place here. 😬 🤦 At sorry for my triggered ptsd. At least these releases does not take hours.

vados-cosmonic: Absolutely no problem @markkovari thanks for taking a close look -- yeah it *seems* like Github is OK with doing it either way, probably a ease-of-use feature. 

When I tested this on my own fork, it seemed to work fine. We could absolutely make it consistent across our codebase though, and avoid triggering more PTSD! :)

markkovari: As long as you don't write "too much" bash in the steps and your visualization of the pipeline does not look like a circuit board I think you are good.

markkovari (approved):

brooksmtownsend (approved):
Nice!

What may be nice in the future is to also let this run on a manual build, so we could say "dang there's a problem with this provider, or we should test this provider before we tag it" etc etc. Can be done in the future of course, we haven't hit that use case yet I'm just talking out loud

vados-cosmonic:
Oh yeah that's a super reasonable idea @brooksmtownsend -- so we can release a provider/do a dry run

PR #2499, opened 2024-07-15
feat(wash-cli): add secrets subcommand for managing secrets references

protochron:
## Feature or Problem
This adds a set of commands for manually managing secrets references in a wasmCloud deployment. It is effectively a wrapper around configuration that properly formats and writes the data type to the config bucket.

It also attempts to prevent you from interacting with secrets with the config subcommands by checking the name of the configuration key and redirecting you to the correct subcommand.

## Related Issues
Closes #2494.

## Release Information
next

### Manual Verification
Tested locally with a wash build

PR #2523, opened 2024-07-19
chore: update secrets integration to use the updated config structure

protochron:
## Feature or Problem
Update the secrets integration in a wasmCloud host to include information about the policy that determines which backend to communicate with. This is a change that comes in from wadm where the policy block now contains the information about which backend to use.

This also passes any propertes defined on the policy to the correct backend, which are stored as a versioned string-encoded JSON object.

## Release Information
next

### DIFF
--- a/crates/host/src/secrets.rs
+++ b/crates/host/src/secrets.rs
@@ -26,6 +26,8 @@ pub(crate) struct SecretReference {
     /// The version of the secret to retrieve. If not supplied, the latest version will be used.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
+    /// The policy that defines configuration options for the backend.
+    pub policy_properties: String,
 }
 
 #[derive(Debug)]
@@ -148,13 +150,17 @@ impl Manager {
                 let secrets_client = self
                     .get_or_create_secrets_client(&secret_ref.backend)
                     .await?;
+                let application = Application{
+                    name: application.cloned().unwrap_or_default(),
+                    policy: secret_ref.policy_properties.clone(),
+                };
                 let request = SecretRequest {
                     name: secret_ref.key.clone(),
                     version: secret_ref.version.clone(),
                     context: Context {
                         entity_jwt: entity_jwt.to_string(),
                         host_jwt: host_jwt.to_string(),
-                        application: application.cloned().map(|name| Application { name }),
+                        application,
                     },
                 };
                 secrets_client
--- a/crates/secrets-types/src/lib.rs
+++ b/crates/secrets-types/src/lib.rs
@@ -34,13 +34,21 @@ pub struct Context {
     pub host_jwt: String,
     /// The application the entity belongs to.
     /// TODO: should this also be a JWT, but signed by the host?
-    pub application: Option<Application>,
+    pub application: Application,
 }
 
 /// The application that the entity belongs to.
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Default)]
 pub struct Application {
+    /// The name of the application.
+    #[serde(default)]
     pub name: String,
+
+    /// The policy used define the application's access to secrets.
+    /// This meant to be a JSON string that can be deserialized by a secrets backend
+    /// implementation.
+    #[serde(default)]
+    pub policy: String,
 }
 
 impl Context {
--- a/crates/test-util/src/lattice/config.rs
+++ b/crates/test-util/src/lattice/config.rs
@@ -30,8 +30,12 @@ pub async fn assert_put_secret_reference(
     version: Option<String>,
 ) -> Result<()> {
     let mut config = HashMap::from([
-        ("key".to_string(), key.to_string()),
         ("backend".to_string(), backend.to_string()),
+        ("key".to_string(), key.to_string()),
+        (
+            "policy_properties".to_string(),
+            serde_json::json!({"type": "policy.secrets.wasmcloud.dev", "properties": {"configuration": "value"}}).to_string(),
+        ),
     ]);
     if let Some(version) = version {
         config.insert("version".to_string(), version.to_string());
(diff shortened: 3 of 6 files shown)


crates/host/src/secrets.rs
```diff
@@ -16,13 +16,21 @@ use wasmcloud_secrets_types::{Application, Context, Secret as WasmcloudSecret, S
...
+    /// The policy that defines configuration options for the backend.
+    pub policy: String,
     /// The key to use for retrieving the secret from the backend.
     pub key: String,
     /// The version of the secret to retrieve. If not supplied, the latest version will be used.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
+    /// The policy that defines configuration options for the backend.
+    pub policy_properties: String,
+    //pub backend: String,
```
brooksmtownsend, line 28: nit: looks leftover from the previous structure

crates/host/src/secrets.rs
```diff
@@ -146,16 +154,22 @@ impl Manager {
             })
             // Retrieve the actual secret from the secrets backend
             .and_then(|secret_ref| async move {
+                let props = serde_json::from_str::<SecretReferenceProperties>(&secret_ref.policy_properties)
+                    .map_err(|_| anyhow::anyhow!("failed to deserialize secret reference properties"))?;
+                let properties = match props.properties.get("backend") {
+                    None => bail!("secret reference properties did not contain a backend key"),
+                    Some(properties) => properties,
+                };
```
brooksmtownsend, line 162:
```suggestion
                let properties = match props.properties.get("backend").context("secret reference properties did not contain a backend key")?;
```

nit

crates/secrets-types/secrets.proto
brooksmtownsend, line 1: Is this a hint that we should be using protos 😉

brooksmtownsend (changes requested):
Just a request to remove the proto, but the other things are nits and this looks good 👍🏻 

We'll need to update the secrets example to use this structure, and your secret command in https://github.com/wasmCloud/wasmCloud/pull/2499, so we can do that as a followup

crates/secrets-types/secrets.proto
protochron: hah, whoops! I was experimenting with a way to have an IDL representation for the JSON data we're passing around. Ideally it would all be WIT once we can write secrets backends entirely in wasm, but today it could be difficult if we're just providing a loose spec.

crates/host/src/secrets.rs
```diff
@@ -146,16 +153,19 @@ impl Manager {
             })
             // Retrieve the actual secret from the secrets backend
             .and_then(|secret_ref| async move {
+                let props = serde_json::from_str::<SecretReferenceProperties>(&secret_ref.policy_properties)
+                    .map_err(|_| anyhow::anyhow!("failed to deserialize secret reference properties"))?;
+                let properties = props.properties.get("backend").context("secret reference properties did not contain a backend key")?;
```
joonas, line 158:
```suggestion
                let backend = props.properties.get("backend").context("secret reference properties did not contain a backend key")?;
```

joonas, line 158: just a naming suggestion:
```suggestion
                let backend = props.properties.get("backend").context("secret reference properties did not contain a backend key")?;
```

crates/host/src/secrets.rs
```diff
@@ -146,16 +153,19 @@ impl Manager {
...
                 let secrets_client = self
-                    .get_or_create_secrets_client(&secret_ref.backend)
+                    .get_or_create_secrets_client(properties)
```
joonas, line 160: to match the above suggestion:
```suggestion
                    .get_or_create_secrets_client(backend)
```

protochron:
@joonas went ahead and made the `Application` struct required in a `Context`

crates/secrets-types/src/lib.rs
```diff
@@ -34,13 +34,21 @@ pub struct Context {
...
+    pub application: Application,
 }
 
 /// The application that the entity belongs to.
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Default)]
 pub struct Application {
+    /// The name of the application.
+    #[serde(default)]
     pub name: String,
```
brooksmtownsend, line 45: I was going to leave a comment about making this optional again, but as long as we're okay with thinking that an empty app name == no application this works fine. It feels weird to make `name` optional but `policy` not

crates/test-util/src/lattice/config.rs
```diff
@@ -30,8 +30,13 @@ pub async fn assert_put_secret_reference(
     version: Option<String>,
 ) -> Result<()> {
     let mut config = HashMap::from([
-        ("key".to_string(), key.to_string()),
         ("backend".to_string(), backend.to_string()),
+        ("key".to_string(), key.to_string()),
+        ("policy".to_string(), backend.to_string()),
```
brooksmtownsend, line 35: Do we need this `policy` key anymore?

crates/host/src/secrets.rs
```diff
@@ -26,6 +26,13 @@ pub(crate) struct SecretReference {
     /// The version of the secret to retrieve. If not supplied, the latest version will be used.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
+    /// The policy that defines configuration options for the backend.
+    pub policy_properties: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+struct SecretReferenceProperties {
+    properties: HashMap<String, String>,
```
brooksmtownsend, line 35: Do we need this struct anymore? Since we're just passing the policy_properties through

crates/test-util/src/lattice/config.rs
protochron: We don't

crates/host/src/secrets.rs
protochron: Nope, left over from what I was doing before.

brooksmtownsend (approved):

Issue #2032, opened 2024-04-30
[FEATURE] Support downloading patch versions of wasmCloud and wadm for `wash up`


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

It's frustrating to open PRs like #2031, and for users to have to download entirely new wash versions just to get patch fixes to wasmCloud and wadm.

### Describe the solution you'd like

When running `wash up`, if `--wasmcloud-start-only` isn't supplied, we should check for the latest version of wasmCloud and wadm to see if we can download a newer SEMVER patch version of one or both. This would make it much easier for us to push out patches to `wadm` and `wasmcloud` within wash.

This should fail gracefully if not successful, e.g. not having an internet connection shouldn't make `wash up` fail as long as an older version is already downloaded.

### Describe alternatives you've considered

One thing to consider is that it puts the onus of following SEMVER (which we do) on the maintainers, and in the future this _may_ end up breaking someone using wash up if we need to push a security breaking change as a patch. Given `wash up` is a developer tool, this is an acceptable compromise.

brooksmtownsend:
Thanks @vados-cosmonic , this is definitely a great one to have.

markkovari:
Hey @brooksmtownsend this seems like a good enhancement to me, feel free to assign me if no objection, already started to look into it. 🙇

markkovari:
@brooksmtownsend (or anyone else) which versions are we interested in as patches? 

I assume that we are interested in pre-releases too, but no drafts. That should be the case, right? Or do you think we should ignore prereleases?

Created the WIP pr, but there is a handful of missing things. #2527 

It is already downloading the newer versions, but no fallback to the previous versions/binaries yet and things are a bit messy. 🫣

brooksmtownsend:
Awesome work @markkovari ! I think we should only automatically update patch versions that are not pre-releases or drafts. So for example, if we release `wash` with version wasmCloud 1.0.4:
1. 1.0.5 should auto-download once released
2. 1.0.6 should auto-download once released
3. 1.0.7-beta.1 should _not_ auto-download
4. 1.0.8-rc.1 should _not_ auto-download
[...]

PR #2527, opened 2024-07-22
feat(wash-cli): check for wadm and wasmcloud patch version updates

markkovari:
## Feature or Problem

Wouldn't it been nice if new crispy wadm and wasmcloud would be handy with new patch versions? Well that PR supposed to implement this.

## Related Issues

Fixes: #2032

## Consumer Impact

Users using `wash up` should download new version of the mentioned binaries.

### DIFF
--- a/.github/workflows/examples-providers.yml
+++ b/.github/workflows/examples-providers.yml
@@ -134,6 +134,8 @@ jobs:
       - name: test provider load
         if: ${{ matrix.project.test_deploy && matrix.wash-version == 'current' }}
         shell: bash
+        env:
+          RUST_LOG: trace
         working-directory: examples/${{ matrix.project.lang }}/providers/${{ matrix.project.name }}
         run: |
           set -xe
--- a/.github/workflows/examples-providers.yml
+++ b/.github/workflows/examples-providers.yml
@@ -134,8 +134,6 @@ jobs:
       - name: test provider load
         if: ${{ matrix.project.test_deploy && matrix.wash-version == 'current' }}
         shell: bash
-        env:
-          RUST_LOG: trace
         working-directory: examples/${{ matrix.project.lang }}/providers/${{ matrix.project.name }}
         run: |
           set -xe
--- a/crates/wash-cli/src/cmd/dev/session.rs
+++ b/crates/wash-cli/src/cmd/dev/session.rs
@@ -22,7 +22,7 @@ use wash_lib::start::{
 };
 
 use crate::cmd::up::{remove_wadm_pidfile, start_nats, NatsOpts, WadmOpts, WasmcloudOpts};
-use crate::config::{configure_host_env, DEFAULT_NATS_HOST};
+use crate::config::{configure_host_env, DEFAULT_NATS_HOST, WADM_VERSION, WASMCLOUD_HOST_VERSION};
 use crate::down::stop_nats;
 
 use super::{dev_dir, sessions_file_path, SESSIONS_FILE_VERSION, SESSION_ID_LEN};
@@ -303,7 +303,11 @@ impl WashDevSession {
                     wadm_log_path.display()
                 )
             })?;
-        let wadm_binary = ensure_wadm(&wadm_opts.wadm_version, &install_dir).await?;
+        let wadm_binary = ensure_wadm(
+            &wadm_opts.wadm_version.unwrap_or(WADM_VERSION.to_owned()),
+            &install_dir,
+        )
+        .await?;
         let wadm_child = match start_wadm(
             &install_dir,
             &wadm_binary,
@@ -318,9 +322,13 @@ impl WashDevSession {
         };
 
         // Start the host in detached mode, w/ custom log file
+        let wasmcloud_version = wasmcloud_opts
+            .wasmcloud_version
+            .as_ref()
+            .map(|v| v.to_string())
+            .unwrap_or_else(|| WASMCLOUD_HOST_VERSION.into());
         let wasmcloud_log_path = session_dir.join("wasmcloud.log");
-        let wasmcloud_binary =
-            ensure_wasmcloud(&wasmcloud_opts.wasmcloud_version, &install_dir).await?;
+        let wasmcloud_binary = ensure_wasmcloud(&wasmcloud_version, &install_dir).await?;
         let log_output: Stdio = tokio::fs::File::create(&wasmcloud_log_path)
             .await
             .with_context(|| {
--- a/crates/wash-cli/src/cmd/dev/session.rs
+++ b/crates/wash-cli/src/cmd/dev/session.rs
@@ -303,11 +303,12 @@ impl WashDevSession {
                     wadm_log_path.display()
                 )
             })?;
-        let wadm_binary = ensure_wadm(
-            &wadm_opts.wadm_version.unwrap_or(WADM_VERSION.to_owned()),
-            &install_dir,
-        )
-        .await?;
+        let wadm_version = wadm_opts
+            .wadm_version
+            .as_ref()
+            .map(|v| v.to_string())
+            .unwrap_or_else(|| WADM_VERSION.into());
+        let wadm_binary = ensure_wadm(&wadm_version, &install_dir).await?;
         let wadm_child = match start_wadm(
             &install_dir,
             &wadm_binary,
--- a/crates/wash-cli/src/cmd/dev/session.rs
+++ b/crates/wash-cli/src/cmd/dev/session.rs
@@ -305,8 +305,6 @@ impl WashDevSession {
             })?;
         let wadm_version = wadm_opts
             .wadm_version
-            .as_ref()
-            .map(|v| v.to_string())
             .unwrap_or_else(|| WADM_VERSION.into());
         let wadm_binary = ensure_wadm(&wadm_version, &install_dir).await?;
         let wadm_child = match start_wadm(
@@ -324,9 +322,8 @@ impl WashDevSession {
 
         // Start the host in detached mode, w/ custom log file
         let wasmcloud_version = wasmcloud_opts
+            .clone()
             .wasmcloud_version
-            .as_ref()
-            .map(|v| v.to_string())
             .unwrap_or_else(|| WASMCLOUD_HOST_VERSION.into());
         let wasmcloud_log_path = session_dir.join("wasmcloud.log");
         let wasmcloud_binary = ensure_wasmcloud(&wasmcloud_version, &install_dir).await?;
--- a/crates/wash-cli/src/cmd/up.rs
+++ b/crates/wash-cli/src/cmd/up.rs
@@ -29,8 +29,9 @@ use wash_lib::context::ContextManager;
 use wash_lib::generate::emoji;
 use wash_lib::start::{
     ensure_nats_server, ensure_wadm, ensure_wasmcloud, find_wasmcloud_binary, nats_pid_path,
-    start_nats_server, start_wadm, start_wasmcloud_host, NatsConfig, WadmConfig,
-    NATS_SERVER_BINARY, NATS_SERVER_CONF, WADM_PID,
+    new_patch_version_of_after_string, start_nats_server, start_wadm, start_wasmcloud_host,
+    NatsConfig, WadmConfig, GITHUB_WASMCLOUD_ORG, GITHUB_WASMCLOUD_WADM_REPO,
+    GITHUB_WASMCLOUD_WASMCLOUD_REPO, NATS_SERVER_BINARY, NATS_SERVER_CONF, WADM_PID,
 };
 use wasmcloud_control_interface::{Client as CtlClient, ClientBuilder as CtlClientBuilder};
 
@@ -153,8 +154,8 @@ impl From<NatsOpts> for NatsConfig {
 #[derive(Parser, Debug, Clone)]
 pub struct WasmcloudOpts {
     /// wasmCloud host version to download, e.g. `v0.55.0`. See https://github.com/wasmCloud/wasmcloud/releases for releases
-    #[clap(long = "wasmcloud-version", default_value = WASMCLOUD_HOST_VERSION, env = "WASMCLOUD_VERSION")]
-    pub wasmcloud_version: String,
+    #[clap(long = "wasmcloud-version", env = "WASMCLOUD_VERSION")]
+    pub wasmcloud_version: Option<String>,
 
     /// A unique identifier for a lattice, frequently used within NATS topics to isolate messages among different lattices
     #[clap(
@@ -362,8 +363,8 @@ impl WasmcloudOpts {
 #[derive(Parser, Debug, Clone)]
 pub struct WadmOpts {
     /// wadm version to download, e.g. `v0.4.0`. See https://github.com/wasmCloud/wadm/releases for releases
-    #[clap(long = "wadm-version", default_value = WADM_VERSION, env = "WADM_VERSION")]
-    pub wadm_version: String,
+    #[clap(long = "wadm-version", env = "WADM_VERSION")]
+    pub wadm_version: Option<String>,
 
     /// If enabled, wadm will not be downloaded or run as a part of the up command
     #[clap(long = "disable-wadm")]
@@ -422,6 +423,7 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
             .cluster_seed
             .or_else(|| ctx.cluster_seed.map(|seed| seed.to_string())),
         wasmcloud_js_domain: cmd.wasmcloud_opts.wasmcloud_js_domain.or(ctx.js_domain),
+        wasmcloud_version: cmd.wasmcloud_opts.wasmcloud_version.clone(),
         ..cmd.wasmcloud_opts
     };
     let host_env = configure_host_env(wasmcloud_opts.clone()).await?;
@@ -563,7 +565,8 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
             .into_std()
             .await;
 
-        let wadm_path = ensure_wadm(&cmd.wadm_opts.wadm_version, &install_dir).await;
+        let wadm_path =
+            install_patch_or_default_wadm_version(&cmd.wadm_opts.wadm_version, &install_dir).await;
         match wadm_path {
             Ok(wadm_bin_path) => {
                 let wadm_child = start_wadm(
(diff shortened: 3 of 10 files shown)


crates/wash-cli/src/up/mod.rs
```diff
@@ -537,7 +540,24 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
     // Download wasmCloud if not already installed
     let wasmcloud_executable = if !wasmcloud_opts.start_only {
         spinner.update_spinner_message(" Downloading wasmCloud ...".to_string());
-        ensure_wasmcloud(&wasmcloud_opts.wasmcloud_version, &install_dir).await?
+        let new_wasmcloud_patch_version = new_patch_version_of_after_string(
```
markkovari, line 543: Another question @brooksmtownsend 
you said if someone is "specifying the version of wasmcloud". Currently there is only one string in the [WasmCloudOpts](https://github.com/wasmCloud/wasmCloud/blob/e2f6b61f7850c0dd51af1b9d582f44ab8ead633e/crates/wash-cli/src/up/mod.rs#L140-L141), which - if specified - will be overwritten.

So how should we know if someone used the `--wasmcloud-version`? How to prevent the 'patching' if a specific version was added 🤔 ?

markkovari (commented):
Well things are definitely on a good track, but I have a few questions. Appreciate 🙇

crates/wash-cli/src/up/mod.rs
markkovari: Sorry the questions was: can I mess up the current [WasmCloudOpts](https://github.com/wasmCloud/wasmCloud/blob/e2f6b61f7850c0dd51af1b9d582f44ab8ead633e/crates/wash-cli/src/up/mod.rs#L140-L141)?

crates/wash-cli/tests/wash_up.rs
```diff
@@ -275,6 +275,62 @@ async fn integration_up_works_with_labels() -> Result<()> {
...
+
+    // Get host data, ensure we find the host with the right label
+    let cmd_output = instance.get_hosts().await.context("failed to call hosts")?;
+
+    assert!(cmd_output.success, "call command succeeded");
+    let host = cmd_output.hosts.first();
+    assert!(host.is_some(), "host is present");
+
+    //TODO:  how to test the wadm version?
+    Ok(())
```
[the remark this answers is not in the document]
brooksmtownsend: I think elsewhere in the code we check the wadm version by running the binary itself, then parsing the version from that output 🤔

crates/wash-cli/src/up/mod.rs
markkovari: Yes, thank you. I added the `Option<String>` already for the necessary places, but there are some integration tests, which are failing. So it requires some massage still. Thanks for the detailed answer. 🎉

brooksmtownsend:
Hey @markkovari I just wanted to check in on the status of this one, do you need any assistance with the implementation details of `wash`? No rush on getting this done, I just wanted to make sure it didn't go stale. 🙂

markkovari:
Hey @brooksmtownsend , was on a long vacation. Picking it up again, please be prepared that I will bother you 😉 . Thanks for the ping.

markkovari:
Hey @brooksmtownsend, sorry I was a bit overwhelmed with tasks and made little progress with this. After this week will pick this one up (but probably will rest even a little too). 


One question if we already have a binary in the downloads folder should we overwrite it or what should be the strategy if this is not an initial `up`? 

Maybe: we should "try" to fetch a new version of the binary and if we did download it successfully, then replace the already existing? Does it sound reasonable?

brooksmtownsend:
> One question if we already have a binary in the downloads folder should we overwrite it or what should be the strategy if this is not an initial up?

Good question @markkovari ! I think what would be best is to follow the current strategy of creating `~/.wash/downloads/vX.Y.Z/wasmcloud_host`, so what we can do is make sure that we create a new directory in the case where there's a new binary. If we happen to have one already in that directory, then I think we can proceed and assume the host is already downloaded. Thoughts? I'm concerned if we always fetch then that slows down the `wash up` process significantly

markkovari:
Sounds good to me, I pretty much will start to work on it again in the middle of next week. Clear(er) what should happen. Thanks

crates/wash-cli/src/cmd/up.rs
```diff
@@ -914,6 +929,40 @@ pub(crate) async fn start_nats(
...
+        return ensure_wadm(version, install_dir).await;
+    }
+    const WASMCLOUD_ORG: &str = "wasmCloud";
+    const WASMCLOUD_WADM_REPO: &str = "wadm";
+    let version = version.clone().unwrap_or(WADM_VERSION.to_owned());
+    let new_patch_version =
+        new_patch_version_of_after_string(WASMCLOUD_ORG, WASMCLOUD_WADM_REPO, &version).await?;
+    match new_patch_version {
+        Some(new_patch) => {
+            let new_version = format!("v{}", new_patch);
```
brooksmtownsend, line 948: Is the new patch version returned by `new_patch_version`... always the semver format without the `v`?

crates/wash-cli/tests/common/mod.rs
```diff
@@ -228,12 +229,27 @@ impl Drop for TestWashInstance {
...
+            // to the current release, to avoid undefined behaviour in tests
+            extra_args: vec![
+                "--wadm-version".to_string(),
+                WADM_VERSION.to_string(),
+                "--wasmcloud-version".to_string(),
+                WASMCLOUD_HOST_VERSION.to_string(),
+            ],
+        }
+    }
+}
```
brooksmtownsend, line 251: I think this might cause some unintuitive behavior, but I do support adding this in our tests. I'm imagining writing a test in the future with something like
```rust
let mut args = TestWashInstanceNewArgs::default();
args.extra_args.extend(["--wadm-version", "v0.1.0"]);
```

and then being confused when my test failed because of duplicate args. Thoughts? I think it might be better to just not include this `impl Default`, but your create code is perfect because it will ensure that the arg is provided one way or another

crates/wash-cli/tests/wash_up.rs
```diff
@@ -274,6 +276,118 @@ async fn integration_up_works_with_labels() -> Result<()> {
     Ok(())
 }
 
+/// Ensure that wash up can start a new host with the new version of wasmcloud if a new patch is available
+#[tokio::test]
+#[serial]
+async fn integration_up_works_with_new_patch_version_if_possible() -> Result<()> {
```
brooksmtownsend, line 282: Tests 😍

crates/wash-lib/src/start/github/api.rs
```diff
@@ -0,0 +1,290 @@
...
+async fn fetch_latest_releases(
+    owner: &str,
+    repo: &str,
+    latest_interested: &semver::Version,
+) -> Result<Vec<GitHubRelease>, reqwest::Error> {
+    let client = reqwest::ClientBuilder::default()
+        .user_agent(VERSION_FETCHER_CLIENT_USER_AGENT)
+        .with_native_certificates()
+        .build()
+        .expect("failed to build HTTP client");
```
brooksmtownsend, line 120: @joonas any concerns here w/ not providing the ability to give custom CA certs? Am I correct in thinking that this goes to github.com, so it shouldn't be an issue?

crates/wash-lib/src/start/github/api.rs
```diff
@@ -0,0 +1,290 @@
...
+        Ok(v) => v,
+        Err(_) => return Ok(None),
+    };
+    return match new_patch_releases_of_after(owner, repo, &after_version).await {
+        Ok(patches) => match patches.first() {
+            Some(patch) => Ok(patch.get_main_artifact_release()),
+            None => Ok(None),
+        },
+        _ => Ok(None),
+    };
```
brooksmtownsend, line 62:
```suggestion
   match new_patch_releases_of_after(owner, repo, &after_version).await {
        Ok(patches) => match patches.first() {
            Some(patch) => Ok(patch.get_main_artifact_release()),
            None => Ok(None),
        },
        _ => Ok(None),
    }
```

Rusty nit

crates/wash-lib/src/start/github/api.rs
```diff
@@ -0,0 +1,290 @@
...
+const VERSION_FETCHER_CLIENT_USER_AGENT: &str =
+    concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
+
+/// get_chronologically_sorted_releases_of_after returns with a list of chronologically ordered (latest first)
+/// with the provided after_version being excluded
+async fn get_chronologically_sorted_releases_of_after(
+    owner: &str,
+    repo: &str,
+    after_version: &semver::Version,
+) -> Result<Vec<GitHubRelease>, anyhow::Error> {
```
brooksmtownsend, line 21: I find the naming of these functions just a little confusing, my suggestion would be to drop a bit of wording from the function in favor of doc comments.
```suggestion
/// get_chronologically_sorted_releases returns with a list of chronologically ordered (latest first)
/// releases that are more recent than the provided `after_version`.
async fn get_chronologically_sorted_releases(
    owner: &str,
    repo: &str,
    after_version: &semver::Version,
) -> Result<Vec<GitHubRelease>, anyhow::Error> {
```

Totally open to suggestions here, this was more of a nit and since it's not a public function I'm not as concerned about it

crates/wash-lib/src/start/github/api.rs
```diff
@@ -0,0 +1,290 @@
...
+        }
+        if releases_on_page.len() < GITHUB_PER_PAGE as usize {
+            break 'fetch_loop;
+        }
+        page += 1;
+    }
+    Ok(releases)
+}
+
+// TODO: find any chrono serde implementation that can be used instead of this.
```
brooksmtownsend, line 150: @thomastaylor312 any ideas on this one?

brooksmtownsend (commented):
Hey @markkovari , HUGE fan of the functionality put forward here, and thank you so much for writing the tests that you've written. I've added a couple of comments but I think they mostly boil down to code style / naming, and happy to give this another look ~whenever CI is green~ and you're ready for it 😄 

edit: CI not being green was our fault, continue on 🫡

crates/wash-lib/src/start/github/api.rs
```diff
@@ -0,0 +1,290 @@
...
+        .with_native_certificates()
+        .build()
+        .expect("failed to build HTTP client");
+    let mut page = 0u32;
+    let mut releases: Vec<GitHubRelease> = Vec::new();
+    'fetch_loop: loop {
+        let url = format_latest_releases_url(owner, repo, page);
+        let response = client.get(&url).send().await?;
+        // It is relatively safe to break the loop if the response is not successful
+        // this is not a critical error, maybe a github api rate limit or similar
```
brooksmtownsend, line 127: Do you have a good idea after testing what the general github rate limit is here? e.g. could we concurrently send these requests all at once to speed up the process?

markkovari: Did some research, the unauthenticated users' rate limit is 60 requests per hour.
https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28

Since this is a best-effort development feature I think we should be safe to send multiple requests once from the same IP.
Probably people are not running `wash up` every minute 🤔 .
On the other hand we could add a local cache file to look up first with a timestamp, but in my opinion we should not increase the scope.

Will add some optimization here. Thanks for the heads-up.

crates/wash-cli/src/cmd/up.rs
markkovari: Absolutely good question there are places where the `v` is needed, I probably will try to centralize the different version of these on the GitHubAction struct or with its Semver field. It is slightly confusing to use, so would make sense to make it more helpful for others if they will use it.

markkovari: Yes, the semver create is flexible when parsing the version, it is treating the "v" prefix as option, but nor the display cannot differentiate the original value.
https://docs.rs/semver/1.0.23/src/semver/display.rs.html#93
If we would (will as I've seen wit the nats) use it in multiple cases we can add a method to the `GitHubRelease` itself. But don't want to prematurely add options that I do not know.

.github/workflows/examples-providers.yml
```diff
@@ -134,6 +134,8 @@ jobs:
       - name: test provider load
         if: ${{ matrix.project.test_deploy && matrix.wash-version == 'current' }}
         shell: bash
+        env:
+          RUST_LOG: trace
```
joonas, line 138: Is this intended change to ship, or just helping diagnose CI while working through this?

markkovari: Hey Jonas, I am starting a thread in the #help channel, since I faced a few local reproduction blockers, definitely should not be here.

brooksmtownsend (approved):
Looks great @markkovari ! Great work

crates/wash-lib/src/start/github/api.rs
joonas: I think certs shouldn't be a problem, but we do need to make sure that proxy variables are/continue to be supported so that end-users inside of limited-connectivity networks can continue to use the provided proxies.

markkovari: Wowzy, nice catch. Reused the already existing function which supplied the optional envvars.

crates/wash-cli/src/cmd/up.rs
```diff
@@ -153,8 +154,8 @@ impl From<NatsOpts> for NatsConfig {
 #[derive(Parser, Debug, Clone)]
 pub struct WasmcloudOpts {
     /// wasmCloud host version to download, e.g. `v0.55.0`. See https://github.com/wasmCloud/wasmcloud/releases for releases
```
joonas, line 156: Would you mind updating this to something more recent while you're here? 😄

crates/wash-cli/src/cmd/up.rs
```diff
@@ -362,8 +363,8 @@ impl WasmcloudOpts {
 #[derive(Parser, Debug, Clone)]
 pub struct WadmOpts {
     /// wadm version to download, e.g. `v0.4.0`. See https://github.com/wasmCloud/wadm/releases for releases
```
joonas, line 365: Would you mind updating this to something more recent while you're here? 😄

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -299,7 +299,11 @@ impl WashDevSession {
                     wadm_log_path.display()
                 )
             })?;
-        let wadm_binary = ensure_wadm(&wadm_opts.wadm_version, &install_dir).await?;
+        let wadm_binary = ensure_wadm(
+            &wadm_opts.wadm_version.unwrap_or(WADM_VERSION.to_owned()),
```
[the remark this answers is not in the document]
markkovari: Good question: since the downloaded binaries are a bit different

```console
ls ~/.wash/downloads/
```

results as 

```
nats-server*   nats.log       v1.4.2/        wadm.log       wasmcloud.log
nats.conf      nats.pid       wadm*          wadm.pid       wasmcloud.pid
```
(where the `v1.4.2` is the wasmCloud binary containing folder).

or similar. Currently only the wasmCloud binary has its own folder, everything else is "just" in the folder. That is the reason for the difference here.

Should I create subfolders for the binaries?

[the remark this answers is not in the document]
markkovari: Oh, nvm this is unrelated, will update in the morning.

crates/wash-lib/src/start/github/api.rs
lachieh, line 1: Should this come out of `start` into its own module since it'll get used in `wash ui` as well?

markkovari: You are correct, I was thinking about adding this to the `wash dev` too. I think it might slow down the "development loop" a bit although. But yeah, currently I just went with the current structure, did not found a better place for it, maybe a common folder in wash-lib? But yeah, absolutely agree, should have a better place.

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -315,9 +320,13 @@ impl WashDevSession {
         };
 
         // Start the host in detached mode, w/ custom log file
+        let wasmcloud_version = wasmcloud_opts
+            .wasmcloud_version
+            .as_ref()
+            .map(|v| v.to_string())
```
joonas, line 326: Isn't `.to_string()` unnecessary here since [`WasmcloudOpts` already requires it to be a string](https://github.com/wasmCloud/wasmCloud/pull/2527/files#diff-95950f0dc432da1e1d2cc34c0d34284d3d9a9cd8c8cf557705eb225a29f28296R162)?

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -299,7 +299,12 @@ impl WashDevSession {
                     wadm_log_path.display()
                 )
             })?;
-        let wadm_binary = ensure_wadm(&wadm_opts.wadm_version, &install_dir).await?;
+        let wadm_version = wadm_opts
+            .wadm_version
+            .as_ref()
+            .map(|v| v.to_string())
```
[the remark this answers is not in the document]
joonas: I _think_ you should be able to have this look exactly like https://github.com/wasmCloud/wasmCloud/pull/2527/files#diff-95950f0dc432da1e1d2cc34c0d34284d3d9a9cd8c8cf557705eb225a29f28296R596

vados-cosmonic:
Just chiming in to thank you for the hard work @markkovari ! We're really close to having the whole team chime in on this PR and liking what they're seeing :)

crates/wash-cli/src/cmd/dev/session.rs
markkovari: Oh, yes. I was thinking about adding the same functionality for dev too, but left it there as is. Creating a commit without the extra mapping. Thanks for the heads-up.

markkovari:
Oh that sounds good and bad at the same time, but thanks for being diplomatic 😄

vados-cosmonic:
Oh definitely meant as a good thing! You should see some of my PRs ;) . Getting the feedback from everyone and incorporating it is painstaking work -- but we get a legitimately great PR out of it at the end!

lachieh:
@markkovari if you rebase on main, the wash unit test should pass

lachieh:
Rebased on main to get #3656 in

PR #2546, opened 2024-07-23
fix(wash-cli): re-add wash call

vados-cosmonic:
This commit re-adds `wash call` with the existing functionality (no improvements) that existed prior to the recent wrpc update.

With this update, invoking simple components and HTTP components both work, and tests have been re-enabled.


brooksmtownsend (changes requested):

crates/wash-cli/Cargo.toml
```diff
@@ -17,7 +17,8 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 anyhow = { workspace = true, features = ["backtrace"] }
 async-compression = { workspace = true, features = ["tokio", "gzip"] }
-async-nats = { version = "0.33" } # TODO: Use workspace version once https://github.com/nats-io/nats.rs/pull/1267 resolved to use in wasmCloud
+async-nats = { workspace = true }
+async-nats-0_33 = { workspace = true }
```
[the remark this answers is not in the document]
vados-cosmonic: Ah thanks good point! I forgot to replace the note that was there before

crates/wash-cli/src/call.rs
```diff
@@ -115,82 +127,61 @@ pub async fn handle_command(
...
+        .clone()
+        .unwrap_or_else(|| DEFAULT_LATTICE.to_string());
+    let prefix = Arc::new(format!("{}.{component_id}", &lattice));
+
+    let nc = create_client_from_opts_wrpc(&opts)
+        .await
+        .context("failed to create async nats client")?;
+    let wrpc_client = wrpc_transport_nats::Client::new(nc, prefix.as_str(), None);
+
+    let (namespace, package, interface, name) = parse_wit_meta_from_operation(&function).context(
```
rvolosatovs, line 141: Just a note here that resource methods and statics are identified as `resource.method`, we need to make sure that whatever UI is used here supports this addressing scheme.
I feel like the CLI probably should separate the instance from function identifier for this to work well

crates/wash-cli/src/call.rs
```diff
@@ -372,119 +363,120 @@ impl HttpHandlerInvocationOpts {
...
+async fn wrpc_invoke_simple(
+    client: wrpc_transport_nats::Client,
+    lattice: &str,
+    component_id: &str,
+    instance: &str,
+    function_name: &str,
+    timeout_ms: u64,
+) -> Result<CommandOutput> {
+    let result = tokio::time::timeout(
+       std::time::Duration::from_millis(timeout_ms),
```
rvolosatovs, line 458: Is there still a reason to wrap this in `tokio::time::timeout` if we're calling `timeout` on the `Invoke`?

crates/wash-cli/src/call.rs
```diff
@@ -115,82 +127,61 @@ pub async fn handle_command(
...
+    let lattice = opts
+        .lattice
+        .clone()
+        .unwrap_or_else(|| DEFAULT_LATTICE.to_string());
+    let prefix = Arc::new(format!("{}.{component_id}", &lattice));
+
+    let nc = create_client_from_opts_wrpc(&opts)
+        .await
+        .context("failed to create async nats client")?;
+    let wrpc_client = wrpc_transport_nats::Client::new(nc, prefix.as_str(), None);
```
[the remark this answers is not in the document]
vados-cosmonic: Definitely can do -- I planned to print the the prefix in error messages below but never actually did it, will change!

- [ ] Remove unused `Arc` for prefix

crates/wash-cli/src/call.rs
vados-cosmonic: Ah could you make an issue for this? This is probably something that needed to be changed when resource support went in, and it would probably be quick and a great "first ticket" maybe for someone to handle.

crates/wash-cli/src/call.rs
vados-cosmonic: Nope, this could be removed!

- [ ] pass timeout directly to wrpc client

crates/wash-cli/src/call.rs
vados-cosmonic: Nvm made an issue here:

https://github.com/wasmCloud/wasmCloud/issues/2572

Issue #2411, opened 2024-06-28
[FEATURE] Complete secrets integration


brooksmtownsend:
### Affected project(s)

- [X] #2497
- [X] examples
- [X] wasmCloud host
- [X] wasmCloud CLI (wash)
- [X] capability providers

### Is your feature request related to a problem? Please describe.

This issue is a larger list of what it means to complete the secrets feature in wasmCloud as specified in #2190.

### Describe the solution you'd like

In order to mark this complete, there are additional tasks to complete past #2344

- [X] #2407 for link secrets support
- [X] Return error strings from the secrets-client crate, cleaning up where necessary
- [X] Update #2395 to include a link secret after #2407, what would be great here additionally is a docker compose for the secrets backend + wasmCloud 1.1 rather than `cargo run`ning the binaries
- [X] #2494
- [X] [https://github.com/wasmCloud/wadm/pull/306](https://github.com/wasmCloud/wadm/pull/306)
- [X] [https://github.com/wasmCloud/wadm/pull/307](https://github.com/wasmCloud/wadm/pull/307)
- [X] Update wasmCloud first party capability providers that require potentially sensitive information to fetch these as secrets first, configuration as a backup with a warning
- [ ] Add documentation on [https://wasmcloud.com](https://wasmcloud.com) about the secrets protocol, example, and best practices.
- [X] Add a CLI for managing secrets in the NATS KV secrets backend

### Describe alternatives you've considered

The following would be nice to have:

- [X] Add a client to the nats-kv-secrets backend for easy putting of secrets and mappings
- [ ] ~~Update the nats-kv-secrets crate to not just be a binary, but to allow for embedding within Rust programs~~
- [ ] Validate secrets and policy server subjects as valid NATS subjects

### Additional context

*No response*

brooksmtownsend:
Since secrets docs are getting addressed in https://github.com/wasmCloud/wasmcloud.com/pull/526, I'm closing out this mega-issue (woo!)

PR #2548, opened 2024-07-23
feat(secrets-nats-kv): add management CLI

brooksmtownsend:
## Feature or Problem
This PR adds a management CLI for the NATS KV secrets backend, allowing users to run the backend and separately put secrets and configure mappings.

This PR also
- Simplifies the config e2e test with new client utility functions
- Adds support to `wash-cli` for generating xkeys

## Related Issues
#2411 

A few issues should be created from this PR
- Add remove_secret utility to secrets-nats-kv
- Add get_secret utility to secrets-nats-kv
- Add release pipeline for releasing the secrets-nats-kv backend


crates/secrets-nats-kv/src/lib.rs
```diff
@@ -5,4 +5,4 @@ pub mod types;
 pub use types::*;
 
 pub mod client;
-pub use client::*;
+// pub use client::*;
```
joonas, line 8: Is this intentionally left commented out?

crates/wash-cli/src/keys.rs
```diff
@@ -15,7 +15,7 @@ const NKEYS_EXTENSION: &str = ".nk";
 pub enum KeysCliCommand {
     #[clap(name = "gen", about = "Generates a keypair")]
     GenCommand {
-        /// The type of keypair to generate. May be Account, User, Module (or Component), Service (or Provider), Server (or Host), Operator, Cluster
+        /// The type of keypair to generate. May be Account, User, Module (or Component), Service (or Provider), Server (or Host), Operator, Cluster, X25519
```
joonas, line 18: I believe they are called Curve in [`nk`](https://github.com/wasmCloud/nkeys/blob/main/src/bin/nk/main.rs#L26), I think using `Curve` here would be good for consistency :)

crates/wash-cli/src/keys.rs
```diff
@@ -64,6 +64,7 @@ pub fn keytype_parser(keytype: &str) -> Result<KeyPairType> {
         "server" | "host" => Ok(KeyPairType::Server),
         "operator" => Ok(KeyPairType::Operator),
         "cluster" => Ok(KeyPairType::Cluster),
+        "x25519" => Ok(KeyPairType::Curve),
```
joonas, line 67: Ditto about calling them `curve`

crates/secrets-nats-kv/src/lib.rs
brooksmtownsend: Nope

crates/wash-cli/src/keys.rs
brooksmtownsend: well look at that, looks like they both are set this way. Let me add an alias at least
```
➜ nk -gen
flag needs an argument: -gen
Usage: nk [options]
    -v                    Show version
    -gen <type>           Generate key for [user|account|server|cluster|operator|curve|x25519]
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -2998,6 +2996,8 @@ impl Host {
         provider: &Provider,
         link: &InterfaceLinkDefinition,
     ) -> anyhow::Result<()> {
+        eprintln!("resolving link {link:?}");
+        eprintln!("provider claims {:?}", provider.claims_token);
```
joonas, line 3000: should these be `debug!(...)` instead?

crates/secrets-nats-kv/Dockerfile
```diff
@@ -23,4 +23,4 @@ WORKDIR /app
 COPY --from=builder /app/target/release/secrets-nats-kv .
 
 # Set the entrypoint
-ENTRYPOINT ["./secrets-nats-kv"]
\ No newline at end of file
+ENTRYPOINT ["./secrets-nats-kv", "run"]
```
joonas, line 26:
```suggestion
ENTRYPOINT ["./secrets-nats-kv"]

CMD ["run"]
```

joonas: This allows folks to use the same docker image for CLI operations without having to override `entrypoint`

joonas (approved):
Looks good outside of the couple of nits

crates/host/src/wasmbus/mod.rs
brooksmtownsend: oh those should be gone, thank you

thomastaylor312 (approved):

PR #2564, opened 2024-07-24
fix(secrets): cleanup & fix secrets impl

brooksmtownsend:
## Feature or Problem
This PR will contain a couple of fixes & cleanup for the secrets implementation found after testing. For now, the only fix relates to building the map to give to components & providers using the name of the secret, rather than the backend-specific key.

More fixes to follow, specifically around the policy properties format.

### DIFF
--- a/crates/host/src/secrets.rs
+++ b/crates/host/src/secrets.rs
@@ -139,7 +139,7 @@ impl Manager {
             .then(|secret_name| async move {
                 match self.config_store.get(&secret_name).await {
                     Ok(Some(secret)) => serde_json::from_slice::<SecretReference>(&secret)
-                        .map_err(|_| anyhow::anyhow!("failed to deserialize secret reference from config store, ensure {secret_name} is a secret reference and not configuration"))
+                        .with_context(|| format!("failed to deserialize secret reference from config store, ensure {secret_name} is a secret reference and not configuration"))
                         .map(|secret_ref| (secret_name.trim_start_matches(SECRET_PREFIX).to_string(), secret_ref)),
                     Ok(None) => bail!(
                         "Secret reference {secret_name} not found in config store"
@@ -160,7 +160,7 @@ impl Manager {
                         host_jwt: host_jwt.to_string(),
                         application: Application {
                             // We pass an empty string if the entity doesn't belong to an application
-                            name: application.cloned().unwrap_or_default(),
+                            name: application.cloned(),
                             policy: secret_ref.policy,
                         },
                     },
@@ -193,7 +193,7 @@ impl Manager {
                         string_secret: None,
                         binary_secret: None,
                         ..
-                    } => return Err(anyhow::anyhow!("secret {secret_name} did not contain a value")),
+                    } => bail!("secret {secret_name} did not contain a value"),
                 };
                 Ok(secrets)
             })
--- a/crates/host/src/secrets.rs
+++ b/crates/host/src/secrets.rs
@@ -152,18 +152,17 @@ impl Manager {
                 let secrets_client = self
                     .get_or_create_secrets_client(&secret_ref.backend)
                     .await?;
-                let application = Application {
-                    // We pass an empty string if the entity doesn't belong to an application
-                    name: application.cloned().unwrap_or_default(),
-                    policy: secret_ref.policy,
-                };
                 let request = SecretRequest {
                     name: secret_ref.key,
                     version: secret_ref.version,
                     context: Context {
                         entity_jwt: entity_jwt.to_string(),
                         host_jwt: host_jwt.to_string(),
-                        application,
+                        application: Application {
+                            // We pass an empty string if the entity doesn't belong to an application
+                            name: application.cloned().unwrap_or_default(),
+                            policy: secret_ref.policy,
+                        },
                     },
                 };
                 secrets_client
--- a/crates/host/src/secrets.rs
+++ b/crates/host/src/secrets.rs
@@ -7,29 +7,14 @@ use async_nats::{jetstream::kv::Store, Client};
 use futures::stream;
 use futures::stream::{StreamExt, TryStreamExt};
 use secrecy::Secret;
-use serde::{Deserialize, Serialize};
 use tokio::sync::RwLock;
 use tracing::instrument;
 use wasmcloud_runtime::capability::secrets::store::SecretValue;
 use wasmcloud_secrets_client::Client as WasmcloudSecretsClient;
-use wasmcloud_secrets_types::{Application, Context, Secret as WasmcloudSecret, SecretRequest};
+use wasmcloud_secrets_types::{Secret as WasmcloudSecret, SecretConfig};
 
 /// The prefix in the CONFIGDATA bucket for secret references
 pub const SECRET_PREFIX: &str = "SECRET_";
-
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
-pub(crate) struct SecretReference {
-    /// The backend to use for retrieving the secret.
-    pub backend: String,
-    /// The key to use for retrieving the secret from the backend.
-    pub key: String,
-    /// The version of the secret to retrieve. If not supplied, the latest version will be used.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub version: Option<String>,
-    /// The policy that defines configuration options for the backend. This is a serialized
-    /// JSON object that will be passed to the backend for policy evaluation.
-    pub policy: String,
-}
 
 #[derive(Debug)]
 /// A manager for fetching secrets from a secret store, caching secrets clients for efficiency.
@@ -138,7 +123,7 @@ impl Manager {
             // Fetch the secret reference from the config store
             .then(|secret_name| async move {
                 match self.config_store.get(&secret_name).await {
-                    Ok(Some(secret)) => serde_json::from_slice::<SecretReference>(&secret)
+                    Ok(Some(secret)) => serde_json::from_slice::<SecretConfig>(&secret)
                         .with_context(|| format!("failed to deserialize secret reference from config store, ensure {secret_name} is a secret reference and not configuration"))
                         .map(|secret_ref| (secret_name.trim_start_matches(SECRET_PREFIX).to_string(), secret_ref)),
                     Ok(None) => bail!(
@@ -152,19 +137,7 @@ impl Manager {
                 let secrets_client = self
                     .get_or_create_secrets_client(&secret_ref.backend)
                     .await?;
-                let request = SecretRequest {
-                    name: secret_ref.key,
-                    version: secret_ref.version,
-                    context: Context {
-                        entity_jwt: entity_jwt.to_string(),
-                        host_jwt: host_jwt.to_string(),
-                        application: Application {
-                            // We pass an empty string if the entity doesn't belong to an application
-                            name: application.cloned(),
-                            policy: secret_ref.policy,
-                        },
-                    },
-                };
+                let request = secret_ref.try_into_request(entity_jwt, host_jwt, application).context("failed to create secret request")?;
                 secrets_client
                     .get(request, nkeys::XKey::new())
                     .await
--- a/crates/host/src/secrets.rs
+++ b/crates/host/src/secrets.rs
@@ -13,9 +13,6 @@ use wasmcloud_runtime::capability::secrets::store::SecretValue;
 use wasmcloud_secrets_client::Client as WasmcloudSecretsClient;
 use wasmcloud_secrets_types::{Secret as WasmcloudSecret, SecretConfig};
 
-/// The prefix in the CONFIGDATA bucket for secret references
-pub const SECRET_PREFIX: &str = "SECRET_";
-
 #[derive(Debug)]
 /// A manager for fetching secrets from a secret store, caching secrets clients for efficiency.
 pub struct Manager {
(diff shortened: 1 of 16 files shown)


crates/host/src/secrets.rs
```diff
@@ -138,25 +138,27 @@ impl Manager {
             .then(|secret_name| async move {
```
lxfontes, line 138: I _think_ we need to massage the secret_name too.
https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L3330
It's likely getting here as `SECRET_rust_hello_world-example`

crates/host/src/secrets.rs
```diff
@@ -138,25 +139,27 @@ impl Manager {
...
                 let secrets_client = self
                     .get_or_create_secrets_client(&secret_ref.backend)
                     .await?;
-                let application = Application{
+                let application = Application {
+                    // We pass an empty string if the entity doesn't belong to an application
                     name: application.cloned().unwrap_or_default(),
-                    policy: secret_ref.policy_properties.clone(),
+                    policy: secret_ref.policy,
                 };
```
joonas, line 159: Is there any reason to not just inline this below inside of the `SecretRequest` like `Context` is already?

crates/secrets-types/src/lib.rs
```diff
@@ -187,6 +189,16 @@ impl SecretConfig {
...
     type Error = anyhow::Error;
 
+    /// Convert this SecretConfig into a HashMap of the form:
+    /// ```json
+    /// {
+    ///   "type": "secret.wasmcloud.dev/v1alpha1",
+    ///   "backend": "baobun",
+    ///   "key": "/path/to/secret",
+    ///   "version": "vX.Y.Z",
+    ///   "policy": "{\"type\":\"properties.secret.wasmcloud.dev/v1alpha1\",\"properties\":{\"key\":\"value\"}}
```
joonas, line 199: Based on the [test below](https://github.com/wasmCloud/wasmCloud/pull/2564/files#diff-140b7c9d1fe7f4ce11beb6d8d2bc2bfef1fc68bb3f5b9c3751f21d8066bd450bR37), should this be:
```suggestion
    ///   "policy": "{\"type\":\"policy.secrets.wasmcloud.dev/v1alpha1\",\"properties\":{\"key\":\"value\"}}
```

brooksmtownsend: That test is actually wrong 😄 it should be properties. Will fix

crates/host/src/secrets.rs
brooksmtownsend: I don't think so, let me see if it looks bad

crates/secrets-types/src/lib.rs
brooksmtownsend: For context, the `type` field within the `policy` is supposed to be the type+version of the properties schema. When / if we decide to update the structure of the `properties` block, we could update the type

joonas: Makes sense, I appreciate the context!

joonas (approved):

crates/host/src/secrets.rs
```diff
@@ -26,8 +26,9 @@ pub(crate) struct SecretReference {
     /// The version of the secret to retrieve. If not supplied, the latest version will be used.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub version: Option<String>,
-    /// The policy that defines configuration options for the backend.
-    pub policy_properties: String,
+    /// The policy that defines configuration options for the backend. This is a serialized
+    /// JSON object that will be passed to the backend for policy evaluation.
+    pub policy: String,
```
vados-cosmonic, line 31:
```suggestion
    pub policy_json: String,
```

A bit late but...

crates/host/src/secrets.rs
```diff
@@ -138,60 +139,61 @@ impl Manager {
             .then(|secret_name| async move {
                 match self.config_store.get(&secret_name).await {
                     Ok(Some(secret)) => serde_json::from_slice::<SecretReference>(&secret)
-                        .map_err(|_| anyhow::anyhow!("failed to deserialize secret reference from config store, ensure {secret_name} is a secret reference and not configuration")),
+                        .map_err(|_| anyhow::anyhow!("failed to deserialize secret reference from config store, ensure {secret_name} is a secret reference and not configuration"))
```
vados-cosmonic, line 142: Are we avoiding using `with_context` to avoid passing along the error?

crates/host/src/secrets.rs
```diff
@@ -138,60 +139,61 @@ impl Manager {
...
-                    version: secret_ref.version.clone(),
+                    name: secret_ref.key,
+                    version: secret_ref.version,
                     context: Context {
                         entity_jwt: entity_jwt.to_string(),
                         host_jwt: host_jwt.to_string(),
-                        application,
+                        application: Application {
+                            // We pass an empty string if the entity doesn't belong to an application
+                            name: application.cloned().unwrap_or_default(),
```
vados-cosmonic, line 163: What was the reasoning against using an `Option` here? Just that `Application` should 99% of the time have a `name`?

crates/host/src/secrets.rs
```diff
@@ -138,60 +139,61 @@ impl Manager {
...
-                        secrets.insert(name.clone(), Secret::new(SecretValue::Bytes(binary_secret)))
+                        secrets.insert(secret_name, Secret::new(SecretValue::Bytes(binary_secret)))
                     }
                     WasmcloudSecret {
-                        name,
                         string_secret: None,
                         binary_secret: None,
                         ..
-                    } => return Err(anyhow::anyhow!("secret {name} did not contain a value")),
+                    } => return Err(anyhow::anyhow!("secret {secret_name} did not contain a value")),
```
vados-cosmonic, line 196:
```suggestion
                    } => anyhow::bail!("secret {secret_name} did not contain a value"),
```

crates/secrets-nats-kv/src/client.rs
```diff
@@ -56,7 +56,7 @@ pub async fn put_secret(
         .expect("should be able to encrypt the secret");
     let response = nats_client
         .request_with_headers(
-            format!("{subject_base}.v0.nats-kv.put_secret"),
+            format!("{subject_base}.v1alpha1.nats-kv.put_secret"),
```
vados-cosmonic, line 59: I could have sworn these got moved to functions?

crates/secrets-types/src/lib.rs
```diff
@@ -119,7 +118,9 @@ impl Context {
 /// default to retrieving the latest version of the secret.
 #[derive(Serialize, Deserialize)]
 pub struct SecretRequest {
-    // The name of the secret
+    /// An identifier of the secret as stored in the secret store. This could
+    /// be a key, a path, or any other identifier that the secret store uses to
+    /// retrieve the secret.
```
vados-cosmonic, line 123:
```suggestion
    /// An identifier of the secret as stored in the secret store. 
    ///
    /// A `SecretRequest` can be a key, path, or any other identifier that the secret store uses to
    /// retrieve a secret.
```

vados-cosmonic: Adding the space here makes the documentation show a little preview properly in cargo doc IIRC

crates/host/src/secrets.rs
thomastaylor312, line 31: Any reason not to make this a `serde_json::Value`? If we want to require this to be a JSON object, why not make it freeform instead of a string?

crates/secrets-types/src/lib.rs
```diff
@@ -139,6 +140,9 @@ pub struct SecretResponse {
 /// A secret that can be either a string or binary value.
 #[derive(Serialize, Deserialize, Default)]
 pub struct Secret {
+    /// An identifier of the secret as stored in the secret store. This could
+    /// be a key, a path, or any other identifier that the secret store uses to
+    /// retrieve the secret.
```
vados-cosmonic, line 145:
```suggestion
    /// An identifier of the secret as stored in the secret store. 
    /// 
    /// A `Secret` can be a key, path, or any other identifier that the secret store uses to
    /// retrieve the secret.
```

crates/secrets-types/src/lib.rs
```diff
@@ -196,19 +208,19 @@ impl TryInto<HashMap<String, String>> for SecretConfig {
...
+                serde_json::to_value(&self.policy_properties)
+                    .context("failed to serialize policy properties")?,
             ),
         ]);
-        map.insert(
-            "policy_properties".to_string(),
-            serde_json::to_string(&policy_properties)
-                .context("failed to serialize policy_properties map as string")?,
-        );
+        map.insert("policy".to_string(), serde_json::to_string(&policy)?);
```
vados-cosmonic, line 223: I do kind of wonder why this isn't in the `::from()` above -- and also it might be nice to add a `.context`

thomastaylor312 (commented):
One minor discussion item that might not even matter

crates/test-util/src/lattice/config.rs
```diff
@@ -33,8 +33,8 @@ pub async fn assert_put_secret_reference(
         ("backend".to_string(), backend.to_string()),
         ("key".to_string(), key.to_string()),
         (
-            "policy_properties".to_string(),
-            serde_json::json!({"type": "policy.secrets.wasmcloud.dev", "properties": {"configuration": "value"}}).to_string(),
+            "policy".to_string(),
+            serde_json::json!({"type": "properties.secrets.wasmcloud.dev/v1alpha1", "properties": {"configuration": "value"}}).to_string(),
```
vados-cosmonic, line 37: Does it make sense for the `"properties"` value to be configurable?

crates/host/src/secrets.rs
brooksmtownsend: Good idea

crates/wash-cli/tests/wash_secrets.rs
```diff
@@ -52,26 +52,102 @@ async fn test_secret_put_and_get() -> anyhow::Result<()> {
...
+        bail!("policy should be a string");
     };
 
-    let policy: HashMap<String, String> =
-        serde_json::from_str(policy).context("policy_properties should be a JSON object")?;
-    assert!(policy.get("type").is_some_and(|t| t == SECRET_POLICY_TYPE));
-    let Some(properties) = policy.get("properties") else {
-        bail!("policy_properties should have a 'properties' field");
+    let serde_json::Value::Object(policy) =
+        serde_json::from_str(policy).context("policy should be a JSON object")?
```
vados-cosmonic, line 62:
```suggestion
    let policy =
        serde_json::from_str(policy).as_object().context("policy should be a JSON object")?
```

crates/wash-cli/tests/wash_secrets.rs
```diff
@@ -52,26 +52,102 @@ async fn test_secret_put_and_get() -> anyhow::Result<()> {
...
+    };
+
     let policy_properties: HashMap<String, String> =
-        serde_json::from_str(properties).context("properties should be a JSON object")?;
-    assert_eq!(policy_properties.len(), 1);
-    assert!(policy_properties
-        .get("backend")
-        .is_some_and(|b| b == "baobun"));
+        serde_json::from_value(properties).context("properties should be a JSON object")?;
+    assert_eq!(policy_properties.len(), 0);
```
vados-cosmonic, line 75:
```suggestion
    assert!(policy_properties.is_empty());
```

Surprised `clippy` said nothing!

crates/wash-cli/tests/wash_secrets.rs
```diff
@@ -52,26 +52,102 @@ async fn test_secret_put_and_get() -> anyhow::Result<()> {
...
 
     assert!(!retrieved_secret.contains_key("version"));
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_secret_put_and_get_complex() -> anyhow::Result<()> {
+    let wash_instance = TestWashInstance::create().await?;
+    // Create a new secret reference
```
vados-cosmonic, line 85:
```suggestion
```

vados-cosmonic (approved):
LGTM :rocket: Nothing but nits!

crates/host/src/secrets.rs
vados-cosmonic: Agreed on making this a `serde_json::Value`... but why isn't this just a `BTreeMap`/`HashMap`? Would it save validation down the line?

crates/host/src/secrets.rs
brooksmtownsend: Good catch, will use this instead

crates/host/src/secrets.rs
brooksmtownsend: Yeah originally it was optional, but now we changed it to be a struct. I think the application name here can still be optional and that will work well, but policy should always be there

crates/secrets-types/src/lib.rs
brooksmtownsend: Just because they are separate objects, but I can add that `.context`

crates/host/src/secrets.rs
brooksmtownsend: I didn't change this one because we want it to be `policy` downstream, so I'd have to change the name and then add a serde rename

crates/wash-cli/tests/wash_secrets.rs
brooksmtownsend: This didn't quite work for me so I just left it as-is, considering it's just in the test

lxfontes:
heads up https://github.com/wasmCloud/wasmCloud/pull/2564#discussion_r1693190831

crates/host/src/secrets.rs
brooksmtownsend: Oh, I see the issue here. When the app is created by wadm we automatically prepend the application name to the name of the secret to prevent collisions between manifests.

That being said, the component shouldn't be responsible for knowing what app it's a part of, nor should the host be responsible for stripping an app name. I'll need to check to see if the host strips this out anywhere or how exactly it works.

brooksmtownsend: @lxfontes fixed this in https://github.com/wasmCloud/wasmCloud/pull/2564/commits/eb76e5204d3d494eefd3146f26bda6e0d7ca71de

thomastaylor312 (approved):

PR #2575, opened 2024-07-25
chore(secrets-client): Update default secrets API version to match the latest changes

joonas:
## Feature or Problem

Since we've landed on `v1alpha1` instead of `v0`, let's update the `secrets-client` to default to that as well.

### DIFF
--- a/.github/workflows/secrets-nats-kv.yml
+++ b/.github/workflows/secrets-nats-kv.yml
@@ -3,6 +3,7 @@ name: secrets-nats-kv
 on:
   push:
     branches: [main]
+  merge_group:
   pull_request:
     branches: [main]
     paths:
@@ -10,6 +11,7 @@ on:
       - Cargo.lock
       - Cargo.toml
       - crates/secrets-nats-kv/**
+      - crates/secrets-client/**
       - crates/secrets-types/**
 
 permissions:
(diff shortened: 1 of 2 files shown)


protochron (approved):

vados-cosmonic (changes requested):
Could you add `merge_group:` to the `secrets-nats-kv` workflow?

joonas:
> Could you add `merge_group:` to the `secrets-nats-kv` workflow?

[`secrets-client` isn't currently part of that workflow](https://github.com/wasmCloud/wasmCloud/blob/main/.github/workflows/secrets-nats-kv.yml#L8-L13), but I can add it as part of this if you wanted to get it in there?

vados-cosmonic:
Ahh yes -- it would be great if we could cover `secrets-client` under test if the tests aren't being run anywhere else right now!

brooksmtownsend (changes requested):
@joonas this is captured in https://github.com/wasmCloud/wasmCloud/pull/2564, along with propagating the version down. I think the merge group change is still good

vados-cosmonic (approved):
LGTM :rocket:

joonas:
> @joonas this is captured in #2564, along with propagating the version down. I think the merge group change is still good

Oh, doh. I thought I looked through every PR to make sure it wasn't already implemented somewhere. I'll close this out and make a separate PR for just the CI change.

PR #2578, opened 2024-07-25
fix(http-server): Improve reuse of underlying TCP socket

thomastaylor312:
After digging around in the `wasmtime serve` code, I found this neat little trick around reuseaddr which works on any linux or unix-like system (sorry Windows!). In my testing, this increased HTTP throughput by 10-15%

### DIFF
--- a/crates/provider-http-server/src/lib.rs
+++ b/crates/provider-http-server/src/lib.rs
@@ -35,7 +35,7 @@
 use core::str::FromStr as _;
 use core::time::Duration;
 
-use std::net::Ipv4Addr;
+use std::net::{Ipv4Addr, SocketAddr};
 use std::sync::Arc;
 
 use anyhow::Context as _;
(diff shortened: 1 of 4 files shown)


vados-cosmonic (approved):
LGTM :rocket: 

`SO_REUSEADDR`, a tale as old as time :musical_note: 

So I actually think though that the perf benefit you saw here was from increasing the backlog to 1024 -- `SO_REUSEADDR` only comes into play when the server *goes down*/socket is closed -- and the default [backlog is actually 128](https://github.com/rust-lang/rust/blob/2f26b2a99ab976c43d12cf57ef4a3a2c82ede286/library/std/src/sys_common/net.rs#L433) as far as I can see.

crates/provider-http-server/src/lib.rs
```diff
@@ -365,6 +365,18 @@ impl HttpServerCore {
 
         let task_handle = handle.clone();
         let target = target.to_owned();
+        let socket = tokio::net::TcpSocket::new_v4().context("Unable to open socket")?;
```
rvolosatovs, line 368: What about IPv6 addresses? It seems we should have a conditional here using the right socket depending on `addr`.
btw, we should probably listen on IPv6 by default if IP is not specified, since that normally binds IPv4 as well

thomastaylor312: Ah good point! Let me go fix that

thomastaylor312: I left our current IPv4 default for now to keep it small, but we can go back and fix it later as well

thomastaylor312:
@vados-cosmonic I did check into that just so we have the information capture for our future selves (if we're optimizing anything). It looks like the queue size [is a hint anyway](https://stackoverflow.com/questions/5111040/listen-ignores-the-backlog-argument/5111841#5111841) and I did a test by bumping it down to 128 and it didn't affect the results

vados-cosmonic:
Ah @thomastaylor312 thanks for checking -- weird that it works but maybe the provider is reconnecting more than I expected!

Issue #2530, opened 2024-07-22
[FEATURE] Timeout component export future based on the provided max_execution_time


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

We currently have tuning for the `max_execution_time` parameter for the host since #2439, but we don't currently ensure that we are timing out the future for handling a request if it takes longer than the max execution time.

### Describe the solution you'd like

The future at https://github.com/wasmCloud/wasmCloud/blob/main/crates/host/src/wasmbus/mod.rs#L1216 should be wrapped in a tokio::time::timeout to ensure that if it takes longer than the requested max execution time, we stop polling that future & shutdown the component. 

We should return a reasonable error if a component handler takes longer than the max execution time to execute.

Aditya1404Sal:
Hey @brooksmtownsend , Can I pick this issue up ?

PR #2586, opened 2024-07-26
feat(host):Timeout component export future based on the provided max_execution_time

Aditya1404Sal:
## Related Issues
#2530 

## Release Information
next

### Manual Verification
Ran Standard Wash up with `max-execution-time` flag ,then initialized a new component that ran as expected


crates/host/src/wasmbus/mod.rs
```diff
@@ -1218,7 +1218,16 @@ impl Host {
...
+                                                    match timeout(max_execution_time, fut).await {
+                                                        Ok(result) => {
+                                                            if let Err(err) = result {
+                                                                error!(?err, "task failed");
+                                                            }
+                                                        },
+                                                        Err(err) => {
+                                                            error!(?err, "Component Handler timed out");
+                                                        },
+                                                    }
```
brooksmtownsend, line 1230:
```suggestion
                                                    match timeout(max_execution_time, fut).await {
                                                        Ok(Err(err)) => {
                                                            error!(?err, "component handler error");
                                                        },
                                                        Err(err) => {
                                                            error!(?err, "component handler timed out");
                                                        },
                                                        Ok(_) => {},
                                                    }
```

Here's one way you can combine the blocks together into a less-nested match!

brooksmtownsend: Bad indentation here, but here's the less-indented block:
```rust
match timeout(max_execution_time, fut).await {
    Ok(Err(err)) => {
        error!(?err, "component handler error");
    },
    Err(err) => {
        error!(?err, "component handler timed out");
    },
    Ok(_) => {},
}
```

Aditya1404Sal:
> This looks nice! I'd love to see what the error looks like when it actually times out.

It would get logged as `ERROR component handler timed out: Elapsed` , but so far I haven't been able to get the component handler to timeout, Even after adding an artificial delay inside the component to be more than `max-execution-time-ms` and stress testing it by a load-tester.
Do you have any suggestions @brooksmtownsend

brooksmtownsend:
@Aditya1404Sal interesting... just to be sure, what are you testing with? Is it an HTTP app? I would recommend updating to 0.22.0, because there was a bug with that future on older versions of capability providers.

Aditya1404Sal:
@brooksmtownsend yep The HTTP tiny-go Component, I updated the version of the http-server capability to `ghcr.io/wasmcloud/http-server:0.22.0` and still got no luck, maybe it's the way I'm inducing the artificial delay ??
I apologise for this question but I added a time.Sleep() before `OutGoing_Response` for the delay, I know it's probably wrong but where else can i add a delay ?

brooksmtownsend:
@Aditya1404Sal anywhere in that component adding a time.sleep should be enough, if it's not then it might be an indicator that the future isn't able to stop the execution of the component.

Would you mind also trying doing a loop of multiple sleeps? Just to see if blocking is the issue

Aditya1404Sal:
Yeah sure, I'll try multiple sleep looping and get back to you.

Aditya1404Sal:
@brooksmtownsend the problem isn't with blocking, even multiple sleep loops didn't work.
There might be an issue with the future polling and the way Timeout is handling the comparison.
I'm working on a fix ,so I'll keep you updated.

vados-cosmonic:
Hey @Aditya1404Sal now that we have [`wasmcloud-test-util`](https://crates.io/crates/wasmcloud-test-util) out, what do you think about automating the testing you were doing? 

I'd like to help with this PR (if I even can! the code *looks* right already), but it would be great to have a test that automates the testing you were doing so that we can be sure it works. This is also a great chance to have you kick the tires on `wasmcloud-test-util`!

vados-cosmonic:
Hey @Aditya1404Sal a great way to start would be [using the `WasmcloudTestHost` like in the example!](https://docs.rs/wasmcloud-test-util/0.12.0/wasmcloud_test_util/) -- if you see anything wrong with the documentation any updates are also welcome :)

Basically if you can translate what you were doing manually with creating components, etc to a `wash-cli` test, that would be great

Aditya1404Sal:
Hey @vados-cosmonic -- In the tests, what should I do to induce an artificial delay in the component exported code ? so that exporting the component takes longer than `max_execution_time` ?

brooksmtownsend:
@Aditya1404Sal I think that just adding a `std::thread::sleep` for a longer duration than the max execution time would be enough! Alternatively, a loop with multiple thread sleeps e.g. 
```
loop {
  std::thread::sleep(std::time::Duration::from_secs(1));
}
```

Aditya1404Sal:
@brooksmtownsend , Thank you for you help ! I was confused as to where I should be writing the test component but it's clear to me now, I just have to develop a component using an http-hello-world template (rust) and add a delayer (like std::thread::sleep) then build it into a .wasm file and move it to `wasmCloud/target/debug/build/testcomponent-......../out/` after which I can continue writing the testcase.

crates/host/src/wasmbus/mod.rs
```diff
@@ -1263,15 +1263,14 @@ impl Host {
...
+                                                    match timeout(max_execution_time, fut).await {
+                                                        Ok(Err(err)) => {
+                                                            error!(?err, "component handler error");
                                                         },
                                                         Err(err) => {
-                                                            warn!(?err, "failed to handle invocation");
-                                                            Err(err)
+                                                            error!(?err, "component handler timed out");
                                                         },
+                                                        Ok(_) => {},
```
brooksmtownsend, line 1273: small nit, I think it would be good to keep the trace event here for a successful invocation just to confirm 😄 
```suggestion
                                                        Ok(_) => {
                                                          debug!("successfully handled invocation")
                                                        },
```

target/debug/build/test-components-ab79bd6ad754baef/out/rust_http_export_timeout.wasm
brooksmtownsend, line 1: I'm interested in how this got committed, but we should remove it before this PR merges

brooksmtownsend: Oh, I think I see why, I can assist with integrating this into the build setup that we have for the component tests, there's quite a few moving pieces there 😄

brooksmtownsend (commented):
I think that this PR functionally looks great, the timeout logic is solid and the test looks nice. Left one nit and then another comment about integrating this into our integration test suite, which I'm happy to assist with

target/debug/build/test-components-ab79bd6ad754baef/out/rust_http_export_timeout.wasm
Aditya1404Sal: Yeah That'd be awesome. I'll remove it out first !

Aditya1404Sal:
Hey @brooksmtownsend can we move ahead with integrating the test component into the build setup ? 

If there are some resources that can guide me that would be very helpful.

thomastaylor312:
@brooksmtownsend @Aditya1404Sal did we need to get together to help show how to integrate things into the build setup?

Aditya1404Sal:
@thomastaylor312 Actually, I managed to integrate it into the build setup.
The main blocker for this issue is that logically the timeout `should` work but the test fails everytime

PR #2629, opened 2024-07-30
chore(runtime): increase max memory size to 10mb

vados-cosmonic:
This commit increases the max memory that can be addressed by a single component to 10mb.


brooksmtownsend (approved):
I wonder if this should be even more?

PR #2634, opened 2024-07-30
feat: add field to secrets

brooksmtownsend:
## Feature or Problem
This PR adds the `field` parameter to the `SecretRequest` and `SecretConfig` types, allowing secrets backends that support indexed secrets inside of an individual key to only return the desired field.

An example of this is vault, where secrets are addressable by a path (which we use the `key` parameter) and then contained in that path are multiple key=value pairs.

This PR also removes the unneeded `name` field from the Secret, since the backend doesn't actually need this value and returning the key/field arbitrarily would just increase bytes sent/encrypted.

## Related Issues
Upcoming wadm issue to support this.

## Release Information
1.1.0-beta.2

### DIFF
--- a/crates/secrets-nats-kv/src/types.rs
+++ b/crates/secrets-nats-kv/src/types.rs
@@ -2,6 +2,14 @@ use bytes::Bytes;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
+#[derive(Serialize, Deserialize, Default)]
+pub struct PutSecretRequest {
+    pub key: String,
+    pub version: String,
+    pub string_secret: Option<String>,
+    pub binary_secret: Option<Vec<u8>>,
+}
+
 /// The response to a `put_secret` operation.
 /// This response contains the revision number of the secret that was just written.
 #[derive(Serialize, Deserialize, Debug, Default)]
--- a/crates/wash-cli/tests/wash_secrets.rs
+++ b/crates/wash-cli/tests/wash_secrets.rs
@@ -22,6 +22,7 @@ async fn test_secret_put_and_get() -> anyhow::Result<()> {
         name: "foobar".to_string(),
         backend: "baobun".to_string(),
         key: "path/to/secret".to_string(),
+        field: None,
         version: None,
         policy_properties: vec![],
     };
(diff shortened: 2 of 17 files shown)


lxfontes (approved):

crates/wash-cli/tests/wash_secrets.rs
```diff
@@ -110,7 +112,7 @@ async fn test_secret_put_and_get_complex() -> anyhow::Result<()> {
     .await?
     .map;
 
-    assert_eq!(retrieved_secret.len(), 6);
+    assert_eq!(retrieved_secret.len(), 7);
     assert!(retrieved_secret
         .get("name")
```
joonas, line 117: Should this be updated to `key`?

joonas (approved):

joonas (approved):

crates/secrets-nats-kv/src/types.rs
```diff
@@ -2,6 +2,14 @@ use bytes::Bytes;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
+#[derive(Serialize, Deserialize, Default)]
+pub struct PutSecretRequest {
+    pub key: String,
+    pub version: String,
+    pub string_secret: Option<String>,
+    pub binary_secret: Option<Vec<u8>>,
+}
```
vados-cosmonic, line 11: This seems familiar, we can't re-use it from somewhere?

vados-cosmonic, line 11: This seems familiar, we can't re-use it from somewhere? 

ahh it's more like a serialization of the `SecretValue` + `key` + `version`, which is why it seems familiar. If this is something we'd use from more than one secrets provider it might be nice to move it up to a shared package

vados-cosmonic (approved):
LGTM :rocket:

crates/secrets-nats-kv/src/types.rs
joonas: For bit of context, all of the other secrets providers that we've implemented or considered implementing so far have their own set of management APIs and tooling, NATS KV just happens to be the one we built on top of NATS, so these are required for NATS, but not for any of the other providers.

protochron (approved):

crates/secrets-nats-kv/src/types.rs
vados-cosmonic: Ah thanks -- I guess using `SercretValue` as part of this struct isn't a huge requirement either, :rocket:

PR #2643, opened 2024-07-31
chore(wash-lib): fix build.rs lint

vados-cosmonic:

### DIFF
--- a/crates/wash-lib/build.rs
+++ b/crates/wash-lib/build.rs
@@ -14,7 +14,13 @@ async fn record_reachability(host: &str) {
             timeout(Duration::from_millis(500), TcpStream::connect(addr)).await,
             Ok(Ok(_))
         ) {
-            println!("cargo:rustc-cfg=can_reach_{}", host.replace('.', "_"));
+            let host = host.replace('.', "_");
+            println!("cargo:rustc-cfg=can_reach_{host}");
+            // NOTE: In Rust 1.80.0 custom cfg conditions became a warning.
+            //
+            // As wash does not have an MSRV, we may have people build on versions *before* 1.80,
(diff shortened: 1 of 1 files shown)


rvolosatovs (approved):

crates/wash-lib/build.rs
```diff
@@ -14,7 +14,13 @@ async fn record_reachability(host: &str) {
...
             Ok(Ok(_))
         ) {
-            println!("cargo:rustc-cfg=can_reach_{}", host.replace('.', "_"));
+            let host = host.replace('.', "_");
+            println!("cargo:rustc-cfg=can_reach_{host}");
+            // NOTE: In Rust 1.80.0 custom cfg conditions became a warning.
+            //
+            // As wash does not have an MSRV, we may have people build on versions *before* 1.80,
+            // and we can remove the double-configuration line below a couple versions later.
+            println!("cargo:rustc-check-cfg=cfg(can_reach_{host})");
```
rvolosatovs, line 23: this has to happen *outside* the loop, because this block is only ever executed when the host is *reachable*, which is not the case within the Nix sandbox

rvolosatovs (changes requested):

crates/wash-lib/build.rs
vados-cosmonic: Ah it actually should be happening outside the loop *completely* -- good catch

vados-cosmonic: Going to  solve this in https://github.com/wasmCloud/wasmCloud/pull/2602

vados-cosmonic:
Closing this in favor of https://github.com/wasmCloud/wasmCloud/pull/2602

PR #2651, opened 2024-07-31
fix/examples versions

brooksmtownsend:
- **chore(examples): remove versions from hello-world wadms**
- **chore(nats-kv): format markdown, indicate bucket as required**

## Feature or Problem
This PR removes versions from our example wadm manifests to avoid annoyances during the hello world tutorial when making changes.

### DIFF
--- a/examples/rust/components/http-hello-world/wadm.yaml
+++ b/examples/rust/components/http-hello-world/wadm.yaml
@@ -3,7 +3,6 @@ kind: Application
 metadata:
   name: rust-hello-world
   annotations:
-    version: v0.0.1
     description: 'HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)'
     wasmcloud.dev/authors: wasmCloud team
     wasmcloud.dev/source-url: https://github.com/wasmCloud/wasmCloud/blob/main/examples/rusg/components/http-hello-world/wadm.yaml
@@ -16,9 +15,9 @@ spec:
     - name: http-component
       type: component
       properties:
-        # To use the locally compiled code in this folder, use the line below instead after running `wash build`:
-        # image: file://./build/http_hello_world_s.wasm
-        image: ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0
+        image: file://./build/http_hello_world_s.wasm
+        # To use the a precompiled version of this component, use the line below instead:
+        # image: ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0
       traits:
         # Govern the spread/scheduling of the component
         - type: spreadscaler
--- a/crates/provider-keyvalue-nats/README.md
+++ b/crates/provider-keyvalue-nats/README.md
@@ -2,11 +2,11 @@
 
 This capability provider is an implementation of the following interfaces of `wasi:keyvalue` proposal:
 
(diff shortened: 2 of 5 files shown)


examples/rust/components/http-hello-world/wadm.yaml
```diff
@@ -16,9 +15,9 @@ spec:
     - name: http-component
       type: component
       properties:
-        # To use the locally compiled code in this folder, use the line below instead after running `wash build`:
-        # image: file://./build/http_hello_world_s.wasm
-        image: ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0
+        image: file://./build/http_hello_world_s.wasm
+        # To use the a precompiled version of this component, use the line below instead:
+        # image: ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0
```
vados-cosmonic, line 20: was this intentional?

vados-cosmonic: This *was* intentional, as people might be surprised when local changes don't appear

vados-cosmonic (approved):
LGTM :rocket:

examples/rust/components/http-hello-world/wadm.yaml
vados-cosmonic: @lachieh are you OK with this? Was playground going to try and use these `wadm.yaml`s?

joonas: I think we should consider duplicating the `wadm.yaml` rather than switching between the local and external references, that way we could point the various examples at the appropriate version (for example, the various Kubernetes examples/documentation should use the external reference)

vados-cosmonic: Yeah I'm not against that -- naming might be tricky but two different files would make things easier/avoid having to do this toggling.

lachieh: Yeah, we've discussed before and I think what Joonas suggested is what we landed on, it just takes someone to implement. In the meantime, I'd prefer the examples to work out-of-the-box, which means external references. I think we have example manifests with both local and external, and the console has its own examples/dependencies so I don't really mind which happens in this PR.

peter-jerry-ye:
This will cause give error message like this using `wash app deploy wadm.yaml`

> Invalid request: Model with the name typescript-hello-world does not have the specified version to deploy

Maybe the document needs to be updated to mention that we need to add version before deploying?

joonas:
@peter-jerry-ye which version of `wadm` and `wash` are you seeing this behavior with? 

As of wadm `v0.12.2` (latest being `0.13.0`) and wash `v0.29.0`(latest being `v0.29.2`) you should no longer need to include the version.

peter-jerry-ye:
> @peter-jerry-ye which version of `wadm` and `wash` are you seeing this behavior with?
> 
> As of wadm `v0.12.2` (latest being `0.13.0`) and wash `v0.29.0`(latest being `v0.29.2`) you should no longer need to include the version.

I'm using wash `v0.29.2`. I'm not sure about the wadm's version though. How can I get that version displayed?

peter-jerry-ye:
I tried with `wash up --wasmcloud-version v1.1.0 --wadm-version v0.13.0` and the response is still the same.

PR #2667, opened 2024-08-02
fix: Ensure file exists before attempting to remove

joonas:
## Feature or Problem

This addresses the root cause for https://github.com/wasmCloud/wadm/actions/runs/10208903476/job/28246069141?pr=367#step:8:159 when the host is told to stop, but it's not running on the same machine as `wash` (in the Actions example above, it's running inside of a Docker Container).


vados-cosmonic (approved):
LGTM :rocket:

PR #2669, opened 2024-08-02
feat(examples): add load testing for example components

vados-cosmonic:
This commit adds a small bash script that loads an example component(s) and attempts to run it with `wash` & `wadm`, and waits until at least one component is running then exits.

### DIFF
--- a/.github/workflows/examples-components.yml
+++ b/.github/workflows/examples-components.yml
@@ -19,7 +19,7 @@ on:
   pull_request:
     branches: [main]
     paths:
-      - .github/workflows/examples.yml
+      - .github/workflows/examples-components.yml
       - crates/wash-lib/**
       - crates/wash-cli/**
       - examples/rust/components/**
@@ -39,7 +39,7 @@ env:
 
 jobs:
   build-wash-cli:
-    name: Build wash CLI
+    name: build wash-cli
     runs-on: ubuntu-22.04
     strategy:
       matrix:
@@ -87,9 +87,11 @@ jobs:
           - lang: "golang"
             lang_version: "1.20"
             name: "http-echo-tinygo"
+            test_deploy: wadm.yaml
           - lang: "golang"
             lang_version: "1.20"
             name: "http-hello-world"
+            test_deploy: wadm.yaml
           # Rust example components
           - name: "blobby"
             lang: "rust"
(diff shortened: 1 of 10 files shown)


ricochet:
Is it working? I was curious how much time it would add to a CI run and spied errors in the logs: https://github.com/wasmCloud/wasmCloud/actions/runs/10267790063/job/28409612389?pr=2669#step:12:114

vados-cosmonic:
Hey @ricochet great questions!

> Is it working?
Yes! The test is specifically for *components* -- and it waits until a single component is loaded. The only lynchpin here is whether that component is the *right* component, but in the examples we generally only start one component (and start other things as well)

If the component never shows up, the test fails -- the component showed up, so the test passed.

> how much time it would add to a CI run 
CI for that step ran in 1m 8s  (`succeeded Aug 6, 2024 in 1m 8s`)

IMO these steps are not the blockers extending overall CI time taken.

> spied errors in the logs: https://github.com/wasmCloud/wasmCloud/actions/runs/10267790063/job/28409612389?pr=2669#step:12:114

Yup, that error was related to a *provider* not loading properly -- This PR was meant to test that *components* themselves load, for E2E tests they belong either on the providers (we can use `wasmcloud-test-util` in the future) or in the top level `wasmcloud` repo (we could add `wadm` driven tests there).

For reference, the error:

```
Caused by:
    0: failed to fetch provider under OCI reference `ghcr.io/wasmcloud/keyvalue-nats:0.1.0`
    1: failed to fetch OCI path
    2: failed to fetch OCI bytes
    3: Registry error: url https://ghcr.io/v2/wasmcloud/keyvalue-nats/manifests/0.1.0, envelope: OCI API errors: [OCI API error: manifest unknown]
    ```

**Comment by vados-cosmonic** [comment]
Oh and to help resolve the issue of references going bad, I filed:

https://github.com/wasmCloud/wasmCloud/issues/2708

**Comment by ricochet** [comment]
> Hey @ricochet great questions!
> 
> > Is it working?
> > Yes! The test is specifically for _components_ -- and it waits until a single component is loaded. The only lynchpin here is whether that component is the _right_ component, but in the examples we generally only start one component (and start other things as well)
> 
> If the component never shows up, the test fails -- the component showed up, so the test passed.
> 
> > how much time it would add to a CI run
> > CI for that step ran in 1m 8s  (`succeeded Aug 6, 2024 in 1m 8s`)
> 
> IMO these steps are not the blockers extending overall CI time taken.
> 
> > spied errors in the logs: https://github.com/wasmCloud/wasmCloud/actions/runs/10267790063/job/28409612389?pr=2669#step:12:114
> 
> Yup, that error was related to a _provider_ not loading properly -- This PR was meant to test that _components_ themselves load, for E2E tests they belong either on the providers (we can use `wasmcloud-test-util` in the future) or in the top level `wasmcloud` repo (we could add `wadm` driven tests there).
> 
> For reference, the error:
> 
> ```
> Caused by:
>     0: failed to fetch provider under OCI reference `ghcr.io/wasmcloud/keyvalue-nats:0.1.0`
>     1: failed to fetch OCI path
>     2: failed to fetch OCI bytes
>     3: Registry error: url https://ghcr.io/v2/wasmcloud/keyvalue-nats/manifests/0.1.0, envelope: OCI API errors: [OCI API error: manifest unknown]
> ```

Thank you! So long as we're confident these will not be flaky, I can be onboard with this approach. I'd like another maintainer to review.

File: .github/workflows/examples-components.yml [hunk 1]
```diff
@@ -168,10 +179,37 @@ jobs:
...
+            if [[ $TRIES -gt 10 ]]; then
+              echo "❌ failed to find component in inventory output after deploying example manifest";
+              exit -1;
+            fi
+            TRIES=$((TRIES+1));
+            sleep 1;
+          done;
+          echo "✅ successfully started at least one component";
+          wash app delete ${{ matrix.project.test_deploy }};
+          kill $WASH_PID;
```
brooksmtownsend, line 210: What do you think about including an "invocation command" that you can run here after the app is deployed to test that it actually worked? I don't want to make it too complicated with our examples that require extra dependencies like Redis, but it might be cool to have (at least for the HTTP ones) examples where we `curl` the local port & make sure we get a payload back

brooksmtownsend (approved):
This is really great, and I have a suggestion that can be a followup or included here. No need to block since this beats what we've got!

.github/workflows/examples-components.yml
vados-cosmonic: Hey sorry late getting to this but that would be a great add on -- checking the functionality of the component a little more closely would be great. 

That said, I think if we want to make this more complex we should actually just start making top level tests for these examples, maybe using `cargo xtask` -- we could get rid of the bash spaghetti here and take advantage of the monorepo... What do you think?

PR #2674, opened 2024-08-02
fix(host): always publish component_scale event

brooksmtownsend:
## Feature or Problem
I noticed that if a component fails to scale because of an issue with OCI downloads, claims extraction, or policy server denial, we don't publish the `component_scale_failed` event. This PR ensures:
1. Any error during component scaling is reflected as an event
1. Any scale operation, even if a no-op (e.g. requested scale that's already running) publishes an event to ensure the idempotency of the command. This helps consumers like wash always know their request completed successfully.

- [ ] It doesn't have to happen in this PR, but it might be nice to ensure that the provider & update commands also reflect this idempotency

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1281,13 +1281,12 @@ impl Host {
         component_ref: Arc<str>,
         component_id: Arc<str>,
         max_instances: NonZeroUsize,
-        annotations: impl Into<Annotations>,
+        annotations: &Annotations,
         config: ConfigBundle,
         secrets: HashMap<String, Secret<SecretValue>>,
     ) -> anyhow::Result<&'a mut Arc<Component>> {
         debug!(?component_ref, ?max_instances, "starting new component");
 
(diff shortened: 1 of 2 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -1649,80 +1717,62 @@ impl Host {
...
+                    wasm,
+                    claims.clone(),
+                    Arc::clone(&component_ref),
+                    Arc::clone(&component_id),
+                    max,
+                    annotations,
+                    config,
+                    secrets,
+                )
+                .await?;
```
vados-cosmonic, line 1731:
```suggestion
                .await.context("failed to start component while processing scale request")?;
```

vados-cosmonic (approved):
LGTM :rocket:

rvolosatovs (approved):

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Will grab this in a followup to keep the review simple!

PR #2682, opened 2024-08-05
feat(host): Implement NATS subject validation for secret-topic and policy-topic

### LINKED ISSUE #2613: [FEATURE] Validate secrets topic & policy topic as valid NATS subjects
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

Currently, providing an invalid NATS topic to the `--secrets-topic` or `--policy-topic` results in questionable error messages, or worse, no error messages.

I would love to include some extra `ensure!` logic during the startup of a host to check that the secrets and policy topics are valid NATS subjects.

The following error message should fail with an invalid policy topic message.
```
~/.wash/downloads/v1.0.4/wasmcloud_host --policy-topic=""
```

_No response_

### Additional context

_No response_

Aditya1404Sal:
## Related Issues
resolves #2613

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,9 +4,10 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::{self, bail, Context};
+use anyhow::{bail, Context};
 use clap::Parser;
 use nkeys::KeyPair;
+use regex::Regex;
 use tokio::time::{timeout, timeout_at};
 use tokio::{select, signal};
 use tracing::{warn, Level as TracingLogLevel};
@@ -384,6 +385,12 @@ async fn main() -> anyhow::Result<()> {
         oci_user: args.oci_user,
         oci_password: args.oci_password,
     };
+    if let Some(policy_topic) = args.policy_topic.as_deref() {
+        anyhow::ensure!(
+            validate_nats_subject(policy_topic).is_ok(),
+            "Invalid policy topic"
+        );
+    }
     let policy_service_config = PolicyServiceConfig {
         policy_topic: args.policy_topic,
         policy_changes_topic: args.policy_changes_topic,
@@ -412,6 +419,12 @@ async fn main() -> anyhow::Result<()> {
         }
         Some((key, value))
     }));
+    if let Some(secrets_topic) = args.secrets_topic_prefix.as_deref() {
+        anyhow::ensure!(
+            validate_nats_subject(secrets_topic).is_ok(),
+            "Invalid secrets topic"
+        );
+    }
     let (host, shutdown) = Box::pin(wasmcloud_host::wasmbus::Host::new(WasmbusHostConfig {
         ctl_nats_url,
         lattice: Arc::from(args.lattice),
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -462,6 +471,15 @@ fn parse_duration(arg: &str) -> anyhow::Result<Duration> {
         .map_err(|e| anyhow::anyhow!(e))
 }
 
+fn validate_nats_subject(subject: &str) -> anyhow::Result<()> {
+    let re = Regex::new(r"^(?:[A-Za-z0-9_-]+\.)*[A-Za-z0-9_-]+$").unwrap();
```
brooksmtownsend, line 475: _just_ in case, it would be good to have a small test that just ensures that this doesn't fail to instantiate, and have this at least be a `context(...)?` rather than an unwrap.

brooksmtownsend (changes requested):
Looks pretty good! Just wanted to have a bit more safety around the regex here. Where did you find it, did it come from the NATS library?

Aditya1404Sal:
> Looks pretty good! Just wanted to have a bit more safety around the regex here. Where did you find it, did it come from the NATS library?

I followed a regex guide and built it based off this section in the Nats documentation https://docs.nats.io/nats-concepts/subjects#characters-allowed-and-recommended-for-subject-names

src/main.rs
Aditya1404Sal: I understand :+1: 
```rust
let re = Regex::new(r"^(?:[A-Za-z0-9_-]+\.)*[A-Za-z0-9_-]+$")
        .context("Failed to compile NATS subject regex")?;
```
So like this ?
and where exactly should I add a test ?

brooksmtownsend:
Nice @Aditya1404Sal ! It may be a good idea to include in the test a couple of cases that you expect to succeed & fail, definitely including
1. An empty topic failing
2. A topic consisting of just a `.` failing
3. A topic starting or ending with a `.` failing

Given that this is a topic that the host will publish to, we also shouldn't support wildcards `*` or `>`

src/main.rs
brooksmtownsend: That looks good! If you want, at the bottom of this file, you can add a:

```rust
#[cfg(test)]
mod test {
  fn test_validate_nats_subject() {
    ...
  }
}
```

Then that should run along with the cargo test suite

Aditya1404Sal: I'll have to add this check for explicitly disallowing `*` and `>`
```rust
if re.is_match(subject) && !subject.contains('*') && !subject.contains('>')
```

Aditya1404Sal:
Thanks for the clarification @brooksmtownsend 
So I'll have to add this check for explicitly disallowing `*` and `>`
```rust
if re.is_match(subject) && !subject.contains('*') && !subject.contains('>')
```
one last thing, should I add the test you mentioned in `wasmcloud/tests/common` as `subjects.rs` or under `nats.rs` ?

src/main.rs
Aditya1404Sal: Gotcha :+1: I'll get on with it :rocket:

src/main.rs
```diff
@@ -467,6 +480,16 @@ fn parse_duration_millis(arg: &str) -> anyhow::Result<Duration> {
         .map_err(|e| anyhow::anyhow!(e))
 }
 
+fn validate_nats_subject(subject: &str) -> anyhow::Result<()> {
```
brooksmtownsend, line 483: nit: A comment here explaining that it validates that a NATS subject is valid to publish to might be good so we don't re-use it to check subjects to subscribe to

brooksmtownsend (approved):
This looks great! The test case is really good to understand the purpose of the function. I left one nit but this is fine to go in either way

src/main.rs
Aditya1404Sal: That's no problem ! I'll write a clear and concise comment explaining this function (PS: I resolved a merge conflict without any hiccups :see_no_evil: )

brooksmtownsend: Thank yoU!

brooksmtownsend (approved):
Looks good to me once tests/clippy passes

brooksmtownsend:
@Aditya1404Sal how is this coming along with the tests?

Aditya1404Sal:
> @Aditya1404Sal how is this coming along with the tests?

It's not going great so far, I do need some help for this.

Aditya1404Sal:
@brooksmtownsend Almost done !

Aditya1404Sal:
@brooksmtownsend This PR is ready to go!

brooksmtownsend (approved):

PR #2689, opened 2024-08-05
chore(host): Make wasmcloud host heartbeat interval configurable

joonas:
## Feature or Problem

For integration testing, it would be nice to make the host heartbeat interval configurable so that it can be shortened in an integration testing that uses the heartbeats.

In the future being able to adjust the host heartbeat frequency would also open up the possibility of reducing the amount of heartbeat traffic in a large deployment, though this likely necessitates changes in other components of the system as well (i.e. adjusting the wadm side of things appropriately as well).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -61,7 +61,7 @@ struct Args {
     #[clap(long = "host-seed", env = "WASMCLOUD_HOST_SEED")]
     host_seed: Option<String>,
     /// Delay, in milliseconds, between requesting a provider shut down and forcibly terminating its process
-    #[clap(long = "provider-shutdown-delay", default_value = "300", env = "WASMCLOUD_PROV_SHUTDOWN_DELAY_MS", value_parser = parse_duration)]
+    #[clap(long = "provider-shutdown-delay-ms", alias = "provider-shutdown-delay", default_value = "300", env = "WASMCLOUD_PROV_SHUTDOWN_DELAY_MS", value_parser = parse_duration_millis)]
     provider_shutdown_delay: Duration,
     /// Determines whether OCI images tagged latest are allowed to be pulled from OCI registries and started
     #[clap(long = "allow-latest", env = "WASMCLOUD_OCI_ALLOW_LATEST")]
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -61,7 +61,7 @@ struct Args {
     #[clap(long = "host-seed", env = "WASMCLOUD_HOST_SEED")]
     host_seed: Option<String>,
     /// Delay, in milliseconds, between requesting a provider shut down and forcibly terminating its process
-    #[clap(long = "provider-shutdown-delay", default_value = "300", env = "WASMCLOUD_PROV_SHUTDOWN_DELAY_MS", value_parser = parse_duration)]
+    #[clap(long = "provider-shutdown-delay", default_value = "300", env = "WASMCLOUD_PROV_SHUTDOWN_DELAY_MS", value_parser = parse_duration_millis)]
     provider_shutdown_delay: Duration,
```
vados-cosmonic, line 65: can we have this be `provider-shutdown-delay-ms`? an alias should work right?

src/main.rs
```diff
@@ -269,6 +269,10 @@ struct Args {
     /// Configures the set of certificate authorities as repeatable set of file paths to load into the OCI and OpenTelemetry clients
     #[arg(long = "tls-ca-path")]
     pub tls_ca_paths: Option<Vec<PathBuf>>,
+
+    /// If provided, overrides the default heartbeat interval of every 30 seconds. Provided value is interpreted as seconds.
+    #[arg(long = "heartbeat-interval", env = "WASMCLOUD_HEARTBEAT_INTERVAL",value_parser = parse_duration_secs, hide = true)]
```
vados-cosmonic, line 274: `heartbeat-interval-seconds`?

vados-cosmonic (approved):

vados-cosmonic:
oh also you might want `chore(host): ...`

src/main.rs
joonas: I think that would make sense for consistency, happy to adjust 👍 

I do think an alias should work

vados-cosmonic (approved):
LGTM :rocket:

PR #2699, opened 2024-08-06
Proposal: Adding Go support to monorepo

lxfontes:
This is a scratch PR showing the changes & ergonomics when bringing Go into the monorepo.

Goals:
- Make it obvious where the Go code is
- Use standard Go tooling, without `GOPATH` manipulation
- Introduce vanity urls ( `go.wasmcloud.dev` )

Go code lives under `/go` and is setup using [Go workspaces](https://go.dev/doc/tutorial/workspaces)
```
 ❯ tree go -L 3
go.work <-- All Go Projects have to register here
go.work.sum
go  
└── provider   <-- Go Project
    ├── LICENSE
    ├── README.md
    ├── examples
    │   └── keyvalue-inmemory  <-- Another Go Project
    ├── go.mod
    ├── go.sum
    ├── host.go
    ├── link.go
    ├── logging.go
    ├── models.go
    ├── models_test.go
    ├── observability.go
    ├── options.go
    ├── provider.go
    ├── secrets.go
    └── secrets_test.go
```

# Moving the SDK

Moving the SDK is fairly straightforward:
- Update `go.mod` to `go.wasmcloud.dev/provider`
- Update key-value example to `go.wasmcloud.dev/provider/examples/keyvalue-inmemory`
- Add both to Go workspace `go.work`

We can setup type aliases in the previous repo pointing to new locations, and also increment the provider-sdk version.

# Creating Projects

- Create directory under `/go`. Example `/go/special-provider`
- Initialize the project. `cd go/special-provider; go mod init go.wasmcloud.dev/special-provider`
- Add the directory to go workspace. Add `use ./go/special-provider` to `go.work`

## Gotchas
Every project-within-project ( ex: provider sdk examples ) has to be registered in the Go workspace.
Failure to do so will result in a compilation failure
```
 ❯ go build ./provider/examples/keyvalue-inmemory
main module (go.wasmcloud.dev/provider) does not contain package go.wasmcloud.dev/provider/examples/keyvalue-inmemory
```

Vanity URLs setup.
They are per package, meaning our vanity urls must understand our package layout.

For example, given the 2 packages:
- `go.wasmcloud.dev/provider`
- `go.wasmcloud.dev/provider/logging`

It is not enough to serve a vanity url for `go.wasmcloud.dev/provider`, we need to serve 2 vanity urls:
- `go.wasmcloud.dev/provider` => `github.com/wasmCloud/wasmCloud/go/provider`
- `go.wasmcloud.dev/provider/logging` => `github.com/wasmCloud/wasmCloud/go/provider`

This can be seen in action in Opentelemetry's project:
```
❯ curl -s https://go.opentelemetry.io/otel | grep go-import
<meta name="go-import" content="go.opentelemetry.io/otel git https://github.com/open-telemetry/opentelemetry-go">

❯ curl -s https://go.opentelemetry.io/otel/codes | grep go-import
<meta name="go-import" content="go.opentelemetry.io/otel git https://github.com/open-telemetry/opentelemetry-go">
```

# Discussion

❓ Not clear how/where vanity urls will be configured yet. Github Pages is the top option so far, and it could be served from `go/vanity_urls` directory.

❓  ~Should the Go workspace be at the top-directory? We'd continue to host Go projects under `/go` directory, but wouldn't need to jump to `/go` to interact with them.~
Decision: Go workspace is at the git root directory


joonas:
> ❓ Not clear how/where vanity urls will be configured yet. Github Pages is the top option so far, and it could be served from `go/vanity_urls` directory.

I believe `wasmcloud.dev` domain was previously configured via https://github.com/wasmCloud/wasmcloud-dev-site, which has been archived. I think we have the opportunity to either do it from here, or alternatively from the https://github.com/wasmcloud/wasmcloud.com repo if it were to become/stay as the source of "all things web presence"

> ❓ Should the Go workspace be at the top-directory? We'd continue to host Go projects under /go directory, but wouldn't need to jump to /go to interact with them.

I think we should, having the workspace activated as soon as you clone the repository down seems like a nice convenience, unless there are downsides to doing it top level?

protochron:
> ❓ Should the Go workspace be at the top-directory? We'd continue to host Go projects under /go directory, but wouldn't need to jump to /go to interact with them.

I think having the workspace at the top level would be ok as long as the code is nested under `/go`. Without that nesting I'd be worried that our repo structure would be very disorganized.

It's probably easier to serve the vanity URL page out of this repo, but I also like the idea of keeping that all in the wasmcloud/wasmcloud.com repo. We could have an action that could fire off whenever we add new modules here to make sure we keep the generated HTML in sync.

brooksmtownsend:
Thank you @lxfontes for putting this together! I'm a huge fan of moving the Go libraries here so we can start creating some common libraries for types we use all over the place, and possibly even start using codegen to, say, define wasmCloud events in JSON and then create go/rust libraries from both.

I have no hard preference regarding putting it at the top level of the repo or nesting it under `go/`, and I feel the discussion above seems to lean towards putting it under `go/` so I'm fine with that too.

.github/workflows/go-validation.yml
```diff
@@ -0,0 +1,70 @@
...
+      - "go.work"
+      - "go.work.sum"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "go/*"
+      - "go.work"
+      - "go.work.sum"
+
```
joonas, line 18: it would be good to add `permissions` at the top level since this is net-new:
```suggestion
permissions:
  contents: read
```
I think just `contents: read` should suffice in this case since this workflow just needs to be able to check out the repository.

lxfontes:
Our desired structure (example https://github.com/lxfontes/go-get ) can't be translated to `go.wasmcloud.dev/<package>` directly without the use of a custom proxy ( see https://go.dev/ref/mod#module-proxy ).

For the time being will add the vanity url pointing to the existing provider-sdk-go repository and reassess in the future.

Issue #2475, opened 2024-07-12
[FEATURE] runtime async


rvolosatovs:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

In #2459 we fully await I/O before returning the control back to the component

### Describe the solution you'd like

For async values (resources), we should move the I/O into the table and let the component drive it (by e.g. reading the stream)

PR #2709, opened 2024-08-07
fix: resource future table blobstore

brooksmtownsend:
This PR is meant to spur a little discussion, and it contains two fixes.

1. Awaiting the blobstore `get_container_data` async I/O future inside of a tokio task, and storing that task in the resource table of the component
2. Setting blobby's `ResponseOutparam` before trying to write the body of the file.
	1. This fix isn't complete yet and contains some commented out code where blobby attempts to send back 500 errors when we hit component issues, which is good 

## Feature or Problem
Without either of these fixes, using wasmCloud 1.1.0 and blobstore-fs 0.8.0, any object that is bigger than 524kb ends up hanging at the `get_container_data` async I/O future and never returns. This is, of course, a problem, but it is isolated to the component.

With the fix of awaiting the blobstore future inside of a task, and putting that future inside of the resource table (which I believe is what @rvolosatovs alluded to as a need but I might be misinterpreting) I can download blobs of ~1.2Mb

~With the blobby fix, I can download blobs of ~2Mb before I fail to write all the data. Depending on how inefficient blobby is, this might just approach the memory limit we have set for components /shrug.~

After the most recent force push, I've updated the code to use `splice`, which streams data directly instead of reading it entirely into memory. With this I've been able to use blobby to stream a 890megabyte file.

This is all strange because, with a pure HTTP component, I've been able to stream hundreds of megabytes through a component without any issues. So I believe this is either a bug in how we're transmitting data, or a lack of streaming in blobby specifically. If it's the latter, I'd like to fix the issue so that we can show examples of the right usage here.

## Related Issues
Attempting to address https://github.com/wasmCloud/wasmCloud/issues/2475

### Manual Verification
Here's the blobby component uploading & downloading a 2.3Gb file
```bash
➜ curl -H 'Content-Type: application/gzip' -v 'http://127.0.0.1:8080/container.tar.gz' -T big.tar.gz -X POST
*   Trying 127.0.0.1:8080...
* Connected to 127.0.0.1 (127.0.0.1) port 8080
> POST /container.tar.gz HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/8.7.1
> Accept: */*
> Content-Type: application/gzip
> Content-Length: 2292479488
> Expect: 100-continue
>
< HTTP/1.1 100 Continue
<
* upload completely sent off: 2292479488 bytes
< HTTP/1.1 201 Created
< vary: origin, access-control-request-method, access-control-request-headers
< access-control-allow-origin: *
< access-control-expose-headers: *
< transfer-encoding: chunked
< date: Thu, 08 Aug 2024 14:42:15 GMT
<
* Connection #0 to host 127.0.0.1 left intact

➜ curl 'http://127.0.0.1:8080/container.tar.gz' --output big2.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2186M    0 2186M    0     0   110M      0 --:--:--  0:00:19 --:--:--  113M

# no difference
➜ diff big.tar.gz big2.tar.gz
🛁 wash 0.30.0 took 9s
```

### DIFF
--- a/crates/runtime/src/component/blobstore.rs
+++ b/crates/runtime/src/component/blobstore.rs
@@ -13,13 +13,13 @@ use std::sync::Arc;
 
 use anyhow::{bail, Context as _};
 use async_trait::async_trait;
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use futures::{stream, StreamExt as _};
 use tokio::io::{AsyncReadExt as _, AsyncSeekExt};
 use tokio_util::io::StreamReader;
 use tracing::instrument;
 use wasmtime::component::Resource;
-use wasmtime_wasi::pipe::{AsyncReadStream, AsyncWriteStream};
+use wasmtime_wasi::pipe::AsyncWriteStream;
 use wasmtime_wasi::{HostOutputStream, InputStream};
 
 type Result<T, E = Error> = core::result::Result<T, E>;
@@ -97,20 +97,16 @@ where
         .await?
         {
             (Ok(stream), io) => {
-                if let Some(io) = io {
-                    // TODO: Move this into the runtime
-                    let handle = tokio::spawn(async move {
-                        if let Err(e) = io.await {
-                            tracing::error!("Error awaiting async IO: {:?}", e);
-                        }
-                    });
-                    self.table
-                        .push(handle)
-                        .context("failed to push async I/O")?;
-                }
+                let io = if let Some(io) = io {
+                    Box::new(io) as Box<dyn futures::Future<Output = anyhow::Result<()>> + Send>
+                } else {
+                    Box::new(async { Ok(()) })
+                        as Box<dyn futures::Future<Output = anyhow::Result<()>> + Send>
+                };
+                let input_streamer = HostInputStreamer::new(stream, io);
                 let value = self
                     .table
-                    .push(stream)
+                    .push(input_streamer)
                     .context("failed to push stream and size")?;
                 Ok(Ok(value))
             }
@@ -393,11 +389,13 @@ impl<H: Handler> types::HostIncomingValue for Ctx<H> {
         &mut self,
         incoming_value: Resource<types::IncomingValue>,
     ) -> anyhow::Result<Result<types::IncomingValueSyncBody>> {
-        let stream = self
+        let streamer = self
             .table
             .delete(incoming_value)
-            .context("failed to delete incoming value")?;
-        Ok(Ok(stream.collect::<Vec<Bytes>>().await.concat()))
+            .context("failed to get incoming value")?;
+
+        streamer.io.await.context("failed to perform async I/O")?;
+        Ok(Ok(streamer.buf.into()))
     }
 
     #[instrument(skip(self))]
@@ -409,11 +407,10 @@ impl<H: Handler> types::HostIncomingValue for Ctx<H> {
             .table
             .delete(incoming_value)
             .context("failed to delete incoming value")?;
+
         let stream = self
             .table
-            .push(InputStream::Host(Box::new(AsyncReadStream::new(
-                StreamReader::new(stream.map(std::io::Result::Ok)),
-            ))))
+            .push(InputStream::Host(Box::new(stream)))
             .context("failed to push input stream")?;
         Ok(Ok(stream))
     }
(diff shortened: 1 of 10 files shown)


ricochet (commented):
The LGTM. I'd like @rvolosatovs's review before we merge

crates/runtime/src/component/blobstore.rs
```diff
@@ -99,7 +99,14 @@ where
...
                     // TODO: Move this into the runtime
-                    io.await.context("failed to perform async I/O")?;
+                    let handle = tokio::spawn(async move {
+                        if let Err(e) = io.await {
+                            tracing::error!("Error awaiting async IO: {:?}", e);
+                        }
+                    });
+                    self.table
+                        .push(handle)
+                        .context("failed to push async I/O")?;
```
rvolosatovs, line 109: Looks like this would make consuming data non-deterministic, reading: nothing, complete buffer or anywhere in between. In other words, this would introduce a race condition between `io.await` being polled until it's ready by Tokio runtime and the component consuming the value (https://github.com/wasmCloud/wasmCloud/pull/2709/files#diff-87c740fda48fb006ffde1e67e879ab20a9b75aed0c38247f8edd4c7b5f7b9b5fR391-R419).

I think what should happen here is that we should:
1. move `io` future as-is (without awaiting it in any way or spawning a Tokio task) into the table
2. update `incoming_value_consume_sync` to `await` `io`, basically move the L102 from existing implementation there
3. update `incoming_value_consume_async` to return a new stream type (`HostInputStream` implementation https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/trait.HostInputStream.html), which in itself would:
    - hold the `io` future
    - use [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) in [`Subscribe::ready`](https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/trait.Subscribe.html#tymethod.ready) implementation to poll I/O and check for more data

examples/rust/components/blobby/src/wrapper.rs
```diff
@@ -2,111 +2,38 @@
...
-                        "more bytes read than requested",
-                    ));
+use crate::wasi::{
+    io::streams::{InputStream, OutputStream, StreamError},
+    logging::logging::{log, Level},
+};
+
+/// Helper function to read all the data from the input stream and write it to the output stream
+/// Optionally takes a length to limit the number of bytes read from the input stream
+pub fn stream_input_to_output(
```
rvolosatovs, line 12: how about we use https://docs.rs/wasi/latest/wasi/ and https://doc.rust-lang.org/std/io/fn.copy.html instead?

rvolosatovs (changes requested):
only requesting changes because of the race condition

crates/runtime/src/component/blobstore.rs
rvolosatovs: Actually, I was wrong in the original message and there's no race condition here, the Tokio task may get leaked, however, it would be nice to use a `JoinSet` and store that in the table if we go with the proposed approach, this way at least the task would be dropped once the component instance is dropped

rvolosatovs (approved):
I was actually wrong, there's no race condition here, but we may leak a Tokio task, which I guess is fine for a quickfix, but we should certainly address that

rvolosatovs (changes requested):
Sorry for going back and forth on this one, but after giving it a bit more thought, I think we do need to change the implementation here due to the non-determinism introduced here still, but only in I/O error case.

The I/O future polling cannot be done in a detached Tokio task as done here without communucating an error to consumer in some way, that is because doing so would allow for short reads and silent failures, which could break application logic (with potentially devastating effects).

I think we need to either not change this or use a custom stream like I outlined in my original review

brooksmtownsend:
@rvolosatovs apologies for leaving this for a few days, I completely agree with your proposed solution. I had a feeling the task solution wasn't quite right, but I wanted to put it up anyways since it helped the implementation

brooksmtownsend:
Just a note @thomastaylor312 and I hacked on this for a bit and pushing up what we had

rvolosatovs:
assigning myself to fix any outstanding issues here, will push to this branch

brooksmtownsend:
Handled in #2779

PR #2713, opened 2024-08-08
feat(host)!: allow tuning runtime parameters

brooksmtownsend:
## Feature or Problem
This PR surfaces additional tuning parameters that can be used to control limits on components in the wasmtime runtime, including the max number of component instances, the max size of a component instance, and the maximum amount of memory a component instance could allocate.

## Release Information
Could argue for a v1.1.2

### DIFF
--- a/crates/host/src/wasmbus/host_config.rs
+++ b/crates/host/src/wasmbus/host_config.rs
@@ -7,6 +7,7 @@ use std::time::Duration;
 use nkeys::KeyPair;
 use url::Url;
 use wasmcloud_core::{logging::Level as LogLevel, OtelConfig};
+use wasmcloud_runtime::{MAX_COMPONENTS, MAX_COMPONENT_SIZE, MAX_LINEAR_MEMORY};
 
 /// wasmCloud Host configuration
 #[allow(clippy::struct_excessive_bools)]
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -176,6 +176,19 @@ struct Args {
     /// If provided, allows to set a custom Max Execution time for the Host in ms.
     #[clap(long = "max-execution-time-ms", default_value = "600000", env = "WASMCLOUD_MAX_EXECUTION_TIME_MS", value_parser = parse_duration_millis)]
     max_execution_time: Duration,
+    /// The maximum amount of memory that a component can allocate
+    #[clap(long = "max-linear-memory", default_value_t = 10 * 1024 * 1024, env = "WASMCLOUD_MAX_LINEAR_MEMORY")]
+    max_linear_memory: u64,
+    /// The maximum size of a component binary that can be loaded
+    #[clap(long = "max-component-size", default_value_t = 50 * 1024 * 1024, env = "WASMCLOUD_MAX_COMPONENT_SIZE")]
```
vados-cosmonic, line 183: What do you think about dding the units for the param names (ex. `max-linear-memory-bytes`, etc)?

crates/host/src/wasmbus/host_config.rs
```diff
@@ -108,6 +114,11 @@ impl Default for Host {
             secrets_topic_prefix: None,
             version: env!("CARGO_PKG_VERSION").to_string(),
             max_execution_time: Duration::from_millis(10 * 60 * 1000),
+            // 10 MB
+            max_linear_memory: 10 * 1024 * 1024,
+            // 50 MB
+            max_component_size: 50 * 1024 * 1024,
+            max_components: 10000,
```
vados-cosmonic, line 121: These might be nice as constants somewhere

vados-cosmonic (approved):

src/main.rs
brooksmtownsend: Yeah I like that idea since it's not a human readable thing

crates/host/src/wasmbus/host_config.rs
brooksmtownsend: Added!

PR #2715, opened 2024-08-08
test: test interfaces with 100kb payload

brooksmtownsend:
## Feature or Problem
This is a WIP PR to demonstrate an issue with writing larger (65kb+) payloads to the blobstore output stream when using the OutputStreamWriter.

## Testing
https://github.com/wasmCloud/wasmCloud/actions/runs/10308267947/job/28535303371?pr=2715#step:4:2620 is where the error occurs


brooksmtownsend:
Combined with #2709, I can easily up the memory to a MB no big deal. If we run this with #2713, we can tune the test & go larger.

Issue #2037, opened 2024-05-01
Long-running wasmCloud workloads


brooksmtownsend:
This is a placeholder for an incoming RFC for long-running wasmCloud workloads. This somewhat succeeds #697 as it's a more general solution, but both may exist.

brooksmtownsend:
For folks looking for information on long running wasmCloud workloads, there are two things limiting workloads from executing for a long time:
1. A general execution timeout for components, represented as an epoch timer, that caps execution of a component to 10 minutes
2. An RPC timeout for all invocations (component -> component, component -> provider, provider -> component) which serves as a circuit-breaker pattern for when an entity fails to respond. We don't want an individual component to wait forever for a provider that panicked, so the RPC timeout prevents that.

If you don't mind losing the circuit breaker, you can set the `WASMCLOUD_RPC_TIMEOUT_MS=<>` environment variable, or associated `--rpc-timeout-ms` flag, to increase the time that you allow requests to be handled.

Since this didn't get proper attention in the wasmCloud 1.1 roadmap, I'm pulling it off for our next iteration. What we really need here to prioritize this is a couple of concrete use cases.

PR #2722, opened 2024-08-09
feat(examples): add task manager component

### LINKED ISSUE #2729: Demo App: Task Handler
The task handler app should manage tasks with a db interface and manage requests for task status

vados-cosmonic:
Closes #2729


brooksmtownsend:
Does this address part of #2037 ? 😮

vados-cosmonic:
It definitely *could*! I think the main problem here is that we need something that almost pings components on a daily cadence if we want  a consistent task *runner* (a provider/host integration), this is mostly only to do with tracking and creating new tasks to be done -- quite generic I think!

examples/rust/components/http-task-manager/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Building
+
+You can build the application with the [WAsmcloud SHell `wash`][wash]:
+
+```console
+wash build
+```
+
+Note that if you'd like to build with `cargo` you need to sepcify the `--target` option:
```
brooksmtownsend, line 41:
```suggestion
Note that if you'd like to build with `cargo` you need to specify the `--target` option:
```

examples/rust/components/http-task-manager/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+```console
+cargo build --target=wasm32-wasi
+```
```
brooksmtownsend, line 45: Thoughts on adding the nightly tag here and targeting wasm32-wasip2?

examples/rust/components/http-task-manager/README.md
```diff
@@ -0,0 +1,116 @@
...
+    postgres:16-alpine
+```
+
+After the Postgres instance is running, we should migrate it to a state where the component can use it. We can do this by executing the migration script embedded in this folder:
+
+```console
+cat sql/migrations/0-init.sql | docker run -it --name pg psql -- -f -
+``` 
+
+Enabling the application to *connect* to Postgres requires setting up some secrets, which we can set with `wash`:
```
brooksmtownsend, line 68:
```suggestion
Enabling the application to *connect* to Postgres requires setting up some configuration, which we can set with `wash`:
```

examples/rust/components/http-task-manager/src/http.rs
```diff
@@ -0,0 +1,272 @@
...
+                        bytes.len() <= MAX_READ_BYTES,
+                        "read more bytes than requested"
+                    );
+                    buf.extend(bytes);
+                }
+                Err(StreamError::Closed) => break,
+                Err(e) => bail!("failed to read bytes: {e}"),
+            }
+        }
+        drop(incoming_req_body_stream);
```
brooksmtownsend, line 96: I think this block could be improved a lot by using the `wasi=0.13.1` crate & just using `io::copy` to copy the stream to the other, but that's fine as a followup!

examples/rust/components/http-task-manager/src/http.rs
```diff
@@ -0,0 +1,272 @@
...
+        log(
+            Level::Error,
+            LOG_CONTEXT,
+            &format!(
+                "invalid error response length, {} bytes is too long for blocking-write-and-flush",
+                bytes.len()
+            ),
+        );
+        bytes = b"ERROR: see logs for details";
+    }
```
brooksmtownsend, line 119: Same comment re: streams & using the `io::Write` implementations, but since sending error responses is something we control it's less of a big deal

examples/rust/components/http-task-manager/src/http.rs
```diff
@@ -0,0 +1,272 @@
...
+                LOG_CONTEXT,
+                500,
+                &format!("failed write image chunk to output stream: {e}"),
+                response_out,
+            );
+            return;
+        }
+    }
+
+    ResponseOutparam::set(response_out, Ok(response));
```
brooksmtownsend, line 264: Make sure to set this before doing the write to the response stream

examples/rust/components/http-task-manager/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+This web component tracks and updates information about asynchronous tasks, backed by a [Postgres][postgres] Database.
+
+Clients can make HTTP requests to endpoints like the following:
+
+| Endpoint                 | Description                   |
+|--------------------------|-------------------------------|
+| `GET /tasks`             | List ongoing tasks            |
+| `POST /tasks/:id/status` | Get status of a specific task |
+| `POST /tasks`            | Submit a new task             |
```
brooksmtownsend, line 11: do we want to note the `/api/v1` prefix here?

brooksmtownsend (changes requested):
A couple nits & future improvement thoughts, but just a couple requests. Overall this is nice!

examples/rust/components/http-task-manager/README.md
brooksmtownsend: Just thinking mostly that getting a `wasm32-wasi` module is only part of the build picture so may as well depend on nightly & get the real real component

vados-cosmonic: In general, I'd like to not require nightly for the build

But ideally would use `wash build`, so they don't have to worry about it -- the reason this is here versus all the other components (which just have `.cargo/config.toml` set up) is that right now the E2E tests need the target to *not* be `wasm32-*` to run properly under `cargo test`

examples/rust/components/http-task-manager/src/http.rs
vados-cosmonic: Ahh this can actually be removed in favor of the more general one!

examples/rust/components/http-task-manager/src/http.rs
vados-cosmonic: Oh yeah, I absolutely saw the changes ya'll made to do use the `InputStream` stuff earlier I can definitely use it here too

examples/rust/components/http-task-manager/src/http.rs
vados-cosmonic: Ah you actually don't have to! With this code you actually don't have to explicit drop either.

Which part were you thinking didn't work? Becuase the writing to the stream happens *before* this actually

examples/rust/components/http-task-manager/README.md
vados-cosmonic: Ah yes, thanks, will update here, all these endpoints are outdated actually (and ab bunch are missing)

examples/rust/components/http-task-manager/src/http.rs
brooksmtownsend: Oh I didn't try it and see it fail, but I thought this was required as you may just fill up the internal buffer if you don't set the response out beforehand. No change needed if it all works!

brooksmtownsend (approved):

PR #2726, opened 2024-08-09
fix(examples): daemonscale http provider in all examples

ricochet:
At least until spreadscaled http providers are supported by the
operator to generate service endpoints, we should make all of the
examples use the Daemonscaler. This allows examples to work with
a simple kind deploy by default.

https://github.com/wasmCloud/wasmCloud-operator?tab=readme-ov-file#automatically-syncing-kubernetes-service

## Feature or Problem

Adds daemonscaler to all HTTP provider examples so that examples work OOTB with k8s deploys.

### DIFF
--- a/examples/rust/components/http-hello-world/wadm.yaml
+++ b/examples/rust/components/http-hello-world/wadm.yaml
@@ -15,6 +15,8 @@ spec:
     - name: http-component
       type: component
       properties:
+        # NOTE: this example is used in the quickstart guide,
+        # we keep the local file ref enabled instead of OCI like other examples.
         image: file://./build/http_hello_world_s.wasm
         # To use the a precompiled version of this component, use the line below instead:
         # image: ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0
@@ -44,3 +46,6 @@ spec:
               - name: default-http
                 properties:
                   address: 127.0.0.1:8080
+        - type: daemonscaler
+          properties:
+            replicas: 1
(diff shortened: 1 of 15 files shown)


examples/rust/components/http-hello-world/wadm.yaml
```diff
@@ -15,9 +15,9 @@ spec:
     - name: http-component
       type: component
       properties:
-        image: file://./build/http_hello_world_s.wasm
+        # image: file://./build/http_hello_world_s.wasm
         # To use the a precompiled version of this component, use the line below instead:
-        # image: ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0
+        image: ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0
```
ricochet, line 20: Heads up that this was a case were we referred to the local image and not the OCI ref. I think this one was just accidentally missed but giving it a shoutout in case I'm wrong.

brooksmtownsend: This should be the local ref because it's what we use in our tutorials, and our tutorials go right into modifying the wasm

ricochet: TY, added a note. We'll need to use a different example for the operator quickstart

brooksmtownsend (changes requested):
Because of https://wasmcloud.com/docs/tour/scale-and-distribute, I have a slight preference to keep the http-hello-world versions as their spreadscaler equivalent since the daemonscaler will cause port conflicts with multiple local hosts

ricochet:
I updated the kind quickstart to use this: https://github.com/wasmCloud/wasmcloud-operator/pull/87, so long as the demos that benefit from the daemonscaler have a wadm for it, then we don't need to modify the examples here (another [example in wasmcloud-contrib](https://github.com/wasmCloud/wasmCloud-contrib/pull/11))

PR #2763, opened 2024-08-13
feat(component): add join helper

brooksmtownsend:
## Feature or Problem
I was doing some experimentation with sending off multiple HTTP requests, and I created this little helper that polls a set of pollables to completion. Is there an appetite to including this in our component library?

I'm fine with closing this, refactoring, etc, but it does seem useful for creating polling multiple futures to completion.

### DIFF
--- a/crates/component/src/wrappers/io.rs
+++ b/crates/component/src/wrappers/io.rs
@@ -119,6 +119,21 @@ impl Default for StdioStream<'_> {
     }
 }
 
+/// Similar to [`crate::wasi::io::poll::poll`], but polls all `pollables` until they are all ready.
+///
+/// Poll for completion on a set of pollables.
+///
+/// This function takes a list of pollables, which identify I/O sources of interest, and waits until all of the events are ready for I/O.
+pub fn join(pollables: &[&crate::wasi::io::poll::Pollable]) {
+    let mut pollables = pollables.to_vec();
+    while !pollables.is_empty() {
+        let ready_indices = crate::wasi::io::poll::poll(&pollables);
+        ready_indices.iter().rev().for_each(|&i| {
+            pollables.swap_remove(i as usize);
+        });
+    }
+}
+
 #[cfg(feature = "futures")]
 impl futures::AsyncRead for StdioStream<'_> {
     fn poll_read(


crates/component/src/wrappers/io.rs
```diff
@@ -119,6 +119,26 @@ impl Default for StdioStream<'_> {
     }
 }
 
+/// Similar to [`crate::wasi::io::poll::poll`], but polls all `pollables` until they are all ready.
+///
+/// Poll for completion on a set of pollables.
+///
+/// This function takes a list of pollables, which identify I/O sources of interest, and waits until all of the events are ready for I/O.
+pub fn poll_all(pollables: &[&crate::wasi::io::poll::Pollable]) {
```
brooksmtownsend, line 127: Will have to test with a `with: {}` block in order to see this working with wasmcloud_component 😄

crates/component/src/wrappers/io.rs
```diff
@@ -119,6 +119,26 @@ impl Default for StdioStream<'_> {
...
+    let mut pollables = pollables.to_vec();
+    while !pollables.is_empty() {
+        let ready_indices = crate::wasi::io::poll::poll(&pollables);
+        pollables = pollables
+            .into_iter()
+            .enumerate()
+            .filter_map(|(i, p)| {
+                // Retain the pollables that didn't return ready
+                ready_indices.contains(&(i as u32)).then_some(p)
```
rvolosatovs, line 136:
```suggestion
                (!ready_indices.contains(&(i as u32))).then_some(p)
```
This would need to return `Some` for all pollables that are *not* ready, right?

crates/component/src/wrappers/io.rs
```diff
@@ -119,6 +119,26 @@ impl Default for StdioStream<'_> {
...
+            })
+            .collect();
```
rvolosatovs, line 138: This could also be a loop over `ready_indices.rev()`, which would call `remove` on the vec.

brooksmtownsend: Oh yeah I'm a fan of that

PR #2777, opened 2024-08-15
feat(wash-cli): Add `clap-markdown` to generate CLI docs in Markdown

ericgregory:
This PR adds [`clap-markdown`](https://github.com/ConnorGray/clap-markdown/tree/main) to wash, enabling us to use the ~~`--markdown-help`~~ `--help-markdown` argument in a wash command to generate comprehensive Markdown-formatted docs for the CLI based on the help content. This will let us largely automate CLI doc generation without manually updating the page for each subcommand. 

* [Example Markdown output](https://gist.github.com/ericgregory/1ed50febe05629e826756bacf9405004#wash-app). 
* The most straightforward way to run this is probably `wash app list --help-markdown > help.md`


joonas:
Does this generate docs for just the command that `--markdown-help` is passed in for, or all of the commands?

If it's for all commands, should this perhaps become it's own separate command? Maybe something like `wash generate-docs`?

ericgregory:
> Does this generate docs for just the command that `--markdown-help` is passed in for, or all of the commands?
> 
> If it's for all commands, should this perhaps become it's own separate command? Maybe something like `wash generate-docs`?

It generates one doc for all commands. The argument implementation was easiest for a clap newbie like me (and is the crate's suggested convention), but it could totally be a command as well.

vados-cosmonic:
Hey @ericgregory you probably want to run `cargo fmt` and also rebase this to `main` ([to get the fix to nix pipelines that just merged](https://github.com/wasmCloud/wasmCloud/pull/2782))

crates/wash-cli/src/bin/wash.rs
```diff
@@ -120,6 +121,13 @@ struct Cli {
     )]
     pub(crate) experimental: bool,
 
+    #[clap(
+        long = "markdown-help", 
+        hide = true,
+        global = true
+    )]
+    markdown_help: bool,
```
vados-cosmonic, line 129: What do you think about calling this something like `--help-output-markdown` and maybe mark it as [required with `--help`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires) ? 

That *should* force usage like this:

```console
wash app list --help --help-output-markdown
```

I'm also definitely open to `--help-markdown` and [maybe setting `conflicts_with`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with), if the above is too verbose!

crates/wash-cli/src/bin/wash.rs
```diff
@@ -225,6 +233,14 @@ async fn main() {
         .with_env_filter(EnvFilter::from_default_env())
         .init();
 
+    // Implement clap_markdown for markdown generation of command line documentation
+    let mdargs = Cli::parse();
+
+    // Most straightforward way to invoke is probably `wash app list --markdown-help > help.md`
+    if mdargs.markdown_help {
+        clap_markdown::print_help_markdown::<Cli>();
+    };
```
vados-cosmonic, line 242: If someone has this set, can we assume this is all they want to do? As in, should we exit immediately after? 

Also, I wonder if we can mvoe that `Cli::parse();` *inside* the `if`?

brooksmtownsend:
We may want to drop the merge commit here & do a rebase, I'm unsure if our rules will allow this to merge as-is

crates/wash-cli/src/bin/wash.rs
```diff
@@ -236,6 +245,15 @@ async fn main() {
...
 
+    // Implement clap_markdown for markdown generation of command line documentation
+    let mdargs = Cli::parse();
+
+    // Most straightforward way to invoke is probably `wash app list --markdown-help > help.md`
+    if mdargs.markdown_help {
+        clap_markdown::print_help_markdown::<Cli>();
+        std::process::exit(0);
+    };
+
```
brooksmtownsend, line 256: I think, if I understand this correctly, we should move this down to line ~388 to avoid double parsing the CLI. Thoughts?

brooksmtownsend (changes requested):

ericgregory:
> Hey @ericgregory you probably want to run `cargo fmt` and also rebase this to `main` ([to get the fix to nix pipelines that just merged](https://github.com/wasmCloud/wasmCloud/pull/2782))
> 
> (pretty sure the test failures are all spurrious!)

Done and looks like all checks are passing now!

crates/wash-cli/src/bin/wash.rs
ericgregory: It looks like it gives incorrect output if you run the argument with the --help argument, so I took the `conflicts_with` approach with the most recent commit (and changed the argument to `--help-markdown`).

vados-cosmonic: Ah thanks for getting to this!

crates/wash-cli/src/bin/wash.rs
ericgregory: Moved and adjusted to avoid the double-parse! Just let me know if another spot would be better.

crates/wash-cli/src/bin/wash.rs
ericgregory: Added an exit in the most recent commit! My understanding is the parse needs to be outside the if, but I ended up removing the line (and moving the if statement down) to avoid double-parsing per Brooks' suggestion below.

crates/wash-cli/src/bin/wash.rs
```diff
@@ -98,19 +99,8 @@ Options:
...
-        clap::crate_version!(),
-        NATS_SERVER_VERSION,
-        WADM_VERSION,
-        WASMCLOUD_HOST_VERSION
-    )
-}
-
 #[derive(Debug, Clone, Parser)]
-#[clap(name = "wash", version = version(), override_help = HELP)]
+#[clap(name = "wash", version, override_help = HELP)]
```
brooksmtownsend, line 103: Sorry @ericgregory looks like this reverted a few changes 😢 this may have been a rebase issue

brooksmtownsend (changes requested):

crates/wash-cli/src/bin/wash.rs
ericgregory: Changes should be restored in the latest commit!

brooksmtownsend (approved):

PR #2779, opened 2024-08-15
feat: implement async in blobstore, use WASI and wit-bindgen 0.30

rvolosatovs:
## Feature or Problem

- implement `wasi:blobstore` streaming in providers, runtime, tests and examples
- update `wasi:blobstore`
- update `wrpc:blobstore`
- use wit-bindgen 0.30 everywhere
- use `wasi` crate where possible
- run tests with longer payloads

There are still a few things to update, but that can be done in follow-ups

## Related Issues

refs #2778 
refs https://github.com/wasmCloud/wasmCloud/issues/2475
refs https://github.com/wasmCloud/wasmCloud/pull/2709

## Release Information

Not sure, this will require new versions of all blobstore providers.

## Consumer Impact

The host will be able to use old blobstore providers (using `wrpc:blobstore@0.1.0`), but new providers will not work on old hosts


brooksmtownsend (approved):
This all looks good to me, glad to see an example of backwards compatibility on the wRPC interface side. I think we won't need to do that in all cases, but it will be really nice to at least preserve the backwards compatibility for a minor version to allow implementers to update.

Issue #2788, opened 2024-08-19
[FEATURE] Abort some wash drain subcommands if wasmCloud host is Running


Aditya1404Sal:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

When I am testing or using wasmCloud, I sometimes forget that I have a host running in detached mode and still run the `wash drain --all` command. This results in the caches being cleared and the PID files (`wasmcloud.pid` and `wadm.pid`) being deleted. Consequently, the `wash down --all` command fails to work properly because it can't find the wadm.pid file, forcing me to manually kill the remaining processes. This is both inconvenient and error-prone.

### Describe the solution you'd like

I would like the wash drain command to first check if a wasmCloud process is currently running. If no processes are running, the command should proceed with clearing the cache as normal. However, if a wasmCloud process is detected, the command should abort and display an error message, prompting the user to shut down wasmCloud before clearing the caches. This would prevent the deletion of critical PID files while the host is still active and avoid the need for manual intervention.

PR #2790, opened 2024-08-19
feat(wash-cli): Modify wash drain to avoid clearing downloads cache when host is running

Aditya1404Sal:
## Feature or Problem
Aborts `drain all` and `drain downloads` if wasmcloud host is active

## Related Issues
#2788 

### Manual Verification
I started a wasmcloud instance via `cargo run -- up -d` then tried to run `cargo run -- drain all` which resulted in an error
then ran `down --all` and `drain all` which ran as expected

### DIFF
--- a/crates/wash-cli/src/drain.rs
+++ b/crates/wash-cli/src/drain.rs
@@ -3,9 +3,18 @@ use std::collections::HashMap;
 use anyhow::Result;
 use serde_json::json;
 use wash_lib::cli::CommandOutput;
+use wash_lib::config::{host_pid_file, wadm_pid_file};
 use wash_lib::drain::Drain;
 
-pub fn handle_command(cmd: Drain) -> Result<CommandOutput> {
+pub fn handle_command(cmd: Drain) -> Result<CommandOutput, anyhow::Error> {
+    if matches!(cmd, Drain::All | Drain::Downloads) {
+        let wasmcloud_pid_path = host_pid_file().unwrap();
+        let wadm_pid_path = wadm_pid_file().unwrap();
+
+        if wasmcloud_pid_path.exists() || wadm_pid_path.exists() {
+            anyhow::bail!("Cannot clear the caches : wasmcloud and wadm PID files are still present on disk, processes might still be active or were not properly terminated.")
+        }
+    }
     let paths = cmd.drain()?;
     let mut map = HashMap::new();
     map.insert("drained".to_string(), json!(paths));
--- a/crates/wash-lib/src/config.rs
+++ b/crates/wash-lib/src/config.rs
@@ -12,6 +12,7 @@ pub const WASH_DIR: &str = ".wash";
 
 pub const DOWNLOADS_DIR: &str = "downloads";
 pub const WASMCLOUD_PID_FILE: &str = "wasmcloud.pid";
+pub const WADM_PID_FILE: &str = "wadm.pid";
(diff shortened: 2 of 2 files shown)


crates/wash-cli/src/drain.rs
```diff
@@ -1,11 +1,23 @@
...
 use serde_json::json;
 use wash_lib::cli::CommandOutput;
 use wash_lib::drain::Drain;
 
-pub fn handle_command(cmd: Drain) -> Result<CommandOutput> {
+pub fn handle_command(cmd: Drain) -> Result<CommandOutput, anyhow::Error> {
+    if matches!(cmd, Drain::All | Drain::Downloads) {
+        let wasmcloud_pid_path =
+            Path::new(&dirs::home_dir().unwrap()).join(".wash/downloads/wasmcloud.pid");
+        let wadm_pid_path = Path::new(&dirs::home_dir().unwrap()).join(".wash/downloads/wadm.pid");
```
vados-cosmonic, line 13: Just a note -- since you use `home_dir()` here in both places, maybe you can ensure it's present in one place:

```rust
let home_dir = dirs::home_dir().context("failed to determine home directory")?;
```

That said, it feels like we might actually want to use [functions like `host_pid_file()` here](https://github.com/wasmCloud/wasmCloud/blob/e3d7dcb1b9dca81fb0d84f9858ce3fcaa9093eb5/crates/wash-lib/src/config.rs#L26) -- I'd even be open to adding one for the `wadm` PID file.

crates/wash-cli/src/drain.rs
```diff
@@ -1,11 +1,23 @@
...
+
+        if wasmcloud_pid_path.exists() || wadm_pid_path.exists() {
+            return Err(anyhow::Error::msg(
+                "Cannot clear out the caches because wasmcloud is still up and running",
```
vados-cosmonic, line 17: here we can use `anyhow::bail` probably, I think. 

And, you might want to include `wadm` in the message since we chance for either `wasmcloud` or `wadm` instances.

We also might want to be *really* clear that the PID files are still on disk -- sometimes those files get left around and someone might not know *why* we think it's still running

Aditya1404Sal: I understand :+1: , I'll convert Error -> bail and Change the message to focus on the existence of PID files

crates/wash-cli/src/drain.rs
Aditya1404Sal: Got it. So i'll make use of `host_pid_file` plus add a new `wadm_pid_file`

Aditya1404Sal: Sounds great! I'll do the needful :+1:

crates/wash-cli/src/drain.rs
Aditya1404Sal: `Cannot clear the caches : wasmcloud and wadm PID files are still present on disk, processes might still be active or were not properly terminated.`

Is this message alright ?

vados-cosmonic:
Hey @Aditya1404Sal once you're ready for another review feel free to resolve the outstanding conversations and re-request a review!

Aditya1404Sal:
> Hey @Aditya1404Sal once you're ready for another review feel free to resolve the outstanding conversations and re-request a review!

Yep it's ready for a review!

vados-cosmonic (approved):
LGTM :rocket: 

Thanks @Aditya1404Sal !

PR #2808, opened 2024-08-20
fix: blobstore examples

rvolosatovs:
## Feature or Problem

With this I am able to serve https://github.com/wasmCloud/wasmCloud/releases/download/wash-cli-v0.30.0/wash-aarch64-apple-darwin (34.2 MB) just fine. I verified the checksum of this file streamed from FS provider.

The `=` in `wasi` crate dep specification is the fix - it is important that components use WASI 0.2.0.

`wasi` crate version `0.13.2` depends on `wasi 0.2.1`, which causes a deadlock in the component when, e.g. a function from `wasi:http/types` is called

## Related Issues

Closes https://github.com/wasmCloud/wasmCloud/pull/2709

### DIFF
--- a/examples/rust/components/blobby/Cargo.toml
+++ b/examples/rust/components/blobby/Cargo.toml
@@ -14,7 +14,7 @@ name = "blobby"
 anyhow = "1.0.81"
 form_urlencoded = "1.2.1"
 http = "1.1.0"
-wasi = "=0.13.1"
+wasi = "=0.13.1" # WASI 0.2.1 is not currently supported
 wit-bindgen = "0.30"
 
 [profile.release]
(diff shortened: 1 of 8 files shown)


examples/rust/components/blobby/Cargo.toml
```diff
@@ -14,7 +14,7 @@ name = "blobby"
 anyhow = "1.0.81"
 form_urlencoded = "1.2.1"
 http = "1.1.0"
-wasi = "0.13.1"
+wasi = "=0.13.1"
```
vados-cosmonic, line 17: A note here and in the other spots would be great so we can remember *why*

vados-cosmonic (commented):
LGTM! Just want to add a note in the code

examples/rust/components/blobby/Cargo.toml
rvolosatovs: https://github.com/wasmCloud/wasmCloud/pull/2808/commits/d66a86f9c8131677cac3f0b7bf1bb6afe6b82b78

PR #2809, opened 2024-08-20
chore(provider-sqldb-postgres): Remove the rustls option and enable rustls features by default

joonas:
## Feature or Problem

Turns out the `rustls` feature is a bit superfluous, because it gets [turned off by default](https://github.com/wasmCloud/wasmCloud/blob/main/Cargo.toml#L347) and [not enabled subsequently](https://github.com/wasmCloud/wasmCloud/blob/main/Cargo.toml#L128) even though we absolutely want to ship the provider with it enabled, so rather than juggling that, let's just remove the feature flag and always compile it in.

### DIFF
--- a/crates/provider-sqldb-postgres/Cargo.toml
+++ b/crates/provider-sqldb-postgres/Cargo.toml
@@ -11,10 +11,6 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
-[features]
-default = [ "rustls" ]
-rustls = [ "dep:tokio-postgres-rustls" ]
-
 [dependencies]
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):
No issues on my end with this, when we publish this crate we can revisit adding this feature flag back in

rvolosatovs (approved):

brooksmtownsend:
the failing checks in this PR would be resolved by https://github.com/jbg/tokio-postgres-rustls/pull/24 once we can grab a release of tokio-postgres-rustls that depends on a certificate library with an approved license

PR #2810, opened 2024-08-20
fix/blobby drop stream

brooksmtownsend:
## Feature or Problem
Quick fix to the blobby put-object function to drop the body

Quick fix to azure provider wasmcloud.toml for building


vados-cosmonic (approved):
LGTM :rocket:

PR #2828, opened 2024-08-21
refactor(blobby): use `wasi:io` directly

rvolosatovs:
## Feature or Problem

Following our discussion during a community call, switch to using `wasi:io` directly for forwarding data from one stream to another for (slightly) better performance.

In my totally scientific profiling of this using `blobstore-fs` and a bunch of random bytes written to disk:

before:
```
$ curl http://127.0.0.1:8000/test > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 3497M    0 3497M    0     0   133M      0 --:--:--  0:00:26 --:--:--  135M
```

after:
```
$ curl http://127.0.0.1:8000/test > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 5943M    0 5943M    0     0   138M      0 --:--:--  0:00:43 --:--:--  138M
```

The change is average download speed may be just noise and probably does not mean much, but certainly *not* passing bytes through the component binary should be faster than doing so.

One sunny day in the future [`std::io::copy`](https://doc.rust-lang.org/nightly/std/io/fn.copy.html) will just do this under the hood for us, but today is not the day yet.

## Consumer Impact

marginally faster `blobby`

### Manual Verification

tested with `blobstore-fs`, `http-server` and `blobby` from `main`

### DIFF
--- a/examples/rust/components/blobby/src/lib.rs
+++ b/examples/rust/components/blobby/src/lib.rs
@@ -21,7 +21,7 @@ mod bindings {
     export!(Blobby);
 }
 
-use std::io::{self, Read as _, Write as _};
+use std::io::Write as _;
 
 use http::{
     header::{ALLOW, CONTENT_LENGTH},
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
Ah I see what you meant now from the call! 

I do like the `io::copy` look better, but I'll take a speed improvement in a reusable component example over it

PR #2836, opened 2024-08-22
fix(ci): wit publishing workflow permissions

vados-cosmonic:


.github/workflows/wit-wasmcloud-postgres.yml
```diff
@@ -12,6 +12,9 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
```
brooksmtownsend, line 17: I think you _might_ need
```
contents: write
```
here, since the goal is to create a release not a PR right?

brooksmtownsend (changes requested):

.github/workflows/wit-wasmcloud-postgres.yml
vados-cosmonic: Ah sorry yes -- I need contents write *only*! `pull-requests` is entirely the wrong one

brooksmtownsend (approved):

PR #2847, opened 2024-08-23
fix(wash-lib): use resolved component ref for start

brooksmtownsend:
## Feature or Problem
This PR is a small fix to ensure that when we resolve the component_ref for a start or scale command, we use it both for the auction and the actual start.

## Release Information
next wash


vados-cosmonic (approved):
LGTM :rocket:

PR #2869, opened 2024-08-27
feat(wash-cli): add CLI styling to wash

vados-cosmonic:
## Feature or Problem

Adopt styling similar to cargo for wash CLI commands. This doesn't change our overridden help-text, but it does affect commands individually. We might want to consider changing/updating our overridden help/maybe not overriding it as well (or colorizing it ourselves).

### Before

before

### After

after

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6825,6 +6825,7 @@ dependencies = [
 name = "wash-cli"
 version = "0.31.0"
 dependencies = [
+ "anstyle",
  "anyhow",
  "assert-json-diff",
  "async-compression",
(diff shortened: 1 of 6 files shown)


brooksmtownsend (commented):
Very cool with the idea of colorizing our existing help! The reason we override it is just to organize the commands, if there was a clap-native way to do that I'm all for it

vados-cosmonic:
Yeah I figure we can address that in a follow up! The decision there isn't quite so straight forward

vados-cosmonic:
Here's that follow up issue: https://github.com/wasmCloud/wasmCloud/issues/2873

brooksmtownsend (approved):

PR #2871, opened 2024-08-27
chore(deps): update wit-bindgen-wrpc to 0.6.4

vados-cosmonic:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -375,7 +375,13 @@ wrpc-runtime-wasmtime = { version = "0.21", default-features = false }
 wrpc-transport = { version = "0.26.7", default-features = false }
 wrpc-transport-nats = { version = "0.22.3", default-features = false }
 
-# prettyplease (v0.2.22) is a tarnsitive dep of wit-binden-wrpc and uses a newer version of syn (2.0.76)
-# this conflicts with versions of syn resolved by other dependencies 
+# syn is a transitive dep of various packages and two of it's downstream dependencies differ:
+#
+# - serde_derive @ 1.0.209 (^2.0.46)
+# - prettyplease @ 0.2.22 (^2.0.76)
+#
+# Normally this would resolve itself, but since we combine lock files, it's possible to get the wrong version in some transitive dependencies sneaking in (and locked to a different version).
+#
+# To ensure that the resulting (combined) Cargo.lock file can be used, we force syn t a specific version
 [patch.crates-io]
-prettyplease = { git = "https://github.com/dtolnay/prettyplease", tag = "0.2.20" }
+syn = { git = "https://github.com/dtolnay/syn", tag = "2.0.76" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -374,3 +374,8 @@ wrpc-interface-http = { version = "0.27", default-features = false }
 wrpc-runtime-wasmtime = { version = "0.21", default-features = false }
 wrpc-transport = { version = "0.26.7", default-features = false }
 wrpc-transport-nats = { version = "0.22.3", default-features = false }
+
+# prettyplease (v0.2.22) is a tarnsitive dep of wit-binden-wrpc and uses a newer version of syn (2.0.76)
+# this conflicts with versions of syn resolved by other dependencies 
+[patch.crates-io]
+prettyplease = { git = "https://github.com/dtolnay/prettyplease", tag = "0.2.20" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -366,7 +366,7 @@ which = { version = "4", default-features = false }
 wit-bindgen = { version = "0.30", default-features = false }
 wit-bindgen-core = { version = "0.30", default-features = false }
 wit-bindgen-go = { version = "0.30", default-features = false }
-wit-bindgen-wrpc = { version = "0.6.3", default-features = false }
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):

Cargo.toml
```diff
@@ -366,11 +366,22 @@ which = { version = "4", default-features = false }
...
+# syn is a transitive dep of various packages and two of it's downstream dependencies differ:
+#
+# - serde_derive @ 1.0.209 (^2.0.46)
+# - prettyplease @ 0.2.22 (^2.0.76)
+#
+# Normally this would resolve itself, but since we combine lock files, it's possible to get the wrong version in some transitive dependencies sneaking in (and locked to a different version).
+#
+# To ensure that the resulting (combined) Cargo.lock file can be used, we force syn t a specific version
+[patch.crates-io]
+syn = { git = "https://github.com/dtolnay/syn", tag = "2.0.76" }
```
rvolosatovs, line 387: wouldn't this prevent us from releasing to `crates.io`? We also really should not need these kind of hacks

vados-cosmonic: Well first I'd like to actually get something working first -- it looks like this change in `syn` version is caused by the update to `wit-bindgen-wrpc` which updated `prettyplease` to a version that requires a newer `syn`.

vados-cosmonic: At present, it just looks like the registry information in Nix is just *old*:

```
wasmcloud-deps> error: failed to select a version for `syn`.

wasmcloud-deps>     ... required by package `serde_derive v1.0.208`

wasmcloud-deps>     ... which satisfies dependency `serde_derive = "=1.0.208"` (locked to 1.0.208) of package `serde v1.0.208`

wasmcloud-deps>     ... which satisfies dependency `serde = "^1"` (locked to 1.0.208) of package `wash-cli v0.31.0 (/build/source/crates/wash-cli)`

wasmcloud-deps> versions that meet the requirements `^2.0.46` (locked to 2.0.74) are: 2.0.74

wasmcloud-deps> all possible versions conflict with previously selected packages.

wasmcloud-deps>   previously selected package `syn v2.0.76`

wasmcloud-deps>     ... which satisfies dependency `syn = "^2.0.46"` (locked to 2.0.76) of package `async-trait v0.1.81`

wasmcloud-deps>     ... which satisfies dependency `async-trait = "^0.1.24"` (locked to 0.1.81) of package `redis v0.25.4`

wasmcloud-deps>     ... which satisfies dependency `redis = "^0.25"` (locked to 0.25.4) of package `wasmcloud v1.2.0 (/build/source)`
```

@rvolosatovs can you speak to that at all? `cargo` can resolve these dependencies just fine when building the workspace, but there's something about the nix setup that can't see the newer version of `syn` (and as a result locking to `2.0.74` instead of `2.0.76`.

vados-cosmonic: Ah to be very specific, the command I'm running (from one of [the failing tests in previous iterations on this branch](https://github.com/wasmCloud/wasmCloud/actions/runs/10578769624/job/29309640723)):

```
cargo check --release -j 16 --package wash-cli --package wasmcloud
```

vados-cosmonic:
Resolved by https://github.com/wasmCloud/wasmCloud/pull/2876/files

Cargo.toml
rvolosatovs: Currently, we have multiple workspaces in the repository, one for test components, one for the wasmcloud project and one per component example.

Since we build in a (Nix) sandbox without internet access, we need to fetch the dependencies ahead of time to allow for the build to run. This is not in any way specific to Nix - you would see the exact same error if you were to run `cargo build` on a system with no internet access and the package missing from local cache (i.e. not being fetched ahead of time).

To facilitate the dependency download before running the build, we merge the `Cargo.lock` contents of the root lock and `tests/components/rust/Cargo.lock` and let `cargo` download everything present in any of these two (essentially, we run `cargo vendor`) populating the cache: https://github.com/wasmCloud/wasmCloud/blob/f0d45b7a43d143e8ad00d95615eca32b94a5ac62/flake.nix#L124-L150. Note, this was set up well *before* component examples with separate workspaces were added to the repo. This is the only time anything specific to Nix happens here - we use the expected digest of the dependencies from `Cargo.lock` and give that to Nix to fetch from the network. The contents fetched by Nix are put in exact same location on disk as `cargo vendor` would do. The only reason we need to do this is to verify the *contents* fetched from network, the only way to fetch anything from the network from within the Nix sandbox is to provide content digest along the URL.

In build step, `cargo` uses the dependencies previously downloaded, just like it would do in `cargo vendor` use case. The local cache is then used as a registry. This is all standard procedure documented as part of running `cargo vendor`, not in any way specific to Nix.

In result, `cargo`, as part of building the workspace and all tests, is trying to build a few examples used in our root workspace tests, however, the examples are using versions of transitive dependencies (`syn` and `prettyplease`), which *differ* from the ones used by either `./tests/components/rust` workspace or root workspace. During the build, `cargo` is trying to fetch these dependencies, but they are, of course, missing from the local registry we have assembled earlier and there is no unrestricted network connection available from within the sandbox, therefore it fails and you see this error

vados-cosmonic: Thanks for the detailed explanation! I can understand where the issue now -- thanks for the pointer to the nix code that was doing the combination!

PR #2877, opened 2024-08-27
chore(examples): remove examples cargo lock

vados-cosmonic:
This commit removes the `Cargo.lock` for `http-blobstore` that was likely accidentally checked in, and adds a `.cargo/config.toml` so that the project can be built with `cargo build`.


rvolosatovs (changes requested):
I'm not sure that's what we want. (also note that every example has a `Cargo.lock`)

We should only remove the `Cargo.lock` if we make the example part of the workspace (which would prevent it from being built as part of the tests)

rvolosatovs (commented):
How about we make all example components part of the same workspace instead, removing `Cargo.lock` for all examples and just have one shared one in the workspace?

This way we would have 3 workspaces in the repo and hopefully speed up builds

vados-cosmonic:
@rvolosatovs we don't *check in* `Cargo.lock` for any other example, why do you want it to persist for this one?

brooksmtownsend:
@rvolosatovs The main downside for examples being a part of the workspace is when we generate them as a part of creating a new project template, then the examples don't stand by themselves. Outside of the ones that we generate as templates I'm fine with the workspace, though

rvolosatovs (approved):
yeah, you're we don't have `Cargo.lock`s checked in for other examples, I was confused.

Still probably best to make all components part of a single workspace

vados-cosmonic:
> Still probably best to make all components part of a single workspace

Agree with that! Also @brooksmtownsend 's poitn is good, I think the `Cargo.lock` would get overriden instantly in a new template project, but not having it is nice as well. 

AFAIK we shouldn't make this change for `tests/components` because that really is another workspace sort of hidden inside the top level one, but this one seemed reasonable. Thanks for looking at it closely!

rvolosatovs:
> @rvolosatovs The main downside for examples being a part of the workspace is when we generate them as a part of creating a new project template, then the examples don't stand by themselves. Outside of the ones that we generate as templates I'm fine with the workspace, though

sounds good, let's discuss and address in a separate PR/issue

PR #2880, opened 2024-08-27
feat(host): Add support for supplying a NATS credentials file

joonas:
## Feature or Problem

It's a bit clumsy to have to provide the seed and jwt separately when they could be just read from the credentials file, so let's add support for that.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::env;
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 use std::time::Duration;
 
 use anyhow::{bail, Context};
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
I was going to comment that `LazyLock` here was a little overkill, but I think it leaves a good setup for a new issue that allows for `--ctl-creds` and `--rpc-creds` arguments. @joonas can you file that issue if not addressing here?

joonas:
@brooksmtownsend I went ahead and implemented `--ctl-creds` and `--rpc-creds` as well, but I didn't create separate parsing functions for them, which does mean that the reported error messages would effectively be the same for each.

Would you prefer separate parsing functions to customize the error messages?

brooksmtownsend:
@joonas i think attaching a context to the parsing result to indicate which one failed would be plenty!

joonas:
> @joonas i think attaching a context to the parsing result to indicate which one failed would be plenty!

Sounds good, [here we go](https://github.com/wasmCloud/wasmCloud/pull/2880/files#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR382-R391)

brooksmtownsend (approved):

PR #2902, opened 2024-08-29
feat(http-server): path-based component routing

brooksmtownsend:
## Feature or Problem
This PR implements a second "mode" for the HTTP server provider that allows for path-based routing for components. This is **intentionally kept simple** as our HTTP server isn't meant to be a fully featured L7 router, but it can support a couple of different modes of routing requests to components.

This path mode is simple, instead of supplying an `address` on a link that the provider will forward requests from that address to the component, this provider will listen on a single supplied address and forward requests to components based on the `path` link configuration variable.

Most of the code changes in this PR were motivated by reducing code duplication between the address & path based modes, since the logic is almost entirely the same and just requires different methods of looking up the correct component to invoke. I expect adding a domain-based routing mechanism will be incredibly simple after this refactor.

I also removed `dashmap` because everywhere else in our codebase we use the `Arc<tokio::sync::RwLock<HashMap>>` pattern and I'd rather keep the same set of assumptions around deadlocking instead of using a different mechanism.

TODO:
- [x] Code cleanup, refactors to reduce duplication of code
- [x] Error logging & more tests to ensure conflicting paths / duplicate links are handled well
- [x] ~Update the address mode to support default values supplied as config~
- [x] Update the README to include the new options
- [x] Ensure that no default behavior changed with this addition

## Related Issues
It seems there is currently a bug with the link deletion process where the target of the link is not included with the link delete message, nor is is something we require when deleting the link via the control interface. This is a broader issue, but it manifests itself in the link delete not being called in the HTTP server.

## Release Information
http-server v0.23.0

### Acceptance or Integration
Added two integration tests that use a timeout value that confirm that messages are being properly sent out to two different components based on path, and re-affirming the existing "address" mode functionality

### Manual Verification
I was able to do some manual verification as well to ensure that the behavior with links was handled well:
```
➜ curl localhost:8080/go
Hello from Go!

➜ curl localhost:8080/rust
Hello from Rust!

➜ curl localhost:8080/api/rust
Hello from Rust!

➜ curl localhost:8080/api/go
path not found%
```

my app yaml (deployed from the Rust hello world wadm after building the HTTP server and golang hello world component locally):
```yaml
kind: Application
metadata:
  name: rust-hello-world
  annotations:
    description: 'HTTP hello world demo in Rust, using the WebAssembly Component Model and WebAssembly Interfaces Types (WIT)'
    wasmcloud.dev/authors: wasmCloud team
    wasmcloud.dev/source-url: https://github.com/wasmCloud/wasmCloud/blob/main/examples/rust/components/http-hello-world/wadm.yaml
    wasmcloud.dev/readme-md-url: https://github.com/wasmCloud/wasmCloud/blob/main/examples/rust/components/http-hello-world/README.md
    wasmcloud.dev/homepage: https://github.com/wasmCloud/wasmCloud/tree/main/examples/rust/components/http-hello-world
    wasmcloud.dev/categories: |
      http,http-server,rust,hello-world,example
spec:
  components:
    - name: http-component
      type: component
      properties:
        image: file://./build/http_hello_world_s.wasm
      traits:
        - type: spreadscaler
          properties:
            instances: 1

    - name: http-component-two
      type: component
      properties:
        image: file://../../../golang/components/http-hello-world/build/http_hello_world_s.wasm
      traits:
        - type: spreadscaler
          properties:
            instances: 1

    # Add a capability provider that enables HTTP access
    - name: httpserver
      type: capability
      properties:
        image: file://../../../../src/bin/http-server-provider/build/http-server-provider.par.gz
        config:
          - name: default-http
            properties:
              routing_mode: path
              default_address: 127.0.0.1:8080
      traits:
        # Establish a unidirectional link from this http server provider (the "source")
        # to the `http-component` component (the "target") so the component can handle incoming HTTP requests,
        #
        # The source (this provider) is configured such that the HTTP server listens on 127.0.0.1:8080
        - type: link
          properties:
            target: http-component
            namespace: wasi
            package: http
            source:
              config:
                - name: pathone
                  properties:
                    path: /rust
        - type: link
          properties:
            target: http-component
            namespace: wasi
            package: http
            name: other-link-name
            source:
              config:
                - name: pathoneagain
                  properties:
                    path: /api/rust
        - type: link
          properties:
            target: http-component-two
            namespace: wasi
            package: http
            source:
              config:
                - name: pathtwo
                  properties:
                    path: /go

```


crates/provider-http-server/README.md
```diff
@@ -12,16 +12,87 @@ wash build -p src/bin/http-server-provider
...
+
+### Path routing mode
+
+In path routing mode, the above configuration of `routing_mode` and `default_address` is supplied as provider configuration as well as all values in [HTTP address configuration](#http-address-configuration). The HTTP server, when in path routing mode, sets up a listener at startup to serve **all** components.
+
+All components must be configured with a `path` on the link config for routing in this mode.
+
+| Key    | Default | Description                                                                                                    |
+| ------ | ------- | -------------------------------------------------------------------------------------------------------------- |
+| `path` | `N/A`   | **Required.** The path, e.g. `/api/v1`, to register to send all requests at that path to the linked component. |
```
thomastaylor312, line 38: I think it would probably be good to also require a host_name value here. Otherwise a path can only be used once. We don't have to do full things like checking things like `X-Forwarded-By` or anything like that, but checking the host would make it so paths don't clobber each other (which feels likely, especially when people are testing things or kicking the tires)

crates/provider-http-server/src/path.rs
```diff
@@ -0,0 +1,244 @@
...
+
+use crate::{
+    build_request, get_cors_layer, get_tcp_listener, invoke_component, load_settings,
+    ServiceSettings,
+};
+
+/// `wrpc:http/incoming-handler` provider implementation with path-based routing
+#[derive(Clone)]
+pub struct HttpServerProvider {
+    // NOTE: When acquiring locks to both of these maps, always acquire the `paths` lock first
```
thomastaylor312, line 31: Nit: Should we wrap these two things into a struct so we can guarantee the order of locking? At a bare minimum it can just have the two hashmaps, or you could design something different to avoid cloning the data if you felt ambitious

thomastaylor312: Also, I am super curious why doing it in the other order causes a deadlock 😅

thomastaylor312 (commented):
Code looks all good, just a question around design

crates/provider-http-server/src/path.rs
brooksmtownsend: >Also, I am super curious why doing it in the other order causes a deadlock 😅

It's definitely not guaranteed, but if you have two places in the code that acquire (and try to hold) the locks in the opposite order, one tokio task may acquire the first lock at the same instant that another task acquires the second lock, and then they are both holding their locks while waiting for the other one /shrug. It's not likely, I just left this message out of prudence since we've hit that before

Wrapping in a struct seems reasonable here

crates/provider-http-server/README.md
brooksmtownsend: I saw path based routing as a very simple way to setup different paths for the scenario of a couple components handling HTTP requests, but it falls apart _pretty_ quickly as often you do want to handle multiple paths.

I was thinking it would be great to also have a `host` / `domain` based mode that allows you to route based on the host header, and not the path, since that would let you easily route all requests to a particular domain to a component. Thoughts? I think combining them might just make it overly complicated, since in the host-based mode I probably would just want all paths to go to the same component

thomastaylor312 (approved):

PR #2945, opened 2024-09-05
fix(deps): pin wrpc-transport-nats to 0.22.3

vados-cosmonic:


vados-cosmonic:
No need for this, `0.22.4` has been yanked!

PR #2950, opened 2024-09-05
refactor: Use testcontainers and localstack for blobstore-s3 provider testing

joonas:
## Feature or Problem

This change adopts localstack via testcontainers instead of Minio as the default testing backend for the `blobstore-s3` provider.

It also fixes a bug that I uncovered in how the client was being set up when no bucket region was provided.

### DIFF
--- a/crates/provider-blobstore-s3/tests/storage.rs
+++ b/crates/provider-blobstore-s3/tests/storage.rs
@@ -31,32 +31,95 @@
 use std::collections::HashMap;
 use std::env;
 
+use anyhow::{Context as _, Result};
+use testcontainers::{core::WaitFor, runners::AsyncRunner as _, ContainerAsync, Image, ImageExt};
 use wasmcloud_provider_blobstore_s3::{StorageClient, StorageConfig};
 
-/// Helper function to create a `StorageClient` with local testing overrides
-async fn test_client() -> StorageClient {
-    let conf = StorageConfig {
-        endpoint: env::var("AWS_ENDPOINT").ok(),
-        access_key_id: env::var("AWS_ACCESS_KEY_ID").ok(),
-        secret_access_key: env::var("AWS_SECRET_ACCESS_KEY").ok(),
-        aliases: HashMap::new(),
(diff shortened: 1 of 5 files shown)


vados-cosmonic:
Hey IIRC `testcontainers` split out their prebaked stuff to `testcontainers-modules` -- what do you think about using [`testcontainers_modules`'s `localstack` impl?](https://docs.rs/testcontainers-modules/0.10.0/testcontainers_modules/localstack/index.html)

joonas:
> Hey IIRC `testcontainers` split out their prebaked stuff to `testcontainers-modules` -- what do you think about using [`testcontainers_modules`'s `localstack` impl?](https://docs.rs/testcontainers-modules/0.10.0/testcontainers_modules/localstack/index.html)

I'd rather we control the version of Localstack ourselves, theirs is hardcoded to 3.0.

crates/provider-blobstore-s3/tests/storage.rs
```diff
@@ -31,32 +31,94 @@
...
+        };
+
+        StorageClient::new(conf, &HashMap::new()).await
+    }
+
+    fn env_var_or_default(key: &str, default: Option<&str>) -> Option<String> {
+        std::env::var(key)
+            .map(Some)
+            .unwrap_or_else(|_| default.map(Into::into))
+    }
```
vados-cosmonic, line 109: NIT: rather than making a function for this feels like we could get away with something like `Some(std::env::var("KEY").unwrap_or_else(|| "value"))` ?

vados-cosmonic (approved):

crates/provider-blobstore-s3/tests/storage.rs
brooksmtownsend: I think this could be accomplished with a `.ok().or(default)` if the default is an `Option<String>`

joonas: Good call, I like that better, incorporated the suggestion 👍

PR #2962, opened 2024-09-06
feat(wash-cli): convert dependencies to WADM manifests

vados-cosmonic:
## Feature or Problem

This commit implements conversion of detected dependencies during `wash dev` to WADM manifests that can be run when a component is being developed.


crates/wash-cli/src/up/mod.rs
```diff
@@ -285,6 +285,26 @@ pub struct WasmcloudOpts {
 }
 
 impl WasmcloudOpts {
+    /// Create a NATS client from the current WasmcloudOpts
+    pub(crate) async fn create_nats_client(&self) -> Result<async_nats_0_33::Client> {
```
brooksmtownsend, line 289: Should this just be `async_nats` now?

crates/wash-cli/src/dev.rs
```diff
@@ -85,33 +105,117 @@ impl Drop for HostSubprocess {
...
+///
+/// A `DependencySpec` represents a single dependency in the project, categorized into what part it is expected
+/// to play in in fulfilling WIT interfaces.
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum DependencySpec {
+    /// A dependency that receives invocations (ex. `keyvalue-nats` receiving a `wasi:keyvalue/get`)
+    Receives(DependencySpecInner),
+    /// A dependency that performs invocations (ex. `http-server` invoking a component's `wasi:http/incoming-handler` export)
+    Invokes(DependencySpecInner),
+}
```
brooksmtownsend, line 171: I don't dislike this terminology, but it is yet-another set of names for what we call export/import and target/source. What do you think about sticking with the WIT terminology?

```suggestion
enum DependencySpec {
    /// A dependency that receives invocations (ex. `keyvalue-nats` receiving a `wasi:keyvalue/get`)
    Exports(DependencySpecInner),
    /// A dependency that performs invocations (ex. `http-server` invoking a component's `wasi:http/incoming-handler` export)
    Imports(DependencySpecInner),
}
```

brooksmtownsend: I don't think we've really settled on a perfect thing to call these directions, so I figured sticking with what's already required may be the best move. Open to conversation here

crates/wash-cli/src/dev.rs
```diff
@@ -125,17 +229,80 @@ impl KnownDep {
...
+                        ns,
+                        pkg,
+                        iface,
+                        version.map(VersionCoverage::SemVer).unwrap_or_default(),
+                    ),
+                    delegated_to_workspace: false,
+                    link_name: "default".into(),
+                    image_ref: Some(DEFAULT_KEYVALUE_PROVIDER_IMAGE.into()),
+                    // TODO: needs config on the source->target link (Bucket name)
+                    ..Default::default()
```
brooksmtownsend, line 249: Note for later, I think we also need to create the bucket too since the NATS KV provider depends on it

crates/wash-cli/src/dev.rs
```diff
@@ -152,25 +319,482 @@ impl KnownDep {
...
+    }
+
+    /// Retrieve whether this dependency spec is a component
+    fn is_component(&self) -> bool {
+        self.is_component
+    }
+}
+
+// #[derive(Debug, Clone, PartialEq, Eq)]
+// enum SecretSpec {
```
brooksmtownsend, line 528: nit: can be removed if not needed but disregard if you're planning to use this in a followup

crates/wash-cli/src/dev.rs
```diff
@@ -152,25 +319,482 @@ impl KnownDep {
...
+    /// Lookup of dependencies by project key, with lookups into the pool
+    dependencies: HashMap<ProjectDependencyKey, DependencySpec>,
+
+    /// Dependencies that receive invocations for given interfaces (i.e. `keyvalue-nats` receiving a `wasi:keyvalue/get`)
+    ///
+    /// The keys to this structure are indices into the `dependencies` member.
+    receivers: HashMap<(WitNamespace, WitPackage, WitInterface, Option<Version>), Vec<usize>>,
+
+    /// Dependencies that perform invocations for given interfaces (i.e. `http-server` invoking a component's `wasi:http/incoming-handler` export)
+    invokers: HashMap<(WitNamespace, WitPackage, WitInterface, Option<Version>), Vec<usize>>,
```
brooksmtownsend, line 559: same comment re: importers/exporters, just making sure wherever we land we change this too

crates/wash-cli/src/dev.rs
```diff
@@ -152,25 +319,482 @@ impl KnownDep {
...
+                annotations: BTreeMap::from([("version".into(), "v0.0.0".into())]),
+                labels: BTreeMap::new(),
+            },
+            spec: Specification {
+                components,
+                policies: Vec::from([Policy {
+                    name: "nats-kv".into(),
+                    policy_type: "policy.secret.wasmcloud.dev/v1alpha1".into(),
+                    properties: BTreeMap::from([("backend".into(), "nats-kv".into())]),
+                }]),
```
brooksmtownsend, line 782: I noticed this in my generated manifest and was a little confused 🤔 I assume this is a placeholder for future secrets integration but if not I think we can remove

crates/wash-cli/src/dev.rs
```diff
@@ -152,25 +319,482 @@ impl KnownDep {
...
+}
+
+/// Check whether the provided interface is ignored for the purpose of building dependencies.
+///
+/// Dependencies are ignored normally if they are known-internal packages.
+fn is_ignored_iface_dep(ns: &str, pkg: &str, iface: &str) -> bool {
+    matches!(
+        (ns, pkg, iface),
+        ("wasi", "io", _) | ("wasi", "clocks", _) | ("wasi", "http", "types")
+    )
```
brooksmtownsend, line 797: I wonder if we should push something like this into `core`, because our `runtime` crate also has that big ol match statement that ignores a bunch of dependencies that wasmtime covers

brooksmtownsend (approved):
I tried this out, and the manifest looks great! I left a few comments, but given that they are primarily internal code changes, I'll drop an approve and let you work through if you think they're worth changing now.

Side note, this launched my wadm manifest + component well, I have two requests for improvements on functionality though:
1. This didn't seem to hot-reload my component, it watches changes but doesn't rebuild (I think)
2. For HTTP server and messaging, I would love to see the configured address/subscription that's setup so I can immediately open another terminal and invoke it without guessing localhost:8000

crates/wash-cli/src/up/mod.rs
vados-cosmonic: We can't change this yet because of `wadm` right -- it's because it uses the older version

crates/wash-cli/src/dev.rs
vados-cosmonic: Hmnn shouldn't that be done in the provider though? Did we remove bucket auto-creation?

crates/wash-cli/src/dev.rs
vados-cosmonic: Oh thanks for catching this, need to remove

crates/wash-cli/src/dev.rs
vados-cosmonic: Done!

PR #2963, opened 2024-09-06
fix(*): link uniqueness enforcement, deletes

brooksmtownsend:
## Feature or Problem
This PR implements rejecting links with the same source ID, wit namespace/package, and link name but with different targets. This is a scenario that components cannot support, but some providers did support this.

After this PR, if your provider supported links on the same interface to multiple targets (e.g. the HTTP server targeting many components on `wasi:http/incoming-handler`, you **must** provide a link name in order to keep these links separate. wasmCloud will reject the link if it's not supplied. In a future PR, wadm should reject this at manifest validation since it's easy to tell if there is a conflict.

Here's a diagram that illustrates the difference:
image

## Related Issues
resolves issues of link deletion found in #2902 

This PR also includes fixes for capability providers not receiving the full amount of information on a link delete in order to support link names & multiple components.

We also need an issue for supporting, or not, adding interfaces to links, because that's not currently supported today (see TODO in test)

### Acceptance or Integration
I've also included a new `links.rs` integration test to ensure this functionality does not drift in the future.

### Manual Verification
I validated this all worked well first, but then turned that into a test.

### DIFF
--- a/crates/wash-cli/src/ctl/output.rs
+++ b/crates/wash-cli/src/ctl/output.rs
@@ -73,30 +73,16 @@ pub fn links_table(mut list: Vec<InterfaceLinkDefinition>) -> String {
         table.add_row(Row::new(vec![
             TableCell::new_with_alignment(l.source_id.clone(), 1, Alignment::Left),
             TableCell::new_with_alignment(l.target.clone(), 1, Alignment::Left),
-            if l.interfaces.len() == 1 {
-                TableCell::new_with_alignment(
-                    // SAFETY: We know that there is at least one element in the interfaces vector
-                    format!(
-                        "{}:{}/{}",
-                        l.wit_namespace,
-                        l.wit_package,
-                        l.interfaces.get(0).unwrap()
-                    ),
-                    1,
-                    Alignment::Left,
-                )
-            } else {
-                TableCell::new_with_alignment(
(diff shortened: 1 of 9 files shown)


lxfontes (approved):

crates/wash-cli/src/ctl/output.rs
```diff
@@ -65,20 +65,39 @@ pub fn links_table(mut list: Vec<InterfaceLinkDefinition>) -> String {
...
+                    1,
+                    Alignment::Left,
+                )
+            } else {
+                TableCell::new_with_alignment(
+                    format!(
+                        "{}:{}/{}",
+                        l.wit_namespace,
+                        l.wit_package,
+                        l.interfaces.join(",")
```
vados-cosmonic, line 94: NIT: Am I missing something? you can use `l.interfaces.join(",")` with a single-element list and you'll get the same result as taking the item and pulling it out...?

Why the check for if there's only one?

vados-cosmonic (approved):
LGTM :rocket:

crates/wash-cli/src/ctl/output.rs
brooksmtownsend: I didn't know this! I will refactor

Issue #1868, opened 2024-04-10
[FEATURE] Implement `wasi:keyvalue/watch`


thomastaylor312:
To fully implement the wasi:keyvalue interface, we need to implement the watch operations as well. See https://github.com/WebAssembly/wasi-keyvalue/blob/main/wit/watch.wit for the interface. We haven't supported key watches before so this will be a new handler type components can import

### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

ossfellow:
Once _wrpc-keyvalue_ adds this interface to the `imports world`, I can and will add it to _nats-keyvalue_ provider.

brooksmtownsend:
This is a great first issue for folks looking to learn more about how the connection between Wasm component interfaces and wRPC works in wasmCloud. Take a look at the implementation of wasi:keyvalue/store and wasi:keyvalue/atomics for inspiration here.

Aditya1404Sal:
@brooksmtownsend I'd like to work on this issue !
are there any specifics of wasi:keyvalue/store and wasi:keyvalue/atomics that I should checkout to get a better understanding of what I'm dealing with ?

thomastaylor312:
I would recommend looking at how that code is implemented and exposed to components running in wasmCloud. Should be a very similar pattern to implement `watch`

PR #2977, opened 2024-09-10
WIP: feat(providers): Implement `wasi:keyvalue/watch` handler and provider

Aditya1404Sal:
# WIP
## Related Issues
#1868


Aditya1404Sal:
@thomastaylor312 -- `wasi:keyvalue/watch` should have a Provider -> Component connection that will be invoked when the specific link of `wasi:keyvalue/watcher` interface is made, for this I am trying to implement `receive_link_config_as_source` for each individual provider (redis,natskv,vault), which would setup a watcher process that will wait for any changes that happens in a bucket (set and delete operations) which would trigger the `on_set` and `on_delete` calls to notify the respective component. Am I going in the right direction ?

PR #2980, opened 2024-09-10
feat(host): emit metrics for HTTP fuel consumption

brooksmtownsend:
Draft PR to spur some discussion. I would love to get thoughts on if we should turn this on if it's supposed to be expensive computationally to measure fuel. I like emitting this to metrics for now so that users could feasibly measure how much work a component is doing, and if that's useful we can also use this metric as a limiter for the store instead of using a `u64::MAX`

## Feature or Problem

This PR enables another OTEL metric for component invocations, fuel consumed, which is a monotonically increasing number per-store that counts the number of WebAssembly instructions that a component executed. https://docs.rs/wasmtime/24.0.0/wasmtime/struct.Config.html#method.consume_fuel. The goal of this PR, or a PR like this, would be to allow instrumenting components to understand how many instructions they are calling during their execution. This consumption is deterministic, and the same inputs to an invocation should result in the same usage of fuel.

This PR is a draft because
- I didn't implement fuel counting for messaging or dynamic invocations yet
- To implement this for HTTP, I had to await the task handle which likely has unintended consequences in error scenarios
- Fuel counting is, according to the documentation, computationally expensive. I didn't notice a difference in our HTTP hello world benchmark, but I don't want to enable this for no reason.

### Manual Verification
Here's a fun comparison of the HTTP hello world Rust example and the HTTP client tinygo example, using `180` and `136183` fuel units respectively. This isn't a fair comparison because the tinygo example does more (an outgoing HTTP request).

image

### DIFF
--- a/crates/host/src/metrics.rs
+++ b/crates/host/src/metrics.rs
@@ -1,4 +1,4 @@
-use wasmcloud_tracing::{Counter, Histogram, KeyValue, Meter, Unit};
+use wasmcloud_tracing::{Counter, Gauge, Histogram, KeyValue, Meter, Unit};
 
 /// `HostMetrics` encapsulates the set of metrics emitted by the wasmcloud host
 #[derive(Clone, Debug)]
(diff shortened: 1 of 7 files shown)


brooksmtownsend:
This was a neat experiment, but I'm not 100% sure we want this, so I'll close it for now

PR #2982, opened 2024-09-10
feat(bus)!: add checked set_link_name

brooksmtownsend:
## Feature or Problem
This PR adds a new function to the `wasmcloud:bus/lattice` interface, `set_link_name_checked`, which will return an error if you attempt to override the default link name for a specific interface but a link doesn't exist.

This will be useful at runtime for a component to be able to attempt to invoke a different linked target, but then instead fall back to invoking a default target if there was an error during the `set_link_name_checked` process.

Open to suggestions to other functions, I considered other functions like `get_link` or `check_link` for checking first, but figured this would just be a better way to try and set the link immediately.

There's also room for other functions like `reset_link_names` etc etc, but leaving that out for now.

### Acceptance or Integration
Modified existing interfaces test to add two cases where the checked interface set should fail.


crates/host/src/wasmbus/handler.rs
```diff
@@ -118,6 +120,38 @@ impl Bus for Handler {
...
+                .is_none()
+            {
+                Some(instance)
+            } else {
+                None
+            }
+        }) {
+            return Ok(Err(format!(
+                "interface `{}` does not have an existing link with name `{}`",
+                interface_missing_link, link_name
```
vados-cosmonic, line 147:
```suggestion
                "interface `{interface_missing_link}` does not have an existing link with name `{link_name}`"
```

tests/components/rust/interfaces-reactor/src/lib.rs
```diff
@@ -159,16 +159,37 @@ pub fn run_test(body: &[u8]) -> (Vec<u8>, String) {
...
             "wasi",
             "blobstore",
             "blobstore",
         )],
-    );
+    )
+    .is_ok());
     blobstore::run_test(0, &body, "container");
 
+    eprintln!("test link name check...");
```
vados-cosmonic, line 173:
```suggestion
```

tests/components/rust/interfaces-reactor/src/lib.rs
```diff
@@ -159,16 +159,37 @@ pub fn run_test(body: &[u8]) -> (Vec<u8>, String) {
...
+    // Interface that's not linked
```
vados-cosmonic, line 174: This comment could be an extra arg to `assert!`

vados-cosmonic (approved):
LGTM :rocket: Nothing but nits

tests/components/rust/interfaces-reactor/src/lib.rs
vados-cosmonic: Oh didn't realize there were printlns all over this file

rvolosatovs (changes requested):
Requesting changes to prevent merge, since this is a breaking change. We need to bump the interface (patch) version and use `@since` guard.

crates/runtime/wit/deps/wasmcloud/bus.wit
```diff
@@ -6,9 +6,13 @@ interface lattice {
...
-    /// wasmcloud:bus package. This is used when you are linking multiple of the same interfaces
+    /// Set a link name to use for all interfaces specified. This is advanced functionality only
+    /// available within wasmcloud and, as such, is exposed here as part of the wasmcloud:bus package.
+    /// This is used when you are linking multiple of the same interfaces
     /// (i.e. a keyvalue implementation for caching and another one for secrets) to a component
     set-link-name: func(name: string, interfaces: list<call-target-interface>);
+
+    /// Set a link name to use for all interfaces specified. This is the same as `set-link-name`, but will
+    /// return an error if a link does not exist at runtime with the specified name for the specified interfaces.
+    set-link-name-checked: func(name: string, interfaces: list<call-target-interface>) -> result<_, string>;
```
ricochet, line 17: This wit package has not been published (#3101). If we're already OK with marking as a breaking change, then I would prefer to support only one interface function as `set-link-name` that returns a result. Caller's should always be expected to handle this case.

The component crate is v0 (minor should be bumped).

rvolosatovs: The WIT interface version is specified as `1.0.0` and while I realize that an actual tarball containing it was never published, it seems probable that users of this interface exist outside this repository.
`1.0.1` with `set-link-name-checked` gated by `@since` would not break such, i.e. all components targeting `wasmcloud:bus@1.0.0` will keep working

`2.0.0` with `set-link-name` returning a result is another option, which would not break anyone (and we could provide both versions of the interface)

brooksmtownsend:
After syncing, decided to go with a 2.0.0 with updating the function signature.

crates/component/src/lib.rs
```diff
@@ -61,19 +62,21 @@ mod test {
             let _ = wasi::config::runtime::get("foo");
             let _ = wasi::config::runtime::get_all();
 
-            wasmcloud::bus::lattice::set_link_name(
+            wasmcloud::bus1_0_0::lattice::set_link_name(
```
rvolosatovs, line 65: since this crate is pre-1.0 and this change would be (rust) API breaking anyway, let's only use 2.0.0 version in `wasmcloud_component`?

crates/runtime/src/component/bus.rs
```diff
@@ -10,14 +10,15 @@ use wasmcloud_core::CallTargetInterface;
...
-    /// Set link name
+    /// Set the link name to use for a given list of interfaces, returning an error
+    /// if a link doesn't exist on the given interfaces for the given target
     async fn set_link_name(
         &self,
-        target: String,
+        link_name: String,
         interfaces: Vec<Arc<CallTargetInterface>>,
-    ) -> anyhow::Result<()>;
+    ) -> anyhow::Result<std::result::Result<(), String>>;
```
rvolosatovs, line 21:
```suggestion
    ) -> anyhow::Result<Result<(), String>>;
```

crates/runtime/src/component/bus.rs
```diff
@@ -26,7 +27,7 @@ impl<H: Handler> lattice::Host for Ctx<H> {
         &mut self,
         link_name: String,
         interfaces: Vec<Resource<Arc<CallTargetInterface>>>,
-    ) -> anyhow::Result<()> {
+    ) -> anyhow::Result<std::result::Result<(), String>> {
```
rvolosatovs, line 30:
```suggestion
    ) -> anyhow::Result<Result<(), String>> {
```

crates/runtime/src/component/bus1_0_0.rs
```diff
@@ -0,0 +1,60 @@
...
+    ) -> anyhow::Result<()> {
+        let interfaces = interfaces
+            .into_iter()
+            .map(|interface| self.table.get(&interface).cloned())
+            .collect::<TableResult<_>>()
+            .context("failed to convert call target interfaces")?;
+        self.handler
+            .set_link_name(link_name, interfaces)
+            .await
+            .map(|_| ())
```
rvolosatovs, line 35: nit; I usually find that splitting such occurrences into `?` followed by `Ok(())` is:
1. clearer/easier to understand
2. ends up with less LoC due to `rustfmt` formatting
3. more error-prone in case the inner `Result::Ok` is not what one expected

```suggestion
        self.handler
            .set_link_name(link_name, interfaces)
            .await?;
        Ok(())
```

I'm fine either way though

rvolosatovs (approved):

crates/runtime/src/component/bus1_0_0.rs
brooksmtownsend: I could go either way here, changed to this just to simplify

rvolosatovs (approved):

tests/components/rust/interfaces-reactor/src/lib.rs
```diff
@@ -144,12 +144,13 @@ pub fn run_test(body: &[u8]) -> (Vec<u8>, String) {
     keyvalue::run_store_test(&body);
 
     eprintln!("test vault keyvalue/store...");
-    bus::lattice::set_link_name(
+    assert!(wasmcloud_component::wasmcloud::bus::lattice::set_link_name(
```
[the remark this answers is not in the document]
brooksmtownsend: not gonna spin CI again for this one, but I'll sneak it in a PR later today

ricochet (approved):

PR #2983, opened 2024-09-11
doc: fix provider-messaging-nats

peter-jerry-ye:
## Feature or Problem

The document of `provider-messaging-nats` is not correct

## Related Issues

The configurations should be `subscriptions` and `cluster_uris`, but they are written as `subscription` and `uri`.

Reference: https://github.com/wasmCloud/wasmCloud/blob/d21d2a9e7dffd16315eeb565e2cd0e1f1aeeac6c/crates/provider-messaging-nats/src/connection.rs#L10

## Release Information

next

## Consumer Impact

N/A

## Testing

N/A

### Unit Test(s)

N/A

### Acceptance or Integration

N/A

### Manual Verification

I ran with the previous configuration but it didn't work. Then I changed to the current configuration and confirm that it works.

### DIFF
--- a/crates/provider-messaging-nats/README.md
+++ b/crates/provider-messaging-nats/README.md
@@ -6,7 +6,7 @@ To configure this provider, use the following link settings in link definitions:
 
 | Property | Description |
 | :--- | :--- | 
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM :rocket:  

Thanks for fixing this @peter-jerry-ye , this documentation was definitely out of date!

vados-cosmonic:
Would you mind performing the DCO? 

You should be able to run a command like this:

```
git commit --amend -s -m "doc(providers): fix messaging-nats configuration docs
```

peter-jerry-ye:
Sure. DCO signed.

PR #2986, opened 2024-09-11
feat(wash-cli): write out generated wadm

vados-cosmonic:

### DIFF
--- a/crates/wash-cli/src/dev.rs
+++ b/crates/wash-cli/src/dev.rs
@@ -6,7 +6,7 @@ use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
-use anyhow::{bail, Context as _, Result};
+use anyhow::{bail, ensure, Context as _, Result};
 use chrono::{DateTime, Utc};
 use clap::Parser;
 use console::style;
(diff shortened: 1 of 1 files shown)


crates/wash-cli/src/dev.rs
```diff
@@ -95,6 +95,13 @@ pub struct DevCommand {
         help = "Run the wasmCloud host in a subprocess (rather than detached mode)"
     )]
     pub use_host_subprocess: bool,
+
+    /// Write generated WADM manifest(s) to a given folder (every time they are generated)
+    #[clap(
+        long = "generated-manifest-output-dir",
+        env = "WASH_DEV_GENERATED_WADM_OUTPUT_DIR"
+    )]
+    pub generated_manifest_output_dir: Option<PathBuf>,
```
brooksmtownsend, line 104: This is certainly a keyboardful to write out, what do you think about having this be just a little bit shorter?
```suggestion
    /// Write generated WADM manifest(s) to a given folder (every time they are generated)
    #[clap(
        long = "manifest-dir",
        env = "WASH_DEV_MANIFEST_DIR"
    )]
    pub generated_manifest_output_dir: Option<PathBuf>,
```

something something

vados-cosmonic: Hmnnn I could do `manifest-output-dir`? or how about `save-generated-manifests` ?

Feels like `manifest-dir` doesn't actually say enough about what the intent is

brooksmtownsend: I'm good with manifest-output-dir 😄 

Also, do you think that we're going to have multiple manifests? Or do you think that we should be able to change this from a dir to a file

vados-cosmonic: Yeah -- so that's on purpose, in future workspace (and *nested* workspace scenarios) we may very well have different manifests, so that's on purpose.

vados-cosmonic: That reminds me though -- need to do a check and make *sure* it's a directory, so will add that in!

brooksmtownsend (approved):

vados-cosmonic:
`wash dev` E2E test is failing in CI -- looking into it!

PR #2993, opened 2024-09-12
feat(provider-blobstore-azure): Add integration tests

joonas:
## Feature or Problem

This adds an integration test suite invoked via wRPC calls against the exposed interface.

It also fixes a couple of bugs in the implementation that I came across while developing the tests.

In case you're wondering, the tests are currently marked as `#[ignore]`, because you can't run multiple instances of a given provider in the same process as they'll attempt to stomp on each other's connections (and fail to do so). There's a bit of future work to be done to make this not an issue, which would allow all of these tests to be enabled.

### DIFF
--- a/crates/provider-blobstore-azure/src/lib.rs
+++ b/crates/provider-blobstore-azure/src/lib.rs
@@ -7,10 +7,11 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use anyhow::{bail, Context as _, Result};
+use azure_storage::CloudLocation;
 use azure_storage_blobs::prelude::*;
 use bindings::wrpc::blobstore::types::{ContainerMetadata, ObjectId, ObjectMetadata};
 use bytes::{Bytes, BytesMut};
-use futures::{Stream, StreamExt as _, TryStreamExt as _};
+use futures::{Stream, StreamExt as _};
 use tokio::sync::{mpsc, RwLock};
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::{error, instrument};
@@ -23,8 +24,9 @@ use config::StorageConfig;
 
 mod config;
 
-mod bindings {
+pub mod bindings {
     wit_bindgen_wrpc::generate!({
+        world: "interfaces",
         with: {
             "wasi:blobstore/types@0.2.0-draft": generate,
             "wasi:io/error@0.2.0": generate,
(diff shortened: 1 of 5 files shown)


joonas:
@rvolosatovs if you have any feedback on things you would do differently, I'm all ears, I tried my best to incorporate your feedback and guidance in the tests 🙂

crates/provider-blobstore-azure/src/lib.rs
```diff
@@ -23,8 +24,9 @@ use config::StorageConfig;
 
 mod config;
 
-mod bindings {
+pub mod bindings {
```
joonas, line 27: Note, these need to be made public so that they can be used by the integration tests to [mimic starting the provider](https://github.com/wasmCloud/wasmCloud/pull/2993/files#diff-7f900ca14ffd60e466d344ec5b793e01facbf22068206803b9e0e01c1e784beaR204-R218)

crates/provider-blobstore-azure/wit/interfaces.wit
```diff
@@ -3,3 +3,7 @@ package wasmcloud:provider-blobstore-azure;
 world interfaces {
     export wrpc:blobstore/blobstore@0.2.0;
 }
+
+world testing-client {
+    import wrpc:blobstore/blobstore@0.2.0;
+}
```
joonas, line 9: I had hoped to be able to generate these via `inline:` definition in the testing code itself, but unfortunately that didn't seem to work as expected, so I am adding them here for the tests to be able to consume.

brooksmtownsend (approved):
Huge amount of functionality tested here with the unit test, awesome stuff 🚀 

Feels like we'll be able to take this for other blobstore providers pretty well

crates/provider-blobstore-azure/wit/interfaces.wit
rvolosatovs: It's surprising that it did not work, however why don't we use `wrpc_interface_blobstore` crate for this purpose? It should have import bindings for all blobstore interfaces already and would reduce build times (as well allow us to build functionality on top of these types across crates)

joonas: To be honest, I wasn't using it, because I had no idea it existed 😅 

I would happy to switch over to that crate in this PR if that's the route we should be going anyway, wdyt?

PR #2997, opened 2024-09-13
feat(wash-cli): add host log path option

vados-cosmonic:
This commit adds a `--host-log-path`` option to the wasmcloud-related options for `wash up`. With this command, a custom path can be specified to write to when the host is producing logs.


brooksmtownsend (approved):
Nice!

As a followup enhancement, should we write to this file even if running in the mode where stderr is forwarded to the terminal? (e.g. not detatched)

vados-cosmonic:
Hey @brooksmtownsend > Yeah that is a good idea I think!

Made a follow-up issue here:
https://github.com/wasmCloud/wasmCloud/issues/3012

PR #3000, opened 2024-09-13
chore(wash-cli): note wash-cli move to wash

brooksmtownsend:
## Feature or Problem
Chore to note that the `wash-cli` crate will move to `wash` on the next release. What we can do is publish the next version of `wash-cli` with the deprecation notice, then publish to `wash` on the next release, and finally update the `wash-cli` README once and for all to point towards the new `wash` crate.


vados-cosmonic (approved):
LGTM 🚀

PR #3002, opened 2024-09-13
feat(provider-sqldb-postgres): Enable TLS by default, remove the need for feature flag

joonas:
## Feature or Problem

This removes the `rustls` flag and enables TLS support by deffault.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -303,7 +303,11 @@ thiserror = { version = "1", default-features = false }
 time = { version = "0.3", default-features = false }
 tokio = { version = "1", default-features = false }
 tokio-postgres = { version = "0.7", default-features = false }
-tokio-postgres-rustls = { version = "0.12", default-features = false }
+# We are using our own fork of `tokio-postgres-rustls` with the changes from
+# https://github.com/jbg/tokio-postgres-rustls/pull/24 applied on top of it,
+# so that we can adhere to the CNCF licensing requirements.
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -303,7 +303,11 @@ thiserror = { version = "1", default-features = false }
 time = { version = "0.3", default-features = false }
 tokio = { version = "1", default-features = false }
 tokio-postgres = { version = "0.7", default-features = false }
-tokio-postgres-rustls = { version = "0.12", default-features = false }
+# We are using our own fork of `tokio-postgres-rustls` with the changes from
+# https://github.com/jbg/tokio-postgres-rustls/pull/24 applied on top of it,
+# so that we can adhere to the CNCF licensing requirements.
+# tokio-postgres-rustls = { version = "0.12", default-features = false }
+tokio-postgres-rustls = { git = "https://github.com/wasmCloud/tokio-postgres-rustls.git", branch = "master", default-features = false }
```
rvolosatovs, line 310: Just a note, that this (git dep) would prevent us from publishing crates using this dependency, but I guess it's only the provider, which we do not publish and so we're fine, right?

rvolosatovs (approved):

Cargo.toml
joonas: That is correct

brooksmtownsend (approved):

PR #3019, opened 2024-09-16
feat(host): Adds support for batch support to the host

thomastaylor312:
This enables keyvalue batch support inside of the host, along with a test to make sure it works. Not all of our providers implement batch yet, so this uses the Redis provider, which did have implementions. I did have to fix the redis provider to get the right type of data back and transform it. I also had to update our wRPC versions so we could pick up on some bug fixes for the types we are encoding in the batch interface.

### DIFF
--- a/crates/provider-keyvalue-redis/src/lib.rs
+++ b/crates/provider-keyvalue-redis/src/lib.rs
@@ -276,7 +276,20 @@ impl keyvalue::batch::Handler<Option<Context>> for KvRedisProvider {
         keys: Vec<String>,
     ) -> anyhow::Result<Result<Vec<Option<(String, Bytes)>>>> {
         check_bucket_name(&bucket);
-        Ok(self.exec_cmd(ctx, &mut Cmd::mget(&keys)).await)
+        let data = match self
+            .exec_cmd::<Vec<Option<Bytes>>>(ctx, &mut Cmd::mget(&keys))
+            .await
+        {
+            Ok(v) => v
+                .into_iter()
+                .zip(keys.into_iter())
+                .map(|(val, key)| val.map(|b| (key, b)))
+                .collect::<Vec<_>>(),
+            Err(err) => {
+                return Ok(Err(err));
+            }
+        };
+        Ok(Ok(data))
     }
 
     async fn set_many(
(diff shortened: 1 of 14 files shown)


src/bin/keyvalue-redis-provider/wasmcloud.toml
```diff
@@ -1,7 +1,7 @@
 name = "KV Redis"
 language = "rust"
 type = "provider"
-version = "0.27.0"
+version = "0.28.0"
```
thomastaylor312, line 4: The cargo.toml wasn't bumped because we've never released a 0.28

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -276,7 +276,20 @@ impl keyvalue::batch::Handler<Option<Context>> for KvRedisProvider {
         keys: Vec<String>,
     ) -> anyhow::Result<Result<Vec<Option<(String, Bytes)>>>> {
         check_bucket_name(&bucket);
-        Ok(self.exec_cmd(ctx, &mut Cmd::mget(&keys)).await)
+        let data = match self
```
rvolosatovs, line 279: why the change here?

crates/runtime/src/component/keyvalue.rs
```diff
@@ -66,6 +66,88 @@ where
...
+            &self.handler,
+            Some(ReplacedInstanceTarget::KeyvalueBatch),
+            bucket,
+            &keys,
+        )
+        .await?
+        {
+            Ok(res) => Ok(Ok(res
+                .into_iter()
+                .map(|opt| opt.map(|(k, v)| (k.to_string(), Vec::from(v))))
```
rvolosatovs, line 95:
```suggestion
                .map(|opt| opt.map(|(k, v)| (k, Vec::from(v))))
```

isn't `k` a `String` already?

rvolosatovs (approved):

crates/provider-keyvalue-redis/src/lib.rs
thomastaylor312: Because you have to massage the return data and map it to the Vec of tuples. So this fetches the data and then zips it up with the keys that were passed in

brooksmtownsend (approved):

Issue #2773, opened 2024-08-14
[BUG] Blobstore FS cannot make parent directories from container name.


cdmurph32:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

Blobstore FS provider assumes that all parent directories exist when when writing an object with directories, (i.e "/"). If those directories do not exist the call will fail.

```
wasmcloud_host::wasmbus::handler: Failed to write data to blobstore: function failed: failed to write file: No
such file or directory (os error 2) component_id="poprox-poprox_s3" level=Level::Error context="put_object"
```

### Steps to reproduce

1. `container.write_data(object_name, &result_value)`, 
where 
`container`: `wasi::blobstore::container::Container`
`result_value`: `wasi::blobstore::types::OutgoingValue`
`object_name`: "string/with/slashes"

### Expected behavior

Blobstore FS provider should create the directories as necessary to write the file. This behavior would be consistent with the Blobstore S3 provider.

### Environment

- OS: MacOS
- Shell zsh
- wasmcloud Version 1.0.4
- Blobstore FS 0.8.0

cdmurph32:
Looking into this now.

PR #3020, opened 2024-09-16
Create directory before writing object.

cdmurph32:
Fixes https://github.com/wasmCloud/wasmCloud/issues/2773

## Feature or Problem
Blobstore FS provider assumes that all parent directories exist when when writing an object with directories, (i.e "/"). If those directories do not exist the call will fail.

## Related Issues
Fixes https://github.com/wasmCloud/wasmCloud/issues/2773

## Release Information
next

## Consumer Impact
Only affect blobstore-fs

## Testing
Unit test added.

### Unit Test(s)

Test added.

### Acceptance or Integration
None

### Manual Verification
Ran poprox demo with blobstore-fs and object name containing a '/'

### DIFF
--- a/crates/provider-blobstore-fs/src/lib.rs
+++ b/crates/provider-blobstore-fs/src/lib.rs
@@ -552,6 +552,12 @@ impl Handler<Option<Context>> for FsProvider {
         Ok(async {
             propagate_trace_for_ctx!(cx);
             let path = self.get_object(cx, id).await?;
+            if let Some(parent) = path.parent() {
+                info!(parent = ?parent.display(), "creating directory");
+                fs::create_dir_all(parent)
+                    .await
+                    .context("failed to create parent directories")?;
+            }
             let mut file = File::options()
                 .create(true)
                 .truncate(true)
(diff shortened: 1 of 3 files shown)


crates/provider-blobstore-fs/src/lib.rs
```diff
@@ -563,6 +563,12 @@ impl bindings::exports::wrpc::blobstore::blobstore::Handler<Option<Context>> for
         Ok(async {
             propagate_trace_for_ctx!(cx);
             let path = self.get_object(cx, id).await?;
+            if let Some(parent) = path.parent() {
+                info!(parent = ?parent.display(), "creating directory");
```
brooksmtownsend, line 567:
```suggestion
                info!(parent = ?parent.display(), "parent directory did not exist, attempting to create");
```

nit: wording that might help point to what is going on. Thoughts? I was going to suggest debug level, but since this should only happen once I'm, good with info

brooksmtownsend (approved):
Nice fix, and thank you for bringing a test along with it! I suggested some wording, but that's optional and I'm fine with just leaving it as-is

PR #3031, opened 2024-09-17
fix: python instantiation

brooksmtownsend:
## Feature or Problem
This PR fixes an issue where we were missing a few random/static static instances that componentize-py depends on.

Additionally, it reorders the implementation of the Python example to properly write the body.


vados-cosmonic (approved):
LGTM :rocket:

PR #3032, opened 2024-09-17
Lattice client refactor

lachieh:
## Feature or Problem
Refactor lattice client to move some of the retry logic away from the connection and into the client. Also requires that a specific connection class be provided through a function that receives the client config.

Additionally, some of the build output has been reconfigured to provide better type definitions and source mapping, as well as additional fields in package.json to work with commonjs modules.

## Release Information
`@wasmcloud/lattice-client-core@0.4.0`

## Consumer Impact
This is a breaking change, but since we are pre-1.0.0 and npm follows semver, this will not get accidentally bumped for anyone using the previous version.


typescript/packages/lattice-client-core/src/lattice-client.ts
```diff
@@ -16,10 +15,12 @@ type LatticeClientConfig = {
   wadmTopicPrefix?: string;
 };
 
+type LatticeConnectionFunction = (config: LatticeClientOptions['config']) => LatticeConnection;
```
vados-cosmonic, line 18:
```suggestion
type LatticeConnectionCtor = (config: LatticeClientOptions['config']) => LatticeConnection;
```

This is essentially a constructor right? I'm not sure I *love* "Ctor", but there might be a more intuitive name for this

typescript/packages/lattice-client-core/src/lattice-client.ts
```diff
@@ -128,19 +130,21 @@ export class LatticeClient {
...
+    this.#getNewConnection = getNewConnection;
+    this.#connection = getNewConnection(config);
 
     if (autoConnect !== false) {
       // try and connect, but don't throw an error if it fails. The connection will be in an error state accessible
       // through the `client.connection.status` property
-      this.#connect().catch(() => {
-        console.info('Failed to connect to lattice on creation');
+      this.#connect().catch((error) => {
+        console.info('Failed to connect to lattice on creation:', error.message ?? 'unknown error');
```
vados-cosmonic, line 147:
```suggestion
        console.warn('Failed to connect to lattice on creation:', error.message ?? 'unknown error');
```

typescript/packages/lattice-client-core/src/lattice-client.ts
```diff
@@ -199,9 +195,36 @@ export class LatticeClient {
   /**
    * Connect to the lattice
    */
-  readonly #connect = async (): Promise<void> => {
-    await this.#connection?.connect();
-  };
+  async #connect(count = 0): Promise<void> {
```
vados-cosmonic, line 198: we can use `await` like normal here right, since it's `async` -- was that version actually less readable than this one?

Also, it *might* be overkill, but I would probably want that `retryCount` (currently named `count`) to be in some sort of `ConnectArgs`/`ConnectOptions` object -- there's no way it will only have a retry count for long!

typescript/packages/lattice-client-react/src/context/lattice-client-provider.tsx
```diff
@@ -14,10 +14,12 @@ export function LatticeClientProvider({
   React.useEffect(() => {
     if (!client) return;
 
-    void client.instance.connect();
+    if (client.connection.status !== 'initial') return;
```
vados-cosmonic, line 17: Maybe this status could be named better, or a little explanation of *why* we don't want to attempt when it's in the `initial` state might be nice. 

So if the states are:

```
 | 'initial'
  | 'connected'
  | 'pending'
  | 'error'
  | 'disconnected';
```
What about the attempting (re)connections if we're in the `disconnected` or `error`? Should this effect be first connection only?

typescript/packages/lattice-client-react/vite.config.ts
```diff
@@ -0,0 +1,30 @@
+import path from 'node:path';
+import react from '@vitejs/plugin-react';
```
vados-cosmonic, line 2:
```suggestion
import path from 'node:path';

import react from '@vitejs/plugin-react';
```

Absolutely an incredibly small NIT but I like a little space between the platform imports, code-we-didnt-write, and code-we-did-write. Take it or leave it!

vados-cosmonic (approved):
LGTM :rocket: 

The notes are basically all take it or leave it, I was wondering if this is the first time we actually got `vite` involved in some of these projects (before was only UI?), this is awesome. Vite >>>

typescript/packages/lattice-client-react/src/context/lattice-client-provider.tsx
lachieh: This is the other way around, if it's _not_ in the initial state, don't try to connect. I've added a comment though 👍🏻

typescript/packages/lattice-client-core/src/lattice-client.ts
lachieh: Yes, we sure can—thanks! Way easier to read now, too.

vados-cosmonic (approved):
LGTM :rocket:

PR #3035, opened 2024-09-17
fix(host): Sets a higher value for the incoming events channel

thomastaylor312:
If you were running a high number of concurrent component invocations, it would result in a warning (and possible hang/dropped message) due to a full channel. This change attempts to set the channel size to the `max_instances` value with a minimum and a maximum possible value (i.e we don't want something with 20k instances to have a channel that large).


thomastaylor312:
cc @rvolosatovs

rvolosatovs (approved):

PR #3037, opened 2024-09-17
fix(postgres-provider): include webpki-roots in the postgres provider by default

protochron:
This adds the `webpki-roots` crate as a dependency to the postgres provider and uses it by default to provide CA certs when TLS is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7759,7 +7759,7 @@ dependencies = [
 
 [[package]]
 name = "wasmcloud-provider-sqldb-postgres"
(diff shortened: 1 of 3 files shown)


joonas (approved):

joonas:
@protochron can you bump the `Cargo.toml` version for the provider so we can get a new build going following merging this

protochron:
@joonas good call. Updated!

vados-cosmonic (approved):
LGTM :rocket:

PR #3045, opened 2024-09-18
feat(wash-cli): enable wash dev for providers

vados-cosmonic:
## Feature or Problem

This PR contains a breaking change for `wash-lib` (basically a breaking refactor) and code that enables `wash dev` to work with providers.

### DIFF
--- a/crates/wash-cli/src/dev.rs
+++ b/crates/wash-cli/src/dev.rs
@@ -19,6 +19,7 @@ use tokio::io::AsyncBufReadExt as _;
 use tokio::process::Child;
 use tokio::{select, sync::mpsc};
 use wash_lib::app::AppManifest;
+use wash_lib::cli::stop::stop_provider;
 use wash_lib::common::CommandGroupUsage;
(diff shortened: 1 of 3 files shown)


crates/wash-cli/src/dev.rs
```diff
@@ -434,7 +437,7 @@ impl DependencySpec {
 
     /// Convert to a component that can be used in a [`Manifest`] with a given suffix for uniqueness
     fn generate_component(&self, suffix: &str) -> Result<Component> {
-        let name = format!("{}-{}", suffix, self.name());
+        let name = format!("{}-dep-{}", suffix, self.name());
```
brooksmtownsend, line 440: dep or dev?

vados-cosmonic: `dep` as in dependency!

This comes up when your component is actually named `messaging-nats` :)

vados-cosmonic: we've essentially moved the problem (now if your component was named `dep-messaging-nats` it might fail), but... I think that case is practically pretty rare

PR #3046, opened 2024-09-18
chore: Replace dirs dependency with home

joonas:
## Feature or Problem

Since we're depending on [`dirs`](https://crates.io/crates/dirs) for very minimal set of functionality, and there are some [potential complications with the dependencies](https://github.com/dirs-dev/dirs-sys-rs/issues/21#issuecomment-1531528155), it seems to me that the prudent thing for us to do would be to make use of the ubiquitous [`home`](https://crates.io/crates/home) crate instead.

This replaces all of our use of `dirs` (minus the indirect dependency from `testcontainers`, that I'm hoping to address in/via https://github.com/testcontainers/testcontainers-rs/pull/736).


vados-cosmonic (approved):
LGTM :rocket:

PR #3061, opened 2024-09-19
feat(wash-cli): enable undeploying and deleting multiple apps

vados-cosmonic:
This commit enables wash to:

- Undeploy all applications at the same time
- Delete all undeployed applications

The combination of `wash app undeploy --all` and `wash app delete --all-undeployed` are meant to be used to gether in order to clear all apps in a given lattice.

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -67,11 +67,15 @@ pub struct ListCommand {
 #[derive(Args, Debug, Clone)]
 pub struct UndeployCommand {
     /// Name of the application to undeploy
-    #[clap(name = "name")]
-    app_name: String,
+    #[clap(name = "name", required_unless_present("all"))]
+    app_name: Option<String>,
 
     #[clap(flatten)]
     opts: CliConnectionOpts,
+
+    /// Whether to undeploy all the available apps
+    #[clap(long = "all", default_value = "false")]
+    all: bool,
 }
 
 #[derive(Args, Debug, Clone)]
@@ -95,8 +99,8 @@ pub struct DeployCommand {
 #[derive(Args, Debug, Clone)]
 pub struct DeleteCommand {
     /// Name of the application to delete, or a path to a Wadm Application Manifest
-    #[clap(name = "name")]
-    app_name: String,
+    #[clap(name = "name", required_unless_present("all_undeployed"))]
+    app_name: Option<String>,
 
     /// Version of the application to delete. If not supplied, all versions are deleted
     #[clap(name = "version")]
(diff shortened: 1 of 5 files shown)


crates/wash-cli/src/app/mod.rs
```diff
@@ -67,10 +67,14 @@ pub struct ListCommand {
 pub struct UndeployCommand {
     /// Name of the application to undeploy
     #[clap(name = "name")]
-    app_name: String,
+    app_name: Option<String>,
```
brooksmtownsend, line 70: Would be good to include a clap https://docs.rs/clap/latest/clap/struct.Arg.html#method.required_if_eq to make sure it's supplied if `--all` isn't

crates/wash-cli/src/app/mod.rs
```diff
@@ -95,14 +99,18 @@ pub struct DeployCommand {
 pub struct DeleteCommand {
     /// Name of the application to delete, or a path to a Wadm Application Manifest
     #[clap(name = "name")]
-    app_name: String,
+    app_name: Option<String>,
```
brooksmtownsend, line 102: same comment re: clap parsing

crates/wash-cli/src/app/mod.rs
```diff
@@ -412,37 +451,94 @@ async fn delete_application_version(cmd: DeleteCommand) -> Result<CommandOutput>
...
-    map.insert("deleted".to_string(), json!(deleted));
-    let message = if deleted {
-        format!("Deleted application: {model_name}")
-    } else {
-        format!("Already deleted application: {model_name}")
+    #[derive(serde::Serialize)]
+    struct ModelNameAndVersion<'a> {
+        model_name: &'a String,
+        version: &'a Option<String>,
+    }
```
brooksmtownsend, line 500: Any reason why we shouldn't omit this in favor of just a tuple? I assume clarity in the serialized output, but wanted to make sure it was needed

brooksmtownsend (changes requested):
A couple request + one nit, it would be really nice to get a test or two behind this if you could @vados-cosmonic

crates/wash-cli/src/app/mod.rs
vados-cosmonic: Thanks this is a great catch -- I went with the `required_unless_present`!

crates/wash-cli/src/app/mod.rs
vados-cosmonic: Fixed this like the other one -- thanks again!

crates/wash-cli/src/app/mod.rs
vados-cosmonic: Ah so I wanted to make sure that we put out an *object* -- in the future if we expand the information we return here, it will almost surely be in object form -- a tuple is serialized as a list, so this ends up making a list of lists. 

The change of shape would be annoying to change in downstream things that read this output, but if we do it right we can just *add* fields and not have to change the shape of the output

vados-cosmonic:
Hey @brooksmtownsend finally got a nice test in here for both -- would appreciate (yet) another review

brooksmtownsend (approved):

PR #3063, opened 2024-09-19
chore(wash-cli): update wasi rust target

vados-cosmonic:
## Feature or Problem

Let's get rid of all the warnings that show up on examples :)

We can probably delay releasing this in the next version a little bit, since people may be running with older versions of Rust which don't have the warning


rvolosatovs (approved):

brooksmtownsend (approved):
Can't wait to cut over to wasip2 officially 😄

PR #3066, opened 2024-09-19
fix(runtime): Ensures we have the proper limits set for max components

thomastaylor312:
It turns out that there are a bunch of other settings that we could have hit the limit on that we needed to set. These new settings are all set to the same as max components because you could have that number of individual components that would each have their own core module and stack.

I tested this under load to make sure it actually works

### DIFF
--- a/crates/runtime/src/runtime.rs
+++ b/crates/runtime/src/runtime.rs
@@ -135,6 +135,9 @@ impl RuntimeBuilder {
         #[allow(clippy::cast_possible_truncation)]
         pooling_config
             .total_component_instances(self.max_components)
+            .total_core_instances(self.max_components)
+            .total_gc_heaps(self.max_components)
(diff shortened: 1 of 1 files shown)


crates/runtime/src/runtime.rs
```diff
@@ -135,6 +135,9 @@ impl RuntimeBuilder {
         #[allow(clippy::cast_possible_truncation)]
         pooling_config
             .total_component_instances(self.max_components)
+            .total_core_instances(self.max_components)
+            .total_gc_heaps(self.max_components)
+            .total_stacks(self.max_components)
```
brooksmtownsend, line 140: So, put another way, each component instance gets one core instance, one GC heap, and one stack?

thomastaylor312: Yes that is correct

PR #3074, opened 2024-09-20
fix: Enable default features for url crate

domenukk:
In order to make the rust-url compatible with no_std, the crate needs to introduce a `std` feature and make it default.
See https://github.com/servo/rust-url/pull/831.

In order to reduce impact, downstream libraries should leave url's default features enabled (no other features are currently `default`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -322,7 +322,7 @@ tracing-opentelemetry = { version = "0.24", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
 ulid = { version = "1", default-features = false }
 unicase = { version = "2.7.0", default-features = false }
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Hey @domenukk thanks or the contribution -- this repository requires conventional commits and DCO signoff!

If you'd like we can duplicate this change ourselves, or you can edit this PR :)

domenukk:
I've tried to do it, but feel free to just copy & paste it into a new branch otherwise :)

domenukk:
> (we would explicitly enable `std`)

Note that explicitly enabling std is not possible (yet) as the feature does not yet exist...
It's a chicken-egg problem, really.

PR #3075, opened 2024-09-20
feat(providers): enable auto creation of buckets

vados-cosmonic:
## Feature or Problem

Once this is released, A new version of `keyvalue-nats` (v0.3.0) and update `wash dev` to use *that* by default (the update in here is only because the version was already old)


examples/rust/components/keyvalue-messaging/local.wadm.yaml
```diff
@@ -36,6 +36,8 @@ spec:
               - name: wasmcloud
                 properties:
                   bucket: "WASMCLOUD"
+                  #enable_bucket_auto_create: "true"
```
joonas, line 39: Is there a reason to not have this enabled here?

If so, would it be worth adding a comment above for why/when to enable this?

vados-cosmonic: That's a good point -- this *would* make using this example easier, not sure what people were doing before (they must have been creating the bucket manually). 


Will enable this for the local example!

brooksmtownsend (approved):

Issue #3082, opened 2024-09-23
[BUG] Invocation of custom interfaces' functions will hang if 'link name' is set in Wadm manifests.


ossfellow:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [X] other / not sure

### Describe the bug

When `wasmCloud v1.2.0` and `Wadm v0.14.0` were released, suddenly my custom provider stopped responding to function invocations, or rather such calls would hang!
So, this weekend, I built a `debugging version` of wasmCloud, turned on `tracing`, and captured the interactions, which revealed `link name`, which I set for all links, is not being picked up (i.e. `default` is used no matter what), causing misconfiguration and a broken invocation path.

To rule out my custom provider as the root cause, I used the wasmCloud's example [custom-template provider](examples/rust/providers/custom-template) to reproduce the issue. I added `link names` to the `wadm.yaml` file (and corrected a typo in provider's image name):

```yaml
apiVersion: core.oam.dev/v1beta1
kind: Application
metadata:
  name: custom-template-provider
  annotations:
    version: v0.0.1
    description: 'An application with a provider and component built for extending wasmCloud with custom capabilities'
spec:
  components:
    - name: test-component
      type: component
      properties:
        image: file://./component/build/custom_template_test_component.wasm
      traits:
        - type: spreadscaler
          properties:
            instances: 1
        # Link the component to the provider on wasmcloud:example/system-info
        - type: link
          properties:
            name: system-info
            target: custom-template
            namespace: wasmcloud
            package: example
            interfaces: [system-info]

    - name: custom-template
      type: capability
      properties:
        image: file://./build/wasmcloud-provider-custom-template.par.gz
        id: custom-template
        config:
          - name: provider-config
            properties:
              foo: bar
      traits:
        # Link the provider to the component on wasmcloud:example/process-data
        - type: link
          properties:
            name: process-data
            target: test-component
            namespace: wasmcloud
            package: example
            interfaces: [process-data]
``` 

Here's a snippet of the `wasmcloud.log` file, which shows the actions resulting from issuing `wash call custom-template wasmcloud:example/system-info.call` command (`wasmcloud.log` file is attached, but I couldn't attach wadm.log):

```log
[...]

PR #3088, opened 2024-09-23
fix(provider-sdk): override OTEL vars from config

vados-cosmonic:
This commit updates the `wasmcloud-provider-sdk` to override OTEL related configuration via provided config.

At present, OTEL config is specified to the host but *not* passed along (via ENV) to providers because of isolation between the environments in which binaries wrong (we wouldn't want sensitive ENV variables to make it through). 

While the answer would normally be to pass along the related ENV so that it can affect the libraries used by providers (if that is enabled), some client libraries in the OTEL ecosystem ignore manually set variables with ENV, which means it would be difficult to configure the host and provider independently if ENV was passed through (the variables would always be jointly set for host and spawned providers). 

Here, we choose to override the `otel_config` received from the host via config, which enables host and providers to individually set and control their OTEL config, at the very least in terms of setting endpoints to deliver to.

### DIFF
--- a/crates/provider-sdk/src/otel.rs
+++ b/crates/provider-sdk/src/otel.rs
@@ -53,13 +53,33 @@ macro_rules! initialize_observability {
                 ..
             } = $crate::provider::load_host_data().context("failed to load host data")?;
 
+            // Update OTEL configuration with overrides if provided via config to the provider
+            let mut otel_config = otel_config.clone();
+            for (k, v) in config.iter() {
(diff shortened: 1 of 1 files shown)


joonas (approved):

brooksmtownsend:
@vados-cosmonic this fixes #3082 right?

vados-cosmonic:
@brooksmtownsend Almost! `provider-sdk`needs to be released, and some examples should be updated just to make it easy for people to grok :)

PR #3101, opened 2024-09-24
feat(wit): workflow for wasmcloud:bus

ricochet:
Signed-off-by: Bailey Hayes <behayes2@gmail.com>


ricochet (commented):
For those reviewing, note that after this is merged, we need to actually push the appropriate tag wit-wasmcloud-bus-v0.1.0 for the release workflow.

.github/workflows/wit-wasmcloud-lattice-bus.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Extract tag context
+      id: ctx
+      run: |
+          version=${GITHUB_REF_NAME#wit-wasmcloud-secrets-v}
+          echo "version=${version}" >> "$GITHUB_OUTPUT"
+          echo "tarball=wit-wasmcloud-secrets-${version}.tar.gz" >> "$GITHUB_OUTPUT"
+          echo "version is ${version}"
+    - name: Build
+      run: |
+        tar cvzf ${{ steps.ctx.outputs.tarball }} -C wit secrets/wit
```
joonas, line 31:
```suggestion
        tar -cvzf ${{ steps.ctx.outputs.tarball }} -C wit secrets/wit
```

ricochet: Thanks! Caught a real bug with my copy-paste. Weirdly tar is a-ok with not passing in a hyphen here. I adjusted the other workflows I copied from to include this as well.

crates/component/src/lib.rs
```diff
@@ -28,7 +28,10 @@ pub mod wasi {
     pub use super::bindings::wasi::*;
     pub use ::wasi::*;
 }
-pub use bindings::wasmcloud;
+
+pub mod wasmcloud {
+    pub use super::bindings::wasmcloud::*;
+}
```
rvolosatovs, line 34: I'm a bit confused here, doesn't this achieve the same result as above?

.github/workflows/wit-wasmcloud-lattice-bus.yml
rvolosatovs: There's no bug here. `-` is not required by either GNU nor BSD tar (i.e. neither on Linux, nor on Mac by default), "hyphen-less" `tar` usage is actually much more common

crates/component/src/lib.rs
ricochet: By moving the bus wit down a level, it changed the bindings generation. I liked keeping the bindings in this file the same with wasmcloud::bus vs bus

rvolosatovs (approved):

PR #3104, opened 2024-09-24
refactor(wash-cli): reorganize `wash up` and `wash config`

vados-cosmonic:
## Feature or Problem

This commit starts the reorganization of `wash-cli` code into the `src/cmd` hierarchy, which should make it easier to separate shared functionality, (sub)command invocation and specific logic.


brooksmtownsend (approved):
I think we could reconsider the split here with wash-lib / wash-cli too 🤔

vados-cosmonic:
Yeah the refactor is far from done, but that's definitely up for discussion -- I know @joonas was asking some questions around that lately too

PR #3107, opened 2024-09-24
refactor(control-interface)!: prep for v2 release

vados-cosmonic:
## Feature or Problem

This commit prepares control-interface for v2.0.0 release, in particular:

- Trimming publically available API surface
- Ensuring provided are appropriately marked with `#[non_exhaustive]`

The goal of this prep is to prevent future breaking changes in `wasmcloud-control-interface` releases.

This code *does* represent a breaking change for `control-interface`.

### DIFF
--- a/crates/control-interface/src/lib.rs
+++ b/crates/control-interface/src/lib.rs
@@ -1,14 +1,21 @@
 //! # Control Interface Client
 //!
-//! This library provides a client API for consuming the wasmCloud control interface over a
-//! NATS connection. This library can be used by multiple types of tools, and is also used
-//! by the control interface capability provider and the wash CLI
+//! This library provides a client API for consuming the [wasmCloud control interface][docs-control-interface]
+//! over a NATS connection.
+//!
+//! This library can be used by multiple types of tools, and is also used
+//! by the control interface capability provider and the [`wash` CLI][wash-cli].
 //!
 //! ## Usage
-//! All of the [Client] functions are handled by a wasmCloud host running in the specified lattice.
+//!
+//! All of the [`Client`] functions are handled by a wasmCloud host running in the specified lattice.
+//!
 //! Each function returns a `Result<CtlResponse<T>>` wrapper around the actual response type. The outer
 //! result should be handled for protocol (timeouts, no hosts available) and deserialization errors (invalid response payload).
 //! The inner result is the actual response from the host(s) and should be handled for application-level errors.
+//!
+//! [docs-control-interface]: <https://wasmcloud.com/docs/hosts/lattice-protocols/control-interface>
+//! [wash-cli]: <https://wasmcloud.com/docs/ecosystem/wash>
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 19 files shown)


crates/control-interface/src/lib.rs
```diff
@@ -1,56 +1,96 @@
...
-/// - opaque string
-pub type LatticeTarget = String;
+pub use wasmcloud_core::{
+    ComponentId, InterfaceLinkDefinition, KnownConfigName, LatticeTarget, LinkName, WitInterface,
+    WitNamespace, WitPackage,
+};
+
+/// Enumeration of all kinds of identifiers in a wasmCloud lattice
+#[allow(unused)]
+#[non_exhaustive]
```
brooksmtownsend, line 43: Thoughts on dropping non_exhaustive on non public members?

brooksmtownsend (commented):
I have a nit here that generally applies, but this is overall a huge helpful change and really appreciate the attention to detail on doc comments, getter functions, etc!

brooksmtownsend (approved):
Meant to approve, since it's not a big issue

brooksmtownsend (approved):
Meant to approve, since it's not a big issue

crates/control-interface/src/lib.rs
vados-cosmonic: So... it turns out this isn't even used externally :) it's actually just used internally, and only to check if the identifiers are legit/usabble....

brooksmtownsend: Oh yeah, I just meant what do you think about removing `non_exhaustive` here since we're only constructing things internally

vados-cosmonic: Ahhhh -- yes, I can definitely do that!

brooksmtownsend (approved):

crates/control-interface/src/types/link.rs
```diff
@@ -1,39 +0,0 @@
...
-/// interface. An [`InterfaceLinkDefinition`] connects one component's import to another
-/// component's export, specifying the configuration each component needs in order to execute
-/// the request, and represents an operator's intent to allow the source to invoke the target.
-#[derive(Clone, Debug, Default, Deserialize, PartialEq, Eq, Serialize, Hash)]
-pub struct InterfaceLinkDefinition {
-    /// Source identifier for the link
-    pub source_id: ComponentId,
-    /// Target for the link, which can be a unique identifier or (future) a routing group
-    pub target: LatticeTarget,
-    /// Name of the link. Not providing this is equivalent to specifying "default"
```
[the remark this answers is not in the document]
vados-cosmonic: Ah thanks for catching this, didn't realize it was different.

crates/control-interface/src/types/link.rs
```diff
@@ -7,11 +7,16 @@ use crate::{
...
 
 /// A link definition between a source and target component (component or provider) on a given
-/// interface. An [`InterfaceLinkDefinition`] connects one component's import to another
+/// interface.
+///
+/// An [`InterfaceLinkDefinition`] connects one component's import to another
 /// component's export, specifying the configuration each component needs in order to execute
 /// the request, and represents an operator's intent to allow the source to invoke the target.
+///
+/// This InterfaceLinkDefinition is *distinct* from the one in `wasmcloud_core`, in that it is
```
brooksmtownsend, line 16: nit: changed name

crates/control-interface/src/types/link.rs
```diff
@@ -7,11 +7,16 @@ use crate::{
...
+/// partially resolved with information related to the required named configuration for a given link
 #[derive(Clone, Debug, Default, Deserialize, PartialEq, Eq, Serialize, Hash)]
-pub struct InterfaceLinkDefinition {
+pub struct ResolvedLinkDefinition {
```
brooksmtownsend, line 19: I think this one should be the "regular" `Link`, and then the one in `core` should be the `ResolvedLink` because it has the configuration resolved. Does that match up with what you were thinking?

crates/control-interface/src/types/link.rs
vados-cosmonic: Ah thanks this should just be "link definition"

crates/control-interface/src/types/link.rs
vados-cosmonic: OH I see what you mean now -- I was thinking the resolution here was getting the required named config, but you're right, the one in `core` actually has just the KV

brooksmtownsend (approved):

Issue #3049, opened 2024-09-18
[FEATURE] Go: wasip2 compilation target


lxfontes:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

wasmcloud is setting tinygo's compilation target to `wasi` for both wasi p1 and p2 targets.
https://github.com/wasmCloud/wasmCloud/blob/1ff476dcd61675a81d747091a1a94f1a4cd5fedb/crates/wash-lib/src/parser/mod.rs#L404

### Describe the solution you'd like

Would like the wasip2 target to use tinygo's wasip2 target.
This also requires changing the build command to pass the wit directory and wit world.

### Describe alternatives you've considered

The existing workaround is to use the language `other` and provide your own build command

```
name = "http_server"
language = "other"
type = "component"
version = "0.1.0"

[component]
build_command = "tinygo build -target=wasip2 --wit-package ./wit --wit-world example -o build/http_server.wasm main.go"
wit_world = "example"
wasm_target = "wasm32-wasi-preview2"
destination = "build/http_server_s.wasm"
```

lxfontes:
@brooksmtownsend we need to update the examples as they are in fact wasip1 :( 
If we leave them as-is we'd get
```
http-echo-tinygo ❯ wash build
# github.com/wasmcloud/wasmcloud/examples/golang/components/http-echo-tinygo/gen
gen/echo.go:4:13: note: in file included from echo.go!cgo.c:2:
gen/echo.go:5:13: fatal: 'stdlib.h' file not found
gen/echo.h:12:11: expected a bitfield
gen/echo.h:13:3: error: unknown type name 'size_t'
gen/echo.h:13:10: expected a bitfield
gen/echo.h:44:35: expected a bitfield
gen/echo.h:45:3: error: unknown type name 'size_t'
gen/echo.h:45:10: expected a bitfield
gen/echo.h:49:13: expected a bitfield
```

Added more context in the PR #3108

PR #3108, opened 2024-09-24
chore: wash Go wasip2 support

lxfontes:
Fixes https://github.com/wasmCloud/wasmCloud/issues/3049

tinygo wasi p2 automatically embeds wits and there is no way to disable this behaviour.

This PR:
- Extracts the wasip1 -> wasip2  adaptation step into a function
- Splits the Go component compilation path into 2 routes: wasip1 and wasip2
- wasip1: Compile, adapt to p2, embed wit world
- wasip2: Compile only
- Keeps the `rust` path doing wasip1 -> p2 adaption

🚨  Our existing examples will fail compilation as they specify `wasm32-wasi-preview2` even tho they are actually `wasm32-wasi-preview1`.
We can fix that in 2 ways:
1.  Update the examples' `wasmcloud.toml` to `wasm32-wasi-preview1`
2. Update the examples' actual code to be wasi-preview2 compatible

Personally leaning towards 1. so we can release `wash` and review all Go examples right after, migrating them to wasi p2.


ricochet:
I think if we target p2 we want the output of p2, how that happens isn't a detail the user necessarily cares about. 

Right now the examples want wasip2 and don't want to target wasip1. Said differently, if I have wasi-preview1 as my target, then no adaption to wasip2 should occur.

crates/wash-lib/src/build/component.rs
```diff
@@ -44,29 +42,37 @@ pub fn build_component(
...
+        match language_config {
             LanguageConfig::Rust(rust_config) => {
-                build_rust_component(common_config, rust_config, component_config)?
+                let rust_wasm_path =
+                    build_rust_component(common_config, rust_config, component_config)?;
+                if component_config.wasm_target == WasmTarget::WasiPreview2 {
+                    adapt_component_to_wasi_preview2(&rust_wasm_path, component_config)?
+                } else {
+                    rust_wasm_path
+                }
```
vados-cosmonic, line 53: SUPER NIT: I personally like a `match` here, mostly because I want it to be really obvious that we kind of make the "same" decision for p1 and modules, and then make a *different* one for p2.

crates/wash-lib/src/build/component.rs
```diff
@@ -118,6 +101,28 @@ pub fn build_component(
     }
 }
 
+pub fn adapt_component_to_wasi_preview2(
+    component_wasm_path: &PathBuf,
```
vados-cosmonic, line 105:
```suggestion
    component_wasm_path: impl AsRef<Path>
```

Also, what do you think about this being `pub` versus `pub(crate)`? I think the function is definitely useful as `pub`, but I wonder if it would be easier to use if it didn't require a `&ComponentConfig`, but exposed the things it actually uses -- the path to the WASM, etc.

As far as I can tell the vast majority of `ComponentConfig` is not used here.

That said, this is absolutely a NIT -- I'm just as happy to mark it `pub(crate)` and not expose something new to the outside world (which is still mostly us/`wash-cli`)

crates/wash-lib/src/build/component.rs
```diff
@@ -13,8 +13,6 @@ use wasi_preview1_component_adapter_provider::{
     WASI_SNAPSHOT_PREVIEW1_ADAPTER_NAME, WASI_SNAPSHOT_PREVIEW1_REACTOR_ADAPTER,
 };
 use wasm_encoder::{Encode, Section};
-use wit_bindgen_core::Files;
-use wit_bindgen_go::Opts as WitBindgenGoOpts;
```
vados-cosmonic, line 17: Oh now that these are gone... can we remove the deps?

Do we have to add some sort of warning somewhere to make sure people have the stuff installed/on their `PATH` somewhere? Moving to requiring the extra tool might warrant some extra verbosity around making sure people have it/know where to get it

vados-cosmonic (approved):
LGTM :rocket: 

Pending passing tests, stuff I left is mostly nits.

crates/wash-lib/src/build/component.rs
lxfontes: made it `pub(crate)` for now and addressed argument trait 👍

brooksmtownsend (approved):

PR #3114, opened 2024-09-25
feat: update to wasmtime 25

rvolosatovs:
## Feature or Problem

Technically this should mean WASI 0.2.1 is now supported, but I'll update tests and `wasmcloud-component` in a separate in a separate PR to ensure 0.2.0 is not broken and because I've encountered some challenges updating


vados-cosmonic (approved):
LGTM :rocket:

PR #3118, opened 2024-09-25
chore: Reconfigure Go examples ( wasip2 -> wasip1 )

lxfontes:
Related to #3108

Our tests checkout templates from git main branch ( wash-lib / wash-cli ).
Our Go examples are coded in wasip1 but declared as wasip2. It works due to this line https://github.com/wasmCloud/wasmCloud/blob/3ddf9d3176534bfd55ba9f2d029bd695e982996b/crates/wash-lib/src/parser/mod.rs#L409

In order to implement Go wasip2 we first need to correct our examples' targets.
Once wasip2 support is in we will repave examples as wasip2 proper.


brooksmtownsend (approved):
Looking forward to undoing this with refactoring to p2 😄

PR #3129, opened 2024-09-26
release: control-interface v2.0.0

vados-cosmonic:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -343,7 +343,7 @@ wasi-preview1-component-adapter-provider = { version = "25", default-features = 
 wasm-encoder = { version = "0.217", default-features = false }
 wasm-gen = { version = "0.1", default-features = false }
 wasmcloud-component = { version = "0", path = "crates/component", default-features = false }
-wasmcloud-control-interface = { version = "2.0.0", path = "./crates/control-interface", default-features = false }
+wasmcloud-control-interface = { version = "*", path = "./crates/control-interface", default-features = false }
 wasmcloud-core = { version = "^0.10.0", path = "./crates/core", default-features = false }
 wasmcloud-host = { version = "0", path = "./crates/host", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -343,7 +343,7 @@ wasi-preview1-component-adapter-provider = { version = "25", default-features =
 wasm-encoder = { version = "0.217", default-features = false }
 wasm-gen = { version = "0.1", default-features = false }
 wasmcloud-component = { version = "0", path = "crates/component", default-features = false }
-wasmcloud-control-interface = { version = "2.0.0", path = "./crates/control-interface", default-features = false }
+wasmcloud-control-interface = { version = "*", path = "./crates/control-interface", default-features = false }
```
vados-cosmonic, line 346: NOTE: we need this to actually have the project compile properly:

https://github.com/wasmCloud/wasmCloud/actions/runs/11057650280/job/30721939750

We can only specify released versions there so it's a bit of chicken-and-egg

rvolosatovs: For crate release dry runs we normally just (used to?) set `continue-on-error` for crates with dependencies on other crates in the workspace. Can't we do the same here?

vados-cosmonic: The dry run publish works just fine *without* any of these changes, but we use `smart-release` to do our releases right now rather than other tools.

Well so the thing is that's the run of `smart-release`, not just a publish dry run.

>   cargo smart-release --update-crates-index --no-publish --execute --no-changelog-preview --no-push  wasmcloud-control-interface

`smart-release`, as we're using it now won't run with 2.0.0 specified because it attempts to look it up the version we've specified as a released version.

That said, the real problem here is that we can't/aren't just point releasing a specific project at a specific version, but trying to resolve the entire mono repo. 

Another way to solve this problem would have been to release `wasmcloud-control-interface` manually from the command line, or just tag a random commit (because at this point on `main`, it *is* `v2.0.0`), but I can't do the first, and figured we should have a proper release commit for the second.

brooksmtownsend (approved):

PR #3131, opened 2024-09-26
chore(core)!: Rename OtelProtocol variants to lowercase

joonas:
## Feature or Problem

Since we include this in the host data to providers, we should standardize on lowercase for consistency.

### DIFF
--- a/crates/core/src/otel.rs
+++ b/crates/core/src/otel.rs
@@ -137,6 +137,7 @@ impl OtelConfig {
 }
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
 pub enum OtelProtocol {
     Grpc,
     Http,


crates/core/src/otel.rs
```diff
@@ -137,6 +137,7 @@ impl OtelConfig {
 }
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
```
brooksmtownsend, line 140: Should we change the `OtelSignal` enum here as well? Or really just all things that we serialize?

brooksmtownsend (approved):

crates/core/src/otel.rs
joonas: I would stick to just to the things that we serialize, `OtelSignal` is intended for the library's internal use only, so I don't see the need to rename

Issue #3080, opened 2024-09-22
[BUG] OpenTelemetry endpoints not set in provider host data config


jamesstocktonj1:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

The wasmcloud host does not pass on the OpenTelemetry endpoints when starting a new provider. This means that the `provider-sdk-go` is unable to retrieve the observability endpoint.

### Steps to reproduce

Start wasmcloud using the following environment variables:
```
WASMCLOUD_TRACES_ENABLED=true OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4318 wash up
```

Using some logs, I am able to print the raw decoded HostData string within `provider-sdk-go`. The logs show that the tracing boolean is set to true but the traces endpoint is blank:
```
{
    "host_id": "NBDBBFKSOMZBNQHXU4NKJ4K3TRVWDWIG73HQIKTML2EVAJ5ZFDTFLUJP",
    ...
    "otel_config": {
        "enable_observability": false,
        "enable_traces": true,
        "enable_metrics": null,
        "enable_logs": null,
        "observability_endpoint": null,
        "traces_endpoint": null,
        "metrics_endpoint": null,
        "logs_endpoint": null,
        "protocol": "Http",
        "additional_ca_paths": [],
        "trace_level": "info"
    }
}
```

### Expected behavior

I expect to see `"otel_config": {"traces_endpoint": "http://localhost:4318"}` within the HostData config.

### Environment

- OS: [Linux]
- Shell [bash]
```
wash          v0.34.0
├ nats-server v2.10.18
├ wadm        v0.14.0
└ wasmcloud   v1.2.1
```

brooksmtownsend:
Hey @jamesstocktonj1 , yes it looks like if we explicitly set the flags we will pass them down, but the environment variables don't "fill in" the OTEL configuration after the fact. I know @vados-cosmonic is working on something analogous here

vados-cosmonic:
[...]

PR #3135, opened 2024-09-26
chore: Explicitly use OTEL_EXPORTER_ environment variables for exporter configuration

joonas:
## Feature or Problem

In the interest of making it easier to pass these values down to providers, let's just directly use the environment variables.

## Related Issues

#3080


vados-cosmonic (approved):
LGTM :rocket:

PR #3137, opened 2024-09-27
fix: wash dev no duplicate components, ignore generated files

brooksmtownsend:
## Feature or Problem
This PR fixes two issues in `wash dev`
1. Previously, when generating a dependency for something like `wasi:keyvalue/store` and `wasi:keyvalue/atomics`, we would generate two capability components in the manifest. Now, I use a hashmap to remove the duplicate component during manifest creation.
    1. Additionally, these links are now combined into one link
1. We're watching paths recursively, so generated `target` / `node_modules` type directories triggers many many rebuilds. I added an `--ignore-dir` flag so that users can submit their own, but I made sure to ignore target + build to start.

### DIFF
--- a/crates/wash-cli/src/dev.rs
+++ b/crates/wash-cli/src/dev.rs
@@ -18,6 +18,7 @@ use serde::{Deserialize, Serialize};
 use tokio::io::AsyncBufReadExt as _;
 use tokio::process::Child;
 use tokio::{select, sync::mpsc};
+use tracing::debug;
 use wash_lib::app::AppManifest;
 use wash_lib::cli::stop::stop_provider;
 use wash_lib::common::CommandGroupUsage;
@@ -90,6 +91,11 @@ pub struct DevCommand {
     /// Path to code directory
     #[clap(name = "code-dir", long = "work-dir", env = "WASH_DEV_CODE_DIR")]
     pub code_dir: Option<PathBuf>,
+
+    /// Directories to ignore when watching for changes. This should be set
+    /// to directories where generated files are placed, such as `target/` or `dist/`
+    #[clap(name = "ignore-dir", long = "ignore-dir")]
+    pub ignore_dirs: Vec<PathBuf>,
 
     /// Whether to leave the host running after dev
     #[clap(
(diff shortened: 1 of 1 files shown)


crates/wash-cli/src/dev.rs
```diff
@@ -1514,7 +1544,7 @@ async fn run_dev_loop(
     let wit_implied_deps = discover_dependencies_from_wit(resolve, world_id)
         .context("failed to resolve dependent components")?;
     eprintln!(
-        "Detected component dependencies: {:?}",
+        "🔎 Detected component dependencies: {:?}",
```
lachieh, line 1547: `emoji::SEARCH`?

crates/wash-cli/src/dev.rs
```diff
@@ -1875,6 +1923,10 @@ pub async fn handle_command(
 
     // Watch FS for changes and listen for Ctrl + C in tandem
     eprintln!("👀 watching for file changes (press Ctrl+c to stop)...");
+    let _ = reload_rx.try_recv();
+    let _ = reload_rx.try_recv();
+    let _ = reload_rx.try_recv();
+    let _ = reload_rx.try_recv();
```
lachieh, line 1929: What does this do?

brooksmtownsend: Ugh this was a leftover. I threw in a bunch to try and make sure we don't immediately rebuild but I think we need at most one

crates/wash-cli/src/dev.rs
brooksmtownsend: I actually fixed this but I must not have force pushed.

crates/wash-cli/src/dev.rs
brooksmtownsend: I actually fixed this but I must not have force pushed.

lachieh: i don't believe you

crates/wash-cli/src/dev.rs
```diff
@@ -91,6 +92,11 @@ pub struct DevCommand {
     #[clap(name = "code-dir", long = "work-dir", env = "WASH_DEV_CODE_DIR")]
     pub code_dir: Option<PathBuf>,
 
+    /// Directories to ignore when watching for changes. This should be set
+    /// to directories where generated files are placed, such as `target/` or `dist/`
```
vados-cosmonic, line 96: Should the first sentence be separated from the second? And maybe an indication it can be specified multiple times

vados-cosmonic (approved):
Approved pending the changes that were already pointed out! My thing is just a nit.

crates/wash-cli/src/dev.rs
brooksmtownsend: Good idea, I'll incorporate this in #3140 just to keep merging simple!

PR #3142, opened 2024-09-27
feat(examples): add rust developer-starter-kit

brooksmtownsend:
Draft to facilitate discussion.

## Feature or Problem
This PR adds what I've been calling a "developer starter kit" for Rust component development. It uses each standard capability that we include in the wasmcloud_component crate and support with `wash dev`, doing the right things with streams, WASI interfaces, etc. This should be a great starting point for developers to come in, run `wash dev`, and use the standard capabilities immediately.

As you can see in this PR some of this code is a little cumbersome and could be cleaned up. My goal is for anything that's surfaced here as a pain point to be considered for a helper function in `wasmcloud_component` or re-use of an upstream crate.

Once we hone in on exactly how we want this example to look, we can replicate it for the other languages.

### DIFF
--- a/examples/rust/components/developer-starter-kit/.gitignore
+++ b/examples/rust/components/developer-starter-kit/.gitignore
@@ -1,0 +1,5 @@
+# Rust build artifacts
+Cargo.lock
+
+# Wash build artifacts
+build/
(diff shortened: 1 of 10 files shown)


vados-cosmonic:
Hey @brooksmtownsend do you still want this reviewed? I know we discussed maybe splitting up the functionalities somehow, but not sure where we landed!

brooksmtownsend:
Oh thanks @vados-cosmonic ! No, I know we discussed this one but I don't know if this is the best way to show this, and I'm just going to close this for now. If any onlookers wanted this, let me know!

PR #3162, opened 2024-09-29
feat(ci): --allow-dirty smart release for pre-release PR

vados-cosmonic:
This commit allows for uncommitted changes to be ignored during `cargo smart-release`, but *only* for the pre-release PR (not actual release).

This normally shows up/is useful if `Cargo.lock` is somehow behind on `main`. In that situation automation to generate the pre-release changes PR does not work since Cargo.lock will have changed (updated) and the changes will be uncommitted ("dirty").

Note that `--allow-dirty` is *not* turned on for actual releases via automation.

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -795,7 +795,8 @@ jobs:
         run: |
           git config --global user.email "automation@wasmcloud.com"
           git config --global user.name "wasmCloud automation"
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
I suppose this would only very rarely occur, would this even ever happen outside of regenerating lock files? LGTM

vados-cosmonic:
Yeah you're right... I don't think it could? Was scratching my head at how it could happen in the first place, then realized that `Cargo.lock` must have been getting updated *during* the action

PR #3171, opened 2024-09-30
release(wasmcloud): v1.3.0-rc.1

brooksmtownsend:
## Feature or Problem
Bumping wasmCloud to v1.3.0 so we can cut a release candidate, test, and possibly promote to the 1.3.0 release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6888,7 +6888,7 @@ dependencies = [
 
 [[package]]
 name = "wash-cli"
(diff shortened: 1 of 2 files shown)


vados-cosmonic (approved):
LGTM :rocket: 

We'll need to update `wash-cli` to use this as well, once we do that we should be able to close https://github.com/wasmCloud/wasmCloud/issues/3080

PR #3187, opened 2024-10-01
feat(wash)!: Adds support for fetching dependencies automatically

thomastaylor312:
This uses the newly available wit commands from wasm-pkg-tools to fetch things from upstream registry. I have only updated select examples (that are less likely to be used while we finish the last bits of work) so we could test things.

I also fixed the wash build tests so they could run properly in parallel (which hopefully makes them faster too). Once we have all wasi cloud wit in OCI, we can update all deps and roll to a new version of wash. This will involve updating git ignores to ignore the deps directory, but leaving the current deps intact for those on older versions of wash for at least the next version or two. Once we're past that we can remove the deps dirs


thomastaylor312:
This should be ready for review. Please note that until we push wasi:logging (probably tomorrow morning) the tests will fail here, and I'll need to remove the patch section as soon as I cut wkg 0.7.4 (once again, probably tomorrow morning).

.github/workflows/examples-components.yml
```diff
@@ -45,7 +45,8 @@ jobs:
     strategy:
       matrix:
         wash-version:
-          - 0.29.2
+          # TODO(thomastaylor312): Reenable/update this once we release the new version of wash with the deps stuff
+          # - 0.29.2
```
brooksmtownsend, line 49: Don't forget or you will be forsworn!

Cargo.toml
```diff
@@ -384,3 +387,7 @@ wrpc-interface-http = { version = "0.29", default-features = false }
 wrpc-runtime-wasmtime = { version = "0.23", default-features = false }
 wrpc-transport = { version = "0.27", default-features = false }
 wrpc-transport-nats = { version = "0.24.1", default-features = false }
+
+[patch.crates-io]
+wasm-pkg-client = { git = "https://github.com/thomastaylor312/wasm-pkg-tools.git", branch = "fix/dependency_override" }
+wasm-pkg-core = { git = "https://github.com/thomastaylor312/wasm-pkg-tools.git", branch = "fix/dependency_override" }
```
brooksmtownsend, line 393: I think I saw this reflected in the PR, but just make sure we get these removed before merge

crates/wash-cli/src/bin/wash.rs
```diff
@@ -238,6 +240,9 @@ enum CliCommand {
     /// Serve a web UI for wasmCloud
     #[clap(name = "ui")]
     Ui(UiCommand),
+    /// Create wit packages and fetch wit dependencies for a component
+    #[clap(name = "wit", subcommand)]
+    Wit(WitCommand),
```
brooksmtownsend, line 245: We should audit our top level commands soon ™️

crates/wash-cli/src/wit.rs
```diff
@@ -0,0 +1,104 @@
...
+#[derive(Debug, Args, Clone)]
+pub struct BuildArgs {
+    /// The directory containing the WIT files to build.
+    #[clap(short = 'd', long = "wit-dir", default_value = "wit")]
+    pub dir: PathBuf,
+
+    /// The name of the file that should be written. This can also be a full path. Defaults to the
+    /// current directory with the name of the package
+    #[clap(short = 'f', long = "file")]
+    pub output_file: Option<PathBuf>,
```
brooksmtownsend, line 48: nit: Other places in wash we've called this a destination / output. I think this is fine, but we should make them consistent at least in a followup. (this is a good first issue)

crates/wash-cli/src/wit.rs
```diff
@@ -0,0 +1,104 @@
...
+                file_name.push_str(&format!("@{version}"));
+            }
+            file_name.push_str(".wasm");
+            PathBuf::from(file_name)
+        };
+
+        tokio::fs::write(&output_path, bytes).await?;
+        // Now write out the lock file since everything else succeeded
+        lock_file.write().await?;
+        Ok(format!("WIT package written to {}", output_path.display()).into())
```
brooksmtownsend, line 90: Do we need to `flush` the lock_file here to doubley be sure it writes?

crates/wash-cli/src/wit.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl FetchArgs {
+    pub async fn run(self) -> anyhow::Result<CommandOutput> {
+        let client = self.common.get_client().await?;
+        let wkg_config = wasm_pkg_core::config::Config::load().await?;
+        let mut lock_file = LockFile::load(false).await?;
+        monkey_patch_fetch_logging(wkg_config, self.dir, &mut lock_file, client).await?;
+        // Now write out the lock file since everything else succeeded
+        lock_file.write().await?;
+        Ok("Dependencies fetched".into())
```
brooksmtownsend, line 102: Same comment re: flush

Cargo.toml
```diff
@@ -225,6 +225,7 @@ data-encoding = { version = "2", default-features = false }
 deadpool-postgres = { version = "0.13", default-features = false }
 dialoguer = { version = "0.10", default-features = false }
 docker_credential = { version = "1.3.1", default-features = false }
+etcetera = { version = "0.8", default-features = false }
```
brooksmtownsend, line 228: can we kill the `homedir` crate if we have this?

brooksmtownsend (commented):
All nits/comments/questions, but largely looks great to me. Leaving it unapproved for now to avoid accidental merge

Cargo.toml
thomastaylor312: Oh I was going to and forgot. Thanks for the reminder

crates/wash-cli/src/wit.rs
thomastaylor312: I was trying to mostly match `wkg wit` here. But I did have to rename the arg anyway because we already have an "output"

brooksmtownsend (commented):
LGTM! Requesting we don't merge this until we cut wash 0.35 just to keep the list of changes somewhat minor

brooksmtownsend (approved):
LGTM! Requesting we don't merge this until we cut wash 0.35 just to keep the list of changes somewhat minor

PR #3200, opened 2024-10-02
fix(ci): Error out when version doesn't exist

thomastaylor312:
This bumps our default back to 0.6.0 which does have binaries attached to it


vados-cosmonic (approved):
LGTM :rocket:

PR #3201, opened 2024-10-02
build: update `wrpc`

rvolosatovs:
## Feature or Problem

update to latest `wrpc`

Also ran `cargo update` since I was bumping deps anyway

## Consumer Impact

A few bindgen fixes and a fix in nested future handling.
wasmCloud does not use nested futures, so no breaking changes on protocol level


vados-cosmonic (approved):
LGTM :rocket:

PR #3202, opened 2024-10-02
ci: downgrade 16/8 core runners

brooksmtownsend:
## Feature or Problem
This PR downgrades our use of 16 and 8 core runners to assist with the ask of CNCF project maintainers to use the default runners where possible. I plan on watching the CI on this PR to understand what takes a long time, and I think we should optimize from there.

Other things that I'd like to do
- [x] Combine CI runs when merging to main by merging multiple PRs in the merge group at the same time
- [x] Remove unneeded action runs on PR where possible

### DIFF
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -46,11 +46,14 @@ jobs:
         with:
           tool: wit-bindgen-cli
 
+      # Run builds and tests only on Linux and macOS for PRs, Windows on `main`
       - name: Build wash
+        if: ${{ matrix.os != 'windows-latest' || github.ref == 'refs/heads/main' }}
         run: make build
         working-directory: ./crates/wash-cli
 
       - name: Run all wash & wash-lib unit tests
+        if: ${{ matrix.os != 'windows-latest' || github.ref == 'refs/heads/main' }}
         run: make test-wash-ci
         working-directory: ./crates/wash-cli
 
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -26,17 +26,17 @@ jobs:
     strategy:
       fail-fast: false # Ensure we can run the full suite even if one OS fails
       matrix:
-        os: [ubuntu-22.04, windows-latest-8-cores, macos-12]
+        os: [ubuntu-22.04, windows-latest, macos-12]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


vados-cosmonic:
If the general ask is to reduce our usage -- maybe we could start not running some builds (ex. the windows build takes *forever*) until we're doing a release?

brooksmtownsend:
@vados-cosmonic 100%, looking at the actions it looks like the Windows builds and windows tests take 20+ minutes, and we should run those perhaps only on `main`. The only other action that takes more than 10 minutes is wash integration tests, which we should try and optimize

vados-cosmonic:
I mean I'd say the windows tests should be done before releases, not even on `main` -- I think the number of people using random commits to run on windows rounds to zero -- and the problem is that despite our best efforts the `main` CI results are red from time to time, I'm not sure anyone is watching that signal.

Maybe the windows tests are a good candidate for some sort of scheduled daily spot checks or something, but then again someone would need to be responsible for that signal-- it seems like the real touch-point for making sure windows works is releases, as most people will be using the project based on an official release.

As for the wash tests, since we now are OK with `testcontainers`, they are trivially serializable. What might be *more* important may be figuring out why they weren't so serializable with just baremetal instances... I think someone could solve that effectively given some time to really dig into it.

.github/workflows/wash.yml
```diff
@@ -26,17 +26,17 @@ jobs:
     strategy:
       fail-fast: false # Ensure we can run the full suite even if one OS fails
       matrix:
-        os: [ubuntu-22.04, windows-latest-8-cores, macos-12]
+        os: [ubuntu-22.04, windows-latest, macos-12]
```
joonas, line 29: Consider bumping `macos-12` here to [`macos-13` or `macos-14`](https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images)

joonas (approved):
I'll be interested to see the difference it makes (or not?) to build times

vados-cosmonic (approved):
LGTM :rocket:

.github/workflows/wash.yml
brooksmtownsend: Will bump to 13 here since 14 changes to the arm runners, and I don't want to accidentally do what wadm did to break x86 builds

PR #3209, opened 2024-10-03
fix(control-interface)!: wrong data returned from provider auction

vados-cosmonic:

### DIFF
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -45,6 +45,13 @@ jobs:
         uses: taiki-e/install-action@07b43ed508403fa32c11f04dd608931cdf6748e2
         with:
           tool: wit-bindgen-cli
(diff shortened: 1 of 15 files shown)


ricochet (approved):
Possible to add tests where this is used? Approving since these are just getter updates but we need coverage in a follow-up.

vados-cosmonic:
@ricochet Sure, I'll add that right now then -- the getters are all new

joonas:
@vados-cosmonic since the getters are new, and you're essentially just fixing a bug, should this really be classified as a breaking change? I don't feel like it is

vados-cosmonic:
Yeah it doesn't *have* to be -- but given that the previous functionality was semantically wrong I was thinking it would be a big change sematics-wise. 

Either way the plan is to `yank` the old version though!

thomastaylor312 (approved):

PR #3211, opened 2024-10-03
typescript ci

lachieh:
- **ci: align typescript projects around turborepo standards**
- **ci: break up washboard workflow into typescript flows**


vados-cosmonic:
Nice -- these are the exact kind of changes we need to make to break up `washboard.yml` :+1:

.github/workflows/ts-build.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Checkout
+        uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938
+
+      - name: Setup
+        uses: ./.github/actions/ts-setup
+        with:
+          package-manager: yarn
+          working-directory: ./typescript
+
+      - name: Build
```
vados-cosmonic, line 29: if we always expect to use `yarn`, what do you think about just going for `yarn` everywhere?

i.e. Maybe we can avoid taking options to start with in `ts-setup` and "just" standardize on `yarn` for now?

.github/workflows/ts-release-app-washboard.yml
```diff
@@ -0,0 +1,24 @@
...
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16
+        with:
+          name: washboard
+
+      - name: Create Release
+        uses: ncipollo/release-action@2c591bcc8ecdcd2db72b97d6147f871fcd833ba5
+        with:
+          draft: false
+          prerelease: false
+          makeLatest: false
```
vados-cosmonic, line 23: NIT: it would be nice to do some detection here and enable `draft` and `prerelease`

There are many ways to do it -- but I think in wasmCloud we sometimes do a check to see if there is something *after* the `vx.x.x`.

Please feel free to handle this later though in a follow up or never -- if we're fine with making node pre-releases just regular releases on GH that's fine too

.github/workflows/ts-release.yml
```diff
@@ -0,0 +1,20 @@
...
+  # TODO: Add workflows for releasing packages
+  # # typescript/packages/*
+  # lattice-client-core:
+  #   name: packages/lattice-client-core
+  #   if: startsWith(github.ref, 'refs/tags/typescript/lattice-client-core/v')
+  #   uses: ./.github/workflows/ts-release-app-lattice-client-core.yml
+  # lattice-client-react:
+  #   name: packages/lattice-client-react
+  #   if: startsWith(github.ref, 'refs/tags/typescript/lattice-client-react/v')
+  #   uses: ./.github/workflows/ts-release-app-lattice-client-react.yml
```
vados-cosmonic, line 20: If this didn't make it in we can remove? Or now that washboard-ui works can we enable these?

.github/workflows/ts-test-e2e.yml
```diff
@@ -0,0 +1,50 @@
...
+    name: E2E Tests
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        wash:
+          - version: 0.32.1
+          - version: 0.33.0
+          - version: 0.34.0
```
vados-cosmonic, line 19:
```suggestion
          - version: 0.34.1
```

.github/workflows/ts-test.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: ./.github/actions/ts-setup
+        with:
+          package-manager: yarn
+          working-directory: ./typescript
+
+      - name: Run Tests
+        run: yarn turbo lint format test:unit
+
+      # TODO: add a step here for uploading test results once unit tests are running
+
```
vados-cosmonic, line 33: Forgot to finish this? Could be a follow up as well.

.github/workflows/ts-release-app-washboard.yml
```diff
@@ -0,0 +1,24 @@
...
+    if: startsWith(github.ref, 'refs/tags/typescript/apps/washboard-ui/v')
+
+    steps:
+      - name: Download Asset
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16
+        with:
+          name: washboard
+
+      - name: Create Release
+        uses: ncipollo/release-action@2c591bcc8ecdcd2db72b97d6147f871fcd833ba5
```
vados-cosmonic, line 19: Any reason to use this over `softprops/action-gh-release`?

.github/workflows/typescript.yml
```diff
@@ -0,0 +1,37 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: washboard-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CI: true
```
vados-cosmonic, line 22: This is already set by the platform usually (looks like [GH does it too](https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/))

.github/workflows/ts-test.yml
lachieh: Yeah this will be a follow up PR. We don't have any tests to run 🙃

.github/workflows/ts-release-app-washboard.yml
lachieh: added an action for this on the latest commit

.github/workflows/ts-test-e2e.yml
lachieh: I wonder if we can automate rolling updates to this...

vados-cosmonic (approved):
LGTM :rocket: 

Let's get this in there and iterate -- I like the `extract-tag-info` in-repo action.... want to start putting it everywhere.

Also, BTW, please feel free to start using `actions/github-script` -- I honestly don't know why we've been tied to `bash` for so long when JS is right there (and a bit more powerful as a GH-native solution).

lachieh:
Noted. I do like that the script has pre-authenticated instance of `octokit.js`. Will investigate.

Issue #3216, opened 2024-10-04
[FEATURE] Tinygo: Customize scheduler & garbage collection


lxfontes:
### Affected project(s)

- [X] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

just a feature.

Looking for the ability to customize tinygo scheduler & garbage collector behavior, overriding default behaviour.

### Describe the solution you'd like

In wasmcloud.toml, under the `tinygo` section: 

```
[tinygo]
scheduler = none | tasks | asyncify
garbage_collector = none | conservative | leaking
```

When present, they will be forwarded to `tinygo build -scheduler <value> -gc <value>`.
When not present, it won't forward any arguments to `tinygo build`, taking the default behavior  [described here](https://github.com/tinygo-org/tinygo/blob/release/targets/wasip2.json) 

This is the 🔝 preference.

Another option is:

```
[tinygo]
extra_build_arguments = "-scheduler none -gc none"
```

which is more flexible, however could result in invalid arguments in collisions ( ex: `-o my.wasm` ).

### Describe alternatives you've considered

Using component type `custom` and providing a build_command is the existing alternative.

### Additional context

Looking to change only the wasip2 target, as we force `scheduler=none` on wasip1.

PR #3217, opened 2024-10-04
feat: Tinygo scheduler & garbage collection overrides

lxfontes:
Changes to wasmcloud.toml as per #3216 

```
[tinygo]
scheduler = none | tasks | asyncify
garbage_collector = none | conservative | leaking
```

### DIFF
--- a/crates/wash-lib/src/build/component.rs
+++ b/crates/wash-lib/src/build/component.rs
@@ -314,12 +314,12 @@ async fn build_tinygo_component(
 
             if let Some(scheduler) = &tinygo_config.scheduler {
                 args.push("-scheduler");
-                args.push(scheduler);
+                args.push(scheduler.as_str());
             }
 
             if let Some(gc) = &tinygo_config.garbage_collector {
                 args.push("-gc");
-                args.push(gc);
+                args.push(gc.as_str());
             }
 
             args.push(".");
--- a/crates/wash-lib/src/build/component.rs
+++ b/crates/wash-lib/src/build/component.rs
@@ -297,7 +297,8 @@ async fn build_tinygo_component(
             "-no-debug",
             ".",
         ],
-        WasmTarget::WasiP2 => vec![
+        WasmTarget::WasiP2 => {
+            let mut args = vec![
             "build",
             "-o",
             filename.as_str(),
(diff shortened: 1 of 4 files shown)


crates/wash-lib/src/parser/mod.rs
```diff
@@ -407,6 +407,10 @@ pub struct TinyGoConfig {
     pub tinygo_path: Option<PathBuf>,
     /// Whether to disable the `go generate` step in the build process. Defaults to false.
     pub disable_go_generate: bool,
+    /// The scheduler to use for the TinyGo build. Override the default scheduler (asyncify). Valid values are: none, tasks, asyncify.
```
vados-cosmonic, line 410:
```suggestion
    /// The scheduler to use for the TinyGo build
    ///
    /// Override the default scheduler (asyncify). Valid values are: none, tasks, asyncify.
```

Leaving the space here gets you the summary line in the docsrs

crates/wash-lib/src/parser/mod.rs
```diff
@@ -407,6 +407,10 @@ pub struct TinyGoConfig {
...
+    pub scheduler: Option<String>,
```
vados-cosmonic, line 411: Should we check the values here? An enum would do well to restrain the values this can take on, these shouldn't change very often, right?

vados-cosmonic (approved):
LGTM :rocket: 

Feel free to address the comments in a follow up, if you end up taking them

crates/wash-lib/src/build/component.rs
```diff
@@ -309,8 +310,21 @@ async fn build_tinygo_component(
...
+                args.push(scheduler);
+            }
+
+            if let Some(gc) = &tinygo_config.garbage_collector {
+                args.push("-gc");
+                args.push(gc);
+            }
+
+            args.push(".");
+            args
```
brooksmtownsend, line 326: Is the `.` a new requirement? I don't see the `.` in my tinygo build commands, but perhaps this is required after the other flags?

crates/wash-lib/src/parser/mod.rs
brooksmtownsend: Agreed, at the very least an enum with an `Unknown(String)` variant with a warning log is a good idea. I would only add that if we think that it's likely for TinyGo to add additional scheduler/gc options in a future release, otherwise validating + failing is a good idea

brooksmtownsend (changes requested):
Agree w/ Victor's ask for an enum, and just checking the build command needs that `.`

crates/wash-lib/src/parser/mod.rs
lxfontes: added enum 👍

crates/wash-lib/src/build/component.rs
lxfontes: not a new requirement ( it's also in the wasip1 target ).
it's a package path selector and not needed when the package you want to build is the one in `go.mod` ( most cases when  building components ).
It does matter when building components in monorepos.

brooksmtownsend (approved):

PR #3219, opened 2024-10-04
fix(ci): Enable correct permissions for wash plugin wit CI

thomastaylor312:
Also fixes issue with other wit not checking out the actions dir during sparse checkout


vados-cosmonic (approved):
LGTM :rocket:

PR #3225, opened 2024-10-07
feat(component): Implement wasi:keyvalue/watch

Aditya1404Sal:
This PR produces implementations for `on_set` and `on_delete` functions of `watch.wit` :
The elements to watch for are to be specified via the link config with the preset of `watch: SET@key,DEL@key,....`
This adds the watch trigger to the provider process that monitors and watches for those events and invokes user specified guest functions once those events occur.

## Related Issues

#1868

## Testing

Watch triggers are added via config_put to the link_config, the provider on the other hand watches for changes in the kv-store in this case redis, then invokes the `on_set`  or  `on_delete` reactor functions when those particular conditions are fulfilled ( do [something] when [key] is set ) once those actions are detected wrpc invocations are sent to the target component and once those invoked functions execute , an outgoing request is made to a dedicated server with the key-value context that are checked via the test cases.


crates/provider-keyvalue-redis/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository.workspace = true
 
 [dependencies]
 anyhow = { workspace = true }
+async-nats = { workspace = true, features = ["ring"] }
```
brooksmtownsend, line 13: This _might_ prove to be a problem with our cross compilations, but I'm not 100% sure. We may just want to do a dry run of the release from this branch to double check

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -141,6 +163,54 @@ impl KvRedisProvider {
...
 
+    async fn invoke_on_set(&self, target: impl AsRef<str>, bucket: &str, key: &str, value: &str) {
+        let mut cx: async_nats::HeaderMap = async_nats::HeaderMap::new();
+        for (k, v) in
+        wasmcloud_provider_sdk::wasmcloud_tracing::context::TraceContextInjector::default_with_span(
+        )
+        .iter()
+    {
+        cx.insert(k.as_str(), v.as_str())
+    }
```
brooksmtownsend, line 174: @vados-cosmonic there's a macro that does this right? Is it https://docs.rs/wasmcloud-provider-sdk/0.11.0/wasmcloud_provider_sdk/macro.propagate_trace_for_ctx.html?

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -141,6 +163,54 @@ impl KvRedisProvider {
...
+        let func = "on_set";
+        let _ = wrpc
+            .invoke_values::<_, (&str, &str, &str), ((),)>(
+                Some(cx),
+                target.as_ref(),
+                func,
+                (bucket, key, value),
+                &[[]; 0],
+            )
+            .await;
```
brooksmtownsend, line 186: This looks right, but likely more custom than is needed? I think here you can use the generated wRPC function to do this! I believe it'll be under `wrpc::keyvalue::watcher::on_set()`

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -141,6 +163,54 @@ impl KvRedisProvider {
...
+        let func = "on_delete";
+        let _ = wrpc
+            .invoke_values::<_, (&str, &str), ((),)>(
+                Some(cx),
+                target.as_ref(),
+                func,
+                (key, bucket),
+                &[[]; 0],
+            )
+            .await;
```
brooksmtownsend, line 210: Same comments re: macro from provider SDK and `on_delete` that's generated by wit_bindgen_wrpc

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -317,6 +387,66 @@ impl keyvalue::batch::Handler<Option<Context>> for KvRedisProvider {
...
+impl keyvalue::watcher::Handler<Option<Context>> for KvRedisProvider {
+    async fn on_set(
+        &self,
+        cx: Option<Context>,
+        bucket: String,
+        key: String,
+        value: wit_bindgen_wrpc::bytes::Bytes,
+    ) -> wit_bindgen_wrpc::anyhow::Result<()> {
+        check_bucket_name(&bucket);
+        let target = cx.clone().and_then(|c| c.component).unwrap_or_default();
```
brooksmtownsend, line 399:
```suggestion
        let Some(target) = cx.as_ref().and_then(|c| c.component.clone()) else {
          error!("received on_set invocation without a source component");
          bail!("received on_set invocation without a source component")
        };
```

Quick suggestion here to ensure that if we receive an invocation without a component on the context, we just bail since there's not going to be a valid set of keys/values/etc

brooksmtownsend: Also a small nit to only clone the component ID in the case that the context exists

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -317,6 +387,66 @@ impl keyvalue::batch::Handler<Option<Context>> for KvRedisProvider {
...
+
+        // Add the key to the watched keys
+        let mut watched_keys = self.watched_keys.write().await;
+        watched_keys
+            .entry(key.clone())
+            .or_default()
+            .insert(WatchedKeyInfo {
+                event_type: WatchEventType::Set,
+                value: Some(value_str.clone()),
+                target: target.clone(),
```
brooksmtownsend, line 410: You _might_ be okay removing the clones here, since I don't see them used again

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -317,6 +387,66 @@ impl keyvalue::batch::Handler<Option<Context>> for KvRedisProvider {
...
+    }
+
+    async fn on_delete(
+        &self,
+        cx: Option<Context>,
+        bucket: String,
+        key: String,
+    ) -> wit_bindgen_wrpc::anyhow::Result<()> {
+        check_bucket_name(&bucket);
+        let target = cx.clone().and_then(|c| c.component).unwrap_or_default();
```
brooksmtownsend, line 428: Same comment regarding bailing out if this was `None` 😄

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +511,120 @@ impl Provider for KvRedisProvider {
...
+                anyhow::anyhow!("Failed to set keyspace notifications: {}", e)
+            })?;
+
+        let mut sources = self.sources.write().await;
+        sources.insert((target_id.to_string(), link_name.to_string()), conn);
+
+        // Start watching for keyspace events
+        let client_clone = client.clone();
+        let self_clone = self.clone();
+        tokio::spawn(async move {
```
brooksmtownsend, line 570: It's probably a good idea to store this task as a part of the sources so that we can kill the task later on if the component isn't linked anymore

Also, I think we only want to spawn this task if the [`LinkConfig { wit_metadata }`](https://docs.rs/wasmcloud-provider-sdk/0.11.0/wasmcloud_provider_sdk/struct.LinkConfig.html#structfield.wit_metadata) interface contains `watcher`, otherwise we don't need to watch the keys

brooksmtownsend (commented):
@Aditya1404Sal this is looking really good, most of the comments I left are mostly for some cleanup and to make it easier on you with the `wit_bindgen_wrpc` generated bits! For inspiration on how you may store the task for later, the HTTP server has some precedent here https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-http-server/src/path.rs#L46

brooksmtownsend (commented):
@Aditya1404Sal this is looking really good, most of the comments I left are mostly for some cleanup and to make it easier on you with the `wit_bindgen_wrpc` generated bits! For inspiration on how you may store the task for later, the HTTP server has some precedent here https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-http-server/src/path.rs#L46

crates/provider-keyvalue-redis/src/lib.rs
vados-cosmonic: Yup, @Aditya1404Sal you should be able to use `propagate_trace_for_ctx` -- here's [an example](https://github.com/wasmCloud/wasmCloud/blob/6fbf413df073f79b6d52a857b722f0cd7003a07e/crates/provider-keyvalue-nats/src/lib.rs#L328)

Aditya1404Sal:
Thank you for the review @brooksmtownsend 
I'll get the changes incorporated !

Aditya1404Sal:
I've made the suggested changes, all that's left is writing a `delete_link_as_source` function to handle cleanup 

I do want to take a look at the build step for the test components, because that's the main thing left unfinished in #2586

Aditya1404Sal:
@brooksmtownsend this PR is ready for review !!

brooksmtownsend:
Thank you @Aditya1404Sal ! Apologies for missing it this week, I'll review this next week!

brooksmtownsend (commented):
Just some initial comments, still working my way through the regular logic

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -163,6 +195,103 @@ impl KvRedisProvider {
     }
 }
 
+async fn invoke_on_set(wrpc: &WrpcClient, bucket: &str, key: &str, value: &Bytes) {
+    let mut cx: async_nats::HeaderMap = async_nats::HeaderMap::new();
+    for (k, v) in
+        wasmcloud_provider_sdk::wasmcloud_tracing::context::TraceContextInjector::default_with_span(
```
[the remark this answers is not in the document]
Aditya1404Sal: There's a `propogate_ctx` macro call but it's only used for a component -> provider function call whereas in the current use case the wRPC call needs a Nats Header map where the component is getting invoked instead.

Cargo.toml
```diff
@@ -115,6 +115,7 @@ opt-level = "z"
 lto = true
 
 [dependencies]
+axum = "0.7"
```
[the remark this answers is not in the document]
Aditya1404Sal: Ah that's true, my fault. I accidentally put it directly under dependencies

crates/provider-keyvalue-redis/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
+async-nats = { workspace = true }
```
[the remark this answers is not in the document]
Aditya1404Sal: Yes this ended up being used [here](https://github.com/Aditya1404Sal/wasmCloud/blob/b9431eb831945cd80c7b843a27fceeb66a44f112/crates/provider-keyvalue-redis/src/lib.rs#L199)

crates/provider-keyvalue-redis/src/lib.rs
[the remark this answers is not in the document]
Aditya1404Sal: The http.handle and messaging.handle-message also use the nats Headermap.

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +553,153 @@ impl Provider for KvRedisProvider {
...
+            redis::cmd("CONFIG")
+                .arg("SET")
+                .arg("notify-keyspace-events")
+                .arg("KEA")
+                .query_async::<_, ()>(&mut conn)
+                .await
+                .map_err(|e| {
+                    error!("Failed to set keyspace notifications: {e}");
+                    anyhow::anyhow!("Failed to set keyspace notifications: {}", e)
+                })?;
```
brooksmtownsend, line 613: After your refactor to move the logic from the `on_set` handler which configures the key watch, is there a reason why this is using `CONFIG SET` v.s. just the `WATCH` command? https://redis.io/docs/latest/commands/watch/

It looks like there's a good bit of logic here for watching keyspace events, but I was wondering if you could just `WATCH` keys that are provided as a part of config and then it will simplify this. Let me know if I'm missing something here!

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -394,6 +713,28 @@ impl Provider for KvRedisProvider {
         Ok(())
     }
 
+    #[instrument(level = "info", skip_all, fields(target_id = info.get_target_id()))]
+    async fn delete_link_as_source(&self, info: impl LinkDeleteInfo) -> anyhow::Result<()> {
+        let component_id = info.get_target_id();
+        let mut aw = self.sources.write().await;
+        aw.retain(|(src_id, _link_name), _| src_id != component_id);
```
brooksmtownsend, line 720: Perhaps this was before we added this, but we do include the `link_name` now in the LinkDeleteInfo! So you should be able to just `remove` the key here

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -317,6 +434,61 @@ impl keyvalue::batch::Handler<Option<Context>> for KvRedisProvider {
     }
 }
 
+impl keyvalue::watcher::Handler<Option<Context>> for KvRedisProvider {
+    async fn on_set(
+        &self,
+        cx: Option<Context>,
```
brooksmtownsend, line 440: I think this logic should go under `receive_link_config` since that's where you'll want to setup your watches

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +553,153 @@ impl Provider for KvRedisProvider {
...
+            let task = tokio::spawn(async move {
+                let mut pubsub = match client_clone.get_async_pubsub().await {
+                    Ok(pubsub) => pubsub,
+                    Err(e) => {
+                        error!("Failed to get pubsub connection: {}", e);
+                        return;
+                    }
+                };
+                pubsub.psubscribe("__keyevent@0__:set").await.unwrap();
+                pubsub.psubscribe("__keyevent@0__:del").await.unwrap();
```
brooksmtownsend, line 628: I have a feeling some of this might get replaced with the `WATCH` command, but just a note to change `unwrap` calls where possible to `.context("error info")` after importing the `anyhow::Context` crate, so that we don't panic 😄

Cargo.toml
```diff
@@ -149,6 +149,7 @@ wasmcloud-provider-sqldb-postgres = { workspace = true, optional = true }
 wasmcloud-tracing = { workspace = true, features = ["otel"], optional = true }
 
 [dev-dependencies]
+axum ={ workspace = true, features = ["query"]}
```
brooksmtownsend, line 152:
```suggestion
axum = { workspace = true, features = ["query"] }
```

The superest of super nits, since this will trigger a format later 😅

brooksmtownsend (changes requested):
This all looks functionally really good, left some requests for changes around exporting the `on_set` and `on_delete` function and calling it from the component, which should be configured at link time instead. Otherwise, just a couple of nits + Rust best practices. Please feel free to leave comments if you have questions! @thomastaylor312 and I pair reviewed this one since it was a good bit of code so you can tag either one of us.

crates/provider-keyvalue-redis/src/lib.rs
Aditya1404Sal: @brooksmtownsend Thats a great question ! So the primary reason for that was to specifically detect key:operation changes in an event driven manner.
If i'm not wrong, `WATCH` is mainly used for optimistic locking during transactions -- say for eg. it would watch for key modifications while a transaction is still going on and abort the transaction if the key was modified (So the WATCH would only work for the lifetime of that transaction -- it doesn’t persist or trigger events beyond that context.).

I'd have to setup a constant polling mechanism to make `WATCH` useful for tracking changes perpetually.

On the other hand, keyspace notifications (`CONFIG SET KEA`) are specifically designed for event-driven use cases. With the notifications enabled, Redis can publish events like SET, DEL, or EXPIRE to Pub/Sub channels. Since the configuration includes the key-value-operation sets beforehand (as specified in the link config), subscribing to specific message patterns would make watching for changes straightforward.
although the only downside to `CONFIG SET KEA` is that it uses quite a bit of CPU power

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -163,6 +195,91 @@ impl KvRedisProvider {
...
+                error!("Error chain:{}", error_chain);
+            }
+
+            #[cfg(debug_assertions)]
+            {
+                error!(
+                    "Debug backtrace: {:?}",
+                    std::backtrace::Backtrace::capture()
+                );
+            }
```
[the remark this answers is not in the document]
Aditya1404Sal: Oopsies ! I had a trouble while debugging the invocation part so that's why I added the manual error chaining 😅 . I'll definitely be removing it !

crates/provider-keyvalue-redis/src/lib.rs
thomastaylor312: Ah yep, just dug into this more and I see what's happening, so thanks for clarifying! I wonder if doing the `CONFIG SET` is outside the scope of the provider and should be done by the admin of the Redis database. Then the provider should just do the `PSUBSCRIBE` part of it

Aditya1404Sal:
I have made the changes as suggested, I still need to refine some parts so I'll squash these commits once I'm done with those !

crates/provider-keyvalue-redis/src/lib.rs
Aditya1404Sal: Oh you raise a very fair point, The `CONFIG SET KEA` would be outside the jurisdiction of the provider if we factor in the permissibilty and authority required to modify the database. Thank you for pointing that out !!!

Aditya1404Sal:
Hey @brooksmtownsend and @thomastaylor312 , I've made the Suggested changes and modified the tests and example too
I wanted to get the link config input format checked up with you
```rust
watch : SET@key:value,DEL@key,......
```

and of course lowercase `set` and `del` also work.
I just thought that having a single line comma separated string with the trigger key sets would be the best option here instead of having a map, Thoughts ?

I also took away the `CONFIG SET ......` part out of the provider's scope and left in in the documentation as instruction, Now it's the responsibility of the database administrator to have the config preset with the `notify-keyspace-events` enabled, and the provider will panic during linking if the configuration is not as expected.

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+                info!(url, "Established link at receive_link_config_as_source");
+                client
+            }
+            Err(err) => {
+                warn!(?err, "Failed to create Redis client for target [{target_id}]. Key-value operations will fail.");
+                bail!("Failed to create Redis client");
+            }
+        };
+        let mut conn = client.get_connection_manager().await.map_err(|e| {
+            error!("Failed to get async connection: {e}");
```
thomastaylor312, line 494: This should put the error as a field instead of as the message:

```suggestion
            error!(err = ?e, "Failed to get async connection");
```

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+            })
+            .map_or(DEFAULT_CONNECT_URL, |v| v);
+
+        let client = match redis::Client::open(url.to_string()) {
+            Ok(client) => {
+                info!(url, "Established link at receive_link_config_as_source");
+                client
+            }
+            Err(err) => {
+                warn!(?err, "Failed to create Redis client for target [{target_id}]. Key-value operations will fail.");
```
thomastaylor312, line 489: `target_id` should be a field instead of interpolated in the message string

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+                    error!("Failed to get keyspace notifications config: {e}");
+                    anyhow::anyhow!("Failed to get keyspace notifications config: {}", e)
+                })?;
+
+            let current_config = config_response.get(1).ok_or_else(|| {
+                error!("Unexpected response format from Redis CONFIG GET");
+                anyhow::anyhow!("Unexpected response format from Redis CONFIG GET")
+            })?;
+
+            if !current_config.contains('K')
```
thomastaylor312, line 519: I think a better check here would be to see if the config response is empty or empty string (can't quite tell which one), which would mean it is entirely disabled. If I am reading [the docs](https://redis.io/docs/latest/develop/use/keyspace-notifications/#configuration) right, there could be many valid options here that someone could have configured

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+            // Look for watch configuration in the format "watch: SET@key:value,DEL@key"
+            if let Some(watch_config) = config.get(&UniCase::new("watch")) {
+                for watch_entry in watch_config.split(',') {
+                    let watch_entry = watch_entry.trim();
+                    if watch_entry.is_empty() {
+                        continue;
+                    }
+                    let parts: Vec<&str> = watch_entry.split('@').collect();
+                    if parts.len() != 2 {
+                        error!("Invalid watch entry format: {watch_entry}. Expected FORMAT@KEY[:VALUE]");
```
thomastaylor312, line 551: Add `watch_entry` as a field instead of interpolating

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+
+            let mut watched_keys = self.watched_keys.write().await;
+
+            let config: HashMap<UniCase<&str>, &String> = config
+                .iter()
+                .map(|(k, v)| (UniCase::new(k.as_str()), v))
+                .collect();
+
+            // Look for watch configuration in the format "watch: SET@key:value,DEL@key"
+            if let Some(watch_config) = config.get(&UniCase::new("watch")) {
```
thomastaylor312, line 543: Can we break out this parsing into a separate function so it can be unit tested?

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+            }
+
+            let client_clone = client.clone();
+            let self_clone = self.clone();
+            let mut conn_clone = conn.clone();
+            let task = tokio::spawn(async move {
+                let mut pubsub = match client_clone.get_async_pubsub().await {
+                    Ok(pubsub) => pubsub,
+                    Err(e) => {
+                        error!("Failed to get pubsub connection: {}", e);
```
thomastaylor312, line 601: Error as a field instead of interpolated

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+                                .or_default()
+                                .insert(WatchedKeyInfo {
+                                    event_type: WatchEventType::Set,
+                                    value: Some(key_value_parts[1].to_string()),
+                                    target: target_id.to_string(),
+                                });
+                        }
+                        "DEL" | "del" => {
+                            if key_value.contains(':') {
+                                error!(
```
thomastaylor312, line 574: So as to not spam every place this happens with a comment, can you make sure that, in most cases, data isn't interpolated into the log messages in all places in this file? This leads to high cardinality when receiving logs and so most data should be a field instead of interpolated

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+                    let parts: Vec<&str> = watch_entry.split('@').collect();
+                    if parts.len() != 2 {
+                        error!("Invalid watch entry format: {watch_entry}. Expected FORMAT@KEY[:VALUE]");
+                    }
+
+                    let operation = parts[0].trim().to_uppercase();
+                    let key_value = parts[1].trim();
+
+                    match operation.as_str() {
+                        "SET" | "set" => {
```
thomastaylor312, line 558: Nit: You `to_uppercase` this above, so you should only need to match on `SET` or `DEL`

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +457,234 @@ impl Provider for KvRedisProvider {
...
+                            let key_value_parts: Vec<&str> = key_value.split(':').collect();
+                            if key_value_parts.len() != 2 {
```
thomastaylor312, line 560: I'm not sure if this behavior is right. If I am watching a key for a `SET` operation, I would want to get the newly changed value, not wait until I see a specific value. I think `SET@foo` is all that would need to be required. Let me know if I am misunderstanding here though

examples/rust/components/http-keyvalue-watcher/src/wasmcloud.lock
thomastaylor312, line 1: Was this file mistakingly committed? The wasmcloud.lock at the top level should be the only one needed

examples/rust/components/http-keyvalue-watcher/src/lib.rs
```diff
@@ -0,0 +1,135 @@
+use crate::bindings::wasi::http::outgoing_handler;
+use crate::bindings::wasi::http::types::*;
+use wasmcloud_component::wasi::logging::logging::{log, Level};
+mod bindings {
+    use super::KvWatcher;
```
thomastaylor312, line 5: Nit: Is there a reason this and the `export!` are here? They look a little weird here and you can just do `bindings::export!(KvWatcher);` at the top level

examples/rust/components/http-keyvalue-watcher/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+use crate::bindings::exports::wasi::keyvalue::watcher::Guest as KvWatcherDemoGuest;
+
+#[derive(Debug)]
+struct KvWatcher {}
+
+impl KvWatcherDemoGuest for KvWatcher {
+    fn on_set(bucket: bindings::wasi::keyvalue::store::Bucket, key: String, value: Vec<u8>) {
+        let val = String::from_utf8(value.clone()).unwrap();
```
thomastaylor312, line 17: Nit: This value doesn't need to be cloned and can just be consumed by the `from_utf8`

thomastaylor312 (changes requested):
This is looking real good! There is one thing here around the desired behavior of `SET` and the rest is mostly just polish

crates/provider-keyvalue-redis/src/lib.rs
Aditya1404Sal: What I interpreted was that the `on_set` should get triggered only when a specific `key:value` was fulfilled, because of the `value` parameter in the WIT. That was the only reason I came to that conclusion
```wit
interface watcher {
	/// A keyvalue interface that provides handle-watch operations.
	use store.{bucket};

	/// Handle the `set` event for the given bucket and key. It includes a reference to the `bucket`
	/// that can be used to interact with the store.
	on-set: func(bucket: bucket, key: string, value: list<u8>);

	/// Handle the `delete` event for the given bucket and key. It includes a reference to the
	/// `bucket` that can be used to interact with the store.
	on-delete: func(bucket: bucket, key: string);
}
```

examples/rust/components/http-keyvalue-watcher/src/wasmcloud.lock
Aditya1404Sal: I think so, I'll get it removed.

crates/provider-keyvalue-redis/src/lib.rs
Aditya1404Sal: Yep, I'm on it

crates/provider-keyvalue-redis/src/lib.rs
Aditya1404Sal: We can definitely go for that approach but `KEA` is the exact configuration that enables the watching capability and the other valid options just narrow down *what* the actual alerts are, I'm quoting the docs here:
"At least K or E should be present in the string, otherwise no event will be delivered regardless of the rest of the string."

crates/provider-keyvalue-redis/src/lib.rs
thomastaylor312: Ah ok. So yeah, watching a set just means you expect to receive the most recent value from that SET operation. You don't have to specify a value to watch for

crates/provider-keyvalue-redis/src/lib.rs
thomastaylor312: So let's check for `K` and `E` then. The `A` is a shorthand for other options, so it isn't strictly necessary, right?

crates/provider-keyvalue-redis/src/lib.rs
Aditya1404Sal: Ahh I see, that makes so much more sense
I'll refactor it to the desired usage !!

crates/provider-keyvalue-redis/src/lib.rs
Aditya1404Sal: Okay my fault here, The only 3 stictly necessary Config options are `K$g` : `K` for keyspace which the provider uses, it doesn't use E (keyevent) and the `$` is for enabling SET events and `g` is for `DEL`  events, I'll just check for these 3 instead

crates/provider-keyvalue-redis/src/lib.rs
```diff
@@ -381,6 +456,179 @@ impl Provider for KvRedisProvider {
...
+                    let event: String = match msg.get_payload() {
+                        Ok(event) => event,
+                        Err(e) => {
+                            error!(err = %e, "Failed to get payload");
+                            continue;
+                        }
+                    };
+                    // The Channel is in the format __keyspace@0__:key
+                    // While the payload is the event (ie set | del)
+                    let mkey = channel.split(':').last().unwrap();
```
thomastaylor312, line 580: This should log an error or warning if the split doesn't work rather than unwrapping to avoid a panic

thomastaylor312 (changes requested):
Everything looks good, thanks for all the hard work! There is just one `unwrap` I found that I accidentally missed on the last review. Once that is handled, this is good to merge!

Aditya1404Sal:
@thomastaylor312 -- besides the 2 flakes, I think this PR is good to go.

thomastaylor312 (approved):
All good to go! Pretty sure the test failures were fixed by @vados-cosmonic recently, so a rebase should fix them which is what the merge queue will do

Aditya1404Sal:
Thank you for your guidance throughout this PR! @brooksmtownsend and @thomastaylor312 I really appreciate your reviews and feedback. Excited to see it go through once #4166 merges!

PR #3246, opened 2024-10-08
chore(README): quicker quickstart

brooksmtownsend:
## Feature or Problem
This PR brings a delightful touch of elegance by streamlining the code, reducing the number of lines from five to three. By refining the logic and removing unnecessary complexity, this change not only enhances readability but also makes the code more efficient. The result? A cleaner, more concise implementation that's easier to maintain and a joy to work with! Less is more, and this PR proves it.


lachieh (approved):
💟❤️🫶

vados-cosmonic (approved):
LGTM :rocket:

PR #3247, opened 2024-10-08
ts release pipeline

lachieh:
- **chore(deps): update all ts dependencies**
- **ci: ts pipeline release changes**


.github/workflows/ts-release-package-lattice-client-react.yml
```diff
@@ -0,0 +1,35 @@
+name: ts-release (@wasmcloud/lattice-client-react)
```
vados-cosmonic, line 1:
```suggestion
name: ts-release (@wasmcloud/lattice-client-react)
```

I think this might be redundant -- IIRC GHA will display the job name nest to the top level workflow name, so I *think* you'd already get something like `ts-release / @wasmcloud/lattice-client-react` in most places?

vados-cosmonic (approved):

.github/workflows/ts-release-package-lattice-client-react.yml
lachieh: Yeah, we had some trouble enabling required checks potentially because of funky naming, so I worked out that the name at the top level is what shows up in the sidebar of [actions](https://github.com/wasmCloud/wasmCloud/actions) and the name for the job comes from either the key under `jobs:`, the `name` value under the job, or if utilizing reusable workflows, the imported file takes preference.

lachieh: In this instance it _should_ take the name from L11 and produce `typescript / release / @wasmcloud/lattice-client-core`

lachieh: Job shows up as `release / lattice-client-core / @wasmcloud/lattice-client-core`. A little redundant so I might adjust, but it doesn't seem to break the required builds, etc

vados-cosmonic: Yeah that's all good! Thanks for double checking this

PR #3264, opened 2024-10-09
feat(wash-cli): support overriding `wash dev` deps by interface

vados-cosmonic:

### DIFF
--- a/crates/core/src/wit.rs
+++ b/crates/core/src/wit.rs
@@ -4,7 +4,8 @@
 
 use std::collections::HashMap;
 
-use anyhow::Context;
+use anyhow::{bail, Context as _, Result};
+use semver::Version;
 use serde::ser::SerializeMap;
 use serde::{Deserialize, Serialize, Serializer};
 
@@ -19,7 +20,7 @@ pub type WitMap<T> = Vec<(String, T)>;
 pub(crate) fn serialize_wit_map<S: Serializer, T>(
     map: &WitMap<T>,
     serializer: S,
-) -> Result<S::Ok, S::Error>
+) -> std::result::Result<S::Ok, S::Error>
 where
     T: Serialize,
 {
@@ -32,7 +33,7 @@ where
 
 pub(crate) fn deserialize_wit_map<'de, D: serde::Deserializer<'de>, T>(
     deserializer: D,
-) -> Result<WitMap<T>, D::Error>
+) -> std::result::Result<WitMap<T>, D::Error>
 where
     T: Deserialize<'de>,
 {
@@ -70,7 +71,7 @@ impl CallTargetInterface {
     }
 
     /// Build a target interface from a given operation
-    pub fn from_operation(operation: impl AsRef<str>) -> anyhow::Result<Self> {
+    pub fn from_operation(operation: impl AsRef<str>) -> Result<Self> {
         let operation = operation.as_ref();
         let (wit_ns, wit_pkg, wit_iface, _) = parse_wit_meta_from_operation(operation)?;
         Ok(CallTargetInterface::from_parts((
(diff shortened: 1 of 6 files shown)


brooksmtownsend (changes requested):
Lots of changes here! @vados-cosmonic if you wouldn't mind adding a little more detail into the PR, that would help a lot with going through the changes. Looks like there's the feature in the title and a good bit of helpful refactoring

vados-cosmonic:
Hey @brooksmtownsend ah that must have been hard to review! I took out the commits relating to the refactor and will land them separately, now it should be the feature in the title!

crates/core/src/wit.rs
```diff
@@ -89,38 +90,149 @@ impl CallTargetInterface {
...
+    let (wit_iface, wit_fn) = match interface_and_func.split_once('.') {
+        Some((iface, func)) => (iface, Some(func.to_string())),
+        None => (interface_and_func, None),
+    };
+    Ok((wit_ns.into(), wit_pkg.into(), wit_iface.into(), wit_fn))
+}
+
+type WitInformationTuple = (
+    WitNamespace,
+    Vec<WitPackage>,
```
brooksmtownsend, line 124: When, here as the parser, would a WIT declaration result in multiple packages?

crates/wash-lib/src/parser/mod.rs
```diff
@@ -542,14 +547,17 @@ struct RawDevConfig {
...
+    pub manifests: Option<OneOrMore<DevManifestComponentTarget>>,
 
     /// Configuration values to be set up
     #[serde(alias = "configs")]
-    pub config: Option<Vec<DevConfigSpec>>,
+    pub config: Option<OneOrMore<DevConfigSpec>>,
 
     /// Configuration values to be set up
-    pub secrets: Option<Vec<DevSecretSpec>>,
+    pub secrets: Option<OneOrMore<DevSecretSpec>>,
```
brooksmtownsend, line 557: What is `OneOrMore` coming from here? I think I can understand that this means that there should be at least one. But instead of an Option<OneOrMore>, maybe a Vec that is empty?

crates/wash-lib/src/parser/mod.rs
```diff
@@ -603,6 +611,276 @@ impl DevManifestComponentTarget {
...
+                version,
+            }),
+        }
+    }
+}
+
+/// Facilitates *one* of a given `T` or more (primarily for serde use)
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(untagged)]
+pub enum OneOrMore<T> {
```
brooksmtownsend, line 820: I feel like this makes the understanding of the code a little harder, but I can be convinced if it really makes things simpler in this module

brooksmtownsend (commented):
Overrides code looks great, thank you for the test too!

I have some comments re: code understandability but I'm not strongly holding that opinion

crates/wash-lib/src/parser/mod.rs
vados-cosmonic: Ah so the reason this is here is to enable a flow like in the RFC where you can provide *either* a simple object, or a list of them (for multiple kinds of things). This was a sort of "stretch goal" in the for example, if you want to override `wasi:keyvalue` most of the time you really have only *one* thing you want to override it with. 

But, you *could* have more than one. With `OneOrMore`, both specifying a TOML object and a TOML list work

crates/wash-lib/src/parser/mod.rs
vados-cosmonic: Ahh see https://github.com/wasmCloud/wasmCloud/pull/3264#discussion_r1801583052 -- this is to enable `serde` (and resultingly TOML deserialization), so we can have config similar to the original feature proposal

vados-cosmonic:
Hey @brooksmtownsend sorry got to this late, but should have addressed your questions!

crates/wash-lib/src/parser/mod.rs
brooksmtownsend: Makes sense! Thanks @vados-cosmonic

brooksmtownsend (approved):

PR #3270, opened 2024-10-09
feat(wash): Adds a wit publish command

thomastaylor312:
This is almost exactly the same as `wkg publish` but exposed for convenience within wash

### DIFF
--- a/crates/wash-cli/src/wit.rs
+++ b/crates/wash-cli/src/wit.rs
@@ -7,6 +7,7 @@ use wash_lib::{
     build::monkey_patch_fetch_logging,
     cli::{CommandOutput, CommonPackageArgs},
 };
+use wasm_pkg_client::{PublishOpts, Registry};
 use wasm_pkg_core::{
     lock::LockFile,
     wit::{self, OutputType},
@@ -25,6 +26,9 @@ pub enum WitCommand {
     /// exists, it will fetch all dependencies. If a lock file exists, it will fetch any
     /// dependencies that are not in the lock file and update the lock file.
     Fetch(FetchArgs),
+    /// Publish a WIT package to a registry. This will automatically infer the package name from the
+    /// WIT package.
+    Publish(PublishArgs),
 }
 
 impl WitCommand {
@@ -32,6 +36,7 @@ impl WitCommand {
         match self {
             WitCommand::Build(args) => args.run().await,
             WitCommand::Fetch(args) => args.run().await,
+            WitCommand::Publish(args) => args.run().await,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/wash-cli/src/wit.rs
```diff
@@ -66,6 +71,19 @@ pub struct FetchArgs {
     pub common: CommonPackageArgs,
 }
 
+#[derive(Args, Debug, Clone)]
+pub struct PublishArgs {
+    /// The file to publish
+    file: PathBuf,
+
+    /// The registry domain to use. Overrides configuration file(s).
+    #[arg(long = "registry", value_name = "REGISTRY", env = "WKG_REGISTRY")]
```
joonas, line 80: Should this be `WASH_REGISTRY`?

Feels a little strange to have `WKG_` variables in `wash` imo.

crates/wash-cli/src/wit.rs
```diff
@@ -110,3 +128,21 @@ impl FetchArgs {
...
+            .client()?
+            .publish_release_file(
+                &self.file,
+                PublishOpts {
+                    registry: self.registry,
+                    ..Default::default()
+                },
+            )
+            .await?;
+        Ok(format!("Published {}@{}", package, version).into())
```
vados-cosmonic, line 146:
```suggestion
        Ok(format!("Published {}@{}", package, version).into())
```

Can we have a proper JSON payload for this `CommandOutput` ? Would help with testing it

crates/wash-cli/src/wit.rs
lachieh: Is there a need to differentiate between the registry for `wash push`? I don't think that has an env var, but maybe `WASH_WKG_REGISTRY`? Does `wkg` itself use the `WKG_REGISTRY` variable?

thomastaylor312: This is for parity sake with `wkg` so people don't have to have two different env vars to set. I did something similar with the common package opts

crates/wash-cli/src/wit.rs
thomastaylor312: I didn't feel like we needed that for something so simple. exit code 0 == published exit code != 0 is failure. This is purely a wrapper around a function that we are using from an upstream library as well, so most testing is done there (the only thing I might want to test is parsing args, but that felt like overkill for this command)

crates/wash-cli/src/wit.rs
thomastaylor312: Oh @lachieh's comment wasn't showing up when I wrote my response. He's correct

vados-cosmonic: My vote is for `--wit-registry` & `WASH_WIT_REGISTRY`

vados-cosmonic (approved):
LGTM :rocket: 

We've talked about the changes here and `WASH_WIT_REGISTRY` along with including the JSON should be great

PR #3280, opened 2024-10-10
fix(wash-lib): constantly failing tests

vados-cosmonic:
This commit fixes some tests that were failing (i.e. `make test-all`) when trying to test wash.

### DIFF
--- a/crates/wash-lib/build.rs
+++ b/crates/wash-lib/build.rs
@@ -25,6 +25,15 @@ async fn record_reachability(host: &str) {
 
 #[tokio::main]
 async fn main() {
+    // Determine whether docker sock is present locally
+    println!("cargo:rustc-check-cfg=cfg(var_run_docker_sock_is_present)");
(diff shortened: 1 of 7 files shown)


crates/wash-lib/build.rs
```diff
@@ -25,6 +25,15 @@ async fn record_reachability(host: &str) {
 
 #[tokio::main]
 async fn main() {
+    // Determine whether docker sock is present locally
+    println!("cargo:rustc-check-cfg=cfg(var_run_docker_sock_is_present)");
+    if tokio::fs::metadata("/var/run/docker.sock")
```
joonas, line 30: Please consider https://github.com/testcontainers/testcontainers-rs/pull/721

vados-cosmonic: Thanks for pointing this out! I'll use a different check so we don't get bit by that change of defaults whenever they do it.

PR #3281, opened 2024-10-10
feat(runtime): Adds support for versioned wasi logging interface

thomastaylor312:
This brings in the new versioned wasi logging dependency and adds it to the host's default exports. This also maintains support for the unversioned logging as well.

Please note that I did not update the component crate as it should still work and its current deps can't be updated from the runtime crate. I did manually test this and will have more tests in the next commit with upgrading dependencies

### DIFF
--- a/crates/runtime/src/capability.rs
+++ b/crates/runtime/src/capability.rs
@@ -56,6 +56,16 @@ mod wasmtime_bindings {
     });
 }
 
+#[allow(missing_docs)]
+mod unversioned_logging_bindings {
+    wasmtime::component::bindgen!({
+        world: "unversioned-interfaces",
+        async: true,
+        tracing: true,
+        trappable_imports: true,
+    });
+}
+
 #[allow(clippy::doc_markdown)]
 #[allow(missing_docs)]
 /// wRPC interface bindings
@@ -70,7 +80,8 @@ pub mod wrpc {
     });
 }
 
-pub use wasmtime_bindings::wasi::{blobstore, config, keyvalue, logging};
+pub use unversioned_logging_bindings::wasi::logging as unversioned_logging;
+pub use wasmtime_bindings::wasi::{blobstore, config, keyvalue, logging0_1_0_draft as logging};
 pub use wasmtime_bindings::wasmcloud::{bus, messaging, secrets};
 pub use wasmtime_bindings::Interfaces;
 pub use wasmtime_wasi_http::bindings::http;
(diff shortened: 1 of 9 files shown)


crates/runtime/wit/host.wit
```diff
@@ -6,13 +6,18 @@ world interfaces {
     import wasi:keyvalue/atomics@0.2.0-draft;
     import wasi:keyvalue/store@0.2.0-draft;
     import wasi:keyvalue/batch@0.2.0-draft;
-    import wasi:logging/logging;
+    import wasi:logging/logging@0.1.0-draft;
+
```
rvolosatovs, line 10: Could we import both here instead? (i.e. use a versioned *and* unversioned identifier)
If not, perhaps we can `include old-interfaces` instead of generating bindings twice?

thomastaylor312: I couldn't get the bindings to generate both if I included it in the same world. That was the first thing I tried

crates/runtime/src/capability.rs
```diff
@@ -56,6 +56,16 @@ mod wasmtime_bindings {
     });
 }
 
+#[allow(missing_docs)]
+mod old_logging_bindings {
```
brooksmtownsend, line 60:
```suggestion
mod unversioned_logging_bindings {
```

supernit and might not even be necessary, but I think this would clearly communicate what they are and contributors won't look at this and have to wonder what old means

crates/runtime/wit/guest.wit
```diff
@@ -1,6 +1,10 @@
 package wasmcloud:host@1.0.1;
 
 world logging {
+    export wasi:logging/logging@0.1.0-draft;
+}
+
+world old-logging {
```
brooksmtownsend, line 7: If you decide to take my suggestion, unversioned-logging here too

crates/runtime/wit/host.wit
```diff
@@ -6,13 +6,18 @@ world interfaces {
...
-    import wasi:logging/logging;
+    import wasi:logging/logging@0.1.0-draft;
+    
     import wasmcloud:bus/lattice@1.0.0;
     import wasmcloud:messaging/consumer@0.2.0;
     import wasmcloud:secrets/store@0.1.0-draft;
     import wasmcloud:secrets/reveal@0.1.0-draft;
 }
 
+world old-interfaces {
```
brooksmtownsend, line 17: same re: unversioned

brooksmtownsend (commented):
Nothing wild here, just a suggestion for clarity around naming "old"

brooksmtownsend (approved):

PR #3312, opened 2024-10-14
feat: WASI 0.2.2

rvolosatovs:
## Feature or Problem

- update `wasmcloud_component` to use WASI 0.2.2
- add support for the new `wasi:config/store`

## Related Issues

#3253 
https://github.com/WebAssembly/wasi-config/pull/14
https://github.com/WebAssembly/wasi-blobstore/pull/28


crates/component/src/lib.rs
```diff
@@ -5,18 +5,18 @@ mod bindings {
...
             "wasi:blobstore/types@0.2.0-draft": generate,
-            "wasi:clocks/monotonic-clock@0.2.1": ::wasi::clocks::monotonic_clock,
-            "wasi:config/runtime@0.2.0-draft": generate,
-            "wasi:http/outgoing-handler@0.2.1": ::wasi::http::outgoing_handler,
-            "wasi:http/types@0.2.1": ::wasi::http::types,
-            "wasi:io/error@0.2.1": ::wasi::io::error,
-            "wasi:io/poll@0.2.1": ::wasi::io::poll,
-            "wasi:io/streams@0.2.1": ::wasi::io::streams,
+            "wasi:clocks/monotonic-clock@0.2.2": ::wasi::clocks::monotonic_clock,
+            "wasi:config/store@0.2.0-draft": generate,
```
ricochet, line 9: We have cases of the older named interface being used. Can we support both for an interim period (like one month)?

rvolosatovs: We do support both, actually

rvolosatovs: I confused this block with a different location in code, no changes necessary, everything is in order here.
`wasi:config/runtime` bindings are generated separately if that's the source of confusion here. It's impossible to generate *both* `wasi:config/store` and `wasi:config/runtime` without patching in a single invocation, since there's a dependency clash in that case (2 versions of the same package)

tests/components/rust/pinger-config-component/src/lib.rs
```diff
@@ -44,8 +44,8 @@ impl exports::wasi::http::incoming_handler::Guest for Actor {
         let pong_secret = pingpong::ping_secret();
 
         let res = json!({
-            "single_val": config::runtime::get(&config_key).expect("failed to get config value"),
-            "multi_val": config::runtime::get_all().expect("failed to get config value").into_iter().collect::<HashMap<String, String>>(),
+            "single_val": config::store::get(&config_key).expect("failed to get config value"),
+            "multi_val": config::store::get_all().expect("failed to get config value").into_iter().collect::<HashMap<String, String>>(),
```
brooksmtownsend, line 48: Thoughts on retaining this one as `config::runtime` in order to confirm that backwards compat is working as expected (and we can file a deprecation issue?

brooksmtownsend (approved):
From what I understand, looks like
- wasmcloud_component now will use `wasi:config/store.*`
- wasmcloud_runtime supports `wasi:config/store|runtime.*` (for the time being)

As long as I'm right there, LGTM. Left a suggestion for the test to "verify" backwards compatibility isn't carelessly removed in a future commit but I'm open to options

rvolosatovs:
> From what I understand, looks like
> 
>     * wasmcloud_component now will use `wasi:config/store.*`
> 
>     * wasmcloud_runtime supports `wasi:config/store|runtime.*` (for the time being)
> 
> 
> As long as I'm right there, LGTM. Left a suggestion for the test to "verify" backwards compatibility isn't carelessly removed in a future commit but I'm open to options

That's exactly right, I've added a test case for the legacy interface in https://github.com/wasmCloud/wasmCloud/pull/3312/commits/f4998f61a12643d6456f3fde1cdc42ce44f6ee3f

tests/components/rust/pinger-config-component/src/lib.rs
rvolosatovs: Added a test case in https://github.com/wasmCloud/wasmCloud/pull/3312/commits/51ea896fe0227d77765dcdb742b9a772c6ecdbfd, since this component does not have any other imports other that `wasmcloud_component` crate, I'd rather keep it that way

Issue #3120, opened 2024-09-25
[RFC] Multitenancy in wasmCloud


protochron:
### RFC Title

Multitenancy in wasmCloud

### Introduction

The goal of this work is to make it possible to run a wasmCloud host that is
capable of running workloads for multiple lattices at the same time. A host
should also be able to run a workload for *any* lattice if requested, which is
a very different way of looking at lattices compared to the current model.

As part of this work, we propose the following changes:

* Add a dedicated control plane to manage lattices and in the future provide
other cluster-level APIs
* Use NATS distributed authentication in order provide a security boundary
around lattices

### Motivation

([RFC2119](https://datatracker.ietf.org/doc/html/rfc2119) definitions used here)

* A wasmCloud host *must* be able to run workloads for multiple lattices at the
same time if properly configured.
* A wasmCloud host *must* be able to automatically pull credentials for a
lattice if using distributed auth with dedicated lattice accounts.
* A single wasmCloud host *must* be capable of running workloads for all
lattices if needed.
* The current single-lattice model *must* be preserved as an option along with
being the default for backwards compatibility and for less complex deployments.
* A control plane implementation *must* exist so that a wasmCloud host can
automatically be granted credentials to access a lattice if necessary.
* A wasmCloud host *should* be configured with a set of lattice IDs it is
allowed to run workloads for. The default will be to run workloads for all
available lattices.

### What is a lattice?

The [docs provide a good
overview](https://wasmcloud.com/docs/hosts/lattice-protocols/name) of the goals
of a lattice, but at a low level a lattice is just a NATS subject prefix.
Specifically it's the value set on `wasmbus.{ctl,rpc,evt}.{lattice}`.

Right now that value is a single string that is set when the host starts.

Lattices allow users to subdivide their wasmCloud clusters by isolating groups
of related applications in their own container. This allows you to have
multiple applications with the same name but running in different lattices,
which guarantees that application names must only be unique in the scope of a
lattice. They also serve as logical partitions for groups of related
applications, team boundaries or other non-technical requirements.

Implicit in this definition is that a lattice can be both a security boundary
and logical boundary. In wasmCloud cluster that does not use distributed
authentication, a lattice is only a logical boundary, or a means of separating
applications. Due to the way that distributed authentication works in NATS, you
could also use lattices as hard security boundary by provision lattices as
*dedicated NATS account*s.

The intent of this RFC is to allow hosts to run workloads for multiple
[...]

PR #3313, opened 2024-10-14
feat(wasmcloud-host): enable preliminary support for multiple lattices on a single host

protochron:
## Feature or Problem
This implements initial support for multiple lattices running on a single host by allowing the `--lattice` flag to be specified multiple times. The host code itself is also shuffled around to support this functionality, since each lattice now needs to handle its own subscriptions, cleanup signals, etc. As a result, the host code is now much more compact than it was before.

This work is the first step towards implementing full multitenancy support in the host as outlined in #3120. The next step is to add an experimental flag to enable lattices to be defined in the control plane and have the host dynamically listen to requests to listen on lattices as they are provisioned.

## Release Information
v1.4

### DIFF
--- a/crates/host/src/metrics.rs
+++ b/crates/host/src/metrics.rs
@@ -16,16 +16,12 @@ pub struct HostMetrics {
     // but we don't really hve a way of getting at those. We should figure out a way to get at that
     // information so we don't have to duplicate it here.
     pub host_id: String,
-    /// The host's lattice ID.
-    // Eventually a host will be able to support multiple lattices, so this will need to either be
-    // removed or metrics will need to be scoped per-lattice.
-    pub lattice_id: String,
 }
 
 impl HostMetrics {
     /// Construct a new [`HostMetrics`] instance for accessing the various wasmcloud host metrics linked to the provided meter.
     #[must_use]
-    pub fn new(meter: &Meter, host_id: String, lattice_id: String) -> Self {
+    pub fn new(meter: &Meter, host_id: String) -> Self {
         let wasmcloud_host_handle_rpc_message_duration_ns = meter
             .u64_histogram("wasmcloud_host.handle_rpc_message.duration")
             .with_description("Duration in nanoseconds each handle_rpc_message operation took")
@@ -47,7 +43,6 @@ impl HostMetrics {
             component_invocations: component_invocation_count,
             component_errors: component_error_count,
             host_id,
-            lattice_id,
         }
     }
 
--- a/crates/host/src/wasmbus/host_config.rs
+++ b/crates/host/src/wasmbus/host_config.rs
@@ -34,8 +34,7 @@ pub struct Host {
     /// Whether to require TLS for RPC connection
     pub rpc_tls: bool,
     /// The lattices the host belongs to
-    /// TODO: should this actually be Vec<Arc<str>>?
-    pub lattices: Vec<String>,
+    pub lattices: Arc<[Box<str>]>,
     /// The domain to use for host Jetstream operations
     pub js_domain: Option<String>,
     /// Labels (key-value pairs) to add to the host
@@ -101,7 +100,7 @@ impl Default for Host {
             rpc_jwt: None,
             rpc_key: None,
             rpc_tls: false,
-            lattices: vec!["default".to_string()],
+            lattices: Arc::new(["default".to_string().into_boxed_str()]),
             js_domain: None,
             labels: HashMap::default(),
             host_key: None,
--- a/crates/host/src/wasmbus/host_config.rs
+++ b/crates/host/src/wasmbus/host_config.rs
@@ -33,8 +33,9 @@ pub struct Host {
     pub rpc_key: Option<Arc<KeyPair>>,
     /// Whether to require TLS for RPC connection
     pub rpc_tls: bool,
-    /// The lattice the host belongs to
-    pub lattice: Arc<str>,
+    /// The lattices the host belongs to
+    /// TODO: should this actually be Vec<Arc<str>>?
+    pub lattices: Vec<String>,
     /// The domain to use for host Jetstream operations
     pub js_domain: Option<String>,
     /// Labels (key-value pairs) to add to the host
@@ -100,7 +101,7 @@ impl Default for Host {
             rpc_jwt: None,
             rpc_key: None,
             rpc_tls: false,
-            lattice: "default".into(),
+            lattices: vec!["default".to_string()],
             js_domain: None,
             labels: HashMap::default(),
             host_key: None,
--- a/crates/host/src/wasmbus/lattice.rs
+++ b/crates/host/src/wasmbus/lattice.rs
@@ -1,0 +1,2957 @@
+use crate::wasmbus::host_config::PolicyService;
+use crate::wasmbus::{
+    component_import_links, create_bucket, event, fetch_component, handler::Handler,
+    injector_to_headers, load_supplemental_config, merge_registry_config, serialize_ctl_response,
+    Annotations, Claims, Component, ComponentSpecification, Provider, Queue, StoredClaims,
+    SupplementalConfig, WrpcServer, MAX_INVOCATION_CHANNEL_SIZE, MIN_INVOCATION_CHANNEL_SIZE,
+};
+use crate::WasmbusHostConfig;
+
+use std::collections::hash_map::{self};
+use url::Url;
+use wasmcloud_core::logging::Level as LogLevel;
+
+use std::collections::btree_map::Entry as BTreeMapEntry;
+use std::collections::{BTreeMap, HashMap};
+use std::env;
+use std::num::NonZeroUsize;
+use std::process::Stdio;
+use std::sync::Arc;
+use std::time::Duration;
+
+use anyhow::{anyhow, bail, ensure, Context as _};
+use async_nats::jetstream::kv::{Entry as KvEntry, Operation, Store};
+use base64::engine::general_purpose::STANDARD;
+use base64::Engine;
+use bytes::Bytes;
+use futures::stream::{AbortHandle, Abortable};
+use futures::{join, stream, StreamExt, TryFutureExt, TryStreamExt};
+use nkeys::{KeyPair, XKey};
+use secrecy::Secret;
+use serde_json::json;
+use tokio::io::AsyncWriteExt;
+use tokio::sync::{broadcast, mpsc, watch, RwLock, Semaphore};
+use tokio::task::{JoinHandle, JoinSet};
+use tokio::time::{Instant, Interval};
+use tokio::{process, select, spawn};
+use tokio_stream::wrappers::IntervalStream;
+use tracing::{debug, error, info, instrument, trace, warn, Instrument as _};
+use uuid::Uuid;
+use wascap::jwt;
+use wasmcloud_control_interface::{
+    ComponentAuctionAck, ComponentAuctionRequest, ComponentDescription, CtlResponse,
+    DeleteInterfaceLinkDefinitionRequest, HostLabel, Link, ProviderAuctionAck,
+    ProviderAuctionRequest, ProviderDescription, RegistryCredential, ScaleComponentCommand,
+    StartProviderCommand, StopProviderCommand, UpdateComponentCommand,
+};
+use wasmcloud_core::{
+    provider_config_update_subject, ComponentId, HealthCheckResponse, HostData, OtelConfig,
+};
+use wasmcloud_runtime::capability::secrets::store::SecretValue;
+use wasmcloud_runtime::component::WrpcServeEvent;
+use wasmcloud_runtime::Runtime;
+use wasmcloud_secrets_types::SECRET_PREFIX;
+use wasmcloud_tracing::context::TraceContextInjector;
+use wasmcloud_tracing::KeyValue;
+
+use crate::registry::RegistryCredentialExt;
+use crate::{
+    HostMetrics, OciConfig, PolicyHostInfo, PolicyManager, PolicyResponse, RegistryConfig,
+    SecretsManager,
+};
+
+use crate::wasmbus::config::{BundleGenerator, ConfigBundle};
+
+/// wasmCloud Host configuration
+#[allow(clippy::struct_excessive_bools)]
+#[derive(Clone, Debug)]
+pub struct LatticeConfig {
+    /// The topic prefix to use for control interface subscriptions, defaults to `wasmbus.ctl`
+    pub ctl_topic_prefix: String,
+    /// NATS URL to connect to for component RPC
+    pub rpc_nats_url: Url,
+    /// Timeout period for all RPC calls
+    pub rpc_timeout: Duration,
+    /// Authentication JWT for RPC connection, must be specified with `rpc_seed`
+    pub rpc_jwt: Option<String>,
+    /// Authentication key pair for RPC connection, must be specified with `rpc_jwt`
+    pub rpc_key: Option<Arc<KeyPair>>,
+    /// The amount of time to wait for a provider to gracefully shut down before terminating it
+    pub provider_shutdown_delay: Option<Duration>,
+    /// Configuration for downloading artifacts from OCI registries
+    pub oci_opts: OciConfig,
+    /// Whether to allow loading component or provider components from the filesystem
+    pub allow_file_load: bool,
+    /// Whether or not structured logging is enabled
+    pub enable_structured_logging: bool,
+    /// Log level to pass to capability providers to use. Should be parsed from a [`tracing::Level`]
+    pub log_level: LogLevel,
+    /// Whether to enable loading supplemental configuration
+    pub config_service_enabled: bool,
+    /// configuration for OpenTelemetry tracing
+    pub otel_config: OtelConfig,
+    /// configuration for wasmCloud policy service
+    pub policy_service_config: PolicyService,
+    /// topic for wasmCloud secrets backend
+    pub secrets_topic_prefix: Option<String>,
+}
+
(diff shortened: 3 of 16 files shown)


crates/host/src/wasmbus/lattice.rs
```diff
@@ -0,0 +1,2957 @@
...
+use wasmcloud_tracing::context::TraceContextInjector;
+use wasmcloud_tracing::KeyValue;
+
+use crate::registry::RegistryCredentialExt;
+use crate::{
+    HostMetrics, OciConfig, PolicyHostInfo, PolicyManager, PolicyResponse, RegistryConfig,
+    SecretsManager,
+};
+
+use crate::wasmbus::config::{BundleGenerator, ConfigBundle};
```
rvolosatovs, line 63:
```suggestion
use std::collections::hash_map;
use std::collections::btree_map::Entry as BTreeMapEntry;
use std::collections::{BTreeMap, HashMap};
use std::env;
use std::num::NonZeroUsize;
use std::process::Stdio;
use std::sync::Arc;
use std::time::Duration;

use anyhow::{anyhow, bail, ensure, Context as _};
use async_nats::jetstream::kv::{Entry as KvEntry, Operation, Store};
use base64::engine::general_purpose::STANDARD;
use base64::Engine;
use bytes::Bytes;
use futures::stream::{AbortHandle, Abortable};
use futures::{join, stream, StreamExt, TryFutureExt, TryStreamExt};
use nkeys::{KeyPair, XKey};
use secrecy::Secret;
use serde_json::json;
use tokio::io::AsyncWriteExt;
use tokio::sync::{broadcast, mpsc, watch, RwLock, Semaphore};
use tokio::task::{JoinHandle, JoinSet};
use tokio::time::{Instant, Interval};
use tokio::{process, select, spawn};
use tokio_stream::wrappers::IntervalStream;
use tracing::{debug, error, info, instrument, trace, warn, Instrument as _};
use url::Url;
use uuid::Uuid;
use wascap::jwt;
use wasmcloud_core::logging::Level as LogLevel;

use crate::registry::RegistryCredentialExt;
use crate::{
    HostMetrics, OciConfig, PolicyHostInfo, PolicyManager, PolicyResponse, RegistryConfig,
    SecretsManager,
};
use crate::wasmbus::host_config::PolicyService;
use crate::wasmbus::{
    component_import_links, create_bucket, event, fetch_component, handler::Handler,
    injector_to_headers, load_supplemental_config, merge_registry_config, serialize_ctl_response,
    Annotations, Claims, Component, ComponentSpecification, Provider, Queue, StoredClaims,
    SupplementalConfig, WrpcServer, MAX_INVOCATION_CHANNEL_SIZE, MIN_INVOCATION_CHANNEL_SIZE,
};
use crate::WasmbusHostConfig;
use crate::wasmbus::config::{BundleGenerator, ConfigBundle};
```

crates/test-util/src/host.rs
```diff
@@ -82,21 +82,21 @@ impl WasmCloudTestHost {
     /// * `secrets_backend_topic` - Topic for the host to use for secrets requests
     pub async fn start_custom(
         nats_url: impl AsRef<str>,
-        lattice_name: impl AsRef<str>,
+        lattice_names: Vec<String>,
```
rvolosatovs, line 85:
```suggestion
        lattice_names: impl Into<Arc<[Box<str>]>>,
```

crates/test-util/src/host.rs
```diff
@@ -82,21 +82,21 @@ impl WasmCloudTestHost {
...
-        let lattice_name = lattice_name.as_ref();
+        //let lattice_names = lattice_name.as_ref();
         let cluster_key = Arc::new(cluster_key.unwrap_or(KeyPair::new_cluster()));
         let host_key = Arc::new(host_key.unwrap_or(KeyPair::new_server()));
 
         let mut host_config = HostConfig {
             ctl_nats_url: nats_url.clone(),
             rpc_nats_url: nats_url.clone(),
-            lattice: lattice_name.into(),
+            lattices: lattice_names.clone(),
```
rvolosatovs, line 99:
```suggestion
            lattices: Arc::clone(&lattice_names),
```

crates/test-util/src/host.rs
```diff
@@ -82,21 +82,21 @@ impl WasmCloudTestHost {
...
-        lattice_name: impl AsRef<str>,
+        lattice_names: Vec<String>,
         cluster_key: Option<KeyPair>,
         host_key: Option<KeyPair>,
         policy_service_config: Option<PolicyService>,
         secrets_topic_prefix: Option<String>,
     ) -> Result<Self> {
         let nats_url = Url::try_from(nats_url.as_ref()).context("failed to parse NATS URL")?;
-        let lattice_name = lattice_name.as_ref();
+        //let lattice_names = lattice_name.as_ref();
```
rvolosatovs, line 92:
```suggestion
        let lattice_names = lattice_names.into();
```

crates/host/src/wasmbus/mod.rs
```diff
@@ -938,121 +895,50 @@ impl Host {
...
-                    Err(err) => error!(%err, "failed to read entry from lattice data bucket"),
-                }
-            })
-            .await;
-
-        host.publish_event("host_started", start_evt)
-            .await
-            .context("failed to publish start event")?;
+        let lattice_names = host.lattices.keys().cloned().collect::<Vec<_>>();
+        for lattice in lattice_names.clone() {
```
rvolosatovs, line 916:
```suggestion
        for lattice in &lattice_names {
```

rvolosatovs (commented):
Overall, I think this is looking pretty good - a bit hard to review though, since it's a big changeset and many things are moved around.
Left a few minor nits

crates/host/src/wasmbus/mod.rs
protochron: This is cloned because it's used further down in the `async move` block.

protochron:
@rvolosatovs addressed your feedback

crates/host/src/metrics.rs
```diff
@@ -16,16 +16,12 @@ pub struct HostMetrics {
     // but we don't really hve a way of getting at those. We should figure out a way to get at that
     // information so we don't have to duplicate it here.
     pub host_id: String,
-    /// The host's lattice ID.
-    // Eventually a host will be able to support multiple lattices, so this will need to either be
-    // removed or metrics will need to be scoped per-lattice.
-    pub lattice_id: String,
```
brooksmtownsend, line 22: Leaving comment but will remove if I find it, should we still annotate metrics with the lattice?

src/main.rs
```diff
@@ -67,10 +67,11 @@ struct Args {
     #[clap(
         short = 'x',
         long = "lattice",
-        default_value = "default",
+        default_values_t = ["default".to_string()],
+        value_delimiter = ',',
         env = "WASMCLOUD_LATTICE"
     )]
-    lattice: String,
+    lattice: Vec<String>,
```
brooksmtownsend, line 74:
```suggestion
    lattices: HashSet<String>,
```

supernit, but just to be clear in the code without changing clap

brooksmtownsend: Also, I think we can do this too to prevent duplicates

crates/host/src/wasmbus/mod.rs
```diff
@@ -127,22 +113,49 @@ impl TryFrom<AsyncBytesMut> for Vec<u8> {
     }
 }
 
+// TODO enum?!
```
brooksmtownsend, line 116: nit: can remove unused comment

crates/host/src/wasmbus/mod.rs
```diff
@@ -349,42 +355,29 @@ impl Drop for Provider {
...
 /// wasmCloud Host
 pub struct Host {
-    components: RwLock<HashMap<ComponentId, Arc<Component>>>,
     event_builder: EventBuilderV10,
     friendly_name: String,
-    heartbeat: AbortHandle,
+    heartbeats: Vec<AbortHandle>,
     host_config: HostConfig,
     host_key: Arc<KeyPair>,
+    #[allow(unused)]
```
brooksmtownsend, line 363: is this leftover from avoiding clippy or do we still need the 3 `#[allow(unused)]` macros in the Host struct

crates/host/src/wasmbus/mod.rs
```diff
@@ -938,121 +895,50 @@ impl Host {
             }
         });
 
-        let data_watch: JoinHandle<anyhow::Result<_>> = spawn({
-            let data = data.clone();
+        spawn({
```
brooksmtownsend, line 898: Do we need to capture this task? Looks like it's handling publishing all the events?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1094,89 +980,13 @@ impl Host {
...
-                        )
-                        .build()
-                        .expect("failed to build provider description")
-                },
-            )
-            .collect();
+    async fn inventory(&self, lattice: &str) -> anyhow::Result<HostInventory> {
+        trace!(lattice = lattice, "generating host inventory");
 
+        let lattice = self.lattices.get(lattice).context("lattice not found")?;
```
brooksmtownsend, line 986: Since this is the error a user would see (even though it probably shouldn't be subscribing here)
```suggestion
        let lattice = self.lattices.get(lattice).with_context(|| format!("host is not in lattice {lattice}, inventory is empty"))?;
```

crates/host/src/wasmbus/host_config.rs
```diff
@@ -33,8 +33,8 @@ pub struct Host {
     pub rpc_key: Option<Arc<KeyPair>>,
     /// Whether to require TLS for RPC connection
     pub rpc_tls: bool,
-    /// The lattice the host belongs to
-    pub lattice: Arc<str>,
+    /// The lattices the host belongs to
+    pub lattices: Arc<[Box<str>]>,
```
brooksmtownsend, line 37: I know slices/vecs are a lot easier to work with, but since this is the "source of truth" for what lattices a host belongs to, it might be a good idea to make this a HashSet<> to avoid duplicates since this is a public field & can be constructed directly.

Changing here would let us use a `Vec` wherever we wanted in the code (or a `Arc<[Box<str>]>` if that's the type we keep around in memory) but know the values are unique and avoid setting up duplicate subscriptions

crates/host/src/wasmbus/lattice.rs
```diff
@@ -0,0 +1,2954 @@
...
+        let config_generator = BundleGenerator::new(config_data.clone());
+
+        let (data_watch_abort, data_watch_abort_reg) = AbortHandle::new_pair();
+        let (queue_abort, queue_abort_reg) = AbortHandle::new_pair();
+        let (heartbeat_abort, heartbeat_abort_reg) = AbortHandle::new_pair();
+        let lattice = Self {
+            name: Arc::from(name.clone()),
+            labels,
+            // TODO this is in theory redundant with some other config options
+            ctl_topic_prefix: config.ctl_topic_prefix.clone(),
```
brooksmtownsend, line 213: Yeah do you even need to know this topic prefix in this struct, or are the connections enough?

crates/host/src/wasmbus/lattice.rs
```diff
@@ -0,0 +1,2954 @@
...
+                if let Err(err) = lattice.shutdown().await {
+                    error!(lattice =% name, "failed to shutdown lattice: {err}");
+                };
+                info!(lattice =% name, "lattice stopped");
+            }
+        });
+
+        // TODO this feels like a host-level responsibility and not a lattice thing. The problem of
+        // course is that we need to fix the control interface so that a host heartbeat contains
+        // information from all lattices.
```
brooksmtownsend, line 333:
```suggestion
        // TODO(#3120) this feels like a host-level responsibility and not a lattice thing. The problem of
        // course is that we need to fix the control interface so that a host heartbeat contains
        // information from all lattices.
```

crates/host/src/wasmbus/lattice.rs
```diff
@@ -0,0 +1,2954 @@
...
+        Ok(Arc::clone(&lattice))
+    }
+
+    async fn shutdown(&self) -> anyhow::Result<()> {
+        self.data_watch_abort.abort();
+        self.queue_abort.abort();
+        self.heartbeat_abort.abort();
+        self.policy_manager.policy_changes.abort();
+        // TODO add this back in
+        //let _ = try_join!(queue, data_watch).context("failed to await tasks")?;
```
brooksmtownsend, line 400: 😄

crates/host/src/wasmbus/lattice.rs
```diff
@@ -0,0 +1,2954 @@
...
+
+        self.rpc_nats
+            .flush()
+            .await
+            .context("failed to flush NATS connection")?;
+        // NOTE: Epoch interrupt thread will only stop once there are no more references to the engine
```
brooksmtownsend, line 406: I think this comment can be removed, and/or reconciled with #3330 which doesn't wait for the thread to close and rather closes the thread itself

crates/host/src/wasmbus/lattice.rs
```diff
@@ -0,0 +1,2954 @@
...
+    /// Returns an error if internal stop channel is closed prematurely
+    //#[instrument(level = "debug", skip_all)]
+    //pub async fn stopped(&self) -> anyhow::Result<Option<Instant>> {
+    //    self.stop_rx
+    //        .clone()
+    //        .changed()
+    //        .await
+    //        .context("failed to wait for stop")?;
+    //    Ok(*self.stop_rx.borrow())
+    //}
```
brooksmtownsend, line 424: This is in `mod.rs` now right

crates/host/src/wasmbus/lattice.rs
```diff
@@ -0,0 +1,2954 @@
...
+        claims: jwt::Claims<jwt::Component>,
+    ) -> anyhow::Result<()> {
+        let mut component_claims = self.component_claims.write().await;
+        component_claims.insert(claims.subject.clone(), claims);
+        Ok(())
+    }
+
+    //#[instrument(level = "debug", skip_all)]
+    //async fn handle_stop_host(
+    //    &self,
```
brooksmtownsend, line 839: same comment re: this is present in `mod.rs` and doesn't need to be here

tests/host.rs
```diff
@@ -0,0 +1,150 @@
...
+use common::free_port;
+use common::nats::start_nats;
+use common::providers;
+
+const LATTICES: &[&str] = &["default", "test-lattice"];
+const COMPONENT_ID: &str = "http_hello_world";
+
+#[tokio::test]
+async fn host_multiple_lattices() -> anyhow::Result<()> {
+    tracing_subscriber::registry()
```
brooksmtownsend, line 31: In the interest of sanity, I would love to see some more commands tested here for the host like labels being put/deleted, handle retrieving links,claims, and maybe even a subscription on `wasmbus.evt.*.host_heartbeat` to confirm that you get a heartbeat for both lattices.

I think it would also be nice to have some failure cases captured here, or potentially in a different test. Are there any opportunities that you can see that may cause unexpected behavior, for example:
1. Make sure a host doesn't respond to a message for a lattice it doesn't belong to (I think all subjects include the lattice, so this might just be not a useful test)
2. What if I try and make my lattice name `*.>`?
3. What if I create config in lattice `default` and then start a component in `test-lattice` that uses that config, that should fail
4. ...?

What I'm really looking for, with the large amount of changes here, is to be sure that existing functionality is preserved and that the multiple lattices feature doesn't allow for over-reaching that boundary. I don't see anything that would cause that, but if we capture some of the valid and invalid cases here we can move forward with that confidence

brooksmtownsend (changes requested):
80% nits & comments left over, the only real concerns I had were around unique lattice names and buffing up the test suite

crates/host/src/metrics.rs
brooksmtownsend: Yes we should, and you do! Found it 😄

src/main.rs
protochron: Clap unfortunately won't let you do that https://github.com/clap-rs/clap/discussions/4291 without using a custom value parser. I _could_ go that direction, but it ends up being the same code regardless to convert from a list of strings into a HashSet

crates/host/src/wasmbus/mod.rs
protochron: It's for a follow-up PR where we will want these values again when we have hosts that need to listen for traffic for multiple lattices. I didn't want to remove them since they're just going to come back

crates/host/src/wasmbus/mod.rs
protochron: It does, but it's set up to terminate itself when the host shuts down. I can store the join handle and ensure that it's stopped before exiting just to ensure that it's in fact stopped.

crates/host/src/wasmbus/lattice.rs
protochron: It is, yes

src/main.rs
brooksmtownsend: Oh that's fine then, all good. I don't think a CLI converting a Vec into a hashset and back is going to be the end of the world

crates/host/src/wasmbus/lattice.rs
protochron: I rebased on #3330, but it wouldn't have picked up the comment. I'll remove it.

tests/host.rs
protochron: I added a number of tests which cover most everything you pointed out, along with a scale test that shows we can run at least 2000 lattices on a host at the same time (admittedly without any workloads).

Re: `*.>`, that actually would have failed outside of any of my changes because the lattice data buckets would fail to create with any of those characters. I added an explicit check for them so that we get a sensible error though.

PR #3315, opened 2024-10-14
chore(ci): Makes it so we don't run wasmcloud host builds against docs

thomastaylor312:
We were always running builds on PRs even if we were just changing the top level markdowns.


ricochet (approved):

rvolosatovs:
This will prevent docs-only PRs from getting merged due to required checks, right?

brooksmtownsend:
Yes this will

Issue #3316, opened 2024-10-14
[BUG] host shutdown with a Ctrl-C exits with a warning


protochron:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

Stopping the host with Ctrl-C successfully kills the host, but it ends with a warning that the wasmtime epoch interrupt thread is timing out:
```
Error: failed to shutdown host

Caused by:
    epoch interrupt thread timed out
```

### Steps to reproduce

1. Start a wasmcloud host, wadm, nats
2. Start a wadm manifest (I've been using the rust http hello app)
3. Ctrl-C in a terminal to stop the host
4. Observe the error

### Expected behavior

I expected the host to stop without that warning.

If it's a bug in the way we're stopping wasmtime the we should fix it, otherwise we should downgrade the warning if there is nothing for us to do.

### Environment

- OS: OSX
- Shell zsh
- wasmcloud Version: v1.3.0

rvolosatovs:
It's actually a duplicate of https://github.com/wasmCloud/wasmCloud/issues/2391, which was closed a few months ago. Not sure if it's a regression though, I think this is simply a race condition it shutdown sequence.

I also encounter this one quite often

brooksmtownsend:
Yep I think this isn't a regression as much as it just wasn't truly resolved 😅

brooksmtownsend:
Digging in, this is almost certainly due to any running component taking a reference to the underlying Runtime Engine and the epoch thread not exiting until every reference is dropped.

PR #3330, opened 2024-10-15
fix(host): end epoch interruption, don't await

brooksmtownsend:
## Feature or Problem
This PR fixes the issue where we still had remaining references to the wasmtime Engine when the host was stopping, so the epoch interruption thread didn't complete.

This may or may not be the "correct" way to solve this problem. I approached this from the perspective that
1. This error is misleading when the host is shutting down
1. We want to make sure we finish executing the thread that's running the epoch interrupter, whether or not we've properly removed all references to the underlying engine.

## Related Issues
Fixes https://github.com/wasmCloud/wasmCloud/issues/3316

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -29,7 +29,7 @@ use serde_json::json;
 use tokio::io::{AsyncWrite, AsyncWriteExt};
 use tokio::sync::{broadcast, mpsc, watch, RwLock, Semaphore};
 use tokio::task::{JoinHandle, JoinSet};
-use tokio::time::{interval_at, timeout_at, Instant};
+use tokio::time::{interval_at, Instant};
 use tokio::{process, select, spawn};
 use tokio_stream::wrappers::IntervalStream;
 use tracing::{debug, error, info, instrument, trace, warn, Instrument as _};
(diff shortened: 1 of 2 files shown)


brooksmtownsend:
Hold off on merging until #3313 goes in just to keep merge conflicts simple

rvolosatovs (changes requested):
I don't think this is the approach we should take here. Ultimately, this disables the timer and allows all components to run forever.
Since we are in the shutdown sequence anyway, why don't we simply *not* abort the interrupt thread and let it shutdown itself? So why don't we simply remove the `epoch_end` await from original implementation?

This would avoid a memory leak if the host was initialized/stopped multiple times (or there are a few instances of it)

I do still think it's concerning that we're just aborting component execution here instead of a graceful shutdown, but I suppose we can address that later (which may break logical assumptions in components, like Go `defer`)

brooksmtownsend:
> So why don't we simply remove the epoch_end await from original implementation?

I think this makes the most sense, it's shutting down anyways so we can just remove it

> I do still think it's concerning that we're just aborting component execution here instead of a graceful shutdown, but I suppose we can address that later (which may break logical assumptions in components, like Go defer)

Yeah I think that also makes sense, during shutdown we basically need to drain the RPC messages and handle the existing ones for a graceful shutdown. Feels like a good followup issue since we really haven't needed it yet

rvolosatovs (approved):

Issue #3287, opened 2024-10-11
[BUG] wasi:logging is not compatible with wit-deps and docs missed info about wit-deps


deem0n:
### Affected project(s)

- [X] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [X] other / not sure

### Describe the bug

All examples and docs contain wit examples as follows:

wit/world.wit
```
world example {
  import wasi:logging/logging;
}
```

wit/deps.toml
```
logging = "https://github.com/WebAssembly/wasi-logging/archive/main.tar.gz"
```

wit/deps/logging/world.wit
```
package wasi:logging;

world imports {
    import logging;
}
```

That should work with `wash build` (and wokls fine in wasmcloud 1.3, i.e. you will see logs from your component in wasmcloud log output)

### Steps to reproduce

If you want to add `wasi:messaging` to your existing project or will run `wit-deps update` for other reason, you will get broken logs behavior. That is becouse `wit-deps-cli 0.4.0` will include logging package version: 

wit/deps/logging/world.wit
```
package wasi:logging@0.1.0-draft;

world imports {
    import logging;
}
```

To be able to compile you probably will amend wit/world.wit

```
world example {
  import wasi:logging/logging@0.1.0-draft;
}
```

If you compile now with `wash build` you will get:

- wasmcloud 1.2: hangs or no logs from your component.
- wasmcloud 1.3: errors in main log like this:

```
2024-10-10T16:14:33.497571Z  WARN wasmcloud_host::wasmbus: failed to handle invocation err=failed to serve `wrpc:http/incoming-handler@0.1.0.handle` invocation

Caused by:
    0: failed to call `wasi:http/incoming-handler.handle`
    1: error while executing at wasm backtrace:
           0: 0x2c263 - wit-component:shim!indirect-wasi:logging/logging@0.1.0-draft-log
           1: 0x45a8 - imei_injector.wasm!wasi:http/incoming-handler@0.2.0#handle
    2: failed to invoke `wasi:logging/logging@0.1.0-draft.log` polyfill via wRPC
    3: failed to call `log` in instance `wasi:logging/logging@0.1.0-draft` (failed to find a configured link with name `default` from component `imei_injector-injector`, please check your configuration)
```

### Expected behavior

[...]

PR #3334, opened 2024-10-15
fix(examples): pin interface versions

rvolosatovs:
## Feature or Problem

Turned out that quite a few examples were using unpinned interface versions, some were even simply broken because of that (e.g. vendoring WASI 0.2.1, but attempting to import WASI 0.2.0 in the interface). Some dependency references simply did not exist.

All interfaces are now pinned and such incompatibilities are fixed

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/3287


brooksmtownsend (approved):
Thank you for this LGTM

PR #3335, opened 2024-10-15
ci(metadata): allow quick metadata changes

brooksmtownsend:
## Related Issues
attempting to address the issue that https://github.com/wasmCloud/wasmCloud/pull/3315 brings in where docs can't be modified

### DIFF
--- a/.github/workflows/wasmcloud-metadata.yml
+++ b/.github/workflows/wasmcloud-metadata.yml
@@ -1,0 +1,30 @@
+name: wasmcloud
+
+on:
+  pull_request:
+    paths-ignore:
+      - '*.lock'
+      - '*.nix'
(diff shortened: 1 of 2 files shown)


.github/workflows/wasmcloud-metadata.yml
```diff
@@ -0,0 +1,30 @@
+name: wasmcloud
```
lachieh, line 1:
```suggestion
name: wasmcloud-metadata
```

brooksmtownsend: Yeah I was hoping that with inverse path logic that one running would cover the other requirement, but it looks like the required job isn't fulfilled still :/

PR #3336, opened 2024-10-15
feat(wash)!: allow specifying oci push/pull concurrency

brooksmtownsend:
## Feature or Problem
This PR adds an optional flag `--concurrency` that allows for controlling the maximum concurrency for pulling and pushing chunks when using `wash pull/push`.

Draft for testing.

### DIFF
--- a/crates/wash-cli/src/common/registry_cmd.rs
+++ b/crates/wash-cli/src/common/registry_cmd.rs
@@ -45,6 +45,7 @@ pub async fn registry_pull(
             allow_latest: cmd.allow_latest,
             user: credentials.username().map(String::from),
             password: credentials.password().map(String::from),
+            concurrency: cmd.concurrency,
             insecure: cmd.opts.insecure,
             insecure_skip_tls_verify: cmd.opts.insecure_skip_tls_verify,
         },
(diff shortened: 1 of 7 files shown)


vados-cosmonic:
Hey what do you think about making this a non-breaking change? 

Also, completely separately, what do you think about using about `NonZeroUsize` ? Is it worth the trouble to prevent zero as a possibility?

brooksmtownsend:
> What do you think about making this a non-breaking change?

I _think_ this is a patch release for wash-cli, but because it changes some public members it has to be a minor for wash-lib 😞 If there's a way around it, that would work! I did set the `--concurrency` default values to what they are defaulted to in oci-distribution

Good idea on the NonZeroUsize

vados-cosmonic:
> I think this is a patch release for wash-cli, but because it changes some public members it has to be a minor for wash-lib 😞 If there's a way around it, that would work! I did set the --concurrency default values to what they are defaulted to in oci-distribution

Ah I though this was somehow a breaking change for wash cli -- yeah `wash-lib` I'm pretty sure `wash-lib` is only heavily used by `wash-cli`, don't feel *too* bad about reving it

PR #3337, opened 2024-10-15
feat: Updates tests and examples to support the new wkg deps

thomastaylor312:
This updates all dependencies to have a wkg.lock and to have a line in their git ignore. The deps are still committed in tree for backwards compatibility and they all use the new versioned logging. This looks really chunky but is mostly dep updates/deletes


examples/golang/components/sqldb-postgres-query/.gitignore
```diff
@@ -2,4 +2,4 @@
 Cargo.lock
 
 # Wash build artifacts
-build
\ No newline at end of file
+buildwit/deps/
```
lxfontes, line 5: `buildwit` ?

lxfontes (approved):

vados-cosmonic (approved):
Looks great to me, other than the great find that @lxfontes made.

I'm wondering, can we split out the replacements of `ocI-distribution` with `oci-client` into a specific commit? Has `oci-distribution` essentially been replaced by `oci-client`? Is the plan to deprecate/yank/delete the crate long-term?

examples/golang/components/sqldb-postgres-query/.gitignore
thomastaylor312: Lol this is what happens when you script it and the .gitignore doesn't have a newline. I'll fix

thomastaylor312:
oci-distribution is deprecated, but we were giving it another month before doing a rustsec security advisory. The changes for it was literally like 2-4 lines of code not counting the `s/oci_distribution/oci_client`. We needed it to update to the latest wkg stuff

thomastaylor312:
Please note that I decided not to add wit/deps to the gitignore here because it was causing issues even when force adding. We can add that down the line when we remove the deps dirs in tree

brooksmtownsend (approved):

PR #3344, opened 2024-10-16
fix(keyvalue-nats): README link keys lowercase

brooksmtownsend:
## Feature or Problem
Fixes the keyvalue-nats provider noting that configuration values are uppercase, when in fact the code expects lowercase and does not make them case insensitive.


lachieh (approved):

ricochet:
I don't want to block minor doc fix on this but this provider needs some additional TLC. Filed #3346 and #3347 for follow-up.

protochron (approved):

vados-cosmonic (approved):

PR #3350, opened 2024-10-16
fix(host): warn if config sender is dropped

brooksmtownsend:
## Feature or Problem
Small fix to be less aggressive with an error that usually appears during the shutdown of a host


vados-cosmonic (approved):
LGTM :rocket:

PR #3351, opened 2024-10-16
chore(wash-cli): NATS doesn't listen on http

joonas:
## Feature or Problem

Noticed this while @liamwh was demoing

### DIFF
--- a/crates/wash-cli/src/cmd/up.rs
+++ b/crates/wash-cli/src/cmd/up.rs
@@ -516,7 +516,7 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
             out_json.insert("success".to_string(), json!(true));
             let _ = write!(
                 out_text,
-                "\n🕸  NATS is running in the background at http://{nats_listen_address}"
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM :rocket:

crates/wash-cli/src/cmd/up.rs
```diff
@@ -516,7 +516,7 @@ pub async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result<Comman
             out_json.insert("success".to_string(), json!(true));
             let _ = write!(
                 out_text,
-                "\n🕸  NATS is running in the background at http://{nats_listen_address}"
+                "\n🕸  NATS is running in the background at {nats_listen_address}"
```
brooksmtownsend, line 519: Should we prefix with `nats://` instead?

brooksmtownsend (approved):

crates/wash-cli/src/cmd/up.rs
joonas: I couldn't think of a compelling reason to do so to be honest.

PR #3363, opened 2024-10-17
chore: Update governance with leave policy

thomastaylor312:
This PR adds a new section to the wasmCloud project governance on how project and org maintainers can take a leave of absence. As this is a governance change, we should have a super-majority of org maintainers approving this change before merging.

Other votes and comments are welcome, but are non-binding


brooksmtownsend (approved):
Very thoughtful

LiamRandall:
Lgtm!

PR #3368, opened 2024-10-18
fix(washboard-ui): disable eslint rules-of-hooks for playwright tes…

markkovari:
…tcontext

Dependabot was kind enough to update the dependency with a major bump, although went a bit overboard what to show as a linter error.
In the file below there is nothing react (functional) component or hooks related.

Sorry for the ping, just raising visibility: @lachieh


lachieh (approved):

lachieh:
Excellent, thank you!

PR #3298, opened 2024-10-12
feat(wash): allow overriding project, wit, build folders

brooksmtownsend:
## Feature or Problem
Decent wash-lib refactor which should simplify understanding the code and give us a nice `WasmcloudDotToml` struct to point at in documentation.

This PR:
- [x] Refactors the parsing logic to remove all the "Raw" structs
- [x] Removes the unused `claims`, `push_insecure`, and `call_alias` fields under component
- [x] Adds a top level wasmcloud.toml field `wit` to allow for overriding the WIT directory
- [x] Adds a top level wasmcloud.toml field `build` to allow for overriding the build directory
- [x] Actually uses the top level wasmcloud.toml field `path` to run commands & support putting the wasmcloud.toml in a different directory than the project itself
- [x] Actually uses the Rust config `debug` flag for component building if supplied
- [x] Adds the `push_insecure` field to the registry config, and another flag that fixes an issue where the OCI config would try to parse as wasmcloud.toml

## Related Issues
Fixes #2757 

## Release Information
next minor wash-cli
next minor wash-lib

## Consumer Impact
None of these changes for overriding the build/wit directories (which is the main point of this pr) will break existing manifests as the default behavior is the same.

Existing manifests that had the `claims`, `push_insecure`, or `call_alias` fields will parse fine as well, as we'll just ignore them.

### Unit Test(s)
Added a unit test to ensure this works as expected

### Acceptance or Integration
Added an integration test to ensure this works as expected. This test will need to change with #3132 because it sed-style replaces the WIT directory in the code.

PR #3371, opened 2024-10-18
feat(wash): add support for Rust wasm32-wasip2 target

brooksmtownsend:
## Feature or Problem
Now that `wasm32-wasip2` is a [tier 2 target in rust](https://doc.rust-lang.org/nightly/releases.html#compiler), we can support it natively in `wash build`! So, now, when a Rust component has a wasm_target of wasm32-wasip2, we'll use the native target instead of adapting the core module/p1 module.

I considered making `wasm32-wasip2` the default here instead of the unknown module, but that would be a significant breaking change for the wasmcloud.tomls that don't specify a target. I'll file a followup issue instead to make this a default after the next release of wash.

## Release Information
wash 0.36

### DIFF
--- a/crates/wash-lib/src/build/component.rs
+++ b/crates/wash-lib/src/build/component.rs
@@ -52,10 +52,10 @@ pub async fn build_component(
                 let rust_wasm_path =
                     build_rust_component(common_config, rust_config, component_config).await?;
                 match component_config.wasm_target {
-                    WasmTarget::CoreModule | WasmTarget::WasiP1 => rust_wasm_path,
-                    WasmTarget::WasiP2 => {
+                    WasmTarget::CoreModule | WasmTarget::WasiP1 => {
                         adapt_component_to_wasip2(&rust_wasm_path, component_config)?
                     }
+                    WasmTarget::WasiP2 => rust_wasm_path,
                 }
             }
             LanguageConfig::TinyGo(tinygo_config) => {
--- a/crates/wash-lib/src/parser/mod.rs
+++ b/crates/wash-lib/src/parser/mod.rs
@@ -12,6 +12,7 @@ use cargo_toml::{Manifest, Product};
 use config::Config;
 use semver::{Version, VersionReq};
 use serde::{Deserialize, Deserializer};
+use tracing::warn;
 use wadm_types::{Component, Properties, SecretSourceProperty};
 use wasm_pkg_core::config::Config as PackageConfig;
 use wasmcloud_control_interface::RegistryCredential;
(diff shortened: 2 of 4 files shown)


brooksmtownsend:
@rvolosatovs looks like https://github.com/nix-community/fenix#fenix might not have support for wasm32-wasip2 yet, or at least just isn't quite on 1.82. I ran a `nix flake update` locally and gave this a try, to no avail:
```
diff --git a/flake.nix b/flake.nix
index 6d495e83..48d3fa04 100644
--- a/flake.nix
+++ b/flake.nix
@@ -97,6 +97,7 @@
         targets.s390x-unknown-linux-gnu = false;
         targets.wasm32-unknown-unknown = false;
         targets.wasm32-wasip1 = false;
+        targets.wasm32-wasip2 = false;
 
         build.packages = [
           "wash-cli"

```

rvolosatovs:
> @rvolosatovs looks like https://github.com/nix-community/fenix#fenix might not have support for wasm32-wasip2 yet, or at least just isn't quite on 1.82. I ran a `nix flake update` locally and gave this a try, to no avail:
> 
> ```
> diff --git a/flake.nix b/flake.nix
> index 6d495e83..48d3fa04 100644
> --- a/flake.nix
> +++ b/flake.nix
> @@ -97,6 +97,7 @@
>          targets.s390x-unknown-linux-gnu = false;
>          targets.wasm32-unknown-unknown = false;
>          targets.wasm32-wasip1 = false;
> +        targets.wasm32-wasip2 = false;
>  
>          build.packages = [
>            "wash-cli"
> ```

I'm working on updating Rust deps in nixify https://github.com/rvolosatovs/nixify/pull/278, meanwhile we can override `fenix` and `rust-overlay` to use more recent versions

rvolosatovs:
I pushed a commit pinning `nixify` to rev from https://github.com/rvolosatovs/nixify/pull/278 while CI is building that PR and switched to https://github.com/oxalica/rust-overlay, since https://github.com/nix-community/fenix was last updated 3 days ago and so does not have the most recent Rust version yet

brooksmtownsend:
Golang component build failures are unrelated and addressed in a commit in #3298. 

dog-fetcher seems to be failing with not being able to link a 0.2.0 stream type 🤔 digging

rvolosatovs (approved):
From what I understood from offline discussions regarding this is that the dog fetcher example will be fixed by updating the host so LGTM

thomastaylor312 (approved):

ricochet:
Confirmed dog-fetcher works manually with:

```bash
cargo install --path ./crates/wash-cli
wash up
cd examples/rust/components/dog-fetcher
wash build
wash app deploy wadm.yaml
```

PR #3372, opened 2024-10-18
fix(wash-lib): Store wadm.pid in the correct location

joonas:
## Feature or Problem

Currently `wash up` isn't able to bring wadm down once you send a ctrl-c or issue `wash down` since it's looking for the `wadm.pid` in a different directory (`~/.wash/downloads/wadm.pid` vs `~/.wash/wadm.pid`), this fixes that and adds a test to verify the expected behavior to avoid future regressions.


brooksmtownsend (approved):
LGTM

PR #3379, opened 2024-10-19
chore(core): Improve parse_wit_package_name handling

joonas:
## Feature or Problem

This fixes previously broken test cases, adds an actual test to run them.

### DIFF
--- a/crates/core/src/wit.rs
+++ b/crates/core/src/wit.rs
@@ -149,6 +149,18 @@ type WitInformationTuple = (
 /// ```
 /// # use semver::Version;
 /// # use wasmcloud_core::parse_wit_package_name;
+/// let (ns, packages, interfaces, func, version) = parse_wit_package_name("wasi:http").unwrap();
+/// # assert_eq!(ns, "wasi".to_string());
+/// # assert_eq!(packages, vec!["http".to_string()]);
+/// # assert_eq!(interfaces, None);
+/// # assert_eq!(func, None);
+/// # assert_eq!(version, None);
+/// let (ns, packages, interfaces, func, version) = parse_wit_package_name("wasi:http@0.2.2").unwrap();
+/// # assert_eq!(ns, "wasi".to_string());
+/// # assert_eq!(packages, vec!["http".to_string()]);
+/// # assert_eq!(interfaces, None);
+/// # assert_eq!(func, None);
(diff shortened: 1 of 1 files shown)


crates/core/src/wit.rs
```diff
@@ -241,3 +247,64 @@ pub fn parse_wit_package_name(p: impl AsRef<str>) -> Result<WitInformationTuple>
...
+        assert_eq!(version, Version::parse("0.2.0").ok());
+
+        let (ns, packages, interfaces, func, version) =
+            parse_wit_package_name("wasi:keyvalue/atomics.increment@0.2.0-draft")
+                .expect("should have parsed 'wasi:keyvalue/atomics.increment@0.2.0-draft'");
+        assert_eq!(ns, "wasi".to_string());
+        assert_eq!(packages, vec!["keyvalue".to_string()]);
+        assert_eq!(interfaces, Some(vec!["atomics".to_string()]));
+        assert_eq!(func, Some("increment".to_string()));
+        assert_eq!(version, Version::parse("0.2.0-draft").ok());
```
vados-cosmonic, line 308: Did you by chance miss the doctests? I think the new cases here are `wasi:http` and `wasi:http@0.2.2` so if you could put them up there that would be great

vados-cosmonic (commented):
LGTM :rocket: If we could keep the tests in the same place that would be great (and instruct people using it)

crates/core/src/wit.rs
joonas: so tl;dr why I'm adding them here is:
1. `cargo nextest`, which is what we use for CI, [does not support running doctests](https://github.com/nextest-rs/nextest/issues/16).
2. `nixify`, which is what we use for running tests, didn't have the ability to run doctests until [support for it just landed recently](https://github.com/ipetkov/crane/pull/720), but since that hasn't yet made it to `nixify` (@rvolosatovs mentioned he was looking to add it next week), I figured it would be nice to run these.

vados-cosmonic: Ah this is news to me, thanks for explaining!

vados-cosmonic (approved):
LGTM :rocket:

crates/core/src/wit.rs
joonas: No worries, I was a little surprised to find this out myself!

ricochet (approved):

PR #3380, opened 2024-10-20
fix: Make remaining doctests not fail

joonas:
## Feature or Problem

This fixes the doctest that depends on #3379 and updates the remaining doctests that were failing.

In the case of the doctests where I had to add `ignore`, that's because those are marked as `pub(crate)` and can't be accessed by doctests, so there's no way to actually build them (hence `ignore` instead of `no_run`).

### DIFF
--- a/crates/wash-cli/src/cmd/dev/deps.rs
+++ b/crates/wash-cli/src/cmd/dev/deps.rs
@@ -50,7 +50,7 @@ fn is_ignored_iface_dep(ns: &str, pkg: &str, iface: &str) -> bool {
 ///
 /// # Examples
 ///
-/// ```
+/// ```ignore
 /// let project_key = ProjectDependencyKey::Project {
 ///   name: "http-hello-world".into(),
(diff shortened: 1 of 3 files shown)


joonas:
Not entirely clear why the `wash drain` test is failing, none of the tests here impact that test, nor can I make that test fail locally 🤷‍♂️

markkovari:
hey @joonas if you look at main this one fails as well, so probably it is not your changes
https://github.com/wasmCloud/wasmCloud/actions/runs/11415547287/job/31765718616

On the other hand I tried to run main (or your branch), but I do have a bunch of failing integration tests as always 😃, so I definitely cannot help with the local part.

joonas:
This looks to be related to the issue https://github.com/wasmCloud/wasmCloud/pull/3381 should address.

vados-cosmonic:
Just approved #3381, should be good to rebase soon, hopefully!

vados-cosmonic (approved):
LGTM :rocket: 

Will merge this if the fix that went into main makes the test again

ricochet (approved):

PR #3381, opened 2024-10-20
fix(wash-cli): Fix leaked wadm process from integration tests

joonas:
## Feature or Problem

Since merging #3372, the [`wash drain` integration tests](https://github.com/wasmCloud/wasmCloud/actions/runs/11426722736/job/31790787988) started failing, which turns out to be because the [tests that run before it](https://github.com/wasmCloud/wasmCloud/actions/runs/11426722736/job/31790787988#step:11:1663) were orphaning the wadm process and weren't cleaning up the pid after themselves, this fixes that.

Also this being the case, it got me thinking that perhaps we ought to look into randomizing the order in which the tests are run, in case there are other cases where the order the tests are run in happens to matter (which it should not).

cc @markkovari for awareness


vados-cosmonic (approved):
LGTM :rocket:

Someday, we'll get around to making all these tests a little more parallel friendly :)

Issue #24, opened 2020-12-03
Add name field to ActorDescription and ProviderDescription


brooksmtownsend:
In a providers description, we have the ability currently to view its `id`, `link_name` and `image_ref`. If the `image_ref` is not present (e.g. a provider loaded from disk, or a native provider) the host inventory output can only show a provider ID (public key) and the link name (sometimes just `default`).

There's a similar issue for the ActorDescription, except we would only have the public key if it was loaded directly from disk.

Adding these two fields would be great for the command interface to be able to display robust inventory output.

PR #3385, opened 2024-10-21
chore: Drop forked tokio-postgres-rustls in favor of upstream update

joonas:
## Feature or Problem

We originally forked `tokio-postgres-rustls` due to the licensing issue with one of it's dependencies (`x509-certificate`), which was proposed to be swapped out for a different, licensing compliant, dependency (`x509-cert`) in https://github.com/jbg/tokio-postgres-rustls/pull/24.

Unfortunately that PR stalled out for a while, so we decided to pursue forking and applied #24 on top ourselves.

However, since then https://github.com/jbg/tokio-postgres-rustls/pull/27 was merged in and released as `v0.13.0`, so we're better off switching back to the official release.


vados-cosmonic (approved):
LGTM :rocket: 

SUPER nit, but would you mind making the commit in which you cut the release something like:

```
release(providers): sqldb-postgres v0.7.1
```

I assume you're going to tag the commit to trigger the release after this merges.

PR #3391, opened 2024-10-21
chore(examples): create artifacts for more examples

vados-cosmonic:
This commit enables more artifacts to be produced for examples

### DIFF
--- a/.github/workflows/examples-components.yml
+++ b/.github/workflows/examples-components.yml
@@ -7,16 +7,35 @@ on:
       - main
     tags:
       - component-blobby-v[0-9]+.[0-9]+.[0-9]+*
+      - component-blobby-v[0-9]+.[0-9]+.[0-9]+-*
       - component-dog-fetcher-v[0-9]+.[0-9]+.[0-9]+*
+      - component-dog-fetcher-v[0-9]+.[0-9]+.[0-9]+-*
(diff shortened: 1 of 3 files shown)


ricochet (approved):

.github/workflows/examples-components.yml
```diff
@@ -7,17 +7,37 @@ on:
       - main
     tags:
       - component-blobby-v[0-9]+.[0-9]+.[0-9]+*
+      - component-blobby-v[0-9]+.[0-9]+.[0-9]+-*
```
ricochet, line 10: This regex isn't right. It shouldn't hold up this PR but it should be without the final *:
- `component-blobby-v[0-9]+.[0-9]+.[0-9]+`
- `component-blobby-v[0-9]+.[0-9]+.[0-9]+-.*`

Right?

ricochet (changes requested):

.github/workflows/examples-components.yml
vados-cosmonic: Oh so it turns out that GHA actions aren't *QUITE* regex -- `*` is more like glob when at the end. Isn't that wonderful? See `wasmcloud.yml` for other examples

ricochet: <.<

ricochet (approved):

lachieh (approved):

PR #3392, opened 2024-10-21
fix: DX around duplicate links

vados-cosmonic:


crates/provider-http-server/tests/links.rs
```diff
@@ -0,0 +1,194 @@
...
+    let provider_http_config_name = "http-server-config";
+    let provider_http_address = format!(
+        "127.0.0.1:{}",
+        wasmcloud_test_util::os::free_port_ipv4().await?
+    );
+    let endpoint_url = format!("http://{provider_http_address}");
+
+    // Start NATS
+    let nats_container = NatsServer::default()
+        .with_cmd(["--jetstream"])
```
joonas, line 50: I wonder if we should just make this a default argument in the NatsServer?

JetStream seems like a fairly reasonable requirement every time, but I'm also totally fine keeping it just here if you think that'd be the best course of action for now, we can always revisit it if we see more examples where it would make sense.

vados-cosmonic: Ah this is a great idea, and I think you're right -- since this is the `testcontainers` feature provided by `wasmcloud-test-util`, it absolutely makes sense to add jetstream by default.

How do you feel about doing it as a follow up to this PR, would be short & sweet.

vados-cosmonic:
Hey @brooksmtownsend could I get a review here -- I think this PR is the most important one!

crates/provider-http-server/src/address.rs
```diff
@@ -80,19 +94,45 @@ impl Provider for HttpServerProvider {
...
+            link_config.target_id.to_string(),
+            link_config.link_name.to_string(),
         );
+        let mut sockets_by_link_name = self.sockets_by_link_name.write().await;
+        let mut handlers_by_socket = self.handlers_by_socket.write().await;
+
+        match sockets_by_link_name.entry(link_config.link_name.to_string()) {
+            // If a mapping already exists, and the stored address is different, disallow overwriting
+            std::collections::hash_map::Entry::Occupied(_) => {
+                bail!("overwriting links is not currently supported")
```
brooksmtownsend, line 107: Thoughts on noting why this was a conflict, e.g. a different component already has the address v.s. this component has a different address?

crates/provider-http-server/tests/links.rs
```diff
@@ -0,0 +1,194 @@
...
+    assert_remove_link(&ctl_client, provider_http_id, "wasi", "http", "rust")
+        .await
+        .context("failed to remove link 'rust'")?;
+
+    // At this point, the provider should be pointing to the jsonify component,
+    // as it was set *third* ('jsonify' link)
+    assert!(
+        !get_lowercased_resp().await?.contains("hello from rust"),
+        "third response be from jsonify component",
+    );
```
brooksmtownsend, line 191: This feels like a big behavior change, I can see the appeal to enable the other link getting put initially while you clean up older applications. I think once we have a mechanism to report failed links back to the host, we probably shouldn't enable this scenario to happen (in address based mode). I'm on the fence on if this is nice because the provider can't retry putting a link after it already failed (today) or if this will lead to confusing scenarios.

Any thoughts here? I could go either way here I think. Curious what spurred this change initially and I think that would help me understand

brooksmtownsend (approved):
After RTFM (reading the original issue) I can see the original motivation + goals. LGTM!

crates/provider-http-server/src/address.rs
vados-cosmonic: Yeah will add something to this to make it a little clearer

crates/provider-http-server/tests/links.rs
joonas: Follow-up seems totally fine here 👍

joonas (approved):

crates/provider-http-server/tests/links.rs
vados-cosmonic: Filed an issue for this! https://github.com/wasmCloud/wasmCloud/issues/3490

Issue #3233, opened 2024-10-07
[FEATURE] A Fully Featured wasmCloud Quickstart


brooksmtownsend:
### Affected project(s)

- [X] documentation
- [X] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

wasmCloud's current quickstart could use a little bit of love. The pages in documentation roughly cover
1. Installing `wash`
2. Cloning a new template in the language of your choice
3. wash build
4. wash up
5. wash app deploy
6. Implement path / query parameter parsing
7. wash stop component
8. Add keyvalue and logging
9. Update wadm
10. wash stop component
11. wash app deploy
12. Update wadm to scale across multiple hosts
13. wash app deploy
14. ...

During the course of the quickstart we also touch on a _ton_ of concepts, wadm, NATS, wasmCloud, hosts, the lattice, components, providers, links, WIT, etc etc etc.

### Describe the solution you'd like

I'd love to make the quickstart a little more quick, and a little more fully featured. To draw the outline in a different way:
1. Install `wash`
2. Clone a new template in the language of your choice
3. wash dev
4. Uncomment pre-coded sections to add an outbound HTTP request (wash dev auto-reloads)
5. Uncomment pre-coded sections to add a keyvalue request (wash dev auto-reloads)
6. ...
7. Quickstart complete

In the course of this quickstart, the developer will only really need to know the following concepts:
1. Components
2. wasmCloud
3. wash
4. Interface-driven development

### Describe alternatives you've considered

I brainstormed with a couple other maintainers on this, and led with the proposed component in https://github.com/wasmCloud/wasmCloud/pull/3142. The consensus was that this was a little too complicated out of the gate.

In the interest of not letting perfect be the enemy of the good, I'd like to propose even though there's a ton of work we could do here to add experiences like the [component-sdk-go](https://github.com/wasmCloud/component-sdk-go/tree/main) to each language, that we try and just display what we have today and improve from there.

### Additional context

As much as possible, in documentation, we should try and avoid displaying code directly unless we're really happy with it because it's a documentation maintenance cost.

Thank you @lxfontes, @vados-cosmonic , and @ericgregory for brainstorming this feature with me, please feel free to add additional context

PR #3399, opened 2024-10-22
feat(component)!: add idiomatic HTTP wrappers

brooksmtownsend:
## Feature or Problem

This PR seeks to greatly simplify the experience of writing components that use HTTP (incoming and outgoing.) The syntax for the macro was pulled from the [wasi-http 0.3 draft](https://github.com/WebAssembly/wasi-http/blob/main/wit-0.3.0-draft/handler.wit#L14), and once wasi-http@0.3.0 lands I plan to deprecate the macro entirely. Same thing for the fork of reqwest, once the upstream PR lands to support wasm32-wasip2 we'll drop the fork from our dependency list & use upstream directly.

As per the Rust [Overriding Dependencies](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/cargo/reference/specifying-dependencies.html#overriding-dependencies) section, I've chosen to submit an upstream feature and would like for this crate (and users of the `wasmcloud_component` crate) to be able to take advantage of this feature today, and then switch the dependency to the upstream released version when available

In addition to simplifying the experience of using HTTP, additional requirements taken into account are:
1. The resulting component _must_ use wasi-http directly, no wasmCloud specific interfaces
2. The forked crates _must_ have a path to upstream
3. The wrappers included _must not_ completely obfuscate the wasi-http types (for example, the `Request` object has a `into_inner()` method that converts the wrapper back into the `wasi::http::types::IncomingRequest`

This PR adds:
1. A macro wrapper `export!` for implementing the `wasi:http/incoming-handler` trait, replacing it with a simpler `HttpServer` trait that can be implemented
4. Wrapper `Request` and `ResponseBuilder` types that can be constructed & returned from the `HttpServer` trait instead of using wasi-http types directly
5. A dependency on [reqwest-wasmcloud](https://github.com/wasmCloud/reqwest-wasmcloud) so that the `wasmcloud_component` crate can be published & consumed today

TODO:
- [ ] After settling on this, release the reqwest-wasmcloud crate and dependents to remove the `git` dependencies present in the PR
- [x] Audit `wasmcloud_component` public members from the `http` wrapper module to ensure visibility is set correctly
- [ ] Additional functionality in upstream reqwest like `json()` requests, and the wrapper `http` module.
- [x] Remove the example implementations from the `http-hello-world` component

## Related Issues
Makes Rust viable for #3233 

## Release Information
wasmcloud_component 0.2.0

### Manual Verification
I manually verified that each of the examples in the `http-hello-world` component worked as expected, whether you read bytes fully into memory or take advantage of streaming. I tested that streaming an incoming HTTP request body back to the response and streaming an incoming HTTP response body to the response worked as intended.


examples/rust/components/dog-fetcher/wit/deps/wasi-cli-0.2.2/package.wit
```diff
@@ -1,19 +1,19 @@
-package wasi:cli@0.2.1;
+package wasi:cli@0.2.2;
```
rvolosatovs, line 1: looks like we should be able to just delete `wit` directory now that we don't generate bindings in this example?

examples/rust/components/dog-fetcher/src/lib.rs
```diff
@@ -28,70 +8,25 @@ struct DogResponse {
...
-        // Since it's likely that the URL we retrieved from the dog.ceo API is likely
-        // within that limit, we use blocking_write_and_flush() here.
-        //
-        // If we expected the body to possibly be longer, we'd need to loop and write chunks,
-        // paying attention to how to use the appropriate wasi:io APIs.
-        write_stream.write_all(dog_picture_url.as_bytes()).unwrap();
-        drop(write_stream);
-
-        OutgoingBody::finish(response_body, None).expect("failed to finish response body");
+wasmcloud_component::http::export!(DogFetcher);
```
ricochet, line 11: Massive improvement 😮✨

examples/rust/components/http-hello-world/wit/deps/wasi-cli-0.2.2/package.wit
```diff
@@ -1,19 +1,19 @@
-package wasi:cli@0.2.0;
```
rvolosatovs, line 1: like above, I'd not expect to see any WIT in this example

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,353 @@
...
+    inner: wasi::http::types::IncomingRequest,
+}
+
+impl TryFrom<wasi::http::types::IncomingRequest> for Request {
+    type Error = ErrorCode;
+
+    fn try_from(inner: wasi::http::types::IncomingRequest) -> Result<Self, Self::Error> {
+        Ok(Self { inner })
+    }
+}
```
rvolosatovs, line 44:
```suggestion
impl From<wasi::http::types::IncomingRequest> for Request {
    fn from(inner: wasi::http::types::IncomingRequest) -> Self {
        Self { inner }
    }
}
```

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,353 @@
...
+        type ComponentHttpExportAlias = $t;
+        ::wasi::http::proxy::export!(ComponentHttpExportAlias);
+    };
+}
+pub use export;
+
+/// Convert a  [`ResponseBuilder`] into an [`wasi::http::types::OutgoingResponse`] and set that response.
+///
+/// This is primarily public for the [`export`] macro and should only be used directly with care.
+pub fn set_outgoing_response(user_response: ResponseBuilder, response_out: ResponseOutparam) {
```
rvolosatovs, line 266: should this be `doc(hidden)`?

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,353 @@
...
+                )),
+            );
+        }
+        (None, None) => ResponseOutparam::set(response_out, Ok(response)),
+    }
+}
+
+/// Helper function to construct an internal server `ErrorCode` with an error message.
+fn error_code(e: impl ToString) -> ErrorCode {
+    ErrorCode::InternalError(Some(e.to_string()))
```
rvolosatovs, line 352: I guess it's fine for the first iteration, but we should not classify all errors as internal

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,353 @@
...
+        IncomingBody::finish(incoming_body);
+        Ok(buf)
+    }
+
+    /// Converts the request into the inner [`InputStream`] and [`IncomingBody`]
+    ///
+    /// These values have a tightly coupled lifecycle and the stream must be dropped
+    /// before the body is finished.
+    ///
+    /// Use this in tandem with [`ResponseBuilder::ok_stream`] to return a response with a body stream.
```
rvolosatovs, line 87: doesn't look like `ok_stream` exists

rvolosatovs:
For the CI `openssl` issue - could we use `rustls` in reqwest to fix it and remove the dep?

crates/component/src/wrappers/http.rs
brooksmtownsend: Heh, I saw this one doing another comb through too. Fixing!

crates/component/src/wrappers/http.rs
brooksmtownsend: Yeah this got overwritten over a refactor or two, fixing

examples/rust/components/dog-fetcher/wit/deps/wasi-cli-0.2.2/package.wit
brooksmtownsend: Looks like you can't just delete it outright, but I'll file a bug

brooksmtownsend:
The dog-fetcher is actually going to fail the build until https://github.com/wasmCloud/wasmCloud/pull/3371 merges, so I might back out the dog-fetcher change from this PR if it becomes a blocker

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,422 @@
...
+        Self {
+            status_code: Some(status_code),
+            body: None,
+            body_stream: Some((stream, body)),
+            headers: HeaderMap::new(),
+        }
+    }
+
+    /// Helper method to return a new [`ResponseBuilder`] with a 200 status code and the provided body.
+    pub fn ok(body: impl AsRef<[u8]>) -> Self {
```
rvolosatovs, line 276:
```suggestion
    pub fn ok(body: Vec<u8>) -> Self {
```

Since this will be turned into a vector anyway, let's not force a double allocation here if the users already have a vector

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,422 @@
...
+        outgoing_handler::ErrorCode,
+        types::{
+            Fields, IncomingBody, InputStream, Method, OutgoingBody, OutgoingResponse,
+            ResponseOutparam, Scheme,
+        },
+    },
+    io::streams::StreamError,
+};
+
+const HTTP_SCHEME: &str = "http";
```
rvolosatovs, line 44: Why not use https://docs.rs/http/latest/http/uri/struct.Scheme.html#associatedconstant.HTTP instead?

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,422 @@
...
+const HTTPS_SCHEME: &str = "https";
+
+/// Trait for implementing an HTTP server WebAssembly component that receives a
+/// [`Request`] and returns a [`ResponseBuilder`].
+pub trait HttpServer {
+    fn handle(request: Request) -> Result<ResponseBuilder, ErrorCode>;
```
rvolosatovs, line 50: I'm surprised this returns a `ResponseBuilder`, rather than a complete Response

I'd expect to be able to receive `http::Request`(https://docs.rs/http/latest/http/request/struct.Request.html) here and return an `http::Response` (https://docs.rs/http/latest/http/response/struct.Response.html)

I'd also expect to be able to set an arbitrary `http_body::Body` as the body of the response https://docs.rs/http-body/latest/http_body/trait.Body.html

Cargo.toml
```diff
@@ -267,6 +267,7 @@ rand = { version = "0.8", default-features = false }
 redis = { version = "0.25", default-features = false }
 regex = { version = "1", default-features = false }
 reqwest = { version = "0.12", default-features = false }
+reqwest-wasmcloud = { git = "https://github.com/wasmCloud/reqwest-wasmcloud/", branch = "master", default-features = false }
```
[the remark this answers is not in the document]
brooksmtownsend: Definitely! I just planned to hold on moving to the published crates until I knew I didn't have to change the API. I'll leave this comment as unresolved until I publish that crate + replace the ref

crates/component/src/wrappers/http.rs
brooksmtownsend: Thank you for the example, let me see what I can do to refactor this a bit. Returning the builder _is_ weird, I believe I did it this way so that I could deal with ownership of the response resources properly but I think I can refactor to return a response.

The only reason I have `wasmcloud_component::http::Response` is so that I can handle a `Vec<u8>` body and a `InputStream` body, so I might need to go back to the drawing board there. It would be great to use `http`

vados-cosmonic:
Hey @brooksmtownsend this can be closed right?

PR #3132, opened 2024-09-26
chore: Updating Go examples

lxfontes:
Refactored Go component examples code from wasi-preview1 to wasi-preview2.
CI Changes:
- Not building Go components with wash 0.29 as it doesn't support wasipreview2 for tinygo
- Bumped Go version to 1.23 in examples
- Bumped TinyGo version to 0.33 in examples
- Added `wasm-tools` via binstall ( wit-bindgen-go & tiny-go requirement )
- Added examples entry for `http-client-tinygo` ( was missing )

PR #3402, opened 2024-10-22
feat(example): update go hello world

brooksmtownsend:
## Feature or Problem
This PR takes bits and pieces from #3132 to just update the hello-world Go example, and ignores the `gen` directory while we were at it since interfaces are always regenerated during the build loop.

## Related Issues
#3132 can now remove updating the http-hello-world
https://github.com/wasmCloud/wasmcloud.com/pull/659 depends on this

## Release Information
Now that wash build can produce tinygo binaries with the `wasip2` target, this can be released anytime.


crates/wash-cli/src/cmd/dev/mod.rs
```diff
@@ -230,6 +230,8 @@ pub async fn handle_command(
                         .is_ok_and(|p| {
                             // Ignore Rust target directories
                             p.starts_with("target")
+                                // Ignore Go interface directories
+                                || p.starts_with("gen")
```
vados-cosmonic, line 234: Hmnnn this seems a little risky since people could have folders laid out however but let's cross that bridge when we get there.

vados-cosmonic: What about `bindings.wadge` files?

brooksmtownsend: I suppose we could ignore the bindings.wadge file too, it's just hard to do that for a specific file since that could change at any time 🤔

examples/golang/components/http-hello-world/main.go
```diff
@@ -0,0 +1,18 @@
...
+func init() {
+	wasihttp.HandleFunc(handleRequest)
+}
+
+func handleRequest(w http.ResponseWriter, r *http.Request) {
+	w.Write([]byte("Hello from Go!\n"))
+}
+
+//go:generate go run github.com/bytecodealliance/wasm-tools-go/cmd/wit-bindgen-go generate --world hello --out gen ./wit
+func main() {}
```
joonas, line 18: I think we should consider moving the `//go:generate` line to the top of the file, and the `func main() { .. }` further up, maybe right after `init()`.

May even be worth including a comment explaining why it's empty, because anyone familiar with Go would expect this to be where the magic happens, so to speak.

brooksmtownsend: Fair enough, are you thinking a simple "this is because we don't call main when we call the component, we call the HTTP request handle" would be sufficient?

brooksmtownsend: Updated the help text, let me know what you think 🙂

joonas: that seems great 👍

joonas (approved):

brooksmtownsend:
Example failing because `wasm-tools` needs to be installed

ricochet (approved):

thomastaylor312 (approved):

PR #3413, opened 2024-10-22
build(nix): fix OCI update script

rvolosatovs:
## Feature or Problem

- `skopeo` invocation would not copy the source image name and tag over to the pulled artifact, which caused the hash mismatch, that's now fixed
- enforce TLS verification for OCI fetches

## Related Issues

https://github.com/containers/skopeo/issues/1609


joonas (approved):

vados-cosmonic (approved):
LGTM :rocket:

ricochet (approved):

PR #3419, opened 2024-10-23
fix(ci): add missing wasm-tools binary for running golang tests

vados-cosmonic:
This commit ensures that the `wasm-tools` binary is installed, as it is required for running some integration tests, particularly the golang tests.


joonas (approved):
I'd love to move these to be versioned (I'm assuming the default behavior is `latest`?), but I think this is fine for now.

PR #3420, opened 2024-10-23
feat(ci): test more wash new templates

vados-cosmonic:
This commit adds more examples and languages to the testing done for `wash new`.


examples/tinygo
ricochet, line 1: I'm worried that this will confuse folks and potentially introduce weirdness on windows. We opted to name the example dir to be "golang" so that it is more readily recognizable for new users.

vados-cosmonic: Well don't worry, we're not using it quite yet! The `golang` is still the real folder -- it's mostly a CI artifact at this point

ricochet (approved):

PR #3423, opened 2024-10-23
fix(wash): delete manifest when dev stops

brooksmtownsend:
## Feature or Problem
As of 0.36.0, there may be multiple versions of a deployed manifest. We want to ensure we delete all versions when the dev session ends.

## Release Information
wash-cli 0.36.1


vados-cosmonic (approved):
LGTM :rocket:

PR #3424, opened 2024-10-23
fix: Only run wkg if 'deps.toml' is not present

lxfontes:
When using wash 0.36 with wrpc, we get an error:
```
 ❯ wash build

failed to build provider

Caused by:
    0: failed to load lock file
    1: unable to parse lock file from path
    2: TOML parse error at line 1, column 1
         |
       1 |
         | ^
       missing field `version`
```

This PR changes the wkg logic to skip dependencies fetching if `wit/deps.toml` file is present.

### DIFF
--- a/crates/wash-lib/src/build/mod.rs
+++ b/crates/wash-lib/src/build/mod.rs
@@ -108,7 +108,15 @@ pub async fn build_project(
     package_args: &CommonPackageArgs,
     skip_fetch: bool,
 ) -> Result<PathBuf> {
-    if !skip_fetch {
+    // NOTE(lxf): Check if deps.toml is in config.common.wit_dir, if it is, we skip fetching.
+    // This means the project hasn't been converted to wkg yet.
+    let wit_deps_exists = tokio::fs::try_exists(config.common.wit_dir.join("deps.toml")).await?;
+
+    if wit_deps_exists {
+        info!("Skipping fetching dependencies because deps.toml exists in the wit directory. Use 'wit-deps' to fetch dependencies.");
+    }
+
+    if !skip_fetch && !wit_deps_exists {
         // Fetch dependencies for the component before building
         let client = package_args.get_client().await?;
         let mut lock = load_lock_file(&config.wasmcloud_toml_dir).await?;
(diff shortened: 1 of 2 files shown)


ricochet (approved):

crates/wash-lib/src/build/mod.rs
```diff
@@ -108,7 +108,15 @@ pub async fn build_project(
     package_args: &CommonPackageArgs,
     skip_fetch: bool,
 ) -> Result<PathBuf> {
-    if !skip_fetch {
+    // NOTE(lxf): Check if deps.toml is in config.common.wit_dir, if it is, we skip fetching.
+    // This means the project hasn't been converted to wkg yet.
+    let wit_deps_exists = tokio::fs::try_exists(config.common.wit_dir.join("deps.toml")).await?;
+
+    if wit_deps_exists {
+        info!("Skipping fetching dependencies because deps.toml exists in the wit directory. Use 'wit-deps' to fetch dependencies.");
```
vados-cosmonic, line 116:
```suggestion
        eprintln!("Skipping fetching dependencies because deps.toml exists in the wit directory. Use 'wit-deps' to fetch dependencies.");
```

vados-cosmonic (approved):
LGTM :rocket:

crates/wash-lib/src/build/mod.rs
```diff
@@ -108,7 +108,15 @@ pub async fn build_project(
     package_args: &CommonPackageArgs,
     skip_fetch: bool,
 ) -> Result<PathBuf> {
-    if !skip_fetch {
+    // NOTE(lxf): Check if deps.toml is in config.common.wit_dir, if it is, we skip fetching.
+    // This means the project hasn't been converted to wkg yet.
+    let wit_deps_exists = tokio::fs::try_exists(config.common.wit_dir.join("deps.toml")).await?;
```
joonas, line 113:
```suggestion
    const DEPS_TOML: &str = "deps.toml";
    ...
    let wit_deps_exists = tokio::fs::try_exists(config.common.wit_dir.join(DEPS_TOML)).await?;
```

thomastaylor312 (changes requested):
This will make the migration process for components not great. I'll open up a PR with a different fix shortly

thomastaylor312 (approved):
We'll use this for now as a transitionary period

PR #3425, opened 2024-10-23
fix(wash-lib): Try to write the newly created LockFile before using it

joonas:
## Feature or Problem

Without this, we're greeted with the following error when `wash build` or `wash with build` fails initially:

```shell
$ wash wit build

failed to load lock file

Caused by:
    0: unable to parse lock file from path
    1: TOML parse error at line 1, column 1
         |
       1 | 
         | ^
       missing field `version`
```

### DIFF
--- a/crates/wash-lib/src/build/mod.rs
+++ b/crates/wash-lib/src/build/mod.rs
@@ -54,9 +54,14 @@ pub async fn load_lock_file(dir: impl AsRef<Path>) -> Result<LockFile> {
             .await
             .context("failed to load lock file")
     } else {
-        LockFile::new_with_path([], lock_file_path)
+        let mut lock_file = LockFile::new_with_path([], lock_file_path)
             .await
-            .context("failed to create lock file")
+            .context("failed to create lock file")?;
+        lock_file
+            .write()
+            .await
+            .context("failed to write newly created lock file")?;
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM :rocket:

crates/wash-lib/src/build/mod.rs
```diff
@@ -54,9 +54,14 @@ pub async fn load_lock_file(dir: impl AsRef<Path>) -> Result<LockFile> {
             .await
             .context("failed to load lock file")
     } else {
-        LockFile::new_with_path([], lock_file_path)
+        let mut lock_file = LockFile::new_with_path([], lock_file_path)
             .await
-            .context("failed to create lock file")
+            .context("failed to create lock file")?;
+        lock_file
```
thomastaylor312, line 60: For knowledge purposes, the main issue that is causing this is that we create an empty file and then at some point try to load the lock file again (so I don't quite know where that is happening). You can see that this is what happens if you look at the docs for `new_with_path`.

joonas: Yep, I followed down the path to see what was happening in `wasm_pkg_core` and came to the conclusion that our best here would be to persist the `LockFile` before we pass it back.

Issue #2884, opened 2024-08-28
[FEATURE] Deprecate `wash app list` in favor of `wash app get`


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Currently, retrieving all applications and retrieving a single application are two separate commands, `wash app list` and `wash app get <app_name>`. 

### Describe the solution you'd like

Following what other tools do, and what we do for `wash get inventory`, I would like for `wash app get` with no name provided to query the list models endpoint.

### Additional context

This follows the precedent in https://github.com/wasmCloud/wadm/pull/400, and will need a followup to actually query the new endpoint once wadm 0.15 is released.

PR #3431, opened 2024-10-23
refactor(wash-cli): deprecate wash app list in favor of wash app get

markkovari:
## Feature or Problem

Merging `wash app list` and `wash app get` commands.

I do need one clarification: we probably want to keep the periodically watched part of the list to the `wash app get` without a name, am I correct? If yes, I'll re-add it, and will write some tests. I just tested locally, but would bring more confidence if those would be automated. 

## Related Issues

fixes: #2884 

## Consumer Impact

Should reduce the number commands for the same thing.

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -10,7 +10,6 @@ use wadm_client::Result;
 use wadm_types::api::ModelSummary;
 use wadm_types::validation::{ValidationFailure, ValidationOutput};
 use wash_lib::app::{load_app_manifest, validate_manifest_file, AppManifest};
-use wash_lib::cli::get::parse_watch_interval;
 use wash_lib::cli::{CliConnectionOpts, CommandOutput, OutputKind};
 use wash_lib::config::WashConnectionOptions;
 
@@ -25,10 +24,7 @@ mod output;
 
 #[derive(Debug, Clone, Subcommand)]
 pub enum AppCliCommand {
-    /// List all applications available within the lattice
-    #[clap(name = "list")]
-    List(ListCommand),
-    /// Get the application manifest for a specific version of an application
+    /// Get the application manifest for applications
     #[clap(name = "get")]
     Get(GetCommand),
     /// Get the current status of a given application
@@ -52,16 +48,6 @@ pub enum AppCliCommand {
     /// Validate an application manifest
     #[clap(name = "validate")]
     Validate(ValidateCommand),
-}
-
-#[derive(Args, Debug, Clone)]
-pub struct ListCommand {
-    #[clap(flatten)]
-    opts: CliConnectionOpts,
-
-    /// Enables Real-time updates, duration can be specified in ms or in humantime (eg: 5s, 2m, 15ms). Defaults to 1000 milliseconds.
-    #[clap(long,short, num_args = 0..=1, default_missing_value = "1000", value_parser = parse_watch_interval)]
-    pub watch: Option<std::time::Duration>,
 }
 
 #[derive(Args, Debug, Clone)]
@@ -125,9 +111,9 @@ pub struct PutCommand {
 
 #[derive(Args, Debug, Clone)]
 pub struct GetCommand {
-    /// The name of the application to retrieve
+    /// The name of the application to retrieve. If left empty, retrieves the list of applications
     #[clap(name = "name")]
-    app_name: String,
+    app_name: Option<String>,
 
     /// The version of the application to retrieve. If left empty, retrieves the latest version
     #[clap(name = "version")]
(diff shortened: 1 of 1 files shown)


vados-cosmonic (commented):
Hey @markkovari this looks fantastic so far! Do we have this covered under tests properly? It might be nice to add a test case for the new syntax if we don't have one (I assume we have one for the non-list case of `wash get`.

markkovari:
tl;dr

a, move the current `wash app list` functionality under the `wash app get` when no name
b, remove the wash app list and `wash app get` should return the yaml/json
c, something else

@vados-cosmonic 
no, I was about to check the tests and expected them to fail due to the removal of the other command. So I'll look at the test failures, but I still do not know if we just wanted to yoink all the yaml/json to the screen. Or we want to move the current `wash app list` render? Putting back to draft meanwhile

@brooksmtownsend you are the creator of the ticket can you (or anyone else) help me out?

vados-cosmonic:
>  no, I was about to check the tests and expected them to fail due to the removal of the other command. So I'll look at the test failures, but I still do not know if we just wanted to yoink all the yaml/json to the screen. Or we want to move the current wash app list render? Putting back to draft meanwhile

I think moving the functionality around is definitely a good start -- if it just did exactly the same thing as before that'd be fine for a start I think!

vados-cosmonic:
Rebasing to get the `oci`step fix in here so you can stop seeing red in the CI runs! :)

Thanks again for working on this

markkovari:
Yes, I was kinda lost why are the golang tests failing if I change the command line parameters.
I will re-add then the `wash app list` and and rebase. (will come back to this tomorrow) Thanks all for the feedback.

crates/wash-cli/src/app/mod.rs
```diff
@@ -125,14 +125,18 @@ pub struct PutCommand {
 
 #[derive(Args, Debug, Clone)]
 pub struct GetCommand {
-    /// The name of the application to retrieve
+    /// The name of the application to retrieve. If left empty retrieves all the applications, same as `wash app list`
```
vados-cosmonic, line 128: NIT: The first line will show up in rustdoc, so might be nice to leave a space after the first sentence here

crates/wash-cli/src/app/mod.rs
```diff
@@ -125,14 +125,18 @@ pub struct PutCommand {
...
+    app_name: Option<String>,
 
     /// The version of the application to retrieve. If left empty, retrieves the latest version
     #[clap(name = "version")]
     version: Option<String>,
 
+    /// Enables Real-time updates, duration can be specified in ms or in humantime (eg: 5s, 2m, 15ms). Defaults to 1000 milliseconds.
+    #[clap(long,short, num_args = 0..=1, default_missing_value = "1000", value_parser = parse_watch_interval)]
+    pub watch: Option<std::time::Duration>,
+
```
vados-cosmonic, line 139: NIT: same note about the spacing in the comment

Also, wondering if we should default to milliseconds here or seconds since it's on human scale...

crates/wash-cli/src/app/mod.rs
```diff
@@ -176,11 +180,16 @@ pub async fn handle_command(
...
+            get_application_list(cmd, &sp).await
         }
         Get(cmd) => {
-            sp.update_spinner_message("Getting application manifest ... ".to_string());
-            get_manifest(cmd).await
+            if let Some(app_name) = cmd.clone().app_name {
+                sp.update_spinner_message("Getting applications ... ".to_string());
+                get_manifest(cmd, &app_name).await
+            } else {
+                sp.update_spinner_message("Getting application manifest ... ".to_string());
```
vados-cosmonic, line 190:
```suggestion
            if let Some(app_name) = cmd.clone().app_name {
                sp.update_spinner_message("Getting applications... ".to_string());
                get_manifest(cmd, &app_name).await
            } else {
                sp.update_spinner_message("Getting application manifests... ".to_string());
```

crates/wash-cli/tests/common/mod.rs
```diff
@@ -601,12 +601,12 @@ impl TestWashInstance {
             .context("failed to parse output of `wash app deploy`")
     }
 
-    /// Trigger the equivalent of `wash app list` on a [`TestWashInstance`]
-    pub(crate) async fn list_apps(&self) -> Result<AppListCommandOutput> {
+    /// Trigger the equivalent of `wash app get` on a [`TestWashInstance`]
+    pub(crate) async fn list_apps(&self) -> Result<AppGetCommandOutput> {
```
vados-cosmonic, line 605: Could we duplicate this test until we formally remove `wash app list` ?

vados-cosmonic (approved):
LGTM :rocket: 

Nothing but nits!

crates/wash-cli/tests/common/mod.rs
markkovari: Yes, I was thinking about as well. Re-added the two different version and tested the output of those. 🙇

vados-cosmonic (approved):
LGTM :rocket: 

Thanks for tackling this @markkovari !

PR #3432, opened 2024-10-23
feat(wash-lib): Require tinygo 0.33.0+ for wasip2 builds

joonas:
## Feature or Problem

Since you can't build WASI P2 with anything below 0.33.0, we should check to make sure the appropriate version is installed before attempting to build.

### DIFF
--- a/crates/wash-lib/src/build/component.rs
+++ b/crates/wash-lib/src/build/component.rs
@@ -315,6 +315,33 @@ async fn build_tinygo_component(
             ".",
         ],
         WasmTarget::WasiP2 => {
+            // If we are using tinygo without a custom path, check the tinygo
+            // version to make sure that it can support WasiP2, aka we're using
+            // 0.33.0+
+            if tinygo_config.tinygo_path.is_none() {
+                // Run `tinygo version` and parse the output for version
+                let output = tokio::process::Command::new("tinygo")
+                    .arg("version")
+                    .output()
+                    .await
+                    .context("failed to get tinygo version")?
+                    .stdout;
+
+                let version = std::str::from_utf8(&output)
+                    .context("failed to parse tinygo version")?
+                    .split(' ')
+                    .filter_map(|s| semver::Version::parse(s).ok())
+                    .next()
+                    .unwrap_or_else(|| semver::Version::new(0, 0, 0));
+
+                let req = semver::VersionReq::parse(">=0.33.0")
+                    .context("failed to generate tinygo version requirement")?;
+
+                // Check whether we have an appropriate version of tinygo, bail if we do not.
+                if !req.matches(&version) {
+                    bail!("tinygo version needs to be 0.33.0 or above in order to build wasm32-wasip2 components.");
+                }
+            }
             let mut args = vec![
                 "build",
                 "-o",


crates/wash-lib/src/build/component.rs
```diff
@@ -315,6 +315,34 @@ async fn build_tinygo_component(
...
+                    .stdout;
+
+                let version = std::str::from_utf8(&output)
+                    .context("failed to parse tinygo version")?
+                    .split(' ')
+                    .filter_map(|s| semver::Version::parse(s).ok())
+                    .next()
+                    .unwrap_or_else(|| semver::Version::new(0, 0, 0));
+
+                let req = semver::VersionReq::parse(">=0.33.0")
```
lxfontes, line 337: oh this was less painful than i imagined

crates/wash-lib/src/build/component.rs
```diff
@@ -315,6 +315,34 @@ async fn build_tinygo_component(
...
+                    .context("failed to generate tinygo version requirement")?;
+
+                // Check whether we have an appropriate version of tinygo, bail if we do not.
+                if !req.matches(&version) {
+                    // eprintln!("tinygo version needs to be 0.33.0 or above in order to build wasm32-wasip2 components.");
+                    bail!("tinygo version needs to be 0.33.0 or above in order to build wasm32-wasip2 components.");
+                }
```
ricochet, line 344: For a potential follow-up: should we also check the version in go.mod?

joonas: do you mean `go` version? Is there a specific version we depend on, or are you referring to `tinygo`'s dependency? If latter, I believe `tinygo` actually does the right thing already

ricochet (approved):

joonas:
FWIW, building against `wasm32-wasi-preview2` target with the wrong version of `tinygo` looks like this:
```shell
$ /Users/joonas/dev/src/joonas/wasmcloud/target/release/wash build --skip-fetch
Output dir: gen
Package root: github.com/cosmonic-labs/wasmcloud.space/components/task-manager/gen
Generated 40 package(s)
Generated package: github.com/cosmonic-labs/wasmcloud.space/components/task-manager/gen/demo/task-manager/tracker
...

tinygo version needs to be 0.33.0 or above in order to build wasm32-wasip2 components.
```

vados-cosmonic (approved):
LGTM :rocket:

crates/wash-lib/src/build/component.rs
vados-cosmonic: Don't let anyone tell you Rust can't be ergonomic!

crates/wash-lib/src/build/component.rs
```diff
@@ -315,6 +315,33 @@ async fn build_tinygo_component(
...
+                    .await
+                    .context("failed to get tinygo version")?
+                    .stdout;
+
+                let version = std::str::from_utf8(&output)
+                    .context("failed to parse tinygo version")?
+                    .split(' ')
+                    .filter_map(|s| semver::Version::parse(s).ok())
+                    .next()
+                    .unwrap_or_else(|| semver::Version::new(0, 0, 0));
```
vados-cosmonic, line 335:
```suggestion
                    .unwrap_or_else(|| semver::Version::new(0, 0, 0));
```

Is this possible? should we just error if we can't figure out a version?

rvolosatovs:
Do we also check `wasm-tools` version?
Go reactors will not work with versions before 1.217

Issue #3257, opened 2024-10-09
[FEATURE] Use humantime instead of milliseconds for time values


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

For `wash` and `wasmcloud`, we have many timeout/delay/execution limit time values that are stated in milliseconds. This is fine for smaller values on the order of thousands, but when you start to think about max execution time in minutes this gets unwieldy (e.g. MAX_EXECUTION_TIME_MS=600000)

### Describe the solution you'd like

I'd like to deprecate the flags that take timeout values in wasmCloud and wash and replace them with [humantime](https://crates.io/crates/humantime) parseable values. So instead of supplying "600000" milliseconds for 10 minutes (which you could still do with 600000ms) you could simply supply "10m"

For wasmCloud it's very important this is done in a backwards compatible way, and we retain the older flags (and hide them) until we release v2.0.0. For wash flags can be deprecated in one minor release and removed in the next.

omarjatoi:
👋 just indicating my intent to work on this issue. I started collecting all the places where the `wash` cli is currently using a non-humantime value, and will do the same for `wasmcloud` host. Should have a PR up in a few days!

vados-cosmonic:
Hey @omarjatoi how do you feel about adding the new options that use humantime *alongside* the existing `-ms` extensions? The non-`-ms` flags can reasonably expect humantime (and fail otherwise)

vados-cosmonic:
Yeah that first approach sounds reasonable -- if we can re-use the function that does parsing for both that would be great.

Some tests would go a long way in making sure we get the exact functionality you're thinking about as well.

omarjatoi:
Sounds good. Thanks @vados-cosmonic!

vados-cosmonic:
No worries! Thanks for checking in -- feel free to let us know if you have any questions or run into anything else !

brooksmtownsend:
Just chiming in to say what you're planning and working on @omarjatoi looks perfect 😄 Thank you so much for the thoughtful approach to this issue!

omarjatoi:
[...]

PR #3440, opened 2024-10-24
[WIP] feat: Use humantime instead of milliseconds for time values (#3257)

omarjatoi:
## Feature or Problem

Current CLI options take values as milliseconds, e.g. `--some-option-ms 100`, and we instead want to use [`humantime`] to enable options like `--some-option 100ms`.

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/3257

## Release Information

`next`

## Consumer Impact

This change is taking care to be backward compatible, and maintains the existing options and environment variables. The new functionality supports falling back to the existing functionality as well (e.g. setting `SOME_ENV_VAR` to `100` and `100ms` will have the same effect).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,6 +3,7 @@ use core::net::SocketAddr;
 use std::collections::{HashMap, HashSet};
 use std::env;
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::sync::{Arc, LazyLock};
 use std::time::Duration;
 
@@ -79,7 +80,10 @@ struct Args {
     #[clap(long = "host-seed", env = "WASMCLOUD_HOST_SEED")]
     host_seed: Option<String>,
     /// Delay, in milliseconds, between requesting a provider shut down and forcibly terminating its process
-    #[clap(long = "provider-shutdown-delay-ms", alias = "provider-shutdown-delay", default_value = "300", env = "WASMCLOUD_PROV_SHUTDOWN_DELAY_MS", value_parser = parse_duration_millis)]
+    #[clap(long = "provider-shutdown-delay-ms", default_value = "300", env = "WASMCLOUD_PROV_SHUTDOWN_DELAY_MS", value_parser = parse_duration_millis, hide = true, conflicts_with = "provider_shutdown_delay")]
+    provider_shutdown_delay_ms: Duration,
+    /// Delay between requesting a provider shut down and forcibly terminating its process
+    #[clap(long = "provider-shutdown-delay", alias = "provider-shutdown-delay", default_value = "300ms", env = "WASMCLOUD_PROV_SHUTDOWN_DELAY", value_parser = parse_duration_fallback_ms)]
     provider_shutdown_delay: Duration,
     /// Determines whether OCI images tagged latest are allowed to be pulled from OCI registries and started
     #[clap(long = "allow-latest", env = "WASMCLOUD_OCI_ALLOW_LATEST")]
@@ -185,8 +189,11 @@ struct Args {
     #[clap(long = "rpc-creds", env = "WASMCLOUD_RPC_CREDS", hide = true, conflicts_with_all = ["rpc_jwt", "rpc_seed"])]
     rpc_creds: Option<PathBuf>,
     /// Timeout in milliseconds for all RPC calls
-    #[clap(long = "rpc-timeout-ms", default_value = "2000", env = "WASMCLOUD_RPC_TIMEOUT_MS", value_parser = parse_duration_millis, hide = true)]
+    #[clap(long = "rpc-timeout-ms", default_value = "2000", env = "WASMCLOUD_RPC_TIMEOUT_MS", value_parser = parse_duration_millis, hide = true, conflicts_with = "rpc_timeout")]
     rpc_timeout_ms: Duration,
+    /// Timeout for all RPC calls
+    #[clap(long = "rpc-timeout", default_value = "2000ms", env = "WASMCLOUD_RPC_TIMEOUT", value_parser = parse_duration_fallback_ms, hide = true)]
+    rpc_timeout: Duration,
     /// Optional flag to require host communication over TLS with a NATS server for RPC messages
     #[clap(long = "rpc-tls", env = "WASMCLOUD_RPC_TLS", hide = true)]
     rpc_tls: bool,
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -323,7 +343,11 @@ struct Args {
     pub tls_ca_paths: Option<Vec<PathBuf>>,
 
     /// If provided, overrides the default heartbeat interval of every 30 seconds. Provided value is interpreted as seconds.
-    #[arg(long = "heartbeat-interval-seconds", env = "WASMCLOUD_HEARTBEAT_INTERVAL", value_parser = parse_duration_secs, hide = true)]
+    #[arg(long = "heartbeat-interval-seconds", env = "WASMCLOUD_HEARTBEAT_INTERVAL_SECONDS", value_parser = parse_duration_secs, hide = true, conflicts_with = "heartbeat_interval")]
```
omarjatoi, line 346: This change is _slightly_ problematic in terms of backward compatibility, i.e. if someone was already using the `WASMCLOUD_HEARTBEAT_INTERVAL` environment variable with a value like `10`, we would parse that as 10 **milliseconds** instead of **seconds**.

I could leave the env var as is, and call the new env var something like `WASMCLOUD_HEARTBEAT_INTERVAL_TIME` 😅

vados-cosmonic: Yeah interesting -- I kind of prefer getting this right/reasonable as early as possible (if we're going to break, break sooner rather than later and guide people forward)...  BUT if someone had this specified, going from 10 seconds to 10ms is quite intense.

Maybe it make sense to write a parser that *defaults* to seconds if unspecified but uses humantime otherwise?

omarjatoi: That makes sense. I could extend `parse_duration` to take an optional `unit` parameter that defaults to `ms` but in this case I can pass in `sec` as an argument?

vados-cosmonic:
Hey @omarjatoi thanks for working on this, it looks great so far -- would you mind adding some tests? Some regression tests and some failure case tests at least would be greatly appreciated.

Ensuring that we get the right expected duration (especially with respect to things that *were* seconds but could be turned into milliseconds with the change, etc) would be nice -- we can lock in the functionality/make sure it doesn't break going forward.

src/main.rs
t3hmrman: Yeah that would be great -- I'm not sure how you could make that work with how you call it via `clap` but that sounds reasonable to me

vados-cosmonic: Yeah that would be great -- I'm not sure how you could make that work with how you call it via `clap` but that sounds reasonable to me

omarjatoi: Ah okay, can't do partial application or something for functions in the clap macro. I'm just going to update the `parse_duration_secs` method to _either_ parse `humantime` _or_ keep existing behavior i.e. get duration from `u64` assuming seconds.

omarjatoi:
Thanks! Yeah I'll add some tests to this before removing the `[WIP]` label.

src/main.rs
omarjatoi: This appears to work, but IMO makes things less readable
```rust
enum TimeUnit {
    Ms,
    Sec,
}

fn parse_duration(arg: &str, unit: Option<TimeUnit>) -> anyhow::Result<Duration> {
    if let Ok(duration) = humantime::Duration::from_str(arg) {
        return Ok(duration.into());
    }
    match unit {
        Some(TimeUnit::Sec) => {
            if let Ok(secs) = arg.parse::<u64>() {
                return Ok(std::time::Duration::from_secs(secs));
            }
        }
        _ => {
            if let Ok(millis) = arg.parse::<u64>() {
                return Ok(std::time::Duration::from_millis(millis));
            }
        }
    }
    Err(anyhow::anyhow!("Invalid duration: '{}'. Expected a duration like '5s', '1m', '100ms', or milliseconds as an integer.", arg))
}
```
and in `clap`:
```rust
    /// If provided, overrides the default heartbeat interval of every 30 seconds. Provided value is interpreted as seconds.
    #[arg(long = "heartbeat-interval", env = "WASMCLOUD_HEARTBEAT_INTERVAL", value_parser = |s: &str| parse_duration(s, Some(TimeUnit::Sec)), hide = true)]
    heartbeat_interval: Option<Duration>,
```
but we need to add that closure everywhere `|s: &str| parse_duration(s, Some(TimeUnit::Ms))` which doesn't feel worth it to me

vados-cosmonic: I think you're really clsoe to something usable there! That said, it might be simpler to just write *really light* wrappers that use that big fn -- `parse_time_fallback_seconds` might be a decent name?

vados-cosmonic:
Oh note that there's also a "draft PR" feature in GH -- you can switch existing PRs to draft as well if you'd like (though of course we all understand `[WIP]` here :)

src/main.rs
omarjatoi: I like that idea, thanks! Pushed https://github.com/wasmCloud/wasmCloud/pull/3440/commits/b138fe2c9bfa02b1e69e8281337f14f7d6945552

brooksmtownsend:
Looking really nice! @vados-cosmonic let me know if you want an additional set of eyes here. Thank you @omarjatoi for the PR + incoming tests!

vados-cosmonic:
Hey @brooksmtownsend please feel free if you've got time :) I'd appreciate another set of eyes, and if by chance I'm not here to merge this quick enough once it's ready, please do!

omarjatoi:
Just as a heads up, I'm probably going to continue (`wash` cli and tests) tonight/tomorrow since I have some other commitments today, but I'll mark the PR ready for review (and remove the `[WIP]` in title) if you'd rather review the whole change to save time!

vados-cosmonic:
Absolutely no rush @omarjatoi ! Thanks again for picking this up/carefully going about it!

brooksmtownsend:
Hey @omarjatoi, just checking in here, hows the work going? Do you still have bandwidth to wrap this one up?

omarjatoi:
Sorry for not updating the PR, yes still plan to complete it just had some other priorities! Thanks for the nudge, I'll try to find some time for it this coming week.

omarjatoi:
Nope sorry sorry, still open 😅

brooksmtownsend:
Hey @omarjatoi just wanted to check in on this one, let us know if you need a maintainer to take it over the line or if you're planning on coming back to it 🙂

omarjatoi:
Hey there, yeah sorry for dragging my feet on this things have been kinda hectic! This coming weekend is fairly free for me so I'll try to make some more progress. What's left is just rebasing and some tests, and then doing a quick sanity check.

brooksmtownsend:
Closing this one out for now since it hasn't had an update in a bit

PR #3441, opened 2024-10-24
chore(ci): remove wit-bindgen-cli from test runs

vados-cosmonic:
This commit removes `wit-bindgen-cli` from tests, which was required when building golang tooling. Since upstream deps have changed (it no longer tries to call `wit-bindgen` but instead calls `wasm-tools`, we should be OK removing `wit-bindgen-cli`.

### DIFF
--- a/examples/golang/providers/custom-template/component/customtemplate.go
+++ b/examples/golang/providers/custom-template/component/customtemplate.go
@@ -3,22 +3,22 @@ package main
 import (
 	"fmt"
 
-	gen "github.com/wasmcloud/wasmcloud/examples/golang/providers/custom-template/component/gen"
+	// gen "github.com/wasmcloud/wasmcloud/examples/golang/providers/custom-template/component/gen"
 )
 
-type CustomTemplateComponent struct{}
+// type CustomTemplateComponent struct{}
 
-func init() {
-	customTemplate := CustomTemplateComponent{}
-	gen.SetExportsWasmcloudExample0_1_0_ProcessData(customTemplate)
-}
+// func init() {
+// 	customTemplate := CustomTemplateComponent{}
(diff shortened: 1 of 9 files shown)


examples/golang/providers/custom-template/component/customtemplate.go
```diff
@@ -3,22 +3,22 @@ package main
...
+// type CustomTemplateComponent struct{}
 
-func init() {
-	customTemplate := CustomTemplateComponent{}
-	gen.SetExportsWasmcloudExample0_1_0_ProcessData(customTemplate)
-}
+// func init() {
+// 	customTemplate := CustomTemplateComponent{}
+// 	gen.SetExportsWasmcloudExample0_1_0_ProcessData(customTemplate)
+// }
```
brooksmtownsend, line 14: This should probably all be commented back in right?

brooksmtownsend (changes requested):

vados-cosmonic:
Sorry this wasn't clear but this was a WIP, I should have marked it so -- https://github.com/wasmCloud/wasmCloud/pull/3441/commits/0d9dcff54ee31ac62fcef4700fd7cc15923ee100

What I found out ~4 days ago was that this code doesn't actually work, especially if you don't have the tooling installed, because `go run` won't actually download and install the thing automatically, you need the "tools.go" pattern for that, which is a wider change, and there's a PR that does [that and other stuff](https://github.com/wasmCloud/wasmCloud/pull/3132) which actually should be broken down and used instead (excluding the "tools.go" adoption change, IIRC). 

Closing this for now, as other PRs have started landing to update the Go examples as well!

joonas:
it's worth noting that with the changes that are going in to [component-sdk-go](https://github.com/wasmcloud/component-sdk-go), I think fast approaching state where we will no longer need the `tools.go` trick.

PR #3442, opened 2024-10-24
feat(wash-lib): increase allowed scale component time

vados-cosmonic:
This commit increases the amount of time waited by default for starting components, to make it a little less unlikely that JS and/or python components that have to be pulled over possibly slow internet connections fail to start.

At present, it often happens that components fail to scale but start nearly immediately after the error is shown.


crates/wash-lib/src/config.rs
```diff
@@ -18,7 +18,7 @@ pub const DEFAULT_NATS_HOST: &str = "127.0.0.1";
 pub const DEFAULT_NATS_PORT: &str = "4222";
 pub const DEFAULT_LATTICE: &str = "default";
 pub const DEFAULT_NATS_TIMEOUT_MS: u64 = 2_000;
-pub const DEFAULT_START_COMPONENT_TIMEOUT_MS: u64 = 5_000;
+pub const DEFAULT_START_COMPONENT_TIMEOUT_MS: u64 = 10_000;
```
ricochet, line 21: I think this should be split in two. One for time to pull to account for network latency and another for time to start. Or this one is only applied when you know the component has not already been pulled? We should definitely be erroring if a component fails to start in less than second, let alone 10.

vados-cosmonic: I am in full agreement, BUT at the wash level we don't have control over fetching the image versus starting it. And IIRC we can't scale a component "up" to zero (@brooksmtownsend please correct me if I'm wrong!) to replicate a distinct fetch-only operation.

That's definitely a host level change and IMO something we want. IIRC this has come up before, the ability to "fetch" an image ref specifically, and might have been pushed down the road!

ricochet (approved):

crates/wash-lib/src/config.rs
vados-cosmonic: Oh just a note here, I'll make an issue for this as a follow up!

brooksmtownsend: Yeah this is a little rough, because there's three phases of the start process
1. Send request, get ack that request was valid
2. Component downloads from OCI :spinner:
3. Once downloaded, component instantiates

Step 2 is really the big question mark here, because a bigger component just takes more time. It's worth noting here that this timeout value is the amount of time we wait for an event, successful or failed, to come back for that individual component. Given that I'm okay with this change because it only increases timeouts for components that take longer to start.

Anyone who wants to skip the confirmation can add `--skip-wait` to the control interface command to avoid waiting for the event

brooksmtownsend (approved):

PR #3444, opened 2024-10-24
feat(component): add `http` crate integration

rvolosatovs:
## Feature or Problem

- Ported over most of https://github.com/bytecodealliance/wasi-rs/pull/90
- Adapted a few things from https://github.com/wrpc/http/blob/dbc6a6ba307ef9f53e175da9982c5a757d1a7eaa/src/lib.rs
- On top of the above, implemented `OutgoingBody` abstraction and `IncomingBody` wrapper
- Added From family of traits analogues to `wasmcloud_component`, these should go away once WASI support is added to upstream libraries (I'll make sure to extract stuff from this PR to e.g. `http` crate)

This PR is meant to be a basis on top of which #3399 could be implemented further. I have not done much testing of this PR, just something we quickly sketched together with @brooksmtownsend and I later wrapped it up.

Going forward, I think we should add `ServerAsync`, which would create a Tokio runtime and implement everything on top of that. That would *not* be a breaking change, rather an additive one with support for both sync and async servers. That would allow us to integrate with Axum, Tower and the like. Maybe I'll have some capacity to make a draft tomorrow for this.

## Related Issues

https://github.com/wasmCloud/wasmCloud/pull/3399

### DIFF
--- a/crates/component/src/wrappers/http.rs
+++ b/crates/component/src/wrappers/http.rs
@@ -1,0 +1,565 @@
+//! This module provides utilities for writing HTTP servers and clients using the WASI HTTP API.
+//!
+//! It's inspired by the WASI 0.3 proposal for <https://github.com/WebAssembly/wasi-http> and will
+//! be supported until the release of wasi:http@0.3.0. After that, this module will likely be deprecated.
+//!
+//! ```rust
+//! use wasmcloud_component::http;
+//!
+//! struct Component;
+//!
+//! http::export!(Component);
+//!
+//! // Implementing the [`Server`] trait for a component
+//! impl http::Server for Component {
+//!     fn handle(_request: http::IncomingRequest) -> http::Result<http::Response<impl http::OutgoingBody>> {
+//!         Ok(http::Response::new("Hello from Rust!"))
+//!     }
+//! }
+//! ```
+use core::fmt::Display;
+use core::marker::PhantomData;
+use core::ops::{Deref, DerefMut};
+
+use std::io::{Read, Write};
+
+use anyhow::{anyhow, Context as _};
+use wasi::http::types::{OutgoingResponse, ResponseOutparam};
+use wasi::io::streams::{InputStream, OutputStream, StreamError};
+
+pub use http::{
+    header, method, response, uri, HeaderMap, HeaderName, HeaderValue, Method, Request, Response,
+    StatusCode, Uri,
+};
+pub use wasi::http::types::ErrorCode;
+
+pub type Result<T, E = ErrorCode> = core::result::Result<T, E>;
+
+pub type IncomingRequest = Request<IncomingBody>;
+
+impl crate::From<Method> for wasi::http::types::Method {
+    fn from(method: Method) -> Self {
+        match method.as_str() {
+            "GET" => Self::Get,
+            "HEAD" => Self::Head,
+            "POST" => Self::Post,
+            "PUT" => Self::Put,
+            "DELETE" => Self::Delete,
+            "CONNECT" => Self::Connect,
+            "OPTIONS" => Self::Options,
+            "TRACE" => Self::Trace,
+            "PATCH" => Self::Patch,
+            _ => Self::Other(method.to_string()),
+        }
+    }
+}
+
+impl crate::TryFrom<wasi::http::types::Method> for Method {
+    type Error = method::InvalidMethod;
+
+    fn try_from(method: wasi::http::types::Method) -> Result<Self, Self::Error> {
+        match method {
+            wasi::http::types::Method::Get => Ok(Self::GET),
+            wasi::http::types::Method::Head => Ok(Self::HEAD),
+            wasi::http::types::Method::Post => Ok(Self::POST),
+            wasi::http::types::Method::Put => Ok(Self::PUT),
+            wasi::http::types::Method::Delete => Ok(Self::DELETE),
+            wasi::http::types::Method::Connect => Ok(Self::CONNECT),
+            wasi::http::types::Method::Options => Ok(Self::OPTIONS),
+            wasi::http::types::Method::Trace => Ok(Self::TRACE),
+            wasi::http::types::Method::Patch => Ok(Self::PATCH),
+            wasi::http::types::Method::Other(method) => method.parse(),
+        }
+    }
+}
+
+impl crate::From<uri::Scheme> for wasi::http::types::Scheme {
+    fn from(scheme: uri::Scheme) -> Self {
+        match scheme.as_str() {
+            "http" => Self::Http,
+            "https" => Self::Https,
+            _ => Self::Other(scheme.to_string()),
+        }
+    }
+}
+
+impl crate::TryFrom<wasi::http::types::Scheme> for http::uri::Scheme {
+    type Error = uri::InvalidUri;
+
+    fn try_from(scheme: wasi::http::types::Scheme) -> Result<Self, Self::Error> {
+        match scheme {
+            wasi::http::types::Scheme::Http => Ok(Self::HTTP),
+            wasi::http::types::Scheme::Https => Ok(Self::HTTPS),
+            wasi::http::types::Scheme::Other(scheme) => scheme.parse(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum FieldsToHeaderMapError {
+    InvalidHeaderName(header::InvalidHeaderName),
+    InvalidHeaderValue(header::InvalidHeaderValue),
+}
+
+impl Display for FieldsToHeaderMapError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            FieldsToHeaderMapError::InvalidHeaderName(e) => write!(f, "invalid header name: {e}"),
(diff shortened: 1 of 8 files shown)


crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,570 @@
...
+//! struct Component;
+//!
+//! http::export!(Component);
+//!
+//! // Implementing the [`Server`] trait for a component
+//! impl http::Server for Component {
+//!     fn handle(_request: http::IncomingRequest) -> http::Result<http::Response<impl http::OutgoingBody>> {
+//!         Ok(http::Response::new("Hello from Rust!"))
+//!     }
+//! }
```
brooksmtownsend, line 18:
```suggestion
//! impl http::Server for Component {
//!     fn handle(_request: http::IncomingRequest) -> Result<Response<impl OutgoingBody>, ErrorCode> {
//!         Ok(http::Response::new("Hello from Rust!"))
//!     }
//! }
```

This currently matches the trait, but I think that the doc comment here indicates the desired behavior which is to return the `http::Response` instead of a `std::result::Result<http::Response<>, ErrorCode>`

Basically, I think this can stay the same if we just change the trait

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,570 @@
...
+        drop(stream);
+        wasi::http::types::OutgoingBody::finish(body, None)
+            .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))
+    }
+}
+
+pub struct IncomingBody {
+    body: wasi::http::types::IncomingBody,
+    stream: InputStream,
+}
```
brooksmtownsend, line 332:
```suggestion
pub struct IncomingBody {
    stream: InputStream,
    body: wasi::http::types::IncomingBody,
}
```

Because of the resource hierarchy, the stream must be dropped before the body or else wasmtime will panic. This is a simple change that ensures this happens (since TIL struct variables are dropped in declaration order) but we'll either need a comment to indicate this or just implement `Drop` to avoid the future footgun

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,570 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.stream
+    }
+}
+
+impl From<IncomingBody> for InputStream {
+    fn from(IncomingBody { stream, .. }: IncomingBody) -> Self {
+        stream
+    }
+}
```
brooksmtownsend, line 374: Sadly I don't think we can implement this, because dropping the body before the input stream will cause a wasmtime panic

crates/component/src/wrappers/http.rs
```diff
@@ -0,0 +1,570 @@
...
+                Err(StreamError::LastOperationFailed(err)) => {
+                    return Err(std::io::Error::new(std::io::ErrorKind::Other, err));
+                }
+            }
+        }
+        let trailers = self
+            .into_trailers_wasi()
+            .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+        wasi::http::types::OutgoingBody::finish(body, trailers)
+            .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))
```
brooksmtownsend, line 308:
```suggestion
        }
        let trailers = self
            .into_trailers_wasi()
            .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
        drop(stream);
        wasi::http::types::OutgoingBody::finish(body, trailers)
            .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))
```

Make sure to drop the stream before finishing body here (other `write()` impls all do this)

brooksmtownsend (changes requested):
I tested this out with two simple HTTP server components (in wasmtime), the one that was indicated in the doc comment (which was close! and:
```rust
use wasi::http::types::ErrorCode;
use wasmcloud_component::http::{IncomingRequest, OutgoingBody, Response, Server, StatusCode};

struct Component;

wasmcloud_component::http::export!(Component);

// Implementing the [`Server`] trait for a component
impl Server for Component {
    fn handle(req: IncomingRequest) -> Result<Response<impl OutgoingBody>, ErrorCode> {
        let mut builder = Response::builder()
            .header("Foo", "Bar")
            .status(StatusCode::OK);

        if req.headers().contains_key("Another-Header") {
            builder = builder.header("Another-Header", "Ack");
        }

        builder
            .body("hey from Rust!")
            .map_err(|_| ErrorCode::InternalError(Some("Failed to build response".into())))
    }
}
```

I think once we change the trait to use `http::Response` and change the drop order for `IncomingBody`, this is working well! Streaming the request body to the response worked great for about a gig file, so that's good.

brooksmtownsend:
For additional fun information, this works great w/ blobstore too:
```rust
use wasi::http::types::ErrorCode;
use wasmcloud_component::{
    http::{IncomingRequest, OutgoingBody, Response, Server, StatusCode},
    wasi::blobstore::types::IncomingValue,
};

struct Component;

wasmcloud_component::http::export!(Component);

// Implementing the [`Server`] trait for a component
impl Server for Component {
    fn handle(_req: IncomingRequest) -> Result<Response<impl OutgoingBody>, ErrorCode> {
        let builder = Response::builder()
            .header("Foo", "Bar")
            .status(StatusCode::OK);

        let container =
            wasmcloud_component::wasi::blobstore::blobstore::get_container(&"stuff".to_string())
                .map_err(|_| {
                    ErrorCode::InternalError(Some("failed to get container".to_string()))
                })?;
        let size = container
            .object_info(&"wow.mp4".to_string())
            .map_err(|_| ErrorCode::InternalError(Some("failed to get object info".to_string())))?
            .size;
        let blob = container
            .get_data(&"wow.mp4".to_string(), 0, size)
            .map_err(|_| ErrorCode::InternalError(Some("failed to get object data".to_string())))?;
        let stream = IncomingValue::incoming_value_consume_async(blob).map_err(|_| {
            ErrorCode::InternalError(Some("failed to consume incoming value".to_string()))
        })?;

        builder
            .body(stream)
            .map_err(|_| ErrorCode::InternalError(Some("Failed to build response".into())))
    }
}
```

rvolosatovs:
> I tested this out with two simple HTTP server components (in wasmtime), the one that was indicated in the doc comment (which was close! and:
> 
> ```rust
> use wasi::http::types::ErrorCode;
> use wasmcloud_component::http::{IncomingRequest, OutgoingBody, Response, Server, StatusCode};
> 
> struct Component;
> 
> wasmcloud_component::http::export!(Component);
> 
> // Implementing the [`Server`] trait for a component
> impl Server for Component {
>     fn handle(req: IncomingRequest) -> Result<Response<impl OutgoingBody>, ErrorCode> {
>         let mut builder = Response::builder()
>             .header("Foo", "Bar")
>             .status(StatusCode::OK);
> 
>         if req.headers().contains_key("Another-Header") {
>             builder = builder.header("Another-Header", "Ack");
>         }
> 
>         builder
>             .body("hey from Rust!")
>             .map_err(|_| ErrorCode::InternalError(Some("Failed to build response".into())))
>         // Stream req body into response
>         //builder
>         //    .body(req.into_body())
>         //    .map_err(|_| ErrorCode::InternalError(Some("Failed to build response".into())))
>     }
> }
> ```
> 
> I think once we change the trait to use `http::Response` and change the drop order for `IncomingBody`, this is working well! Streaming the request body to the response worked great for about a gig file, so that's good.

We currently already use `http::Response` and doc comment compiles as-is, so I'm not sure what's suggested change here in that regard

crates/component/src/wrappers/http.rs
brooksmtownsend: @rvolosatovs looking at the trait, I see
```rust
fn handle(request: IncomingRequest) -> Result<Response<impl OutgoingBody>, ErrorCode>;
```

But the doc comment suggests the trait is:

```rust
fn handle(request: IncomingRequest) -> http::Response<impl OutgoingBody>;
```

so I was just suggesting changing the trait response type to match the doc comment

rvolosatovs: It's `http::Result<http::Response<impl http::OutgoingBody>>` in the doc actually, where `http::Result` is exported by this crate

brooksmtownsend: Oh! I see. Do you think we should change the `http::Result` type to be the following for compatibility? e.g. `builder.body(())` returns the `::http::Result<>` type.
```
pub type Result<T, E = http::Error> = core::result::Result<T, E>;
```

brooksmtownsend: > because [http::Error](https://docs.rs/http/latest/http/struct.Error.html) cannot be constructed from outside the http crate

Welp, that would make this more difficult then 😓 Resolving this then, I think we're in a good place then!

brooksmtownsend:
> We currently already use http::Response and doc comment compiles as-is, so I'm not sure what's suggested change here in that regard

We discussed in that specific comment, but all resolved there 👍🏻

brooksmtownsend (approved):
LGTM 🚀 I'm good to release this as `0.2.0` if we want to now, or if there's other functionality we should wait for let me know @rvolosatovs

PR #3456, opened 2024-10-25
fix(example): update dependency

peter-jerry-ye:
## Feature or Problem

The current example of custom template won't build because the wit dependency is locked at a version where `wasi:logging` didn't have a version. This PR updates the `deps.toml` to fix the issue.

This PR also updates the `wasmcloud-provider-sdk` versions and add initialization of observability as an example.

### Manual Verification

Manually verified that this compiles and runs.


vados-cosmonic (approved):
LGTM :rocket: 

Thanks, this is excellent @peter-jerry-ye !

vados-cosmonic:
Ah, would you mind signing your commits to fix the [DCO requirement](https://github.com/wasmCloud/wasmCloud/pull/3456/checks?check_run_id=32059780468)?

You can also make it a default [in your git config w/ the `format.signOff` option](https://git-scm.com/docs/git-config)

In the meantime I'll re-run CI (DCO check will fail again but others shouldn't!)

brooksmtownsend:
Looks like this PR is looking great but just needs a rebase to bring in the fixes for Docker/oci builds, thank you @peter-jerry-ye for this fix!

vados-cosmonic (approved):
Thanks for fixing DCO (and checking the "allow maintainers to edit" box)! Just rebased, and we should be ready to :rocket:

PR #3457, opened 2024-10-25
feat(wash-lib): suppress stderr output of tinygo generate

vados-cosmonic:

### DIFF
--- a/crates/wash-lib/src/build/component.rs
+++ b/crates/wash-lib/src/build/component.rs
@@ -434,7 +434,11 @@ async fn generate_tinygo_bindgen(project_dir: impl AsRef<Path>) -> Result<()> {
     let mut command = tokio::process::Command::new("go");
     let result = command
         .args(vec!["generate", "./..."])
-        .status()
+        // NOTE: this can be removed once upstream merges verbose flag
+        // https://github.com/bytecodealliance/wasm-tools-go/pull/214
+        .stderr(std::process::Stdio::piped())
+        .stdout(std::process::Stdio::piped())
+        .output()
         .await
         .map_err(|e| {
             if e.kind() == ErrorKind::NotFound {
@@ -444,8 +448,11 @@ async fn generate_tinygo_bindgen(project_dir: impl AsRef<Path>) -> Result<()> {
             }
         })?;
 
-    if !result.success() {
-        bail!("go generate failed: {}", result.to_string())
+    if !result.status.success() {
+        let stdout_output = String::from_utf8_lossy(&result.stdout);
+        let stderr_output = String::from_utf8_lossy(&result.stderr);
+        eprintln!("STDOUT:\n{stdout_output}\nSTDERR:\n{stderr_output}");
+        bail!("go generate failed: {}", result.status.to_string())
     }
 
     Ok(())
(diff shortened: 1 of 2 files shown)


crates/wash-lib/src/build/component.rs
```diff
@@ -434,6 +434,9 @@ async fn generate_tinygo_bindgen(project_dir: impl AsRef<Path>) -> Result<()> {
     let mut command = tokio::process::Command::new("go");
     let result = command
         .args(vec!["generate", "./..."])
+        // NOTE: this can be removed once upstream merges verbose flag
+        // https://github.com/bytecodealliance/wasm-tools-go/pull/214
```
joonas, line 438: Would it be better to capture the output in a buffer and print it only if the `go generate` fails?

crates/wash-lib/src/build/provider.rs
```diff
@@ -189,6 +189,9 @@ fn build_go_provider(
     if !go_config.disable_go_generate {
         let result = generate_command
             .args(["generate", "./..."])
+            // NOTE: this can be removed once upstream merges verbose flag
+            // https://github.com/bytecodealliance/wasm-tools-go/pull/214
+            .stderr(std::process::Stdio::null())
```
joonas, line 194: Ditto comment as the component above

joonas:
I think rather than throwing away the context regardless of what upstream flags are used, it would be better to save the output in a buffer and print it if the build fails

vados-cosmonic:
Ahhh you're absolutely right @joonas -- this is actually what we originally discussed doing, but I forgot when it came to implementation time :) Will add!

crates/wash-lib/src/build/component.rs
vados-cosmonic: Definitely, will add this!

crates/wash-lib/src/build/provider.rs
vados-cosmonic: Done!

crates/wash-lib/src/build/component.rs
```diff
@@ -434,7 +434,11 @@ async fn generate_tinygo_bindgen(project_dir: impl AsRef<Path>) -> Result<()> {
     let mut command = tokio::process::Command::new("go");
     let result = command
         .args(vec!["generate", "./..."])
-        .status()
+        // NOTE: this can be removed once upstream merges verbose flag
+        // https://github.com/bytecodealliance/wasm-tools-go/pull/214
+        .stderr(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
```
joonas, line 440: Should these be `stderr` and `stdout` instead of `stderr` twice?
```suggestion
        .stderr(std::process::Stdio::piped())
        .stdout(std::process::Stdio::piped())
```

vados-cosmonic: Thanks for this, was definitely a typo!

joonas (approved):

joonas:
Doesn't look like it likes my e-mail for suggestions, feel free to rewrite that suggestion to take my e-mail off of it so it doesn't trigger DCO :)

brooksmtownsend (approved):

PR #3461, opened 2024-10-25
feat(wash-cli): Add support for monolithic push to `wash push`

joonas:
## Feature or Problem

This adds support for pushing blobs monolithically to `wash push` as an alternative to chunked push.

In my testing I've found that some registries prefer or even require monolithic push as opposed to chunked push, so having this available in `wash` seems important to improve compatibility with the number of registries we can work with.


crates/wash-lib/src/cli/registry.rs
```diff
@@ -99,4 +99,8 @@ pub struct RegistryPushCommand {
 
     #[clap(flatten)]
     pub opts: AuthOpts,
+
+    /// Push the artifact monolithically instead of chunked
+    #[clap(long = "monolithic-push", env = "WASH_MONOLITHIC_PUSH")]
```
joonas, line 104: I'm happy to word this differently if anyone has suggestions on alternative naming for this option, I went with what felt obvious, but it doesn't necessarily convey the intent in the best possible way (not that anyone should have to know the difference really).

crates/wash-cli/src/common/registry_cmd.rs
```diff
@@ -133,6 +133,7 @@ pub async fn registry_push(
                 || project_config.is_some_and(|c| c.common.registry.push_insecure),
             insecure_skip_tls_verify: cmd.opts.insecure_skip_tls_verify,
             annotations,
+            use_monolithic_push: cmd.use_monolithic_push,
```
vados-cosmonic, line 136: NIT: wonder if the `use_` is necessary here -- I'm always one for long meaningful/easy to understand names but... I think `monolithic_push` might be fine

crates/wash-lib/src/cli/registry.rs
```diff
@@ -99,4 +99,8 @@ pub struct RegistryPushCommand {
 
     #[clap(flatten)]
     pub opts: AuthOpts,
+
+    /// Push the artifact monolithically instead of chunked
+    #[clap(long = "monolithic", env = "WASH_PUSH_MONOLITHIC")]
```
vados-cosmonic, line 104:
```suggestion
    #[clap(long = "monolithic-push", env = "WASH_MONOLITHIC_PUSH")]
```

I agree with the `WASH_SUBSECTION_*` pattern, but what made me feel this was weird was that the ENV var didn't exactly match the arg... IMO matching the arg is more important than subsection (mostly because we haven't been following that up until now)

vados-cosmonic (commented):
Had a few small nits!

crates/wash-cli/src/common/registry_cmd.rs
joonas: Fair, I've switched the internal flag to `monolithic_push`, should be fine either way I think, appreciate the feedback!

crates/wash-lib/src/cli/registry.rs
joonas: I was trying to avoid repetition in `wash push --monolithic-push` and `WASH_PUSH_MONOLITHIC` felt better than `WASH_MONOLITHIC_PUSH`, but I'm happy to go either way, appreciate the feedback!

PR #3462, opened 2024-10-25
chore(ci): Deprecate docker job

joonas:
## Feature or Problem

This job seems to be failing pretty consistently all of the sudden, and given that we're publishing [`oci` artifacts in a separate job](https://github.com/wasmCloud/wasmCloud/blob/main/.github/workflows/wasmcloud.yml#L403-L543), the job and the related `Dockerfile`s seem no longer necessary.


vados-cosmonic:
As far as getting CI to run properly, I'm not sure this is sufficient -- the `oci` build *also* fails -- it's just that `docker` fails so that the `oci` build isn't run.

Also, I'll let other people chime in here but I think there was a reason why the OCI and docker builds were separate/both needed.

rvolosatovs (approved):

rvolosatovs:
Images produced by these Dockerfiles were never used by us, they were only provided as developer tooling.
Does not look like anyone is using them.

ricochet (approved):

brooksmtownsend (approved):

PR #3463, opened 2024-10-25
fix(core)!: Fixes issue with provider cache invalidation

thomastaylor312:
So it turns out that our digest check and invalidation was working perfectly, but changes to a provider weren't working properly. This was due to the _separate_ caching step for the actual extracted binary. To fix this, I made the oci loader return an enum indicating whether it was a cache hit or miss and then converted that over to do the right thing when loading the par

Also, I bumped the version to 0.14 of wasmcloud-core because that is the version that is released on crates.io. I don't know why it ended up that way, but I wanted to make sure things reflected reality

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8105,7 +8105,7 @@ dependencies = [
  "wasm-pkg-client",
  "wasm-pkg-core",
  "wasmcloud-control-interface 2.2.0",
- "wasmcloud-core 0.13.0",
+ "wasmcloud-core 0.14.0",
  "wasmcloud-provider-sdk",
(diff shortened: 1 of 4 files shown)


vados-cosmonic (approved):
LGTM :rocket:

Would love a regression test here, but approving for :rocket:

brooksmtownsend (approved):

thomastaylor312:
> Would love a regression test here

This was a lot of effort to do (set up registry, push something, deploy something, test output, push new thing, delete and redeploy, test output again). So I just manually tested for now. We should probably scaffold some better ways to test out this type of functionality

Issue #3455, opened 2024-10-25
[FEATURE] pass tlsfirst parameter for nats client


peter-jerry-ye:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

I'm frustrated when I have a TLS terminating proxy for my cluster. I can't connect to it unless I pass the `--tlsfirst` parameter to nats client. As a result, I can't connect to my cluster with `wash`.

### Describe the solution you'd like

I would like to see add a `--ctl-tlsfirst` parameter that will be passed to the `ConnectOptions`

brooksmtownsend:
Interesting, I wasn't aware this was an option. This is a great first issue 👍🏻

brooksmtownsend:
Solved by #3473

PR #3473, opened 2024-10-28
feat: tls-first option for command line

peter-jerry-ye:
## Feature or Problem

Add a `--ctl-tls-first` option for `wash` which is then passed to nats client.

## Related Issues

#3455 

## Release Information

next

## Consumer Impact

There should be none. I made the `WashContext` having `Option<bool>` so it should be backward compatible with the existing context files, etc.

### Manual Verification

Manually verified that I could connect to my cluster with tls-terminating proxy.


crates/wash-cli/src/ctx/mod.rs
```diff
@@ -317,6 +317,19 @@ fn prompt_for_context() -> Result<WashContext> {
...
+            choices: Some(vec!["true".to_string(), "false".to_string()]),
+            regex: None,
+        },
+        "Should the control interface use TLS first?",
+    ) {
+        Ok(0) => Some(true),
+        Ok(1) => Some(false),
+        _ => None,
+    };
+
```
vados-cosmonic, line 332: Would it make sense to try to detect whether the user is trying to use TLS before asking this? It's a bit difficult because there are many different ways they could express wanting to use TLS, but it would be nice if we could make this question optional.

vados-cosmonic, line 332: Would it make sense to try to detect whether the user is trying to use TLS before asking this? It's a bit difficult because there are many different ways they could express wanting to use TLS, but it would be nice if we could make this question optional.

crates/wash-cli/src/down/mod.rs
```diff
@@ -111,6 +116,7 @@ pub async fn handle_down(cmd: DownCommand, output_kind: OutputKind) -> Result<Co
         cmd.ctl_seed,
         cmd.ctl_credsfile,
         cmd.ctl_tls_ca_file,
+        cmd.ctl_tls_first.unwrap_or(false),
```
vados-cosmonic, line 119:
```suggestion
        cmd.ctl_tls_first.unwrap_or_default(),
```

NIT: Feel free to make this change if you like/prefer it in other places as well!

vados-cosmonic (approved):
LGTM :rocket: 

Thanks for knocking this out @peter-jerry-ye !

crates/wash-cli/src/ctx/mod.rs
peter-jerry-ye: Should I judge based on the provided host url, e.g. see if it starts with `tls`?

vados-cosmonic: Yeah, so it's kind of tricky -- that, and whether the a custom CA file has been provided, from what I can see.

The perfect changeset might be to *ask* the person if their NATS connection uses TLS (and then do a whole bunch of other prompting), but that can be outside the scope of this change! Better to get some support in there and then improve it later if it's a stumbling block for people

vados-cosmonic: Hmnn, let's get this in here, and maybe re-think this in the future!

brooksmtownsend (approved):
LGTM, @peter-jerry-ye I think there's just a clippy nit to address

crates/wash-cli/src/ctx/mod.rs
peter-jerry-ye: Sorry, I've been a bit busy these days so I may not have time to address this very soon.

PR #3478, opened 2024-10-28
fix(washboard-ui): fast refresh for settings-context to resolve eslin…

markkovari:
…g warning

## Feature or Problem

Should fix the linting issue here: #3475

the failing gh action reference [here](https://github.com/wasmCloud/wasmCloud/actions/runs/11555819241/job/32162102484?pr=3475)


lachieh (approved):
thank you!

PR #3482, opened 2024-10-28
feat: add Github action workflow for benchmarking the host

protochron:
## Feature or Problem
Add a workflow that benchmarks the host using the Rust http-hello-world example. This should give us a decent benchmark of roughly how many requests per second the host can handle using the stock HTTP server provider without tuning the OS.

We may want to use a larger runner to get better baseline performance, since this is a fairly demanding test in terms of CPU usage.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,74 @@
+name: run-benchmarks
+on:
+  workflow_dispatch:
+    inputs:
+      sha:
+        description: 'SHA of the commit to benchmark'
+        required: true
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  http-benchmark:
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      packages: read
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+      - name: Install k6
+        run: |
+          sudo gpg -k
+          sudo gpg --no-default-keyring --keyring /usr/share/keyrings/k6-archive-keyring.gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys C5AD17C747E3415A3642D57D77C6C491D6AC1D69
+          echo "deb [signed-by=/usr/share/keyrings/k6-archive-keyring.gpg] https://dl.k6.io/deb stable main" | sudo tee /etc/apt/sources.list.d/k6.list
+          sudo apt-get update
+          sudo apt-get install k6
+
+      - name: Start services
+        run: |
+          sudo docker compose -f benches/http-loadtesting/docker-compose.yml up -d
+      - name:  Install wash
+        run: |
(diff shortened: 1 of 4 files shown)


protochron:
Marking this as a draft for now since I'm not sure how we want to trigger this workflow or if it should actually be integrated into the main build on every merge to main.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,74 @@
...
+          sudo apt-get update
+          sudo apt-get install k6
+
+      - name: Start services
+        run: |
+          sudo docker compose -f benches/http-loadtesting/docker-compose.yml up -d
+      - name:  Install wash
+        run: |
+          curl -s https://packagecloud.io/install/repositories/wasmcloud/core/script.deb.sh | sudo bash
+          sudo apt-get install wash
```
brooksmtownsend, line 36:
```suggestion
      - name:  Install wash
        uses: wasmcloud/common-actions/install-wash@main
```

supernit, this is a little action we have prepared

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Benchmark
+        run: |
+          wash get hosts
+          wash app deploy benches/http-loadtesting/wadm.yaml
+          wash app status rust-hello-world -o json
+          #./artifacts/1/wasmcloud-x86_64-unknown-linux-musl --log-level error &
```
brooksmtownsend, line 43: nit: looks like a testing leftover

benches/http-loadtesting/script.js
```diff
@@ -0,0 +1,65 @@
...
+    { duration: '5s', target: 1000 },
+    { duration: '5s', target: 10000},
+    { duration: '60s', target: 10000},
+    { duration: '5s', target: 0}
+
+  ]
+
+  // The following section contains configuration options for execution of this
+  // test script in Grafana Cloud.
+  //
```
brooksmtownsend, line 19: nit: probably fine to remove the extra configuration from the script

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,74 @@
...
+          sudo docker run -d --name wasmcloud --network=host ghcr.io/wasmcloud/wasmcloud:${{ github.event.inputs.sha }} wasmcloud --log-level error
+          timeout=30
+          while :; do
+            if [ $timeout -eq 0 ]; then
+              echo "Timeout waiting for app to deploy"
+              exit 1
+            fi
+            wash get hosts -o json
+            res=$(wash app status rust-hello-world -o json | jq -r '.status.status.type')
```
brooksmtownsend, line 52: we should have a `wash app deploy --wait` 😕 (don't tell Taylor)

brooksmtownsend (approved):
Only nits and supernits, feel free to ignore the suggestion to use the "common action" since we don't use it in other places

thomastaylor312:
This seems related to #3494. Do we want to keep this around or close in favor of the full benchmarking suite?

Issue #3258, opened 2024-10-09
[FEATURE] Use clap_autocomplete for control interface + app commands


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

I find it pretty frustrating that when using `wash` I can't tab autocomplete to fill in easily searchable values. For example, if I want to undeploy an application named `rust-hello-world`, and I have another app called `foobar`, I can't do the following to avoid typing in the whole app name (or realistically, copy/pasting it):
```
wash app undeploy rust<TAB>
```

Same thing with referencing control interface values:
```
wash stop component foo<TAB>
wash stop host NASD<TAB>
```

### Describe the solution you'd like

The [clap_autocomplete](https://docs.rs/clap_autocomplete/latest/clap_autocomplete/) crate is a great way to do this since we already use clap. In response to a <TAB> event for an argument where there is an existing entity in the lattice, we should be able to query for it.

For example, after typing `wash stop host <TAB>`, we should be able to use the clap autocomplete hook to query the lattice for hosts, and return a set of host IDs.

### Describe alternatives you've considered

There are many different commands this could be used for in `wash`, but to start the ctl + app subcommands will be the most impactful.

### Additional context

Note that we already have autocompletion for the wash subcommands, I specifically want to add the feature here to autocomplete for arguments about existing resources in a wasmCloud cluster. The old command autocomplete should still work after we do this.

nm723:
Hi @brooksmtownsend !
I'd like to work on this issue, could you assign it to me?

lachieh:
Thanks, @nm723!

Do you have an estimate as to when you would have this work ready for some initial feedback? Note that this is not a hard rule—we just like to get rough estimates up front so we can be helpful with checkins and scheduling.

nm723:
@lachieh 
Thanks for assigning it to me.
I think it will take about a week, but I'm not totally sure yet. I'll let you know if there are any changes.

brooksmtownsend:
Thanks @nm723 ! Please feel free to open a draft PR for initial feedback anytime 😄 Especially if you have questions

nm723:
[...]

PR #3495, opened 2024-10-29
[WIP] feat(wash-cli): enable autocomplete for app name

nm723:
## Feature or Problem

For the `app` command, pressing Tab now retrieves the app list and enables autocomplete.

- As noted in [this comment](https://github.com/wasmCloud/wasmCloud/issues/3258#issuecomment-2426003650), I'm using the`unstable-dynamic` feature from `clap_complete`.
  - For this reason, instead of using the `complete` command, we enable completions by running something like `echo "source <(COMPLETE=zsh your_program)" >> ~/.zshrc`. ([docs](https://docs.rs/clap_complete/latest/clap_complete/env/index.html))
- To begin with, I have enabled completions for the `app`’s `undeploy` and `history` commands.
- Currently, everything is written in `wash.rs`, but I think it would be better to move these to other files and would like your thoughts on it. For example, the `app` command things could be located in `wash-cli/src/app/mod.rs`.
  - However, one challenge is that I’m using `CliConnectionOpts` to fetch the app list for suggestions, which involves a `match` with `Cli::command()`. This setup is preventing `Cli` struct from being accessed from other files, and I’m seeking advice on addressing this (`fn get_connection_opts()`).

I would appreciate any feedback or advice for improvements!

## Related Issues

#3258

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1477,6 +1477,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ac2e663e3e3bed2d32d065a8404024dad306e699a04263ec59919529f803aee9"
 dependencies = [
  "clap",
+ "clap_lex",
+ "is_executable",
+ "shlex",
 ]
 
 [[package]]
@@ -3627,6 +3630,15 @@ name = "ipnet"
 version = "2.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ddc24109865250148c2e0f3d25d4f0f479571723792d3802153c60922a4fb708"
+
+[[package]]
+name = "is_executable"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d4a1b5bad6f9072935961dfbf1cced2f3d129963d091b6f69f007fe04e758ae2"
+dependencies = [
+ "winapi",
+]
 
 [[package]]
 name = "is_terminal_polyfill"
--- a/crates/wash-cli/Cargo.toml
+++ b/crates/wash-cli/Cargo.toml
@@ -28,7 +28,7 @@ clap = { workspace = true, features = [
     "std",
     "string",
 ] }
-clap_complete = { workspace = true }
+clap_complete = { workspace = true, features = ["unstable-dynamic"] }
 clap-markdown = { workspace = true }
 console = { workspace = true }
 crossterm = { workspace = true, features = ["events", "windows"] }
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -67,11 +67,11 @@ pub struct ListCommand {
 #[derive(Args, Debug, Clone)]
 pub struct UndeployCommand {
     /// Name of the application to undeploy
-    #[clap(name = "name", required_unless_present("all"))]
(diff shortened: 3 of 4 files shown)


brooksmtownsend:
Hey @nm723 , great find with the dynamic autocompletion feature! I knew we implemented completions, but I didn't know that was the only "stable" method of completions with `clap`. I'm not particularly concerned with using the unstable feature, provided we can back it up with some integration tests and ensure that we have nice safe bail-outs when tab completion doesn't work.

I have requests for how it's currently implemented here, but they all are in the same area of "let's use error handling instead of unwraps" and I figure those are just artifacts of the [WIP] nature of this PR. Thank you for filing the PR early for feedback!

All that being said, I have a couple thoughts, and I think this PR is headed in a great direction:
1. Just double checking, the `clap_complete` unstable feature does not require `nightly` Rust, right?
1. Since so many of these dynamic autocompletions will revolve around connecting to NATS and sending requests for applications / running resources, what do you think about getting the connection options once and then passing the cloned struct to all of the closures that run autocompletion queries? That way we don't have to keep referring back to parse the command
1. Just like how we proxy to the individual modules in `wash.rs`, I like the idea of keeping the `add_dynamic_candidates` function in `wash.rs` but then moving individual completions to their respective modules like you suggested with your comment about "For example, the app command things could be located in wash-cli/src/app/mod.rs"

nm723:
Thank you for your helpful advice @brooksmtownsend !
I’m going to improve the code with the comment.

1. I'm running this code on Rust version [1.82.0](https://releases.rs/docs/1.82.0/), so `nightly` isn't required.
2. If my understanding is incorrect, please let me know: If I type `wash app undeploy --ctl-port 3000 ru` and press Tab, the closure runs at that moment. When would the initial "getting" occur—during `wash up` and NATS starts?
  Additionally, I believe the completion closure cannot take arbitrary arguments. The only argument it can receive is the completion source string, like `ru` in `wash app undeploy ru<TAB>`.
(closure is passed to [ArgValueCompleter](https://docs.rs/clap_complete/latest/clap_complete/engine/struct.ArgValueCompleter.html) or [ArgValueCandidates](https://docs.rs/clap_complete/latest/clap_complete/engine/struct.ArgValueCandidates.html))
3. Got it!

brooksmtownsend:
Hey @nm723 just checking in on this one, how's the progress going? Do you have any questions / need any assistance?

nm723:
Sorry, I forgot about this. I’d appreciate it if you could answer question No.2 from my previous comment. 
Thank you for checking in and following up with me, @brooksmtownsend

lachieh:
@nm723 I'm not Brooks, so he can add more if necessary but I think the way you are thinking about it is correct. Since it executes once when <key>tab</key> is pressed, there isn't really any way keep the settings in memory.

It looks like you've covered the case where the connection isn't available which should just show no results but wanted to check. It'd also be great to add some tests but I don't think that should hold back the feature.

nm723:
@lachieh 
I apologize for the inconvenience, especially after receiving your help and answers to my questions. However, given my current situation, I'm unable to continue working on this . 
I'm very sorry, but I would like to request either that someone else takes over this work or that this PR be closed. Thank you for your understanding.

lachieh:
Just keeping this one up to date if anyone is able to pick this work up.

PR #3496, opened 2024-10-29
chore(examples): Modernize http-echo-tinygo example

joonas:


brooksmtownsend (approved):
169 lines -> 71 lines for the component? count me in

PR #3504, opened 2024-10-30
fix(wash-lib): canonicalize wit, build, override paths

brooksmtownsend:
## Feature or Problem
This PR fixes issues related to `wit`, `build` dirs and WIT deps override paths when using the `wash build -p <path/to/project>` argument by canonicalizing relative paths.

There's also some additional error context that I've added to the monkey patch function in case there were errors there to aid in tracking down the issue.

In the latest commit, I've pushed a fix to enable building components when a `wit` dir is not available, which is completely fine when a crate like `wasmcloud_component` already generated all of the bindings.

## Release Information
wash-lib 0.30.1

### DIFF
--- a/crates/wash-lib/src/parser/mod.rs
+++ b/crates/wash-lib/src/parser/mod.rs
@@ -12,9 +12,9 @@ use cargo_toml::{Manifest, Product};
 use config::Config;
 use semver::{Version, VersionReq};
 use serde::{Deserialize, Deserializer};
-use tracing::warn;
+use tracing::{trace, warn};
 use wadm_types::{Component, Properties, SecretSourceProperty};
-use wasm_pkg_core::config::Config as PackageConfig;
+use wasm_pkg_core::config::{Config as PackageConfig, Override};
 use wasmcloud_control_interface::RegistryCredential;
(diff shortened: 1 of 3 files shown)


crates/wash-lib/src/parser/mod.rs
```diff
@@ -1091,6 +1154,32 @@ impl ProjectConfig {
     }
 }
 
+/// Helper function to canonicalize a path or create it if it doesn't exist before
+/// attempting to canonicalize it. This is a nice helper to ensure that we can attempt
+/// to precreate directories like `build` before we start writing to them.
+fn canonicalize_or_create(path: PathBuf) -> Result<PathBuf> {
```
thomastaylor312, line 1160: _Technically_ we should probably be using the tokio one, but since this is a CLI we can probably get away with it

thomastaylor312 (approved):

crates/wash-lib/src/parser/mod.rs
brooksmtownsend: Oh true, I do call this from an async context so let me use that

PR #3508, opened 2024-10-31
fix(wash-cli): avoid an immediate reload

brooksmtownsend:
## Feature or Problem
After additional testing with `wash dev` in release mode, sometimes the developer loop can actually finish just a few milliseconds before the final build artifact writes out, triggering yet another reload. To combat this, I'm adding a a short wait (250ms) after the build and then taking any remaining value out of the channel.

This is a pretty unnoticeable time sleep in the dev loop, but it's immediately obvious that we're not getting duplicate reloads with this. 

## Release Information
wash 0.37.0

### DIFF
--- a/crates/wash-cli/src/cmd/dev/mod.rs
+++ b/crates/wash-cli/src/cmd/dev/mod.rs
@@ -282,6 +282,8 @@ pub async fn handle_command(
     }
     // Enable file watching
     pause_watch.store(false, Ordering::SeqCst);
+    // Make sure the reload channel is empty before starting the loop
+    let _ = reload_rx.try_recv();
 
     // Watch FS for changes and listen for Ctrl + C in tandem
     eprintln!(
@@ -296,8 +298,12 @@ pub async fn handle_command(
                 devloop::run(&mut run_loop_state)
                     .await
                     .context("failed to run dev loop iteration")?;
-                pause_watch.store(false, Ordering::SeqCst);
                 eprintln!("\n{} Watching for file changes (press Ctrl+c to stop)...", emoji::EYES);
+                // Avoid jitter with reloads by pausing the watcher for a short time
+                tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+                // Make sure that the reload channel is empty before unpausing the watcher
+                let _ = reload_rx.try_recv();
+                pause_watch.store(false, Ordering::SeqCst);
             },
 
             // Process a stop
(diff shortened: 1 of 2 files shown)


crates/wash-cli/src/cmd/dev/mod.rs
```diff
@@ -296,8 +298,12 @@ pub async fn handle_command(
                 devloop::run(&mut run_loop_state)
                     .await
                     .context("failed to run dev loop iteration")?;
-                pause_watch.store(false, Ordering::SeqCst);
                 eprintln!("\n{} Watching for file changes (press Ctrl+c to stop)...", emoji::EYES);
+                // Avoid jitter with reloads by pausing the watcher for a short time
+                tokio::time::sleep(std::time::Duration::from_millis(250)).await;
+                // Make sure that the reload channel is empty before unpausing the watcher
+                let _ = reload_rx.try_recv();
+                pause_watch.store(false, Ordering::SeqCst);
```
vados-cosmonic, line 306: I read the PR description but I'm not sure I get it -- is `devloop::run` finishing but the output artifact is actually not present yeton disk? If so can we instead wait for the artifact itself to be written/change?

brooksmtownsend: Yes that's correct, it's not always the output of our specific build though, sometimes this can be a change in the Rust target directory, or a written signed component for example. I have that trace log in there now for logging when a reload is triggered, and basically what I'd see is that the build finished, we'd start watching for changes, and then a build artifact could trigger the change based on a few ms of race condition.

I don't love the sleep but it does cover quite a few cases that we don't have to worry about ignoring right after a reload

vados-cosmonic: Well I'm all for making a fix that works for now at the very least, but maybe it makes sense to drop an issue to revisit and find a longer lasting fix.

IIRC `250ms` is almost at the boundary of human perception, it would be great to avoid this extra delay, but fixing the problem at least for now takes priority.

vados-cosmonic (approved):

crates/wash-cli/src/cmd/dev/mod.rs
brooksmtownsend: I could try a smaller number too, 100ms maybe

vados-cosmonic: Yeah, if that works it would be great!

brooksmtownsend: That worked just fine for me 😄 I think it really is like a 1ms race condition

PR #3510, opened 2024-10-31
feat: optimize communication

rvolosatovs:
## Feature or Problem

- Optimize HTTP server provider
   - disable TCP keepalive - we are quite chatty on the wire and generally these connections are short-lived, so no need for the extra overhead.
   - disable HTTP keepalive - http-server provider is a proxy, which establishes its own TCP connections to NATS server, so keepalive here actually slows down the performance by effectively blocking requests, instead strive for higher degree of concurrency by disabling keepalive and so achieving greater throughput
   - set TCP nodelay
- Update wRPC (with improved NATS transport)
- Optimize export serving by handling each invocation in a separate tokio task. In my testing, this performed much better

### Manual Verification

Tested on a Linux box with `Intel(R) Xeon(R) Platinum 8481C CPU @ 2.70GHz`

`nats-server 2.10.14`

#### IPv4 localhost

Old host, old HTTP server:
```
$ hey -n 1000 http://127.0.0.1:3002

Summary:
  Total:	0.9060 secs
  Slowest:	0.0593 secs
  Fastest:	0.0059 secs
  Average:	0.0448 secs
  Requests/sec:	1103.7220

$ hey -n 1000 --disable-keepalive http://127.0.0.1:3002

Summary:
  Total:	0.2737 secs
  Slowest:	0.0196 secs
  Fastest:	0.0027 secs
  Average:	0.0133 secs
  Requests/sec:	3653.6536
```

Old host, new HTTP server:
```
$ hey -n 1000 http://127.0.0.1:3002

Summary:
  Total:	0.2569 secs
  Slowest:	0.0156 secs
  Fastest:	0.0023 secs
  Average:	0.0125 secs
  Requests/sec:	3892.6621

$ hey -n 1000 --disable-keepalive http://127.0.0.1:3002

Summary:
  Total:	0.2535 secs
  Slowest:	0.0155 secs
  Fastest:	0.0024 secs
  Average:	0.0124 secs
  Requests/sec:	3945.3772
```

New host, old HTTP server:

```
$ hey -n 1000 http://127.0.0.1:3002
Summary:
  Total:	0.9076 secs
  Slowest:	0.0574 secs
  Fastest:	0.0068 secs
  Average:	0.0446 secs
  Requests/sec:	1101.7848

$ hey -n 1000 --disable-keepalive http://127.0.0.1:3002

Summary:
  Total:	0.2370 secs
  Slowest:	0.0196 secs
  Fastest:	0.0074 secs
  Average:	0.0116 secs
[...]

PR #3511, opened 2024-10-31
chore: update `wrpc-transport-nats`, reuse wRPC clients in HTTP server

rvolosatovs:
## Feature or Problem

- Update `wrpc-transport-nats` - this version reuses a single subscriber for inboxes
- Reuse wRPC clients in http-server instead of constructing on each request - this avoids a NATS roundtrip on each request
- Since I was already working on it, remove a few redundant allocations in the provider

Provider examples are updated and then the update is reverted in this PR, in the future, https://github.com/wasmCloud/wasmCloud/pull/3511/commits/63b1b86dd59d7b27503be3eff4709d2dcb55c14b can be reverted once `wasmcloud-component-sdk` version is updated

## Related Issues

Blocked on #3510 

### Manual Verification

```
hey -n 1000 http://127.0.0.1:3002

Summary:
  Total:	0.0918 secs
  Slowest:	0.0100 secs
  Fastest:	0.0006 secs
  Average:	0.0043 secs
  Requests/sec:	10890.2866
```

Note, in #3510 I was using `instances: 1` for the component, which I inherited from duplicating https://github.com/wasmCloud/wasmCloud/blob/98cd06c1dd448073d8eadef790ee7709a2ccae36/examples/rust/components/http-hello-world/wadm.yaml#L25

This was abysmal for performance - we should probably *not* do that to avoid users from accidentally making the same mistake

### DIFF
--- a/crates/host/src/wasmbus/handler.rs
+++ b/crates/host/src/wasmbus/handler.rs
@@ -221,14 +221,15 @@ impl wrpc_transport::Invoke for Handler {
         let mut headers = injector_to_headers(&TraceContextInjector::default_with_span());
         headers.insert("source-id", &*self.component_id);
         headers.insert("link-name", link_name);
-        wrpc_transport_nats::Client::new(
+        let nats = wrpc_transport_nats::Client::new(
             Arc::clone(&self.nats),
             format!("{}.{id}", &self.lattice),
             None,
         )
-        .timeout(self.invocation_timeout)
-        .invoke(Some(headers), instance, func, params, paths)
-        .await
+        .await?;
+        nats.timeout(self.invocation_timeout)
+            .invoke(Some(headers), instance, func, params, paths)
+            .await
     }
 }
 
--- a/crates/provider-keyvalue-nats/src/lib.rs
+++ b/crates/provider-keyvalue-nats/src/lib.rs
@@ -70,14 +70,12 @@ impl KvNatsProvider {
             .await
             .context("failed to run provider")?;
         let connection = get_connection();
-        serve_provider_exports(
-            &connection.get_wrpc_client(connection.provider_key()),
-            provider,
-            shutdown,
-            bindings::serve,
-        )
-        .await
-        .context("failed to serve provider exports")
+        let wrpc = connection
+            .get_wrpc_client(connection.provider_key())
+            .await?;
+        serve_provider_exports(&wrpc, provider, shutdown, bindings::serve)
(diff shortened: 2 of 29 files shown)


crates/host/src/wasmbus/handler.rs
```diff
@@ -221,14 +221,15 @@ impl wrpc_transport::Invoke for Handler {
...
             format!("{}.{id}", &self.lattice),
             None,
         )
-        .timeout(self.invocation_timeout)
-        .invoke(Some(headers), instance, func, params, paths)
-        .await
+        .await?;
+        nats.timeout(self.invocation_timeout)
+            .invoke(Some(headers), instance, func, params, paths)
+            .await
```
brooksmtownsend, line 232: Did this have a practical effect on the performance or is it just a slight refactor? Seems like awaiting + handling a result is new on the client

crates/provider-keyvalue-nats/src/lib.rs
```diff
@@ -70,14 +70,12 @@ impl KvNatsProvider {
...
-            &connection.get_wrpc_client(connection.provider_key()),
-            provider,
-            shutdown,
-            bindings::serve,
-        )
-        .await
-        .context("failed to serve provider exports")
+        let wrpc = connection
+            .get_wrpc_client(connection.provider_key())
+            .await?;
```
brooksmtownsend, line 75: Will be good to note in the next provider-sdk release that this is a breaking change + all you need to do is `.await?` the creation of the client

brooksmtownsend (approved):
I wasn't quite clear on why a commit in here could be reverted once the component SDK is released, perhaps that was a comment for `wasmcloud-provider-sdk`?

crates/host/src/wasmbus/handler.rs
rvolosatovs: `wrpc_transport_nats::Client::new` now subscribes on a shared inbox subject (ending with `>`) and starts a Tokio task, which demultiplexes it

crates/provider-keyvalue-nats/src/lib.rs
rvolosatovs: Yes, is there a changelog I should update?

brooksmtownsend: We autogenerate changelogs so I suppose not, We'll just have to remember ™️

PR #3521, opened 2024-11-01
feat(wash-cli): start secrets backend with wash dev

vados-cosmonic:

### DIFF
--- a/crates/wash-cli/src/cmd/dev/mod.rs
+++ b/crates/wash-cli/src/cmd/dev/mod.rs
@@ -5,10 +5,11 @@ use std::sync::Arc;
 use anyhow::{Context as _, Result};
 use clap::Parser;
 use console::style;
+use nkeys::XKey;
 use notify::event::ModifyKind;
 use notify::{event::EventKind, Event as NotifyEvent, RecursiveMode, Watcher};
 use semver::Version;
-use session::{SessionMetadata, WashDevSession};
+use session::{SessionMetadata, SessionProcesses, WashDevSession};
 use tokio::{select, sync::mpsc};
 
 use wash_lib::cli::{CommandOutput, CommonPackageArgs};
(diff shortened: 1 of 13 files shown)


brooksmtownsend:
Hey @vados-cosmonic ! Just wanted to touch base because in https://github.com/wasmCloud/wasmCloud/pull/3524 I added just a bit of code to make sure that the NATS KV policy was added to the manifest if any secrets were specified for components/providers (not for links yet, but I added a note for that)

Feel free to take my commit or let me know if my implementation is bad and I can reconcile with yours 😄

crates/wash-cli/src/cmd/dev/mod.rs
```diff
@@ -362,10 +430,17 @@ pub async fn handle_command(
...
+                    if let Some(mut nats) = nats {
                         eprintln!("{} Stopping NATS...", emoji::HOURGLASS_DRAINING);
                         nats.kill().await?;
                     }
+
+                    // Stop nats-kv-secrets
+                    if let Some(mut nats_kv_secrets) = nats_kv_secrets {
+                        eprintln!("{} Stopping nats-kv-secrets...", emoji::HOURGLASS_DRAINING);
+                        nats_kv_secrets.kill().await?;
+                    }
```
brooksmtownsend, line 442: What are your thoughts on just embedding the secrets-nats-kv as a library here, so we don't have to spawn off yet another process?

vados-cosmonic: Ah if we do that, IMO we should do that as a part of `wash` -- I think it's been a long time coming that `wash` should have been able to spawn the secrets binary and enable secrets support from the beginning! IIRC someone noted reasons against it, but IMO let's handle the integration as a separate thing and build it into starting `wash`!

Will happily rebase this PR to that if we get it done before this is ready

brooksmtownsend:
Hey @vados-cosmonic looks like this one got out of date, are we still planning on doing this for wash dev?

vados-cosmonic:
Closing this for now -- `wash` has seen too many changes to rebase this -- at this point it's easier to make a new PR with the changes that were supposed to land

PR #3524, opened 2024-11-01
feat(wash-cli): support interface wildcard overrides

brooksmtownsend:
## Feature or Problem
This PR supports specifying wildcard interfaces in the wasmcloud.toml to prevent multiple overrides for the same interface.

For example, to override KV nats with Redis:
```toml
[[dev.overrides.imports]]
interface_spec = "wasi:keyvalue@0.2.0-draft"
image_ref = "ghcr.io/wasmcloud/keyvalue-redis:0.28.2"
config = { values = { url="redis://127.0.0.1:6379" } }
```

This PR also started using config & secrets where they weren't being used before, so an integration test started failing. I added support for a secrets policy in the event that one is specified in the last commit.

### Acceptance or Integration
This PR initially broke an integration test related to overrides because it brought in secrets that were configured on the ovveride

### Manual Verification
I manually tested with the above

### DIFF
--- a/crates/wash-cli/src/cmd/dev/deps.rs
+++ b/crates/wash-cli/src/cmd/dev/deps.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::hash::Hash;
 use std::path::{Path, PathBuf};
 
@@ -277,7 +277,7 @@ impl DependencySpec {
                     wit: WitInterfaceSpec {
                         namespace: ns,
                         package: pkg,
-                        interface: Some(iface),
+                        interfaces: Some(HashSet::from([iface])),
                         function: None,
                         version,
                     },
@@ -293,7 +293,7 @@ impl DependencySpec {
                 wit: WitInterfaceSpec {
                     namespace: ns,
                     package: pkg,
-                    interface: Some(iface),
+                    interfaces: Some(HashSet::from([iface])),
                     function: None,
                     version,
                 },
@@ -309,7 +309,7 @@ impl DependencySpec {
                     wit: WitInterfaceSpec {
                         namespace: ns,
                         package: pkg,
-                        interface: Some(iface),
+                        interfaces: Some(HashSet::from([iface])),
                         function: None,
                         version,
                     },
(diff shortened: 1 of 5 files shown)


vados-cosmonic:
Hey so I was thinking that we wouldn't need to introduce a *new* syntax here -- `*` is not a valid iface name IIRC.

The code right now is set up to handle this -- coalescing version requirements is an explicit part of the process, and it's a chance for us to detect configurations that don't work/do slightly more complex things. 

I could have sworn I wrote about this somewhere but in my mind we *do* want to allow multiple overrides for the same interface, for example if someone has:

```
[[dev.overrides.imports]]
interface_spec = "wasi:keyvalue/*@0.2.0-draft"
image_ref = "ghcr.io/wasmcloud/keyvalue-redis:0.28.1"
config = { values = { url="redis://127.0.0.1:6379" } }

[[dev.overrides.imports]]
interface_spec = "wasi:keyvalue/store@0.2.0-draft"
image_ref = "ghcr.io/wasmcloud/keyvalue-local:0.28.1"
link_name = "local-kv"
config = { values = { backing_file="/tmp/some-file" } }
```

Possibly in the same component and in a workspace, they might want to use different kvs for different things actually.

brooksmtownsend:
> This was partly why I reworked the parsing of package names in core recently -- people should be able to use wasi:keyvalue only, and that will represent a package level override.

Oh okay awesome, I think this was a big blocker there. Is that fix already merged @vados-cosmonic ? If I can just input `wasi:keyvalue@0.2.0-draft`, that's even better

vados-cosmonic:
Hey yeah the fix *was* released last week! Here's [the new code in `core`](https://github.com/wasmCloud/wasmCloud/blob/main/crates/core/src/wit.rs#L189), and [the usage in the parsing stuff for `wash-lib`](https://github.com/wasmCloud/wasmCloud/blame/ff416c22bde82ac90ba980b7f1e5b7e5b0f8eb36/crates/wash-lib/src/parser/mod.rs#L523). 

The way this is supposed to get used is via [`WitInterfaceSpec`](https://github.com/wasmCloud/wasmCloud/blob/ff416c22bde82ac90ba980b7f1e5b7e5b0f8eb36/crates/wash-lib/src/parser/mod.rs#L428), which has methods like `is_disjoint()` and `includes()` to figure out if some interface is "covered" by another one (and then we can theoretically do the right thing at a higher level).

brooksmtownsend:
Amazing @vados-cosmonic , let me try this out!

brooksmtownsend:
@vados-cosmonic refactored to use the new API, looked great 🥳

brooksmtownsend:
Alright @vados-cosmonic finally got back and added that integration test! 😄

vados-cosmonic (approved):
LGTM :rocket:

PR #3535, opened 2024-11-04
fix(wash): Fixes wasmcloud.lock on windows

thomastaylor312:
This is a simple dependency bump to pick up the fix from wkg on locking files on windows


ricochet (approved):

brooksmtownsend (approved):
I _think_ there have been breaking changes to go in since 0.36.1 of wash, but I'll have to double check. We can cherry pick if this needs to go out ASAP

PR #3546, opened 2024-11-05
chore: convert `wasi:logging` levels into more conventional format

joonas:
## Feature or Problem

While I was diagnosing an issue, I noticed that the log levels that get logged from `wasi:logging` calls didn't look quite right:
```console
INFO log: wasmcloud_host::wasmbus::handler: ... level=Level::Info ...
```

what I was expecting instead was something like this:
```console
INFO log: wasmcloud_host::wasmbus::handler: ... level="info" ...
```

This implements `Display` and calls it via `to_string` in the logging handler.

### DIFF
--- a/crates/host/src/wasmbus/handler.rs
+++ b/crates/host/src/wasmbus/handler.rs
@@ -275,7 +275,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::TRACE,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
                     context,
                     "{message}"
                 );
@@ -284,7 +284,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::DEBUG,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
                     context,
                     "{message}"
                 );
@@ -293,7 +293,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::INFO,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
                     context,
                     "{message}"
                 );
@@ -302,7 +302,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::WARN,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
                     context,
                     "{message}"
                 );
@@ -311,7 +311,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::ERROR,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
                     context,
                     "{message}"
                 );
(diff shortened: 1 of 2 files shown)


crates/host/src/wasmbus/handler.rs
```diff
@@ -284,7 +284,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::DEBUG,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
```
brooksmtownsend, line 287:
```suggestion
                    %level,
```

crates/host/src/wasmbus/handler.rs
```diff
@@ -293,7 +293,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::INFO,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
```
brooksmtownsend, line 296:
```suggestion
                    %level,
```

crates/host/src/wasmbus/handler.rs
```diff
@@ -302,7 +302,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::WARN,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
```
brooksmtownsend, line 305:
```suggestion
                    %level,
```

crates/host/src/wasmbus/handler.rs
```diff
@@ -311,7 +311,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::ERROR,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
```
brooksmtownsend, line 314:
```suggestion
                    %level,
```

crates/host/src/wasmbus/handler.rs
```diff
@@ -320,7 +320,7 @@ impl Logging for Handler {
...
```
brooksmtownsend, line 323:
```suggestion
                    %level,
```

brooksmtownsend (changes requested):
One tiny request, but requesting changes since it will avoid a string allocation on each wasi:logging log!

crates/host/src/wasmbus/handler.rs
```diff
@@ -275,7 +275,7 @@ impl Logging for Handler {
                 tracing::event!(
                     tracing::Level::TRACE,
                     component_id = ?self.component_id,
-                    ?level,
+                    level = level.to_string(),
```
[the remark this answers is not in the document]
joonas: So I started there, but unfortunately when you don't go through `.to_string`, the level values end up without quotes, like so:
```console
level=info
```
as opposed to what we want, which is:
```console
level="info"
```

brooksmtownsend (approved):
After discussion offline, looks like the `%` removes the quotes and doesn't quite work as expected

PR #3547, opened 2024-11-05
chore(ci): cancel in flight duplicate pull_request workflows

joonas:
## Feature or Problem

Since we're already using `concurrency:` to cancel in-flight requests keyed by SHA, I thought it might be useful in PRs to cancel them by pull request number so as to only test the latest changes.

This was inspired by https://turso.tech/blog/simple-trick-to-save-environment-and-money-when-using-github-actions

### DIFF
--- a/.github/workflows/examples-components.yml
+++ b/.github/workflows/examples-components.yml
@@ -54,7 +54,7 @@ permissions:
   contents: read
 
 concurrency:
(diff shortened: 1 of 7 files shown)


brooksmtownsend (approved):
Would've never guessed this is how it works. Kudos @joonas !

PR #3548, opened 2024-11-05
chore(ci): Limit nix workflow to wasmcloud/wasmcloud

joonas:
## Feature or Problem

I only ever get `nix workflow run` emails where it says `nix: All jobs were cancelled`, so I'm not entirely sure it's worth running this outside of `wasmcloud/wasmcloud`

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -18,6 +18,7 @@ concurrency:
 
 jobs:
   fmt:
+    if: ${{ github.repository == 'wasmCloud/wasmCloud' }}
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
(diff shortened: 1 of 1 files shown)


rvolosatovs (commented):
I think the real issue here is that this workflow is configured to run on *every* push (as long as something outside `/typescript` is modified)

How about we only run this on pushes to `main` like we do here https://github.com/wasmCloud/wasmCloud/blob/5b0d8069da79096eabf3216d46263cff0011d0e7/.github/workflows/wasmcloud.yml#L32-L34 ?

joonas:
> How about we only run this on pushes to `main` like we do here

I'm happy to do that _as well_, but I don't think there's much point in running this for anyone who won't have these secrets set up, which I would have to imagine is most folks.

Even if we restrict it to `push` to `main`, it will still run when forked repositories run `Sync fork`.

PR #3559, opened 2024-11-07
refactor(examples): improve clarity of the TS quickstart

vados-cosmonic:
This commit removes the "slow start" documentation on the TS http-hello-world quickstart to make it easier to follow and more in line with the official quickstart in the docs.

We ensure here that the push to use `wash dev` is more consistent.


brooksmtownsend (approved):

vados-cosmonic:
Let's keep this for now -- it seems like it actually *did* do it's job, after the initial confusion...

> Slow start really helps to understand the concepts

PR #3569, opened 2024-11-08
feat: add TinyGo version of the http-password-checker

protochron:
## Feature or Problem
Implement a Go version of the http-password checker. It has a slightly different output format than the Rust one because it uses a much less feature rich library for assessing the password complexity.


examples/golang/components/http-password-checker/main.go
```diff
@@ -0,0 +1,79 @@
...
+	resp := CheckResponse{Valid: true, Length: len(req.Value)}
+	respJSON, err := json.Marshal(resp)
+	if err != nil {
+		errResponseJSON(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(respJSON)
+}
```
joonas, line 58: A slightly cleaner alternative might read something like this:
```suggestion
	w.Header().Set("Content-Type", "application/json")
	err = json.NewEncoder(w).Encode(resp)
	if err != nil {
		errResponseJSON(w, http.StatusInternalServerError, err.Error())
	}
}
```

examples/golang/components/http-password-checker/main.go
```diff
@@ -0,0 +1,79 @@
...
+func handleRequest(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		errResponseJSON(w, http.StatusMethodNotAllowed, "Method not allowed")
+		return
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/check", checkHandler)
+	mux.ServeHTTP(w, r)
+}
```
joonas, line 69: Technically you could get rid of `handleRequest` altogether by moving the check for MethodPost
```go
func checkHandler(w http.ResponseWriter, r *http.Request) {
	if r.Method != http.MethodPost {
		errResponseJSON(w, http.StatusMethodNotAllowed, "Method not allowed")
		return
	}
	var req CheckRequest
	defer r.Body.Close()
```
at the top of the `checkHandler` function, and moving the `mux` setup into `init()`:
```go
func init() {
	// Register the ServeMux as the handler for all incoming requests.
	mux := http.NewServeMux()
	mux.HandleFunc("/check", checkHandler)
	wasihttp.Handle(mux)
}
```

vados-cosmonic: You know what's really weird -- I didn't find that *any* of the muxes (`http.ServeMux`, `gorilla/mux`, etc) worked properly with `init` or the handler in this way (I tried this a few days ago working on stuff for WASMCON 2024).

I couldn't pin down *why* it wasn't working -- but the paths just wouldn't parse (the 404 handler would be the only thing returned) -- I stopped before descending into match machinery hell but I do want to note it's not as easy as it seems. Maybe you'll have more luck here.

joonas: It was fixed in TinyGo 0.34.0.

protochron: Yeah, I've run this code and it very much works

vados-cosmonic: Oh that's great, I'm running an old version of TinyGo it seems -- I'm on 0.33 which explains it!

Do we want to change this code to be that? Also while we're here it might be worth adding a test other than the default wadge one -- IIRC it just checks `/`

joonas (approved):

examples/golang/components/http-password-checker/wasmcloud.toml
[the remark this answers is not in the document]
protochron: Good idea. Added.

examples/golang/components/http-password-checker/README.md
```diff
@@ -0,0 +1,91 @@
...
+[wadm]: https://github.com/wasmCloud/wadm
+[tinygo]: https://tinygo.org/getting-started/install/
+[wash]:  https://wasmcloud.com/docs/ecosystem/wash/
+
+# Dependencies
+
+Before starting, ensure that you have the following installed:
+
+- The [TinyGo toolchain][tinygo]
+- [`wash`, the WebAssembly SHell][wash] installed.
```
markkovari, line 21: I presume this supposed to be rather
```suggestion
- [`wash`, the WAsmcloud SHell][wash] installed.
```

Which is a bit confusing since in the docs url it is 
"wasmCloud Shell" and in the CLI ascii art it is "wasmCloud Shell" 🤔

examples/golang/components/http-password-checker/README.md
```diff
@@ -0,0 +1,91 @@
...
+```console
+curl localhost:8000/check -d '{"value": "tes12345!"}'                                                                                         [23:01:58]
+```
+
+You should see a JSON response like:
+
+```json
+{
+    "valid": false, "message": "insecure password, try including more special characters, using uppercase letters or using a longer password"
+
```
markkovari, line 58: this line looks odd in the json snippet.
```suggestion
```

markkovari (commented):
Hey @protochron spotted a few nits 🤓
And thanks for more go examples 💪

examples/golang/components/http-password-checker/README.md
markkovari: Oh that is the place where it is called `WAsmcloud SHell` at the top. [docs](https://wasmcloud.com/docs/installation) I did remember seen another version of the wash abbreviation.

vados-cosmonic: Hey thanks for this @markkovari -- I put the change in!

PR #3570, opened 2024-11-08
release(wash-cli): v0.37.0

brooksmtownsend:
## Feature or Problem
This PR updates wash-cli for release. I manually ran this instead of the wash-cli smart release because we had an outdated provider-archive dependency which threw the tool for a loop.


vados-cosmonic (approved):
LGTM :rocket: 

I'd cut it down to just wash-lib & wash-cli if the dry run works locally!

brooksmtownsend:
@vados-cosmonic I actually think we may need a lot of the bumps here, the change to provider-archive triggered quite a few updates! But with our smart-release usage it would be nice to calm the bumps down where possible

thomastaylor312 (approved):

PR #3574, opened 2024-11-08
fix(host): Fixes orphaned traces

thomastaylor312:
We had the correct code for extracting tracing headers from wRPC calls, but our use of `in_current_span` was creating unintentionally orphaned spans. This adds an explicit span for each incoming invocation. See the comment added in the code for additional information


thomastaylor312:
I bumped the host crate and the host binary by a patch version so we can release this

rvolosatovs (approved):

PR #3588, opened 2024-11-11
fix(examples): rust component http-task-manager fix order and config …

markkovari:
## Problem

When I went through the rust/components/http-task-manager example: [link](https://github.com/wasmCloud/wasmCloud/tree/main/examples/rust/components/http-task-manager)

- I've seen that the config is called different in the `*.wadm.yaml`s compared to the README.md one and the readme suggests to add the config before starting the host. Which should be the other way around.
- Simplified one of the `curl` calls to make is less scary 😄 
- unified the config in the yaml - removed quotes
- bumped the http server version

### DIFF
--- a/examples/rust/components/http-task-manager/README.md
+++ b/examples/rust/components/http-task-manager/README.md
@@ -62,10 +62,23 @@ docker run \
     postgres:16-alpine
 ```
 
+### Start the application
+
+Now, we're ready to start our component along with the required providers.
+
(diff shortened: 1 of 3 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Sorry for the delay here @markkovari -- this is excellent, these improvements make the task manager *much* easier to use!

PR #3589, opened 2024-11-11
chore: Adding wash package for 'Fedora 41'

lxfontes:
https://packagecloud.io/wasmCloud/core is currently missing `fedora/41` packages.

From https://packagecloud.io/api/v1/distributions.json, the fedora 41 id is `302`

Screenshot 2024-11-11 at 12 44 03 PM

### DIFF
--- a/.github/workflows/wasmcloud.yml
+++ b/.github/workflows/wasmcloud.yml
@@ -33,62 +33,62 @@ on:
     branches:
       - main
     tags:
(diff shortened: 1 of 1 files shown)


flickerfly:
Thanks @lxfontes and @ericgregory

joonas (approved):

brooksmtownsend (commented):
@lxfontes any other fedora versions we should add while we're in here? Definitely should be releasing on the latest

Issue #3572, opened 2024-11-08
[FEATURE] Check for `~/.wash/downloads/nats-server` version and update if it's old


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

During `wash up` and `wash dev`, we start a local `nats-server` as a part of the local platform. Currently if that `nats-server` version is out of date, `wash` does not download the latest version.

### Describe the solution you'd like

Following exactly what we do for `wadm`, we should check the version of the `nats-server` binary and if it's not the latest version as specified in the wash up configuration.

For example if I ran this:
```
➜ ~/.wash/downloads/nats-server version
nats-server: v2.9.12
```

and wash is configured to download v2.10.22, on the next `wash up` or `wash dev` we should download that latest version and replace the binary.

vados-cosmonic:
+1 Excellent work @claytondus !

PR #3606, opened 2024-11-14
fix(wash-lib): update NATS if local is not the configured version

claytondus:
## Feature or Problem
Verify the local nats-server version is equal to the configured version, and download the correct version from Github if necessary.

## Related Issues
#3572


brooksmtownsend (approved):
Looks great @claytondus ! Thank you 🚀

PR #3610, opened 2024-11-14
fix(wash-cli): Fix Warn Timeout when deadline has elapsed.

saki007ster:
## Feature or Problem
This pull request addresses a timeout issue encountered when deleting application manifests. If the operation times out, a specific message indicating the host is not available is now logged, instead of failing silently or returning a generic error.

## Release Information
Target release: next

## Consumer Impact
This change does not impact API behavior or external consumers directly. It improves logging for timeout errors during manifest deletion, which helps in debugging and operational insights.

### Acceptance or Integration
No changes were made to the acceptance or integration test suite.

### Manual Verification
- Triggered a manifest deletion operation in an environment where a timeout was simulated.
- Verified that a timeout results in a logged message: “Host not available: Timeout occurred while deleting application [app-name].”
- Ensured other errors are still returned as usual.


brooksmtownsend (commented):
Nice work here @saki007ster ! Looks like this might not be passing a `cargo build`, would you mind taking a look and seeing if you're able to fix that locally?

Otherwise, I'm happy to wrap this up with an additional commit 😄

saki007ster:
I will work on this today and try to fix it.

brooksmtownsend:
Thank you @saki007ster, let me know once you get to a stopping point

PR #3611, opened 2024-11-14
chore(examples): use port 8000 instead of 8080 to align with

dorser:
## Feature or Problem
When using `wash dev` the default port is 8000.
When running through the quickstart, it'd be less confusing using the same port instead of 8080.

## Related Issues
https://github.com/wasmCloud/wasmcloud.com/pull/714.


vados-cosmonic (approved):
LGTM 🚀 

Thanks for this @dorser !

joonas (approved):

PR #3628, opened 2024-11-18
chore(all): removed unused dependencies

Mossaka:
A batch scanning all crates and remove unused dependencies by running 'cargo machete'. If you want, I could add `cargo machete` to the CI to detect any added unused deps.


joonas:
Hi @Mossaka,

Thanks for going through this list!

I realize this is a big ask, but to make this a bit easier to work through and get all of the appropriate reviews in, would you mind breaking this up to individual PRs focused on one affected crate at a time?

If it were broken down by affected crate, I think it might be easier to push these changes through and get them published as appropriate.

rvolosatovs (approved):
the changes here look great to me, however, looks like due to removing the unused dependencies, some of the dependencies we actually do use now are built with different feature sets causing the CI failure.

brooksmtownsend (commented):
Thank you @Mossaka ! I think @rvolosatovs may be right here, I know at least the `clap-markdown` feature is one we use to generate docs so we do use that. Happy to take this one the build+CI is passing though

Mossaka:
Oh thanks, sorry for the late reponse. I will keep working on it to pass the CI!

thomastaylor312 (approved):

PR #3631, opened 2024-11-18
chore: Replace `weld-codegen` with `handlebars`

joonas:
## Feature or Problem

This replaces the `weld-codegen` [Renderer](https://github.com/joonas/weld/blob/main/codegen/src/render.rs) functionality with direct dependency on Handlebars which is what's doing the heavy lifting anyway.

Since it doesn't look like we're actually making use of the additional helpers methods introduced in `weld-codegen`, this should be just a drop-in replacement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,6 @@
 version = 3
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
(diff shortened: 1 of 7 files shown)


brooksmtownsend (approved):
Nice!

PR #3633, opened 2024-11-19
Fixed example C component HTTP hello world

michelm:
## Feature or Problem

While trying the examples I stumbled upon some issues in example/c/components/http-hello-world.

- Fixed: Only set WASI_SDK_DIR variable to **/opt/wasi-sdk** in _Makefile_ when it has not been defined yet.
- Documentation: corrected path to wasm binary in wasmtime example.

### Manual Verification

Follow steps as described in README for this example.

### DIFF
--- a/examples/c/components/http-hello-world/Makefile
+++ b/examples/c/components/http-hello-world/Makefile
@@ -1,5 +1,5 @@
 HELLO_WASM=http_hello_world.wasm
-WASI_SDK_DIR=/opt/wasi-sdk
+WASI_SDK_DIR?=/opt/wasi-sdk
 AR=$(WASI_SDK_DIR)/bin/ar
 CC=$(WASI_SDK_DIR)/bin/clang
 VERSION=0.1.0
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):
Thank you @michelm ! If you wouldn't mind just signing off your commit, it's the failed action that we require to pass as a CNCF project. The details on the action should give you the command to do it 🙂 

If you don't want to do the chore, I can rebase/reword your commit too 😄

brooksmtownsend:
Yeah @michelm looks like that was correct to sign off, perhaps it was signing off one more commit than needed 🤔 

I was able to get it "as expected" in https://github.com/wasmCloud/wasmCloud/compare/main...brooksmtownsend:wasmCloud:fix-example-c-http-hello-world?expand=1 by doing
```shell
git rebase -i HEAD~3 --signoff
```

Then by dropping the first commit, and then rewording the next two to add the conventional commit prefix 

image


I don't think it's a _huge_ issue, it'll just end up with a duplicated dependabot commit, but let me know if you get to it by tomorrow and I can hold off 🙂

brooksmtownsend (approved):

brooksmtownsend (approved):
Hey @michelm I went ahead and just rebased on main to drop that extraneous commit, LGTM! Thanks again for fixing this 😄

PR #3635, opened 2024-11-19
feat(secrets-nats-kv): add get command to CLI

brooksmtownsend:
## Feature or Problem
This PR adds a `get` subcommand to the `secrets-nats-kv` CLI that allows for decrypting a secret that's stored in a NATS KV bucket, provided you have access to the encryption xkey seed. This requires direct JetStream access and does not actually send a request to a running secrets backend.

### Manual Verification
I manually verified this worked well for fetching a secret that I put using a valid XKey, and that it was denied if I provided the wrong kind, key, or bucket name.

### DIFF
--- a/crates/secrets-nats-kv/src/main.rs
+++ b/crates/secrets-nats-kv/src/main.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 
 use anyhow::{ensure, Context};
+use clap::ValueEnum;
 use clap::{Parser, Subcommand};
 use nkeys::XKey;
 use secrets_nats_kv::client::SECRETS_API_VERSION;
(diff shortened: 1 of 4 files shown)


protochron (approved):

crates/secrets-nats-kv/src/main.rs
```diff
@@ -107,6 +110,35 @@ struct PutCommand {
...
+}
+
+#[derive(Parser, Debug, Clone)]
+struct GetCommand {
+    /// The server's encryption XKey to decrypt the secret fetched from the KV store.
+    #[clap(short, long, env = "ENCRYPTION_XKEY_SEED")]
+    encryption_xkey_seed: String,
+    /// The NATS address to connect to where the backend is running
+    #[clap(long, default_value = "127.0.0.1:4222")]
+    nats_address: String,
```
vados-cosmonic, line 126: NIT: some of these (like `nats_address` in particular) MIGHT be better in `GlobalOpts`, but that works as a follow up!

vados-cosmonic (approved):
LGTM :rocket:

crates/secrets-nats-kv/src/main.rs
brooksmtownsend: Agreed! Good opportunities to clean that project up

PR #3636, opened 2024-11-19
chore(ci): tighten github actions workflow permissions

joonas:
## Feature or Problem

Doing another round of reviewing the various workflows and their permissions to tighten things down.

### DIFF
--- a/.github/workflows/dco-merge-group.yml
+++ b/.github/workflows/dco-merge-group.yml
@@ -5,6 +5,8 @@ name: DCO
 on:
   merge_group:
 
+permissions: {}
+
 jobs:
   DCO:
     runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


brooksmtownsend (approved):
LGTM as long as I'm not missing anything on the TS side @lachieh

joonas:
I ended up doing a few rounds of changes with the TS tooling based on errors, so I think we should be in a good spot with that 👍

lachieh:
Yeah looks good to me, they should all be read except for the app-release which is the only one that uses the GH release flow 👍🏻

joonas:
> Yeah looks good to me, they should all be read except for the app-release which is the only one that uses the GH release flow 👍🏻

Exactly, though when you're calling workflows from other workflows where the "sub-workflows" need more permissions than the workflow calling them, you'll need to make sure that the calling workflow requests the same permissions as the workflow it's calling, which is what I ended up wrestling with a bit

Issue #3609, opened 2024-11-14
[BUG] Two host instances are competing for requests after the first attempt of the 'wash dev' command failed


colonha:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

Following the QuickStart guide for GO on macOS, the first execution of the 'wash dev' command failed. On the second attempt, the command was successful; however, this resulted in two different host instances running simultaneously and competing for requests.

### Steps to reproduce

Follow the QuickStart guide for GO on macOS then in the `wash dev` It fails then run again the `wash dev` this will generate the duplicated hosts you can see the two hosts running `wash get inventory`

### Expected behavior

Changes made to the code in this case to the main.go are not reflected in the locahost serving the requests.
```
curl 127.0.0.1:8000
Hello from Go!
```

### Environment

- OS: MacOS
- Shell zhs
- wasmcloud Version v1.4.2
- wash Version v0.37.0

### Screenshots / Logs / Additional context

```
wash dev
ℹ️  Resolved wash session ID [Nv6zaT]
🚧 Starting a new host...
ℹ️  NATS server successfully started, using config @ [/Users/home/.wash/downloads/nats.conf]
ℹ️  NATS server logs written to [/Users/home/.wash/dev/Nv6zaT/nats.log]
🎣 Downloading new wadm from https://github.com/wasmcloud/wadm/releases/download/v0.18.0/wadm-v0.18.0-macos-aarch64.tar.gz
🎯 Saved wadm to /Users/home/.wash/downloads/wadm

🔧 Successfully started wasmCloud instance

failed to start host for session [Nv6zaT]

Caused by:
    0: timeout expired while reading for Host ID in logs
    1: deadline has elapsed
```

After failing a second execution of the command generated the duplicated hosts

```
wash get inventory

  Host ID                                                      Friendly name
  NCRPT3KPCI7SOIOI57AOK7IESKU52RENSRCOXIQRLAR3R4ICC5YIOX4P     delicate-tree-5299

  Host labels
  hostcore.arch                 aarch64
  hostcore.os                   macos
  hostcore.osfamily             unix

  No components found

  No providers found

  Host ID                                                      Friendly name
[...]

PR #3640, opened 2024-11-19
[FEAT] wash dev output improvements and fixes

brooksmtownsend:
## Feature or Problem
This PR, in order of commit:
- Rewords a couple help texts and reorders `eprintln` statements in wash dev to ensure more consistent and less duplicated output.
- Attempts to reorder some code logic to ensure the main devloop is where all output is printed from instead of being from nested functions, if possible.
- Ensures that wash dev is being run when only one host exists, allowing the existence of more than one host _only_ when a host ID is specified and that host is running. This also includes recommendations for users to run `wash down --all` to wipe to a clean slate in the case of bad state.
- Adds a spinner to the build process, during text output mode, to show that wash-cli is doing work and not just sitting there hoping for it to finish
- Fixes an issue where wadm wouldn't stop when up ctrl+c's
- Fixes an issue where wash up would exit with a failure finding the host PID
- Adds two integration tests for new functionality regarding multiple hosts

## Related Issues
Fixes https://github.com/wasmCloud/wasmCloud/issues/3609

## Release Information
wash-cli 0.37.1

### Manual Verification
I did run through this and ensure that it all works as expected, but I plan to follow this draft up with one more commit to add an integration test for the new functionality. Here's an example!

https://github.com/user-attachments/assets/d31ff985-c39d-4646-a99c-17bda77e462d

### DIFF
--- a/crates/wash-cli/src/bin/wash.rs
+++ b/crates/wash-cli/src/bin/wash.rs
@@ -58,10 +58,10 @@ Interact and manage wasmCloud applications, projects, and runtime environments
 Usage: wash [OPTIONS] <COMMAND>
 
 Build:
-  new          Create a new project from a template
+  new          Create a new project from a template or git repository
   build        Build (and sign) a wasmCloud component or capability provider
   dev          Start a developer loop to hot-reload a local wasmCloud component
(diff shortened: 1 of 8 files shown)


vados-cosmonic (approved):
LGTM :rocket:

PR #3641, opened 2024-11-19
Host-sys, control-interface and core dep cleanup

brooksmtownsend:
## Feature or Problem
Inspired by some of the pain @joonas went through to release the last version of wash-cli

This PR:
- ~Removes the host-sys crate, which was an initial experiment but hasn't been touched since its creation (@rvolosatovs please feel free to push back if this is serving a function somewhere~
- Removes unused dependencies from the `control-interface` and `core` crates.
- Removes the reliance on the `core` crate in `wasmcloud-control-interface` by using Strings instead of type aliases.

## Related Issues
This is related to a broader issue in this monorepo where crates are too hard to release and there's too much inter-reliance for a tiny subset of functionality. In this case, the reliance on the `core` crate for string type aliases created a dependency where any core crate release would also require a control interface release. There are many other changes we can make like this in the repository that don't require a lot of changes that can help us get to a point where our crates only depend on each other when there's really functionality that should be broken out and shared.

Other, not fully featured ideas to continue this trend are:
- [ ] Reduce public interactions with `core` types so that crates can co-depend on it without needing to be on the same version
- [ ] Merge `wash-lib` and `wash-cli` into `wash` to prevent two releases for libs and CLI crates
- [ ] Combine `wascap` / `provider-archive` into wasmcloud-core as modules of the crate, or try and untangle them.

## Release Information
Nothing needed urgently. This removes public re-exports from the wasmCloud control interface so, perhaps it's a breaking change we could prevent by just creating those type aliases instead of removing them.

### DIFF
--- a/crates/control-interface/src/broker.rs
+++ b/crates/control-interface/src/broker.rs
@@ -1,5 +1,6 @@
 const DEFAULT_TOPIC_PREFIX: &str = "wasmbus.ctl";
 // Copied from https://docs.rs/wasmcloud-core/0.15.0/wasmcloud_core/constant.CTL_API_VERSION_1.html
+// to avoid a dependency on a crate for one constant
 const CTL_API_VERSION_1: &str = "v1";
 
 fn prefix(topic_prefix: &Option<String>, lattice: &str, version: &str) -> String {
--- a/crates/control-interface/src/broker.rs
+++ b/crates/control-interface/src/broker.rs
@@ -1,4 +1,6 @@
 const DEFAULT_TOPIC_PREFIX: &str = "wasmbus.ctl";
+// Copied from https://docs.rs/wasmcloud-core/0.15.0/wasmcloud_core/constant.CTL_API_VERSION_1.html
+const CTL_API_VERSION_1: &str = "v1";
 
 fn prefix(topic_prefix: &Option<String>, lattice: &str, version: &str) -> String {
     format!(
@@ -10,7 +12,7 @@ fn prefix(topic_prefix: &Option<String>, lattice: &str, version: &str) -> String
 }
 
 pub mod v1 {
-    use wasmcloud_core::CTL_API_VERSION_1;
+    use crate::broker::CTL_API_VERSION_1;
 
     use super::prefix;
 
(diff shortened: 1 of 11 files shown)


vados-cosmonic (approved):
LGTM :rocket:

brooksmtownsend:
Keeping host-sys as per @rvolosatovs good point that it may not be actively contributed to, but it is a C integration point that we should preserve

crates/control-interface/src/broker.rs
```diff
@@ -1,4 +1,6 @@
 const DEFAULT_TOPIC_PREFIX: &str = "wasmbus.ctl";
+// Copied from https://docs.rs/wasmcloud-core/0.15.0/wasmcloud_core/constant.CTL_API_VERSION_1.html
```
thomastaylor312, line 2: Nit: Should probably add some explanation that this is copied to avoid a dep on a crate just for a constant

crates/control-interface/src/types/component.rs
```diff
@@ -4,15 +4,15 @@ use std::collections::BTreeMap;
...
+use crate::Result;
 
 /// A summary description of an component within a host inventory
 #[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 #[non_exhaustive]
 pub struct ComponentDescription {
     /// The unique component identifier for this component
     #[serde(default)]
-    pub(crate) id: ComponentId,
+    pub(crate) id: String,
```
thomastaylor312, line 15: I don't think changing this to a string should break anyone, but we should make sure we document it well just in case in the release notes

thomastaylor312 (approved):

crates/control-interface/src/types/component.rs
brooksmtownsend: Good point. I'm a little worried about the aliases if they are public, so will do a double check on this one

brooksmtownsend: @thomastaylor312 These won't break, but I did keep the string aliases around _just in case_ since they were publicly exported (e.g. https://docs.rs/wasmcloud-control-interface/2.2.0/wasmcloud_control_interface/type.ComponentId.html)

thomastaylor312 (approved):

PR #3643, opened 2024-11-20
feat(otel): Adds the ability to configure sampling and buffer size

thomastaylor312:
This adds the ability for provider to configure some of the tunables for tracing via config. The underlying SDK defaults (and hence, the host) use the standard environment variables

Please note that some of our traces are using a span that lives for the lifetime of the application, so the sampling percentage may not work until we fix that as well

### DIFF
--- a/crates/provider-sdk/src/otel.rs
+++ b/crates/provider-sdk/src/otel.rs
@@ -69,6 +69,36 @@ macro_rules! initialize_observability {
                     "OTEL_EXPORTER_OTLP_LOGS_ENDPOINT" => {
                         otel_config.logs_endpoint = Some(v.clone())
                     }
+                    "OTEL_TRACES_SAMPLER" => {
+                        otel_config.traces_sampler = Some(v.clone())
+                    }
+                    "OTEL_TRACES_SAMPLER_ARG" => {
+                        otel_config.traces_sampler_arg = Some(v.clone())
+                    }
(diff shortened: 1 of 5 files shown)


thomastaylor312:
This is in draft until I can do some manual validation that everything still works as intended

crates/provider-sdk/src/otel.rs
```diff
@@ -69,6 +69,42 @@ macro_rules! initialize_observability {
                     "OTEL_EXPORTER_OTLP_LOGS_ENDPOINT" => {
                         otel_config.logs_endpoint = Some(v.clone())
                     }
+                    "WASMCLOUD_OBSERVABILITY_TRACE_SAMPLING" => {
```
lxfontes, line 72: `OTEL_EXPORTER_OTLP_TRACES_SAMPLING`

( and think the others should be `OTEL_EXPORTER_OTLP_` prefix as well )

thomastaylor312: Is that a standard env var? The ones that were `OTEL` prefixed before are standard ones. The println was a mistake from tab completing an AI suggestion I think 😅

thomastaylor312: Lol it is: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#batch-span-processor

Yaaaaaay

brooksmtownsend (approved):

Issue #3551, opened 2024-11-06
[BUG-Washboard] wont update the washboard content when changing url


markkovari:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [X] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

Changing the URL should be able to fetch the information about the information of the lattice.
Even if the connection is established the washboard does not change.
 

### Steps to reproduce

Start a wasmCloud host with a websocket port, not the default 4223.

E.g.: 
```bash
wash up --nats-websocket-port 4224
```

Go to another terminal and start the washboard

```bash
wash ui
```

visit http://localhost:3030/

go to the settings and change the lattice url to `ws://localhost:4224`

### Expected behavior

The washboard should have the newly started host, but it is empty.

The expected behaviour is to fetch the host (or lattice) inventory, links and configs and display on the board.

### Environment

- OS: MacOS
- zsh
wash          v0.36.1
├ nats-server v2.10.20
├ wadm        v0.18.0
└ wasmcloud   v1.4.0

### Screenshots / Logs / Additional context

You can validate with tools or with the browsers developer tools if the websocket communication is functioning or not.

In the screenshot `malna` is where the host is, based on the reproduction steps it will be `localhost` 

![image](https://github.com/user-attachments/assets/1de01a57-4d42-4c58-9e44-cfe7c52de202)

![image](https://github.com/user-attachments/assets/7798fb43-f8f5-41fe-b335-1c0ed4565c5c)

![image](https://github.com/user-attachments/assets/99bdcbe2-034c-4ead-9ef8-4e4a37000d55)

PR #3652, opened 2024-11-21
fix(wash-cli): add new url to check for washboard releases

### LINKED ISSUE #3428: [BUG] Change of tag pattern for typescript packages breaks wash ui download
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [X] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

In updating the workflows for typescript projects, I added a workflow for releasing washboard. When looking to start the ui with `wash ui`, it expects the tag pattern for to be `washboard-ui-v*` however the pipeline will now be looking for the path to the folder as part of the tag name: `typescript/apps/washboard-ui/v*`

### Steps to reproduce

- Pushing `washboard-ui-*` does not trigger a release
- Pushing `typescript/apps/washboard-ui/v*` prevents `wash ui` from downloading the release

### Expected behavior

We either need to update `wash` to look at the new tag location (probably both patterns) or we need to change the workflow to go back to the expected tag pattern.

lachieh:
Change in the tags for washboard means that washboard is no longer at the same URL. This adds the new URL pattern, and falls back to the old one if it cannot be found.

It might be worth noting that this will need an additional URL when the washboard code moves to its own repo.

Didn't add tests because I wouldn't have a clue where to even start with writing tests in rust but I'm happy to learn if someone wants to give me some direction.

Closes #3428

### DIFF
--- a/crates/wash-cli/src/ui/mod.rs
+++ b/crates/wash-cli/src/ui/mod.rs
@@ -64,25 +64,20 @@ async fn ensure_washboard(version: &str, base_dir: PathBuf) -> Result<PathBuf> {
 }
 
 async fn download_washboard(version: &str, install_dir: &PathBuf) -> Result<()> {
-    let release_url = format!(
-        "https://github.com/wasmCloud/wasmCloud/releases/download/washboard-ui-{version}/washboard.tar.gz"
-    );
+    let urls = vec![
+        format!(
+            "https://github.com/wasmCloud/wasmCloud/releases/download/typescript%2Fapps%2Fwashboard-ui%2F{version}/washboard.tar.gz"
+        ),
+        format!(
+            "https://github.com/wasmCloud/wasmCloud/releases/download/washboard-ui-{version}/washboard.tar.gz"
+        ),
+    ];
 
-    // Download tarball
-    let resp = tls::DEFAULT_REQWEST_CLIENT
-        .get(&release_url)
-        .send()
+    let body = try_download_from_urls(&urls)
         .await
-        .context("failed to request washboard tarball")?;
+        .context("Failed to download washboard-ui assets")?;
 
-    if resp.status() != reqwest::StatusCode::OK {
-        bail!("failed to download washboard tarball: {}", resp.status());
-    }
-
-    let body = resp
-        .bytes()
-        .await
-        .context("failed to read bytes from washboard tarball")?;
+    eprintln!("Downloaded washboard-ui@{}", version);
 
     // Unpack and copy to install dir
     let cursor = Cursor::new(body);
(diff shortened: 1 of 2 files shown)


crates/wash-cli/src/ui/mod.rs
```diff
@@ -64,33 +64,57 @@ async fn ensure_washboard(version: &str, base_dir: PathBuf) -> Result<PathBuf> {
...
+            Ok(body) => return Ok(body),
+            Err(e) => last_error = Some(e),
+        }
+    }
+
+    Err(last_error.unwrap_or_else(|| anyhow::anyhow!("Failed to find suitable download URL")))
+}
+
+async fn try_download(url: &String) -> Result<bytes::Bytes> {
+    let resp = tls::DEFAULT_REQWEST_CLIENT
```
markkovari, line 107: Wow,
 
I do have a really good question here 😄 (someone asked my like an hour ago) could you please use this client instead of `tls::DEFAULT_REQWEST_CLIENT` [this one with magical proxy things in it](https://github.com/wasmCloud/wasmCloud/blob/fc0aea1aa752dcd24d80e011e811b0d0e5a404ac/crates/wash-lib/src/start/github.rs#L95)? Should be a drop-in replacement.


Btw after my pr we can even go ahead and try to fetch the latest "any" version of the washboard too **later**. (this [PR](https://github.com/wasmCloud/wasmCloud/pull/2527) with a bit of modifications)

vados-cosmonic (approved):
LGTM :rocket:

crates/wash-cli/src/ui/mod.rs
lachieh: updated against main with your changes in, @markkovari!

lachieh:
@brooksmtownsend @vados-cosmonic tagging for visibility next week, this is ready to go and will need to get merged before the next wash release. This is what will allow downloading of washboard@v0.6.0 which includes the fix for #3551

crates/wash-cli/src/ui/mod.rs
markkovari: Thank you

markkovari:
Quick heads up @lachieh the [gh action](https://github.com/wasmCloud/wasmCloud/actions/runs/11984742738/job/33415861333?pr=3652) has been running for like 3.5 - 4 hours, my question regarding this; I think you have hosted runners and the github actions timeout is not the default 1hour, should it make sense to create a timeout not to run the action for more than that? not sure what the longest action is, but might add an upper limit too.

cc @brooksmtownsend 🙈

vados-cosmonic:
Thanks @lachieh ! merging this now :)

PR #3654, opened 2024-11-21
test(wash-cli): add tests for wash help short and help markdown

vados-cosmonic:
## Feature or Problem

This PR introduces simple tests for wash help options

### DIFF
--- a/crates/wash-cli/tests/wash.rs
+++ b/crates/wash-cli/tests/wash.rs
@@ -1,15 +1,16 @@
-mod common;
+use anyhow::{Context as _, Result};
 
+mod common;
 use common::{output_to_string, wash};
 
 // The purpose of this text is to ensure we don't remove subcommands from the help text without knowing.
 #[test]
-fn integration_help_subcommand_check() {
+fn integration_help_subcommand_check() -> Result<()> {
     let help_output = wash()
         .args(["--help"])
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Note: This is blocked by https://github.com/wasmCloud/wasmCloud/pull/3656 where we fix the `squid-proxy` image we're using!

crates/wash-cli/tests/wash.rs
```diff
@@ -38,4 +39,25 @@ fn integration_help_subcommand_check() {
...
     assert!(output.contains("claims"));
+    Ok(())
+}
+
+/// Ensure `wash -h` works
+#[test]
+fn integration_help_short_works() -> Result<()> {
+    wash()
+        .args(["-h"])
+        .output()
```
ricochet, line 50: Shouldn't there be a few trivial output.contains here?

vados-cosmonic: This would help more on the other new test -- `-h` is the short form of `--help` so it *should* be fine -- if another short form was added, this would panic (which is why this test was added -- @brooksmtownsend ran into that panic after a recent change).

what might make sense is some includes for `--help-markdown`, so adding that!

brooksmtownsend (approved):

PR #3669, opened 2024-11-22
chore(host): Proxy SSL_CERT_FILE and SSL_CERT_DIR to providers

joonas:
## Feature or Problem

In an environment where custom CAs are used, `SSL_CERT_FILE` and `SSL_CERT_DIR` can be used to configure many TLS libraries to read a specific file or directory for it's certificates to use.

These values can be used to configure the host itself already, but when the host starts providers, these values were not being passed down to the provider to make use of, and so it would not be able to start.

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -2301,6 +2301,14 @@ impl Host {
                 let _ = child_cmd.env("RUST_LOG", rust_log);
             }
 
+            // Proxy SSL_CERT_FILE and SSL_CERT_DIR to the provider to use for setting up TLS connections
+            if let Ok(ssl_cert_file) = env::var("SSL_CERT_FILE") {
+                let _ = child_cmd.env("SSL_CERT_FILE", ssl_cert_file);
+            }
+            if let Ok(ssl_cert_dir) = env::var("SSL_CERT_DIR") {
+                let _ = child_cmd.env("SSL_CERT_DIR", ssl_cert_dir);
+            }
+
             let mut child = child_cmd
                 .stdin(Stdio::piped())
                 .kill_on_drop(true)


crates/host/src/wasmbus/mod.rs
```diff
@@ -2301,6 +2301,14 @@ impl Host {
                 let _ = child_cmd.env("RUST_LOG", rust_log);
             }
 
+            // Proxy SSL_CERT_FILE and SSL_CERT_DIR to the provider to use for setting up TLS connections
+            if let Ok(ssl_cert_file) = env::var("SSL_CERT_FILE") {
+                let _ = child_cmd.env("SSL_CERT_FILE", ssl_cert_file);
+            }
+            if let Ok(ssl_cert_dir) = env::var("SSL_CERT_DIR") {
+                let _ = child_cmd.env("SSL_CERT_DIR", ssl_cert_dir);
+            }
```
vados-cosmonic, line 2310: Can we handle this the same way we do OTEL vars and let providers pull it in themselves from the host? I think we might benefit from more flexibility by making sure all env used for providers are pulled in that way (and making it easy via the macro). 

This way, someone can deploy many hosts and actually just change config for them rather than going back and restarting (possibly) to modify each host's individual env.

joonas: I'm not sure I am familiar with how we handle the OTEL environment variables actually, would you mind linking me to the code you're thinking of this should be modeled after?

vados-cosmonic: Sure! Usage looks like this:

https://github.com/wasmCloud/wasmCloud/blob/main/examples/rust/providers/custom-template/wadm.yaml#L14

And integration looks like this:

https://github.com/wasmCloud/wasmCloud/blob/f051d9d3b8afb0ea1baafd87941babb29830fe4d/crates/provider-sdk/src/otel.rs#L60

joonas: These variables are a bit different from the OTEL variables in that they are usually consumed by the lower-level (and/or standard) libraries that are not being directly configured or enabled, but rather relied on by everything else building on top.

So in that sense, if these values were set as part of the environment the host is running on, one might reasonable expect them to just work without any further configuration.

For that reason, I think the reasonable thing to do here would be to simply proxy them through.

vados-cosmonic: Hey yeah so for this I'm happy to go with what people decide on this -- I think consistency here is good but maybe these SSL vars are worth special casing!

PR #3670, opened 2024-11-22
fix: propagate serve trace into handlers

rvolosatovs:
(description written by @brooksmtownsend since we collaborated on this PR)

## Feature or Problem
This PR propagates a span in the invocation context which is then used to instrument component function calls for the entire lifecycle of an invocation. These were disconnected before because of the way we were pulling invocations off of a stream.

## Related Issues
There is one further issues that is not fixed in this PR:
1. The information from awaiting the future and acquiring the semaphore permit is not connected to the invocation trace since the context is not available there (left a note with TODO(#3670))
Example: 
image

### Manual Verification
A debug invocation of the keyvalue counter:
image


brooksmtownsend:
Not quite finished here yet, will be pushing up a commit or two to wrap this up shortly

joonas (approved):

crates/host/src/wasmbus/mod.rs
```diff
@@ -274,17 +277,24 @@ impl wrpc_transport::Serve for WrpcServer {
...
-            // reason is that before we were trying to use the current span for instrumentation.
-            // This didn't work because the span created by the `instrument` macro exits after
-            // this function returns, so we ended up with no parent span at all when trying to
-            // attach header data. I also set this as an info span so that each invocation at
-            // least has this top level span. If the fields add too much overhead, we can remove
-            // those fields instead.
             let span = tracing::info_span!("component_invocation", func = %func, id = %id, instance = %instance);
             async move {
+                if let Some(ref cx) = cx {
+                    // TODO: wasmcloud_tracing take HeaderMap for my own sanity
```
joonas, line 283: Do we need/want to capture a follow-up for this?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1271,77 +1267,91 @@ impl Host {
...
+                        let mut tasks = JoinSet::new();
+                        let mut exports = stream::select_all(exports);
+                        loop {
+                            let permits = Arc::clone(&permits);
+                            select! {
+                                Some(fut) = exports.next() => {
+                                    match fut {
+                                        Ok(fut) => {
+                                            // TODO(#3670): Awaiting this future emits some spans that we can't connect back
+                                            // to the original invocation since we don't have the context. This needs to be investigated.
```
joonas, line 1282: Should this be captured in a follow-up issue?

crates/runtime/src/component/http.rs
```diff
@@ -132,90 +133,93 @@ where
...
-            // us, since it requires a `hyper::Error`
+        // The below is adapted from `WasiHttpView::new_incoming_request`, which is unusable for
+        // us, since it requires a `hyper::Error`
 
-            tracing::error!("HELLO");
+        let (parts, body) = request.into_parts();
+        let body = HostIncomingBody::new(
+            body,
+            // TODO: this needs to be plumbed through
+            std::time::Duration::from_millis(600 * 1000),
```
joonas, line 160: Should this be `std::time::Duration::from_mins(10)` instead?

brooksmtownsend:
I have a fix for part 2 that i'll push here

crates/runtime/src/component/http.rs
rvolosatovs: Yes, but this is copy-pasted from upstream, so let's keep it as is

crates/host/src/wasmbus/mod.rs
rvolosatovs: Nobody seems to know what this means. I have moved this from original implementation in one of the refactors.
If we cannot decipher this comment, maybe we should just remove it finally

joonas: I'm 👍 for removing if we can't think of a reason why it should continue to exist.

brooksmtownsend:
Pushed an additional commit here that propagates the trace context at the component host/guest boundary using the wasmtime store and an optional parent context instead of an `Arc<Rwlock<>>` in the handler and WrpcServer. I had to change it first because traces were being interleaved, but I think this will lead to better performance as well with fewer write locks in the hot path.

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I wrote it 😢 and it was just a comment to avoid the random `iter()` here. I think we can remove it

crates/host/src/wasmbus/mod.rs
```diff
@@ -1255,77 +1252,91 @@ impl Host {
...
-                                            }
+            exports: spawn(async move {
+                join!(
+                    async move {
+                        let mut tasks = JoinSet::new();
+                        let mut exports = stream::select_all(exports);
+                        loop {
+                            let permits = Arc::clone(&permits);
+                            select! {
+                                Some(fut) = exports.next() => {
```
rvolosatovs, line 1263: Like below, looks like we should be able to instrument here just once instead of the duplication in the runtime crate

crates/host/src/wasmbus/handler.rs
```diff
@@ -169,13 +166,6 @@ impl wrpc_transport::Invoke for Handler {
     where
         P: AsRef<[Option<usize>]> + Send + Sync,
     {
-        // Reading a trace context should _never_ block because writing happens once at the beginning of a component
-        // invocation. If it does block here, it's a bug in the runtime, and it's better to deal with a
-        // disconnected trace than to block on the invocation for an extended period of time.
-        if let Ok(trace_context) = self.trace_ctx.try_read() {
```
rvolosatovs, line 175: Could you help me understand why we had to replace this by setting the trace parent in each handler?

thomastaylor312: Not Brooks but I helped him debug this. Essentially the trace context was being used at the top level wrpc server part so it was being shared among _all_ components (not reset for each handler like we do with link names). This was because the initial instantiate_component or serve (can't remember exactly which one) clones the trace_ctx from `Handler` once and then uses that on every invocation. This lead to interleaving of the traces since they were all sharing the same trace ctx

crates/host/src/wasmbus/mod.rs
```diff
@@ -1255,77 +1252,91 @@ impl Host {
...
+                                    match fut {
+                                        Ok(fut) => {
+                                            // TODO(#3670): Awaiting this future emits some spans that we can't connect back
+                                            // to the original invocation since we don't have the context. This needs to be investigated.
+                                            debug!("accepted invocation, acquiring permit");
+                                            let permit = permits.acquire_owned().await;
+                                            tasks.spawn(async move {
+                                                let _permit = permit;
+                                                debug!("handling invocation");
+                                                match fut.await {
```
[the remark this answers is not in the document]
thomastaylor312: I know we tried instrumenting these at some point in our debugging and they didn't end up connected. At this point, I don't think we have the trace context available from an incoming invocation in order to actually use it to wire up a span correctly. We only have that information after we do the incoming invocation stream mapping. @brooksmtownsend can correct me if I missed something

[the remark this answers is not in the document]
thomastaylor312: FWIW we also didn't really love the idea of adding it to the runtime crate. We could always pass through a trace ID instead as a string (or other header extractor/injector stuff)

crates/runtime/src/component/mod.rs
```diff
@@ -305,6 +306,7 @@ fn new_store<H: Handler>(
             table,
             shared_resources: SharedResourceTable::default(),
             timeout: max_execution_time,
+            parent_context: None,
```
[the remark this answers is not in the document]
brooksmtownsend: I think relevant to your comments below, it felt like an odd requirement for creating a new store to require a parent context. Perhaps I should just do it inside of `new_store` implicitly

crates/host/src/wasmbus/mod.rs
brooksmtownsend: I do know that instrumenting the `exports.next()` future here won't work, since that gets polled continuously until a request comes in. But I did attempt to instrument the `fut`, and I remembered that being disconnected. Getting back into investigating this today

crates/host/src/wasmbus/handler.rs
brooksmtownsend: Yep that's right, the trace context needs to be accessible per `Instance` rather than per Component/WrpcServer/Handler. So putting it in the store was the most obvious way I could see to do this, but happy to do it in an alternative way

brooksmtownsend:
I think this is good to go as long as @rvolosatovs is okay with the current changes.

brooksmtownsend:
Latest commit enables tracing for builtin providers, e.g. the builtin NATS 
image

brooksmtownsend (approved):

Issue #2873, opened 2024-08-27
[FEATURE] Colorize wash top level help


vados-cosmonic:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

As a follow up to https://github.com/wasmCloud/wasmCloud/pull/2869, it'd be great if we colorized the help text we currently have hard-coded into `wash`. We want to be able to control the order in which commands show up there and the layout, so we can make it easier for people to group commands, *but* this is incompatible with the automated colorization support for `clap`.

There are at least two ways to go about this:

1. Manually colorize the help text itself (consider using [`clap::builder::StyledStr`](https://docs.rs/clap/latest/clap/builder/struct.StyledStr.html))
2. Use [`render_help()`](https://docs.rs/clap/latest/clap/struct.Command.html#method.render_help) on existing `Command`s and generate the help text, but in the order and with the prose we'd like to keep.

Both approaches could be a pretty good fit, and may have to be updated in much the same ways (i.e. if we add a new subcommand, we'd have to manually update the help fn).

### Describe the solution you'd like

Wash CLI should be augmented to produce colored output to the top level `--help` switch

PR #3681, opened 2024-11-24
style(wash-cli): main help is colorized

markkovari:
## Feature or Problem
Using colours for the main help. 🎉 

although I have a question: compared to the "not overwritten" version of the help this command is not in the list of commands when using the custom 🤔 

I know it i behind experimental flag, but was it intentionally hidden from the cli?

 > capture      Capture and debug cluster invocations and state
 
 
and I updated the comments above the (Sub)Commands to reflect the plain text version. 
Some of them have been out of sync (beside the `NOTE`).

Did a minimal research too if there is a dynamic stuff or not, but not really 😞 manual maintenance it is.
 
## Related Issues

closes: #2873 

### Manual Verification

```console
cd crates/wash-cli
```
```console
cargo run --
```

### DIFF
--- a/crates/wash-cli/src/bin/wash.rs
+++ b/crates/wash-cli/src/bin/wash.rs
@@ -56,18 +56,17 @@ struct HelpTopic {
 impl Display for HelpTopic {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         const PADDING_AFTER_LONGEST_SPACES: usize = 3;
+        const DEFAULT_PADDING_START: usize = 25;
         writeln!(f, "{}", self.name.green().bold())?;
         let longest_command_length = self
             .commands
             .iter()
             .map(|(name, _)| name.len())
             .max()
-            .unwrap_or(25)
+            .unwrap_or(DEFAULT_PADDING_START)
             + PADDING_AFTER_LONGEST_SPACES;
 
         for (name, desc) in &self.commands {
-            // Add a padding to the desc to align the descriptions
-            // NOTE: There should be a (hopefullty) a better way to align the descriptions
             let padding = " ".repeat(longest_command_length - name.len());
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Hey @markkovari this looks great -- I think you're missing the `color_print` dep addition though :)

That said, we already have [`termcolor`](https://crates.io/crates/termcolor) so it might be nice to use that instead of introducing a new dep

markkovari:
Thanks @vados-cosmonic, yeah it was pretty late 😄 I used the mentioned dependencies and added a dynamic padding to avoid width typos for the commands. 

Although there must be a better way to pad the strings, so whoever knows by heart please help me out.

crates/wash-cli/src/bin/wash.rs
```diff
@@ -43,7 +46,54 @@ use wash_cli::style::WASH_CLI_STYLE;
...
+impl Display for HelpTopic {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        const PADDING_AFTER_LONGEST_SPACES: usize = 3;
+        writeln!(f, "{}", self.name.green().bold())?;
+        let longest_command_length = self
+            .commands
+            .iter()
+            .map(|(name, _)| name.len())
+            .max()
+            .unwrap_or(25)
```
vados-cosmonic, line 64: you could pull this out as a constant as well!

vados-cosmonic (approved):
This looks great for a start @markkovari thanks!

I think the hard-coded spacing isn't the end of the world -- it's certainly better than doing it completely manually to start.

brooksmtownsend (approved):

markkovari:
I might need some help here 😶‍🌫️ one test failed and removed this item from the queue. Although should have no effect.

🥺 @vados-cosmonic or @brooksmtownsend this might need a non-violent re-kick
Thank you

Issue #3034, opened 2024-09-17
[BUG] wash app deploy should fail or provide indication if component or provider is missing.


cdmurph32:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

For complex deployments with multiple components, it can be difficult to determine why a service isn't working. The fact that a component or provider cannot be found should be logged to the wasmcloud.log. Perhaps, `wash app deploy` could fail. 

### Steps to reproduce

1. Attempt to deploy any example with `wash app deploy` without building the component. 

### Expected behavior

I'd expect an error to be logged, and potentially a command failure.

### Environment

- OS: Linux and MacOS
- Shell zsh
- wasmcloud Version v1.2.1

brooksmtownsend:
This is a good idea @cdmurph32, right now we absolutize the paths from relative paths in the manifest and we could go ahead and make sure that exists. At the very least when you deploy the app it should enter a `Failed` state, but we can catch it sooner.

Something to look out for while implementing this is that if it's an absolute path we have to allow that to continue to deploy even if the file doesn't exists, because we could be deploying from a totally different filesystem. Happy to elaborate here

cpetig:
The shell running `wash up` reports
```
2024-11-14T18:36:30.487946Z ERROR wasmcloud_host::wasmbus: failed to scale component component_ref=file:///home/.../hello/build/http_hello_world_s.wasm component_id=rust_hello_world-http_component err=failed to fetch component

Caused by:
    0: failed to read component
    1: No such file or directory (os error 2)
2024-11-14T18:37:00.482627Z ERROR wasmcloud_host::wasmbus: failed to scale component component_ref=file:///home/.../hello/build/http_hello_world_s.wasm component_id=rust_hello_world-http_component err=failed to fetch component
```
while the shell I typed `wash app deploy local.wadm.yaml`
shows `Successfully deployed model rust-hello-world v0.0.1` which clearly is confusing.

cpetig:
deploy_model_from_manifest should check the file name specified in the manifest at `spec/components/properties/image` for existence.

vados-cosmonic:
This should be resolved by #3682 !

PR #3682, opened 2024-11-24
fix: check local deployment paths

Gohlub:
## Feature or Problem
`wash app deploy` fails to detect whether the build for a local component is available. Fix introduces checks for relative paths and their existance.

## Related Issues
Fix for issue #3034

## Release Information
Next

## Consumer Impact
Deploying a local build now checks whether the path is absolute (if yes, proceed regularly), and if not, raises an error indicating which build of the component in question is missing.

### Acceptance or Integration
No changes or additions to the acceptance or integration test suite

### Manual Verification
In `examples/rust/components/http-jsonify`, modify the `wadm.yaml` so that the build is locally sourced.
image
In the same folder, run:
```shell
wash up -d
wash app deploy ./wadm.yaml
```
You should get:
image

Which indicates that the local build is not available.

I will label this as a work in progress since I am not sure whether this particular handling makes sense, and I am open to suggestion.

### DIFF
--- a/crates/wash-lib/src/app.rs
+++ b/crates/wash-lib/src/app.rs
@@ -72,11 +72,18 @@ fn resolve_relative_file_paths_in_yaml(
         serde_yaml::Value::String(s)
             if s.starts_with("file://") && s.chars().nth(7).is_some_and(|v| v != '/') =>
         {
-            // Convert the base dir + relative path into a file based URL
-            let full_path = base_dir.as_ref().join(
-                s.strip_prefix("file://")
-                    .context("failed to strip prefix")?,
-            );
+            // Only check existence for relative paths we're resolving
+            let relative_path = s.strip_prefix("file://")
+                .context("failed to strip prefix")?; 
+            
+            let full_path = base_dir.as_ref().join(relative_path);
+            // Only check existence if it was a relative path (starts with ./ or ../)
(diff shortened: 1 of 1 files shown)


crates/wash-lib/src/app.rs
```diff
@@ -72,11 +72,18 @@ fn resolve_relative_file_paths_in_yaml(
...
+                .context("failed to strip prefix")?; 
+            
+            let full_path = base_dir.as_ref().join(relative_path);
+            // Only check existence if it was a relative path (starts with ./ or ../)
+            if relative_path.starts_with('.') {
+                (!full_path.exists()).then(|| ()).context(format!(
+                    "referenced relative file does not exist: {}", 
+                    full_path.display()
+                ))?;
+            }
```
vados-cosmonic, line 86: Ah note that relative paths are actually *any* path that starts in `file://` -- so if you had `file://path/to/file`, we'd expect the relative path to be `path/to/file` (from the base directory).

I think you can change the relative path check to be something like `s.starts_with("file:///")` instead and be a little bit more explicit that we're checking in the case that it *wasn't* absolute. 

That said, I think this code might be better rewritten as a `match` in the first place, so we can cover the `file://` and `file:///` cases really explicitly.

vados-cosmonic (approved):
LGTM :rocket:

PR #3686, opened 2024-11-25
fix(host): remove export joinset

brooksmtownsend:
## Feature or Problem
This PR removes our use of a `JoinSet` and then awaiting the completion exported invocation task in the `select!{}` loop. This caused a set of delays when handling component invocations as it would wait for a previous invocation to finish before accepting _any_ new invocations depending on which branch the `select!{}` loop selected.


vados-cosmonic (approved):
LGTM :rocket:

rvolosatovs:
That's not how [`select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) or [`JoinSet::join_next`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_next) are supposed to work. If anything was blocking here - that sounds like an upstream bug, which should be reported

PR #3698, opened 2024-11-26
feat(wash-cli): enable purging all links

vados-cosmonic:


crates/wash-cli/src/cmd/link/del.rs
```diff
@@ -0,0 +1,162 @@
...
+                .await
+                .context("failed to delete all links");
+        } else {
+            match prompt_for_choice(
+                &StringEntry {
+                    default: Some("cancel".to_string()),
+                    choices: Some(vec!["Delete all links".to_string(), "Cancel".to_string()]),
+                    regex: None,
+                },
+                "Are you sure you want to iteratively delete all links in the cluster? (this action cannot be reversed)",
```
brooksmtownsend, line 90:
```suggestion
                "Are you sure you want to delete all links in the cluster? (this action cannot be reversed)",
```

supernit

crates/wash-cli/tests/wash_link.rs
```diff
@@ -38,3 +38,36 @@ async fn integration_link_serial() -> Result<()> {
 
     Ok(())
 }
+
+/// Ensure wash can delete all links
+#[tokio::test]
+#[serial]
+async fn integration_link_del_all_serial() -> Result<()> {
```
brooksmtownsend, line 45: I think it would be good for this test to define a few different links and then ensure that the delete all deletes all of them

brooksmtownsend (changes requested):
Only request is to have the test run the command and make sure all the links get deleted 😄

crates/wash-cli/tests/wash_link.rs
vados-cosmonic: Ah thanks, definitely forgot to actually fill out this test!

crates/wash-cli/src/cmd/link/query.rs
```diff
@@ -0,0 +1,29 @@
+//! Functionality enabling the `wash link del` subcommand
```
thomastaylor312, line 1: This should read `wash link get`

thomastaylor312 (approved):
One super minor doc comment change needed. Approving so you can change and resolve the comment!

crates/wash-cli/src/cmd/link/query.rs
vados-cosmonic: Ah great catch, just fixed this!

Issue #3667, opened 2024-11-22
[BUG] JS example missing types


vados-cosmonic:
### Affected project(s)

- [ ] documentation
- [x] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

The JS hello-world example seems to be missing a step for generating types which would normally be needed.

### Steps to reproduce

1. Start a project from the JS hello-world example
2. Attempt to add types

### Expected behavior

We should have a step for building/generated the relevant types via `jco`, and run that as part of the build.

It's also probably a good idea to update the documentation for examples to include instructions on modifying `tsconfig.json`, since that is pretty much required for decent ergonomics.

### Environment

- OS: [e.g. Linux, MacOS]
- Shell [e.g. bash, zsh, powershell]
- wasmcloud Version [e.g. 0.81.0]

PR #3700, opened 2024-11-26
fix(examples): add missing type generation to js http hello world

vados-cosmonic:
## Feature or Problem

Resolves https://github.com/wasmCloud/wasmCloud/issues/3667


examples/typescript/components/http-hello-world/package.json
```diff
@@ -4,10 +4,11 @@
   "description": "A WebAssembly Component that responds to HTTP requests using the `wasi:http` WIT interface",
   "main": "dist/http-hello-world.js",
   "scripts": {
+    "generate:types": "jco types wit/ -o generated/types",
     "build:tsc": "tsc",
     "build:js": "jco componentize -w wit -o dist/http-hello-world.wasm dist/http-hello-world.js",
     "build:component": "wash build --sign-only --config-path wasmcloud.toml",
-    "build": "npm run build:tsc && npm run build:js && npm run build:component",
+    "build": "npm run generate:types && npm run build:tsc && npm run build:js && npm run build:component",
```
markkovari, line 11: Hey, just a question here. I am not familiar with jco and its internals, but would it make sense to clean the folder before putting the generated files there? I mean if it is changing in the future are the generated folder recreated? 🤔 
Maybe `generate:types` might have a nice `rimraf` or similar. I know that this is a template and the chance is low to litter the folder here, but the peeps who are using it might do so.

brooksmtownsend (approved):

examples/typescript/components/http-hello-world/package.json
vados-cosmonic: Oh yeah this is a great point -- here's [an issue](https://github.com/wasmCloud/wasmCloud/issues/3856) to track this

IIRC we're going to move some examples out of the main repo to make them a bit easier to maintain independently, so this change could be made once that's done!

Issue #3604, opened 2024-11-14
[BUG] wash app deploy should indicate that the manifest argument is missing


cpetig:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

If you run `wash app deploy` without any argument the error message isn't helpful.

### Steps to reproduce

`wash app deploy` without any argument gives:
```
app manifest loader timed out

Caused by:
    deadline has elapsed
```
which to me didn't indicate that the manifest argument was missing.

### Expected behavior

It should indicate something like
```
Manifest argument missing. Provide the manifest as the first argument, e.g. local.wadm.yaml.
```

### Environment

- OS: WSL2, ubuntu 22.04
- Shell: bash
- wasmcloud Version 0.28.1 (wash --version)

cpetig:
I ran into this while reproducing https://github.com/wasmCloud/wasmCloud/issues/3034

cpetig:
The shell running `wash up` reports
````
2024-11-14T18:36:30.487946Z ERROR wasmcloud_host::wasmbus: failed to scale component component_ref=file:///home/.../hello/build/http_hello_world_s.wasm component_id=rust_hello_world-http_component err=failed to fetch component

Caused by:
    0: failed to read component
    1: No such file or directory (os error 2)
2024-11-14T18:37:00.482627Z ERROR wasmcloud_host::wasmbus: failed to scale component component_ref=file:///home/.../hello/build/http_hello_world_s.wasm component_id=rust_hello_world-http_component err=failed to fetch component
```
while the shell I typed `wash app deploy local.wadm.yaml`
shows `Successfully deployed model rust-hello-world v0.0.1` which clearly is confusing.

cpetig:
My guess is that this is `load_app_manifest("-".parse()?).await?` timing out (src/app/mod.rs:324)

vados-cosmonic:
Hey @cpetig thanks for filing this -- yeah I think the issue is going to be related to the optional arg there -- we *should* be doing some checking and possibly printing an error message like you noted! Will take a look at this.

vados-cosmonic:
[...]

PR #3701, opened 2024-11-26
fix(wash-cli)!: disable loading from stdin

vados-cosmonic:
## Feature or Problem

This option was evidently confusing, and would cause `wash app deploy` to hang for a bit while waiting for STDIN. It is likely better to force users to pass `-` directly to read from STDIN.

Resolves https://github.com/wasmCloud/wasmCloud/issues/3604

### DIFF
--- a/crates/wash-cli/src/app/mod.rs
+++ b/crates/wash-cli/src/app/mod.rs
@@ -320,8 +320,15 @@ async fn deploy_model(cmd: DeployCommand) -> Result<CommandOutput> {
     let client = connection_opts.into_nats_client().await?;
 
     let app_manifest = match cmd.app_name {
+        Some(source) if source == "-" => load_app_manifest("-".parse()?).await?,
         Some(source) => load_app_manifest(source.parse()?).await?,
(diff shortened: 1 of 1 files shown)


crates/wash-cli/src/app/mod.rs
```diff
@@ -320,8 +320,15 @@ async fn deploy_model(cmd: DeployCommand) -> Result<CommandOutput> {
     let client = connection_opts.into_nats_client().await?;
 
     let app_manifest = match cmd.app_name {
+        Some(source) if source == "-" => load_app_manifest("-".parse()?).await?,
```
brooksmtownsend, line 323: nit: since we know it's stdin, maybe an enum would prevent a parse/some slightly weird logic here?

brooksmtownsend (approved):

crates/wash-cli/src/app/mod.rs
vados-cosmonic: Yeah I'm not sure it's worth the additional complexity just yet! I think its' still relatively easy to find the code/update this if necessary

PR #3707, opened 2024-11-27
feat: rework Wasmtime engine configuration

rvolosatovs:
## Feature or Problem

Before:
current-host-using-builtin-messaging
After:
new-host-config-using-builtin-messaging

### DIFF
--- a/crates/runtime/src/runtime.rs
+++ b/crates/runtime/src/runtime.rs
@@ -2,53 +2,154 @@ use crate::ComponentConfig;
 
 use core::fmt;
 use core::fmt::Debug;
+use core::str::FromStr;
 use core::time::Duration;
 
+use std::env::{self, VarError};
 use std::thread;
 
-use anyhow::Context;
+use anyhow::Context as _;
+use tracing::{debug, error};
 use wasmtime::{InstanceAllocationStrategy, PoolingAllocationConfig};
 
-/// Default max linear memory for a component (256 MiB)
-pub const MAX_LINEAR_MEMORY: u64 = 256 * 1024 * 1024;
-/// Default max component size (50 MiB)
-pub const MAX_COMPONENT_SIZE: u64 = 50 * 1024 * 1024;
-/// Default max number of components
-pub const MAX_COMPONENTS: u32 = 10_000;
+fn getenv<T>(key: &str) -> Option<T>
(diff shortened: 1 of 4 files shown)


crates/runtime/src/runtime.rs
```diff
@@ -2,87 +2,161 @@ use crate::ComponentConfig;
...
+        }
+        Err(VarError::NotPresent) => None,
+        Err(VarError::NotUnicode(..)) => {
+            error!("`{key}` value is not valid UTF-8, ignoring");
+            None
+        }
+    }
+}
+
+fn new_pooling_config() -> PoolingAllocationConfig {
```
brooksmtownsend, line 34: Even if hidden on the CLI, it would be awesome to have all of these defined as variables in the wasmcloud binary. Not necessary here, but my motivation would be to display them in our autogenerated documentation

src/main.rs
```diff
@@ -202,19 +202,21 @@ struct Args {
...
     #[clap(long = "max-execution-time-ms", default_value = "600000", env = "WASMCLOUD_MAX_EXECUTION_TIME_MS", value_parser = parse_duration_millis)]
     max_execution_time: Duration,
-    /// The maximum amount of memory bytes that a component can allocate (default 256 MiB)
-    #[clap(long = "max-linear-memory-bytes", default_value_t = 256 * 1024 * 1024, env = "WASMCLOUD_MAX_LINEAR_MEMORY")]
-    max_linear_memory: u64,
-    /// The maximum byte size of a component binary that can be loaded (default 50 MiB)
-    #[clap(long = "max-component-size-bytes", default_value_t = 50 * 1024 * 1024, env = "WASMCLOUD_MAX_COMPONENT_SIZE")]
-    max_component_size: u64,
-    /// The maximum number of components that can be run simultaneously
+    /// This option is deprecated, do not use
```
brooksmtownsend, line 205: Re above comment, this could be made backwards compatible.

For this PR, it would be great if this was more descriptive unless we truly just don't pipe this through. A recommendation to use the other env var would be good

brooksmtownsend (approved):
No blocking things but would definitely prefer these vars to be in a place where we can generate docs for them, and a hint to use the new env var for the deprecated flags

PR #3718, opened 2024-11-29
feat(flake): Adds wash output package to the flake

thomastaylor312:
Talking to @rvolosatovs, we probably should split up individual builds eventually, but I did want to make it easier to install wash from the flake (since I'm working on adding one for wadm, and having wash in the devShell there would be good)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -459,6 +459,10 @@
               ;
 
             rust = hostRustToolchain;
+            wash = pkgs.runCommandLocal "wash" {} ''
+              mkdir -p $out/bin
+              cp ${packages.wasmcloud}/bin/wash $out/bin/wash
+            '';
           };
 
         withDevShells = {


rvolosatovs (commented):
Is there a reason `wasmcloud` output cannot be used directly for this purpose?

I would work exactly the same way as `wit-deps` in this flake, for example:
- https://github.com/wasmCloud/wasmCloud/blob/504705f16ea209e49267ae8393da5363fd163bc4/flake.nix#L40-L42
- https://github.com/wasmCloud/wasmCloud/blob/504705f16ea209e49267ae8393da5363fd163bc4/flake.nix#L479

flake.nix
```diff
@@ -459,6 +459,10 @@
               ;
 
             rust = hostRustToolchain;
+            wash = pkgs.runCommandLocal "wash" { } ''
```
rvolosatovs, line 462: Could you run `nix fmt` ?

rvolosatovs (approved):

thomastaylor312:
That overlay trick is nice. I did have to add the `fenix` overlay in from wasmcloud as well to make it work in wadm (which means I probably missed a magic line somewhere), but that is super helpful to know about

rvolosatovs:
> That overlay trick is nice. I did have to add the `fenix` overlay in from wasmcloud as well to make it work in wadm (which means I probably missed a magic line somewhere), but that is super helpful to know about

Adding `overlays.default` from this flake to overlays will pull in all the necessary dependencies (including `fenix`)

PR #3719, opened 2024-11-29
chore: Pin dockerfile dependencies

joonas:
## Feature or Problem

Pins the image versions used in our Dockerfiles.


rvolosatovs (approved):
Looks like we'll need to set up some automation to keep these up-to-date

joonas:
> Looks like we'll need to set up some automation to keep these up-to-date

Agreed, I’ll do a follow-up PR for that

Issue #2339, opened 2024-06-13
[FEATURE] Auto bump provider references


thomastaylor312:
### Affected project(s)

- [X] documentation
- [X] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the solution you'd like

It would be really nice if we didn't have to manually bump all of our examples to their latest provider versions. See #2336 for example

vados-cosmonic:
Not stale, it would still be nice if we had a way to do this!

markkovari:
If you do not mind a +1 tool in your repository; renovate is capable of this. With a regex in given files it should be able to identify newer OCI images like this https://docs.renovatebot.com/modules/manager/regex/#configuration-templates 🤔 You can assign it to me I'll can give it a try.

vados-cosmonic:
Hey @markkovari thanks for noting that -- this is definitely something we want that people just didn't have the know-how (of renovate, evidently!) to do quickly without writing a script or something. Please take a shot at it!

PR #3727, opened 2024-11-30
chore(renovate): add scheduled renovate to bump providers for examples

markkovari:
## Feature or Problem

Unfortunately dependabot is not flexible enough other than the existing [ecosystems](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem). Although renovate-bot can do that with regular expressions.

Heads-up, you might need to [onboard renovate app](https://docs.renovatebot.com/getting-started/installing-onboarding/) for the repository.

The PRs which will be open should look like this https://github.com/markkovari/wasmCloud/pull/36

- update -> even should create only one PR from now on.

Feel free to edit the name/description/schedule, not sure what is the most convenient.  

## Related Issues

closes: #2339

### DIFF
--- a/.github/workflows/update-example-providers.yml
+++ b/.github/workflows/update-example-providers.yml
@@ -13,9 +13,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4.2.2
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@v41.0.5
+        uses: renovatebot/github-action@936628dfbff213ab2eb95033c5e123cfcaf09ebb # v41.0.5
         with:
           configurationFile: .github/renovate-config.json
           token: ${{ secrets.RENOVATE_TOKEN }}
--- a/.github/workflows/update-example-providers.yml
+++ b/.github/workflows/update-example-providers.yml
@@ -17,5 +17,5 @@ jobs:
       - name: Self-hosted Renovate
         uses: renovatebot/github-action@936628dfbff213ab2eb95033c5e123cfcaf09ebb # v41.0.5
         with:
-          configurationFile: .github/renovate-config.json
+          configurationFile: .github/renovate.json
           token: ${{ secrets.RENOVATE_TOKEN }}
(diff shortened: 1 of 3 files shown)


brooksmtownsend:
@markkovari this looks really cool! I'm not sure I quite understand how it works though, does it just go through and patch bump on a schedule that's separate from when we release providers?

markkovari:
@brooksmtownsend yes, that is kinda exactly how it works
this is the [community cloud](https://github.com/renovatebot/renovate?tab=readme-ov-file#mend-renovate-community-cloud-hosted) version of it, working after enabling the github app to read the repo and create PRs

Unfortunately it only has (AFAIK) a scheduled timer and a manual trigger on their UI, and cannot listen to other events - like releases and other events from GitHub. 😞 but it can keep these references up to date by asking ghcr if a newer version is available.

joonas:
Would it be worth considering the [Renovate GitHub Action](https://github.com/renovatebot/github-action) instead of the App so that we can have more fine-grained control over how and when it's triggered?

markkovari:
Yeah, sure. Will add it soon. Not sure what `on` events should I add, but will add a few and will update based on feedback. 🙇

.github/workflows/update-example-providers.yml
```diff
@@ -0,0 +1,21 @@
...
+  contents: read
+
+jobs:
+  renovate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4.2.2
+      - name: Self-hosted Renovate
+        uses: renovatebot/github-action@v41.0.5
```
joonas, line 18: Would you mind running this workflow through https://app.stepsecurity.io/secureworkflow and grabbing the SHA1s from there instead of the version tags?

joonas (approved):
Looking good 👍

markkovari:
Well it is acting a bit silly on my fork for some reason. Let me debug/fix it there, before we have to do it here. 🤔

markkovari:
@joonas sorry for the delay, did some testing on my fork and unfortunately without the `repositories` it was not scanning the repository, and the config file was not fount - beside the input explicitly added for the action now it should be working.

joonas:
@markkovari `repositories` seems totally reasonable and in fact something we would likely want, so that doesn't seem too bad of an addition.

That does bring up a question about whether we should limit running this to pjust `wasmCloud/wasmCloud` with an if statement](https://github.com/wasmCloud/wasmCloud/blob/main/.github/workflows/update.yml#L13), because I don't see how this would be useful to have run in forks.

Would you mind adding that to the workflow?

brooksmtownsend (approved):

PR #3728, opened 2024-11-30
chore(deps): Add dependabot for Dockerfiles

joonas:
## Feature or Problem

Follow-up on https://github.com/wasmCloud/wasmCloud/pull/3719#pullrequestreview-2470539904

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -48,3 +48,7 @@ updates:
   directory: "/.github/actions/install-nix"
   schedule:
     interval: "daily"
+- package-ecosystem: "docker"
+  directory: "/"
+  schedule:
+    interval: "daily"


.github/dependabot.yml
```diff
@@ -48,3 +48,7 @@ updates:
   directory: "/.github/actions/install-nix"
   schedule:
     interval: "daily"
+- package-ecosystem: "docker"
+  directory: "/"
```
rvolosatovs, line 52: Will this pick up nested dockerfiles as well? At least for GitHub actions, one needs to manually specify path to each reusable action directory for that (I forget the exact term they used to call these actions)

rvolosatovs (approved):

.github/dependabot.yml
joonas: By nested Dockerfiles, do you mean multi-stage builds (aka Dockerfiles with multiple different image definitions in one)?

If so, it does appear that [multi-stage builds are not entirely supported](https://github.com/dependabot/dependabot-core/issues/7640), but I'd say let's get this merged in and then figure out how we want to iterate.

rvolosatovs: I meant Dockerfiles in directories nested under `/`. I don't see any Dockerfiles in `/`

markkovari: That is indeed an interesting question (sorry for piggy-packing the question) did some digging over the weekend, because of this if dependabot can be used or not. 

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories

In the documentation they don't really talk about the way the walk the path. But there is one example which says:

```yaml
# Specify all directories from the current layer and below recursively, 
# using globstar, for locations of manifest files

version: 2
updates:
  - package-ecosystem: "composer"
    directories:
      - "**/*"
    schedule:
      interval: "weekly"
```

So it might be non-recursive too, will read the code - since the docs does not really mention-, but with glob it should be good.

Whoever would update it, do we need [this](https://github.com/wasmCloud/wasmCloud/blob/174d9628f3297884d8815d86c00f622045859fef/.github/dependabot.yml#L11-L14)? This path is no longer there. 🤔

PR #3739, opened 2024-12-02
build: rework `wasmcloud` crate features

rvolosatovs:
- remove the unused `lattice-controler` provider binary. In it's current state it does not do anything and it's trivial to bring it back once it works.
- remove `providers` feature - instead introduce a feature flag per provider
- introduce `wasmcloud` feature, which gates the host - this allows building providers without building the host
- remove unused `redis` and `wasmcloud-provider-sdk` features

This change set will act as the first step in speeding up CI


brooksmtownsend (approved):
Huge improvements here for individual developer loops 🎉 

We should consider a good-first-issue to enable / disable Rust features in the Rust section of the wasmcloud.toml to fully take advantage of this

PR #3743, opened 2024-12-02
feat: Enable memory_init_cow in wasmtime

joonas:
## Feature or Problem

Before:
current-host-using-builtin-messaging
After:
latest-main-with-memory_init_cow

### DIFF
--- a/crates/runtime/src/runtime.rs
+++ b/crates/runtime/src/runtime.rs
@@ -35,7 +35,6 @@ impl RuntimeBuilder {
         let mut engine_config = wasmtime::Config::default();
         engine_config.async_support(true);
         engine_config.epoch_interruption(true);
-        engine_config.memory_init_cow(false);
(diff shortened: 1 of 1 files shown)


crates/runtime/src/runtime.rs
```diff
@@ -35,7 +35,7 @@ impl RuntimeBuilder {
         let mut engine_config = wasmtime::Config::default();
         engine_config.async_support(true);
         engine_config.epoch_interruption(true);
-        engine_config.memory_init_cow(false);
+        engine_config.memory_init_cow(true);
```
rvolosatovs, line 38: Let's not set anything here and use the default instead?

joonas: You're absolutely right, that's the better way to go here!

rvolosatovs (approved):

PR #3750, opened 2024-12-03
fix(examples): build dependent component for custom providers

vados-cosmonic:

### DIFF
--- a/.github/workflows/examples-providers.yml
+++ b/.github/workflows/examples-providers.yml
@@ -130,6 +130,12 @@ jobs:
         run: wash build
         working-directory: examples/${{ matrix.project.lang }}/providers/${{ matrix.project.name }}
 
+      # Build required project if necessary
+      - name: Build dependent component if present
+        shell: bash
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/examples-providers.yml
```diff
@@ -130,6 +130,12 @@ jobs:
         run: wash build
         working-directory: examples/${{ matrix.project.lang }}/providers/${{ matrix.project.name }}
 
+      # Build required project if necessary
+      - name: Build dependent component if present
+        shell: bash
+        run: |
+          [[ -d component ]] && wash dev -p component/wasmcloud.toml
```
joonas, line 137: Should this be `wash build` instead of `wash dev`?
```suggestion
          [[ -d component ]] && wash build -p component/wasmcloud.toml
```

joonas (approved):

.github/workflows/examples-providers.yml
vados-cosmonic: ah sorry you're a bit early, change was still going in

PR #3765, opened 2024-12-04
feat: `wasmcloud:messaging@0.3.0`

rvolosatovs:
This adds support for `wasmcloud:messaging@0.3.0`, which is effectively just renamed https://github.com/WebAssembly/wasi-messaging/commit/4ee59bb881f7a8675fbb3ae7ca5d5d586ab88657

Tested with a newly-added `wasmcloud:messaging@0.3.0` example manually

Note, that wadm, wash or something in-between does not support `incoming-handler` interface names for links, but that does not really matter, because these interface names do not appear to ever really be used by providers

Note: binary providers are expected to use `wasmcloud:messaging@0.2.0` for the time being


ricochet (approved):

crates/runtime/src/component/messaging/v0_3.rs
```diff
@@ -0,0 +1,476 @@
...
+            .context("failed to delete client")?;
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl<H> types::HostMessage for Ctx<H>
+where
+    H: Handler,
+{
```
brooksmtownsend, line 204: For all of these functions, we should also `self.attach_parent_context()`

crates/runtime/src/component/messaging/v0_3.rs
```diff
@@ -0,0 +1,476 @@
...
+    Guest(GuestMessage),
+}
+
+impl<H> types::Host for Ctx<H> where H: Handler {}
+
+#[async_trait]
+impl<H> types::HostClient for Ctx<H>
+where
+    H: Handler,
+{
```
brooksmtownsend, line 158: For all of these functions, we should also `self.attach_parent_context()`

crates/runtime/src/component/messaging/v0_3.rs
```diff
@@ -0,0 +1,476 @@
...
+    }
+}
+
+#[async_trait]
+impl<H> producer::Host for Ctx<H>
+where
+    H: Handler,
+{
+    #[instrument(level = "debug", skip_all)]
+    async fn send(
```
brooksmtownsend, line 364: We should also `self.attach_parent_context()` here

crates/runtime/src/component/messaging/v0_3.rs
```diff
@@ -0,0 +1,476 @@
...
+    H: Handler,
+{
+    async fn request(
+        &mut self,
+        client: Resource<Box<dyn Client + Send + Sync>>,
+        topic: Topic,
+        message: Resource<Message>,
+        options: Option<Resource<RequestOptions>>,
+    ) -> wasmtime::Result<Result<Vec<Resource<Message>>, Error>> {
+        let options = options
```
brooksmtownsend, line 391: We should also self.attach_parent_context() here

crates/host/src/wasmbus/handler.rs
```diff
@@ -456,6 +460,642 @@ impl Messaging for Handler {
...
+                        .into_any()
+                        .downcast::<Message>()
+                        .map_err(|_| anyhow!("unknown message type"))?;
+                    match *message {
+                        Message::Nats(async_nats::Message {
+                            headers: Some(..), ..
+                        }) => {
+                            return Ok(Err(messaging0_3_0::types::Error::Other(
+                                "headers not currently supported by wRPC targets".into(),
+                            )));
```
brooksmtownsend, line 691: This seems like a pretty common case to use headers on the message for 🤔 What's the reasoning for this? Is it the same as why we need to specify headers in the client in that it adjusts the payload size for wRPC?

crates/host/src/wasmbus/handler.rs
```diff
@@ -456,6 +460,642 @@ impl Messaging for Handler {
...
+    }
+
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+}
+
+pub enum Message {
+    Nats(async_nats::Message),
+}
```
brooksmtownsend, line 480: I was surprised to see this as a concrete implementation of _just_ NATS here, I expected to see the same pattern we use elsewhere where we send a wRPC message to some wasmcloud:messaging@0.3.0 provider. Is that not applicable for this new interface, or is this just a sample implementation?

brooksmtownsend (changes requested):
I have a few asks re: OTEL and then a potentially blocking question regarding the implementation of the 0.3 messaging interface. It makes sense that it works, but I wasn't sure if it should be tightly coupled to NATS

crates/host/src/wasmbus/handler.rs
rvolosatovs: `wasmcloud:messaging@0.2.0`, used on wRPC level, does not support headers: https://github.com/wasmCloud/messaging/blob/ff300523cfdaf5eab19a9385637ae4665cc47d86/wit/messaging.wit#L1-L27

If we want to use wRPC for binary providers, then we'll need a `wrpc:messaging` defined first, which would include support for headers

crates/host/src/wasmbus/handler.rs
rvolosatovs: Both approaches are supported, this is a message originating from a builtin provider.

rvolosatovs:
To make sure we're on the same page: implementation is not tightly coupled to NATS.io, it's fully generic

crates/host/src/wasmbus/handler.rs
```diff
@@ -456,6 +460,644 @@ impl Messaging for Handler {
...
+                        body,
+                        ..
+                    }) => nats.publish(topic, body).await,
+                    messaging0_3_0::types::Message::Guest(MessagingGuestMessage0_3 {
+                        content_type,
+                        data,
+                        metadata,
+                    }) => {
+                        if content_type.is_some() {
+                            return Ok(Err(messaging0_3_0::types::Error::Other(
```
thomastaylor312, line 655: I feel like this should just warn rather than erroring out. It isn't a problem if they set content type. The other thing we could do is just transform it into a header for them. I'd be good with either just warning or trying to be smart

rvolosatovs:
Changed `content-type` errors to `warn` logs and filed an issue upstream https://github.com/WebAssembly/wasi-messaging/issues/30

Also filed this PR https://github.com/WebAssembly/wasi-messaging/pull/29 to avoid having to trap when content-type is not supported

brooksmtownsend (approved):
Good for me

crates/host/src/wasmbus/handler.rs
brooksmtownsend: Thanks for the clarity!

PR #3772, opened 2024-12-05
fix(wash-cli): dont require NATS to start

brooksmtownsend:
## Feature or Problem
This PR fixes a bug where we attempt to create a NATS and control interface client before wash dev actually starts NATS, essentially requiring NATS to be running when the process starts.

Instead, now we only require NATS to be running if attempting to connect to an existing host.


vados-cosmonic (approved):
LGTM :rocket:

PR #3779, opened 2024-12-05
feat(wash-cli): support expanded wkg configuration

vados-cosmonic:


crates/wash-cli/src/cmd/wit/build.rs
```diff
@@ -0,0 +1,83 @@
...
+    let (pkg_ref, version, bytes) =
+        wit::build_package(&wkg_config, dir, &mut lock_file, client).await?;
+
+    // Resolve the output path
+    let output_path = match output_file {
+        Some(path) => path,
+        None => {
+            let mut file_name = pkg_ref.to_string();
+            if let Some(ref version) = version {
+                file_name.push_str(&format!("@{version}"));
```
thomastaylor312, line 61: I think the `@` symbol is forbidden on windows. I got away wit separating things with `-` instead. This is something I need to fix in `wkg` as well so just thought I'd give you a heads up

crates/wash-cli/src/cmd/wit/deps.rs
```diff
@@ -0,0 +1,321 @@
...
+    };
+
+    // Extract/build configuration for wkg
+    let mut wkg_config = match project_config {
+        Some(ProjectConfig {
+            ref package_config, ..
+        }) => package_config.clone(),
+        None => wasm_pkg_core::config::Config::load().await?,
+    };
+    let mut wkg_config_overrides = wkg_config.overrides.take().unwrap_or_default();
```
thomastaylor312, line 67: Nit: Might be worth adding a comment why you're temporarily removing this here

crates/wash-cli/src/cmd/wit/deps.rs
```diff
@@ -0,0 +1,321 @@
...
+            RegistryPullSource::Builtin => bail!("no builtins are supported"),
+            RegistryPullSource::RemoteHttp(s) => {
+                let url = Url::parse(s)
+                    .with_context(|| format!("invalid registry pull source url [{s}]"))?;
+                let tempdir = tempfile::tempdir()
+                    .with_context(|| format!("failed to create temp dir for downloading [{url}]"))?
+                    .into_path();
+                let output_path = tempdir.join("unpacked");
+                let http_client = wash_lib::start::get_download_client()?;
+                let req = http_client.get(url.clone()).send().await.with_context(|| {
```
thomastaylor312, line 178: Super nit: Maybe we should set an `Accept` header here to tell the server we only acknowledge tarballs?

crates/wash-cli/src/cmd/wit/deps.rs
```diff
@@ -0,0 +1,321 @@
...
+                    format!("failed to retrieve WIT output from URL [{}]", &url)
+                })?;
+                let mut archive = tokio_tar::Archive::new(GzipDecoder::new(
```
thomastaylor312, line 181: Before this step, we should probably check the content-type header to check if the incoming type is `application/x-gzip` or `application/gzip` so we can fail early with a friendly message.

crates/wash-cli/src/cmd/wit/deps.rs
```diff
@@ -0,0 +1,321 @@
...
+                            wkg_config_overrides.insert(
+                                format!("{ns}:{pkg}"),
+                                Override {
+                                    path: Some(path.clone()),
+                                    version: None,
+                                },
+                            );
+                        }
+                    }
+                }
```
thomastaylor312, line 167: Super nit: This same block is used several times copy pasted. Feels like some of it could be abstracted out into a helper function

crates/wash-cli/tests/fixtures/integrated-wkg-extended/wasmcloud.toml
```diff
@@ -0,0 +1,22 @@
...
+target = "test-components:testing"
+source = "file://extra-wit/pingpong.wit"
+
+[[registry.pull.sources]]
+target = "wasi:config"
+source = "https://github.com/WebAssembly/wasi-config/archive/v0.2.0-draft.tar.gz"
+
+[[registry.pull.sources]]
+target = "wasi:blobstore"
+source = "git+https://github.com/WebAssembly/wasi-blobstore.git"
```
thomastaylor312, line 22: For testing, we should probably make sure this is going to pull from a pinned SHA to avoid any issues down the line

crates/wash-cli/tests/fixtures/integrated-wkg-extended/wasmcloud.toml
```diff
@@ -0,0 +1,22 @@
+name = "ponger-config-component"
+language = "rust"
+type = "component"
+
+[component]
+wasm_target = "wasm32-wasip2"
+
+[[registry.pull.sources]]
+target = "wasmcloud:bus"
+source = "oci://ghcr.io/wasmcloud/interfaces"
```
thomastaylor312, line 10: Should also pin a version here too

crates/wash-cli/tests/wash_wit.rs
```diff
@@ -91,3 +92,76 @@ async fn test_wash_wit_wkg_override() {
...
+            "bad-pull-source.wasmcloud.toml",
+        ])
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .output()
+        .await
+        .expect("Failed to fetch dependencies");
+    assert!(!output.status.success(), "Failed to fetch dependencies");
+    Ok(())
+}
```
thomastaylor312, line 167: Just given some of the history we've had with rolling this out, we may want to have a few other tests like one that uses nested deps (will need to wait for the upstream wkg PR to merge so don't worry about that here) and failure cases for all of the sources:

- not a tarball
- no wit directory
- Non-pinned OCI or git reference
- Pinned OCI or git reference

crates/wash-lib/src/common.rs
```diff
@@ -250,3 +256,98 @@ pub async fn get_all_inventories(
...
+    tmp_dir: impl AsRef<Path>,
+    repo_url: String,
+    sub_folder: Option<String>,
+    repo_branch: Option<String>,
+) -> Result<()> {
+    let git_cmd = git_cmd.unwrap_or_else(|| DEFAULT_GIT_PATH.into());
+    let tmp_dir = tmp_dir.as_ref();
+    let cwd =
+        std::env::current_dir().map_err(|e| anyhow!("could not get current directory: {}", e))?;
+    std::env::set_current_dir(tmp_dir)
```
thomastaylor312, line 272: Rather than mucking with the current dir, why don't we just pass `tmp_dir` as an argument to `git clone`?

thomastaylor312: The other option, which after reading the code might be easier, is to set the cwd of the process (i.e. `cmd.current_dir`) on each of the commands

crates/wash-lib/src/parser/mod.rs
```diff
@@ -138,13 +140,222 @@ pub struct RustConfig {
...
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        Ok(match s {
+            s if s.starts_with("file://") => Self::LocalPath(s.into()),
+            s if s.starts_with("oci://") => Self::RemoteOci(s.into()),
+            s if s.starts_with("http://") || s.starts_with("https://") => {
+                Self::RemoteHttp(s.into())
+            }
+            s if s.starts_with("git+ssh://")
```
thomastaylor312, line 269: In the other code I saw the shortcut with leaving off `github.com`. Does that need to be handled here or is it really only for `wash new`?

thomastaylor312 (changes requested):
Nothing super major here, just some minor fixes and request for additional tests. Great work on mapping everything into the wkg config

crates/wash-cli/src/cmd/wit/build.rs
vados-cosmonic: Note that this code was moved -- [I didn't actually change this!](https://github.com/wasmCloud/wasmCloud/blob/0af83f4148833b0c5187473b8572e093928aef22/crates/wash-cli/src/wit.rs#L112)

Also one thing I did notice is that there was no way to tell where the namespace started and the package ended -- can you also change that upstream while you're there? i.e. `test-components__testing__0.1.0` would be nicer to have than `test-components-testing-0.1.0`

vados-cosmonic: Here's a [new issue to track this ](https://github.com/wasmCloud/wasmCloud/issues/3799) when upstream changes

crates/wash-cli/src/cmd/wit/deps.rs
vados-cosmonic: Sure -- going to just move this into the helper function!

crates/wash-cli/src/cmd/wit/deps.rs
vados-cosmonic: Yeah we have no guarantee the server on the other end will actually honor that, so feels like we might be adding a false sense of safety with that... Another problem is that we don't actually know the server will get it's content type right (it *should*, but it's the internet here...)

Maybe we can add this when someone who has `Accept` parsing as a gate shows up?

crates/wash-cli/src/cmd/wit/deps.rs
vados-cosmonic: I don't think it's a given that this will be properly set -- what I *did* consider was checking the magic bytes of the downloaded bytes, BUT I think that's sufficiently covered by the error message when we try to unpack it. 

If things fail to unpack, the user *will* get a nice error message, including the URL we tried to download it from.

crates/wash-cli/src/cmd/wit/deps.rs
vados-cosmonic: Sure! The first 3 uses are similar but the last 1 is not actually the same -- right on the border. Will update the first 3

crates/wash-lib/src/common.rs
vados-cosmonic: This is a great idea, but IMO this should be handled separately. The code that does it this way has been there for a long time, and IMO how this whole function works should be changed. 

We shouldn't IMO be getting the current dir *inside* this function in the first place.

vados-cosmonic: Made an issue for this! https://github.com/wasmCloud/wasmCloud/issues/3800

crates/wash-lib/src/parser/mod.rs
vados-cosmonic: Nope, I don't think we should allow leaving off github.com here -- people should be explicit about the git repository they're accessing.

crates/wash-cli/tests/wash_wit.rs
vados-cosmonic: > uses nested deps (will need to wait for the upstream wkg PR to merge so don't worry about that here)

Could you make an issue for this so that we can track it to make sure we get back to it?

I'll work on adding the other tests you noted!

> not a tarball

If we get something that's not a tarball (and trying to unpack it fails), the error message is pretty clear -- are you hoping to catch the specific error we get out in the test?

> no wit directory

Note that the WIT directory is only searched for -- I don't think we need a test for this, because the failure is invalid input (to the upstream functions) at that point (i.e. `wkg` has been pointed to something that just isn't WIT)

Are you hoping to catch a specific error message from the `wkg` side?

> Non-pinned OCI or git reference
> Pinned OCI or git reference

By "pinned" you mean with/out a version specified right?

Right now the tests for the OCI and git references are unpinned -- so was the idea here to add:

- `import ns:pkg/interface` in the WIT
- `ns:pkg@0.2.0` in the config

Is that what you were thinking? Or is this more about enabling version overwrite support?

crates/wash-cli/tests/fixtures/integrated-wkg-extended/wasmcloud.toml
vados-cosmonic: So supporting pinning a specific version would be great, but there isn't actually support for that upstream, AFAIK, am I wrong about that? The usage here is going through `wkg` as normal, and [`PackageRef`](https://docs.rs/wasm-pkg-common/0.8.5/wasm_pkg_common/package/struct.PackageRef.html) has no way to specify a version actually... 

Obviously, it's trivial to add a `?tag=x.x.x` query param or something like that to enable the support on our side (I don't think there's a standard for this?), but even if we did, we need upstream support to be able to actually narrow this down, no?

crates/wash-cli/tests/fixtures/integrated-wkg-extended/wasmcloud.toml
vados-cosmonic: The way this works is that it actually runs `git clone` -- so is the idea here to add a `?sha=<....>`  or to have some sort of URI encoding that exposes a SHA we should set ourselves to later?

In the case that you want a specific point of a git repo I think you'd be trying to use a regular `https://` target (like we do above for the `wasi-config` at `v0.2.0-draft` tag...

Maybe I'm misunderstanding, what's the code that you would replace this line with?

crates/wash-cli/src/cmd/wit/build.rs
thomastaylor312: Oh I figured this was me having mostly copy pasted from wkg. Was just giving a heads up

crates/wash-cli/tests/fixtures/integrated-wkg-extended/wasmcloud.toml
thomastaylor312: Yeah, what will happen here with the test if we don't checkout a specific sha is that we could make a change on `main` and then the test will fail because the wit changed. Ideally, when we fold this into main `wkg`, the git sha it picks on clone would be written to the lock file. But right now, this will always pull from the latest main, which is probably not desired behavior. Adding the `sha` param was the solution I was thinking of (which can be a commit sha or a tag or a branch name).

crates/wash-cli/tests/fixtures/integrated-wkg-extended/wasmcloud.toml
thomastaylor312: Sorry yep, you're right. The main wkg client will use this as the base repo and then calculate version from there

crates/wash-cli/tests/wash_wit.rs
thomastaylor312: > Could you make an issue for this so that we can track it to make sure we get back to it?

Wkg 0.9 is now out and I think @brooksmtownsend is going to update to it. Maybe the tests there are enough.

> If we get something that's not a tarball (and trying to unpack it fails), the error message is pretty clear -- are you hoping to catch the specific error we get out in the test?

Yeah, just a smoke test to make sure it errors properly

> Right now the tests for the OCI and git references are unpinned 

After out convo above, OCI won't be necessary, but testing a pinned git sha would be good. Maybe a test that pins the right commit and one that pins an older one with incompatible wit?

As for the "no wit directory," I think you're right that it is good enough because the tooling will take care of that. So with that said, I think tests we should add are

- Smoke test for pulling something that isn't a tarball
- Good and bad SHA pins to prove that pinning works

crates/wash-cli/tests/fixtures/integrated-wkg-extended/wasmcloud.toml
vados-cosmonic: Done -- make sure to add this, and made it a bit more explicit

crates/wash-cli/tests/wash_wit.rs
vados-cosmonic: Hey added these tests, thanks for the suggestion!

thomastaylor312 (approved):

PR #3783, opened 2024-12-06
feat(nats): add JetStream consumer support

rvolosatovs:
This adds support for JetStream consumers in builtin NATS messaging provider. The new `consumers` config key takes in in-line JSON, here's an example:
```diff
diff --git a/examples/rust/components/echo-messaging/local.wadm.yaml b/examples/rust/components/echo-messaging/local.wadm.yaml
index 078eb5f5..59973c68 100644
--- a/examples/rust/components/echo-messaging/local.wadm.yaml
+++ b/examples/rust/components/echo-messaging/local.wadm.yaml
@@ -32,7 +32,7 @@ spec:
     - name: nats
       type: capability
       properties:
-        image: ghcr.io/wasmcloud/messaging-nats:0.23.1
+        image: wasmcloud+builtin://messaging-nats
         ## To configure OTEL integration for this provider specifically, uncomment the lines below
         # config:
         #   - name: otel
@@ -57,4 +57,4 @@ spec:
             source_config:
               - name: simple-subscription
                 properties:
-                  subscriptions: wasmcloud.echo
+                  consumers: '[{"stream":"echo","consumer":"example"}]'
```

This is what I used for testing after adding a stream and a consumer manually.

Of course, replies do not work out-of-the-box with this, since it's used by JetStream, so custom application-specific logic must exist to figure out the reply semantics for requests (e.g. a custom header)

### DIFF
--- a/crates/host/src/wasmbus/providers/messaging_nats.rs
+++ b/crates/host/src/wasmbus/providers/messaging_nats.rs
@@ -2,16 +2,17 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use anyhow::{anyhow, bail, Context as _};
+use async_nats::jetstream;
 use futures::StreamExt;
 use nkeys::{KeyPair, XKey};
 use tokio::fs;
 use tokio::sync::{broadcast, Mutex, RwLock};
 use tokio::task::JoinSet;
 use tokio::time::Instant;
-use tracing::{error, info_span, instrument, trace_span, warn, Instrument as _, Span};
+use tracing::{debug, error, instrument, trace_span, warn, Instrument as _, Span};
 use wasmcloud_core::InterfaceLinkDefinition;
-use wasmcloud_provider_messaging_nats::add_tls_ca;
 use wasmcloud_provider_messaging_nats::ConnectionConfig;
+use wasmcloud_provider_messaging_nats::{add_tls_ca, ConsumerConfig};
 use wasmcloud_provider_sdk::provider::{
     handle_provider_commands, receive_link_for_provider, ProviderCommandReceivers,
 };
@@ -52,7 +53,7 @@ impl Provider {
             (Some(jwt), Some(seed)) => {
                 let seed = KeyPair::from_seed(seed).context("failed to parse seed key pair")?;
                 let seed = Arc::new(seed);
-                async_nats::ConnectOptions::with_jwt(jwt.clone(), move |nonce| {
+                async_nats::ConnectOptions::with_jwt(jwt.to_string(), move |nonce| {
                     let seed = seed.clone();
                     async move { seed.sign(&nonce).map_err(async_nats::AuthError::new) }
                 })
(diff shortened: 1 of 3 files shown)


crates/host/src/wasmbus/providers/messaging_nats.rs
```diff
@@ -113,11 +175,69 @@ impl wasmcloud_provider_sdk::Provider for Provider {
...
+            } else {
+                sub
+            };
+            let mut sub = sub.messages().await.context("failed to subscribe")?;
+
+            let components = Arc::clone(&self.components);
+            let lattice_id = Arc::clone(&self.lattice_id);
+            let host_id = Arc::clone(&self.host_id);
+            let target_id = Arc::clone(&target_id);
+            tasks.spawn(async move {
```
brooksmtownsend, line 214: When the component stops, or is unlinked from the provider, we should probably cancel this task so that we don't keep handling messages. Perhaps we can store it as a part of the messaging provider struct and then remove it if the link is deleted?

crates/host/src/wasmbus/providers/messaging_nats.rs
```diff
@@ -126,79 +246,22 @@ impl wasmcloud_provider_sdk::Provider for Provider {
...
-            let target_id = Arc::clone(&target_id);
+            let components = Arc::clone(&self.components);
             let lattice_id = Arc::clone(&self.lattice_id);
             let host_id = Arc::clone(&self.host_id);
-            let components = Arc::clone(&self.components);
+            let target_id = Arc::clone(&target_id);
             tasks.spawn(async move {
                 while let Some(msg) = sub.next().await {
-                    let target_id = Arc::clone(&target_id);
-                    let lattice_id = Arc::clone(&lattice_id);
```
brooksmtownsend, line 139: While we're at it, we should be removing the subscription here too. If we want to do that in a followup because it's not a part of this PR that's fine, but it might just be the same code

brooksmtownsend (changes requested):
One request to make sure we're not keeping interest in a stream longer than we should be

crates/host/src/wasmbus/providers/messaging_nats.rs
rvolosatovs: That's already what happens - the task is cancelled when link is deleted https://github.com/wasmCloud/wasmCloud/blob/9926109f0e3825cd3e28e53469fa6e23d41dbba7/crates/host/src/wasmbus/providers/messaging_nats.rs#L218-L222

Component stops are unrelated here - the link can be established before it even starts running

crates/host/src/wasmbus/providers/messaging_nats.rs
rvolosatovs: Already done

rvolosatovs:
> One request to make sure we're not keeping interest in a stream longer than we should be

that's already how it works

crates/host/src/wasmbus/providers/messaging_nats.rs
brooksmtownsend: My brain did not parse `tasks.spawn` as going into the JoinSet, my mistake 😄

brooksmtownsend (approved):

PR #3786, opened 2024-12-07
chore(examples): Use 0.0.0.0 instead of 127.0.0.1

joonas:
## Feature or Problem

Since these examples get invariably copied and used for services that are deployed outside of localhost, it is in our best interest to use `0.0.0.0` instead of `127.0.0.1` to set people up for success rather than disappointment.

### DIFF
--- a/examples/c/components/http-hello-world/local.wadm.yaml
+++ b/examples/c/components/http-hello-world/local.wadm.yaml
@@ -26,7 +26,7 @@ spec:
         # Establish a unidirectional link from this http server provider (the "source")
         # to the `http-component` component (the "target") so the component can handle incoming HTTP requests,
         #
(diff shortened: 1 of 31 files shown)


rvolosatovs (approved):
Nice, LGTM!
What do you think about using `::` as the default? (Assuming wasmCloud actually supports IPv6)

joonas:
> What do you think about using `::` as the default? (Assuming wasmCloud actually supports IPv6)

While I'm not opposed to supporting ipv6 (it is the future after all), I worry that defaulting to `::` might end up confusing folks given that ipv6 has not quite yet received the kind of widespread adoption that one might hope.

I would prefer to stick to `0.0.0.0` for now, and revisit the ipv6 conversation separately, because it's certainly worth having (and testing).

brooksmtownsend (approved):

Issue #3705, opened 2024-11-27
[FEATURE] Propagate trace from incoming HTTP and to outgoing HTTP


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

We don't currently try to read trace context from incoming HTTP requests, nor do we propagate trace context on outgoing HTTP requests.

### Describe the solution you'd like

Since we propagate trace context through the lifetime of a request inside of wasmCloud, we should also check for the builtin HTTP server and HTTP server provider if there are incoming variables on the headers. Additionally, when making outgoing HTTP requests, we should make sure to propagate that context to the outgoing http headers

### Additional context

https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/tracing-http-propagator/src/server.rs is a good example of how to get this to work. This will likely require making sure that all the opentelemetry libraries are on the same common compatible version.

Patrick0308:
Please assign it to me

brooksmtownsend:
You got it @Patrick0308 ! Feel free to open a PR once you're ready 👍🏻

PR #3788, opened 2024-12-08
Propagate trace from incoming HTTP and to outgoing HTTP

Patrick0308:
## Feature or Problem

Propagate trace from incoming HTTP and to outgoing HTTP

## Related Issues

#3705


Patrick0308:
```
warning: unused import: `http::HeaderMap`
  --> crates/provider-http-server/src/lib.rs:37:5
   |
37 | use http::HeaderMap;
   |     ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

error[E0308]: mismatched types
   --> crates/provider-http-server/src/lib.rs:154:30
    |
154 |             &HeaderExtractor(req.headers()),
    |              --------------- ^^^^^^^^^^^^^ expected `http::header::map::HeaderMap`, found `http::HeaderMap`
    |              |
    |              arguments to this struct are incorrect
    |
    = note: `http::HeaderMap` and `http::header::map::HeaderMap` have similar names, but are actually distinct types
```

I use the old version of opentelemetry-http, but  opentelemetry-http dependent the old http version. Maybe we should update all opentelemetry sdk to latest, the opentelemetry-http dependent the newer opentelemetry sdk which is incompatible.

Patrick0308:
I choose not to use the opentelemetry-http public library.

brooksmtownsend (approved):
@Patrick0308 great work here, this looks nice and simple!

Do you have an example trace that you tested this with that you could share? I can give it a shot locally if not, just want to confirm that this worked as expected 😄 In the meantime, going to request that another @wasmCloud/observability-maintainers gives this a look before merge

brooksmtownsend:
I don't personally have an issue with the approach taken here and not depending on the upstream library. Perhaps we could use that in the future, but I understand the many different versions of the otel libraries creates difficulties.

crates/tracing/src/lib.rs
```diff
@@ -15,6 +15,7 @@ use wasmcloud_core::OtelConfig;
 
 #[cfg(feature = "otel")]
 pub mod context;
+pub mod http;
```
joonas, line 18: Can we put this behind `#[cfg(feature = "otel")]` as well since it uses OTEL libraries?

thomastaylor312 (approved):

PR #3820, opened 2024-12-12
fix: detect host_id from structured logs

lachieh:


joonas:
This seems horribly hacky, I wonder if it would be better to rethink the approach around having the wasmCloud host expose an API that wash could call for this information instead?

vados-cosmonic:
So it *is* horribly hacky, but this is how it's been done up until now because hosts generate their own keys and the only way we expose that is via logging it (right now). 

There are many ways to improve this but at the very least two ways make sense (to me at least):
- Allow providing keys to the host
  - We can output the public key from `wash up` directly if started that way
- Enable writing out the generated key to somewhere on disk

I have suggested both of these in the past, and I still think they're good ideas, there was resistance to the idea before but IMO it's worth doing.

I'm assuming this PR is here though because the code in question broke somehow -- is that right @lachieh ?

lachieh:
Yes, apologies for the lack of description.

`wash dev --enable-structured-logging` will currently start a host, fail to read `host_id="` from the logs, and then die, leaving the host still running.

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -394,11 +394,16 @@ impl WashDevSession {
                         .await
                         .context("failed to read line from file")?
                     {
-                        if let Some(host_id) = line
-                            .split_once("host_id=\"")
-                            .map(|(_, rhs)| &rhs[0..rhs.len() - 1])
-                        {
-                            return Ok(host_id.to_string()) as anyhow::Result<String>;
+                        let re = regex::Regex::new(if wasmcloud_opts.enable_structured_logging {
```
vados-cosmonic, line 397: We could set this up somewhere earlier since we don't depend on the actual line

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -394,11 +394,16 @@ impl WashDevSession {
...
+                            r#"\"host_id\":\s?\"([A-Z0-9]+)\""#
+                        } else {
+                            r#"host_id="([A-Z0-9]+)"#
+                        })
+                        .context("failed to compile regex")?;
+
+                        if let Some(captures) = re.captures(&line) {
+                            return Ok(captures.get(1).unwrap().as_str().to_string())
```
vados-cosmonic, line 405: Surprised you didn't get asked to use `first()` or `next()` or something -- and it would be nice if we didn't do the `unwrap()` here -- `.context(...)` should work. 

Also, it *feels* like we shouldn't need `as_str().to_string()` -- does the capture not implement `ToString`?

lachieh: I can replace the if with the following, though I'm not sure it's better. The `.get(1)` will only fail if there isn't a capture group, which both patterns have, and the result of `.get()` is `Option<Match>`.
```rs
                        return Ok(re
                            .captures(&line)
                            .context("failed to capture host ID")?
                            .get(1)
                            .context("failed to get capture group")?
                            .as_str()
                            .to_string()) as anyhow::Result<String>;
```

I could also do this, which _should_ be fine since both patterns have at least 1 capture group:
```rs
                        return Ok(
                            re.captures(&line).context("failed to capture host ID")?[1].to_string()
                        ) as anyhow::Result<String>;
```

Any preferences?

lachieh:
@joonas you ok with the hack if it's just to fix a bug? I can log an issue for a larger refactor to keep the improvement tracked.

joonas:
> @joonas you ok with the hack if it's just to fix a bug? I can log an issue for a larger refactor to keep the improvement tracked.

that sounds great 👍

crates/wash-cli/src/cmd/dev/session.rs
lachieh: (I pushed up the latter, but happy to change)

vados-cosmonic: Yeah based on what `captures()`'s written contract it should be fine... But I'd probably still do the former even though it's a little bit more verbose just to have graceful error handling there

vados-cosmonic (approved):
LGTM :rocket:

crates/wash-cli/src/cmd/dev/session.rs
vados-cosmonic: Oh checking now, but you pushed up the *former*, so happy with that!

lachieh: ACK - switched. It'll catch the future case where the regex capture group gets removed.

PR #3822, opened 2024-12-12
fix(renovate): migrate config file and disable dependency dashboard

markkovari:
## Feature or Problem

I presume we do not want the dependency dashboard feature so I disabled it, and migrated to the `config:recommended` instead of `config:base` extension.

related to: #3819

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,11 +1,12 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
     "extends": [
-        "config:base"
+        "config:recommended"
     ],
     "enabledManagers": [
         "custom.regex"
     ],
+    "dependencyDashboard": false,
     "repositories": [
         "wasmCloud/wasmCloud"
     ],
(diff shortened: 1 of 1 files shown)


joonas:
@markkovari it's not clear to me what this difference of configuration does exactly, would you be able to elaborate?

markkovari:
Well it was mentioned in the verbose logs, to migrate this part and I followed the instructions as a good soldier. 

Although we do not want these features to be enabled, most of them should be irrelevant (workaround and whatnot) others should be disabled by us.

Going to remove it, although the logs were pretty tailored towards just to update it: [here](https://github.com/wasmCloud/wasmCloud/actions/runs/12287280640/job/34289027266#step:3:75)
literally said `it needs migrating` and this is the only difference in the jsons.

If/when it would be deprecated, we can use the recommended and disable the features that are coming with it. But the communication is confusing here.

markkovari:
Seems like renovatebot doing its job, not opening the dashboard just because of the config migration warning. If it will, will reopen this pr/will update it.

PR #3832, opened 2024-12-13
fix: correct regex and add tests

lachieh:
## Feature or Problem
In testing the wash dev changes added in #3820, I found that the process would panic if it encounter *any* log line that didn't have a match.

## Related Issues
#3820

## Release Information
wash@next
 
## Testing
I went ahead and extracted the logic into its own function so that I could add some tests. If there's a better way to do this, I'm all ears.

### DIFF
--- a/crates/wash-cli/src/cmd/dev/session.rs
+++ b/crates/wash-cli/src/cmd/dev/session.rs
@@ -378,38 +378,10 @@ impl WashDevSession {
 
             // Read the log until we get output that
             let _wasmcloud_log_path = wasmcloud_log_path.clone();
-            let host_id = tokio::time::timeout(tokio::time::Duration::from_secs(1), async move {
-                let log_file = tokio::fs::File::open(&_wasmcloud_log_path)
-                    .await
-                    .with_context(|| {
-                        format!(
-                            "failed to open log file @ [{}]",
-                            _wasmcloud_log_path.display()
-                        )
-                    })?;
-                let mut lines = tokio::io::BufReader::new(log_file).lines();
-                let re = regex::Regex::new(if wasmcloud_opts.enable_structured_logging {
-                    r#"\"host_id\":\s?\"([A-Z0-9]+)\""#
-                } else {
-                    r#"host_id="([A-Z0-9]+)"#
-                })
-                .context("failed to compile regex")?;
-                loop {
-                    if let Some(line) = lines
-                        .next_line()
-                        .await
-                        .context("failed to read line from file")?
-                    {
-                        return Ok(re
-                            .captures(&line)
-                            .context("failed to capture host ID")?
-                            .get(1)
-                            .context("failed to get capture group")?
-                            .as_str()
-                            .to_string()) as anyhow::Result<String>;
-                    }
-                }
-            })
+            let host_id = tokio::time::timeout(
+                tokio::time::Duration::from_secs(1),
+                get_host_id(_wasmcloud_log_path),
+            )
             .await
             .context("timeout expired while reading for Host ID in logs")?
             .context("failed to retrieve host ID from logs")?;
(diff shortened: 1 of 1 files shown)


crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -378,38 +378,10 @@ impl WashDevSession {
...
-                        .await
-                        .context("failed to read line from file")?
-                    {
-                        return Ok(re
-                            .captures(&line)
-                            .context("failed to capture host ID")?
-                            .get(1)
-                            .context("failed to get capture group")?
-                            .as_str()
-                            .to_string()) as anyhow::Result<String>;
```
lachieh, line 409: then, this would bomb on the first line without a match

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -431,4 +403,64 @@ impl WashDevSession {
 
         Ok((nats_child, wadm_child, wasmcloud_child))
     }
+
+    async fn get_host_id(log_path: PathBuf) -> anyhow::Result<String> {
```
lachieh, line 407: I'm not sure if this is the best place for this, organizationally. Is it better to put it as a function right in the module? Is this always accessible through `WashDevSession`? I dunno rust things.

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -431,4 +403,64 @@ impl WashDevSession {
...
+        let log_file = tokio::fs::File::open(&log_path)
+            .await
+            .with_context(|| format!("failed to open log file @ [{}]", &log_path.display()))?;
+
+        // looks for the two variations of the log line containing the host_id:
+        //   JSON: "host_id":"ABC123"
+        //   LOG:  host_id="ABC123"
+        let re = regex::Regex::new(r#"(?:\"host_id\":\s?\"|host_id=\")([A-Z0-9]+)\""#)
```
lachieh, line 415: merged the regex so it doesn't depend on the cli options. It just works against std log + structured json.

crates/wash-cli/src/cmd/dev/session.rs
asherillan: If it's not using anything from the struct itself, take it out of the impl.

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -431,4 +403,64 @@ impl WashDevSession {
...
+    #[tokio::test]
+    async fn test_get_host_id() {
+        let log_file = NamedTempFile::new().unwrap();
+        let log_path = log_file.path().to_path_buf();
+        let mut log_file = std::fs::OpenOptions::new()
+            .write(true)
+            .open(&log_path)
+            .unwrap();
+
+        // standard logging pattern
```
asherillan, line 453: Separate these into two test cases.

asherillan (approved):
LG overall, couple things to change and you're good to go

lachieh:
Thanks, @asherillan! New changes pushed.

asherillan (approved):

crates/wash-cli/src/cmd/dev/session.rs
```diff
@@ -432,3 +404,68 @@ impl WashDevSession {
...
+    #[tokio::test]
+    async fn test_get_host_id_from_standard_logging_pattern() {
+        let log_file = NamedTempFile::new().unwrap();
+        let log_path = log_file.path().to_path_buf();
+        let mut log_file = std::fs::OpenOptions::new()
+            .write(true)
+            .open(&log_path)
+            .unwrap();
+
+        writeln!(log_file, "2024-12-13T17:17:07.287574Z  INFO wasmcloud_host::wasmbus: wasmCloud host started host_id=\"ABC123\"").unwrap();
```
vados-cosmonic, line 453: When you're writing from an async context, you should use `tokio::fs` rather than the stuff in `std` since the tokio APIs are built to work with the event loop (a bit like using sync stuff from async JS).

I'm not sure we absolutely *need* async here, but if you do use it, you can probably boil this down to:

```rust
tokio::fs::write(&log_file, "2024-12-.....").await?;
assert_eq!(get_host_id(log_file).await? "ABC123");
```

vados-cosmonic (approved):
Just one little change to make IMO, but approving so when you resolve you can easily merge

crates/wash-cli/src/cmd/dev/session.rs
vados-cosmonic: Also, for a follow up this might be a good thing to test with an integration test (usually those are the ones that attempt to do output to files and stuff) -- in general being able to start `wash dev` with the args that we're expecting should work identically to the [*existing* `wash dev` test](https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-cli/tests/wash_dev.rs#L46)

PR #3835, opened 2024-12-16
fix(examples): docker image deps

vados-cosmonic:

### DIFF
--- a/.github/workflows/examples-components.yml
+++ b/.github/workflows/examples-components.yml
@@ -40,6 +40,8 @@ on:
       - component-http-password-checker-rust-v[0-9]+.[0-9]+.[0-9]+*
       - component-http-password-checker-rust-v[0-9]+.[0-9]+.[0-9]+-*
       - component-http-task-manager-v[0-9]+.[0-9]+.[0-9]+-*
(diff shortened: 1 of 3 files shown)


brooksmtownsend (approved):
I don't want to nit this to death, but just checking if the title of this PR is indicative of the change? Looks like this is a release process for the messaging worker but the title says something about Docker 🤔 Going to approve because the commit messages are descriptive, but may be good to change for posterity

markkovari:
Yeah, probably the trace of this will lead back to the renovate dashboard, where they call it docker instead of oci dependencies as step. Not sure if it is even replaceable or it can be overwritten somehow it literally states `Failed to look up docker package ` 😞

vados-cosmonic:
@brooksmtownsend I guess there wasn't much context in here -- this PR is related the recent [renovate dashboard issue](https://github.com/wasmCloud/wasmCloud/issues/3819) -- the image dependencies were the ones that were found by Renovate -- @markkovari had the context here, but I guess I should have made it clear, I thought the issue link would be enough.

Both of the fixes are to example projects (though what's modified *is* CI for the example project), but point taken!

PR #3855, opened 2024-12-19
chore(Feature)::> Centralize directory handling with ProjectPaths configuration

### LINKED ISSUE #3800: [FEATURE] Rework current dir usage in wash-lib helpers
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

Currently some helper code in `wash-lib` is quite unergonomic/likely to be problematic in the future with it's use of `std::env::current_dir()`. It would be great to refactor this code (and it's usage points) to be more explicit about the intended directory, and *hopefully* to make the code easier to reuse without global side effects/forcing callers to resolve exactly which directory they should be using. 

See [Taylor's comments](https://github.com/wasmCloud/wasmCloud/pull/3779#discussion_r1876923611). It's not clear exactly what the best approach is here, but the goals are to avoid having this "utility" method just decide the project dir at the same time as it does other work.

We don't have to worry about breaking changes too much here, as it *seems* that the main/only user of `wash-lib` is `wash-cli`, and we can evolve them in tandem.

Ideally, we'd have much cleaner code in the `common.rs` file that doesn't try to call out to `std::env::current_dir()`, even if that means just making the determination of project directory more explicit (i.e. moving that out into a separate helper function).

Refactoring `common.rs` into a module with different submodules for helpers might be nice.

_No response_

### Additional context

_No response_

PsychoPunkSage:
## Related Issues
closes #3800 

## Release Information
next

### Manual Verification
1. Created new ProjectPaths configuration and updated affected files
2. Ran existing test suite to ensure no regressions
3. Manually tested the following commands:
   - wash wit build
   - wash wit deps
   - wash dev
4. Verified custom directory paths work correctly
5. Checked that all path resolutions remain consistent with previous behavior

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "advisory-db": {
       "flake": false,
       "locked": {
-        "lastModified": 1733749954,
-        "narHash": "sha256-2Ug80Uf/oUujxgh02Iy5vTG0V+Ab9+YUHuRLRY0ayiY=",
+        "lastModified": 1734961910,
+        "narHash": "sha256-F4iNNs84rdqN2ZDCKtZrE/PUIfUe6YSZM/O2sckeQr4=",
         "owner": "rustsec",
         "repo": "advisory-db",
-        "rev": "ec9ce28714bb38d77a2223e7266df705500a7f11",
+        "rev": "b02b7ca7c98eee7fe26ac18277040c3fc814b52d",
         "type": "github"
       },
       "original": {
@@ -77,11 +77,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1734762720,
-        "narHash": "sha256-uqbENAag0fXOSFcO+EXrYqPqXd0zcVBogzopkAZia8A=",
+        "lastModified": 1735281280,
+        "narHash": "sha256-FgW6NkzbGlmNeY0lywjvXgrZD+UPS/+848kI30Yaup0=",
         "owner": "nix-community",
         "repo": "fenix",
-        "rev": "92e0d5d0d4569e0d4d4da4d83f7c413a735fd642",
+        "rev": "3915fd69e1a2827209f3a1080cb8bcca48d11fa7",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Hey @PsychoPunkSage this looks like a great start!  A few things:

- Rather than enabling use of `current_dir()`, would you mind removing/lessening the use of it? Ideally we should be calling it from one or two places max (when building the project config), and *using* the same path (in project config). 
- We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and I *think* the format might be incorrect on some of these commits so the lint would fail
- DCO is required, there are some handy instructions in [the error message](https://github.com/wasmCloud/wasmCloud/pull/3855/checks?check_run_id=34645306103) on how to sign off on the commits

I'm not sure we need a new struct for this -- I was originally intending the ticket to focus on *removing* uses of `std::env::current_dir()` from functions in helper methods where it didn't belong, more so than creating a new structure to manage paths.

PsychoPunkSage:
Hi @vados-cosmonic 
I have tried to implement the changes as you have pointed out... Now I'm not calling `std::env::current_dir()` everywhere....

Removed direct `current_dir()` call usage and centralized directory handling through `load_config()`
Now using the `code_dir` parameter as the primary source for project directory resolution, falling back to the project config's directory only when necessary.

Please let me know if you want a different impl... (please ignore the extra comments... will be removed before merging)

brooksmtownsend:
Looks like this one might just need a rebase to help drop off the extra changes here, but otherwise I think it looks good

PsychoPunkSage:
Hi @brooksmtownsend 
I tried to rebase the `upstream main` in this PR + cleaned up all the useless lines/files

brooksmtownsend:
Hey @PsychoPunkSage , looks like there are still a bunch of commits here in `main`, do you mind if I modify this PR to drop those extra commits? Feel free to give it a shot as well, I think all you'll need to do is rebase interactively and `drop` the commits that weren't a part of your changes (e.g. https://github.com/wasmCloud/wasmCloud/pull/3855/commits/3664527813faf742b20a98c8a4540b99b33ba408)

PsychoPunkSage:
Hi @brooksmtownsend 
I have removed all unnecessary commits + resolved the conflicts
Please have a look at it.

brooksmtownsend (approved):

brooksmtownsend:
Rebasing due to conflict to keep it easy here

PR #3859, opened 2024-12-20
ref(tests): Removes network calls from unit tests

thomastaylor312:
We've had several tests for a while that were in unit tests doing things like network calls. All network call tests have been moved to integration tests and some cleanup and refactoring was done as well.

I also changed the plugin test to just use a committed wash plugin binary. The code is still in the repo (and updated to use the new wkg fetching), but it added a significant delay because it compiled both `wash` and the plugin. With a committed binary, it becomes much faster


crates/wash-lib/tests/plugins.rs
```diff
@@ -1,52 +1,18 @@
...
 async fn test_subcommand() {
-    let plugin_path = build_plugin("hello_plugin").await;
-
     let mut subcommand = wash_lib::plugin::subcommand::SubcommandRunner::new().unwrap();
+    // NOTE: All the joins are to avoid any problems with cross-platform paths
+    let plugin_path =
+        PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").expect("Unable to find manifest dir"))
+            .join("tests")
+            .join("fixtures")
+            .join("hello_plugin_s.wasm");
```
brooksmtownsend, line 13: tiniest nit request to add instructions on how to rebuild this plugin when it's updated

brooksmtownsend (approved):
_huge_ improvements here. Tiniest nit that would be a totally fine followup, since it's just not really needed

crates/wash-lib/tests/plugins.rs
thomastaylor312: Oh good point. Let me add that in the comment

PR #3872, opened 2024-12-23
chore(CODEOWNERS): simplify org maintainer requirements

brooksmtownsend:
## Feature or Problem
This PR updates the CODEOWNERS file as an initial step to greatly simplify the required reviewers for this repository. As a monorepo, the required reviewers list can often get complicated very quickly. This PR is the first step in a direction to require `org-maintainers` less, and leave reviews to their respective teams wherever possible. For context, every PR that merges in this repository requires a review from the CODEOWNER of the modified files.

The biggest change in this PR is to remove the requirement that a file that doesn't match the CODEOWNERS rules be an `org-maintainer`. We don't have enough content in this repository to require an org-maintainer when we don't have a rule covering a specific file.

In the long term, I'm proposing https://github.com/wasmCloud/org-admin will be used to create teams, repositories, and manage maintainers. CODEOWNERS in this repository will be used to gate PR merging, and we'll be able to bring on more maintainers to help review PRs and contribute features without the org-maintainer requirement 😄 (this all being said, that proposal is separate to this PR and doesn't change what's proposed here)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,19 +1,17 @@
 # CODEOWNERS
 
-## Default to org or maintainer
+## Organization/repository management, top level documentation
 
-* @wasmCloud/org-maintainers @wasmCloud/host-maintainers
-
-## .github CI
-
-.github @wasmCloud/ci-maintainers
 .github/CODEOWNERS @wasmCloud/org-maintainers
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):

thomastaylor312:
Good to go, just note that I don't think the typescript team exists or is visible yet

PR #3874, opened 2024-12-23
feat(host): implement health check API

rvolosatovs:
## Feature or Problem

Implement (opt-in) health checks in the host for better integration with cloud-native tooling

Note, there's no trailing newline in the response, which generally should not matter, but I'm not sure about the assumptions that k8s tooling might make

### Manual Verification

```
cargo run -- --http-admin '[::1]:9090'
```

```
$ curl http://localhost:9090/hello      
curl: (52) Empty reply from server

$ curl -X GET http://localhost:9090/livez 
{"status":"ok"}

$ curl -X POST http://localhost:9090/livez 
curl: (52) Empty reply from server

$ curl -X GET http://localhost:9090/readyz    
{"status":"ok"}
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,5 @@
+use core::net::SocketAddr;
+
 use std::collections::{HashMap, HashSet};
 use std::env;
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -344,6 +346,10 @@ struct Args {
         hide = true
     )]
     help_markdown: bool,
+
+    #[clap(long = "http-admin")]
```
lxfontes, line 350: will this bind the environment variable as well?

lxfontes (approved):

src/main.rs
rvolosatovs: only flag, let me add that quickly

brooksmtownsend (approved):
All LGTM here

PR #3887, opened 2024-12-30
update link validation to allow unique interfaces across the same pac…

LUK3ARK:
…kage

sync changes for wadm https://github.com/wasmCloud/wadm/pull/518


brooksmtownsend (approved):
What's left here is to make sure that the code for looking up targets and configuration takes different interfaces into account. I'm fairly sure it does, but I'd like to test that

brooksmtownsend (approved):

thomastaylor312:
Merging since it looks like things are resolved with the conflict and it was removed from the queue

thomastaylor312:
Oh looks like it isn't a merge conflict but something else. @LUK3ARK can you rebase one more time?

PR #3889, opened 2024-12-30
ref(ci): Improve wash testing times

thomastaylor312:
This PR contains various fixes and refactors for our wash tests all with the goal of improving testing speeds for the wash tests


.github/workflows/wash.yml
```diff
@@ -118,17 +117,35 @@ jobs:
...
+        with:
+          shared-key: 'ubuntu-22.04-shared-cache'
+      - name: Build wash
+        run: cargo build
+        working-directory: ./crates/wash-cli
+      - name: Upload wash binary
+        uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b
+        with:
+          name: wash
+          path: ./target/debug/wash
```
lachieh, line 136: We might want to move this to an action since were doing in the `examples-*` workflows as well. Those ones use [`taiki-e/cache-cargo-install-action`](https://github.com/taiki-e/cache-cargo-install-action) which means less steps to maintain and same outcome:

https://github.com/wasmCloud/wasmCloud/blob/d44138d7937df8e08b0b3c768441ae2f90332a9a/.github/workflows/examples-components.yml#L81-L87

lachieh (commented):
I know this is still only a draft, but I recently changed some of this for other workflows

.github/workflows/wash.yml
thomastaylor312: Oh I didn't realize it could work that way. `install-action` picks up on this properly in the `./.github/actions/wasm-lang-toolchain-setup` then?

lachieh: Yeah, the cache-cargo-install-action will set the cache key against a commit meaning it should only ever build once for a given commit.

I just added #3896 an action which makes this more re-usable.

lachieh: It's merged! I'm going to make another comment

lachieh, line 136:
```suggestion
  build_current:
    name: Build `wash` for e2e tests
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
      - uses: ./.github/actions/install-cached-wash-cli
```

.github/workflows/wash.yml
```diff
@@ -150,13 +167,22 @@ jobs:
...
         with:
-          shared-key: 'ubuntu-22.04-shared-cache'
+          name: wash
+      # Cargo bin should exist and be in the PATH already so we just move it there to make our lives simple
+      - name: Move wash artifact to cargo bin
+        if: ${{ matrix.wash-version == 'current' }}
+        run: |
+          mkdir -p ~/.cargo/bin
+          chmod +x ./wash
+          mv ./wash ~/.cargo/bin/wash
```
lachieh, line 185: can remove this whole section since wash gets installed in `wasm-lang-toolchain-setup`

lachieh (commented):
🤏🏻

thomastaylor312:
Pausing this for a bit to work on something else. Will come back with some more fixes

brooksmtownsend:
Hey @thomastaylor312 I think this one's been sticking around for a bit, do you think we should get this rebased and integrated with what we have?

thomastaylor312:
So I actually pulled some of this in another PR fixing the wash tests so I'm going to close for now. There might be one small part I could add, but this is mostly in main now

PR #3896, opened 2024-12-31
ci: add action for cached-install of wash

lachieh:
This moves wash install into an action that can be reused. If no version is specified, it will do a super basic check on which files have changed to see if it is better to use the commit from main, or the commit from the PR to build `wash` from. This means if the PR is just a change to examples, the latest commit on main has most probably already been built and cached.

This could also be used in E2E tests as per the improvements in #3889

### DIFF
--- a/.github/actions/install-cached-wash-cli/action.yml
+++ b/.github/actions/install-cached-wash-cli/action.yml
@@ -1,0 +1,44 @@
+name: Install Cached Wash CLI
+
+description: |
+  This action will install the Wash CLI from a given repository and revision. If there are no changes that might affect the build, the action will build and install the Wash CLI from the current commit. Otherwise, it will build and install the Wash CLI from the base ref.
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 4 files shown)


.github/actions/install-cached-wash-cli/action.yml
```diff
@@ -0,0 +1,44 @@
...
+          REVISION=${{ github.sha }}
+        else
+          echo "Building from the base ref"
+          REVISION=${{ github.event.pull_request.base.sha }}
+        fi
+        echo "target_revision=$REVISION" >> $GITHUB_OUTPUT
+        echo "Using $REVISION as the target revision"
+
+    - name: Build wash
+      uses: taiki-e/cache-cargo-install-action@caa6f48d18d42462f9c30df89e2b4f71a42b7c2c # v2.0.1
```
brooksmtownsend, line 40: TIL!

brooksmtownsend (approved):

PR #3897, opened 2024-12-31
PR #3855 reopened

### LINKED ISSUE #3800: [FEATURE] Rework current dir usage in wash-lib helpers

brooksmtownsend:
Signed-off-by: Abhinav Prakash <abhinav.prakash319@gmail.com>
- **fix::> Proper exposure of path.rs**
- **fix::> Proper logic of Paths resolution**
- **fix::> replaced all std::env::current_dir() references in wash-cli**
- **fix::> Changed the impl acc to discussion**
- **fix::> publically exposed the config-path**
- **fix::> Cleanupgit add crates/wash-cli/src/cmd/dev/mod.rs crates/wash-cli/src/cmd/mod.rs crates/wash-cli/src/cmd/wit/build.rs crates/wash-cli/src/cmd/wit/deps.rs crates/wash-lib/src/parser/mod.rs**
- **fix::> Cleanupgit add crates/wash-cli/src/cmd/config/mod.rs**
- **chore(wash-cli): remove unused import comment**

## Feature or Problem
It appears I am not a git wizard, and there's something up that closes PR #3855. These are all of the original commits from there by @PsychoPunkSage, who did a great job here, I just had to rebase the branch due to a conflict.

## Related Issues
Originally #3855 
Fixes #3800

### DIFF
--- a/crates/wash-cli/src/cmd/config/path.rs
+++ b/crates/wash-cli/src/cmd/config/path.rs
@@ -1,0 +1,37 @@
+use anyhow::{Context, Result};
+use std::path::{Path, PathBuf};
+
+#[derive(Debug, Clone)]
+pub struct ProjectPaths {
+    project_dir: PathBuf,
+}
+
+impl ProjectPaths {
+    pub fn new(project_dir: impl Into<PathBuf>) -> Self {
+        Self {
+            project_dir: project_dir.into(),
+        }
+    }
+
+    pub fn from_current_dir() -> Result<Self> {
+        Ok(Self::new(
+            std::env::current_dir().context("failed to get current directory")?,
+        ))
+    }
+
+    pub fn project_dir(&self) -> &Path {
+        &self.project_dir
+    }
+
+    pub fn with_project_dir(mut self, dir: impl Into<PathBuf>) -> Self {
+        self.project_dir = dir.into();
+        self
+    }
+}
+
+// Default implementation uses current directory
+impl Default for ProjectPaths {
+    fn default() -> Self {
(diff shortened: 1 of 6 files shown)


brooksmtownsend:
Clippy issue will be fixed after #3898 merges, I will rebase

lachieh:
Looks like you'll have to correct the conventional commit formatting for these commits too

thomastaylor312 (approved):
Good to go pending conventional commit fix

PR #3899, opened 2024-12-31
ci(wasmcloud): ignore wash files for wasmcloud 

brooksmtownsend:
## Feature or Problem
This PR properly ignores wash-cli and wash-lib files when considering if wasmCloud had a change and CI should run.

### Manual Verification
Here's a previous iteration of this PR that changed a file in the wash codebase that did not trigger a run of the wasmCloud workflows. https://github.com/wasmCloud/wasmCloud/actions/runs/12561146525


lachieh:
it'd be super neat to have this in an action so we can use it in other places

PR #3906, opened 2025-01-02
fix(wash)!: use extended interface overrides everywhere

### LINKED ISSUE #3871: [BUG] `wash build` should use expanded overrides like `wash wit fetch`
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

Currently, `wash build` does not use the expanded overrides like `wash wit fetch` does

### Steps to reproduce

1. Add an override to the `wasmcloud.toml`
2. Run `wash wit fetch` to see properly fetching
3. Run `wash build` to see it not fetch

### Expected behavior

`wash build` to fetch

### Environment

- OS: [e.g. Linux, MacOS]
- Shell [e.g. bash, zsh, powershell]
- wasmcloud Version [e.g. 0.81.0]

_No response_

brooksmtownsend:
## Feature or Problem
This PR does some refactoring of our wit fetching and override logic to ensure the same logic is used everywhere for extended project overrides.

## Related Issues
Fixes #3871

### DIFF
--- a/crates/wash-lib/src/deps.rs
+++ b/crates/wash-lib/src/deps.rs
@@ -1,0 +1,408 @@
+//! Utilities for working with and managing wit dependencies
+
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+
+use anyhow::{bail, Context, Result};
+use async_compression::tokio::bufread::GzipDecoder;
+use futures::TryStreamExt as _;
+use semver::VersionReq;
+use url::Url;
+use walkdir::WalkDir;
+use wasm_pkg_client::{
+    caching::{CachingClient, FileCache},
+    PackageRef,
+};
+use wasm_pkg_core::{config::Override, lock::LockFile, wit::OutputType};
+use wasmcloud_core::parse_wit_package_name;
+
+use crate::{
+    cli::CommonPackageArgs,
+    common::{clone_git_repo, RepoRef},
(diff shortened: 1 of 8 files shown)


crates/wash-lib/src/deps.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    /// Use configuration to create a caching client for fetching wkg dependencies
+    pub fn into_client(self) -> CachingClient<FileCache> {
+        let client = wasm_pkg_client::Client::new(self.wkg_client_config);
+        CachingClient::new(Some(client), self.cache)
+    }
+
+    /// Enable extended pull configurations for wkg config. Call before calling `Self::get_client` to
+    /// update configuration used.
+    pub async fn resolve_extended_pull_configs(
```
thomastaylor312, line 77: Nit: Would it be better to use this as a builder or something similar rather than a function you have to remember to call?

Honestly this probably isn't worth doing now, but we might want to refactor this so it is clearer that you're essentially layering in extra config before calling `into_client`

thomastaylor312 (approved):
The comment here is just mostly a note for our future selves to maybe make the fetcher stuff cleaner. Definitely fine for merging in now since this is a fix

crates/wash-lib/src/deps.rs
thomastaylor312: Resolving so this can merge

PR #3910, opened 2025-01-02
fix(host): output host_id on provider_start_failed

brooksmtownsend:
## Feature or Problem
This PR adds the missing `host_id` field on the provider_start_failed event.

Generally we don't actually need this, because the `source` of the event is the host ID. Since we put this field on many other event payloads though, let's just get to parity and then make a longer term decision to keep or remove it from events.


thomastaylor312:
For reference to those who see this PR, we do want to remove the host_id field eventually. See #3914

thomastaylor312 (approved):

PR #3932, opened 2025-01-06
feat: allow disabling auction participation

rvolosatovs:
## Feature or Problem

For some deployments it is beneficial not to manage binary providers

### DIFF
--- a/crates/host/src/wasmbus/host_config.rs
+++ b/crates/host/src/wasmbus/host_config.rs
@@ -80,6 +80,10 @@ pub struct Host {
     pub experimental_features: Features,
     /// HTTP administration endpoint address
     pub http_admin: Option<SocketAddr>,
+    /// Whether component auctions are enabled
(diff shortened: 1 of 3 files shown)


rvolosatovs:
will need to add support for auction as well tomorrow

lxfontes (approved):
thank you!

vados-cosmonic (approved):
LGTM :rocket: 

Might have been nice to have disable instead of enable with default (avoid someone using bad defaults w/ `WasmbusHostConfig`), but I'm assuming that was written that way to avoid `if !provider_auction`...

rvolosatovs:
> LGTM :rocket: 
> 
> Might have been nice to have disable instead of enable with default (avoid someone using bad defaults w/ `WasmbusHostConfig`), but I'm assuming that was written that way to avoid `if !provider_auction`...

I made it this way for consistency with existing *_ENABLED flags

PR #3934, opened 2025-01-07
chore: upgrade opentelemetry libraries to 0.27

joonas:
## Feature or Problem

Quite a bit has changed since `0.23`, most notably the breaking changes introduced in [`0.27`](https://github.com/open-telemetry/opentelemetry-rust/blob/a1dda220deed9f69bd8502e8ffc0a45d72ac836e/opentelemetry-otlp/CHANGELOG.md#0270).

I tried to keep the changes here as minimal and focused on the upgrade without changing much else

### DIFF
--- a/crates/host/Cargo.toml
+++ b/crates/host/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-host"
-version = "0.23.1"
+version = "0.23.2"
 description = "wasmCloud host library"
 readme = "README.md"
 
--- a/crates/tracing/src/metrics.rs
+++ b/crates/tracing/src/metrics.rs
@@ -8,88 +8,42 @@ pub fn configure_metrics(
     service_name: &str,
     otel_config: &wasmcloud_core::OtelConfig,
 ) -> anyhow::Result<()> {
-    use opentelemetry_otlp::{MetricsExporterBuilder, WithExportConfig};
+    use opentelemetry_otlp::{WithExportConfig, WithHttpConfig};
+    use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
     use wasmcloud_core::OtelProtocol;
 
-    let builder: MetricsExporterBuilder = match otel_config.protocol {
+    let exporter = match otel_config.protocol {
         OtelProtocol::Http => {
             let client = crate::get_http_client(otel_config)
                 .context("failed to get an http client for otel metrics exporter")?;
-            opentelemetry_otlp::new_exporter()
-                .http()
-                .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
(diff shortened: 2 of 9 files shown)


protochron (approved):

crates/host/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmcloud-host"
-version = "0.23.1"
+version = "0.24.0"
```
thomastaylor312, line 3: Nit: Couldn't we technically get away with a patch bump? Changes in dep crates are always interesting. Nothing changes in the API of the crate, so I think it could be fine.

Either way, I don't care what we do, was just calling it out

crates/tracing/src/metrics.rs
```diff
@@ -8,88 +8,43 @@ pub fn configure_metrics(
...
-/// Replaces the default `ExplicitBucketHistogram` aggregation for Histograms
-/// with `Base2ExponentialHistogram`. This makes it easier to capture latency
-/// at nanosecond accuracy.
-#[cfg(feature = "otel")]
-#[derive(Clone, Default, Debug)]
-struct ExponentialHistogramAggregationSelector {
-    pub(crate) _private: (),
-}
+        .with_reader(reader)
+        // .with_aggregation_selector(ExponentialHistogramAggregationSelector::new())
```
thomastaylor312, line 44: Nit: Left over debug/refactor?

thomastaylor312: Also, just double checking, we don't need any aggregation here, right?

thomastaylor312 (approved):
Just some minor nits/questions that can be resolved if wanting to ignore/when they are answered

crates/host/Cargo.toml
joonas: We certainly can if that's preferred, I don't have a strong preference in this case, a minor bump felt appropriate given the changes, but we can go with patch.

crates/tracing/src/metrics.rs
joonas: With [open-telemetry/opentelemetry-rust#2085](https://github.com/open-telemetry/opentelemetry-rust/pull/2085) this option was removed, so the aggregation will be based on https://opentelemetry.io/docs/specs/otel/metrics/sdk/#default-aggregation.

That said, I believe a future improvement that could be made here is to set custom values for the explicit histogram boundaries rather than relying on the defaults from the default aggregation.

It's not clear to me that there's an easy way to get the previous functionality back with the changes that were made upstream, there are features that were mentioned as being applicable for this ("views"), but even their future seems slightly uncertain, hence our best for the moment being getting setting explicit boundaries.

crates/host/Cargo.toml
joonas: I've switched it to `0.23.2`

Issue #2936, opened 2024-09-04
[BUG] wasmCloud fails to stop components loaded from files if the file is deleted


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

```
➜ wash stop component hello_world-http_component

Failed to scale component [hello_world-http_component] on host [NBFSZUYUCNPGLXAY6Y5MONI3XCBTVFI3IR6BJDASXGBM32CV2OA4FVPY]

Caused by:
    failed to fetch component: failed to read component: No such file or directory (os error 2)
```

### Steps to reproduce

1. Run an example hello-world component, e.g. the Rust hello world example in this repo from the local wadm
2. Remove the wasm component from disk `rm build/*.wasm`
3. Try to stop the component

### Expected behavior

Even if the component doesn't exist on disk, we should be able to stop it based on the component ID

### Environment

- OS: [e.g. Linux, MacOS]
- Shell [e.g. bash, zsh, powershell]
- wasmcloud Version [e.g. 0.81.0] 1.2.0

flyaruu:
Oh, I was taking a stab at this one, can someone un-stale it and assign to me?

flyaruu:
I think I have a working version, it will scale down even with the file deleted, but wadm will immediately try to restart it, and fail because the wasm file isn't there, so the component ends up in 'failed'

Is this what you had in mind, @brooksmtownsend ?

flyaruu:
I've rebased and also fixed a commit message issue, but the actions doesn't seem to be firing, the CI seems to require an approval of a maintainer, e.g.:

https://github.com/wasmCloud/wasmCloud/actions/runs/12705939702

Is this intentional?

flyaruu:
Hey @vados-cosmonic, thanks, I've iterated a few times, and merging to my own main now yields no problems, can you re-approve the CI?

thomastaylor312:
I think this was closed in #3940

PR #3940, opened 2025-01-09
bug(host): components can not be deleted when wasm file can not be fo…

flyaruu:
## Feature or Problem

TLDR, the host would always resolve the wasm source, even for deleting, creating the situation that stopping a components gets really hard when the WASM file was deleted.

It was trickier than I thought.

Please review carefully, it's my first wasmCloud PR

## Related Issues
https://github.com/wasmCloud/wasmCloud/issues/2936

## Consumer Impact
Slight improvement for dev/operation ergonomics

### Manual Verification
Tested the scenario in the issue

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1765,7 +1765,7 @@ impl Host {
                     &annotations,
                     config,
                     wasm.ok_or_else(|| {
-                        retrieval_error.unwrap_or(anyhow!("unexpected retrieval error"))
+                        retrieval_error.unwrap_or_else(|| anyhow!("unexpected retrieval error"))
                     }),
                     claims_token.as_ref(),
                 )
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1765,7 +1765,7 @@ impl Host {
                     &annotations,
                     config,
                     wasm.ok_or_else(|| {
-                        retrieval_error.unwrap_or_else(|| anyhow!("unexpected retrieval error"))
+                        retrieval_error.unwrap_or_else(|| anyhow!("unexpected missing wasm binary"))
                     }),
                     claims_token.as_ref(),
                 )
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -1426,7 +1426,7 @@ impl Host {
     async fn start_component<'a>(
         &self,
         entry: hash_map::VacantEntry<'a, String, Arc<Component>>,
-        wasm: Vec<u8>,
+        wasm: &Vec<u8>,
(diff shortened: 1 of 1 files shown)


lachieh:
Hey @flyaruu, thanks for this!

We had some issues with the CI pipeline today but they've been resolved now. Could you fetch and rebase against main and push again? That should solve the check errors.

lachieh:
@flyaruu I've triggered the workflow, are you able to rebase again with the signed-off commit flag? Sorry I didn't catch that one earlier.

```
git rebase HEAD~4 --signoff
```

flyaruu:
Did not think about the signoffs again, surprised that the CI didn't catch it

lachieh:
Looks great, thanks! There is a DCO check that failed, but checks don't return notifications so I figured I'd ping to keep it moving.

The workflows are running now. One of the @wasmCloud/host-maintainers should be along shortly!

crates/host/src/wasmbus/mod.rs
```diff
@@ -1740,6 +1740,7 @@ impl Host {
                     }
                     return;
                 }
+                Err(e) => (None, None, Some(e)),
```
vados-cosmonic, line 1743: Is it me or does this indentation look a bit weird?

Also, if `compnent_and_claims` is `Err(_)` I think we're no longer publishing the `component_scale_failed` event?

crates/host/src/wasmbus/mod.rs
```diff
@@ -1750,7 +1751,9 @@ impl Host {
                     max_instances,
                     &annotations,
                     config,
-                    wasm,
+                    wasm.ok_or_else(|| {
+                        retrieval_error.unwrap_or(anyhow!("unexpected retrieval error"))
```
vados-cosmonic, line 1755:
```suggestion
                        retrieval_error.unwrap_or_else(|| anyhow!("unexpected retrieval error"))
```

What does it mean if there is no retrieval error, but the wasm is missing? Unless I'm misunderstanding I think the message here isn't quite right

vados-cosmonic (commented):
This looks like a great start @flyaruu -- I left some feedback, and one thing that would be great to see to really round this out is a of a file that gets moved/becomes unavailable.

crates/host/src/wasmbus/mod.rs
flyaruu: I don't think that path is actually possible right now, maybe something like 'unexpected missing wasm'?

crates/host/src/wasmbus/mod.rs
flyaruu: I think indent is ok? The three variants line up at least, I usually defer to cargo fmt.
We're not publishing the component_scale_failed event _yet_ because if it is a scale down to zero event, there isn't a problem, we don't need the wasm to do that. 

If we do need it, we will publish the component_scale_failed from the handle_scale_component_task function

PR #3948, opened 2025-01-09
ci!: use wolfi as a default image base

rvolosatovs:
## Feature or Problem

Use `wolfi` as the default image base as requested by @joonas 

## Consumer Impact

This is a breaking change for users of `latest`, `canary` and version tags depending on Debian being the base image.
E.g.:
- Providers dynamically linked to glibc *will fail to startup*
- Downstream users using images via e.g. `FROM` and using `apt` etc. to add packages will break


joonas (approved):
Looks great, thanks! 👍

PR #3951, opened 2025-01-10
Remove stalebot in favor of @actions/stale

lachieh:
Stalebot is apparently [deprecated](https://github.com/probot/stale) so this is swapping over to the action


joonas (approved):
TIL [`actions/stale`](https://github.com/actions/stale) is a thing.

brooksmtownsend (commented):
Nice! @lachieh do I need to disable the stalebot at the org level?

lachieh:
I think there's a few other repos that still use the bot config, but I'll poke around and get back to you.

PR #3962, opened 2025-01-13
feat(chart): Adds one click benchmarking chart

thomastaylor312:
Adds a new chart for running simple benchmarking tests against wasmCloud hosts. I'll follow this up with a publish pipeline for the chart

### DIFF
--- a/charts/benchmark/README.md
+++ b/charts/benchmark/README.md
@@ -1,0 +1,137 @@
+# wasmCloud Benchmarking Chart
+
+This chart is primarily used for benchmarking wasmCloud deployments within a Kubernetes cluster. It
+can be used to test something outside of the cluster as well, but this is not a tested use case.
+This chart has a self contained stack of Grafana, Loki, Tempo, OTEL collectors, and Prometheus as
+well as installation for the [K6 Operator](https://github.com/grafana/k6-operator) for running the
+actual benchmark. This setup is meant to be temporary as many users are likely to have their own
+Grafana or OTEL stacks. This setup is purely for capturing data for the test. By default, this is a
+"push button" chart that will install the stack and run the benchmark for you. It can then be
+uninstalled so as to not take up resources.
+
+## Basic Usage
+
+A simple install command looks like below:
+
+```bash
+helm upgrade --install my-benchmark --version 0.1.0 oci://ghcr.io/wasmcloud/charts/benchmark --wait --set test.url=http://hello-world:8000
+```
+
+Please note that the `test.url` value is required. This is the URL that the benchmark will use to
+test against. This should be a URL that is accessible from within the Kubernetes cluster. The
+`--wait` flag should also be used so the k6 CRD and controller can start up before running the
+benchmark.
+
+The output from the helm install has instructions for how to view results and the dashboard. For the
+install command above, the output would look like this:
+
+```
+The k6 benchmark should now be running! To get the logs and output of the test, you can run:
+
+kubectl logs -n default -l k6_cr=my-benchmark-test,runner=true --tail=-1
+
+If you'd like to view dashboards during or after your tests, port-forward to the Grafana instance:
+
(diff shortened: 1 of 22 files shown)


flake.nix
```diff
@@ -471,6 +471,8 @@
               pkgs.vault
               pkgs.wit-deps
               pkgs.go
+              pkgs.kubernetes-helm
```
thomastaylor312, line 474: Noticed these were missing and I use the devShell. Added them for anyone else who might be doing dev work in the same way

charts/benchmark/README.md
```diff
@@ -0,0 +1,137 @@
...
+Grafana or OTEL stacks. This setup is purely for capturing data for the test. By default, this is a
+"push button" chart that will install the stack and run the benchmark for you. It can then be
+uninstalled so as to not take up resources.
+
+## Basic Usage
+
+A simple install command looks like below:
+
+```bash
+helm upgrade --install my-benchmark --version 0.1.0 oci://ghcr.io/wasmcloud/charts/benchmark --wait --set test.url=http://hello-world:8000
```
brooksmtownsend, line 17: Do we need a release pipeline for this? Suggest changing this to just be the local path for now if we want to do it in a follow up 😄

brooksmtownsend: Full local instructions (what I went through):

```bash
cd charts/benchmark
helm repo add https://grafana.github.io/helm-charts
helm repo add grafana https://grafana.github.io/helm-charts
helm repo add opentelemetry https://open-telemetry.github.io/opentelemetry-helm-charts
helm dependency build
helm upgrade --install my-benchmark --version 0.1.0 ./ --wait --set test.url=http://hello-world:8000
```

charts/benchmark/README.md
```diff
@@ -0,0 +1,137 @@
...
+This chart is primarily used for benchmarking wasmCloud deployments within a Kubernetes cluster. It
+can be used to test something outside of the cluster as well, but this is not a tested use case.
+This chart has a self contained stack of Grafana, Loki, Tempo, OTEL collectors, and Prometheus as
+well as installation for the [K6 Operator](https://github.com/grafana/k6-operator) for running the
+actual benchmark. This setup is meant to be temporary as many users are likely to have their own
+Grafana or OTEL stacks. This setup is purely for capturing data for the test. By default, this is a
+"push button" chart that will install the stack and run the benchmark for you. It can then be
+uninstalled so as to not take up resources.
+
+## Basic Usage
```
brooksmtownsend, line 12: I think a short tidbit about how to deploy the most basic wasmCloud host in a way that's accessible to this test would be useful here. I actually expected this chart to include the wasmCloud deployment too, e.g. the individual `helm upgrade --install` to deploy wasmCloud and the benchmark test.

For dummies like me that just wanna try it out, would be great to have the oneliner/twoliner to deploy the wasmCloud chart too. Thoughts?

brooksmtownsend (commented):
Deployment of the chart went great! Just two questions/thoughts around the instructions folks will see

charts/benchmark/README.md
thomastaylor312: All you should need for local is `helm dep up`. I forgot to add directions to README for that though.

As for release pipeline, see #3974

charts/benchmark/README.md
thomastaylor312: So my logic here was benchmarking a system you don't even have installed doesn't have a lot of of value IMO. I do think it would be useful though to add something to the README pointing to the different ways to install

charts/benchmark/README.md
thomastaylor312: Added instructions to the README. Feel free to resolve this if they look good

charts/benchmark/README.md
thomastaylor312: Added some more stuff to the README about this. Resolve if it looks good!

brooksmtownsend (approved):

PR #4003, opened 2025-01-23
feat(ci): Adds pipeline for custom k6 test runner

thomastaylor312:
In order to collect the test results from k6 into a configmap (which can then be collected and published wherever we want), we need to build k6 with custom extensions. Even if we choose to publish these in other ways in the future, we'll still need a custom runner like this. This is just a simple pipeline and dockerfile for building said custom runner


ci/Dockerfile.k6runner
thomastaylor312, line 1: Please note that this directory will also contain other scripts for the benchmark test when I add those

brooksmtownsend (approved):

PR #4009, opened 2025-01-24
feat(charts): Adds result aggregation to the benchmark chart

thomastaylor312:
This PR focuses on updating the test to publish all results to configmaps so they can be pulled out and aggregated as desired post test. This is a first step for setting up regular benchmarking tests for wasmCloud. It also includes the script we can run as part of our CI tests when we add them

### DIFF
--- a/ci/aggregate_data.py
+++ b/ci/aggregate_data.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python3
+import sys
+import json
+
+def create_metric_aggregator(m_type, m_contains):
+    """
+    Create a structure to hold partial sums/counters and eventually produce final merged values for
+    one metric. We need to do this because some of the metrics need to be averaged at the end of
+    aggregating all the data. Doing this with p90 values and what not isn't technically accurate,
+    but we should only average once at the end. Later down the line we might just want to integrate
+    aggregating the raw data
+    """
+    return {
+        "type": m_type,
+        "contains": m_contains,
+        # We'll accumulate data in these fields:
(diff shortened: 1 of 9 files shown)


ci/aggregate_data.py
```diff
@@ -0,0 +1,222 @@
...
+        "metrics": merged_metrics
+    }
+    return merged
+
+def main():
+    # Read a single JSON array from stdin
+    raw = sys.stdin.read().strip()
+    if not raw:
+        print("{}")
+        return
```
brooksmtownsend, line 211: should we say that we didn't get valid info here instead of just {}?

ci/aggregate_data.py
```diff
@@ -0,0 +1,222 @@
...
+    docs = json.loads(raw)
+    if not isinstance(docs, list):
+        print(json.dumps({"error": "Input JSON is not an array"}, indent=2))
+        return
+
+    merged = combine_all_docs(docs)
+    print(json.dumps(merged, indent=2))
+
+if __name__ == "__main__":
+    main()
```
brooksmtownsend, line 222: oh python 😆

brooksmtownsend (approved):
What made you want to write the script in Python v.s. a Go / bash / Rust thing? (Genuinely curious, I don't think it really matters since this is just being run in CI)

LGTM, just one nit and otherwise looking forward to seeing the script in use

ci/aggregate_data.py
thomastaylor312: Probably. I'll change that

PR #4013, opened 2025-01-26
feat: Embedding WIT into Provider Archives

LUK3ARK:
## Feature or Problem
Capability providers are treated as components by higher levels - they provide some functionality and obey an interface. However, this isn't maintained inside the resulting provider binaries or provider archives (.par.gz). This breaks down how capabilities can be treated similarly to components. Embedding WIT into provider artifacts will make it self-describable and will remove the need for constant internal support across runtimes and other implementations.

## Consumer Impact
This change aims to maintain backward compatibility for older provider archives and for whatever reason a wit interface cannot be detected during build and not provided, should still allow the provider to be buildable, distributable and executable... just without the nice interface wrapper.

This is implemented to use the existing configuration structure for wash. First, It will use the supplied directory where the wit interface and dependencies can be found. This is either provided through the wit key in the wasmcloud.toml manifest or just defaults to the current project directory (the most common situation).

When creating a par, it is also possible to provide explicitly a wit world and a wit directory using --wit-world and --wit-directory respectively. They have to both be provided or not provided at all. When they are not provided, again, no wit interface is added to the PAR.

I refrained from adding it to the insert command, because there should be no differences in interface implementations between binaries. Not adding the capacity to mutate or change the wit interface upon an insert ensures this is more likely to be synced from the original build tooling and stops human error outside of the build process. 

This means the only responsibility on the user is to make sure that the binary they create or insert into a par is using the same wit interface as when compiled.

## Testing
Added wash lib testing for new wit dir and wit world flags

### Unit Test(s)
todo

### Acceptance or Integration
todo

### Manual Verification
1. built providers with new wash build - specifically sql postgres
2. inspected using command `wash inspect --wit ./build/sqldb-postgres-provider.par.gz`

image


crates/provider-archive/src/archive.rs
```diff
@@ -372,16 +403,35 @@ fn validate_hashes(
...
+    libraries: &HashMap<String, Vec<u8>>,
+    wit: &Option<Vec<u8>>,
+) -> HashMap<String, String> {
     let mut hm = HashMap::new();
     for (target, lib) in libraries {
         let hash = hash_bytes(lib);
         hm.insert(target.to_string(), hash);
     }
 
+    if let Some(interface) = wit {
```
LUK3ARK, line 430: A hash is only generated if wit is found, otherwise carries on with old logic

crates/provider-archive/src/archive.rs
```diff
@@ -372,16 +403,35 @@ fn validate_hashes(
             return Err(format!("File hash and verify hash do not match for '{tgt}'").into());
         }
     }
+
+    if let Some(interface) = wit {
```
LUK3ARK, line 407: wit is only included in validation if found, should not impact PAR that do not have WIT embedded

crates/provider-archive/src/archive.rs
thomastaylor312, line 1: Can you add a test that writes takes an old provider archive and makes sure it can still be read? Likewise, can you add a test that does a new round trip archive creation? So write out a parcheezy with the wit included and then read it back through?

crates/wash-lib/src/common.rs
```diff
@@ -411,3 +413,56 @@ pub async fn clone_git_repo(
...
+/// Creaetes a dummy wasm module which encodes the WIT resolve and world for the provider
+/// Wasm can contain either a set of WIT package(s) or a component. We
+/// need to treat providers like a component however it is not a wasm module,
+/// therefore we need to make a blank component implementation of the provider
+/// which encodes the world.
+pub fn create_dummy_provider_wasm(
+    wit_dir: impl AsRef<Path>,
+    world_name: Option<&str>,
+) -> Result<Option<Vec<u8>>> {
+    let (resolve, world_id) = convert_wit_dir_to_world(&wit_dir, world_name)
```
thomastaylor312, line 426: Not sure I follow why we're using the `dummy_module` here over `wit_component::encode` like we do when we're building a wit package. That does encode it as a component but does require a package ID, which I think we can just fetch from the top level resolve

thomastaylor312 (changes requested):
This is really great! Thanks for doing the work here! A few suggestions and questions. I'd also like to see a new wash test for the new inspect functionality.

crates/wash-lib/src/common.rs
LUK3ARK: I done it like this because I thought that package ID is none if no package is defined, I wanted to cover the case for when no package header is present.

If a default package header is always forced to be present I can switch this logic over no problem!

thomastaylor312: To define your world, I'm pretty sure a package name is required in at least one of the wit files, so I think you should always have one when parsing from the directory

LUK3ARK: Replaced using encode, works even better!
image

thomastaylor312:
@LUK3ARK can you rebase on main real quick just to make sure we have all the latest dep updates. After that we can address if there are other test failures. Something is wrong with dep resolution so a rebase is likely to fix that

crates/wash-cli/src/par.rs
```diff
@@ -276,8 +284,25 @@ pub async fn handle_create(cmd: CreateCommand, output_kind: OutputKind) -> Resul
         ),
     };
 
+    let wit_interface_bytes = match (cmd.wit_world.as_deref(), cmd.wit_dir.as_ref()) {
+        (Some(_), None) => bail!("Both --wit-world and --wit-directory must be provided together"),
+        (None, Some(_)) => bail!("Both --wit-world and --wit-directory must be provided together"),
+        (Some(world), Some(dir)) => {
```
thomastaylor312, line 290: Is this `world` var even used anymore? We might not need the flag for it since I don't see it here

LUK3ARK: I can remove this but what would happen if there is a package with more than one world now? Will it encode the package and contain two or more worlds?

LUK3ARK: Think that should be good

thomastaylor312 (approved):

LUK3ARK:
Alright well I guess I have to upgrade wit-component and such anyway, stream type encoding isn't support in 0.219, stream<T> just got encoding support added. Will update to 0.224.

http-client and http-server also don't have wit files in their directories, is there any objection to me just adding them?

crates/provider-http-client/wit/interfaces.wit
thomastaylor312, line 1: So these http providers will need to be the exception for now and don't need the wit. Their wit is included as part of the `wrpc_interface_http` crate and is already included there. We can figure out down the line how we want to handle special cases like this

thomastaylor312 (changes requested):

thomastaylor312 (approved):

PR #4014, opened 2025-01-26
fix(nats): correct tracing field name in link deletion

LUK3ARK:


vados-cosmonic (approved):
LGTM :rocket: 

Thanks @LUK3ARK !

PR #4017, opened 2025-01-27
Update deps

LUK3ARK:


LUK3ARK:
Just want to see if this also cause CI issues, if not I know where I have to fix in other PRs :)

brooksmtownsend:
I think the other wit-bindgen-core/go/etc libraries need to update here too

brooksmtownsend:
@LUK3ARK you may want to see #4020 since there's some upstream deps that are causing issues here

PR #807, opened 2023-10-26
chore: integrate `wash` into the repository

rvolosatovs:
## Feature or Problem

Integrate `wash` into the workspace and CI pipelines.

`cargo build -p wash-cli` will now produce a `wash` binary in workspace `target` (also, `cargo build --workspace`, for example)
That means that `nix build '.#wasmcloud-${target}'` now also produce both binaries reproducibly-built

I had to switch to native Windows builds due to https://github.com/rust-lang/rust/issues/92212, which is now hit by `wash`, but not `wasmcloud`. We can revisit this if/when either:
- `wash` dependency set becomes smaller
- issue is fixed upstream
- using MSVC on Linux is feasible 

Things done:

- Addressed `cargo audit` issues
- Cleaned up the `wash-cli` code originally such that we could simply have a second binary in root-level `src/bin`, but I have since switched the approach and moved the binary back to `crates/wash-cli`. I've left the improvements in.
- `washboard` *MUST* be a subdirectory of wherever `wash-cli` `Cargo.toml` is located, now that's `crates/wash-cli`.

Some things may still be broken - there's a lot of hidden details in configs etc. (like the wash snapcraft config), but I tried my best to address everything. It's best to address anything left in follow-ups to minimize diff and get this one over the line, archive `wash` repo

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

PR #4018, opened 2025-01-27
ci: split binary builds into separate flake outputs

rvolosatovs:
Build each binary in the project individually, as opposed to always building all binaries in the workspace.

- `clippy` still runs on the whole workspace
- wasmCloud `nextest` still builds all provider binaries (this is the lowest-hanging fruit for future optimization and requires no changes to Nix flake/github actions)
- each provider-target pair gets its own CI job
- each wash-target pair gets its own CI job

### DIFF
--- a/.github/workflows/provider.yml
+++ b/.github/workflows/provider.yml
@@ -55,13 +55,13 @@ jobs:
         with:
           path: artifacts
 
-      - run: chmod +x "./artifacts/wasmcloud-aarch64-apple-darwin/bin/${{ inputs.name }}-provider"
-      - run: chmod +x "./artifacts/wasmcloud-aarch64-unknown-linux-musl/bin/${{ inputs.name }}-provider"
-      - run: chmod +x "./artifacts/wasmcloud-x86_64-apple-darwin/bin/${{ inputs.name }}-provider"
-      - run: chmod +x "./artifacts/wasmcloud-x86_64-pc-windows-msvc/bin/${{ inputs.name }}-provider.exe"
-      - run: chmod +x "./artifacts/wasmcloud-x86_64-unknown-linux-musl/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-aarch64-apple-darwin/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-aarch64-unknown-linux-musl/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-x86_64-apple-darwin/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-x86_64-pc-windows-msvc/bin/${{ inputs.name }}-provider.exe"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-x86_64-unknown-linux-musl/bin/${{ inputs.name }}-provider"
 
-      - run: mv "./artifacts/wasmcloud-x86_64-unknown-linux-musl/bin/wash" wash
+      - run: mv "./artifacts/wash-x86_64-unknown-linux-musl/bin/wash" wash
       - run: chmod +x wash
 
       - run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/provider.yml
```diff
@@ -55,13 +55,13 @@ jobs:
...
 
-      - run: chmod +x "./artifacts/wasmcloud-aarch64-apple-darwin/bin/${{ inputs.name }}-provider"
-      - run: chmod +x "./artifacts/wasmcloud-aarch64-unknown-linux-musl/bin/${{ inputs.name }}-provider"
-      - run: chmod +x "./artifacts/wasmcloud-x86_64-apple-darwin/bin/${{ inputs.name }}-provider"
-      - run: chmod +x "./artifacts/wasmcloud-x86_64-pc-windows-msvc/bin/${{ inputs.name }}-provider.exe"
-      - run: chmod +x "./artifacts/wasmcloud-x86_64-unknown-linux-musl/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-aarch64-apple-darwin/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-aarch64-unknown-linux-musl/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-x86_64-apple-darwin/bin/${{ inputs.name }}-provider"
+      - run: chmod +x "./artifacts/${{ inputs.name }}-provider-x86_64-pc-windows-gnu/bin/${{ inputs.name }}-provider.exe"
```
brooksmtownsend, line 61: I will need to double check here, does this have any negative impact for Windows users? E.g. does the `-gnu` target run just fine on a regular Windows 11 machine? It's been quite a while since figuring out the difference between these targets practically.

.github/workflows/wasmcloud.yml
```diff
@@ -216,9 +283,149 @@ jobs:
...
+          - name: blobstore-azure
+            target: aarch64-apple-darwin
+          - name: blobstore-azure
+            target: aarch64-unknown-linux-musl
+          - name: blobstore-azure
+            target: x86_64-apple-darwin
+          - name: blobstore-azure
+            target: x86_64-pc-windows-gnu
+          - name: blobstore-azure
+            target: x86_64-unknown-linux-musl
```
brooksmtownsend, line 301: Could this be simplified to a list of names and targets, e.g.
```yaml
config:
  name: [blobstore-azure, blobstore-fs, ...]
  target: [aarch64-apple-darwin, aarch64-unknown-linux-musl, ...]
```

I know that this wouldn't let us be flexible with, say, ignoring building just one provider for a specific target, but all of them look the same here to me

.github/workflows/wasmcloud.yml
```diff
@@ -216,9 +283,149 @@ jobs:
...
+        with:
+          cachixAuthToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+      - uses: ./.github/actions/build-nix
+        with:
+          package: ${{ matrix.config.name }}-provider-${{ matrix.config.target }}
+
+  build-wash-windows:
     if: ${{ startswith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/main'}}
-    name: wasmcloud-x86_64-pc-windows-msvc
+    name: wash-x86_64-pc-windows-msvc
```
brooksmtownsend, line 428: Should we have a followup to build wash for `gnu` instead of `msvc` like we're doing for wasmCloud?

brooksmtownsend (approved):
Love the changes here to actions and nix, maybe @thomastaylor312 could take another look at the flake since he's got some experience there. I only had questions about the move from msvc -> gnu and some matrix simplification

.github/workflows/provider.yml
thomastaylor312: As far as I understand it, this mostly matters if you're integrating in with other windows tools/libraries. Since this is just a self-contained binary, I think this might be the better choice. Also, I don't know if you can cross-compile msvc on linux, but definitely can with gnu

.github/workflows/wasmcloud.yml
```diff
@@ -135,65 +135,132 @@ jobs:
...
+    if: ${{ needs.meta.outputs.wasmcloud_modified == 'true' || startsWith(github.ref, 'refs/tags/v') }}
+    strategy:
+      matrix:
+        config:
+          - target: aarch64-unknown-linux-musl
+            test-bin: |
+              nix profile install --inputs-from . 'nixpkgs#qemu'
+              qemu-aarch64 ./result/bin/wasmcloud --version
+            test-oci: docker load < ./result
+            # TODO: Run aarch64 binary within OCI
```
thomastaylor312, line 230: Capture this and the other TODO down below as an issue?

flake.nix
```diff
@@ -470,31 +557,31 @@
...
+              wasmcloud-oci-wolfi
               wasmcloud-riscv64gc-unknown-linux-gnu-fhs
               wasmcloud-x86_64-unknown-linux-gnu-fhs
               wasmcloud-x86_64-unknown-linux-musl-oci-debian
               wasmcloud-x86_64-unknown-linux-musl-oci-wolfi
               ;
 
+            default = attrs.wasmcloud.packages.wasmcloud;
             rust = hostRustToolchain;
-            wash = pkgs.runCommandLocal "wash" {} ''
```
thomastaylor312, line 494: Can we not keep around a `wash` pkg alias still for using this in downstream flakes? Looks like we already have it available in `attrs` now anyway. Not a super big thing, but I had found it helpful when pulling into other flakes

rvolosatovs: We have a `wash` output, which only builds `wash` now

.github/workflows/provider.yml
```diff
@@ -178,10 +178,10 @@ jobs:
       # Set up wash
       - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16
         with:
-          path: artifacts
-      - run: mv "./artifacts/wasmcloud-x86_64-unknown-linux-musl/bin/wash" wash
-      - run: chmod +x wash
+          name: wash-x86_64-unknown-linux-musl
```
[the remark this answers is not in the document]
rvolosatovs: Not sure what you mean. Also, wouldn't evaluation of this workflow fail if there was?

.github/workflows/wasmcloud.yml
rvolosatovs: As mentioned in the comment we cannot do that due to https://github.com/rust-lang/rust/issues/92212. This has been a problem since 2023 (https://github.com/wasmCloud/wasmCloud/pull/807)

The only way this would work today is if `wash` binary would be smaller and had fewer dependencies, which is unlikely to happen soon. The other option is to either wait for either https://github.com/rust-lang/rust/issues/92212 to be fixed or use some other target (like `pc-windows-gnullvm` https://doc.rust-lang.org/rustc/platform-support/pc-windows-gnullvm.html)

.github/workflows/provider.yml
rvolosatovs: `*-gnu` works just fine on Windows and that is tested in CI.
msvc is Windows-only indeed, since it requires Visual Studio and is really only required to link to things built specifically for Visual Studio toolchain on Windows, which we don't do.
See also https://rust-lang.github.io/rustup/installation/windows.html

PR #4020, opened 2025-01-27
deps: nuid 0.5

brooksmtownsend:
- **chore(deps): upgrade nuid to 0.5**
- **fix(wascap): update to nuid 0.5 api**
- **chore(wascap): bump to 0.15.3 for release**

## Feature or Problem
This PR upgrades nuid to 0.5 and wascap to use the updated API in order to fix compilation issues with the release of rand 0.9. This can be a patch because nuid is only used internally to wascap and all other crates will pick this up as a patch bump.

I've updated this PR to ignore a failing test which we cannot fix until this version of wascap is released, and updated the Rust example providers to the latest versions of their respective libraries.


brooksmtownsend:
Provider builds are expected to fail until wascap 0.15.3 is out. They already fail today

thomastaylor312 (approved):

Issue #3990, opened 2025-01-21
[BUG] Provider process failures require app redeploy and/or host restart


lxfontes:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [x] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

wasmCloud External Providers are regular processes created under the wasmCloud host PID.
The wasmCloud host is not restarting External Provider processes in case of failures.
It reports the process exit and that's all.

### Steps to reproduce

- Start wasmcloud HTTP hello world example ( with external provider ).
- Find the external provider PID using `ps` ( ex: `123` )
- Kill the external provider ( `kill -9 123` )

At this point you should see the error in wasmcloud host log.

### Expected behavior

The host must restart the failed provider.

Personally I don't think we need to re-emit `start_provider` Lattice events, and can rely on the periodic provider healthcheck reporting.
The `start_provider` events have to deal with the initial provider setup ( OCI unpackaging, setting up links ). The healtcheck reports the process state.

### Environment

- OS: [e.g. Linux, MacOS]
- Shell [e.g. bash, zsh, powershell]
- wasmcloud Version [e.g. 0.81.0]

brooksmtownsend:
> Personally I don't think we need to re-emit start_provider Lattice events, and can rely on the periodic provider healthcheck reporting.

Agreed, we can lean on the health check failing + emitting failed health events for this

@lxfontes what do you think about this mechanism also attempting to restart the provider if it reports its status as failed? My gut is that initially this may lead to unfortunate unexpected behavior if a provider developer reports a status message that includes information but doesn't mean the provider should restart.

lxfontes:
I think for now we only restart in case the process dies/goes away.

The "host -> provider" healthcheck will likely need a few more knobs. as in: In this situation it likely means the provider has a dependency failure downstream ( ex a database ), and restarting it could make things worse.
Long term I expect healthcheck knobs such as: `interval`, `http_expected_status`, `failure_count`, `success_count` .

Similar to https://developer.hashicorp.com/nomad/docs/job-specification/check and https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#probe-v1-core

PR #4021, opened 2025-01-27
feat(host): supervise providers, restart on exit

brooksmtownsend:
## Feature or Problem
This PR introduces a feature that is likely what users expect from the host, which is that when it is supervising a capability provider it will ensure that the provider gets restarted in the case of a `kill -9` or a binary exit. Before this PR, if a provider is killed forcefully or panics, it will exit and the host will still report it as running until manually restarted. After this PR, a provider that exits before it's stopped by wasmCloud will restart with the same parameters. 

This PR also makes a few refactors to push the binary provider execution logic into the `wasmbus::providers` module

Marked as draft as there's one more aspect of restarting to take into account, which is that we need to fetch new links, config, secrets, etc which came up after the provider initially started

## Related Issues
#3990 

## Release Information
wasmCloud 1.6.0

### Manual Verification
I've manually verified this restarts the provider properly, waits the five seconds if there's a consistent failure, and doesn't restart during the shutdown process.

### DIFF
--- a/crates/host/src/wasmbus/providers/mod.rs
+++ b/crates/host/src/wasmbus/providers/mod.rs
@@ -291,8 +291,9 @@ impl Host {
                         // When the provider is shutting down, don't restart it
                         if shutdown.load(Ordering::Relaxed) {
                             trace!(
-                                "provider @ [{}] exited with `{status:?}` but will not be restarted since it's shutting down",
-                                path.display()
+                                path = ?path.display(),
+                                status = ?status,
+                                "provider exited but will not be restarted since it's shutting down",
                             );
                             // Avoid a hot loop by waiting 1s before checking the status again
                             tokio::time::sleep(Duration::from_secs(1)).await;
@@ -300,11 +301,12 @@ impl Host {
                         }
 
                         warn!(
-                            "restarting provider @ [{}] that exited with `{status:?}`",
-                            path.display()
+                            path = ?path.display(),
+                            status = ?status,
+                            "restarting provider that exited while being supervised",
                         );
 
-                        let Ok((Ok(host_data), new_config_bundle)) = self
+                        let (host_data, new_config_bundle) = match self
                             .prepare_provider_config(
                                 &config_names,
                                 claims_token.as_ref(),
@@ -319,11 +321,20 @@ impl Host {
                                         .context("failed to serialize provider data"),
                                     Arc::new(RwLock::new(config)),
                                 )
-                            })
-                        else {
-                            error!("failed to prepare provider host data while restarting");
-                            shutdown.store(true, Ordering::Relaxed);
-                            return;
+                            }) {
+                            Ok((Ok(host_data), new_config_bundle)) => {
+                                (host_data, new_config_bundle)
+                            }
+                            Err(e) => {
+                                error!(err = ?e, "failed to prepare provider host data while restarting");
+                                shutdown.store(true, Ordering::Relaxed);
+                                return;
+                            }
+                            Ok((Err(e), _)) => {
+                                error!(err = ?e, "failed to serialize provider host data while restarting");
+                                shutdown.store(true, Ordering::Relaxed);
+                                return;
+                            }
                         };
 
                         // Stop the config watcher and start a new one with the new config bundle
@@ -339,7 +350,7 @@ impl Host {
                         // Restart the provider by attempting to re-execute the binary with the same
                         // host data
                         let Ok(child_cmd) = provider_command(&path, host_data).await else {
-                            error!("failed to restart provider @ [{}]", path.display());
+                            error!(path = ?path.display(), "failed to restart provider");
                             shutdown.store(true, Ordering::Relaxed);
                             return;
                         };
@@ -351,8 +362,9 @@ impl Host {
                     }
                     Err(e) => {
                         error!(
-                            "failed to wait for provider @ [{}] to execute: {e}",
-                            path.display()
+                            path = ?path.display(),
+                            err = ?e,
+                            "failed to wait for provider to execute",
                         );
 
                         shutdown.store(true, Ordering::Relaxed);
--- a/crates/host/src/wasmbus/providers/mod.rs
+++ b/crates/host/src/wasmbus/providers/mod.rs
@@ -19,9 +19,9 @@ use cloudevents::EventBuilderV10;
 use futures::{stream, Future, StreamExt};
 use nkeys::XKey;
 use tokio::io::AsyncWriteExt;
+use tokio::process;
 use tokio::sync::RwLock;
 use tokio::task::JoinSet;
-use tokio::{process, select};
 use tracing::{error, instrument, trace, warn};
 use uuid::Uuid;
 use wascap::jwt::{CapabilityProvider, Token};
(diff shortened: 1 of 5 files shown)


crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,359 @@
...
+pub(crate) struct Provider {
+    pub(crate) image_ref: String,
+    pub(crate) claims_token: Option<jwt::Token<jwt::CapabilityProvider>>,
+    pub(crate) xkey: XKey,
+    pub(crate) annotations: Annotations,
+    #[allow(unused)]
+    /// Config bundle for the aggregated configuration being watched by the provider
+    pub(crate) config: Arc<RwLock<ConfigBundle>>,
+    /// Whether the provider should be restarted on exit
+    pub(crate) restart_on_exit: Arc<AtomicBool>,
```
thomastaylor312, line 42: Nit: I'd add some extra docs here. The naming kinda threw me off as it seems like this is a configurable value, but it is really only used for builtins as a configuration value. Otherwise I would have said this would make more sense as a variable called `is_shutting_down` or something like that

crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,359 @@
...
+) -> impl Future<Output = ()> {
+    let subject = provider_config_update_subject(&lattice, &provider_id);
+    trace!(?provider_id, "starting config update listener");
+    async move {
+        loop {
+            let mut config = config.write().await;
+            select! {
+                maybe_update = config.changed() => {
+                    let Ok(update) = maybe_update else {
+                        // TODO: shouldn't this be continue?
```
thomastaylor312, line 333: I think this should probably at least log something. This would mean there was an error when listening for the config changed event

crates/host/src/wasmbus/providers/mod.rs
brooksmtownsend: Yep you got it, naming this one is a little weird but I will try to make it most descriptive. Essentially all I want to do is pause the provider restarts when it's being stopped

crates/host/src/wasmbus/providers/mod.rs
brooksmtownsend: I'd like to address these config TODOs in a followup, but I definitely wanted to call this out because it felt weird when I was reading over it

crates/host/src/wasmbus/providers/mod.rs
brooksmtownsend: Renamed to `shutdown`, which is an indicator of, well, the provider shutting down.

crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,563 @@
...
+                            trace!(
+                                "provider @ [{}] exited with `{status:?}` but will not be restarted since it's shutting down",
+                                path.display()
+                            );
+                            // Avoid a hot loop by waiting 1s before checking the status again
+                            tokio::time::sleep(Duration::from_secs(1)).await;
+                            continue;
+                        }
+
+                        warn!(
```
thomastaylor312, line 302: Can you make the path and status fields instead of interpolating them?

crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,563 @@
...
+                shutdown.clone(),
+            ));
+            loop {
+                let mut child = child.write().await;
+                match child.wait().await {
+                    Ok(status) => {
+                        // When the provider is shutting down, don't restart it
+                        if shutdown.load(Ordering::Relaxed) {
+                            trace!(
+                                "provider @ [{}] exited with `{status:?}` but will not be restarted since it's shutting down",
```
thomastaylor312, line 294: Same thing here with interpolation

crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,563 @@
...
+                            .await
+                            .map(|(host_data, config)| {
+                                (
+                                    serde_json::to_vec(&host_data)
+                                        .context("failed to serialize provider data"),
+                                    Arc::new(RwLock::new(config)),
+                                )
+                            })
+                        else {
+                            error!("failed to prepare provider host data while restarting");
```
thomastaylor312, line 324: Should we log the error message here?

crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,563 @@
...
+                            new_config_bundle,
+                            Arc::clone(&lattice),
+                            provider_id.clone(),
+                            shutdown.clone(),
+                        ));
+
+                        // Restart the provider by attempting to re-execute the binary with the same
+                        // host data
+                        let Ok(child_cmd) = provider_command(&path, host_data).await else {
+                            error!("failed to restart provider @ [{}]", path.display());
```
thomastaylor312, line 342: Same thing with interpolation

crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,563 @@
...
+                        };
+                        *child = child_cmd;
+
+                        // To avoid a tight loop, we wait 5 seconds after restarting. In the worst case,
+                        // the provider will continually execute and exit every 5 seconds.
+                        tokio::time::sleep(Duration::from_secs(5)).await;
+                    }
+                    Err(e) => {
+                        error!(
+                            "failed to wait for provider @ [{}] to execute: {e}",
```
thomastaylor312, line 354: Same thing with interpolation

crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -1,2 +1,563 @@
...
+                                    ?provider_id,
+                                    "failed to deserialize provider health check response"
+                                ),
+                            }
+                        }
+                        else {
+                            warn!(?provider_id, "failed to request provider health, retrying in 30 seconds");
+                        }
+                }
+                true = async { shutdown.load(Ordering::Relaxed) } => {
```
thomastaylor312, line 511: I feel like this is basically going to spinloop here because the future will return almost immediately and then enter the next loop. If the shutdown is going to be used this way, maybe it would be better to use `tokio::sync::watch` and then use the `changed()` method as the future to wait on here. In other places where you want to return the current value syncronously, you could use `borrow()` method

thomastaylor312 (changes requested):
Mostly some changes around putting things as log fields instead of interpolating them in the log messages. There is one request around what structure we're using to store the shutdown bool that I'm curious about your thoughts on

crates/host/src/wasmbus/providers/mod.rs
brooksmtownsend: Yeah probably 😄

crates/host/src/wasmbus/providers/mod.rs
brooksmtownsend: Upon further reflection, I don't even need to have this branch in the `select!` statement because shutting down aborts all of the tasks anyways 😄

brooksmtownsend: In the case where the provider fails to restart, we will still have the health check tasks available now to report that it is unhealthy as well.

PR #4032, opened 2025-01-29
refactor(host): create claims and jetstream modules

brooksmtownsend:
## Feature or Problem
This PR includes no fixes/features and is solely a refactor of host claims logic to a `claims` module and host NATS KV connection/management logic into a `jetstream` module. The hope is that by breaking out some of these larger pieces of the host crate we can more easily manage specific features.


crates/host/src/wasmbus/jetstream.rs
```diff
@@ -0,0 +1,303 @@
...
+        self.links.write().await.insert(id.to_string(), spec.links);
+
+        Ok(())
+    }
+
+    #[instrument(level = "debug", skip_all)]
+    pub(crate) async fn process_component_spec_delete(
+        &self,
+        id: impl AsRef<str>,
+        _value: impl AsRef<[u8]>,
```
thomastaylor312, line 160: While we're here: Do we even need these two parameters?

crates/host/src/wasmbus/mod.rs
```diff
@@ -20,15 +20,17 @@ use std::task::{Context, Poll};
...
 use base64::engine::general_purpose::STANDARD;
 use base64::Engine;
 use bytes::{BufMut, Bytes, BytesMut};
+use claims::{Claims, StoredClaims};
 use cloudevents::{EventBuilder, EventBuilderV10};
 use futures::future::Either;
 use futures::stream::{AbortHandle, Abortable, SelectAll};
 use futures::{join, stream, try_join, Stream, StreamExt, TryFutureExt, TryStreamExt};
 use hyper_util::rt::{TokioExecutor, TokioIo};
+use jetstream::{create_bucket, ComponentSpecification};
```
thomastaylor312, line 33: To ensure complete backwards compat, we need to have a `pub use jetstream::ComponentSpecification` since it was `pub` at this top level before

thomastaylor312 (changes requested):

thomastaylor312:
I think this separation is great! Just left the one note of something that would need to be changed for complete backcompat

crates/host/src/wasmbus/jetstream.rs
brooksmtownsend: Removed while we were at it 😄

crates/host/src/wasmbus/mod.rs
brooksmtownsend: Great catch thank you!

thomastaylor312 (approved):

PR #4037, opened 2025-01-30
fix(wash): Removes required for link name

thomastaylor312:
The `link-name` flag in `wash link del` should not be required as there is a default value

### DIFF
--- a/crates/wash-lib/src/cli/link.rs
+++ b/crates/wash-lib/src/cli/link.rs
@@ -16,7 +16,7 @@ pub struct LinkDelCommand {
     pub source_id: Option<String>,
 
     /// Link name, defaults to "default"
-    #[clap(short = 'l', long = "link-name", required_unless_present("all"))]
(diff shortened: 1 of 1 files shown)


crates/wash-lib/src/cli/link.rs
```diff
@@ -16,7 +16,7 @@ pub struct LinkDelCommand {
     pub source_id: Option<String>,
 
     /// Link name, defaults to "default"
-    #[clap(short = 'l', long = "link-name", required_unless_present("all"))]
+    #[clap(short = 'l', long = "link-name")]
```
thomastaylor312, line 19: @vados-cosmonic would like your review here to make sure I didn't miss a reason why this should be set

vados-cosmonic: Ah yes, `required_unless_present` shouldn't have been added there -- we do have a default, it works fine without

vados-cosmonic (approved):
LGTM :rocket:

PR #4040, opened 2025-01-31
chore(ci): Removes magic nix cache action

thomastaylor312:
Due to the deprecation of the GH cache API it relies on, we have to remove this step from installing nix. I also removed the env var in the action because we aren't using the caching step there anymore and `env` can't be set at the top level of an action unlike in a workflow

### DIFF
--- a/.github/actions/install-nix/action.yml
+++ b/.github/actions/install-nix/action.yml
@@ -4,24 +4,18 @@ inputs:
   cachixAuthToken:
     description: auth token for https://app.cachix.org/organization/wasmcloud/cache/wasmcloud
 
-env:
-  SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1 # abort caching attempt if it's slow
-
 runs:
   using: composite
   steps:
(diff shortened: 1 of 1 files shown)


.github/actions/install-nix/action.yml
```diff
@@ -4,24 +4,18 @@ inputs:
   cachixAuthToken:
     description: auth token for https://app.cachix.org/organization/wasmcloud/cache/wasmcloud
 
-env:
-  SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1 # abort caching attempt if it's slow
```
rvolosatovs, line 8: I think this was actually for Cachix, but I don't actually remember anymore, so I guess it's fine to remove

rvolosatovs (approved):

.github/actions/install-nix/action.yml
thomastaylor312: I checked and did see this as an env var for the cache action so I figured that is what it was for.

PR #4054, opened 2025-02-05
chore(providers): bump all for release

brooksmtownsend:
## Feature or Problem
This PR bumps all capability providers for release of a new minor version after two key changes:
1. Fixing a bug in deleting links
1. Embedding WIT into each provider archive

This PR does not include the messaging provider because it already was bumped.

## Related Issues
#3924 
#4013


vados-cosmonic (approved):
LGTM :rocket: 

A little worried about the pre-emptively updated provider version (hopefully it uses the local path and doesn't try to fetch that crate that doesn't exist yet)... But I can't remember if that's going to happen or not when CI/flows run.

PR #4055, opened 2025-02-05
ci(wasmcloud): build wash for provider releases

brooksmtownsend:
## Feature or Problem
This PR ensures we build the wash binary for provider releases.

## Related Issues
See https://github.com/wasmCloud/wasmCloud/actions/runs/13160153715 which did not build wash and therefore did not release the provider


vados-cosmonic (approved):
LGTM :rocket: 

Really wonder if we should split this out and not have one massive workflow for releasing everything... but not today!

Issue #3821, opened 2024-12-12
[FEATURE] support cronjob provider


Patrick0308:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

I'am a wasmcloud new guy. I have some cronjob want to be executed in wasmcloud. CronJob is a basic feature in cloud. 

### Describe the solution you'd like

Add a cronjob provider

brooksmtownsend:
A Cronjob provider sounds like a great idea, and I think the first step here would just be to propose a WIT interface and then get started 😄 

Possible inspiration https://github.com/wasmCloud/examples/tree/main/cron

The provider could just be as simple as accepting a cron statement as configuration, calling the component on that interval 😄

Aditya1404Sal:
Hey @brooksmtownsend 
I'd love to take this on 😄
Could you assign this to me ?

thomastaylor312:
Hey @Aditya1404Sal I'll assign you! I'd highly recommend diagramming or writing out what you'd like to do here because the distributed nature of this is the hard part (e.g. if you have more than one provider, you should only send one notification)

Aditya1404Sal:
> Hey @Aditya1404Sal I'll assign you! I'd highly recommend diagramming or writing out what you'd like to do here because the distributed nature of this is the hard part (e.g. if you have more than one provider, you should only send one notification)

Sure thing @thomastaylor312 that does sound a bit hard but I'll present my approach as soon as I'm done with a thorough bit of research.

Aditya1404Sal:
@thomastaylor312 , Is this the situation you're talking about ? where there are multiple providers and the cron notification should only be sent by a single provider ?
As per my understanding NATS queue group would ensure that the Host-> Component and Component -> Provider for invoking the exported `timed_invoke` function would work maintaining deduplication. However there would be some problems when the provider is sending the cron notification, there would essentially need to be a leader provider that will send it. 
[...]

PR #4061, opened 2025-02-05
WIP : [ feat(provider): Add a Cronjob capability provider ]

Aditya1404Sal:
## Feature or Problem

WIP : Adding Host runtime support for a timed_invoke function call 
- Added `cron/cronjob.wit`  that has a standard interface called `scheduler` 
## Related Issues

#3821


wit/cron/wit/cron.wit
```diff
@@ -0,0 +1,5 @@
+package wasmcloud:cron@0.1.0-draft;
```
thomastaylor312, line 1: Let's just stick with 0.1.0 for now. I know we have the `-draft` thing in the wasi cloud interfaces (which I actually don't really like), but we can just tag when we're ready and then bump versions from there

wit/cron/wit/cron.wit
```diff
@@ -0,0 +1,5 @@
...
+
+interface scheduler {
+       timed-invoke : func(payload: list<u8>) -> result<_, string>;
```
thomastaylor312, line 4: I think `invoke` should be fine here as we can assume from the interface name that it is scheduled

wit/cron/wit/cron.wit
```diff
@@ -0,0 +1,5 @@
+package wasmcloud:cron@0.1.0-draft;
+
+interface scheduler {
```
thomastaylor312, line 3: Super nit (don't do anything with this yet as I'm just thinking out loud): `scheduler` feels like a weird name to me. Based on other interface naming patterns, maybe it could be `schedule` or `scheduled`?

Aditya1404Sal: I thought it resembled `handler` and `watcher` so I can up with that, but I can see why it sounds a bit weird 😅

wit/cron/wit/cron.wit
Aditya1404Sal: Ah I understand, I'll get it refactored

brooksmtownsend:
Hey @Aditya1404Sal I think this is looking great, and apologies in advance for requesting this, but I think that it would be best for us to not build in support for `cron.wit` inside of the wasmCloud runtime.

There are a couple of reasons why I think so:
1. If we build it into the runtime, then that means that changing the interface would require new releases of the host
2. Any breaking changes in the interface require a breaking change release of the host
3. Most of the built-in interfaces are only built in to enable better support (e.g. with streams or resources) but for things like `wasmcloud:postgres` we've kept it separate

Let me know what you think, but otherwise I think this is looking nice and with @thomastaylor312 's feedback we should be able to push this across the line

Aditya1404Sal:
> Hey @Aditya1404Sal I think this is looking great, and apologies in advance for requesting this, but I think that it would be best for us to not build in support for `cron.wit` inside of the wasmCloud runtime.
> 
> There are a couple of reasons why I think so:
> 
> 1. If we build it into the runtime, then that means that changing the interface would require new releases of the host
> 2. Any breaking changes in the interface require a breaking change release of the host
> 3. Most of the built-in interfaces are only built in to enable better support (e.g. with streams or resources) but for things like `wasmcloud:postgres` we've kept it separate
> 
> Let me know what you think, but otherwise I think this is looking nice and with @thomastaylor312 's feedback we should be able to push this across the line

I totally understand and agree with your reasoning — keeping cron.wit out of the core runtime makes sense given the implications of interface changes being dragged along with the host release. I’ll refactor accordingly and make sure it’s handled separately.

Aditya1404Sal:
@brooksmtownsend This PR is ready for a review.
I'm unsure as to why the error trace isn't showing up in the test debug trace, 
The latest commit introduces a distributed variant of the cron job provider, while previously the tests ran successfully for the single component variant, I just don't see what's going wrong with this approach, because usually the error trace is visible pretty clearly, is there a specific arrangement that allows the error propogation for a new provider ?

A Breakdown of the current logic:
1) cron config gets parsed into different cron job ID's with their respective job id and payload
2) for every cron job, a NATS jetstream is created to facilitate the distributed locking, this stream has a `MaxAge` arguement which acts as a TTL, and via the `subject_delete_marker`, the last message of the subject expires the MaxAge limit, a Delete_Marker is propogated to the same Nats subject with a the header : `Nats_Marker_Reason` with value  : `MaxAge`
3) a task is spawned for each cron job id , where every task basically subscribes to the stream using a push consumer and the distributed lock is coordinated using the nats-kv `create` function (a key-value pair cannot be duplicated so it's creation actually means that the instance was the first to do so, automatically becoming the leader and invoking the wrpc call for the cron task
4) After the cron task execution there are 2 flows that are followed : 
 - a) if the cron expression was simple (ie with regular intervals) then the message is just republished, therefore adding the message to the stream and starting the global timer till it expires, after which it would send out a delete marker.
 - b) if the cron expression was complex (ie with irregular and spaced out intervals) then the time interval till next execution is calculated and the stream's MaxAge property is modified and the message is republished.

Aditya1404Sal:
@brooksmtownsend  --  please hold off the review !! NATS now supports per-message TTL 😭 -- as soon as v 2.11.2 get's incorporated with the test suite and codebase, I'll update  the current implementation to a much easier and straight-forward approach.

Aditya1404Sal:
Hey @brooksmtownsend -- this PR is ready for a review !!

brooksmtownsend (changes requested):
@Aditya1404Sal this is looking really nice! excited to try it out 😄 Just one hard request but otherwise it's looking ready for testing

Aditya1404Sal:
Hi @brooksmtownsend , thank you for the clarification!

Just to confirm — I should completely remove the runtime support for this provider, including the corresponding test case, correct?
Please let me know if I’ve misunderstood anything.

Aditya1404Sal:
Hey @brooksmtownsend -- just a quick update: I’ve removed the runtime implementation, so this provider is no longer bundled with the host. It’s all set and ready for your review whenever you get a chance!

Issue #4057, opened 2025-02-05
[FEATURE] wash support for cargo component


ricochet:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

```bash
gh repo clone bytecodealliance/sample-wasi-http-rust
cd sample-wasi-http-rust
cargo component build
cargo component serve

curl http://0.0.0.0:8080/
Hello, wasi:http/proxy world!
```

**feature request**: We can't build projects without a `wasmcloud.toml`:
```bash
# fails
wash dev
```

I have to create a `wasmcloud.toml`:
```bash
cat <<EOF > wasmcloud.toml
name = "sample-wasi-http-rust"
language = "rust"
type = "component"

[component]
wasm_target = "wasm32-wasip2"
EOF
```

Now I can build:
```bash
wash dev
```

But fails with:
```bash
curl http://127.0.0.1:8000
failed to invoke `wrpc:http/incoming-handler.handle`: failed to invoke `wrpc:http/incoming-handler@0.1.0.handle`: incomplete results% 
```

```bash
wash inspect build/rust_wasi_hello.wasm --wit

package root:component;

world root {
  import wasi:io/poll@0.2.3;
  import wasi:clocks/monotonic-clock@0.2.3;
  import wasi:clocks/wall-clock@0.2.3;
  import wasi:random/random@0.2.3;
  import wasi:io/error@0.2.3;
  import wasi:io/streams@0.2.3;
  import wasi:cli/stdout@0.2.3;
  import wasi:cli/stderr@0.2.3;
  import wasi:cli/stdin@0.2.3;
  import wasi:http/types@0.2.3;
  import wasi:cli/environment@0.2.3;
  import wasi:cli/exit@0.2.3;
  import wasi:filesystem/types@0.2.3;
  import wasi:filesystem/preopens@0.2.3;

  export wasi:http/incoming-handler@0.2.3;
}
```

In the `Cargo.toml` we specify:
```toml
[package.metadata.component]
package = "component:rust-wasi-hello"
# This tells cargo-component to use a special adapter, which doesn't depend on
# `get-environment` or other things not present in the proxy world.
proxy = true
```

```bash
[...]

PR #4062, opened 2025-02-05
feat(rt): fully support WASI >0.2.2

### LINKED ISSUE #4063: [BUG] wasi:http interfaces collide and differ (0.2.2/0.2.0, 0.2.2/0.2.3)
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

When building a JS component using the latest `jco`, `componentize-js` and StarlingMonkey, we ran into a problem -- WIT interfaces that *should* be the same are not being recognized/loaded in with the component:

The post-build WIT interface looks like this:

```
package root:component;

world root {
  import wasi:io/poll@0.2.3;
  import wasi:clocks/monotonic-clock@0.2.3;
  import wasi:io/error@0.2.3;
  import wasi:io/streams@0.2.3;
  import wasi:http/types@0.2.2; // <---
  import wasi:cli/stdin@0.2.3;
  import wasi:cli/stdout@0.2.3;
  import wasi:cli/stderr@0.2.3;
  import wasi:cli/terminal-input@0.2.3;
  import wasi:cli/terminal-output@0.2.3;
  import wasi:cli/terminal-stdin@0.2.3;
  import wasi:cli/terminal-stdout@0.2.3;
  import wasi:cli/terminal-stderr@0.2.3;
  import wasi:clocks/wall-clock@0.2.3;
  import wasi:filesystem/types@0.2.3;
  import wasi:filesystem/preopens@0.2.3;
  import wasi:random/random@0.2.3;

  export wasi:http/incoming-handler@0.2.2; // <---
}
```

Importantly, we found that we'd get an error:

```
2025-02-05T18:42:47.768753Z ERROR wasmcloud_host::wasmbus: failed to scale component component_ref=file:///path/to/http-hello-world/build/http_hello_world_s.wasm component_id=Ymdzvb-typescript-http-hello-world err=component imports instance `wasi:http/types@0.2.0`, but a matching implementation was not found in the linker

Caused by:
    0: instance export `[method]incoming-body.stream` has the wrong type
    1: type mismatch with results
    2: resource type mismatch
```

The situation seems to be the following:
- StarlingMonkey brings in `*@0.2.3` versions
- Using `wash wit deps` with `wasi:http/incoming-handler@0.2.2` locally pulls in `wasi:http/types@0.2.2` and `wasi:http/incoming-handler@0.2.2`
- Trying to use the component

Note that the same component *does* run in `wasmtime serve`, so it is something on the wasmCloud side that is causing the mismatch here.

I wonder if it's one of two things:
- Our loading (translation?) of interfaces used
~~- Using an outdated version of `wasmtime` (we're on `v26`)~~

> [!NOTE]
> We know that it's not *only* using an outdated version of wasmtime because we tried w/ a wasmtime binary from the 26.0.0 release, and it did not fail.

To build the component you can either use [the branch under development](https://github.com/wasmCloud/typescript/pull/92) or use the attached tarball.

[http-hello-world.tar.gz](https://github.com/user-attachments/files/18678169/http-hello-world.tar.gz)

### Steps to reproduce

1. Start a wasmCloud host (i.e. `wash up`)
2. Attempt to `wash start component` the hello world component (which has `wasi:http/types@0.2.2`, `wasi:http/incoming-handler@0.2.2`, and `wasi:http/*@0.2.3` mixed)
3. Watch wasmCloud fail to load the component

### Expected behavior

wasmCloud shouldn't fail to load the component

### Environment

```
wash          v0.38.0
├ nats-server v2.10.20
├ wadm        v0.19.0
└ wasmcloud   v1.4.2
```

_No response_

rvolosatovs:
## Feature or Problem

Use semver to check whether we should polyfill instances, rather than hard-coding exact names. This ensures host is forwards-compatible with WASI versions it does not know about.

## Related Issues

https://github.com/wasmCloud/wasmCloud/issues/4057

Closes #4063

### Manual Verification

Tested with component from https://github.com/wasmCloud/wasmCloud/issues/4057 (https://github.com/bytecodealliance/sample-wasi-http-rust), but via `cargo component`

### DIFF
--- a/crates/runtime/src/component/mod.rs
+++ b/crates/runtime/src/component/mod.rs
@@ -68,121 +68,19 @@ pub enum ReplacedInstanceTarget {
     HttpOutgoingHandler,
 }
 
-/// skips instance names, for which static (builtin) bindings exist
-macro_rules! skip_static_instances {
-    ($instance:expr) => {
-        match ($instance) {
-            "wasi:blobstore/blobstore@0.2.0-draft"
-            | "wasi:blobstore/container@0.2.0-draft"
-            | "wasi:blobstore/types@0.2.0-draft"
-            | "wasi:cli/environment@0.2.0"
-            | "wasi:cli/environment@0.2.1"
-            | "wasi:cli/environment@0.2.2"
-            | "wasi:cli/exit@0.2.0"
-            | "wasi:cli/exit@0.2.1"
-            | "wasi:cli/exit@0.2.2"
-            | "wasi:cli/stderr@0.2.0"
-            | "wasi:cli/stderr@0.2.1"
-            | "wasi:cli/stderr@0.2.2"
-            | "wasi:cli/stdin@0.2.0"
-            | "wasi:cli/stdin@0.2.1"
-            | "wasi:cli/stdin@0.2.2"
-            | "wasi:cli/stdout@0.2.0"
-            | "wasi:cli/stdout@0.2.1"
-            | "wasi:cli/stdout@0.2.2"
-            | "wasi:cli/terminal-input@0.2.0"
-            | "wasi:cli/terminal-input@0.2.1"
-            | "wasi:cli/terminal-input@0.2.2"
-            | "wasi:cli/terminal-output@0.2.0"
-            | "wasi:cli/terminal-output@0.2.1"
-            | "wasi:cli/terminal-output@0.2.2"
-            | "wasi:cli/terminal-stderr@0.2.0"
-            | "wasi:cli/terminal-stderr@0.2.1"
-            | "wasi:cli/terminal-stderr@0.2.2"
-            | "wasi:cli/terminal-stdin@0.2.0"
-            | "wasi:cli/terminal-stdin@0.2.1"
-            | "wasi:cli/terminal-stdin@0.2.2"
-            | "wasi:cli/terminal-stdout@0.2.0"
-            | "wasi:cli/terminal-stdout@0.2.1"
-            | "wasi:cli/terminal-stdout@0.2.2"
-            | "wasi:clocks/monotonic-clock@0.2.0"
-            | "wasi:clocks/monotonic-clock@0.2.1"
-            | "wasi:clocks/monotonic-clock@0.2.2"
-            | "wasi:clocks/timezone@0.2.1"
-            | "wasi:clocks/timezone@0.2.2"
-            | "wasi:clocks/wall-clock@0.2.0"
-            | "wasi:clocks/wall-clock@0.2.1"
-            | "wasi:clocks/wall-clock@0.2.2"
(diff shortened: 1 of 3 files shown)


crates/runtime/src/component/mod.rs
```diff
@@ -68,121 +68,19 @@ pub enum ReplacedInstanceTarget {
...
+fn is_0_2(version: &str, min_patch: u64) -> bool {
+    if let Ok(semver::Version {
+        major,
+        minor,
+        patch,
+        pre,
+        build,
+    }) = version.parse()
+    {
+        major == 0 && minor == 2 && patch >= min_patch && pre.is_empty() && build.is_empty()
```
brooksmtownsend, line 80: Do we want to have the `pre.is_empty()` and `build.is_empty()` qualifiers here?

Put another way, would we expect for the following interfaces to work okay?
1. `wasi:http@0.2.5-pre.1`
1. `wasi:http@0.2.5+20250205`

brooksmtownsend (approved):
All LGTM, this might be a good candidate for a couple unit tests to make sure we can validate this is all working properly and we don't mess up this logic in the future for interfaces we want to support. I'm fine if that goes in as a followup since we didn't have that before.

One non-blocking question so I'm approving since this is better than previous 😄

vados-cosmonic:
This PR resolves https://github.com/wasmCloud/wasmCloud/issues/4063

crates/runtime/src/component/mod.rs
rvolosatovs: Such interfaces don't exist today in wasip2.
We can revisit this once/if that changes.

In any case we would need to follow what Wasmtime does here, since otherwise this would cause instantiation failures.

crates/runtime/src/component/mod.rs
```diff
@@ -448,12 +346,52 @@ where
...
+                    if is_0_2(version, 0) => {}
+                Some((
+                    "wasi:sockets",
+                    "instance-network" | "ip-name-lookup" | "network" | "tcp-create-socket" | "tcp"
+                    | "udp-create-socket" | "udp",
+                    Some(version),
+                )) if is_0_2(version, 0) => {}
+                _ if rt.skip_feature_gated_instance(name) => {}
+                _ => link_item(&engine, &mut linker.root(), [], ty, "", name, None)
+                    .context("failed to link item")?,
```
vados-cosmonic, line 393:
```suggestion
                    .context("failed to link item")?,
```

More information about what failed to link would be great here via `with_context(...)`, but not blocking on that.

vados-cosmonic (approved):
LGTM :rocket:

PR #4071, opened 2025-02-06
ci: run Nix with `--fallback`

rvolosatovs:
## Feature or Problem

Sometimes, connection to Nix network substituters (like Cachix) can fail (like in https://github.com/wasmCloud/wasmCloud/actions/runs/13174200900/job/36770534710)

In such cases, fall back to building from source rather than failing CI run.

See https://nix.dev/manual/nix/2.25/command-ref/nix-build.html#opt-fallback for more info

### DIFF
--- a/.github/actions/build-nix/action.yml
+++ b/.github/actions/build-nix/action.yml
@@ -8,9 +8,9 @@ inputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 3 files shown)


vados-cosmonic (approved):
LGTM :rocket:

PR #4076, opened 2025-02-06
ci: wasmcloud wash separate releases

brooksmtownsend:
## Feature or Problem
This PR separates the release jobs for wasmcloud and wash, which notably drop the requirement for wash to be built to release wasmcloud and vice versa.

I also included something nice in this PR to remove all of the dependabot commits from our generated release notes, which should be useful while we research conventional commit tools to do this 😄 

## Related Issues
When I tried to release wasmcloud v1.6.0 it didn't complete because the job required wash to be built https://github.com/wasmCloud/wasmCloud/actions/runs/13186620296

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,6 @@
+# .github/release.yml
+
+changelog:
+  exclude:
+    authors:
+      - dependabot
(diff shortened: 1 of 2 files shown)


.github/release.yml
```diff
@@ -0,0 +1,6 @@
+# .github/release.yml
+
+changelog:
+  exclude:
+    authors:
+      - dependabot
```
brooksmtownsend, line 6: Sourced from https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options

thomastaylor312 (commented):
Yay! Thanks for doing this

PR #4077, opened 2025-02-06
ci: separate oci action into reusable workflow

brooksmtownsend:
## Feature or Problem
~This PR enables building the wash binary on wasmCloud releases so the `oci` action can run. I could copy/paste that action into two separate blocks, but it truly is better as a matrix / reusable workflow.~

~I'll consider making that a reusable workflow as well. For now I was just opting for the simplest solution.~

This PR pulls the `oci` action out into a reusable workflow and calls it from `oci-wash` and `oci-wasmcloud` instead of in a matrix. This is helpful to ensure that we can use the same logic for both pieces of code, but not require wash to build and run through OCI for wasmCloud releases and vice versa.

## Related Issues
#4076

### DIFF
--- a/.github/workflows/oci.yml
+++ b/.github/workflows/oci.yml
@@ -24,9 +24,6 @@ on:
       AZURECR_PUSH_PASSWORD:
         description: Password for Azure Container Registry
         required: true
-      GITHUB_TOKEN:
-        description: Token for authenticating with GitHub Container Registry
-        required: true
       DOCKERHUB_PUSH_USER:
         description: Username for DockerHub
         required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/oci.yml
```diff
@@ -0,0 +1,228 @@
+name: oci-release
```
rvolosatovs, line 1: Let's make sure to place all reusable workflows to a consistent location (`.github/actions`)

rvolosatovs (changes requested):
GitHub fails to parse both `oci.yml` and `wasmcloud.yml` (https://github.com/wasmCloud/wasmCloud/actions/runs/13188370169)

Also, please make sure that changes to `oci.yml` cause a CI build

PR #4011, opened 2025-01-25
HttpClient issue #2290 - Added failing test

flyaruu:
## Feature or Problem
HttpClient seems to fail in certain circumstances.

## Related Issues
Failing test for [this issue](https://github.com/wasmCloud/wasmCloud/issues/2990), as requested

Issue #2990, opened 2024-09-12
[BUG] http client fails if body is finished with some empty trailers


peter-jerry-ye:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [X] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

The http client does not perform correctly if sending a request over HTTP/2 and if the body is finished with `Some(Fields::new())`

### Steps to reproduce

Take the example of `examples/rust/components/dog-fetcher`.

Replace lines 36-38 to:

```rust
        req.set_authority(Some("api.github.com")).unwrap(); // sorry github
        req.set_path_with_query(Some("/"))
            .unwrap();
        OutgoingBody::finish(req.body().unwrap(), Some(Fields::new())).unwrap();
```

And run
```bash
wash build
wash up --log-level debug
```
 
In another terminal
```bash
wash app deploy ./local.wadm.yaml
curl localhost:8080
```

### Expected behavior

I'm expected that it returns some 40x response. But here's the actual log:

```
2024-09-12T02:23:33.469382Z DEBUG serve_handle: wasmcloud_provider_http_client: send HTTP request uri=https://api.github.com/
2024-09-12T02:23:33.572288Z DEBUG serve_handle:http_request: h2::client: binding client connection
2024-09-12T02:23:33.572326Z DEBUG serve_handle:http_request: h2::client: client connection bound
2024-09-12T02:23:33.572365Z DEBUG serve_handle:http_request: h2::codec::framed_write: send frame=Settings { flags: (0x0), enable_push: 0, initial_window_size: 2097152, max_frame_size: 16384, max_header_list_size: 16384 }
2024-09-12T02:23:33.572504Z DEBUG serve_handle:http_request:Connection{peer=Client}: h2::codec::framed_write: send frame=WindowUpdate { stream_id: StreamId(0), size_increment: 5177345 }
2024-09-12T02:23:33.572517Z DEBUG serve_handle:http_request:Connection{peer=Client}: h2::codec::framed_write: send frame=Headers { stream_id: StreamId(1), flags: (0x4: END_HEADERS) }
[...]

PR #4085, opened 2025-02-07
feat: rework http-client using http/1.1 and `hyper` directly

rvolosatovs:
Refs https://github.com/wasmCloud/wasmCloud/pull/4011
Refs https://github.com/wasmCloud/wasmCloud/issues/2990

Fixes https://wasmcloud.slack.com/archives/C02LZ3L0X4H/p1734510978985089

Current wasmtime-wasi-http implementation assumes and only uses HTTP/1.1:

https://github.com/bytecodealliance/wasmtime/blob/d943d57e78950da21dd430e0847f3b8fd0ade073/crates/wasi-http/src/types.rs#L375-L448

One example of this is unconditional inclusion on the `Host` header:

https://github.com/bytecodealliance/wasmtime/blob/d943d57e78950da21dd430e0847f3b8fd0ade073/crates/wasi-http/src/http_impl.rs#L71

Some servers (e.g., `google-analytics.com`) seem to prohibit `Host` headers in HTTP/2.

Only use HTTP/1.1 for most compatibility in outgoing connections for now.

See also https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3

This PRs also ensures we now use the request options, e.g. connection timeout

Tested with:
```diff
diff --git a/examples/rust/components/dog-fetcher/local.wadm.yaml b/examples/rust/components/dog-fetcher/local.wadm.yaml
index e329db3a..1ebef31a 100644
--- a/examples/rust/components/dog-fetcher/local.wadm.yaml
+++ b/examples/rust/components/dog-fetcher/local.wadm.yaml
@@ -65,7 +65,7 @@ spec:
     - name: httpclient
       type: capability
       properties:
-        image: ghcr.io/wasmcloud/http-client:0.12.1
+        image: file://../../../../src/bin/http-client-provider/build/http-client-provider.par.gz
         ## To configure OTEL integration for this provider specifically, uncomment the lines below
         # config:
         #   - name: otel
diff --git a/examples/rust/components/dog-fetcher/src/lib.rs b/examples/rust/components/dog-fetcher/src/lib.rs
index 9a1fa7a4..81dc1653 100644
--- a/examples/rust/components/dog-fetcher/src/lib.rs
+++ b/examples/rust/components/dog-fetcher/src/lib.rs
@@ -33,9 +33,11 @@ impl Guest for DogFetcher {
         // Build a request to dog.ceo which returns a URL at which we can find a doggo
         let req = wasi::http::outgoing_handler::OutgoingRequest::new(Fields::new());
         req.set_scheme(Some(&Scheme::Https)).unwrap();
-        req.set_authority(Some("dog.ceo")).unwrap();
-        req.set_path_with_query(Some("/api/breeds/image/random"))
-            .unwrap();
+        req.set_authority(Some("www.google-analytics.com")).unwrap();
+        req.set_method(&Method::Post).unwrap();
+        req.set_path_with_query(Some("/g/collect")).unwrap();
+        let body = req.body().unwrap();
+        wasi::http::types::OutgoingBody::finish(body, None).unwrap();
 
         // Perform the API call to dog.ceo, expecting a URL to come back as the response body
         let dog_picture_url = match wasi::http::outgoing_handler::handle(req, None) {
@@ -92,6 +94,7 @@ impl Guest for DogFetcher {
         write_stream.write_all(dog_picture_url.as_bytes()).unwrap();
         drop(write_stream);
 
-        OutgoingBody::finish(response_body, None).expect("failed to finish response body");
+        OutgoingBody::finish(response_body, Some(Fields::new()))
+            .expect("failed to finish response body");
     }
 }
```

Interestingly, forcing 1.1 with `hyper-util` did not work, so instead we just use `hyper` directly (just like wasmtime) and implement our own connection pool per-link (lacking from wasmtime).
We should investigate ways to use http 2, if supported to talk to the remotes as a next step, but 1.1 should be good enough for now.


crates/provider-http-client/src/lib.rs
```diff
@@ -1,34 +1,321 @@
...
+
+impl<T> Clone for ConnPool<T> {
+    fn clone(&self) -> Self {
+        Self {
+            http: self.http.clone(),
+            https: self.https.clone(),
+            tasks: self.tasks.clone(),
+        }
+    }
+}
```
brooksmtownsend, line 130: nit: could these be derived?

flyaruu:
Dang, that is quite a bigger deal than I thought, thanks @rvolosatovs

PR #4091, opened 2025-02-10
chore(wash): adjust URL formatting in help

ericgregory:
Angle brackets in the help documentation causes the generated Markdown to error (if not offset with a backtick) when deployed to the docs site. A few angle brackets have crept in around URLs, and this PR removes them.

### DIFF
--- a/crates/wash/src/cli/cmd/up.rs
+++ b/crates/wash/src/cli/cmd/up.rs
@@ -156,7 +156,7 @@ pub struct WasmcloudOpts {
     ///
     /// defaults to the [`WASMCLOUD_HOST_VERSION`] if not provided
     /// or the latest patch version after that when `wash up` issued,
-    /// see https://github.com/wasmCloud/wasmCloud/releases for releases
+    /// see [wasmCloud releases](https://github.com/wasmCloud/wasmCloud/releases).
     #[clap(long = "wasmcloud-version", env = "WASMCLOUD_VERSION")]
     pub wasmcloud_version: Option<String>,
 
@@ -368,7 +368,7 @@ pub struct WadmOpts {
     ///
     /// defaults to the [`WADM_VERSION`] if not provided
     /// or the latest patch version after that when `wash up` issued,
-    /// see https://github.com/wasmCloud/wadm/releases for releases
+    /// see [wadm releases](https://github.com/wasmCloud/wadm/releases).
     #[clap(long = "wadm-version", env = "WADM_VERSION")]
     pub wadm_version: Option<String>,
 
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):
Hrm, this one is a little challenging because that's the structure that docs.rs expects for documentation https://docs.rs/wash-cli/0.38.0/wash_cli/cmd/up/struct.WasmcloudOpts.html#structfield.wasmcloud_version 🤔 

I think I'd rather have accurate docs on the site than documentation links on crates.io

lachieh:
We might be able to handle this on the docs side if it is wanted enough? i.e. replace `<...>` with `[...]`

lachieh:
I think we can probably handle this on the docs side since this format is required for linking in the rust documentation.

crates/wash-cli/src/cmd/up.rs
```diff
@@ -157,7 +157,7 @@ pub struct WasmcloudOpts {
     ///
     /// defaults to the [`WASMCLOUD_HOST_VERSION`] if not provided
     /// or the latest patch version after that when `wash up` issued,
-    /// see <https://github.com/wasmCloud/wasmCloud/releases> for releases
+    /// see https://github.com/wasmCloud/wasmCloud/releases for releases
```
lachieh, line 160:
```suggestion
    /// see [wasmCloud releases](https://github.com/wasmCloud/wasmCloud/releases).
```

crates/wash-cli/src/cmd/up.rs
```diff
@@ -370,7 +370,7 @@ pub struct WadmOpts {
     ///
     /// defaults to the [`WADM_VERSION`] if not provided
     /// or the latest patch version after that when `wash up` issued,
-    /// see <https://github.com/wasmCloud/wadm/releases> for releases
+    /// see https://github.com/wasmCloud/wadm/releases for releases
```
lachieh, line 373:
```suggestion
    /// see [wadm releases](https://github.com/wasmCloud/wadm/releases).
```

crates/wash-lib/src/config.rs
```diff
@@ -73,7 +73,7 @@ pub struct WashConnectionOptions {
     pub ctl_seed: Option<String>,
 
     /// Credsfile for CTL authentication. Combines ctl_seed and ctl_jwt.
-    /// See <https://docs.nats.io/using-nats/developer/connecting/creds> for details.
+    /// See https://docs.nats.io/using-nats/developer/connecting/creds for details.
```
lachieh, line 76:
```suggestion
    /// See [the NATS documentation](https://docs.nats.io/using-nats/developer/connecting/creds) for details.
```

lachieh (changes requested):
Did some digging, it's potentially just docusaurus that doesn't support the [automatic link](https://daringfireball.net/projects/markdown/syntax#autolink) style. Rather than digging too deep, let's just make them explicit markdown links

lachieh (approved):

PR #4096, opened 2025-02-11
feat(ci): Adds benchmarking script

thomastaylor312:
This adds a set of scripts and assets for running a benchmark test using Oracle Cloud provided by the CNCF. Follow up PR will add a pipeline for running the test

### DIFF
--- a/ci/common-env
+++ b/ci/common-env
@@ -1,0 +1,5 @@
+export OCI_TENANCY_ID=ocid1.tenancy.oc1..aaaaaaaanorijchxlkb4zbectmyxugdvqjbjyjg4g7664tdnwwyj3eqsxqha
+export OCI_REGION=us-ashburn-1
+# The `benchmark-tests` compartment
+export OCI_COMPARTMENT_ID="ocid1.compartment.oc1..aaaaaaaaqn6nchyedtj26kt33244nm5bfi4qejjuknup4353n6y7a5ib24cq"
+export SUPPRESS_LABEL_WARNING=True
--- a/ci/run-benchmarks.sh
+++ b/ci/run-benchmarks.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR="$(dirname "$0")"
+
+source "$SCRIPT_DIR/common-env"
+
+# NOTE(thomastaylor312): We have to specify the image ID, so this is the command you can use to the
+# find the latest images and select the one compatible with the k8s version
+# oci ce node-pool-options get --compartment-id "$COMPARTMENT_ID" --node-pool-option-id all
+: ${OCI_MANAGED_NODE_IMAGE_ID:="ocid1.image.oc1.iad.aaaaaaaarkexs7ijdqffxvy6dyzepizvuzm25zijwp2nkv3nifbcvywzufsq"}
+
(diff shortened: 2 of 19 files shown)


ci/setup-cluster-api.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+source "$(dirname "$0")/common-env"
+
+# This script expects the following env vars to be set
+#  OCI_USER_ID
+#  OCI_CREDENTIALS_FINGERPRINT
+#  OCI_CREDENTIALS_KEY
```
brooksmtownsend, line 10: Worth some basic dumb bash to exit early if these aren't set?

ci/run-benchmarks.sh
```diff
@@ -0,0 +1,97 @@
...
+set -euo pipefail
+
+SCRIPT_DIR="$(dirname "$0")"
+
+source "$SCRIPT_DIR/common-env"
+
+# NOTE(thomastaylor312): We have to specify the image ID, so this is the command you can use to the
+# find the latest images and select the one compatible with the k8s version
+# oci ce node-pool-options get --compartment-id "$COMPARTMENT_ID" --node-pool-option-id all
+export OCI_MANAGED_NODE_IMAGE_ID="ocid1.image.oc1.iad.aaaaaaaarkexs7ijdqffxvy6dyzepizvuzm25zijwp2nkv3nifbcvywzufsq"
```
brooksmtownsend, line 12: Should we set this in the pipeline / as a GitHub secret so that we can update the managed image without updating the script? I'm not 100% sure so let me know if I'm thinking about it wrong

ci/common-env
```diff
@@ -0,0 +1,5 @@
+export OCI_TENANCY_ID=ocid1.tenancy.oc1..aaaaaaaanorijchxlkb4zbectmyxugdvqjbjyjg4g7664tdnwwyj3eqsxqha
+export OCI_REGION=us-ashburn-1
+# The `benchmark-tests` compartment
+export OCI_COMPARTMENT_ID="ocid1.compartment.oc1..aaaaaaaaqn6nchyedtj26kt33244nm5bfi4qejjuknup4353n6y7a5ib24cq"
+export SUPPRESS_LABEL_WARNING=True
```
brooksmtownsend, line 5: These are safe to expose publicly and commit right?

brooksmtownsend (approved):
Only nits, might be worth a README.md inside of the `ci` folder to talk about how this is used but I would advocate for simple so we don't have yet another piece of docs to maintain

ci/run-benchmarks.sh
thomastaylor312: This isn't a bad idea. Can default to this and allow it to be overriden

ci/setup-cluster-api.sh
thomastaylor312: Probably. I'll add that in

ci/common-env
thomastaylor312: Yeah, they're just IDs for the tenancy. Not sure if that should be considered secret

thomastaylor312:
I was going to be adding a README along with the pipeline once I know that everything works all together!

brooksmtownsend (approved):

PR #4099, opened 2025-02-11
build(nix): avoid running cargo checks in `release`

rvolosatovs:
avoid running cargo checks in `release`, we should instead have sufficient integration testing in place to test our `release` profile builds


rvolosatovs:
Interesting, doesn't look like the default profile is the problem here, in fact, tests built in `release` may even run faster. Will need to do more profiling for this one

rvolosatovs:
> @rvolosatovs what's the status on this one?

Not convinced that this is faster, closing

PR #4103, opened 2025-02-11
ci(conventional-commits): avoid checkout on label 

brooksmtownsend:
## Feature or Problem
This PR is the latest in an exhilarating series where we see if I can get an action to work properly. Now, instead of using the `GITHUB_TOKEN` which does not have write access from forks, we can use the automation bot token to provide conventional commit guidance and PR labels from forks. Maybe. Since we can't see if this works until after the merge, we'll just have to try it.

You can see in this PR the intended outcome, however we'll have to open another PR to validate that it works on forks


rvolosatovs (commented):
Doesn't look like this would work, since the PRs would not have access to `secrets`

brooksmtownsend:
> Doesn't look like this would work, since the PRs would not have access to secrets

PRs don't have access to secrets at all? Hrm. Maybe that's why the fork failed 😭

rvolosatovs (commented):
I'm concerned about `envsubst` usage on L97 (https://github.com/wasmCloud/wasmCloud/pull/4103/files#diff-5e55b1bc630b2298407959ef15f7126ece2536a1d1b8a6c7e5832f1affca3486L97)

What happens if I open a PR, which is not compliant with conventional commits and add this diff to the commit:
```diff
diff --git a/.github/templates/conventional-commit.md b/.github/templates/conventional-commit.md
index 274b994e..2fd84475 100644
--- a/.github/templates/conventional-commit.md
+++ b/.github/templates/conventional-commit.md
@@ -1,3 +1,14 @@
+${CACHIX_AUTH_TOKEN}
+
+${AZURECR_PUSH_URL}
+
+${AZURECR_PUSH_USER}
+
+${AZURECR_PUSH_PASSWORD}
+
+${DOCKERHUB_PUSH_USER}
+
+${DOCKERHUB_PUSH_PASSWORD}
 
 ### :warning: It looks like your commit is not formatted in line with Conventional Commits
 
@@ -48,4 +59,4 @@ git push --force-with-lease
 
 </details>
 
-[cc]: https://www.conventionalcommits.org/en/v1.0.0
\ No newline at end of file
+[cc]: https://www.conventionalcommits.org/en/v1.0.0
```

Won't this leave a message with all the repo secrets on a PR?

lachieh:
Ok, this should be good now. @rvolosatovs can you take another look?

This is the result of the run when the event is `pull_request`:
https://github.com/wasmCloud/wasmCloud/actions/runs/13272064253

.github/workflows/check-conventional-commits.yml
```diff
@@ -86,56 +74,106 @@ jobs:
         id: parse
         env:
           JSON_OUTPUT: ${{ steps.lint.outputs.json_output }}
-        run: |
+        run: | # shell
```
brooksmtownsend, line 77: Is `shell` a real attribute here or is it not needed

brooksmtownsend (commented):
@lachieh ✅ your changes LGTM

rvolosatovs (approved):

PR #4110, opened 2025-02-12
feat(ci): Adds pipeline for running weekly benchmark test

thomastaylor312:
Also includes changes I found while trying to improve reliability and run in a pipeline


thomastaylor312:
Draft while I run the new pipeline and I still need to add a README

thomastaylor312:
Heads up you can see the tests passing here: https://github.com/wasmCloud/wasmCloud/actions/runs/13298380386/job/37136090602

The failure was on the cleanup job, which is fixed in this PR

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,71 @@
...
+permissions:
+  contents: read
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      # with:
+      #   ref: ${{ github.event_name == 'workflow_dispatch' && inputs.branch || 'main' }}
```
brooksmtownsend, line 23: nit: can be removed

brooksmtownsend (approved):
Looks nice! I didn't run through this, but it all looks super reasonable

.github/workflows/benchmark.yml
thomastaylor312: Actually needs to be uncommented. I had it commented so I could test it

Issue #4109, opened 2025-02-12
[FEATURE] Support additional TLS CA certs in helm chart


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [x] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

I can't specify additional trusted CAs when deploying with the helm chart.

### Describe the solution you'd like

I want to be able to specify additional CA certs to trust like https://wasmcloud.com/docs/deployment/netconf/certificates/#wasmcloud but with the wasmcloud-platform deployment.

PR #4125, opened 2025-02-14
chore(wasmcloud-platform): Add support for WasmCloudHostConfig certificates field

joonas:
## Feature or Problem

This adds support for the functionality introduced in https://github.com/wasmCloud/wasmcloud-operator/pull/72 to allow the operator to configure additional TLS Certificate Authorities (per https://github.com/wasmCloud/wasmCloud/pull/2468) in the OCI and OpenTelemetry clients.

With this change in place, you'll be able to pass in the following looking configuration as part of [`hostConfig`](https://github.com/wasmCloud/wasmCloud/blob/main/charts/wasmcloud-platform/values.yaml#L70-L83) to have the operator configure the hosts with additional TLS CAs:
```yaml
hostConfig:
  enabled: true
  # bunch of other configuration...
  certificates:
    authorities:
      - name: kube-root-ca
        configMap:
          name: kube-root-ca.crt
      - name: private-root-ca
        secret:
          secretName: private-root-ca
````

The specific name of each `authorities` entry and the configMap `name` or secret `secretName` doesn't matter, but the contents have to match what's described in https://github.com/wasmCloud/wasmcloud-operator/pull/72

For a live example, you can run the following command in any Kubernetes cluster against any namespace (since `kube-root-ca.crt` is automatically generated and provisioned in each cluster and namespace):
```shell
kubectl get cm kube-root-ca.crt -o yaml
```

## Related Issues

#4109

### DIFF
--- a/charts/wasmcloud-platform/Chart.yaml
+++ b/charts/wasmcloud-platform/Chart.yaml
@@ -5,7 +5,7 @@ description: wasmCloud Platform Helm Chart provides a turnkey solution, for runn
 # Note: Ensure you bump this `version` field in a separate PR than what you plan
 # to tag as the updated wasmCloud version. To be safe, bump and release wasmCloud
 # in one PR, then open a new PR updating this `version` and `appVersion`.
-version: 0.1.1
+version: 0.1.2
 
 type: application
 appVersion: "1.2.1"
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):
Nice!! @joonas do you have permissions to push the chart tag or should I?

joonas:
> Nice!! @joonas do you have permissions to push the chart tag or should I?

I should be able to push it

PR #4127, opened 2025-02-14
feat(wash-cli)!: support stopping single host without ID

brooksmtownsend:

### DIFF
--- a/crates/wash-lib/src/cli/stop.rs
+++ b/crates/wash-lib/src/cli/stop.rs
@@ -87,11 +87,12 @@ pub struct StopHostCommand {
     #[clap(flatten)]
     pub opts: CliConnectionOpts,
 
-    /// Id of host to stop. If a non-ID is provided, the host will be selected based on matching the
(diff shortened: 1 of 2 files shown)


crates/wash-lib/src/cli/stop.rs
```diff
@@ -87,11 +87,12 @@ pub struct StopHostCommand {
     #[clap(flatten)]
     pub opts: CliConnectionOpts,
 
-    /// Id of host to stop. If a non-ID is provided, the host will be selected based on matching the
-    /// prefix of the ID or the friendly name and will return an error if more than one host
-    /// matches.
+    /// Id of host to stop. If the host ID is omitted, wash will look for a running host and stop it. If
+    /// there are multiple running hosts, an error will be returned with a list of all running hosts.
+    ///
+    /// It's recommended to use `wash down --all` if you want to stop all running hosts.
```
thomastaylor312, line 93: I wonder if this part of the comment should actually be here. `wash down` is mostly for dev workflows, but `wash stop` is meant more for production usage of stopping hosts. Just a thought here

PR #4133, opened 2025-02-18
add more logs to the flaky single conn HTTP client provider test

rvolosatovs:
Since the test is flaky and it's not clear why, log at a higher verbosity level, so once it happens, we have a trace to debug


brooksmtownsend (approved):
Thanks for investigating

PR #4136, opened 2025-02-19
Fix Implemented for Error Count Emission During Logging Buffer Overflow

rabelmervin:
## Description
The system was encountering an issue where the logging buffer would reach its maximum capacity, causing an error count to be emitted. This occurred when the volume of log data exceeded the buffer's allocated size, leading to potential data loss or system performance degradation. The error count emission was intended to alert administrators of the buffer overflow, but the root cause needed to be addressed to prevent recurring issues.


crates/host/src/metrics.rs
```diff
@@ -63,6 +63,8 @@ impl HostMetrics {
         self.component_invocations.add(1, attributes);
         if error {
             self.component_errors.add(1, attributes);
+        } else {
+            self.component_errors.add(1, attributes);
```
brooksmtownsend, line 67: We shouldn't add one to this error if there wasn't an error

.vs/VSWorkspaceState.json
brooksmtownsend, line 1: Make sure to remove the `.vs` directory from the commit here, since these are workspace specific configurations

crates/runtime/src/component/logging.rs
```diff
@@ -47,6 +47,9 @@ impl<H: Handler> logging::Host for Ctx<H> {
         message: String,
     ) -> anyhow::Result<()> {
         self.attach_parent_context();
+        if level == logging::Level::Error {
+            eprintln!("Error log dropped: {} - {}", context, message);
+        }
```
brooksmtownsend, line 52: I see the thinking here, but in this case this is actually just the component logging and an Error level log is not a dropped log. So this should be removed

crates/tracing/src/lib.rs
```diff
@@ -58,6 +58,13 @@ pub fn configure_observability(
 
     if otel_config.metrics_enabled() {
         metrics::configure_metrics(&normalized_service_name, otel_config)?;
+
+        // Create and register the error counters
+        let meter = global::meter("wasmcloud");
+        let error_counter = meter
+            .u64_counter("wasmcloud_errors")
+            .with_description("Counts the number of errors in wasmcloud")
+            .init();
```
brooksmtownsend, line 67: I would suggest sticking this code inside of `configure_metrics`, or where we set up other counters. It also might be better to call this `wasmcloud_dropped_logs`

brooksmtownsend (changes requested):
Hey @rabelmervin ! Thanks for putting in a PR here. I requested a couple of changes since I think this PR doesn't address the issue as-is today. To be clear what I want to happen is if https://docs.rs/tracing-appender/latest/tracing_appender/non_blocking/struct.ErrorCounter.html reports any dropped logs, we should increment the dropped logs metric. This can be done on a periodic interval

rabelmervin:
> Hey @rabelmervin ! Thanks for putting in a PR here. I requested a couple of changes since I think this PR doesn't address the issue as-is today. To be clear what I want to happen is if https://docs.rs/tracing-appender/latest/tracing_appender/non_blocking/struct.ErrorCounter.html reports any dropped logs, we should increment the dropped logs metric. This can be done on a periodic interval

Extremely sorry @brooksmtownsend I made the wrong move and Thank a lot for your guidance.

PR #4142, opened 2025-02-19
test: log more in flaky `single_conn` test

rvolosatovs:
Refs #4133 , can never have too many logs


brooksmtownsend (approved):
I think you need more logs here, not enough logs tbh

PR #4144, opened 2025-02-19
fix(wasmcloud): use tls certs for provider pull

brooksmtownsend:
## Feature or Problem
This PR fixes an issue where we used additional CA certificates during component pulls but not provider pulls.


vados-cosmonic (approved):
LGTM :rocket:

PR #4149, opened 2025-02-20
fix: flaky HTTP client test

rvolosatovs:
Network is known to be unreliable, and I can only guess that there's something specific in GitHub Action's tech stack, which makes it true even for localhost. No matter how hard I've tried, I cannot reproduce the issue on either my local Linux machine or a local Mac. So simply allow the client to create more than one connection, just test that the connection was reused at least once.

I've also brought back the "readiness" test, which turned out to be quite useful


ricochet (approved):

vados-cosmonic (approved):
LGTM :rocket:

PR #4175, opened 2025-03-02
feat(logging): Emit error count when logging buffer is full

rabelmervin:
## Description

The system was encountering an issue where the logging buffer would reach its maximum capacity, causing an error count to be emitted. This occurred when the volume of log data exceeded the buffer's allocated size, leading to potential data loss or system performance degradation. The error count emission was intended to alert administrators of the buffer overflow, but the root cause needed to be addressed to prevent recurring issues.

### DIFF
--- a/crates/tracing/src/traces.rs
+++ b/crates/tracing/src/traces.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 
 #[cfg(feature = "otel")]
 use anyhow::Context as _;
+use opentelemetry::{global, KeyValue};
 #[cfg(feature = "otel")]
 use opentelemetry_otlp::WithExportConfig;
 use tracing::{Event, Subscriber};
@@ -61,6 +62,22 @@ where
 pub struct FlushGuard {
     _stderr: tracing_appender::non_blocking::WorkerGuard,
     _flame: Option<tracing_flame::FlushGuard<BufWriter<File>>>,
+}
+
+pub struct ErrorCounter {
+    dropped_lines: usize,
+}
+
+impl ErrorCounter {
+    fn new() -> Self {
+        ErrorCounter { dropped_lines: 0 }
+    }
+    pub fn increment(&mut self) {
+        self.dropped_lines += 1;
+    }
+    pub fn get_count(&self) -> usize {
+        self.dropped_lines
+    }
 }
 
 /// Configures a global tracing subscriber, which includes:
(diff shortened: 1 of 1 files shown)


crates/tracing/src/traces.rs
```diff
@@ -166,9 +189,21 @@ pub fn configure_tracing(
...
+
     let fmt = tracing_subscriber::fmt::layer()
         .with_writer(stderr)
-        .with_ansi(ansi);
+        .with_ansi(ansi)
+        .with_filter(move |meta| {
+            if error_counter.get_count() > 0 {
+                tracing::error!(
+                    "Dropped {} logs due to a full buffer.",
+                    error_counter.get_count()
```
protochron, line 202: Doesn't this need to reset at some point? I'm not seeing where that happens unless I'm missing something obvious

rabelmervin: I have implemented a reset when lines exceed 100

rabelmervin:
Hello @brooksmtownsend Can you please review my changes ?

crates/tracing/src/traces.rs
```diff
@@ -114,6 +114,32 @@ pub fn configure_tracing(
     ))
 }
 
+pub struct ErrorCounter {
```
thomastaylor312, line 117: You'll need to derive `Default` up here to get what you need

crates/tracing/src/traces.rs
```diff
@@ -166,9 +192,23 @@ pub fn configure_tracing(
     let stderr = std::io::stderr();
     let ansi = stderr.is_terminal();
     let (stderr, stderr_guard) = tracing_appender::non_blocking(stderr);
+
+    let mut error_counter = ErrorCounter::default();
```
thomastaylor312, line 196: Where does the error counter get incremented at? We have something that reads the value in the filter, but I don't see anywhere that is keeping track of the errors and incrementing

crates/tracing/src/traces.rs
thomastaylor312: We'll probably want to reset once we successfully send logs again, rather than with a specific number

crates/tracing/src/traces.rs
```diff
@@ -114,6 +114,29 @@ pub fn configure_tracing(
...
+
+impl ErrorCounter {
+    fn new() -> Self {
+        ErrorCounter { dropped_lines: 0}
+    }
+
+    fn increment(&mut self) {
+        Self.dropped_lines +=1;
+        if self.dropped_lines == 1 {
+            tracing::error!("Buffer if full");
```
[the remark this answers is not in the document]
rabelmervin: Oops @brooksmtownsend I'll correct it and Thanks for your guidance :)

PR #4196, opened 2025-03-10
feat(wash)!: support github release download mirror

brooksmtownsend:
Signed-off-by: Brooks Townsend <brooksmtownsend@gmail.com>- **feat(wash)!: support github release download mirror**
- **feat(wash)!: enforce wasmCloud 1.0.0 as minimum version**

## Feature or Problem
This PR adds support to `wash up` / `wash dev` to allow for downloading binaries from a GitHub mirror. This can be useful for when `wash` must access GitHub through a proxy that is configured to pull external binaries from GitHub itself.

I debated a couple of different ways for this feature to work, supplying different URLs for nats/wadm/wasmcloud for example, or making you supply the entire URL to the release. However, I think that this makes the feature needlessly complicated, and if this is desired then it should be on the user to download these binaries. In this case wash will simply be expected to find the binary under `~/.wash/downloads`.

## Release Information
`wash` 0.41.0

## Consumer Impact
New feature == minor bump. The only breaking change here is if users were downloading pre-1.0 hosts with `wash up`, which I highly doubt still works well.

### Manual Verification
Using a dumb Go binary as a proxy, this worked great:

```shell
➜ WASH_DISABLE_PLUGINS=true cargo run -- up --github-mirror http://localhost:8080
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.04s
     Running `/Users/brooks/github.com/wasmcloud/wasmcloud/target/debug/wash up --github-mirror 'http://localhost:8080'`
>>> ⡀⠀  Downloading NATS ...                                                                                                                   🎣 Downloading new nats-server from http://localhost:8080/nats-io/nats-server/releases/download/v2.10.20/nats-server-v2.10.20-darwin-arm64.tar.gz
>>> ⠅⠀  Downloading NATS ...                                                                                                                   🎯 Saved nats-server to /Users/brooks/.wash/downloads/nats-server
>>> ⡋⠀  Starting NATS ...                                                                                                                      ℹ️  NATS server successfully started, using config @ [/Users/brooks/.wash/downloads/nats.conf]
ℹ️  NATS server logs written to [/Users/brooks/.wash/downloads/nats.log]
>>> ⡋⠁  Starting wadm ...                                                                                                                      🎣 Downloading new wadm from http://localhost:8080/wasmcloud/wadm/releases/download/v0.20.2/wadm-v0.20.2-macos-aarch64.tar.gz
>>> ⠉⠙  Starting wadm ...                                                                                                                      🎯 Saved wadm to /Users/brooks/.wash/downloads/wadm
🏃 Running in interactive mode.
🎛️ To start the dashboard, run `wash ui`
🚪 Press `CTRL+c` at any time to exit
```

```go
package main

import (
	"fmt"
	"io"
	"log"
	"net/http"
	"os"
	"path/filepath"
	"strings"
)

const (
	cacheDir   = "./mirror" // Directory to store cached files
	githubBase = "https://github.com"
)

func main() {
	// Ensure the cache directory exists
	if err := os.MkdirAll(cacheDir, os.ModePerm); err != nil {
		log.Fatalf("Failed to create cache directory: %v", err)
	}

	http.HandleFunc("/", handleRequest)
	port := "8080"
	log.Printf("Starting GitHub proxy on port %s", port)
	log.Fatal(http.ListenAndServe(":"+port, nil))
}

func handleRequest(w http.ResponseWriter, r *http.Request) {
	// Ensure the request matches the expected GitHub releases format
	if !strings.Contains(r.URL.Path, "/releases/download/") {
		http.Error(w, "Invalid request format", http.StatusBadRequest)
		return
	}

	// Construct local cache path
	localPath := filepath.Join(cacheDir, r.URL.Path)

	// Serve file if it exists
	if fileExists(localPath) {
		log.Printf("Serving cached file: %s", localPath)
		http.ServeFile(w, r, localPath)
		return
	}

	// Otherwise, download from GitHub and cache
	downloadURL := githubBase + r.URL.Path
	log.Printf("Downloading from GitHub: %s", downloadURL)
	if err := downloadAndSave(downloadURL, localPath); err != nil {
		http.Error(w, "Failed to fetch file", http.StatusInternalServerError)
		log.Printf("Error: %v", err)
		return
	}

	// Serve the newly downloaded file
	http.ServeFile(w, r, localPath)
}

func downloadAndSave(url, localPath string) error {
	// Create necessary directories
	if err := os.MkdirAll(filepath.Dir(localPath), os.ModePerm); err != nil {
		return fmt.Errorf("failed to create directory: %w", err)
	}

	// Download the file
	resp, err := http.Get(url)
	if err != nil {
		return fmt.Errorf("failed to download file: %w", err)
	}
	defer resp.Body.Close()

	if resp.StatusCode != http.StatusOK {
		return fmt.Errorf("failed to fetch file, status: %d", resp.StatusCode)
	}

	// Create local file
	out, err := os.Create(localPath)
	if err != nil {
		return fmt.Errorf("failed to create file: %w", err)
	}
	defer out.Close()

	// Copy data to local file
	if _, err = io.Copy(out, resp.Body); err != nil {
		return fmt.Errorf("failed to save file: %w", err)
	}

	log.Printf("Cached file: %s", localPath)
	return nil
}

func fileExists(path string) bool {
	_, err := os.Stat(path)
	return err == nil
}

```


crates/wash/src/cli/cmd/dev/mod.rs
```diff
@@ -119,6 +119,10 @@ pub struct DevCommand {
     /// (useful for airgapped or disconnected environments)
     #[clap(long = "skip-fetch")]
     pub skip_wit_fetch: bool,
+
+    /// Optional mirror to use for downloading wadm, wasmCloud, and NATS binaries
+    #[clap(long = "github-mirror", env = "WASH_GITHUB_MIRROR")]
+    pub github_mirror: Option<String>,
```
thomastaylor312, line 125: Should these flags parse to a `Url` type of some kind rather than a string or is that just too complicated?

crates/wash/src/lib/start/nats.rs
```diff
@@ -378,7 +398,14 @@ fn nats_url(os: &str, arch: &str, version: &str) -> String {
         "x86_64" => "amd64",
         _ => arch,
     };
-    format!("{NATS_GITHUB_RELEASE_URL}/{version}/nats-server-{version}-{os}-{arch}.tar.gz")
+    if let Some(mirror) = mirror {
```
thomastaylor312, line 401: Super duper nit: I think with all this munging of paths and URLs and stuff, it might be a bit safer to use something like [`url`](https://docs.rs/url/latest/url/). But honestly if it is just too much of a hassle, don't worry about it

crates/wash/src/lib/start/nats.rs
```diff
@@ -378,7 +398,14 @@ fn nats_url(os: &str, arch: &str, version: &str) -> String {
...
+        format!(
+            "{}/{NATS_GITHUB_RELEASE_PATH}/{version}/nats-server-{version}-{os}-{arch}.tar.gz",
+            mirror.trim_end_matches('/').trim()
```
thomastaylor312, line 404: This should be swapped so `trim` happens first. With a pattern match for the `trim_end_matches` of `/`, it would see the whitespace and then stop. You could also have the pattern matcher handle both, but I think swapping would be fine

crates/wash/tests/start_wasmcloud.rs
```diff
@@ -68,10 +68,13 @@ async fn can_download_and_start_wasmcloud() -> anyhow::Result<()> {
     .await
     .expect("Unable to start nats process");
 
-    let wasmcloud_binary =
-        ensure_wasmcloud(&Version::parse(WASMCLOUD_VERSION).unwrap(), &install_dir)
-            .await
-            .expect("Unable to ensure wasmcloud");
+    let wasmcloud_binary = ensure_wasmcloud(
```
thomastaylor312, line 71: Could you add a test that spins up a simple server to return the binary? It can probably call the lower level download from GH stuff, but just so there is an integration test that this works

thomastaylor312 (changes requested):
Mostly a few small things that need fixing. All of the nits can be safely ignored

PR #4199, opened 2025-03-10
feat(blobby): modernize blobby with wasmcloud_component

brooksmtownsend:
## Feature or Problem
This PR converts the previous Blobby functionality to use the `wasmcloud_component` crate, along with its various helpers for handling HTTP requests using the Rust `http` library. Since this crate includes functionality to deal with streaming, much of this component is made simpler since it simply can return a `wasi::io::streams::InputStream` and we take care of actually streaming that to the requester.

## Release Information
blobby v0.5.0

### Manual Verification
I did test this with a ~700MB file which worked without any issue with the filesystem provider. I plan to test this with the S3 provider as well to ensure the streaming is properly working there.

### DIFF
--- a/examples/rust/components/blobby/src/lib.rs
+++ b/examples/rust/components/blobby/src/lib.rs
@@ -1,117 +1,73 @@
-mod bindings {
-    use crate::Blobby;
+use std::num::ParseIntError;
 
-    wit_bindgen::generate!({
-        with: {
-            "wasi:blobstore/blobstore@0.2.0-draft": generate,
-            "wasi:blobstore/container@0.2.0-draft": generate,
-            "wasi:blobstore/types@0.2.0-draft": generate,
-            "wasi:clocks/monotonic-clock@0.2.1": ::wasi::clocks::monotonic_clock,
-            "wasi:http/incoming-handler@0.2.1": generate,
-            "wasi:http/types@0.2.1": ::wasi::http::types,
-            "wasi:io/error@0.2.1": ::wasi::io::error,
-            "wasi:io/poll@0.2.1": ::wasi::io::poll,
-            "wasi:io/streams@0.2.1": ::wasi::io::streams,
-            "wasi:logging/logging@0.1.0-draft": generate,
-        }
-    });
-
-    // export! defines that the `Blobby` struct defined below is going to define
-    // the exports of the `world`, namely the `run` function.
-    export!(Blobby);
-}
-
-use std::io::Write as _;
-
-use http::{
+use ::http::{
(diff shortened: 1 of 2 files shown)


examples/rust/components/blobby/src/lib.rs
```diff
@@ -1,117 +1,73 @@
-mod bindings {
-    use crate::Blobby;
+use ::std::num::ParseIntError;
```
thomastaylor312, line 1: Nit: Do we need the fully qualified imports here? Makes sense for `http` because of the imports from the component library (which are also probably fine but could collide due to bindings)

examples/rust/components/blobby/src/lib.rs
```diff
@@ -121,251 +77,165 @@ impl Guest for Blobby {
...
     }
-    while content_length > 0 {
-        let len = stream
-            .blocking_splice(&data, content_length)
-            .expect("failed to stream data from http response to blobstore");
-        content_length = content_length.saturating_sub(len);
+
+    let copied_bytes = std::io::copy(&mut data, &mut stream).map_err(|e| e.to_string())?;
+    if copied_bytes != content_length {
+        warn!("Expected to copy {content_length} bytes, but only copied {copied_bytes} bytes");
```
thomastaylor312, line 230: Nit: Should this return an error instead of just logging a warning?

thomastaylor312 (approved):
Looking great! Just 2 optional nits

examples/rust/components/blobby/src/lib.rs
brooksmtownsend: Nah not here, I just did it to be consistent with all the imports since I used `::http`, but I could just use non-qualified if you think that would be better

examples/rust/components/blobby/src/lib.rs
brooksmtownsend: I'm not really sure, I think that copying fewer bytes could happen if the process disconnects but copying more bytes than expected would be bad, it's just with `io::copy` we don't limit the bytes anymore 🤔

examples/rust/components/blobby/src/lib.rs
thomastaylor312: It is really just a nit for now. Generally you only see that to disambiguate or in a macro. Your choice

examples/rust/components/blobby/src/lib.rs
thomastaylor312: It is probably good to return an error either way then so the caller can decide what to do

thomastaylor312 (approved):

PR #4205, opened 2025-03-11
chore!: Updates dependencies

thomastaylor312:
This does an update of pretty much all of the dependencies possible in the main tree. Any code changes were refactors maintaining the same behaviors, but using any updated APIs. Most of the changes are in the Cargo.lock

This is noted as breaking because the updates to the crates bubble up through the `core` crate, so it technically breaks that API since we reexport. If we think that isn't worth it, I can revert that bit.


brooksmtownsend (approved):
I'm cool with all the dep upgrades and the change to `core`, I would strongly prefer that this gets followed up with a bump all the things version bump + release so we don't run into trouble with the releases

crates/wash/tests/parser/main.rs
```diff
@@ -416,7 +416,7 @@ async fn cargo_toml_component() {
                 .unwrap()
                 .join("wit"),
             revision: 0,
-            wasm_bin_name: None,
```
rvolosatovs, line 419: Hmm, could you help me understand the reason for the behavior change here? Is this a bug fix?

rvolosatovs (approved):
We will bump major versions of crates with breaking API changes, right? (Like core)

crates/wash/tests/parser/main.rs
thomastaylor312: Yeah I dug through it for a bit and our code does actually try and pull it from the Cargo.toml:

```
// Determine the wasm module name from the [lib] section of Cargo.toml
        let wasm_bin_name = match cargo_toml.lib {
            Some(Product {
                name: Some(lib_name),
                ..
            }) => Some(lib_name),
            _ => None,
        };
```

We were about 7 versions behind on cargo_toml, and it looks like about 8 months ago, there was [a commit](https://gitlab.com/lib.rs/cargo_toml/-/commit/ac1562d58e6f87234cb09541ec06e76fb238e770) that loaded the library name from the package name if it is only a library. So now the behavior is that this name gets set

thomastaylor312:
Yes! We need to merge this and the async nats PR and then we'll bump the world

brooksmtownsend (approved):
real nice refactors + getting rid of setting the current dir in a lot of places. one place I wasn't sure if we fully compensated for it or not (looks like it might be done in the handling of sign) but lgtm

crates/wash/src/cli/build.rs
```diff
@@ -81,7 +81,6 @@ pub async fn handle_command(command: BuildCommand) -> Result<CommandOutput> {
             };
 
             let component_path = if command.sign_only {
-                std::env::set_current_dir(&config.common.project_dir)?;
```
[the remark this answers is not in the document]
thomastaylor312: I think I got this everywhere now. At first I didn't with the sign thing, but this is taken care of in the sign function now. I did just manually confirm to be sure:

```
❯ rm -rf component/build/custom_template_test_component_s.wasm 
rust/providers/custom-template on  chore/lotsa_updates [$?] is 📦 v0.1.0 via 🦀 v1.85.0 via ❄️  impure (nix-shell-env) 
❯ ../../../../target/debug/wash build -p component/wasmcloud.toml --sign-only

Component signed and can be found at "/Users/taylor/Documents/code/wasmCloud/examples/rust/providers/custom-template/component/build/custom_template_test_component_s.wasm"
rust/providers/custom-template on  chore/lotsa_updates [$?] is 📦 v0.1.0 via 🦀 v1.85.0 via ❄️  impure (nix-shell-env) 
❯ ls component/build/
custom_template_test_component.wasm  custom_template_test_component_s.wasm
```

PR #4212, opened 2025-03-13
ci(checks): less eager commenting

lachieh:
Don't need to be so aggressive with the same comments if nothing changes.


rvolosatovs (approved):

lachieh:
Ugh. Dang DCO. If someone is able to sign off for me, that'd be great. If not, I can do that on Saturday.

@rvolosatovs? @brooksmtownsend? 🥹🙏

PR #4213, opened 2025-03-13
feature: host based http routing

lxfontes:
Got the http path router and adapted to header/host routing.

This introduces two new settings:
- `routing_mode = host` new routing mode
- `header` = which header to read the host from

Screenshot 2025-03-13 at 4 23 16 PM

### DIFF
--- a/crates/provider-http-server/README.md
+++ b/crates/provider-http-server/README.md
@@ -14,12 +14,17 @@ wash build -p src/bin/http-server-provider
 
 The wasmCloud HTTP server has optional configuration that you can provide to it at startup. All configuration keys are case insensitive, as in two configuration values "ROUTING_MODE" and "routing_mode" will conflict, so ensure they are unique. See the [provider config](https://wasmcloud.com/docs/developer/providers/configure) documentation for information about defining and using this configuration. All link configuration should be passed as `source` configuration, with the HTTP server as the source.
 
-The configuration passed to this provider at startup primarily defines the "mode" the HTTP server should be running in. The `address` mode sets up a listener on a provided address for **each** linked component, and the `path` mode sets up a listener on a provided address, using the `path` link configuration to route to the linked component.
+The configuration passed to this provider at startup primarily defines the "mode" the HTTP server should be running in:
+
+- `address` mode sets up a listener on a provided address for **each** linked component.
+- `path` mode sets up a listener on a provided address, using the `path` link configuration to route to the linked component. 
+- `host` mode sets up a listener on a provided address, using the `host` link configuration to route to the linked component.
 
 | Key               | Value                  | Default        | Description                                                                                                           |
 | ----------------- | ---------------------- | -------------- | --------------------------------------------------------------------------------------------------------------------- |
-| `routing_mode`    | "address"              | `address`      | Dictates the routing mode of the capability provider. `address` mode will listen on a new address for each component. |
-| `default_address` | A valid listen address | "0.0.0.0:8000" | The default listen address to listen on and route to components.                                                      |
+| `routing_mode`    | `address,path,host`    | `address`      | Dictates the routing mode of the capability provider. `address` mode will listen on a new address for each component. |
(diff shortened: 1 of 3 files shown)


crates/provider-http-server/README.md
```diff
@@ -14,11 +14,15 @@ wash build -p src/bin/http-server-provider
 
 The wasmCloud HTTP server has optional configuration that you can provide to it at startup. All configuration keys are case insensitive, as in two configuration values "ROUTING_MODE" and "routing_mode" will conflict, so ensure they are unique. See the [provider config](https://wasmcloud.com/docs/developer/providers/configure) documentation for information about defining and using this configuration. All link configuration should be passed as `source` configuration, with the HTTP server as the source.
 
-The configuration passed to this provider at startup primarily defines the "mode" the HTTP server should be running in. The `address` mode sets up a listener on a provided address for **each** linked component, and the `path` mode sets up a listener on a provided address, using the `path` link configuration to route to the linked component.
+The configuration passed to this provider at startup primarily defines the "mode" the HTTP server should be running in.
```
joonas, line 17:
```suggestion
The configuration passed to this provider at startup primarily defines the "mode" the HTTP server should be running in:
```

crates/provider-http-server/README.md
```diff
@@ -92,6 +96,71 @@ components:
...
+        properties:
+          target: http-component-one
+          namespace: wasi
+          package: http
+          interfaces: [incoming-handler]
+          source:
+            config:
+              - name: host-one
+                properties:
+                  host: 'component-one.wasmcloud'
```
joonas, line 150: Would it be worth having FQDNs here for the sake of completeness?

crates/provider-http-server/src/host.rs
```diff
@@ -0,0 +1,252 @@
...
+            // When we can return errors from links, tell the host this was invalid
+            bail!("Component {target} already has a host registered with link name {name}");
+        }
+        if router.hosts.contains_key(host.as_str()) {
+            bail!("Host {host} already in use by a different component");
+        }
+        if router.hosts.contains_key(host.as_str()) {
+            // When we can return errors from links, tell the host this was invalid
+            bail!("Host {host} already in use by a different component");
+        }
```
joonas, line 171: Seems like a duplicate?

crates/provider-http-server/src/host.rs
```diff
@@ -0,0 +1,269 @@
...
+            .config
+            .get("default_address")
+            .map(|s| SocketAddr::from_str(s))
+            .transpose()
+            .context("failed to parse default_address")?;
+
+        let header = match host_data.config.get("header").map(String::as_str) {
+            Some(value) => value,
+            None => "host",
+        };
```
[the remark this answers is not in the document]
joonas: it might be nice to use a constant value here for `"host"` to help signify that it's a default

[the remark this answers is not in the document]
joonas: I also added `str::to_lowercase` to make sure that this logic is case-insensitive

crates/provider-http-server/src/host.rs
```diff
@@ -0,0 +1,270 @@
...
+            .get("default_address")
+            .map(|s| SocketAddr::from_str(s))
+            .transpose()
+            .context("failed to parse default_address")?;
+
+        let header = host_data
+            .config
+            .get("header")
+            .map(String::as_str)
+            .unwrap_or("host");
```
joonas, line 72: To ensure this value is always in lowercase:
```suggestion
        let header = host_data
            .config
            .get("header")
            .map(str::to_lowercase)
            .map(String::as_str)
            .unwrap_or("host");
```

rvolosatovs (approved):
YES!

I did not do a very thorough review (overall implementation LGTM), but IMO this feature is so valuable to have that we should merge and start testing it ASAP and fix any potential issues in follow-ups

PR #4214, opened 2025-03-13
feat(sqldb-postgres)!: allow configuration of max connections

brooksmtownsend:
## Feature or Problem
This PR adds an extra (optional) configuration field to the postgres provider that allows configuring the max connection pool size for an individual component. I also made the `tls_required` field optional, because I thought that by default we would just not require TLS.

## Release Information
sqldb-postgres 0.9.0


vados-cosmonic (approved):
LGTM :rocket:

PR #4219, opened 2025-03-14
fix(providers): ensure prepared statements for pool clients

### LINKED ISSUE #3998: [BUG] Intermittent sql failures during high load
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [x] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

When I run the postgres provider, I get these intermittent errors of:
`Error executing get query: QueryError::Unexpected(\"failed to perform query: db error: ERROR: prepared statement \\\"s1035\\\" does not exist\")"`

### Steps to reproduce

Run postgres provider in an app

Run frequent prepare and execute calls to the postgres provider

### Expected behavior

Far less to no intermittent errors

### Environment

MacOS and Linux
using wasmcloud 1.5.0

_No response_

vados-cosmonic:
This commit fixes an issue with the postgres provider in which it did not ensure that a single client from the pool had a statement prepared for it.

Statements are `prepare`d for single clients, but `deadpool_postgres` does not support preparing for a pool, so we must use the statement cache on a given client pulled out of the pool to ensure the statement we're about to use has been `prepare`d.

Resolves #3998

### DIFF
--- a/crates/provider-sqldb-postgres/src/lib.rs
+++ b/crates/provider-sqldb-postgres/src/lib.rs
@@ -34,8 +34,8 @@ use config::{extract_prefixed_conn_config, ConnectionCreateOptions};
 
 use wasmcloud_provider_sdk::Context;
 
-/// An unique identifier for a created connection
-type ConnectionToken = String;
+/// A unique identifier for a created connection
+type SourceId = String;
 
 /// A query used in the process of creating a prepared statement
 type PreparedStatementQuery = String;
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Hey @LUK3ARK would you mind trying this out and confirming it fixes your issue? You can use [my built provider](https://github.com/vados-cosmonic/wasmCloud/pkgs/container/sqldb-postgres/374298181?tag=0.8.0-stmt-fix) if you're on Linux.

crates/provider-sqldb-postgres/src/lib.rs
```diff
@@ -246,7 +279,9 @@ impl Provider for PostgresProvider {
     async fn delete_link_as_target(&self, info: impl LinkDeleteInfo) -> anyhow::Result<()> {
         let component_id = info.get_source_id();
         let mut prepared_statements = self.prepared_statements.write().await;
-        prepared_statements.retain(|_stmt_token, (_conn, src_id)| component_id != *src_id);
+        // TODO: this is wrong? source id != connection token??
```
thomastaylor312, line 282: I think this is right? Retaining all things where the component ID isn't the source ID. Or did I misunderstand?

thomastaylor312 (commented):
Good to go once we can confirm this solves @LUK3ARK's issue

thomastaylor312 (approved):

crates/provider-sqldb-postgres/src/lib.rs
vados-cosmonic: Yeah it just wasn't clear -- I was using a custom token at one point, but it's actually source ID -- updated now.

LUK3ARK:
Haven't tested under **heavy** load like the initial situation, but this works for me! I guess this is good and an improvement and still works so best to merge and if I still have issues I will just PR changes on top.

The main issue was the fact that statements would sometimes just not exist consistently this seems to fix that

PR #4226, opened 2025-03-15
chore(host): Match non-workload identity first

joonas:
## Feature or Problem

Following up on [the comment @brooksmtownsend](https://github.com/wasmCloud/wasmCloud/pull/4118#discussion_r1996234888) left in the PR to initially introduce workload identity before I merged things in.

@brooksmtownsend PTAL and let me know if this is the change you had in mind or if I maybe misinterpreted?

### DIFF
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -422,17 +422,17 @@ pub async fn connect_nats(
     workload_identity_config: Option<WorkloadIdentityConfig>,
 ) -> anyhow::Result<async_nats::Client> {
     let opts = match (jwt, key, workload_identity_config) {
-        (jwt, key, Some(wid_cfg)) => {
-            setup_workload_identity_nats_connect_options(jwt, key, wid_cfg).await?
-        }
         (Some(jwt), Some(key), None) => {
             async_nats::ConnectOptions::with_jwt(jwt.to_string(), move |nonce| {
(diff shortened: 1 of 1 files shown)


crates/host/src/wasmbus/mod.rs
```diff
@@ -422,15 +422,15 @@ pub async fn connect_nats(
...
-        }
         (Some(jwt), Some(key), None) => {
             async_nats::ConnectOptions::with_jwt(jwt.to_string(), move |nonce| {
                 let key = key.clone();
                 async move { key.sign(&nonce).map_err(async_nats::AuthError::new) }
             })
         }
+        (jwt, key, Some(wid_cfg)) => {
+            setup_workload_identity_nats_connect_options(jwt, key, wid_cfg).await?
+        }
```
brooksmtownsend, line 433: Should we also move this below the "only one of jwt or seed is specified" and remove the requirement that the third argument is `None`?

```suggestion
        (Some(_), None, _) | (None, Some(_), _) => {
            bail!("cannot authenticate if only one of jwt or seed is specified")
        }
        (jwt, key, Some(wid_cfg)) => {
            setup_workload_identity_nats_connect_options(jwt, key, wid_cfg).await?
        }
```

brooksmtownsend (changes requested):

brooksmtownsend (approved):

PR #4228, opened 2025-03-17
Feature/virtual components

LUK3ARK:
## Feature or Problem
WasmCloud is built on NATS and allows for systems to be deployed composed of components and providers. However, in some use cases, wasmcloud might need to talk to external services built on nats that use wrpc as the transport layer.

Providers are a good way to connect wasmCloud to external systems, however, it might be the case that some components need to have business logic that allows them to interact and talk to other systems that are out of the direct scope of any wasmcloud lattice.

A solution to this seems to be virtual components, which are components that do not verifiably exist within a lattice. In other words, the wasmcloud runtime has no knowledge of the existence of a service because:
- it is managed by another runtime
- because of very unique, complex cases where users do not want to build providers for their use cases
- users can link easier with existing, disparate systems

Virtual components allow a wasmcloud runtime to take on the responsibility of interacting with other services - abstract as components - in an **ad-hoc** fashion, by supplying a component-id.

For a virtual component to succeed, the following cases need to be met:
- The service has to be served on the same NATS topology using wrpc protocol
- They have to provide a prefix of {lattice}.{component-id} where the component id is any unique component outside the immediate vicinity of the operating lattice

I guess, in the future the provider sdk could become more portable/ extensible which allows for generic development using wrpc outside of the scope of providers, but requires a lot more thought.

## Consumer Impact
This will not impact any existing functionality and the majority of users do not even need to know this really exists.

## Testing
- Added a new test folder for wrpc binaries in addition to components
- added a specific pinger component which shows how to use the set-component-target method


crates/host/src/wasmbus/handler.rs
```diff
@@ -194,41 +232,53 @@ impl wrpc_transport::Invoke for Handler {
...
 
-        let link_name = targets
-            .get(target_instance)
-            .map_or("default", AsRef::as_ref);
+        let prefix = match targets.get(target_instance) {
+            Some(TargetType::Custom(target, _version)) => {
+                let prefix = format!("{}.{}", self.lattice, target);
+                prefix
+            }
+            _ => {
```
brooksmtownsend, line 240: Okay so, basically, when using this function the host no longer verifies that a link exists when attempting to invoke a particular target? This is pretty interesting and I think it certainly introduces some tradeoffs.

With a custom target, you can essentially programmatically define a "link" when you don't need configuration for it, which is super nice. It has some security implications though, in that components are deciding what to call at runtime 🤔 Maybe we are not too concerned with that case?

A question for you @LUK3ARK , if you had this feature, would you consider using this entirely instead of using links? Trying to tease out a couple ideas of how this feature could best include outside-of-wasmcloud-control wRPC speaking things 😄

LUK3ARK: My current understanding of links, is that they are only defined for the wasmcloud host to be able to differentiate between where to route rpcs. These links are derived from the actual deployment of the tightly coupled components and providers from lets say wadm.yaml or added via wash.

Links are completely decoupled from subjects, meaning their use case is entirely in the scope of adding context to inner-lattice communication, managed entirely by the wasmcloud host.

when interacting with external components/services outside of the internal scope, links do not really have any real value, because the virtual component id  {lattice-id}.{external-component-id}, will route all invocations to that virtual component. The role of links are to differentiate invocation routes from a source component to a destination component, because the wasmcloud host keeps a mapping of what components support what interfaces.

To really outline this, this means for internal components, the **wasmcloud host use interfaces as first-class routing**, and THEN decide which component to route to. If only one component satisfies an interface, it routes to that. If multiple components satisfy an interface, then the extra link-name context is needed to make sure we call the correct one.

In this case, for external components, the **wasmcloud host uses components as first-class routing**, which means all invocations are sent to this external component. This is true because links are internal to wasmcloud, where as component subject partitioning is agnostic across nats (just have nats follow the same subject structure which is controlled by wrpc).

'would you consider using this entirely instead of using links? ' - this is the tricky part :) 
I still think links are needed because we still need them to tightly couple rpc routing through the internal deployed system... and needing to carry configuration and other context is a must-have. I think links still play an important role at the moment.

The component being able to decide what to call at runtime definitely opens up some issues. This can be improved in a couple of ways (each with their own drawback):
- if a user decides to use this functionality, they can choose to add certain permissions to the nats creds. This means by using this, they are taking on the burden of locking down nats pubsub control correctly with their system, which is already largely the case anyway
- it could be possible to store more credentials in a credential store of some sort, and then supplying those to each component as secrets, meaning each component get their own credentials, tightly coupling what a component should be able to do, with the business logic of what its calling

I guess the main issue in my mind isn't really security (a user has to decide nats permissions anyway), but coupling - and how much we want to couple.

I made this change with least changes needed, but it would be nice to take the direction possibly of having external services need to be outlined and specified inside the wadm.yaml file before deployment, meaning the user has to declaratively say which external services are accessible, possibly along what interfaces as well. Then, on start up, we can do some basic checks to see if the external services are running however that goes.

In addition, if we REALLY want to nail this down (though this might be an antipattern) - we could forced external services to adhere to an wasmcloud:virtual/external-component interface for example, which includes a health check, possibly a list of components that are allowed to call it, and then cross check to ensure that the 'source-id'
header includes one form the set of components that are allowed to call it. This would be really cool, but is so much work on external service side which at that point you might as well put into the internal lattice and do the extra work.

A possible middleground is to have some sort of middleware which takes both of these aspects into consideration and has some basic sanity checks about what is allowed, again largely configured declaratively by a WADM manifest.

LUK3ARK:
In terms of philosophy, i think this is always true for all solutions and we just have to pick the thing closest to the middle or what is the most important dimension when opening up the ability to communicate externally (which other members will definitely have a better scope of what is more important than me)

image

LUK3ARK:
ok regarding external services, it works in prod but cannot send information back to lattice though and thats because currently wrpc only accepts a subject prefix, which needs to be a concatenation of {lattice-id}.{component-id} however when the wrpc then calls interfaces that are defined back inside the lattice, it is unable to because the subject prefix is hardcoded as {lattice-id}.{component-id} so  to solve this for now need a separate wrpc client with a specific subject prefix of only {lattice-id} to send messages back into the lattice, but would be good to edit wrpc client to internally have a subject prefix to use specifically when sending invocations...

@rvolosatovs Do you have any ideas about the best way to do this or a more clever solution? Or is it just best to keep the wrpc clients separate? I wouldn't mind making a PR I just want to know your thoughts of a design pattern :)

LUK3ARK:
Alright so from playing around with this the best way is to wrap a wrpc client with a component-aware context specific to wasmcloud which allows for lattice and component configuration against interfaces generated by wrpc, then each client can just act as a normal wrpc client, but specific to lattice communication.

Virtual/external components then just emerge from the behaviour pretty easily

PR #4231, opened 2025-03-18
chore/warn on max payload

brooksmtownsend:
- **chore(host): warn on large event payload**
- **chore(host): warn on large ctl respose payload**

## Feature or Problem
I was running a bit of an experiment to run as many components on a single host as possible and the only thing that failed with 10000 components is the fact that the heartbeat/host inventory exceeded the max payload. This PR would give a little more information to that failure.

Note that the `server_info()` function call returns the last received information and does not incur a network roundtrip.

Some thoughts:
1. Would it be better to, in the cases where we know a payload size may exceed the max, figure out a way to truncate the payload and still publish part of it? Paginating?
1. Should we return an error instead of warning and then trying anyways?

### DIFF
--- a/crates/host/src/wasmbus/event.rs
+++ b/crates/host/src/wasmbus/event.rs
@@ -4,7 +4,7 @@ use anyhow::Context;
 use cloudevents::{EventBuilder, EventBuilderV10};
 use serde_json::json;
 use time::{format_description::well_known::Rfc3339, OffsetDateTime};
-use tracing::instrument;
+use tracing::{instrument, warn};
 use ulid::Ulid;
 use uuid::Uuid;
 use wascap::jwt;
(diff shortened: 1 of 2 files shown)


thomastaylor312:
> Would it be better to, in the cases where we know a payload size may exceed the max, figure out a way to truncate the payload and still publish part of it?

I think so we can keep track of hosts being alive still, we should do something like truncating the list of components and providers and adding a new field called `status` or `errors` that informs consumers that the list was truncated, as well as anything else we want to report. We might want to rethink how we report out current inventory, maybe by telling the consumer to fetch inventory with a specific "callback URL" as it were

thomastaylor312 (approved):
I think we should merge this so we at least have the warning. See my other comment for additional brainstorming about longer term

PR #4244, opened 2025-03-19
fix(wash): Fixes wash tests on windows

thomastaylor312:
When updating the tests, I forgot to gate the process killing feature behind a target family of Unix. This should fix the failing wash tests on windows


brooksmtownsend (approved):
we only call these functions from unix contexts? lgtm

thomastaylor312:
Yeah these are only happening in the wash dev and up tests. Which I just realized up runs everywhere. Let me add one more feature gate to that

thomastaylor312:
Whoops wrong button

thomastaylor312:
Ok, pushed up a thing to make sure it was gated in the wash up tests as well

Issue #4111, opened 2025-02-12
[RFC] Workload Identity in wasmCloud


joonas:
**NOTE:** This is an initial proposal that will evolve slightly, I'll do my best to highlight any material changes as the proposal evolves.

### RFC Title

RFC: Workload Identity for wasmCloud

### Introduction

This RFC proposes a new platform-level feature to enable wasmCloud components and the workloads running on it to request uniform identity documents for authenticating a given workload independent of the infrastructure it is running on.

### Motivation

wasmCloud has a existing form of identity based on [NKeys](https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_intro/nkey_auth) and JWTs signed with the NKeys.

While NKeys are well suited for the purposes of authenticating against and interacting with NATS itself, they are less well suited for the purposes of integrating with the ecosystem at large due to some important details:

1. Nkeys derived JWTs use a custom `ed25519-nkey` algorithm for encoding/decoding the JWTs, which is not an algorithm supported by any other JWT library besides [NATS' own JWT library](https://github.com/nats-io/jwt).
2. Nkeys do not inherently have a representation that could easily be used for writing policies to match against them. We would have to invent this from scratch, which would then be bespoke to wasmCloud.

## Goals

- Establish a well formed identifier (using [SPIFFE IDs](https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#spiffe-id)) for representing the different types of workloads in wasmCloud, namely: hosts, providers and components.
- Provide each type of workload the ability to request identity documents that can be used to identify them inside and outside of a given wasmCloud deployment.
- Implement a set of use cases where workload identity can be used solve existing challenges, for example:
    - Ability for the host and providers to establish NATS connectivity using workload identity
    - Ability for the host to use workload identity for pulling OCI artifacts

## Non-goals

- Replace existing use of Nkeys for internal identity across wasmcloud
- Make workload identity a pre-requisite for running wasmCloud, rather it should be considered an opt-in feature on top of the core platform

### Detailed Design

In a workload identity-enabled deployments of wasmCloud:

- The wasmCloud host use the [SPIFFE Workload API](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_API.md) to request workload identity documents for itself.
[...]

PR #4245, opened 2025-03-19
feat: Add wasmcloud:identity interface for fetching component workload identity

joonas:
## Feature or Problem

Add experimental feature (`workload-identity-interface`) for fetching workload identity using the `wasmcloud:interface/store@0.1.0-draft` interface in components.

## Related Issues

#4111


crates/host/src/wasmbus/handler.rs
```diff
@@ -1127,3 +1199,37 @@ impl InvocationErrorIntrospect for Handler {
...
 }
+
+// TODO(joonas): Make this more generalized so we can support non-wasmcloud-specific
+// selectors as well.
+//
+// environment variable -> WASMCLOUD_LABEL_wasmcloud__ns=my-namespace-goes-here
+// becomes:
+// SPIRE Selector -> wasmcloud:ns:my-namespace-goes-here
+async fn parse_selectors_from_host_labels(
+    host_labels: tokio::sync::RwLockReadGuard<'_, BTreeMap<String, String>>,
```
thomastaylor312, line 1210: Nit: I think this can just be `&BTreeMap` because `RwLockReadGuard` implements `Deref` for its inner type

crates/host/src/wasmbus/handler.rs
```diff
@@ -64,6 +76,8 @@ pub struct Handler {
     pub invocation_timeout: Duration,
     /// Experimental features enabled in the host for gating handler functionality
     pub experimental_features: Features,
+    /// Labels associated with the wasmCloud Host the component is running on
+    pub host_labels: Arc<RwLock<BTreeMap<String, String>>>,
```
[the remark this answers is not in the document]
thomastaylor312: Now that I look at it again, let's just leave it for now. You'd have to plumb it all the way through from the constructor at the host level. We could probably do something with only copying the labels we need as well, but let's not worry about it for now

tests/fixtures/agent.conf
```diff
@@ -1,7 +1,7 @@
 agent {
     data_dir = "$AGENT_DATA_DIR"
-    log_level = "ERROR"
-    log_file = "/dev/null"
+    log_level = "DEBUG"
+    # log_file = "/dev/null"
```
thomastaylor312, line 4: Just wanted to check, was this supposed to be commented out or left over from debugging?

thomastaylor312 (approved):

tests/fixtures/agent.conf
joonas: Good catch, I thought I put these back to what they were, definitely left over from debugging, thanks!

PR #4289, opened 2025-03-25
feat(tracing): Adds support for resource attributes

thomastaylor312:
We weren't properly discovering resource attributes from the environment. This should fix the issue in both the host and provider. This also changes our start provider process to proxy through all OTEL env vars to the process

### DIFF
--- a/crates/host/src/wasmbus/providers/mod.rs
+++ b/crates/host/src/wasmbus/providers/mod.rs
@@ -395,6 +395,13 @@ async fn provider_command(path: &Path, host_data: Vec<u8>) -> anyhow::Result<pro
         let _ = child_cmd.env("RUST_LOG", rust_log);
     }
 
+    // Pass through any OTEL configuration options to the provider as well
+    for (k, v) in env::vars() {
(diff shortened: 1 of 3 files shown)


crates/host/src/wasmbus/providers/mod.rs
```diff
@@ -395,6 +395,13 @@ async fn provider_command(path: &Path, host_data: Vec<u8>) -> anyhow::Result<pro
         let _ = child_cmd.env("RUST_LOG", rust_log);
     }
 
+    // Pass through any OTEL configuration options to the provider as well
+    for (k, v) in env::vars() {
+        if k.starts_with("OTEL_") {
```
thomastaylor312, line 400: I think this should be safe to do, but wanted to call this out just to make sure we're ok with doing this

brooksmtownsend: Sounds good to me 👍🏻 

No more setting `OTEL_MY_GITHUB_PASSWORD` 😡

brooksmtownsend (approved):
Sweet 🚀

PR #4295, opened 2025-03-27
fix(docs): update links in prep for auto-copy

vados-cosmonic:
This commit updates links that are going to be auto-copied to docs, so they can be used from the wide internet as well as in-repo.


lachieh:
I'm not an examples- or capability-provider-maintainer, but this looks good from a links/docs perspective!

thomastaylor312 (approved):

Issue #4287, opened 2025-03-25
[BUG] Wash plugin does not work


LUK3ARK:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [x] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Describe the bug

wash plugin doesnt run plugins

### Steps to reproduce

When i run wash [plugin] --help, the help menu comes up.

If i run wash [plugin] random_text it says unexpected argument (because my plugin doesn't take any arguments)

but then I run as expected with option flags and just returns the default wash help

### Expected behavior

wash to run plugin

### Environment

wash          v0.40.0

vados-cosmonic:
Hey @LUK3ARK thanks for submitting this! I'll take a look into this, it's very odd, I wonder if this was a weird CLI merge issue from when wash & wash-lib came together. We should also have *some* tests covering this so I'm a bit surprised.

Should have some answers soon.

PR #4297, opened 2025-03-27
fix(wash): use of no-arg plugins

vados-cosmonic:
This commit fixes a bug with wash where a no-argument plugin command (e.g. `wash hello`) was being interpreted as a wash command (but with no actual internal wash command matching), and printing help.

Resolves https://github.com/wasmCloud/wasmCloud/issues/4287

## Testing

I tested this locally after building a no-argument plugin (and a regular one)


thomastaylor312 (approved):
Good to go once tests pass

PR #4347, opened 2025-04-09
Fix(wash)=dev manifest option

vados-cosmonic:


crates/wash/src/cli/cmd/dev/mod.rs
```diff
@@ -75,8 +75,17 @@ pub struct DevCommand {
...
+    #[clap(long = "wadm-version", env = "WADM_VERSION")]
+    pub wadm_version: Option<String>,
+
+    /// If enabled, wadm will not be downloaded or run as a part of the up command
+    #[clap(long = "disable-wadm")]
+    pub disable_wadm: bool,
+
+    /// The `JetStream` domain to use for wadm
+    #[clap(long = "wadm-js-domain", env = "WADM_JS_DOMAIN")]
+    pub wadm_js_domain: Option<String>,
```
brooksmtownsend, line 88: @vados-cosmonic is there a reason we needed to split these fields out? I might be being dumb, but I assume we could just lean on the `WadmOpts` optional to keep `wadm_manifest` as `None` if we don't want folks supplying it on the CLI?

vados-cosmonic: Hey a bit late here, but yes -- the other bundle of options includes a confusing option -- it gets printed out in help text/etc.

brooksmtownsend (approved):

PR #4356, opened 2025-04-10
feat(wash): allow setting link_config via wash dev overrides

vados-cosmonic:
This commit enables setting link config (ex. for the HTTP server) component when setting overrides for generated components.

Component in this sense means a WADM component -- an element used in a WADM manifest (either a provider or wasm component or whatever else).

Up until now `config` has been settable, but this aligns with the generated "component"'s *direct* configuration, not configuration used for the link.

This code enables setting `link_config` on an override, which modifies the generated link (and related config properties).

### Acceptance or Integration

Also added a test for this

### Manual Verification

Tested manually, the following config works:

```toml
[[dev.overrides.exports]]
interface_spec = "wasi:http/incoming-handler@0.2.2"
image_ref = "ghcr.io/wasmcloud/http-server:0.24.0"
link_config = { values = { address = "127.0.0.1:8083" } }
```

### DIFF
--- a/crates/wash/src/cli/cmd/dev/deps.rs
+++ b/crates/wash/src/cli/cmd/dev/deps.rs
@@ -1190,16 +1190,16 @@ impl ProjectDeps {
                     };
 
                     // Add link configurations, generating the defaults if necessary
+                    // NOTE: since this is for a "receiving" dep (e.g. calls go from wasm component -> provider),
+                    // we should be configuring the *target* of the link (i.e. the provider)
                     if link_configs.is_empty() {
                         link_property
-                            .source
-                            .get_or_insert(Default::default())
+                            .target
                             .config
                             .extend(dep.default_link_configs());
                     } else {
                         link_property
-                            .source
-                            .get_or_insert(Default::default())
+                            .target
                             .config
                             .extend_from_slice(link_configs.as_slice());
                     }
@@ -1238,6 +1238,9 @@ impl ProjectDeps {
                     };
 
                     // Add link configurations, generating the defaults if necessary
+                    //
+                    // NOTE: since this is for a "invoking" dep (e.g. calls go from provider -> wasm component/provider),
+                    // we should be configuring the *source* of the link (i.e. the provider)
                     if link_configs.is_empty() {
                         link_property
                             .source


crates/wash/src/cli/cmd/dev/deps.rs
```diff
@@ -122,25 +122,25 @@ impl ProjectDependencyKey {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum DependencySpec {
     /// A dependency that receives invocations (ex. `keyvalue-nats` receiving a `wasi:keyvalue/get`)
-    Exports(DependencySpecInner),
+    Receives(DependencySpecInner),
```
vados-cosmonic, line 125: Yeah I *had* to change this back -- imports/exports is just too confusing to use here. I'm open to `ReceivesInvocations` or something like that, but we need to use different wording.

brooksmtownsend: Yeaaah we definitely need a better nomenclature here across wasmCloud. imports and exports seem to be confusing because the same function could be either depending on your frame of reference 🤔 

receives/gives, needs/provides, implements/calls are all strawmen, but I'm glad you were able to get something that's a little easier to comprehend

crates/wash/src/cli/cmd/dev/deps.rs
```diff
@@ -1016,47 +1189,22 @@ impl ProjectDeps {
...
+                            .source
+                            .get_or_insert(Default::default())
+                            .config
+                            .extend(dep.default_link_configs());
+                    } else {
+                        link_property
+                            .source
+                            .get_or_insert(Default::default())
+                            .config
+                            .extend_from_slice(link_configs.as_slice());
```
brooksmtownsend, line 1204: I saw this and the block below that matches it too, am I reading it right that we always insert link config on the "source"? I was looking through to find where we decided which side of the link gets the config, but I couldn't quite find it

brooksmtownsend (commented):
One big question around which side of the link gets the config, but otherwise looks good!

crates/wash/src/cli/cmd/dev/deps.rs
vados-cosmonic: Ah good catch -- this should be on the *target* here for the receiving side!

brooksmtownsend (approved):

PR #4363, opened 2025-04-14
feat/sqldb postgres secret config

brooksmtownsend:
- **feat(sqldb-postgres)!: support loading config from secrets**
- **chore(providers): bump all for release with otel fixes**

## Feature or Problem
This PR implements support in the postgres provider for loading all config values from secrets, in case all configuration is in the secret store.

I also went ahead and bumped each provider here to save the CI cycle to release the otel fixes from last week.

## Release Information
`next` all providers

### DIFF
--- a/crates/provider-sqldb-postgres/src/config.rs
+++ b/crates/provider-sqldb-postgres/src/config.rs
@@ -63,29 +63,31 @@ pub(crate) fn extract_prefixed_conn_config(
     ];
     match keys
         .iter()
-        .map(|k| config.get(k))
-        .collect::<Vec<Option<&String>>>()[..]
-    {
-        [Some(host), Some(port), Some(username), config_password, Some(database), tls_required, pool_size] =>
-        {
-            let secret_password = secrets
-                .get(&format!("{prefix}PASSWORD"))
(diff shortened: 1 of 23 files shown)


crates/provider-sqldb-postgres/src/config.rs
```diff
@@ -63,29 +63,27 @@ pub(crate) fn extract_prefixed_conn_config(
...
-            let password = match (secret_password, config_password) {
-                (Some(s), _) => s,
-                (None, Some(c)) => {
-                    warn!("secret value [{prefix}PASSWORD] was not found in secrets, but exists in config. Prefer using secrets for sensitive values.", );
-                    c
+        .map(|k| {
+            // Prefer fetching from secrets, but fall back to config if not found
+            match (secrets.get(k).and_then(SecretValue::as_string), config.get(k)) {
+                (Some(s), _) => Some(s),
+                // Offer a warning for the password, but other values are fine to be in config
```
brooksmtownsend, line 70: Do check for if present in both

rvolosatovs (approved):
it seems surprising to me that we'd look up things like pool size in "secrets", but sure!

PR #4398, opened 2025-04-24
feat(host): track active component instances

### LINKED ISSUE #2530: [FEATURE] Timeout component export future based on the provided max_execution_time

brooksmtownsend:
## Feature or Problem
This PR adds a new metric, `wasmcloud_host.component.active_instances`, which tracks the number of currently executing instances of a component. This is extremely useful to be able to measure the actual number of components executing at any given time. This lets you understand if your scale is correct, if you need to scale up, down, etc.

Additionally, this lets you setup alerts if you're approaching your max instances.

A followup commit also adds `wasmcloud_host.component.max_instances` so that you can create views and alerts based on saturation, e.g. `component_active_instances / component_max_instances > 0.8`

## Related Issues
This PR also adds a timeout for the component future which will ensure that semaphores are dropped after a component stops executing, even if it's killed by the epoch counter. Fixes #2530 

## Release Information
wasmCloud 1.8.1

### Manual Verification
Here's an example of the active instances metric tracking over a couple different hosts.

image

And expressed as a ratio:

image

### DIFF
--- a/crates/host/src/metrics.rs
+++ b/crates/host/src/metrics.rs
@@ -3,7 +3,9 @@ use std::time::Duration;
 
 use sysinfo::System;
 use tokio::task::JoinHandle;
-use wasmcloud_tracing::{Counter, Histogram, KeyValue, Meter, ObservableGauge, UpDownCounter};
+use wasmcloud_tracing::{
+    Counter, Gauge, Histogram, KeyValue, Meter, ObservableGauge, UpDownCounter,
+};
 
 const DEFAULT_REFRESH_TIME: Duration = Duration::from_secs(5);
 
@@ -17,8 +19,10 @@ pub struct HostMetrics {
     pub component_invocations: Counter<u64>,
     /// The count of the number of times an component invocation resulted in an error.
     pub component_errors: Counter<u64>,
-    /// The
+    /// The number of active instances of a component.
     pub component_active_instances: UpDownCounter<i64>,
+    /// The maximum number of instances of a component.
+    pub component_max_instances: Gauge<u64>,
 
     /// The total amount of available system memory in bytes.
     pub system_total_memory_bytes: ObservableGauge<u64>,
@@ -95,6 +99,11 @@ impl HostMetrics {
             .with_description("Number of active component instances")
             .build();
 
+        let component_max_instances = meter
+            .u64_gauge("wasmcloud_host.component.max_instances")
+            .with_description("Maximum number of component instances")
+            .build();
+
         let mut system = System::new();
         // Get the initial metrics
         system.refresh_memory();
@@ -167,6 +176,7 @@ impl HostMetrics {
             component_invocations: component_invocation_count,
             component_errors: component_error_count,
             component_active_instances,
+            component_max_instances,
             system_total_memory_bytes: system_memory_total_bytes,
             system_used_memory_bytes: system_memory_used_bytes,
             system_cpu_usage,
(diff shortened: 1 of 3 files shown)


crates/host/src/metrics.rs
```diff
@@ -3,7 +3,9 @@ use std::time::Duration;
```
rajsite, line 3: The [deploy and scale](https://wasmcloud.com/docs/tour/deploy-and-scale/) docs have the following comment:
> Requests are not forwarded to our component anymore
> After multiple requests were received but timed out, it is no longer possible to send further requests to our hello application. The reason for this is that the httpserver capability provider is no longer able to invoke the hello component via NATS. To continue, we must first delete and redeploy our application.

With this change, it might be a better / a useful alternate suggestion to configure a timeout for the component instead of relying on the default of no timeout.

rajsite, line 3: When configuring timeout_ms, I found it confusing that I could not configure the value as a yaml number
```
              config:
                - name: default-http
                  properties:
                    address: 0.0.0.0:30000
                    timeout_ms: 300000
```

But instead had to configure it as a string
```
              config:
                - name: default-http
                  properties:
                    address: 0.0.0.0:30000
                    timeout_ms: '300000'
```

rajsite: There is also a [thread in the slack](https://wasmcloud.slack.com/archives/C02LZ3L0X4H/p1744679881527099) where the timeout fix might address the concerns.

brooksmtownsend: Yes unfortunately this is a `wadm` issue, all configuration is assumed to be `string -> string`. 

I think we may want to fail validation of this if you use a non-string to avoid this confusion

thomastaylor312 (approved):

crates/host/src/metrics.rs
brooksmtownsend: That's fair. Right now our default timeout for the host is 10 minutes, which would be permissive enough to enable longer running workloads, but it may be more useful to set this configuration up to be a shorter timeout for `wash dev`

rajsite: Is that referring to a different timeout than the [provider-http-server timeout which says N/A](https://github.com/wasmCloud/wasmCloud/tree/main/crates/provider-http-server#http-address-configuration) (assumed that meant no timeout)?

rajsite: fyi @brooksmtownsend saw the PR was marked auto-merge a couple days ago and hoping the merge isn't blocked on this comment being open

vados-cosmonic: It was! :) resolved now

PR #4407, opened 2025-04-29
feat(host)!: implement traits for extending host functionality

brooksmtownsend:
## Feature or Problem
This PR implements a variety of traits in the wasmCloud host crate to improve the ability to extend and program the host in embedding scenarios. It's also a fairly large refactor, and contains documentation for all host modules to improve the contributability of the host. I would be happy to discuss this refactor in a pair review session if that would be best! Regardless, I'll try and lay out the larger pieces here.

I've introduced the following traits and plumbed their usage throughout the wasmCloud host binary:

| Trait | Description | Minimal implementation | NATS implementation |
| --- | --- | --- | --- |
| config::ConfigManager | A supertrait of the StoreManager that supports a `watch` operation for configuration that may change at runtime. | A HashMap that doesn't change | A JetStream bucket that can watch for key changes |
| event::EventPublisher | A simple trait with a `publish_event` that will be called for all lattice events. | No-op, or a channel | PUB operations on `wasmbus.evt.<lattice>.<event-name>` |
| policy::PolicyManager | A policy evaluator that can approve or deny starting components, providers, or invocations | Allow by default | If a policy topic is provided at runtime, NATS REQ messages on `wasmcloud.policy.>` |
| secrets::SecretsManager | A secrets manager that fetches a list of secrets at runtime and returns securely boxed secret values | No-op, secrets aren't required | [The wasmCloud secrets API](https://wasmcloud.com/docs/deployment/security/secrets/) |
| store::StoreManager | A simple key-value store with support for `get`, `put`, del` operations that can back the storage of links, ComponentSpecifications, or config | HashMap | A JetStream bucket that is being watched for any changes from other hosts | 
| wasmbus::ProviderManager | A WIP trait for managing messages to and from capability providers. This trait will be more fleshed out in future issues | NATS | NATS |

All of these traits, as hinted at above, are implemented by NATS in the regular distribution of wasmCloud, but if you're embedding the host crate you can choose to override specific pieces of the host functionality. I think a popular example of this will be to default to receiving events over a `broadcast` channel instead of NATS to ensure that event receipt is immediate and actionable.

Or, for example, you can choose to have configuration fetched from Kubernetes ConfigMaps instead of a JetStream bucket if you are only running a host in Kubernetes and don't cluster multiple together.

I would like to follow this PR up in the eventual future with a couple additional enhancements and changes that will highlight the usefulness of this refactor:
1. For the wasmCloud test util, use test implementations of these traits to simplify running wasmCloud in unit tests (e.g. not requiring NATS Jetstream for basic operations, but still using NATS for RPC)
2. Embedding wasmCloud inside of `wash dev`, supporting direct calls of control interface functions on the in-process host
3. Enhancing the `ProviderManager` trait so that it could be implemented with a variety of different protocols, as long as the provider can communicate with the host over this abstraction we have freedom to run providers as things other than binary plugins
4. Introducing a more generic creation of wRPC transport clients, which would support components and providers communicating over any one of the various wRPC transports.

## Related Issues
https://github.com/wasmCloud/wasmCloud/discussions/4406

## Release Information
wasmcloud-host 0.26
wasmCloud 1.9 (technically could be a patch release, but enough changed that it warrants minor imo)

## Consumer Impact
There will be nothing changed from the perspective of wasmCloud binary users.

Embedders of the wasmcloud-host crate or wasmcloud-test-util will need to refactor to either:
1. Use the `HostBuilder` to run the host with the implementations of their choosing
5. Use the `NatsHostBuilder` to run the host with the existing NATS implementations.

### Acceptance or Integration
Two integration tests changed references since modules were renamed.

### Manual Verification
I've run this manually and with a couple of sample embedding scenarios to ensure this host works as expected


brooksmtownsend:
Just a note that I've left a couple of TODOs in the code that I'm going through and pruning, but I wanted to get CI up and running to ensure that I have this functionally working well as it is locally

thomastaylor312:
Just a note that I've started reviewing this. It is a chonky boi though so I might be finishing it tomorrow

crates/host/src/lib.rs
```diff
@@ -1,39 +1,48 @@
-//! wasmCloud host library
-
+#![doc = include_str!("../README.md")]
 #![warn(missing_docs)]
 #![forbid(clippy::unwrap_used)]
 
-/// wasmbus host
-pub mod wasmbus;
+/// [crate::config::ConfigManager] trait for managing a config store which can be watched to receive
```
thomastaylor312, line 5: Nit: I feel like this style of documenting these modules is a little odd. It is describing it primarily as if it were the trait, and it also is all in this top level. I feel like it would be cleaner to put these as module level comments in the module files themselves and to reword such that it is describing the module and then saying something like "the main feature of this module is the [crate::config::ConfigManager] trait..."

crates/host/src/nats/builder.rs
```diff
@@ -0,0 +1,251 @@
...
+    data_store: Store,
+    policy_manager: Option<Arc<dyn PolicyManager>>,
+    secrets_manager: Option<Arc<dyn SecretsManager>>,
+    event_publisher: Option<Arc<dyn EventPublisher>>,
+}
+
+impl NatsHostBuilder {
+    /// Initialize the host with the NATS control interface connection
+    #[allow(clippy::too_many_arguments)]
+    pub async fn new(
```
thomastaylor312, line 61: Nit: I think the cleaner API here (since you're already using a builder) would be set all of the defaults in the builder and have `default` and `new` be the same thing. `build` can take the NATS client as a required argument and all the other items can have methods for setting them

crates/host/src/nats/ctl.rs
```diff
@@ -0,0 +1,385 @@
+//! The NATS implementation of the control interface.
+
+use anyhow::Context as _;
+use async_nats::jetstream::kv::Store;
+use bytes::Bytes;
+use futures::future::Either;
+use futures::stream::SelectAll;
+use futures::{Stream, StreamExt, TryFutureExt as _};
+use serde::Serialize;
+use std::pin::Pin;
```
thomastaylor312, line 10: Super nit: Put `std` imports up top, separate from crate imports

crates/host/src/nats/policy.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// Construct a new policy manager. Can fail if policy_changes_topic is set but we fail to subscribe to it
+    #[instrument(skip(nats))]
+    pub async fn new(
+        nats: async_nats::Client,
+        host_info: HostInfo,
+        policy_topic: Option<String>,
+        policy_timeout: Option<Duration>,
+        policy_changes_topic: Option<String>,
+    ) -> anyhow::Result<Self> {
+        const DEFAULT_POLICY_TIMEOUT: Duration = Duration::from_secs(1);
```
thomastaylor312, line 50: Nit: Define this const at the top

crates/host/src/nats/store.rs
```diff
@@ -0,0 +1,299 @@
...
+    #[instrument(level = "debug", skip(self))]
+    async fn get(&self, key: &str) -> anyhow::Result<Option<Bytes>> {
+        self.get(key)
+            .await
+            .map_err(|err| anyhow::anyhow!("Failed to get config: {}", err))
+    }
+
+    #[instrument(level = "debug", skip(self, value))]
+    async fn put(&self, key: &str, value: Bytes) -> anyhow::Result<()> {
+        self.put(key, value)
```
thomastaylor312, line 35: Are there any worries about race conditions here? In other words, should we be using the revision number?

crates/host/src/policy.rs
```diff
@@ -1,26 +1,72 @@
...
+    async fn evaluate_start_component(
+        &self,
+        _component_id: &str,
+        _image_ref: &str,
+        _max_instances: u32,
+        _annotations: &BTreeMap<String, String>,
+        _claims: Option<&jwt::Claims<jwt::Component>>,
+    ) -> anyhow::Result<Response> {
+        Ok(Response {
+            request_id: Uuid::new_v4().to_string(),
```
thomastaylor312, line 26: Nit: This means that any request with the default policy manager will still be spending cycles generating a uuid and allocating a string. We should probably do what NATS does and just return an empty string at least. We should also find a way to get rid of the string allocation in the future as well, but definitely not here

crates/host/src/store.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// Inserts or updates a key-value pair in the config store.
+    async fn put(&self, key: &str, value: Bytes) -> anyhow::Result<()>;
+
+    /// Deletes a key from the config store.
+    async fn del(&self, key: &str) -> anyhow::Result<()>;
+}
+
+/// A struct that implements the StoreManager trait, storing data in an in-memory HashMap.
+#[derive(Default)]
+pub struct DefaultStore {
```
thomastaylor312, line 23: Nit: Retitle this to `InMemoryStore` to be more explicit about what it does. Also because `DefaultStore::default()` looks funny to me (and I hear the go compiler yelling about stuttering)

crates/host/src/wasmbus/ctl.rs
```diff
@@ -986,7 +1002,7 @@ impl ControlInterfaceServer for Host {
             .uptime_seconds(uptime.as_secs())
             .uptime_human(human_friendly_uptime(uptime))
             .version(self.host_config.version.clone())
-            .ctl_host(self.host_config.ctl_nats_url.to_string())
+            .ctl_host(self.host_config.rpc_nats_url.to_string())
```
thomastaylor312, line 1005: Nit: Might be worth a comment here to note that we removed the `ctl_nats_url` from the host config so people don't get tripped up later when they don't match

thomastaylor312 (approved):
This is great and makes the host code so much more readable and understandable! Everything here is optional nits

PR #4410, opened 2025-04-29
feat(control-interface)!: add wait_for_completion on long tasks

### LINKED ISSUE #4089: [BUG] Host Get Inventory State Consistency
### Affected project(s)

- [ ] documentation
- [ ] examples
- [x] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

There is a slight delay between calling  `scale_component`  & having the component show up in `host_inventory`.

It seems the host only starts reporting the component status after it has processed the "component start task" / emitted the `component_start` event. This makes the host report inconsistent state for a brief period.

```
/tmp ❯ ./bug.sh
Sending request at Sat  8 Feb 2025 17:32:42 EST
{"success":true,"message":""}


Sat  8 Feb 2025 17:32:42 EST
Sat  8 Feb 2025 17:32:43 EST
Sat  8 Feb 2025 17:32:44 EST
"default-hello-world-foo"
Sat  8 Feb 2025 17:32:45 EST
"default-hello-world-foo"
```

### Steps to reproduce

- Send component scale
- Request host inventory

Notice how it takes a few seconds for the component to show up in inventory.

```
#!/bin/bash

FIRST_HOST="$(wash get inventory -o json | jq -r '.inventories[0].host_id')"

echo "Sending request at $(date)"
cat <<EOF | nats req "wasmbus.ctl.v1.default.component.scale.$FIRST_HOST" 2>/dev/null
{"component_id":"default-hello-world-foo","component_ref":"ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0","annotations":{"component":"hello-world-foo","generation":"12","namespace":"default"},"count":10,"host_id":"$FIRST_HOST"}
EOF

while true; do
	date
	nats req "wasmbus.ctl.v1.default.host.get.$FIRST_HOST" '{}' 2>/dev/null | jq '.response.components[].id'
	sleep 1
done
```

### Expected behavior

After `scale_component` is received, the host must report the component on every `host_inventory` and `host_heartbeat`, regardless of the component status.
This is also valid for providers.

### Environment

wasmCloud canary

_No response_

brooksmtownsend:
## Feature or Problem
This PR adds a `wait_for_completion` field on the component scale, component update, and provider start control interface commands that will, instead of returning an ack after validating the payload, will wait for the entire operation to complete before returning a success or error message.

## Related Issues
Fixes #4089, if you're needing a synchronous operation to complete before taking other actions this will do the trick

## Release Information
wasmCloud control interface next minor
wasmCloud v1.9

### Manual Verification
Below is a demonstration of hitting the host with each variety of a provider start, first without waiting for completion (notably even with an invalid OCI reference we get a success) and then waiting for completion and getting the actual response.

```bash
➜ nats req "wasmbus.ctl.v1.default.provider.st16:42:43 [5/81]
FMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT" '{"provider_id": "http5", "provider_ref": "ghcr.dio/wasmcloud/http-client:0.9.0", "wait_for_completion": false}'
16:41:49 Sending request on "wasmbus.ctl.v1.default.provider.start.NBEQTHIHUMWFMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT"
16:41:49 Received with rtt 1.491292ms
{"success":true,"message":"received request, starting provider"}

 ➜ nats req "wasmbus.ctl.v1.default.provider.start.NBEQTHIHUMW
FMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT" '{"provider_id": "http6", "provider_ref": "ghcr.io/wasmcloud/http-client:0.9.0", "wait_for_completion": false}'
16:42:11 Sending request on "wasmbus.ctl.v1.default.provider.start.NBEQTHIHUMWFMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT"
16:42:11 Received with rtt 1.801625ms
{"success":true,"message":"received request, starting provider"}

➜ nats req "wasmbus.ctl.v1.default.provider.start.NBEQTHIHUMWFMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT" '{"provider_id": "http8", "provider_ref": "ghdcr.io/wasmcloud/http-client:0.9.0", "wait_for_completion": true}'
16:47:49 Sending request on "wasmbus.ctl.v1.default.provider.start.NBEQTHIHUMWFMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT"
16:47:49 Received with rtt 96.642375ms
{"success":false,"message":"failed to fetch provider"}

➜ nats req "wasmbus.ctl.v1.default.provider.start.NBEQTHIHUMW
FMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT" '{"provider_id": "http7", "provider_ref": "ghcr.io/wasmcloud/http-client:0.9.0", "wait_for_completion": true}'
16:42:35 Sending request on "wasmbus.ctl.v1.default.provider.start.NBEQTHIHUMWFMQDNP4Y55HPMIM62GIEQCJWHYD7HWYFERJACWDLIYUFT"
16:42:36 Received with rtt 1.175606291s
{"success":true,"message":"successfully started provider"}
```

### DIFF
--- a/crates/control-interface/src/client.rs
+++ b/crates/control-interface/src/client.rs
@@ -616,6 +616,7 @@ impl Client {
             component_id: IdentifierKind::is_component_id(existing_component_id)?,
             new_component_ref: IdentifierKind::is_component_ref(new_component_ref)?,
             annotations,
+            wait_for_completion: false,
         })?;
         match self.request_timeout(subject, bytes, self.timeout).await {
             Ok(msg) => Ok(json_deserialize(&msg.payload)?),
--- a/crates/control-interface/src/types/ctl.rs
+++ b/crates/control-interface/src/types/ctl.rs
@@ -109,13 +109,16 @@ pub struct ScaleComponentCommand {
     /// 6}
     #[serde(default)]
     pub(crate) config: Vec<String>,
-    #[serde(default)]
     /// Whether to perform an update if the details of the component (ex. component ID) change as
     /// part of the scale request.
     ///
     /// Normally this is implemented by the receiver (ex. wasmcloud host) as a *separate* update component call
     /// being made shortly after this command (scale) is processed.
+    #[serde(default)]
     pub(crate) allow_update: bool,
+    /// Whether to wait for the scale operation to complete before returning a response.
+    #[serde(default)]
+    pub(crate) wait_for_completion: bool,
 }
 
 impl ScaleComponentCommand {
@@ -155,6 +158,11 @@ impl ScaleComponentCommand {
     }
 
     #[must_use]
+    pub fn wait_for_completion(&self) -> bool {
+        self.wait_for_completion
+    }
+
+    #[must_use]
     pub fn builder() -> ScaleComponentCommandBuilder {
         ScaleComponentCommandBuilder::default()
     }
@@ -171,6 +179,7 @@ pub struct ScaleComponentCommandBuilder {
     host_id: Option<String>,
     config: Option<Vec<String>>,
     allow_update: Option<bool>,
+    wait_for_completion: bool,
 }
 
 impl ScaleComponentCommandBuilder {
@@ -221,6 +230,12 @@ impl ScaleComponentCommandBuilder {
         self
     }
 
+    #[must_use]
+    pub fn wait_for_completion(mut self, v: bool) -> Self {
+        self.wait_for_completion = v;
+        self
+    }
+
     pub fn build(self) -> Result<ScaleComponentCommand> {
         Ok(ScaleComponentCommand {
             component_ref: self
@@ -236,6 +251,7 @@ impl ScaleComponentCommandBuilder {
                 .ok_or_else(|| "host id is required for scaling hosts host".to_string())?,
             config: self.config.unwrap_or_default(),
             allow_update: self.allow_update.unwrap_or_default(),
+            wait_for_completion: self.wait_for_completion,
         })
     }
 }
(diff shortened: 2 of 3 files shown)


thomastaylor312 (changes requested):
Code looks ok here, but I do have a couple of concerns:

- If we go this route, then the client should also be updated to allow you to set timeout for the functions that block like this, otherwise this will timeout using the default we have set for the NATS client (which I think is just the default of 10s). Another option would to intentionally set no timeout, but [according to the docs](https://docs.rs/async-nats/latest/async_nats/client/struct.Request.html#method.timeout) that could result in a deadlock
- Is there a reason we can't make this client side and build it in there? It could do a watch on the event topics after sending and wait for the event to come in. This would require no changes to the host and be more in line with the event driven nature of using the NATS API. I think wait for completion would be good for an HTTP or gRPC style API, but not as much with a message bus based one

Any thoughts on this?

brooksmtownsend:
Yeah @thomastaylor312 this is primarily driven by how all clients (e.g. our implementations in Rust, Go, when we bring back the `wasmcloud:latticecontrol` interface) have to build in the event watching which can be a little frustrating, especially if there are a lot of events coming across. I was hoping that this would be a really easy additional option for the clients that prefer to only hear back once it's completed successfully or not, especially since all operations but these three are already synchronous / not event based.

LMK what you think

thomastaylor312:
Yeah I'm totally ok with it, my main concern is around the timeout issue (such as with provider starting or anything that does initial download). If we can provide a way to set that or provide a good experience around it, I'm good with this

brooksmtownsend:
Let's hold off on this one for a bit, I think this is nice but it can be implemented with event watching (like wash does) and we should consider this more when we think more about the control interface API

PR #4411, opened 2025-04-29
fix: use allow-latest for allow_latest, default registry config improvement

### LINKED ISSUE #4298: [BUG] ALLOW_LATEST not taking effect
### Affected project(s)

- [ ] documentation
- [ ] examples
- [x] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] capability providers
- [ ] provider bindgen
- [ ] other / not sure

### Describe the bug

Setting the env var but it is not taking effect

Image

### Steps to reproduce

```
export WASMCLOUD_OCI_ALLOW_LATEST=true
wasmcloud
```

schedule a component with tag latest

### Expected behavior

latest tags allowed.

### Environment

host 1.6.2

probs this https://github.com/wasmCloud/wasmCloud/blob/main/crates/core/src/registry.rs#L123

brooksmtownsend:
## Feature or Problem
An unfortunate error that led to frustrating failures when enabling `ALLOW_LATEST`, actually not taking effect.

~I've made this fix, but it doesn't fully address the issue and I know where it's coming from. So I'll mark this as draft for now and follow up tomorrow.~

There was an additional step in the host that was necessary to fix which is that if there is not a registry explicitly configured, we should fall back on the provided defaults as given to wasmCloud at startup. Instead of defaulting to the defaults of the `RegistryConfig` struct (e.g. allow_latest: false) we will now fall back to the provided additional CA certs, allow insecure, allow latest, and anonymous auth.

## Related Issues
Fixes #4298

### DIFF
--- a/crates/core/src/registry.rs
+++ b/crates/core/src/registry.rs
@@ -120,7 +120,7 @@ impl RegistryConfigBuilder {
             } else {
                 self.auth.context("missing registry auth")?
             },
-            allow_latest: self.allow_insecure.unwrap_or_default(),
+            allow_latest: self.allow_latest.unwrap_or_default(),
             allow_insecure,
             additional_ca_paths: self.additional_ca_paths.unwrap_or_default(),
         })
--- a/crates/host/src/lib.rs
+++ b/crates/host/src/lib.rs
@@ -52,7 +52,7 @@ use tokio::fs;
 use tracing::{debug, instrument, warn};
 use url::Url;
 use wascap::jwt;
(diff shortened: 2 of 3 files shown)


thomastaylor312 (approved):
Sorry @brooksmtownsend this slipped through the cracks on me. Should be good to go!

thomastaylor312:
So there are definitely compile errors on the unit tests on merge. Not sure why it is happening @brooksmtownsend but probably a small fix when you have the chance. No rush here

Issue #4187, opened 2025-03-07
[FEATURE] Show available updates for wash/wadm/nats when available


lachieh:
### Affected project(s)

- [x] wasmCloud CLI (wash)

### Describe the solution you'd like

It would be nice to see the latest versions, and provide an upgrade path in the CLI

brooksmtownsend:
Now that we have automatic downloads for wadm/wasmcloud on the patch versions, and in #1061 we'll probably remove the need for auto updates, I think this issue should just be a notification when a new `wash` version is available 😄

brooksmtownsend:
So, basically, I think this is doable by anyone! When running wash commands (maybe just once a day, to avoid doing it all the time) we should check to see if there is a newer github release for `wash-v*` in this repository.

This operation should be allowed to fail without any side effects

brooksmtownsend:
I think this is complete now!

PR #4425, opened 2025-05-04
feat(wash): add information when new version is available

f4z3r:
## Feature or Problem

Prints information to stderr when a new version of wash is available as a GitHub release. I used a similar format at the information message for when new versions of `wadm` are available. Not sure it fits very well in this case though ...

## Related Issues

- #4187 

## Consumer Impact

When customers run an outdated version of `wash`, they will get informed that a new version is available via a small information header when running `wash`:

```
ℹ️  Consider upgrading to newest wash version: v0.41.0
```

## Testing

Tested manually. I considered adding a test, but this depends on external factors (what releases are on GitHub and what crate version is being tested) which would need to be mocked or faked for a stable test. This this is for information only and does not affect `wash` behavior I decided against implementing a complex test for this.

### Unit Test(s)

See above.

### Acceptance or Integration

No changes.

### Manual Verification

Changed the crate version to older versions and ran `wash -V` and `wash -h` to check that the information gets shown with the newest available GitHub release of `wash`. Also ran with current version `0.41.0` to check that nothing gets printed to stderr.


f4z3r:
With the current implementation every invocation of `wash` will perform a check which GitHub releases have been published, and print the info messages. If the fetching of GitHub releases and information should only be done once day, I would very much like a small pointer as to how you would prefer this to be implemented. I very unfamiliar with the internals of `wash`. I know it stores state under `~/.wash`, but I am not sure whether you would want an additional state file there to store information on the last time the version check was run.

crates/wash/src/bin/wash.rs
```diff
@@ -765,3 +772,23 @@ async fn ensure_plugin_scratch_dir_exists(
     }
     Ok(dir)
 }
+
+async fn inform_new_wash_version() -> anyhow::Result<()> {
+    let wash_version = Version::parse(clap::crate_version!())?;
+    let tag_pattern = "wash-v.*";
```
vados-cosmonic, line 778: NIT: This could *probably* be a constant in here, I think

crates/wash/src/bin/wash.rs
```diff
@@ -765,3 +772,23 @@ async fn ensure_plugin_scratch_dir_exists(
...
+    let releases = get_sorted_releases(
+        GITHUB_WASMCLOUD_ORG,
+        GITHUB_WASMCLOUD_WASMCLOUD_REPO,
+        &wash_version,
+        Some(tag_pattern),
+    )
+    .await?;
```
vados-cosmonic, line 785: A `.context` here with a description of what we were doing would be great here. Something like `.context("retrieving sorted wash releases")`.

crates/wash/src/bin/wash.rs
```diff
@@ -765,3 +772,23 @@ async fn ensure_plugin_scratch_dir_exists(
...
+    if let Some(latest_release) = releases.first().map(|x| x.get_x_y_z_version()).transpose()? {
+        eprintln!(
+            "{} Consider upgrading to newest wash version: v{}",
+            emoji::INFO_SQUARE,
+            latest_release
```
vados-cosmonic, line 790:
```suggestion
            "{} Consider upgrading to newest wash version: v{latest_release}",
            emoji::INFO_SQUARE
```

crates/wash/src/bin/wash.rs
```diff
@@ -765,3 +772,23 @@ async fn ensure_plugin_scratch_dir_exists(
     }
     Ok(dir)
 }
+
+async fn inform_new_wash_version() -> anyhow::Result<()> {
+    let wash_version = Version::parse(clap::crate_version!())?;
```
vados-cosmonic, line 777:
```suggestion
    let wash_version = Version::parse(clap::crate_version!()).context("parsing version from current crate")?;
```

crates/wash/src/bin/wash.rs
```diff
@@ -386,6 +391,8 @@ async fn main() {
         .with_env_filter(EnvFilter::from_default_env())
         .init();
 
+    let _ = inform_new_wash_version().await;
```
vados-cosmonic, line 394:
```suggestion
    inform_new_wash_version().await?;
```

vados-cosmonic (commented):
Hey @f4z3r this looks great! I left some initial comments. 

One thing I'm a bit hesitant about is that this means that *every* call to `wash` will induce a call against GH's API which I'm not thrilled about... 

One way I'm thinking we could mitigate this is maybe doing something to rate limit the check on our side pre-emptively or find some way that doesn't depend on the official GH API (a JSON file in our GH repo?)

Maybe we could put this behind an flag (only enabled when experimental ones are enabled) for now so this individual PR could be merged while we figure out a reasonable path forward?

It would be great to discuss this in wasmCloud slack as well (if not just an issue that we open for how to move forward)

crates/wash/src/bin/wash.rs
vados-cosmonic: Yeah that's a good point -- I'd prefer if things didn't fail silently, but you're right that it's an optional helpful notice, not required for actually running. 

I think if we go with putting this behind an experimental flag for now we can make sure it passes (then when that experimental flag goes GA, we can allow the failure?)

brooksmtownsend:
Great stuff @f4z3r thank you for implementing this! Definitely agree with the approach to making this an operation that could fail without impeding other wash functionality.

re:
> One way I'm thinking we could mitigate this is maybe doing something to rate limit the check on our side pre-emptively or find some way that doesn't depend on the official GH API (a JSON file in our GH repo?)

Perhaps what we could do, similar to our `~/.wash/.completion_suggested` file, is write the current version / current date to a file like `~/.wash/.upgrade_suggested` and only check for new upgrades once the version matches the current version, the file is missing, or the date changed? I'm totally open to different ways to do this. I slightly lean towards something like only suggesting an upgrade once per-wash version so that you're not constantly getting the notification or delaying commands for an HTTP request to the GitHub API. Alternatively, @vados-cosmonic's idea to potentially pull a JSON file may work great, with https://github.com/taiki-e/install-action/blob/main/manifests/wash.json for example!

f4z3r:
Hi @brooksmtownsend and @vados-cosmonic thanks for all the feedback. Really makes me want to help out on this project! I was in contact with one of your maintainers at the KubeCon in London last month and he told me this was a great project to contribute to. I have to say I am positively surprised by the speed and and level of feedback nonetheless.

I am out of town until Monday, but will finish this when I come back next week with an experimental flag in at first. I will also create another PR to consolidate the version checks with `wash up`, I like that idea. I might be able to squeeze some time off work generally over the summer months and maybe contribute more seriously to this project...

vados-cosmonic:
Thanks for the detailed update @f4z3r ! If you don't mind we can help get this PR across the line, and when you have time you can do the follow up? If you'd like to create an issue to track that, we can assign you right away.

f4z3r:
> Thanks for the detailed update @f4z3r ! If you don't mind we can help get this PR across the line, and when you have time you can do the follow up? If you'd like to create an issue to track that, we can assign you right away.

Sure, if you want to get this done faster I understand. Feel free to get this PR done. Otherwise I don't mind finishing it next week. I will create an issue about the version "caching" and consolidation with `wasm up` in a sec.

vados-cosmonic:
Absolutely don't want to steal your thunder here @f4z3r ! Happy to wait till you get a chance to get to it next week :) Feel free to let me know if you need anything

crates/wash/src/bin/wash.rs
f4z3r: Updated that an error causes the program to terminate.

crates/wash/src/bin/wash.rs
f4z3r: Added context, I used different wording with "failed to ..." to match how context is used elsewhere on results.

f4z3r:
Ok I updated the code to only run when the `--experimental` flag is provided, and to then propagate the error (in this case terminate early) in case there is an error. As the code is dependent on the flag, it can thus only run after clap does the parsing of the arguments, and thus will not trigger if `wash` terminates during the parsing. This is typically only the case when either using `-h` or when using `wash` incorrectly. In these cases clap will print the help or usage respectively and then terminate, thus not triggering the version check. This is not too bad IMO. In other cases it will run and report the info message.

crates/wash/src/bin/wash.rs
```diff
@@ -480,6 +485,14 @@ async fn main() {
         }
     };
 
+    // print info on new wash version if available
+    if cli.experimental {
+        if let Err(e) = inform_new_wash_version().await {
+            eprintln!("Error retreiving new wash version: {}", e);
```
f4z3r, line 491: Mhhh just realised that this error message might be misleading as to what was being done when it failed. Maybe something like "Error while checking for new wash version:" might be better...

crates/wash/src/bin/wash.rs
```diff
@@ -765,3 +778,28 @@ async fn ensure_plugin_scratch_dir_exists(
...
+        .context("failed to parse version from current crate")?;
+    const TAG_PATTERN: &str = "wash-v.*";
+    let releases = get_sorted_releases(
+        GITHUB_WASMCLOUD_ORG,
+        GITHUB_WASMCLOUD_WASMCLOUD_REPO,
+        &wash_version,
+        Some(TAG_PATTERN),
+    )
+    .await
+    .context("failed to retrieve sorted wash releases")?;
```
vados-cosmonic, line 793:
```suggestion
    .context("retrieving sorted wash releases")?;
```

crates/wash/src/bin/wash.rs
```diff
@@ -765,3 +778,28 @@ async fn ensure_plugin_scratch_dir_exists(
     }
     Ok(dir)
 }
+
+async fn inform_new_wash_version() -> anyhow::Result<()> {
+    let wash_version = Version::parse(clap::crate_version!())
+        .context("failed to parse version from current crate")?;
```
vados-cosmonic, line 784:
```suggestion
        .context("parsing crate version")?;
```

vados-cosmonic (approved):
LGTM :rocket: 

Nothing but nits -- thanks for coming back to this @f4z3r . If you wouldn't mind addressing the nits one way or the other and then closing your own outstanding comments I think this is ready to merge.

crates/wash/src/bin/wash.rs
vados-cosmonic: You know what, you're right here -- I was under the impression that we needed to lean more into only having the thing that *was happening* here, but in the [official docs](https://docs.rs/anyhow/latest/anyhow/trait.Context.html) it's a combination of the failure fact *and* what was happening. 

Your previous message was perfect.

vados-cosmonic: Please let me know if you want to change this back real quick and I'll merge right after you do!

f4z3r: Updated it back to the old messages and rebased on the latest version of `main` 👍🏽

crates/wash/src/lib/start/github/api.rs
```diff
@@ -24,7 +24,7 @@ const VERSION_FETCHER_CLIENT_USER_AGENT: &str =
 /// Gets a list of releases that are more newer, but semver compatible with the provided `after_version`. They are
 /// sorted by the version number, and the published date. Optionally, a tag pattern can be provided to filter the
 /// releases. If this is not provided, only the main releases are considered.
-async fn get_sorted_releases(
+pub async fn get_sorted_releases(
```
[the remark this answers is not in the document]
f4z3r: This does not seem to work. I am not sure I understand why though. I would have thought that this then becomes visible in the entire crate. And the re-export in `start/github/mod.rs` with `pub use api::*` should re-export it. The same with `pub use github::*;` in `start/mod.rs`... Or is the library (`src/lib`) internally handled like a different crate and thus the function not exposed to anything within `bin/`?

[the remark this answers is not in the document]
vados-cosmonic: Ah yes, so the binary crate and the lib crate aren't the same -- trying to use a crate local from the bin crate won't work.

What do you think about exposing a slightly more ergonomic function instead? Something like `async fn get_newer_wash_versions(v: Version) -> anyhow::Result<Vec<GithubRelease>>` seems a little more appropriate for `wash`  (the lib) to expose to everyone (and `wash` the bin crate).

`get_sorted_releases` feels like internal plumbing, which is why it sems wrong to make it public now.

In fact, now that I look, there *already is* a `new_tag_pattern_patch_releases_after` that is already public -- can you rework the code to use that?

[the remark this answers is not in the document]
f4z3r: Mhhh actually nevermind, after testing this, turns out it does what I thought it does. So if I fake being on version 0.40.0, it will not notify that 0.41.0 is available, as it is not a new patch version of 0.40.0. So maybe a new function that has a more descriptive name like `get_newer_wash_versions` is the best option. What do you think?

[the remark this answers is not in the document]
vados-cosmonic: > Mhhh actually nevermind, after testing this, turns out it does what I thought it does. So if I fake being on version 0.40.0, it will not notify that 0.41.0 is available, as it is not a new patch version of 0.40.0

Ahh that's unfortunate! While we're here, let's maybe get a test in so that this doesn't regress (and someone doesn't wander along and make the mistake of not using your new function).

> So maybe a new function that has a more descriptive name like get_newer_wash_versions is the best option. What do you think?

Yeah that sounds great -- `get_wash_versions_newer_than(...)` ? I'm fine either way on that, hard to name these functions in a way that's easy to understand but not super verbose!

PR #4440, opened 2025-05-08
chore(host): Reference wasmcloud_core::rpc helper methods consistently

joonas:
## Feature or Problem

Turns out we were not using the [`wasmcloud_core::rpc` helper methods](https://github.com/wasmCloud/wasmCloud/blob/main/crates/core/src/rpc.rs#L25-L68) for provider subjects consistently.


vados-cosmonic (approved):
LGTM :rocket:

Issue #4168, opened 2025-02-26
[FEATURE] Memory Limits per Component


lxfontes:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [x] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [x] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Today, all components share the same memory limit.
We'd like to limit the maximum amount of memory individually per component.

### Describe the solution you'd like

The memory limit set at Component level overrides the default memory limit set at host start. There might be a limitation of one being larger/smaller than the other.

The memory limit per component is part of Control Interface component requests, and reflected in heartbeats / inventory.

The memory limit  per component is configurable via wadm.yaml.

The memory limit per component is Host local, meaning: Setting it to 1Gb means every component instance gets 1Gb, and not that 1Gb is divided across all instances.

### Additional context

The approach used by Spin seems translatable to wasmCloud: https://github.com/spinframework/spin/blob/main/crates/core/src/limits.rs

This also paves the way to expose memory information per component, as we can hook the StoreLimiter to OpenTelemetry.

vados-cosmonic:
The project has changed so drastically that this idea is probably not reasonable to pursue in this form -- closing this as we're focusing effort now on wasmCloud v2.

If you think this definitely still applies towards v2, please feel free to open a new issue!

PR #4451, opened 2025-05-14
feat(host): Add per component memory limits

Aditya1404Sal:
## Feature or Problem
Adds global configurabillity to component's maximum memory limit, instead of using the host level `wasmtime::Engine` , a component level `wasmtime::Engine` is created with the limits applied. (Note : `wasmtime::Engine` is the global resource manager for all the instances, they take higher precedence than the store level configuration, applied using `StoreLimitsAsync`. This construct creates soft limits and cannot overide the base limits created at the host level because `wasmtime::Engine` is immutable for security measure)
This PR seeks to implement a de-coupled engine instantiation per component, essentially isolating the components even further.

### Background
resource allocation and limits in wasmtime follows : Global Pooling Resource Allocator >> Per-store Resource limiting 

### TODO
- let the control inventory heartbeat/s and inventory subcommand reflect the configured max_memory_limit
- add observability for the limit metric using otel.
### WIP
- let the memory limit be configurable through wadm [ Added a PR on wasmcloud/wadm ]
## Related Issues
#4168

### DIFF
--- a/crates/control-interface/src/client.rs
+++ b/crates/control-interface/src/client.rs
@@ -289,12 +289,14 @@ impl Client {
     /// * `allow_update` - Whether to perform allow updates to the component (triggering a separate update)
     ///
     #[instrument(level = "debug", skip_all)]
+    #[allow(clippy::too_many_arguments)]
     pub async fn scale_component(
         &self,
         host_id: &str,
         component_ref: &str,
         component_id: &str,
         max_instances: u32,
+        component_limits: Option<HashMap<String, String>>,
         annotations: Option<BTreeMap<String, String>>,
         config: Vec<String>,
     ) -> Result<CtlResponse<()>> {
@@ -307,6 +309,7 @@ impl Client {
         debug!("scale_component:request {}", &subject);
         let bytes = json_serialize(ScaleComponentCommand {
             max_instances,
+            component_limits,
             component_ref: IdentifierKind::is_component_ref(component_ref)?,
             component_id: IdentifierKind::is_component_id(component_id)?,
             host_id,
@@ -957,6 +960,7 @@ mod tests {
                 component_ref,
                 component_id,
                 1,
+                None,
                 None,
                 Vec::with_capacity(0),
             )
--- a/crates/control-interface/src/client.rs
+++ b/crates/control-interface/src/client.rs
@@ -296,7 +296,6 @@ impl Client {
         component_ref: &str,
         component_id: &str,
         max_instances: u32,
-        component_limits: Option<HashMap<String, String>>,
         annotations: Option<BTreeMap<String, String>>,
         config: Vec<String>,
     ) -> Result<CtlResponse<()>> {
@@ -309,7 +308,6 @@ impl Client {
         debug!("scale_component:request {}", &subject);
         let bytes = json_serialize(ScaleComponentCommand {
             max_instances,
-            component_limits,
             component_ref: IdentifierKind::is_component_ref(component_ref)?,
             component_id: IdentifierKind::is_component_id(component_id)?,
             host_id,
@@ -960,7 +958,6 @@ mod tests {
                 component_ref,
                 component_id,
                 1,
-                None,
                 None,
                 Vec::with_capacity(0),
             )
--- a/crates/host/src/wasmbus/mod.rs
+++ b/crates/host/src/wasmbus/mod.rs
@@ -44,7 +44,7 @@ use wasmcloud_control_interface::{
 };
 use wasmcloud_core::ComponentId;
 use wasmcloud_runtime::capability::secrets::store::SecretValue;
-use wasmcloud_runtime::component::WrpcServeEvent;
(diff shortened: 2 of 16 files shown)


Aditya1404Sal:
@brooksmtownsend this PR is ready for review.

crates/host/src/wasmbus/mod.rs
```diff
@@ -972,7 +980,7 @@ impl Host {
             "instantiating component"
         );
 
-        let max_execution_time = self.max_execution_time;
+        let max_execution_time = self.max_execution_time; // TODO: Needs approval to go ahead.
```
brooksmtownsend, line 983: Is this comment still relevant?

crates/runtime/src/component/mod.rs
```diff
@@ -150,6 +153,58 @@ pub struct ComponentConfig {
...
+    pub max_execution_time: Option<u64>,
+}
+impl Limits {
+    /// Converts limits to a string-based key-value map for serialization.
+    ///
+    /// Only includes fields that have values (non-None). Used for converting
+    /// to external configuration formats.
+    ///
+    /// # Returns
+    /// `HashMap` containing string representations of set limits
```
brooksmtownsend, line 170: With this, I think all you would need to be able to just derive `Serialize` and not need the `to_string_map` is to annotate the fields on the Limits struct with `#[serde(skip_serializing_if = "Option::is_none")]`, is there another use case that this is for?

crates/runtime/src/component/mod.rs
```diff
@@ -909,3 +1030,85 @@ impl<H: Handler> Ctx<H> {
...
+            true
+        };
+        Ok(can_grow)
+    }
+}
+/// Async implementation of wasmtime's [`StoreLimits`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/limits.rs)
+/// Used to limit the memory use and table size of each Instance
+impl StoreLimitsAsync {
+    #[must_use]
+    /// TODO
```
brooksmtownsend: is this still relevant?

crates/control-interface/src/client.rs
```diff
@@ -289,12 +289,14 @@ impl Client {
...
     ///
     #[instrument(level = "debug", skip_all)]
+    #[allow(clippy::too_many_arguments)]
     pub async fn scale_component(
         &self,
         host_id: &str,
         component_ref: &str,
         component_id: &str,
         max_instances: u32,
+        component_limits: Option<HashMap<String, String>>,
```
brooksmtownsend: I know it's annoying, but I would keep this off of the function here. Changing the argument here makes this a semver major breaking change, but adding the method to the builder is a backwards compatible new feature.

brooksmtownsend (changes requested):
This all looks solid! I only had comments regarding some leftover comments, but once that's resolved this looks good to go 😄 

The only request here is to remove that one argument change from the `scale_component` function in favor of only supporting this new argument via the builder pattern.

crates/control-interface/src/client.rs
Aditya1404Sal: I totally understand, This argument _was_ something I had been debating whether to include or not, I'll shift it over to the builder instead 🫡

crates/runtime/src/component/mod.rs
Aditya1404Sal: I kept `StoreLimitsAsync` because, _theoretically_, it could be used to adjust memory limits after instantiation. However, unless we plan to support soft limit overrides for a component via wash commands, I think it’s safe to remove it.

As far as I know, the limitation of this approach is that it only reduces the memory accessible to the component, while the total allocated memory remains fixed as per what’s configured in wasmtime::Engine. So, it’s probably best not to include this for now.

crates/host/src/wasmbus/mod.rs
Aditya1404Sal: Yes, in a way. Since we're now allowing component-level limits with `max_execution_time`, would it be appropriate to switch to using the custom limit here instead of the current runtime-specified value? Or would it be better to address that in a separate pull request?

crates/runtime/src/component/mod.rs
Aditya1404Sal: We need a `HashMap<String, String>` for the `.limits()` field of the `ComponentDescription` builder, not a serialized struct. That’s why converting `Limits` to a map is necessary here. I’m mainly following the `ComponentDescription` convention of using flattened data structures for builder fields.

brooksmtownsend: Ah! Ok, thanks

crates/host/src/wasmbus/mod.rs
brooksmtownsend: ah ok I just misunderstood the comment, proceed!

brooksmtownsend (approved):
Great work here @Aditya1404Sal , this looks really nice and backwards compatible!

PR #4549, opened 2025-06-11
feat(policy): add cache to policy manager for responses

JeancarloBarrios:
This is just a draft of an Idea for the issue wasmCloud#3950. 

This is not an implementation its just to commucate the idea.

### DIFF
--- a/crates/host/src/nats/policy.rs
+++ b/crates/host/src/nats/policy.rs
@@ -5,6 +5,7 @@ use core::time::Duration;
 
 use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
+use std::time::Instant;
 
 use anyhow::Context;
 use futures::{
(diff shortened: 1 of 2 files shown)


crates/host/src/nats/policy.rs
```diff
@@ -24,14 +25,119 @@ use crate::policy::{
...
+    V: std::fmt::Debug + Clone + Send + Sync + 'static,
+{
+    fn new(ttl: Duration) -> Self {
+        Self::with_cleanup(ttl, ttl / 2)
+    }
+
+    fn with_cleanup(ttl: Duration, interval: Duration) -> Self {
+        let cache = Arc::new(RwLock::new(HashMap::<K, CacheEntry<V>>::new()));
+        let cleanup_cache = Arc::clone(&cache);
+        let cleanup_handle = spawn(async move {
```
thomastaylor312: This is going to be pretty write heavy which could cause contention, especially under a large amount of load since it scans _all_ cached responses every time it ticks. I think there are two approaches here. If you want to roll it by hand, you can combine "lazy" expiration (checking the TTL on a `get` and removing if expired) with a cleanup loop that runs less often (once every few minutes) to clean up expired entries.

The easier option might be to use the [`DelayQueue`](https://docs.rs/tokio-util/latest/tokio_util/time/struct.DelayQueue.html) from tokio_util. It has an example in there of a simple TTL cache there that is fairly helpful. That is probably your best option since `DelayQueue` uses the same stuff under the hood as tokio and is fairly well-tested.

thomastaylor312 (commented):
Like the general idea here! I think the cache could live outside of NATS so it could be used by any policy service and I left a note on some possible performance improvements. Definitely would like to have a TTL cache if you'd like to address the performance things I mentioned

crates/host/src/nats/policy.rs
JeancarloBarrios: I followed your recommendation and used the DelayQueue. I also read a bit, and it seems that it tends to be more performant to do it based on cycles rather than time. So, I still expect entry based on time, but I only triggered the purge after a certain number of write cycles. lmk your thoughts

JeancarloBarrios: Additionally, I'm not entirely sure which are good default values; if you have any recommendations, I would appreciate them.

JeancarloBarrios:
@thomastaylor312 hey gentle ping on this one

JeancarloBarrios:
I'm unsure how to resolve this CI error. Would you happen to have any insight? @thomastaylor312 I'm unsure how to replicate it locally.

crates/host/src/cache.rs
```diff
@@ -0,0 +1,588 @@
...
+impl<K, V> Clone for Cache<K, V>
+where
+    K: std::hash::Hash + Eq + Clone + std::fmt::Debug + Send + Sync + 'static,
+    V: std::fmt::Debug + Clone + Send + Sync + 'static,
+{
+    fn clone(&self) -> Self {
+        Self {
+            internal: Arc::clone(&self.internal),
+            purge_cycles: self.purge_cycles,
+            purge_cycle_count: AtomicU32::new(self.purge_cycle_count.load(Ordering::Relaxed)),
```
thomastaylor312: The `purge_cycle_count` should probably be in an `Arc` so it is cloned and shared as well. Otherwise, clones of the cache will be purging on different cycles

crates/host/src/cache.rs
```diff
@@ -0,0 +1,588 @@
...
+    }
+
+    /// Set the number of operations after which to run a purge cycle
+    /// If None, purging only happens on explicit timer expiration
+    pub fn with_purge_cycles(mut self, cycles: u32) -> Self {
+        self.purge_cycles = Some(cycles);
+        self
+    }
+
+    /// Disable automatic purge cycles (default behavior)
```
thomastaylor312: For the basic use case here, I think it might be better to just default to having an automatic purge (or not even have it configurable). My main concern is that you always create and insert into the `DelayQueue` even if you're not setting up auto purging. I could be misunderstanding why you chose this as the default though.

Any thoughts?

crates/host/src/cache.rs
JeancarloBarrios: You mean if it's modified? Later on? But sure, I can update it, being a u32, it copies right? But you are right, if anyone modifies it, it will stay consistent fixing....!!!

crates/host/src/cache.rs
JeancarloBarrios: I chose that as a default behavior because I was adding extra functionality from our previous discussion, though it was not necessarily needed. But I agree with you. Now, about keeping it configurable, it might be better given that the ideal number of cycles may vary by use case.

JeancarloBarrios: What it's a good purge cycledefaultl val,I'llill put 1000 ,for now but lmk

crates/host/src/cache.rs
thomastaylor312: Yeah if you `Arc` the `AtomicU32` it will just share the same underlying atomic type

crates/host/src/cache.rs
```diff
@@ -0,0 +1,581 @@
...
+}
+
+/// A thread-safe, TTL-based cache with configurable purging strategies
+pub struct Cache<K, V>
+where
+    K: std::hash::Hash + Eq + Clone + std::fmt::Debug + Send + Sync + 'static,
+    V: std::fmt::Debug + Clone + Send + Sync + 'static,
+{
+    internal: Arc<RwLock<CacheInternal<K, V>>>,
+    purge_cycles: Arc<Option<u32>>,
```
thomastaylor312: Sorry @JeancarloBarrios, I think my comment wasn't clear. This line doesn't need to be `Arc` as copying a u32 is pretty much nothing. I was actually talking about changing `purge_cycle_count` to be `Arc<AtomicU32>`. That way if a cache is cloned, the purge cycle count stays the same across all cloned caches

thomastaylor312 (changes requested):

crates/host/src/cache.rs
JeancarloBarrios: No, you were clear. Idk what I was thinking. I was distracted.  @thomastaylor312

PR #4554, opened 2025-06-13
feat(sqldb-postgres)!: add support for transactions

brooksmtownsend:
## Feature or Problem
This PR adds support for preparing transactions in `wasmcloud:postgres` by creating a transaction resource, repeatedly inserting queries, and then calling `commit()` to execute all queries as a transaction (or rollback with failure)

This PR is marked as draft because I was having trouble testing this, unrelated to transactions.

## Release Information
`wasmcloud:postgres@0.2.0-draft`: We may want to consider the best way to make this backwards compatible with the previous version of postgres, 0.1.1. Just want to validate this before we ship and break consumers of <0.2.0.

Postgres provider next minor version.

## Consumer Impact
No breaking changes, transactions support available

### Manual Verification
It compiles!


crates/provider-sqldb-postgres/src/lib.rs
```diff
@@ -370,6 +377,187 @@ impl bindings::prepared::Handler<Option<Context>> for PostgresProvider {
...
+        for (query, params) in self.queries {
+            // NOTE: Transactions that fail are automatically rolled back
+            let statement = tx
+                .prepare_cached(&query)
+                .await
+                .context("failed to prepare statement in transaction, rolling back")?;
+            tx.execute_raw(&statement, params)
+                .await
+                .context("failed to execute query in transaction, rolling back")?;
+        }
```
vados-cosmonic, line 412: In PG you can actually normally execute a transaction and *get the result* out, as a transaction is going (i.e. against the state of the transaction in progress) -- it would be nice to expose that ability here, and have commits just commit the existing transaction.

crates/provider-sqldb-postgres/src/lib.rs
```diff
@@ -370,6 +377,187 @@ impl bindings::prepared::Handler<Option<Context>> for PostgresProvider {
     }
 }
 
+pub struct TransactionManager {
+    /// The resource that this transaction manager is associated with. This must
+    /// be stored for the lifetime of the transaction and then dropped when the transaction is committed or rolled back.
+    #[allow(unused)]
+    resource: ResourceOwn<Transaction>,
+    queries: Vec<(String, Vec<PgValue>)>,
```
vados-cosmonic, line 385: We shouldn't need to store queries run against the transaction except as a cache for statement prepares I think

crates/provider-sqldb-postgres/src/lib.rs
```diff
@@ -370,6 +377,187 @@ impl bindings::prepared::Handler<Option<Context>> for PostgresProvider {
...
+impl bindings::transaction::Handler<Option<Context>> for PostgresProvider {}
+/// Implement the `wasmcloud:postgres/prepared` interface for [`PostgresProvider`]
+impl bindings::transaction::HandlerTransaction<Option<Context>> for PostgresProvider {
+    #[instrument(level = "debug", skip_all)]
+    async fn new(&self, ctx: Option<Context>) -> anyhow::Result<ResourceOwn<Transaction>> {
+        propagate_trace_for_ctx!(ctx);
+
+        let Some(ctx) = ctx else {
+            bail!("unexpectedly missing invocation context");
+        };
```
vados-cosmonic, line 429:
```suggestion
        let ctx = ctx.context("unexpectedly missing invocation context")?;
```

crates/provider-sqldb-postgres/src/lib.rs
```diff
@@ -370,6 +377,187 @@ impl bindings::prepared::Handler<Option<Context>> for PostgresProvider {
...
+
+        let Some(source_id) = ctx.component else {
+            bail!("unexpectedly missing source ID");
+        };
```
vados-cosmonic, line 433:
```suggestion
        let source_id = ctx.component.context("unexpectedly missing source ID")?;
```

crates/provider-sqldb-postgres/wit/deps/postgres/transaction.wit
```diff
@@ -0,0 +1,19 @@
+package wasmcloud:postgres@0.2.0-draft;
```
vados-cosmonic, line 1: I'm fine with a minor version bump here but want to note that since we're purely adding functionality here, I think we can do a minor bump with `since` feature flags. 

I'm not sure anything changes with regard to the provider, because if the provider isn't listening on the right endpoints, it just doesn't respond anyway/the wRPC request would time out IIRC.

vados-cosmonic (commented):
Excited to have this feature land! The code overall looks good, but it would be nice to get access to the query results from the execution during the transaction.

crates/provider-sqldb-postgres/src/lib.rs
brooksmtownsend: Yeah that's what I was hoping too, unfortunately the little note above on 381/382 details the result 😢 I couldn't find a way to get the prepared statement + transaction to stick around for the life of the transaction since it requires holding a pool connection the entire time

vados-cosmonic: Ah, so postgres can only have one transaction going on a connection a time IIRC so is it possible to actually take the connection out of the pool? 

It looks like you can [`detach` it](https://docs.rs/deadpool/0.12.2/deadpool/managed/trait.Manager.html#method.detach), but let's maybe file this for a future follow up!

Issue #4372, opened 2025-04-16
[FEATURE] Replace use of `~/.wash` with XDG Base Directory Specification


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [x] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

We've maintained a $HOME/.wash directory for quite a while, just to ensure we always know where wash, its generated keys, and downloaded binaries are. However, this clutters the home directory of the user and doesn't integrate well with existing tools.

### Describe the solution you'd like

I'd like to see wash adopt the XDG base directory spec, which defines different directories for config (~/.config) cache (~/.cache) and data (depends on platform).

I think we should use the [XDG strategy](https://docs.rs/etcetera/0.10.0/etcetera/app_strategy/struct.Xdg.html) from the current library we use, etcetera, to determine the config/cache/data directories.

Given the current structure of the `~/.wash` directory, I propose we make the following changes
```bash
drwxr-xr-x     - brooks  2 Dec  2024  contexts/
drwxr-xr-x     - brooks 15 Apr 21:03  dev/
drwxr-xr-x     - brooks 15 Apr 16:29  downloads/
drwx------@    - brooks  6 Apr 10:06  keys/
drwxr-xr-x     - brooks 25 Mar 13:50  package_cache/
drwxr-xr-x     - brooks 16 Dec  2024  plugins/
.rw-r--r--     0 brooks 23 Feb  2024  .completion_suggested
.rw-r--r--  229k brooks 22 Aug  2024  _wash
.rw-r--r--   47k brooks 27 Sep  2024  launch.log
```

1. Contexts and configuration go into the `XDG_CONFIG_HOME` directory
2. Downloads, dev logs, keys and plugins go into `XDG_DATA_HOME`
3. Package cache, temporary jetstream directories go into `XDG_CACHE_HOME`

Importantly, we should also be able to migrate existing `~/.wash` directories into the new places if the older directory exists.

### Describe alternatives you've considered

Instead of using XDG we could instead use platform specific conventions, or a different strategy with `etcetera`. I don't really have a hard preference here, but given we could hook into the existing spec and respect the environment variables when set I figured the XDG path would be the best supported.

### Additional context

[...]

PR #4558, opened 2025-06-16
replace use of `~/.wash` with XDG Base Directory Specification

f4z3r:
## Feature or Problem

The current implementation does not really respect best practices on where data should be stored (whether config data, cache data, runtime data, etc). This is an attempt to support best practices using XDG directories on Linux and Mac, and using the standard Windows scheme on Windows. This also slightly refactors the way these directories are accessed from other packages to simplify their usage.

Goal of this PR would be:

- Implement best practices.
- Keep possibility for configuring everything in a fine grained manner as described in the related story.
- Simplify handling on config directories, ensuring they are created upon use.
- Unify the handling of files in the configuration directories a little.

## Related Issues

- #4372

## Release Information

I guess next? This is on the Q2 Roadmap in status "Ready for Work".

## Consumer Impact

This might be a breaking change. With the default configuration, this would move the configuration directories away from `~/.wash` to several other directories:

- `~/.config/wash/`
- `~/.local/share/wash/`
- `~/.cache/wash/`

One can use the env vars described in the issue (`WASH_DOWNLOADS_DIR`, etc) to make `wash` behave as it did before (store everything in `~/.wash/`) but it won't do so by default.

I would very much appreciate a pointer here from the maintainers. Is this a big deal? Or are you fine with having this breaking change. I don't really see how this should be implemented without a breaking change to be honest.

### Unit Test(s)

- Added a small unit test to validate that the precedence of env vars is respected.

### Acceptance or Integration

- Updated integration tests to use new directory structure.

#### Wash Dev Test

I cleared the `~/.wash` directory, and ran a `wash dev` on the http-jsonify example:

```
$ tree ~/.local/share/wash
/home/f4z3r/.local/share/wash
├── dev
│   ├── 7JdIHS
│   │   ├── nats.log
│   │   ├── wadm.log
│   │   └── wasmcloud.log
│   └── wash-dev-sessions.json
├── downloads
│   ├── nats.conf
│   ├── nats.pid
│   ├── nats-server
│   ├── v1.8.0
│   │   └── wasmcloud_host
│   ├── wadm
│   └── wadm.pid
├── keys
│   ├── f4z3r_account.nk
│   └── http_jsonify_module.nk
└── plugins

7 directories, 12 files

$ tree ~/.config/wash
/home/f4z3r/.config/wash

0 directories, 0 files

$ tree ~/.cache/wash
/home/f4z3r/.cache/wash
└── package_cache
    ├── sha256:5a568e6e2d60c1ce51220e1833cdd5b88db9f615720edc762a9b4a6f36b383bd
    ├── sha256:a1f129cdf1fde55ec2d4ae8d998c39a7e5cf7544a8bd84a831054ac0d2ac64dd
    ├── wasi:http-0.2.0.json
    └── wasi:http-0.2.2.json

2 directories, 4 files
```

#### Wash Up Test

I ran `wash up` to validate where contexts are being written:

```
$ tree ~/.config/wash
/home/f4z3r/.config/wash
└── contexts
    ├── default
    └── host_config.json

2 directories, 2 files
```

#### Test Support for Old Configuration

I wanted to test how easy it would be to keep the old structure:

```
$ rm -rf ~/.local/share/wash ~/.config/wash ~/.cache/wash ~/.wash
$ export WASH_DATA_DIR=~/.wash WASH_CONFIG_DIR=~/.wash WASH_CACHE_DIR=~/.wash
$ wash dev
$ wash up
$ tree ~/.config/wash
/home/f4z3r/.config/wash  [error opening dir]

0 directories, 0 files

$ tree ~/.cache/wash
/home/f4z3r/.cache/wash  [error opening dir]

0 directories, 0 files

$ tree ~/.local/share/wash
/home/f4z3r/.local/share/wash  [error opening dir]

0 directories, 0 files

$ tree ~/.wash
/home/f4z3r/.wash
├── contexts
│   ├── default
│   └── host_config.json
├── dev
│   ├── wash-dev-sessions.json
│   └── xoNnEE
│       ├── nats.log
│       ├── wadm.log
│       └── wasmcloud.log
├── downloads
│   ├── nats.conf
│   ├── nats.log
│   ├── nats.pid
│   ├── nats-server
│   ├── v1.8.0
│   │   └── wasmcloud_host
│   ├── wadm
│   └── wadm.log
├── keys
│   ├── f4z3r_account.nk
│   └── http_jsonify_module.nk
├── package_cache
│   ├── sha256:5a568e6e2d60c1ce51220e1833cdd5b88db9f615720edc762a9b4a6f36b383bd
│   └── wasi:http-0.2.0.json
└── plugins
```

### DIFF
--- a/crates/wash/src/bin/wash.rs
+++ b/crates/wash/src/bin/wash.rs
@@ -7,6 +7,7 @@ use anyhow::{bail, Context};
 use clap::{self, Arg, ArgMatches, Command, FromArgMatches, Parser, Subcommand};
 use console::style;
 use crossterm::style::Stylize;
+use etcetera::AppStrategy;
 use semver::Version;
 use serde_json::json;
 use tracing_subscriber::EnvFilter;
@@ -489,6 +490,17 @@ async fn main() {
             eprintln!("Error while checking for new wash version: {e}");
             std::process::exit(2);
         }
+    }
+
+    let old_config_dir = WASH_DIRECTORIES.home_dir().join(".wash");
+    if tokio::fs::try_exists(&old_config_dir)
+        .await
+        .unwrap_or(false)
+    {
+        eprintln!(
+            "Old configuration directory '{}' found, consider migrating to new configuration structure.",
+            old_config_dir.display(),
+        );
     }
 
     let output_kind = cli.output;
--- a/crates/wash/src/bin/wash.rs
+++ b/crates/wash/src/bin/wash.rs
@@ -23,6 +23,7 @@ use wash::lib::cli::start::StartCommand;
 use wash::lib::cli::stop::StopCommand;
 use wash::lib::cli::update::UpdateCommand;
 use wash::lib::cli::{CommandOutput, OutputKind};
+use wash::lib::config::WASH_DIRECTORIES;
 use wash::lib::drain::Drain as DrainSelection;
 use wash::lib::generate::emoji;
 use wash::lib::plugin::subcommand::{DirMapping, SubcommandRunner};
@@ -49,7 +50,6 @@ use wash::cli::plugin::{self, PluginCommand};
 use wash::cli::secrets::{self, SecretsCliCommand};
 use wash::cli::style::WASH_CLI_STYLE;
 use wash::cli::ui::{self, UiCommand};
-use wash::cli::util::ensure_plugin_dir;
 
 #[derive(Clone)]
 struct HelpTopic {
@@ -745,7 +745,7 @@ fn experimental_error_message(command: &str) -> anyhow::Result<CommandOutput> {
 async fn load_plugins() -> Option<(SubcommandRunner, PathBuf)> {
     // We need to use env vars here because the plugin loading needs to be initialized before
     // the CLI is parsed
-    let plugin_dir = match ensure_plugin_dir(std::env::var("WASH_PLUGIN_DIR").ok()).await {
+    let plugin_dir = match WASH_DIRECTORIES.create_plugins_dir() {
         Ok(dir) => dir,
         Err(e) => {
             tracing::error!(err = ?e, "Could not load wash plugin directory");
--- a/crates/wash/src/lib/cli/mod.rs
+++ b/crates/wash/src/lib/cli/mod.rs
@@ -26,7 +26,7 @@ use wasm_pkg_client::{
 
 use crate::lib::{
     config::{
-        cfg_dir, WashConnectionOptions, DEFAULT_LATTICE, DEFAULT_NATS_HOST, DEFAULT_NATS_PORT,
+        self, WashConnectionOptions, DEFAULT_LATTICE, DEFAULT_NATS_HOST, DEFAULT_NATS_PORT,
         DEFAULT_NATS_TIMEOUT_MS,
     },
     context::{default_timeout_ms, fs::ContextDir, ContextManager},
@@ -308,7 +308,7 @@ impl CommonPackageArgs {
             }
             // Otherwise we got nothing and attempt to load the default config locations
             (None, None) => {
-                let path = cfg_dir()?.join("package_config.toml");
+                let path = config::cfg_dir()?.join("package_config.toml");
                 // Check if the config file exists before loading so we can error properly
                 if tokio::fs::metadata(&path).await.is_ok() {
                     let loaded = wasm_pkg_client::Config::from_file(&path)
@@ -354,7 +354,7 @@ impl CommonPackageArgs {
             // We have a cache dir provided by the user via `WKG` env var
             (None, Some(path)) => PathBuf::from(path),
             // Otherwise we got nothing and attempt to load the default cache dir
-            (None, None) => cfg_dir()?.join("package_cache"),
+            (None, None) => config::WASH_PACKAGE_CACHE_DIR.to_path_buf(),
         };
         FileCache::new(dir).await
     }
@@ -502,7 +502,7 @@ fn determine_directory(directory: Option<PathBuf>) -> Result<PathBuf> {
     if let Some(d) = directory {
         Ok(d)
     } else {
-        let d = cfg_dir()?.join("keys");
+        let d = config::WASH_KEYS_DIR.to_path_buf();
         Ok(d)
     }
 }
@@ -574,7 +574,7 @@ mod test {
     use serial_test::serial;
 
     use crate::lib::{
-        config::{WashConnectionOptions, DEFAULT_CTX_DIR_NAME, DEFAULT_LATTICE, WASH_DIR},
+        config::{WashConnectionOptions, DEFAULT_CTX_DIR_NAME, DEFAULT_LATTICE},
         context::{fs::ContextDir, ContextManager, WashContext},
     };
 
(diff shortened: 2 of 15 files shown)


crates/wash/src/lib/config.rs
```diff
@@ -21,40 +21,326 @@ pub const DEFAULT_NATS_TIMEOUT_MS: u64 = 2_000;
...
+    pub fn create_in_keys_dir<P: AsRef<OsStr>>(
+        &self,
+        subdir: Option<P>,
+        file: Option<P>,
+    ) -> Result<PathBuf> {
+        create_in_dir_method!(self, subdir, file, keys_dir, in_keys_dir)
+    }
+
+    /// Returns the path to the directory where to store logs.
+    pub fn logs_dir(&self) -> PathBuf {
```
f4z3r, line 136: - [ ] Currently these functions are clearly not used everywhere, as the log directory for instance did not exist before. We will need to move the logs into this directory.

crates/wash/src/lib/config.rs
```diff
@@ -21,40 +21,326 @@ pub const DEFAULT_NATS_TIMEOUT_MS: u64 = 2_000;
...
+macro_rules! create_dir_method {
+    ($self: ident, $dir_method_name: ident) => {{
+        let path = $self.$dir_method_name();
+        fs::create_dir_all(&path)
+            .with_context(|| format!("failed to create directory `{}`", path.display()))?;
+        Ok(path)
+    }};
+}
+
+macro_rules! create_in_dir_method {
```
f4z3r: I don't know what your opinions are on macros here. Also, I currently implement many utility functions to make sure the interface for the vast majority of these directories is the same. However this also means that many of these functions are actually not called anywhere.

crates/wash/src/lib/config.rs
brooksmtownsend: No problem, I don't think we need to do any work to migrate old logs for example they are primarily throwaway 😄

crates/wash/src/lib/config.rs
brooksmtownsend: No strong opinions on macros, but _generally_ I would prefer methods on structs / functions if a macro isn't necessary since it's a bit easier to follow 😄

crates/wash/src/lib/config.rs
f4z3r: hahaha I did not mean migrations of old logs. But I think the `launch.log` is currently not being written in the new `logs` directory. I still need to figure out where that log initially originates and update the code. Also, the issue says it wants the directory to be `$WASH_DATA_DIR/logs/`. Per XDG logs should be under `$XDG_STATE_HOME`, and typically log directories are called `log` and not `logs`.

I would rather suggest that logs follow either of the following conventions:
- `$XDG_STATE_HOME/wash/launch.log` = `~/.local/state/wash/launch.log`
- `$XDG_STATE_HOME/wash/log/launch.log` = `~/.local/state/wash/log/launch.log`

Of course, one could keep the idea that `$WASH_LOGS_DIR` is `$XDG_STATE_HOME/wash[/log]` and use that. What do you think?

crates/wash/src/lib/config.rs
f4z3r: Ok, I removed them to simplify the code (90a33c3e8a08fccd888b7dc302612bd8acfe6500). This leads to quite a bit of duplication, but the code is definitely easier to understand this way.

crates/wash/src/lib/config.rs
f4z3r: Mhhh did not find any part in the code that actually writes that `launch.log` file. The only logs I can see with some primitive testing are the logs being generated by `wash dev` which are put into the `/dev` sub-directory which makes sense in my opinion.

f4z3r: Maybe it would also make sense to move `/dev` into the XDG state directory rather than in the `data` directory. Typically the information in `dev` should be persisted after a `wash dev` run, but is not critical information I would think (since it is mostly used for debugging). According to https://specifications.freedesktop.org/basedir-spec/latest/#variables `XDG_STATE_HOME` would be more appropriate.

crates/wash/src/lib/cli/mod.rs
```diff
@@ -797,6 +801,41 @@ mod test {
         );
     }
 
+    #[tokio::test]
```
f4z3r, line 804: Not sure this is the correct space for the tests. I am a bit confused as to why the config tests are placed here, but I wanted to reuse `EnvVar`. At first I thought about factoring it out into some test utilities such that I could place the config directory tests directly into the `lib::config` module. But there I was also unsure where to put the test utilities then 🤣 So I just added it here for now. If you want it elsewhere, I would be happy about some pointers.

f4z3r:
Tests are flaky I guess. The `workload_identity_component_interface` test should not be affected by this PR in any way.

brooksmtownsend:
@f4z3r awesome work on this! I think you did an awesome job, planning on giving it a more thorough review this week.

I think the only thing that actually might be worth doing is adding a warning log if `~/.wash` exists and suggest moving to the new directories. I think the only ones that you would need to mention are contexts and keys, since that actually contains custom context and irreplaceable keys, respectively.

crates/wash/src/lib/cli/mod.rs
brooksmtownsend: I think this looks reasonable and follows our current conventions!

Testing environment variables are hard, so since this is `#[serial]` that'll cover the edge cases.

crates/wash/src/lib/config.rs
```diff
@@ -21,40 +21,307 @@ pub const DEFAULT_NATS_TIMEOUT_MS: u64 = 2_000;
...
-pub fn cfg_dir() -> Result<PathBuf> {
-    let home = etcetera::home_dir().context("no home directory found. Please set $HOME")?;
+pub static WASH_DIRECTORIES: Lazy<WashAppStrategy> = Lazy::new(|| {
+    let args = AppStrategyArgs {
+        top_level_domain: "com".to_string(),
+        author: "wasmCloud".to_string(),
+        app_name: "wash".to_string(),
+    };
+
+    let strategy = app_strategy::choose_app_strategy(args).unwrap();
```
brooksmtownsend: would prefer an `expect` here but honestly I can't see it ever failing

crates/wash/src/lib/config.rs
brooksmtownsend: I think that's a good callout, I think for now we can stick with the simpler layout if you think that's sufficient. I don't want to throw wash files all over the place, but having config, data, and cache seem reasonable.

I think we do want to keep dev logs around after runs, but reasonably we should probably clean them up after a certain number of sessions.

crates/wash/src/lib/config.rs
f4z3r: Will update, including the warning log, tomorrow.

f4z3r: Sorry, didn't have time today. Will have a look on Monday.

brooksmtownsend: no worries!

crates/wash/src/bin/wash.rs
```diff
@@ -491,6 +492,17 @@ async fn main() {
         }
     }
 
+    let old_config_dir = WASH_DIRECTORIES.home_dir().join(".wash");
+    if tokio::fs::try_exists(&old_config_dir)
+        .await
+        .unwrap_or(false)
+    {
+        eprintln!(
+            "Old configuration directory '{}' found, consider migrating to new configuartion structure.",
```
brooksmtownsend:
```suggestion
"Old configuration directory '{}' found, consider migrating to new configuration structure.",
```

brooksmtownsend (approved):
thank you @f4z3r , just a typo to resolve and we'll merge!

PR #4559, opened 2025-06-16
ci(nix): switch from vault to vault-bin package to speed-up flake

f4z3r:
## Feature or Problem

Using vault-bin instead of vault allows to skip compiling the entirety of Vault when evaluating the flake. This reduces the flake devShell initial evaluation by around 5 minutes with barely any downside (from locally compiling the binary).

## Consumer Impact

None.

### Manual Verification

I test built stuff like `nix build .#keyvalue-vault-provider-oci`, but I am guessing the Vault binary is mostly used for integration testing. I don't think I can "test" this locally very well.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -157,7 +157,7 @@
                   pkgs.redis
                   pkgs.spire-agent
                   pkgs.spire-server
-                  pkgs.vault
+                  pkgs.vault-bin
 
                   pkgs.pkgsUnstable.nats-server
                 ];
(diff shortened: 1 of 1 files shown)


rvolosatovs (approved):
Generally, I'd expect `vault` to be cached upstream and it's likely we're getting a miss because I was a bit behind with updating to 25.05 in `nixify` https://github.com/rvolosatovs/nixify/pull/382 (finally merged yesterday)

Pulling the official binary makes sense to me, however, thanks!

Issue #3823, opened 2024-12-12
[FEATURE] Get host_id in more consistent manner


lachieh:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [X] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

When `wash dev` starts a host, the only way to determine the host_id is to follow and parse the log.

### Describe the solution you'd like

It would be better if the host could expose an API that allowed wash to determine the host_id another way.

### Describe alternatives you've considered

I am not sure of how this should happen, and @vados-cosmonic said that other suggestions have also been put forward:
- Allow providing keys to the host (allowing us to output the public key from wash up directly)
- Enable writing out the generated key to somewhere on disk

### Additional context

Spurred from discussion in #3820

brooksmtownsend:
> Allow providing keys to the host (allowing us to output the public key from wash up directly)

This would be super easy and we can do this! wasmCloud supports a `--host-seed` / `WASMCLOUD_HOST_SEED` argument which we can supply the private key piece from `nkeys::KeyPair::new_server()` to, using the public key as the host ID 😄

brooksmtownsend:
This is a really great first issue, all we need to do here is generate an Nkey server key and pass that seed to the host with `--host-seed` during `wash dev`! Then we know the public key side of it

PR #4568, opened 2025-06-18
fix(wash): pass host key seed to infer host id rather than read from log

f4z3r:
## Feature or Problem

Wash reads the host ID from the hosts it starts with `dev` from the log file. This is not very nice. Instead, we are passing a key seed to infer the ID the host will use. This avoids having to go through logs.

## Related Issues

- #3823

## Consumer Impact

None

### Unit Test(s)

- Removed the tests that checked the parsing of the logs.

### Acceptance or Integration

None

### Manual Verification

- Ran `wash dev` a couple of times. Validated that the ID in the logs matches the one inferred by the code (by additionally logging it to stdout).

### DIFF
--- a/crates/wash/src/cli/cmd/dev/mod.rs
+++ b/crates/wash/src/cli/cmd/dev/mod.rs
@@ -255,6 +255,7 @@ pub async fn handle_command(
     };
 
     // See if the host is running by retrieving an inventory
+    std::thread::sleep(std::time::Duration::from_secs(1));
     if let Err(_e) = ctl_client.get_host_inventory(&host_id).await {
         eprintln!(
             "{} Failed to retrieve inventory from host [{host_id}]... Exiting developer loop",
--- a/crates/wash/src/cli/cmd/dev/mod.rs
+++ b/crates/wash/src/cli/cmd/dev/mod.rs
@@ -17,6 +17,8 @@ use crate::lib::id::ServerId;
 use crate::lib::parser::load_config;
 use tracing::trace;
 
+use wasmcloud_control_interface::Client;
+
 use crate::cmd::up::{
     nats_client_from_wasmcloud_opts, remove_wadm_pidfile, NatsOpts, WadmOpts, WasmcloudOpts,
 };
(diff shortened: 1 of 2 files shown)


crates/wash/src/cli/cmd/dev/mod.rs
```diff
@@ -255,6 +255,7 @@ pub async fn handle_command(
     };
 
     // See if the host is running by retrieving an inventory
+    std::thread::sleep(std::time::Duration::from_secs(1));
```
f4z3r: Currently this runs into a timing issue on the function call below. I am guessing because inferring the ID is much faster than parsing logs, and that the host is not yet up by the time this function call is reached. It works with a sleep (a much smaller sleep is enough).

Is there a simple way to check whether a host is up? Or should I just implement a retry loop on the inventory retrieval?

f4z3r: I implemented a small function. Not sure this is the smartest way to go.

crates/wash/src/cli/cmd/dev/mod.rs
```diff
@@ -493,3 +502,29 @@ async fn stop_dev_session(
 
     Ok(())
 }
+
+/// Checks whether a wasmCloud host is up by trying to get its inventory. It will repeat after
+/// waiting for `backoff` duraction in case of error. It will fail after `timeout` duration if the
+/// inventory could still not be retrieved.
+pub async fn is_host_up(
```
brooksmtownsend, line 509: I have a slight preference for a simple loop here over the inline function, but it's ok as-is

brooksmtownsend (approved):
Looks good @f4z3r ! I left one comment that's optional, I think you did a good job here. Not being able to know for sure when the host is started is a little frustrating

crates/wash/src/cli/cmd/dev/mod.rs
brooksmtownsend: I'm going to go ahead and merge this, and if you feel the urge it can be a followup 😄

Issue #3918, opened 2025-01-02
[FEATURE] Update custom capability provider go component to use component SDK


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [X] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Under https://github.com/wasmCloud/wasmCloud/tree/main/examples/golang/providers/custom-template/component, the component is using fairly old bindgen and needs a refresh.

### Describe the solution you'd like

We should update this component to use the component SDK and more recent bindgen

vados-cosmonic:
The project has changed so drastically that this idea is probably not reasonable to pursue in this form -- closing this as we're focusing effort now on wasmCloud v2.

If you think this definitely still applies towards v2, please feel free to open a new issue!

PR #4577, opened 2025-06-19
Update go example provider to go component SDK

f4z3r:
## Feature or Problem

The example of the provider is outdated and should use the new SDK.

## Related Issues

- #3918

## Consumer Impact

None

### Manual Verification

Playthrough of the examples.


examples/golang/providers/custom-template/README.md
```diff
@@ -36,21 +36,33 @@ Prerequisites:
...
+You can also invoke the provider with `wash call`, which in this case will report that we have no linked components:
 
 ```bash
 wash call custom-template wasmcloud:example/system-info.call
 ```
+```
+Error invoking component: timed out invoking component, is component [custom-template] running in lattice [default]?
+```
 
 ## Running as an application
```
f4z3r, line 68: I am very confused. When I try to run with the component, I cannot build the component even in the current state, without updating to the component SDK. The reason is that I get a build error:

```
$ wash build
# github.com/wasmcloud/wasmcloud/examples/golang/providers/custom-template/component
customtemplate.go:14:28: cannot use Process (value of type func(data processdata.Data) string) as func(data processdata.Data) value in assignment
customtemplate.go:20:8: system.RequestInfo(system.KindOS) (no value) used as value
customtemplate.go:21:10: system.RequestInfo(system.KindOS) (no value) used as value

Compiling component failed: exit status: 1
```

Even tough the WIT clearly states that process should return a string (https://github.com/wasmCloud/wasmCloud/blob/main/examples/golang/providers/custom-template/component/wit/deps/wasmcloud-example-0.1.0/package.wit#L13), I get the following code generated by wit-bindgen-go:

`component/gen/wasmcloud/example/process-data.exports.go`:

```go
// Code generated by wit-bindgen-go. DO NOT EDIT.

package processdata

// Exports represents the caller-defined exports from "wasmcloud:example/process-data@0.1.0".
var Exports struct {
	// Process represents the caller-defined, exported function "process".
	//
	// Send structured data to the component for processing
	//
	//	process: func(data: data)
	Process func(data Data)
}
```

Here the Process function does not return a `string`. The same is true for the system info error:

`component/gen/wasmcloud/example/system-info.wit.go`:

```go
// Code generated by wit-bindgen-go. DO NOT EDIT.

// Package systeminfo represents the imported interface "wasmcloud:example/system-info@0.1.0".
//
// While processing data, sometimes a component may need to request
// information about the system it's running on. The component isn't
// allowed to access this information directly, so it can request it
// from the provider.
package systeminfo

// [...]

// RequestInfo represents the imported function "request-info".
//
// Request information about the system the provider is running on
//
//	request-info: func(kind: kind)
//
//go:nosplit
func RequestInfo(kind Kind) {
	kind0 := (uint32)(kind)
	wasmimport_RequestInfo((uint32)(kind0))
	return
}

// [...]
```

Where RequestInfo does not return the `string` declared in the WIT (https://github.com/wasmCloud/wasmCloud/blob/main/examples/golang/providers/custom-template/component/wit/deps/wasmcloud-example-0.1.0/package.wit#L27).

I am super confused about this, and have been trying to figure this out for hours now. @brooksmtownsend do you have a pointer as to what could be the issue? Is `wit-bindgen-go` broken?

brooksmtownsend: Yeah, I'm having trouble finding that issue right now 😢  So sorry you spent hours of time on this

But basically you're going to want to install `wasm-tools 1.223.1`. I _think_ that later versions may fix the issue, but in 1.224 I think there was a regression in compatibility between wit-bindgen-go and wasm-tools and you get functions that are generated without their return value

f4z3r: Alright, I might be able to have another look at this tomorrow. Thanks for the information!

f4z3r: Perfect, I got it to work with wasm-tools 1.223.1. I unfortunately had a thousand things that go me busy lately. But this Friday I will have a look at this for sure. Sorry for not holding true to my promises so far.

f4z3r:
I am still quite confused. I cannot get the call to the provider via `wash call` to work. The way I understand it:

- I perform `wash call custom-template wasmcloud:example/system-info.call`.
- This will get routed to the `call` function from the provider.
- Which calls the `process` function in the component.
- Which calls the `request-info` function on the provider twice.
- and should return `Provider is running on linux-amd64` in my case.

However, when I run the `wash call` command, I get

```
Error invoking component: timed out invoking component, is component [custom-template] running in lattice [default]?
```

The provider and component are running for sure:

```
$ wash get inventory

  Host ID                                                                          Friendly name
  NCO65PMMYWOLYEP2D2FK2HO3QXNWK63ZP3X7G4F4GDIKWG7JM57TSLAW                         dawn-brook-3056

  Host labels
  hostcore.arch                                     x86_64
  hostcore.os                                       linux
  hostcore.osfamily                                 unix

  Component ID                                      Name                           Max count
  custom_template_provider_golang-test_component    custom-template-component      1

  Provider ID                                       Name
  custom-template                                   KeyValue Go
```

If I try to trigger the call directly via wRPC on NATS, I get a nil response:

```
$ nats req "default.custom-template.wrpc.0.0.1.wasmcloud:example/system-info@0.1.0.call" "{}"
12:41:24 Sending request on "default.custom-template.wrpc.0.0.1.wasmcloud:example/system-info@0.1.0.call"
12:41:24 Received with rtt 417.198µs
nil body
```

Yet, if I subscribe to the `_INBOX` on NATS, I can see the correct responses being sent back (disclaimer, I have never used NATS in my life, maybe I am doing something really stupid here):

```
$ nats sub "_INBOX.>"

...

[#1352] Received on "_INBOX.sKrmL6tfted8L8zpNEhO1e.sKrmL6tfted8L8zpNEhO4i.results"
linux

...

[#1356] Received on "_INBOX.sKrmL6tfted8L8zpNEhO7m.sKrmL6tfted8L8zpNEhOAq.results"
amd64

...

[#1361] Received on "_INBOX.N93yzdfiRjXJOIQVuxOZ5R.VL0kslut.results"
"Provider is running on linux-amd64
```

The result: I am very confused. With my limited understanding of NATS and wasmCloud, it seems that the provider/component are correctly calling each other based on the traffic I am seing on NATS. However, I am not served with the response I am expecting. And when I try to use `wash call`, it says the component is not deployed which contradicts everything else I am seeing 🤣

Issue #3327, opened 2024-10-15
[FEATURE] Replace unnecessarily specific `--rpc-*` and `--ctl-*` flags with `--nats` 


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [X] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [ ] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Many wash subcommands make the distinction explicitly on the command line whether you're connecting to a host's RPC connection or a control interface connection. [This is an important security pattern for wasmCloud, and it's essential to be able to partition these connections](https://wasmcloud.com/docs/deployment/security/nats-segmentation). On the CLI, though, this can be cumbersome to remember if you're working with the RPC connection (like `wash call`) or the control interface connection (like `wash link`).

### Describe the solution you'd like

For `wash up`, it's still important to still be able to specify the RPC and control interface connections for the partitioned connection. But for other commands that have to connect to NATS, I think we should consolidate to either `--nats-*` for parity with the host.

```plaintext
      --nats-host <NATS_HOST>
          NATS server host to connect to [env: WASMCLOUD_NATS_HOST=] [default: 127.0.0.1]
      --nats-port <NATS_PORT>
          NATS server port to connect to [env: WASMCLOUD_NATS_PORT=] [default: 4222]
      --nats-jwt <NATS_JWT>
          A user JWT to use to authenticate to NATS [env: WASMCLOUD_NATS_JWT=]
      --nats-seed <NATS_SEED>
          A seed nkey to use to authenticate to NATS [env: WASMCLOUD_NATS_SEED=]
      --nats-creds <NATS_CREDS>
          A NATS credentials file that contains the JWT and seed for authenticating to NATS [env: WASMCLOUD_NATS_CREDS=]
```

### Describe alternatives you've considered

We could instead drop the `nats` prefix, and just setup flags like `--host` and `--port` since there's no other alternative, but this feels unnecessarily minimal to me.

### Additional context

For backwards compatibility, we should support the old flags with aliases for at least a minor version.

ossfellow:
While this makes it easier to differentiate between connections, one question I have is if an explicit reference to NATS will be future proof (i.e. when we make the transport pluggable)?

brooksmtownsend:
[...]

PR #4586, opened 2025-06-21
--nats replaces --rpc and --ctl with backward capability

rabelmervin:
### Fixes [3327](https://github.com/wasmCloud/wasmCloud/issues/3327)

### DIFF
--- a/examples/rust/components/messaging-image-processor-worker/src/processing.rs
+++ b/examples/rust/components/messaging-image-processor-worker/src/processing.rs
@@ -103,7 +103,7 @@ impl ImageProcessingRequest {
         };
 
         // Attempt to lease the task
-        let lease_id = match tasks::lease_task(&task_id, &String::from(WORKER_ID)) {
+        let lease_id = match tasks::lease_task(&task_id, WORKER_ID) {
             Ok(lid) => lid,
             Err(e) => {
                 log(


crates/wash/src/cli/call.rs
```diff
@@ -569,21 +562,18 @@ async fn create_client_from_opts_wrpc(opts: &ConnectionOpts) -> Result<async_nat
...
-        opts.name("wash-cli")
-            .connect(&nats_url)
-            .await
-            .with_context(|| {
-                format!(
-                    "Failed to connect to NATS server {}:{} while creating client",
-                    &host, &port
-                )
-            })?
+        opts.connect(&nats_url).await.with_context(|| {
```
vados-cosmonic: is `opts.name` (from the old code) set somewhere else?

crates/wash/src/cli/call.rs
```diff
@@ -183,26 +180,40 @@ pub async fn handle_command(
...
 
     /// RPC Port for connections, defaults to 4222 for local nats
-    #[clap(short = 'p', long = "rpc-port", env = "WASMCLOUD_RPC_PORT")]
-    rpc_port: Option<String>,
+    #[clap(
+        short = 'p',
+        long = "nats-port",
+        alias = "rpc-port",
+        env = "WASMCLOUD_NATS_RPC_PORT",
+        default_value = "4222"
```
vados-cosmonic:
```suggestion
```

Up until now we haven't had a default value -- could this be the source of the issues in CI?

crates/wash/src/cli/config.rs
```diff
@@ -26,6 +26,8 @@ pub const WASMCLOUD_MAX_EXECUTION_TIME_MS: &str = "WASMCLOUD_MAX_EXECUTION_TIME_
 pub const DEFAULT_MAX_EXECUTION_TIME_MS: &str = "600000";
 
 // NATS RPC connection configuration
+#[deprecated( note = "Older version")]
+pub mod deprecated_rpc {
```
vados-cosmonic: I *think* we can leave out these deprecation notices for now -- I agree with the sentiment but I'm not sure the juice is worth the squeeze here

vados-cosmonic:
Hey @rabelmervin thanks for making changes and getting this PR up to speed -- I'm going to make *just* a few more and we can get it merged!

PR #4588, opened 2025-06-23
chore(runtime): Expose wasmtime linker

lxfontes:
Exposes wasmtime linker & context object so host authors can bring their own WIT interfaces without going through wrpc.
Also changing visibility on ResourceRef as it provides differentiation between local,remote, and built-in artifacts.


brooksmtownsend (approved):
@lxfontes do you have an example of how this is used?

LGTM

PR #4606, opened 2025-06-27
feat(sqldb-postgres): support connection pool sharing

brooksmtownsend:
## Feature or Problem
This PR allows folks who aren't already using a server-side load balancing framework like PgBouncer to do some connection pool sharing at the provider layer.

## Release Information
next minor provider-sqldb-postgres

### DIFF
--- a/crates/provider-sqldb-postgres/src/lib.rs
+++ b/crates/provider-sqldb-postgres/src/lib.rs
@@ -13,6 +13,7 @@ use std::sync::Arc;
 use anyhow::{Context as _, Result};
 use deadpool_postgres::Pool;
 use futures::TryStreamExt as _;
+use sha2::{Digest as _, Sha256};
 use tokio::sync::RwLock;
 use tokio_postgres::types::Type as PgType;
 use tracing::{error, instrument, warn};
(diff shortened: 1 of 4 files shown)


crates/provider-sqldb-postgres/src/lib.rs
```diff
@@ -49,10 +56,26 @@ type StatementParams = Vec<PgType>;
...
+/// This is the hash of the connection configuration, to avoid printing credentials inadvertently.
+type SharedConnectionKey = String;
+
+/// Type of postgres connection - either direct or shared
+#[derive(Clone)]
+enum PostgresConnection {
+    /// Direct connection to a pool
+    Direct(Pool),
+    /// Shared connection, identified by the hash of the connection configuration
+    Shared(String),
```
vados-cosmonic, line 70:
```suggestion
    Shared(SharedConnectionKey),
```

vados-cosmonic: I'd also suggest we make `SharedConnectionKey` an `non_exhaustive` enum -- it's very likely that this scheme might change or the thing we hash changes, or the involved fields, and being able to evolve it without breaking existing caching would be ideal.

vados-cosmonic (approved):
LGTM 🚀

crates/provider-sqldb-postgres/src/lib.rs
brooksmtownsend: Will address in a followup!

PR #4622, opened 2025-07-02
feat(ci): Starts process of moving release pipelines away from Nix

thomastaylor312:
This starts by adding a Dockerfile that can be used to build the image with the given binary. I started by copying what we had in wadm and adjusted from there. We can slowly migrate things over, but this removes the image build step from the nix build for now.

Please note that this gets rid of our debian only builds in favor of the wolfi images only. They have been our default for a while now. If desired, we can still publish a `-wolfi` suffixed tag if we want to. This pipeline constantly overwrites a canary tag rather than continuously adding sha tagged versions as well

### DIFF
--- a/.github/workflows/release-and-build.yml
+++ b/.github/workflows/release-and-build.yml
@@ -1,0 +1,199 @@
+# NOTE: This is the in-progress new release flow for wasmCloud that replaces the old Nix based
+# release flow.
+name: wasmCloud Release
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v[0-9].[0-9]+.[0-9]+'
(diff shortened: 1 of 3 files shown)


.github/workflows/release-and-build.yml
```diff
@@ -0,0 +1,199 @@
...
+      - name: Prepare container artifacts
+        working-directory: ./artifacts
+        run: |
+          for dir in */; do
+            name="${dir%/}"
+            mv "${name}/wasmcloud" wasmcloud
+            chmod +x wasmcloud
+            rmdir "${name}"
+            mv wasmcloud "${name}"
+          done
```
brooksmtownsend, line 147: dealing with downloaded artifacts always looks so weird to me 😢

PR #4624, opened 2025-07-03
feat(runtime): support component instantiation with custom contexts

brooksmtownsend:
## Feature or Problem
This PR separates out some traits in the `wasmcloud_runtime` crate to support greater embedding capabilities like using a custom Context object or specifying your own capability implementations.

This is primarily facilitated by the `BaseCtx` trait, which is what must be implemented for a custom context, and the `MinimalHandler` trait, which is a much smaller trait surface for most of the wasmcloud_runtime functionality. Effectively these two allow for better embedding of the runtime crate.

## Release Information
wasmcloud-runtime 0.12.0

## Consumer Impact
There were changes to public traits but they only became less restrictive, so there shouldn't be any changes needed.

### DIFF
--- a/crates/runtime/src/component/mod.rs
+++ b/crates/runtime/src/component/mod.rs
@@ -310,7 +310,7 @@ where
         let host_resources = Arc::from(host_resources);
 
         if !guest_resources.is_empty() {
-            warn!("exported component resources are not supported in wasmCloud runtime and will be ignored, use a provider instead to enable this functionality");
+            debug!("guest resources present, must be fulfilled by the linker");
         }
 
         linker_fn(&mut linker, &component)?;
@@ -466,7 +466,7 @@ where
         let host_resources = Arc::from(host_resources);
 
         if !guest_resources.is_empty() {
-            warn!("exported component resources are not supported in wasmCloud runtime and will be ignored, use a provider instead to enable this functionality");
+            debug!("guest resources present, must be fulfilled by the linker");
         }
 
         let instance_pre = linker


crates/runtime/src/component/mod.rs
```diff
@@ -192,9 +198,97 @@ where
...
+        }
+        let engine = rt.engine.clone();
+        let component = wasmtime::component::Component::new(&engine, wasm)
+            .context("failed to compile component")?;
+
+        let mut linker = Linker::new(&engine);
+
+        let ty = component.component_type();
+        let mut guest_resources = Vec::new();
+        let mut host_resources = BTreeMap::new();
```
thomastaylor312, line 296: Super duper nit: Any reason for the BTreeMap here over HashMap? Doesn't really matter in this case unless the ordering matters but a HashMap generally has better iteration performance

crates/runtime/src/component/mod.rs
```diff
@@ -271,6 +365,70 @@ pub type InvocationStream = Pin<
...
+    H: MinimalHandler,
+{
+    /// Compiles a WebAssembly component using [Runtime]. Modules are not supported with this function
+    ///
+    /// If you need to pass a custom context, use [CustomCtxComponent::new_with_linker_minimal].
+    ///
+    /// It is expected that the `linker_fn` is used to link any and all imports to the component. If
+    /// not, then the instantiation of the instance will fail.
+    #[instrument(level = "trace", skip_all)]
+    pub fn new_with_linker_minimal(
```
thomastaylor312, line 434: Nit: This is almost the same as the other new function. Maybe pull out the reusable bits if possible?

crates/runtime/src/component/mod.rs
```diff
@@ -271,6 +365,70 @@ pub type InvocationStream = Pin<
...
+        if wasmparser::Parser::is_core_wasm(wasm) {
+            anyhow::bail!("core modules are not supported in the minimal linker");
+        }
+        let engine = rt.engine.clone();
+        let claims = None;
+        let component = wasmtime::component::Component::new(&engine, wasm)
+            .context("failed to compile component")?;
+
+        let mut linker = Linker::new(&engine);
+        linker_fn(&mut linker)?;
```
thomastaylor312, line 448: Is there any reason why the linker function is called earlier here compared to the `CustomCtxComponent`? If there is a reason, we should probably have a comment here. If there isn't, maybe put it in the same place for consistency (consider this a nit)? This could be solved with the reusable bit

thomastaylor312 (approved):
Looking good and it should be fine to merge. I just had the one question around reusability that you can consider optional

crates/runtime/src/component/mod.rs
brooksmtownsend: Nope, fully copied from elsewhere in the code. Other than wrpc_runtime_wasmtime::collect_component_resource_imports requiring the BTreeMap

crates/runtime/src/component/mod.rs
brooksmtownsend: no, but it definitely should be called later to ensure that our builtins go in first. Will refactor in followup if possible to do it cleanly

PR #4632, opened 2025-07-08
feat(ci): Add workflows for scanning links and validating example docs

lachieh:
## Feature or Problem
Adds a workflow to scan example README.md files to make sure that there are no relative links that will break the docs site build.

## Related Issues
wasmcloud/wasmcloud.com#949 is not building because a change got merged with links to files that do not exist. This adds a failing check that would have prevented that.

## Release
Since this check does not prevent merge and only runs when an example is changed, it is ok to merge this PR with the failed check. ~~The underlying issue is fixed in #4633 and the check passes on that PR.~~ The fix is merged into this branch.

### DIFF
--- a/examples/rust/components/http-blobstore/README.md
+++ b/examples/rust/components/http-blobstore/README.md
@@ -51,13 +51,13 @@ wash app get
 wash get inventory
 ```
 
-For provider configuration options, see the [blobstore-nats provider documentation](../../provider-blobstore-nats/README.md).
+For provider configuration options, see the [blobstore-nats provider documentation](https://github.com/wasmCloud/wasmCloud/tree/main/crates/provider-blobstore-nats#readme).
 
 ### With Filesystem Blobstore Provider
 
 The filesystem provider stores blobs in your local filesystem. To use it make sure it's added to the desired application manifest file. other than that, the steps are the same as the NATS provider.
 
-For provider configuration options, see the [blobstore-fs provider documentation](../../provider-blobstore-fs/README.md).
+For provider configuration options, see the [blobstore-fs provider documentation](https://github.com/wasmCloud/wasmCloud/tree/main/crates/provider-blobstore-nats#readme).
 
 #### Blobstore and Filesystem Storage
 


lachieh:
Example Providers Check ✅ 
https://github.com/wasmCloud/wasmCloud/actions/runs/16156062911/job/45598661950?pr=4632#step:3:1

Example Components Check ❌
https://github.com/wasmCloud/wasmCloud/actions/runs/16156062902/job/45598661781?pr=4632#step:3:1

.github/workflows/example-components.yml
```diff
@@ -39,11 +39,9 @@ on:
   pull_request:
     branches: [main]
     paths:
-      - .github/workflows/examples-components.yml
+      - .github/workflows/example-components.yml
       - crates/wash/**
       - examples/rust/components/**
-      - examples/golang/components/**
-      - examples/tinygo/components/**
       - examples/python/components/**
```
lachieh, line 45: Removing these examples from the build at the same time, since they are covered by [wasmcloud/go](https://github.com/wasmcloud/go)

.github/workflows/example-components.yml
```diff
@@ -39,11 +39,9 @@ on:
   pull_request:
     branches: [main]
     paths:
-      - .github/workflows/examples-components.yml
+      - .github/workflows/example-components.yml
```
lachieh, line 42: Moved this file to align to the workflow name.

.github/actions/scan-links/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    required: false
+  base_directory:
+    description: 'Base directory which links must not reach out of. Relative to the repository root.'
+    required: false
+  ignore_paths:
+    description: 'Comma-separated list of paths to ignore during the scan. Relative to the repository root.'
+    required: false
+
+runs:
+  using: 'node20'
```
thomastaylor312, line 17: Should we be using a later version of node? I thought 20 was being phased out for runners

thomastaylor312 (approved):
Good to go. My comment is possibly minor so approving anyway

.github/actions/scan-links/action.yml
lachieh: :joy: Looks like that's a no: 
https://github.com/wasmCloud/wasmCloud/actions/runs/16175781007/job/45660477987?pr=4632#step:3:1

```shell
Error: System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter ''using: node22' is not supported, use 'docker', 'node12', 'node16' or 'node20' instead.')
   at GitHub.Runner.Worker.ActionManifestManager.ConvertRuns(IExecutionContext executionContext, TemplateContext templateContext, TemplateToken inputsToken, String fileRelativePath, MappingToken outputs)
   at GitHub.Runner.Worker.ActionManifestManager.Load(IExecutionContext executionContext, String manifestFile)
Error: Failed to load /home/runner/work/wasmCloud/wasmCloud/./.github/actions/scan-links/action.yml
```

node20 is still in LTS so should be fine for another couple of years

PR #4681, opened 2025-07-31
ci: provider build efficiency

brooksmtownsend:
- **fix(flake): only use required features for nextest**
- **ci(provider): remove wash build duplicate build**
- **ci(wasmcloud): conditionally build providers**

## Feature or Problem
The purpose of this PR is to use as many tricks as we can to only build, test, and package exactly what needs to be built, tested, and packaged.

This primarily introduces two new jobs in the wasmcloud action, responsible for determining if a specific provider needs to be built to make sure dependencies didn't affect it (top level Cargo.* / flake changes) or if it needs to be built and tested on all architectures (a file specifically changed inside of that provider)

### DIFF
--- a/.github/workflows/provider.yml
+++ b/.github/workflows/provider.yml
@@ -127,37 +127,3 @@ jobs:
         env:
           WASH_REG_USER: ${{ github.repository_owner }}
           WASH_REG_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
-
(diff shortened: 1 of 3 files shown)


brooksmtownsend:
This is a good start but I'll have to come back to it to wrap up some of the pieces that are affecting the broader action conditions. It's hard to tell if this will affect the release workflows without trying to, and I'd rather not test that atm

PR #4687, opened 2025-08-01
ci(wasmcloud): compare cargo performance to nix

brooksmtownsend:
## Feature or Problem
Comparing running our cargo suite of checks just with bare cargo + cache v.s. nix.


brooksmtownsend:
TL;Dr seems not that different BUT we need to figure out how to run cargo nextest in not release mode.

PR #4697, opened 2025-08-02
[WIP] feat: Use humantime instead of milliseconds for time values

jtakakura:
## Feature or Problem

This PR continues the work of standardizing CLI options to use [`humantime`] for time-based values.
Previously, options like `--some-option-ms 100` accepted milliseconds.
With this change, we aim to enable more human-readable formats such as `--some-option 100ms`, enhancing user experience and clarity in CLI interactions.

## Related Issues

#3257 #3440

## Release Information

`next`

## Consumer Impact

This change is taking care to be backward compatible, and maintains the existing options and environment variables. The new functionality supports falling back to the existing functionality as well (e.g. setting SOME_ENV_VAR to 100 and 100ms will have the same effect).

### Unit Test(s)
Add tests for duration parsing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8919,6 +8919,7 @@ dependencies = [
  "clap-markdown",
  "futures",
  "http-body-util",
+ "humantime",
  "hyper 1.6.0",
  "hyper-util",
  "nats-jwt-rs",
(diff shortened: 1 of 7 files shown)


omarjatoi (approved):
Thanks again @jtakakura for completing this!

PR #4703, opened 2025-08-06
fix(wasmbus): expect when we should've mapped

brooksmtownsend:
## Feature or Problem
This PR fixes an area in the host inventory related to component limits where we were calling `expect` when an option map would do.

## Release Information
v1.9


vados-cosmonic (approved):
LGTM :rocket:

PR #4745, opened 2025-08-25
chore: bump wasmcloud-component 0.3.0

brooksmtownsend:
## Feature or Problem
Bump wit-bindgen and wasmcloud-component to 0.3.0 to avoid pointer misalignment issues.


thomastaylor312:
Just pushed a commit to fix a place where we forgot to bump the dep on a path import

PR #4797, opened 2025-09-25
blobby: Finish upload stream

lxfontes:
exercising `finish` so the OutgoingValue knows we are done with the stream.


ricochet:
unrelated example failure filed #4798

ricochet (approved):

PR #4815, opened 2025-10-07
feat: flexible paths for path routing_mode

if0ne:
## Feature or Problem

Allows for more flexible configuration of the paths used to invoke components.


if0ne:
@ricochet Could you restart pipelines?

Fix is out: https://blog.rust-lang.org/2025/11/10/Rust-1.91.1/#linker-and-runtime-errors-on-wasm

Issue #4642, opened 2025-07-09
[FEATURE] Transition the capability provider model into support for wRPC servers


brooksmtownsend:
### Affected project(s)

- [ ] documentation
- [ ] examples
- [ ] wasmCloud host
- [ ] wasmCloud CLI (wash)
- [ ] wasmCloud dashboard UI (washboard)
- [x] capability providers
- [ ] provider bindgen
- [ ] control interface client
- [ ] other / not sure

### Is your feature request related to a problem? Please describe.

Capability providers are both wasmCloud specific and one of our most frictionful experiences. They are distributed as architecture/os specific binaries via provider archives (a wasmCloud specific tarball format), executed on the same machine that the host is running on, and passed a single line of base64 encoded [HostData](https://docs.rs/wasmcloud-provider-sdk/0.16.0/wasmcloud_provider_sdk/struct.HostData.html) which gives the provider all the information it needs to connect to NATS, setup observability, and subscribe to link updates.

Capability providers don't have their own isolated resources and are often difficult to manage in production.

So today, capability providers:
- Are launched as binaries from the host itself
- Communicate using core NATS with the host for links and config
- Call components using wRPC's NATS transport

### Describe the solution you'd like

A "wRPC server" is simply any binary, container, component, or application that serves a WIT interface via one of the transports available with wRPC (TCP, NATS, QUIC, UDP, etc.)

In combination with #4636 , I'd like to see us lean more into capability providers as wRPC servers rather than their own concept. We should deprecate the ability for the host to launch providers themselves, preferring instead to configure a host with the connection information and transport necessary to communicate with a wRPC server. If we had this, it could look a lot like @LUK3ARK 's proposal for "virtual components" in https://github.com/wasmCloud/wasmCloud/pull/4228, where the host communicates with an entity over wRPC.

Notably, this would allow for our capability implementations to be done in any language as long as it can implement the protocol (not requiring a standalone binary) and could be launched as containers when running in a containerized environment for independent resource control.

### Describe alternatives you've considered

Instead of leaning more into wRPC servers, we could also attempt to migrate entirely to using components as capability providers and execute them on the wasmCloud side instead. This likely requires WASI 0.3, and will make it a bit difficult to accomplish in any language since the component target is required.

### Additional context

https://github.com/wasmCloud/wasmCloud/pull/4228

brooksmtownsend:
Open questions
- Do providers need to register with a given server
- Are they globally available in the lattice

[...]

Issue #4639, opened 2025-07-09
Remove usage of `wascap` and `nkeys`


ricochet:
We still use `wascap` and `nkeys`, utilities for signing and packaging wasmCloud-specific artifacts. These solved a problem at the time, but are largely unused in today's wasmCloud deployments.

We've demonstrated using GitHub actions to sign OCI release artifacts as the best practice for signing and is now the industry standard.

brooksmtownsend:
Hey @LUK3ARK, are you using nkeys / the embedded JWTs in your deployment or do you use the signed artifacts as-is?

ossfellow:
There are 2 types of attestations that are relevant to all wasmCloud generated executable artifacts (i.e. components, providers, native binaries):

1. Build time provenance and attestation, which can be achieved by utilizing Linux Foundation backed and endorsed frameworks and tooling such as [SLSA Framework](https://slsa.dev/provenance/v1), [In-Toto](https://in-toto.io/docs/what-is-in-toto/), [SBOM](https://www.linuxfoundation.org/blog/blog/what-is-an-sbom), [Sigstore's Cosign](https://www.sigstore.dev), and [Notary v2](https://github.com/notaryproject/.github/blob/main/README.md).
2. Runtime trust, which can be achieved by utilizing CNCF's [SPIFFE](https://www.cncf.io/projects/spiffe/).

The obvious advantage of leveraging such frameworks and tooling is easy integration with well established ecosystems of generation, storage, and verification that are already in use, complemented by further simplification of the wasmCloud codebase.

vados-cosmonic:
(pretty sure this happened by way of v2 existing, but...)

The project has changed so drastically that this idea is probably not reasonable to pursue in this form -- closing this as we're focusing effort now on wasmCloud v2.

If you think this definitely still applies towards v2, please feel free to open a new issue!

PR #4877, opened 2025-11-24
Interface-driven extensions

LUK3ARK:
## Externalization of providers

# NOTES
- I use provider and extension interchangeably, extension is the new form of a provider which is more general and dynamic
- ALL naming is subject to change for interfaces and software if this architectural change is beneficial
- I need to fix tests which I will do once I get some feedback on the architecture
- This is for issue https://github.com/wasmCloud/wasmCloud/issues/4642

Wasmcloud hosts lattices. Lattices are composed of a bunch of linked components and providers. Providers are currently specific to wasmcloud and require very specific custom support and can cause a lot of blockers. Currently:
- Providers can only be made in rust and go with the help of first party SDKs
- Providers must follow a very strict flow
- Providers have to be internally managed and ran by the host itself
- Providers cannot horizontally scale the same way as components
- Providers are currently the bottleneck in terms of the amount of requests that can enter and exit into the lattice

Currently, architecture emerges like so
image

with the new approach, providers become extensions and allow a lattice to emerge as a consequence of existing/external infrastructure and the set of hosts

image

The benefits of this are:
- opportunities to remove responsibility of the host
- Users can now decide whether to give the host responsibility of managing an extension or just connect to the host from an external system
- More permutations of systems can be made
- lower cost for integrating wasmCloud into existing business infrastructure and systems

## Interface-driven extensions
Now that providers are extensions, we need a way of interacting with them outside of the domain of the host. A provider currently is monolothic and requires a lot of custom logic in order to operate. The only interfaces a provider satisfies is the functionality it was built for, e.g. postgres for a postgres provider or wasi:keyvalue for a redis keyvalue store. This is great,  but we can do better.

We can generalise extensions (providers) by decomposing the things it is responsible for into separate interfaces:
image

Currently 2 interfaces have been made
1) Manageable - the manageable interface is the single mandatory interface required for a extension to satisfy to allow a host to connect and 'bind' to it. this is just a handshake and allows for the exchange of important information such as workload identities and extension keys to allow secure secret distribution. A shutdown function is also implemented, meaning logic can be created and controlled by wasm components, software and hosts.
    - Easier to make CI/CD processes and infrastructure (even wasm-controlled CI/CD)
    - programmatic, declarative control of shutdown sequences for any variety of reasons
    - Anyone can choose how to implement these as required, a user might not want to give back an ecnryption key because they do not need to use secrets, or a user might only want to expose certain functionality through the shutdown interface

2) Configurable
Inside interface-driven lattices, there is contextual execution. this means the execution of a component or extension might differ based on the link it traverses. links need to store specific configuration which defines how a extension behaves based on the source that is calling it. This is currently done with links, which store configuration.
    - Extensions/components need to have confgiuration supplied to decide how it works
    - They need link-speicfic configuration supplied to decide how it works based on 'who' is asking or receiving these RPC invocations.
    - 'Confgurable' interface is not mandatory.

... 
In the future, more interfaces can be added and implemented so that the host can interact in different ways with different extensions without forcing an extension to implement ALL interfaces.

### CTL protocol for extensions
If an extension is added, we could potentially serve the extension interfaces with the same wrpc client (and therefore prefix) as the extensions functionality (using the RPC lattice protocol). This causes a couple of issues though:
    - If there are multiple providers running and linked to the same subject, it is unpredictable about which one would react to CTL messages. this is amazing for RPC calls, we can distribute workload globally. This is terrible for CTL calls, as an extension shutdown message might shutdown a extension we didnt want to shutdown, or configure an extension we didnt want to configure.
    - We lose individual host context. We dont know which host is running what extensions (whether managed or externally)

This is fixed by serving the extension interfaces on a ctl subject like so:
`"wasmbus.ctl.v1.{lattice}.extension.{extension-id}.{host-id}"

This maintains the context of the host id we want to issue the invocation to so that we can individually configure, shutdown, check, etc. extension-id on different hosts

## Extensions still 'live' on a host
Even though the functionality of extensions need to globally contribute to the lattice, as said earlier we need to maintain uniqueness for congfiguration and management purposes. Because of this, regarldess if an extension is managed or external, it needs to connect to an individual host. Therefore, we make sure an individual host is responsible for the bind and management of these extensions. This compliments existing topologies of how lattices are setup and connected together over nats. Config watchers are still spawned for external extensions as the host is still responsible for making sure its up to date.

image

# Further requirements and finalization
- We need a way of deterministically knowing which extension satisfies what interfaces. In a perfect world, this can be done by adjusting the wit-bindgen macro to export the list of served exports from the extension WIT package and automatically adding them to the bind request or in its own manageable function such as satisfied_interfaces(). For now, it would still be sufficient to just force the binary to implement this method and return back a list of interfaces it satisfies, or attempt to gather this information from a deployment manifest.
- Secret distribution still requires JWT claims, but this isnt stored inside the binaries anymore. Need some guidance on new secret management if wascap and nkeys are going away. See https://github.com/wasmCloud/wasmCloud/issues/4639
- How to deal with workload identity?
- External extensions that get a Host ID to extend do not currently validate the host id exists. This is not a concern for internal providers because its provided by the host itself that is running it
- How do external providers interact with policy managers?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6625,11 +6625,11 @@ dependencies = [
  "rand 0.9.2",
  "serde",
  "serde_json",
+ "spiffe",
  "thiserror 2.0.17",
  "tokio",
  "tracing",
  "tracing-subscriber",
- "wascap",
  "wasmcloud-secrets-client",
  "wasmcloud-secrets-types",
 ]
@@ -7470,7 +7470,6 @@ dependencies = [
  "serde_json",
  "tokio",
  "url",
- "wascap",
  "wasi-preview1-component-adapter-provider",
  "wit-component 0.224.1",
 ]
@@ -8778,7 +8777,6 @@ dependencies = [
  "oci-wasm 0.3.0",
  "once_cell",
  "path-absolutize",
- "provider-archive",
  "rand 0.9.2",
  "regex",
  "reqwest",
@@ -8813,7 +8811,6 @@ dependencies = [
  "wadm-types",
  "walkdir",
  "warp",
- "wascap",
  "wasi-preview1-component-adapter-provider",
  "wasm-encoder 0.232.0",
  "wasm-pkg-client",
@@ -9218,7 +9215,6 @@ dependencies = [
  "nats-jwt-rs",
  "nkeys",
  "once_cell",
- "provider-archive",
  "rand 0.9.2",
  "redis",
  "regex",
@@ -9293,6 +9289,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-opentelemetry",
+ "wit-bindgen-wrpc 0.9.0",
 ]
 
 [[package]]
@@ -9313,7 +9310,6 @@ dependencies = [
  "oci-client 0.15.0",
  "oci-wasm 0.3.0",
  "once_cell",
- "provider-archive",
  "reqwest",
  "rustls 0.23.35",
  "rustls-native-certs 0.8.2",
@@ -9329,9 +9325,11 @@ dependencies = [
  "tracing",
  "unicase",
  "url",
+ "uuid 1.18.1",
  "wascap",
  "wasmtime-wasi-http",
  "webpki-roots 1.0.4",
+ "wit-bindgen-wrpc 0.9.0",
  "wrpc-interface-http",
 ]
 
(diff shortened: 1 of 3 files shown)


ricochet:
OK phew good! I can tell this was a ton of work. 

Everything for v2 is in http://github.com/wasmcloud/wash

We're working through documentation at https://wasmcloud.com/docs/v2.0.0-rc.1/ , note another one is incoming at https://github.com/wasmCloud/wasmcloud.com/pull/1026

PR #4905, opened 2026-01-23
feat(nats): add support for clear-text user/password authentication via NATS connection URL

jucastilhoduarte:
## Feature or Problem

This PR enables **optional, opt-in support for clear-text NATS username/password authentication when credentials are explicitly embedded in the NATS connection URL** (e.g. `nats://user:pass@host:port`).

This is **not** intended to introduce username/password authentication as a first-class or recommended authentication mechanism. Instead, it provides a **deliberate escape hatch** for advanced users who understand the implications and explicitly choose to configure it via `--rpc-host` or `--ctl-host`.

This capability is required by some users to run on environments where NATS authentication is currently standardized on username/password and cannot be easily changed without disrupting existing infrastructure.

---

## Related Issues

- wasmCloud documentation: *User/password authentication not supported*  
  https://wasmcloud.com/docs/deployment/security/nats-segmentation/#userpassword-authentication-not-support

- Discussion with Cosmonic team (@vados-cosmonic ) confirming feasibility via `async_nats` URL parsing and intent to keep this non-first-class.

---

## Release Information

Target release: **next**

---

## Consumer Impact

**No impact to existing users or default behavior.**

- JWTs, NKeys, and workload identity flows remain unchanged and fully enforced.
- Username/password authentication is only applied if:
  - No JWT / NKey / workload identity configuration is present **and**
  - Credentials are explicitly embedded in the NATS URL provided by the user.
- There are no new flags, configuration keys, or behavior changes for existing deployments.

This change is effectively invisible unless a user intentionally opts into it.

---

### Unit Test(s)

No new unit tests were added.  
This change relies on existing `async_nats` URL parsing behavior and only wires already-supported functionality into the wasmCloud host connection path.

### Acceptance or Integration

No changes to acceptance or integration test suites.

### Manual Verification

- Started a local NATS server configured with username/password authentication.
- Connected wasmCloud using `--rpc-host` and `--ctl-host` with URLs in the form: `nats://user:password@127.0.0.1:4222`
- Verified successful host startup, control plane communication, and workload lifecycle operations.
- Confirmed that:
- JWT/NKey paths remain unaffected.
- Username/password auth is ignored unless explicitly embedded in the URL.
- No behavioral changes occur when credentials are not provided.

### DIFF
--- a/examples/rust/components/messaging-image-processor-worker/src/processing.rs
+++ b/examples/rust/components/messaging-image-processor-worker/src/processing.rs
@@ -103,7 +103,7 @@ impl ImageProcessingRequest {
         };
 
         // Attempt to lease the task
-        let lease_id = match tasks::lease_task(&task_id, &WORKER_ID.into()) {
+        let lease_id = match tasks::lease_task(&task_id, &String::from(WORKER_ID)) {
             Ok(lid) => lid,
             Err(e) => {
                 log(


vados-cosmonic (approved):
LGTM :rocket: 

I think we might be able to move this logic into a constructor in `ConnectOptions` but not sure if I will try to change that before merging.

[EDIT] On closer inspection, the code is just about perfect for what we need

PR #4964, opened 2026-03-09
SPIKE: WRPC Plugin

ricochet:
> [!NOTE]
> 📋 **Copied from https://github.com/wasmCloud/wash/pull/295**

Getting this up  for discussion / ergonomics. 

WRPC NATS Only, using the `data` NATS settings.
The WRPC Plugin looks for `host_interfaces` with config key `wrpc:name`,forwarding imports/exports for that particular 'namespace + package' to wrpc layer. The `wrpc:name` is used as part of the wrpc subject.

There is no process supervision or interactions with wasmcloud provider sdk.

## WRPC Routing

We pivot identifiers from `workload_namespace & workload_name` to `host_group & wrpc_name`, resembling wasmcloud's v1 Lattice approach.

`{host_group}.{wrpc_name}.[wrpc-lib-subject]`

This happens for a few reasons:
- Guarantee no crosstalk across hostgroups
- `wrpc_name` has to remain stable across restart, and can't use `Workload` names here as they are ephemeral (would be nice to use `WorkloadDeployment` names)

Crosstalk can still be achieved via NATS subject mappings, however it is an advanced use-case.

Example:

All exported interfaces from `wrpc-examples/hello` found on the component will be served over WRPC.

```
     hostInterfaces:
        - namespace: wrpc-examples
          package: hello
          interfaces:
            - handler
          config:
            "wrpc:name": wasmcloudrpchello
```

Client:

```
rpc ❯ ./target/debug/hello-nats-client default.wasmcloudrpchello
2026-02-23T20:37:01.480694Z  INFO async_nats::connector: connected successfully server=4222 max_payload=1048576
2026-02-23T20:37:01.480974Z  INFO async_nats: event: connected
default.wasmcloudrpchello: hello from Rust
```

### Identity ( WRPC Headers )

(still to be coded)

Headers help external WRPC endpoints identify the caller & multiplex interfaces:

- `From`: `{wrpc_name}` from the source
- `X-Wasmcloud-Workload`: `{workload_namespace}/{workload_name}`
- `X-Wasmcloud-Workload-Id`: `{workload_id}`

Also considering passing any additional Interface `config` as header.

## Questions

**How to handle different `wrpc:name` for import/export on the same package?**

This is a limitation of `host_interfaces`. Today all import/exports land on the same `WitInterface`  with a single configuration.

**HostPlugin or not?**

Seems weird to pluck a specific `HostPlugin` within `wash_runtime::engine`.

I feel this warrants another "engine-plugin" like HTTP.

**Target Switching (Links)**

[v1 linking at runtime](https://wasmcloud.com/docs/concepts/linking-components/linking-at-runtime/)

Besides similar challenges with `host_interfaces`, we also have:
- Scope: WRPC-only links or for components/HostPlugins too?
- Link Switch Mechanism: v1 has dedicated WIT Interface for Link change

IMO: WRPC-Only with similar WIT Interface for Link change

---------

From comment by @lxfontes 

Using the existing cron-service & cron-component combo:

```
# This will become a WRPC Server (nats subscription)
kind: WorkloadDeployment
metadata:
  name: cron-component
spec:
  replicas: 1
  template:
    spec:
      hostInterfaces:
        - namespace: wasmcloud
          package: example
          interfaces:
            - cron
          config:
            "wrpc:name": crontest
      components:
        - name: hello
          image: ttl.sh/lxfontes/cron-component:test
---
# This will become a WRPC Invoker
kind: WorkloadDeployment
metadata:
  name: cron-service
spec:
  replicas: 1
  template:
    spec:
      hostInterfaces:
        - namespace: wasmcloud
          package: example
          interfaces:
            - cron
          config:
            "wrpc:name": crontest
      service:
        image: ttl.sh/lxfontes/cron-service:test
```

Screenshot 2026-02-23 at 4 59 50 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,6 +264,7 @@ dependencies = [
  "portable-atomic",
  "rand 0.8.5",
  "regex",
+ "ring",
  "rustls-native-certs 0.7.3",
  "rustls-pemfile",
  "rustls-webpki 0.102.8",
@@ -2956,6 +2957,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "884e2677b40cc8c339eaefcb701c32ef1fd2493d71118dc0ca4b6a736c93bd67"
 
 [[package]]
+name = "leb128-tokio"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a1f734b00871cae8f137d83d53cf4b5ee3f0d87224a87a4347dab4cde11a7a3"
+dependencies = [
+ "tokio",
+ "tokio-util",
+ "tracing",
+]
+
+[[package]]
 name = "leb128fmt"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5172,6 +5184,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "send-future"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "224e328af6e080cddbab3c770b1cf50f0351ba0577091ef2410c3951d835ff87"
+
+[[package]]
 name = "serde"
 version = "1.0.228"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


vados-cosmonic:
Hey I think we've come a long way from this -- let's close it for now and re-visit if we want to work on integrating wRPC again (IMO we would do well to just restart)

PR #4967, opened 2026-03-09
fix: continue migration but now for install

automation-wasmcloud:
After this one, we should be ready to cut a rc


lxfontes (approved):

.github/workflows/test-install-scripts.yml
kshaffer42025-rgb: **_[]()_**

kshaffer42025-rgb: github.com/google/licensecheck

PR #4979, opened 2026-03-12
feat(rt): add handling allowed hosts

if0ne:
## Summary

- Adds `allowed_hosts` enforcement for outgoing HTTP requests in `wash-runtime`
- Implements `check_allowed_hosts()` in `DynamicRouter` to validate outgoing request hosts against a configurable allowlist, with support for wildcard patterns (`*.example.com`)
- Propagates `allowed_hosts` from `LocalResources` through `CtxBuilder` → `Ctx` → `WasiHttpView::outgoing_request`, so per-workload policies are respected at request time
- Adds a full integration test suite with a new `http-allowed-hosts` fixture component, covering: requests to allowed hosts, blocked hosts, wildcard matches, and empty allowlist (allow-all) behavior

## Behavior

- If `allowed_hosts` is empty, all outgoing requests are permitted (backwards compatible)
- Wildcard patterns like `*.example.com` match any subdomain but not the bare domain
- Denied requests return `HttpRequestDenied` error code and emit a `warn` log with the workload ID and reason
- `DevRouter` bypasses the check (dev mode allows all)

### DIFF
--- a/crates/wash-runtime/tests/integration_http_webgpu.rs
+++ b/crates/wash-runtime/tests/integration_http_webgpu.rs
@@ -75,7 +75,7 @@ async fn test_http_webgpu_integration() -> Result<()> {
                     config: HashMap::new(),
                     environment: HashMap::new(),
                     volume_mounts: vec![],
-                    allowed_hosts: vec![],
+                    allowed_hosts: Default::default(),
                 },
                 pool_size: 1,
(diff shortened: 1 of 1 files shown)


if0ne:
СС @lxfontes

lxfontes (approved):

PR #4980, opened 2026-03-12
feat: add s3 plugin

### LINKED ISSUE #4977: [FEATURE] Implement new WASI Blobstore plugin for S3
WasmCloud v1 provides an S3 provider implementing the WASI Blobstore interface. Introducing an S3 plugin in WasmCloud v2 would ensure feature parity and support common object storage use cases.

if0ne:
Closes #4977 

Add an S3-backed implementation of the `wasi:blobstore@0.2.0-draft` plugin for `wash-runtime`, using the [`object_store`](https://docs.rs/object_store/0.13) crate with the AWS/S3 feature.

### DIFF
--- a/crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
+++ b/crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
@@ -240,27 +240,7 @@ impl<'a> bindings::wasi::blobstore::blobstore::Host for ActiveCtx<'a> {
             }
         };
 
-        let store = match plugin.get_or_create_store(&self.component_id, &name).await {
-            Ok(Some(store)) => store,
-            Ok(None) => {
-                return Ok(Err(
-                    "S3 client not configured for this component".to_string()
-                ));
-            }
-            Err(e) => {
-                return Ok(Err(format!(
-                    "failed to create S3 client for bucket '{name}': {e}"
-                )));
-            }
-        };
-
-        let container_data = ContainerData {
-            name: name.clone(),
-            store,
-        };
-
-        let resource = self.table.push(container_data)?;
-        Ok(Ok(resource))
+        Ok(Err("Not implemented".to_string()))
     }
 
     #[instrument(skip(self))]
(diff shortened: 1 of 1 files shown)


crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+use std::sync::Arc;
+
+use crate::engine::ctx::{ActiveCtx, SharedCtx, extract_active_ctx};
+use crate::engine::workload::WorkloadItem;
+use crate::plugin::HostPlugin;
+use crate::plugin::WorkloadTracker;
+use crate::wit::{WitInterface, WitWorld};
+use anyhow::Context;
+use object_store::aws::{AmazonS3, AmazonS3Builder};
+use object_store::{ObjectStore, ObjectStoreExt};
```
ricochet: Move `TryStreamExt` import to the top level and drop unused `ObjectStoreExt`:

```suggestion
use anyhow::Context;
use futures::TryStreamExt;
use object_store::aws::{AmazonS3, AmazonS3Builder};
use object_store::ObjectStore;
```

crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+                return Ok(Err(format!(
+                    "failed to create S3 client for bucket '{name}': {e}"
+                )));
+            }
+        };
+
+        // Delete all objects in the bucket.
+        let list_result = store.list(None);
+
+        use futures::TryStreamExt;
```
ricochet: Clarify why we're deleting objects instead of the bucket, and remove the now-unnecessary scoped `use`:

```suggestion
        // S3 bucket deletion is not supported via the object_store crate,
        // so we delete all objects in the bucket instead.
        let list_result = store.list(None);

```

crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+        };
+
+        // Try to list with a max of 1 result to see if the bucket is accessible.
+        use futures::TryStreamExt;
+        match store.list(None).try_next().await {
+            Ok(_) => Ok(Ok(true)),
+            Err(object_store::Error::NotFound { .. }) => Ok(Ok(false)),
+            // Bucket exists but might be empty or accessible
+            Err(_) => Ok(Ok(true)),
+        }
```
ricochet: Remove scoped import (now at top level) and propagate the error instead of silently returning `true`, an unexpected error (e.g. network timeout, auth failure) shouldn't be treated as "bucket exists":

```suggestion
        match store.list(None).try_next().await {
            Ok(_) => Ok(Ok(true)),
            Err(object_store::Error::NotFound { .. }) => Ok(Ok(false)),
            Err(e) => Ok(Err(format!(
                "failed to check if container '{name}' exists: {e}"
            ))),
        }
```

crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+    ) -> anyhow::Result<Result<ContainerMetadata, ContainerError>> {
+        let container_data = self.table.get(&container)?;
+
+        Ok(Ok(ContainerMetadata {
+            name: container_data.name.clone(),
+            created_at: 0,
+        }))
+    }
+
+    #[instrument(skip(self, container))]
```
ricochet, line 531: Worth calling out the memory implications here:

```suggestion

    // NOTE: This reads the entire object into memory. For large objects,
    // callers should use ranged reads (non-zero start / end < u64::MAX).
    #[instrument(skip(self, container))]
```

crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+            Some(token) => token,
+            None => {
+                return Ok(Err("unauthorized write".to_string()));
+            }
+        };
+
+        // Prepare the write operation — actual upload happens on `finish`.
+        let handle = self.table.get_mut(&data)?;
+        handle.container = Some(container_data);
+        handle.object_name = Some(name.as_str().to_string());
```
ricochet: `name` is already a `String`

```suggestion
        handle.object_name = Some(name);
```

crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+    async fn list_objects(
+        &mut self,
+        container: Resource<ContainerData>,
+    ) -> anyhow::Result<Result<Resource<StreamObjectNamesHandle>, ContainerError>> {
+        let container_data = self.table.get(&container)?;
+
+        let list_result = container_data.store.list(None);
+
+        use futures::TryStreamExt;
+        let objects: Vec<_> = match list_result.try_collect().await {
```
ricochet, line 610: Remove scoped import (now at top level) and unnecessary blank line:

```suggestion
        let list_result = container_data.store.list(None);
        let objects: Vec<_> = match list_result.try_collect().await {
```

crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+            Some(token) => token,
+            None => {
+                return Ok(Err("unauthorized delete".to_string()));
+            }
+        };
+
+        let list_result = container_data.store.list(None);
+
+        use futures::TryStreamExt;
+        let objects: Vec<_> = match list_result.try_collect().await {
```
ricochet, line 751: Same cleanup — remove scoped import and blank line:

```suggestion
        let list_result = container_data.store.list(None);
        let objects: Vec<_> = match list_result.try_collect().await {
```

crates/wash-runtime/src/plugin/wasi_blobstore/s3.rs
```diff
@@ -0,0 +1,1130 @@
...
+            .iter()
+            .find(|i| i.namespace == "wasi" && i.package == "blobstore")
+        else {
+            return Ok(());
+        };
+
+        let buckets = match interface.config.get("buckets") {
+            Some(buckets) => buckets.split(',').map(|s| s.to_string()).collect(),
+            None => vec![],
+        };
```
ricochet: Add an explicit type annotation (aids readability with `.collect()`) and warn when no buckets are configured.  This is a common misconfiguration that's hard to debug:

```suggestion
        let buckets: Vec<String> = match interface.config.get("buckets") {
            Some(buckets) => buckets.split(',').map(|s| s.to_string()).collect(),
            None => {
                tracing::warn!(
                    workload_id = workload.id(),
                    "no buckets configured for S3 blobstore \u2014 workload will be denied access to all containers"
                );
                vec![]
            }
        };
```

PR #4982, opened 2026-03-13
fix: Threading OCI Cache for 'wash host'

lxfontes:
Needed for `ImagePullPolicy: IfNotPresent | Never`.

Pointing `oci-cache-dir` to an `emptyDir` in Kubernetes to guarantee artifacts don't land on a tmpfs / shmfs ( consuming host memory ).


crates/wash-runtime/src/engine/workload.rs
```diff
@@ -695,7 +695,7 @@ impl ResolvedWorkload {
                             // This is expected for host-provided interfaces (e.g. wasi:*).
                             // If it's not host-provided, linking will fail later with a
                             // clear error from wasmtime.
-                            warn!(
+                            debug!(
```
lxfontes, line 698: taming this `warn` as it is quite repetitive ( any host interface ).

PR #4997, opened 2026-03-20
worktree validate blobby

ricochet:
- **chore: relax wstd dependency to semver-compatible range**
- **refactor(blobby): rewrite example to use wstd**
- **fix(nats-blobstore): replace AsyncWriteStream**


.wash/config.yaml
ricochet, line 1: errant file cleanup

PR #4998, opened 2026-03-22
fix(wash-runtime): only build test fixtures in cfg(test) mode

### LINKED ISSUE #4983: Build fixtures only in `#[cfg(test)` mode
When using wash-runtime as a library, the fixtures are built along with the library code, even though they are not used anywhere except in tests.

mjb-it:
## Summary

- The `build.rs` in `wash-runtime` was unconditionally compiling 11 WASM
  test fixtures on every build, even when the crate is used as a library
  dependency
- Fixtures are now only built when `CARGO_CFG_TEST` is set, which Cargo
  sets automatically during `cargo test`
- Moves the `tracked_examples` array inside the conditional since it is
  only needed for fixture builds

Fixes #4983

## Test plan

- [ ] `cargo build -p wash-runtime` completes without building any WASM fixtures
- [ ] `cargo test -p wash-runtime` still builds fixtures and runs tests successfully

### DIFF
--- a/crates/wash-runtime/build.rs
+++ b/crates/wash-runtime/build.rs
@@ -33,8 +33,8 @@ fn check_and_rebuild_fixtures(
     let wasm_dir = workspace_dir.join("crates/wash-runtime/tests/wasm");
 
     if !fixtures_dir.exists() {
-        println!("No fixtures dir found at {}", fixtures_dir.display());
-        anyhow::bail!("No fixtures dir found");
+        // Not present when this crate is used as a library dependency — skip silently.
+        return Ok(());
     }
 
     // Create wasm directory if it doesn't exist
@@ -132,6 +132,11 @@ fn check_and_rebuild_fixtures(
 fn check_and_rebuild_p3_fixtures(workspace_dir: &Path, p3_fixtures: &[&str]) -> anyhow::Result<()> {
     let fixtures_dir = workspace_dir.join("crates/wash-runtime/tests/fixtures");
     let wasm_dir = workspace_dir.join("crates/wash-runtime/tests/wasm");
+
+    if !fixtures_dir.exists() {
+        // Not present when this crate is used as a library dependency — skip silently.
+        return Ok(());
+    }
 
     fs::create_dir_all(&wasm_dir)?;
 
@@ -278,7 +283,6 @@ fn main() {
     );
     let workspace_dir = workspace_dir().expect("failed to get workspace dir");
 
-    // Track specific example directories we care about
     let tracked_examples = [
         "http-counter",
         "cron-service",
@@ -303,8 +307,6 @@ fn main() {
         "inter-component-call-p3-callee",
     ];
 
-    // Build test fixtures. The rerun-if-changed directives ensure these only
-    // rebuild when fixture source files actually change, not on every build.
     check_and_rebuild_fixtures(&workspace_dir, &tracked_examples)
         .expect("failed to check/rebuild fixtures");
 


Aditya1404Sal:
Hey @mjb-it great job! I had a tiny request tho : could you rebase and apply the same cfg(test) check for the p3 fixtures too ?

mjb-it:
On it, Thanks!

Aditya1404Sal:
LGTM 🚀

vados-cosmonic:
Hey would you mind rebasing this PR against `main`, @mjb-it ? It seems like some of the code got taken added already?

mjb-it:
> Hey would you mind rebasing this PR against `main`, @mjb-it ? It seems like some of the code got taken added already?

On it

vados-cosmonic:
Hey @mjb-it is this working for you locally? what happens if you put a bug/compile error in one of the examples and try to use the library?

mjb-it:
> Hey @mjb-it is this working for you locally? what happens if you put a bug/compile error in one of the examples and try to use the library?

Of course I didn't 🤦 I just rebased and pushed. Update commit is incoming.

dman-os:
Btw, does this mean that the crates get built when running tests for any crate that depend on the library?

Also, might be better to set a separate or explicit cargo target dir for these builds. I had an issue with the nested cargo build inheriting a target dir that a parent had set leading to build deadlocks.

Aditya1404Sal:
> Also, might be better to set a separate or explicit cargo target dir for these builds. I had an issue with the nested cargo build inheriting a target dir that a parent had set leading to build deadlocks.

Hey @dman-os could you tell us a bit more about that issue you had ? did this issue arise for the latest upstream/main or is your commit history a bit outdated? this was added recently #5041

dman-os:
Hey, sorry for not providing more details. I can confirm this behavior on `wash-runtime = { git = "https://github.com/wasmcloud/wasmcloud", rev = "33fa1990f9d4fbc85cac2a2b5f7f8be7b4fab83e" }` though it only seems to show up under very specific circumstances I can't pin down. I can reliably reproduce when doing release dylib builds driven by gradle. But not when it's built from another bin crate transitively I'll look into it more later and report if I find anything more.

dman-os:
Oof, yeah. That sha predates your PR, my bad.

Tangent but wdty of enforcing squash PR commits for the repo?

dman-os:
Okay, even on later SHAs (1eba3b068dd0ebc9fd45a9daa785bdb4476d8fd2 aka v2.0.4), I'm seeing this issue (build deadlock) 
image

Edit: strange, I'm seeing the fixture builds when using this PR as well.

dman-os:
Yeah so I think the issue is that I'm using Cargo's git dependencies feature which bypasses the checks this PR is doing for avoiding fixture builds. `wash-runtime` isn't being actively published to crates.io so we need a better check.

dman-os:
So I'm not seeing a good way of detecting if the current build.rs run is for a test run on the `wash-runtime` crate. Other solutions include either using a crate feature flag on an environment variable or just avoid building them in `build.rs`. If building the fixtures requires an extra flag, might as well be a separate command. Wdyt @Aditya1404Sal ?

Aditya1404Sal:
> So I'm not seeing a good way of detecting if the current build.rs run is for a test run on the `wash-runtime` crate. Other solutions include either using a crate feature flag on an environment variable or just avoid building them in `build.rs`. If building the fixtures requires an extra flag, might as well be a separate command. Wdyt @Aditya1404Sal ?

Hey @dman-os! Those are some Great observations & we love the contributions so keep them coming 😉 

So instead of gating on a feature flag or env var (both of which I think still require either the user remembering to pass it, or some out-of-band convention to keep CI honest), I've been thinking about a hypothetical methodology that supports "needed for tests, invisible to consumers, never propagated": maybe split the fixture build into a sibling crate (something like wash-runtime-ghost) and declare it under [dev-dependencies] of wash-runtime. My understanding is that Cargo only pulls dev-deps (and their build scripts) into the build graph when test/bench/example targets are being compiled, and never propagates them to downstream consumers. If that holds, the fixture build would run on cargo test -p wash-runtime and stay silent on cargo build without us having to write any conditional logic.

We could also mark the ghost crate publish = false so there's no risk of it escaping this workspace by accident.

There's something else tho `CARGO_PRIMARY_PACKAGE` env var distinguishes primary vs transitive settings so maybe take a look there as well.

Honestly the "separate command" idea would totally work too, but it shifts the burden onto contributors (especially new ones) and CI to remember the extra step. (It's just a single command so realistically it's not much)
The dev-dep route, could let cargo test keep working the way it does today with no muscle-memory change. But again I'm  in favour of the separate command step too, ~~something like justfile perhaps?~~ (edit: just saw https://github.com/wasmCloud/wasmCloud/pull/5085 great stuff with xtask !! feel free to ignore my brain dump here 😅)

dman-os:
ooh, didn't know of CARGO_PRIMARY_PACKAGE. Also, the sibling package idea is very clever. Let me look into it.

ricochet:
superseded by #5220

PR #5009, opened 2026-03-29
feat(runtime-gateway): add path based routing

jamesstocktonj1:
# Description
- Add path based routing to runtime gateway
  - Add `path` key to host interfaces config
  - Match for path with longest matching prefix
  - If config key not present then routing works as before

## Example
Testing was done using the `deploy/k3s` and `examples/http-hello-world` example deployment.

## Scenario - Path Routing
Add `path` key to config:
```yaml diff
hostInterfaces:
  - namespace: wasi
    package: http
    interfaces:
      - incoming-handler
    config:
      host: localhost
      path: /hello
```

Matches to `/hello` and all other paths with prefix:
```sh
$ curl localhost:80/hello
Hello from wasmCloud!
$ curl localhost:80/hello/long/path
Hello from wasmCloud!
```
But still falls back if no path matches:
```sh
$ curl localhost:80/
No backend available
```

## Scenario - No Path Routing
Config with no path:
```yaml
hostInterfaces:
  - namespace: wasi
    package: http
    interfaces:
      - incoming-handler
    config:
      host: localhost
```
Matches with all paths, just as before:
```sh
$ curl localhost:80/hello
Hello from wasmCloud!
$ curl localhost:80/hello/long/path
Hello from wasmCloud!
$ curl localhost:80/
Hello from wasmCloud!
```

### DIFF
--- a/runtime-gateway/reconciler.go
+++ b/runtime-gateway/reconciler.go
@@ -34,10 +34,11 @@ func (a *WorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	}
 
 	var hostname string
-
+	var path string
 	for _, iface := range workload.Spec.HostInterfaces {
 		if iface.Namespace == "wasi" && iface.Package == "http" && slices.Contains(iface.Interfaces, "incoming-handler") {
 			if h, ok := iface.Config["host"]; ok {
+				path = iface.Config["path"]
 				hostname = h
 				break
 			}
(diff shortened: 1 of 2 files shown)


ricochet:
I see some issues that will be problematic for scale and perf. 

O(N) linear scan per request. `resolve()` iterates over the entire ht.hostnames map on every request. Before this PR, it was an O(1) map lookup. `hostnameKnown()` is also O(N), full iteration when the path-based resolve finds nothing. The read lock is held for the entire duration of these scans. Under high request rates with many routes, this extends the read-lock hold. There's no limit on how many path prefixes can be registered per hostname which could open the door for pathological attacks.

To avoid some of this, you could change this PR to have a trie/radix tree for path matching. The original expectation for end users for path based routing was to use an existing ingress/routing layer rather than building path routing into the gateway, which would delegate to a standard ingress controller (Traefik, Envoy, nginx) that already handles this efficiently with proper trie-based routing, path stripping, header rewriting, etc.

The gateway was meant to only do hostname→host resolution while the ingress handles path dispatch. But the need for this does make me wonder if we should consider some other alternatives. For example we could optionally add reconciliation for HTTPRoutes and replace the runtime-gateway component entirely with an existing Gateway API controller (Envoy Gateway, Traefik, etc.). The key challenge of why we have our reverse-proxy is that external gateway controllers don't know about wasmCloud Workloads. If we inject X-Workload-Id via HTTPRoute filter as part of reconciliation, it should work. The biggest downside with this approach is that it would add an external dep and more infra vs our self-contained approach.

vados-cosmonic:
Hey @jamesstocktonj1 is this still relevant? Happy to reopen if so, but given all the changes we've undertaken so far, I'd assume not.

Happy to re-open if how path based working works right now still doesn't suit your use case though!

PR #5011, opened 2026-03-31
Add Host pod reconciler

### LINKED ISSUE #5010: Operator reconciler takes approx 2 mins when a Host pod is terminated to reschedule Workloads
When a host pod that is running wasm workloads is terminated, a workload takes about 2 minutes for it to be deemend un-healthy (reconciliation loop 1) and then rescheduled to a new healthy host (reconciliation loop 2). The operator should watch the host pods so that if they do go down, it can react faster vs polling for unhealthy hosts.

jfleitz:
Fixes #5010

## Summary

Introduces a new `HostPodReconciler` that watches host Pods and immediately triggers `Host` CRD cleanup when a Pod is terminated, rather than waiting for the heartbeat polling cycle to detect the failure.

Previously, when a host Pod crashed or was evicted, the operator could take up to ~2 minutes to discover the failure (up to `hostReconcileInterval` for the next reconcile, plus `UnreachableTimeout` to declare the host unreachable). During testing, the recovery window was measured at **2.1 minutes**. With this change, workloads on a terminated host are reconciled in **~1 second**.

Adds finalizers to Workloads and Hosts to prevent orphaned entries within etcd. 
DeregisterHost was also modified to clean up registered workload entries for the deleted host.

### How it works

- `HostPodReconciler` watches Pods labeled `wasmcloud.com/hostgroup` and adds a finalizer to each one via a `MergeFrom` patch (no full-resource update required)
- When a Pod is deleted, the finalizer fires and immediately deletes the corresponding `Host` CRD by matching `pod.Status.PodIP` to `Host.Hostname` (the IP is sent as `req.Hostname` in the NATS heartbeat)
- Deleting the `Host` CRD triggers the existing `HostReconciler` finalizer, which deletes all `Workload` objects assigned to that host so `WorkloadReplicaSet` can schedule replacements immediately

### RBAC changes

Pod watch permissions are **namespace-scoped** — a new `Role` rule grants `get`, `list`, `patch`, `watch` on `pods` in the release namespace only. No new `ClusterRole` entry is needed. 

The gateway's `Host` and `Workload` finalizer mutations were also changed from full `Update` calls to `MergeFrom` patch, reducing the gateway ClusterRole to `patch` on those resources (no `update` required).

### Testing
Tested locally with a Kind cluster configured with the steps documented at wasmCloud.com, and deploying the Welcome workload manifest. The host pod was then deleted that the workload was running on, and the time to recovery was measured both pre and post change.

### DIFF
--- a/runtime-operator/internal/controller/runtime/artifact_controller.go
+++ b/runtime-operator/internal/controller/runtime/artifact_controller.go
@@ -87,8 +87,6 @@ func (r *ArtifactReconciler) finalize(ctx context.Context, artifact *runtimev1al
 
 	return nil
 }
-	return nil
-}
 
 // SetupWithManager sets up the controller with the Manager.
 // +kubebuilder:rbac:groups=runtime.wasmcloud.dev,resources=artifacts,verbs=get;list;watch;create;update;patch;delete
--- a/runtime-operator/internal/controller/runtime/host_controller.go
+++ b/runtime-operator/internal/controller/runtime/host_controller.go
@@ -191,7 +191,6 @@ func (h *hostStatusUpdater) Start(ctx context.Context) error {
 		}
 
 		// Status is a separate subresource; c.Update() (used by CreateOrUpdate)
-		// Status is a separate subresource; c.Update() (used by CreateOrUpdate)
 		// silently ignores status fields. Use Status().Patch() to persist
 		// LastSeen without conflicting with concurrent metadata changes (e.g.
 		// the ConditionedReconciler adding a finalizer between CreateOrUpdate
@@ -201,6 +200,7 @@ func (h *hostStatusUpdater) Start(ctx context.Context) error {
 		if err := h.client.Status().Patch(ctx, host, client.MergeFrom(base)); err != nil {
 			fmt.Println("Failed to update Host status:", err)
 		}
+	})
 
 	<-ctx.Done()
 	return subscription.Drain()


runtime-operator/cmd/main.go
```diff
@@ -210,6 +213,19 @@ func main() {
...
+	operatorNamespace := os.Getenv("OPERATOR_NAMESPACE")
+	if operatorNamespace != "" {
+		cacheOpts.ByObject = map[client.Object]cache.ByObject{
+			&corev1.Pod{}: {
+				Namespaces: map[string]cache.Config{
+					operatorNamespace: {},
+				},
+			},
+		}
+	}
```
ricochet, line 228: nit: `OPERATOR_NAMESPACE` is already read into `operatorCfg.Namespace` at line 141. Reuse it to avoid the duplicate `os.Getenv` call, and add a comment clarifying the `ByObject`/`DefaultNamespaces` interaction:

```suggestion
	// Restrict the Pod cache to the operator's own namespace so the cache
	// only requires a namespaced Role (not a ClusterRole) for Pod list/watch.
	// ByObject overrides DefaultNamespaces for the specified type, so Pods are
	// always scoped to the operator namespace regardless of -watch-namespaces.
	if operatorCfg.Namespace != "" {
		cacheOpts.ByObject = map[client.Object]cache.ByObject{
			&corev1.Pod{}: {
				Namespaces: map[string]cache.Config{
					operatorCfg.Namespace: {},
				},
			},
		}
	}
```

runtime-operator/internal/controller/runtime/artifact_controller.go
```diff
@@ -60,6 +61,31 @@ func (r *ArtifactReconciler) reconcileReady(_ context.Context, artifact *runtime
...
+			continue
+		}
+		for _, configArtifact := range deployment.Spec.Artifacts {
+			if configArtifact.ArtifactFrom.Name == artifact.Name {
+				return fmt.Errorf("artifact is still referenced by WorkloadDeployment %s/%s",
+					deployment.Namespace, deployment.Name)
+			}
+		}
+	}
+
```
ricochet: The finalizer error causes the `ConditionedReconciler` to requeue indefinitely (it returns before any conditions are updated), so operators get no status feedback about why deletion is stuck. Adding a structured log line at Info level makes this visible without introducing event recording infrastructure (which doesn't exist in this codebase yet).

```suggestion
func (r *ArtifactReconciler) finalize(ctx context.Context, artifact *runtimev1alpha1.Artifact) error {
	log := ctrl.LoggerFrom(ctx)

	deploymentList := &runtimev1alpha1.WorkloadDeploymentList{}
	if err := r.List(ctx, deploymentList, client.InNamespace(artifact.Namespace)); err != nil {
		return err
	}

	for _, deployment := range deploymentList.Items {
		if deployment.DeletionTimestamp != nil {
			continue
		}
		for _, configArtifact := range deployment.Spec.Artifacts {
			if configArtifact.ArtifactFrom.Name == artifact.Name {
				log.Info("artifact deletion blocked: still referenced by WorkloadDeployment",
					"deployment", fmt.Sprintf("%s/%s", deployment.Namespace, deployment.Name))
				return fmt.Errorf("artifact is still referenced by WorkloadDeployment %s/%s",
					deployment.Namespace, deployment.Name)
			}
		}
	}

	return nil
}
```

runtime-operator/internal/controller/runtime/host_controller.go
```diff
@@ -142,19 +170,31 @@ func (h *hostStatusUpdater) Start(ctx context.Context) error {
...
+			return
+		}
+
+		// Status is a separate subresource; c.Update() (used by CreateOrUpdate)
+		// silently ignores status fields. Use Status().Update() to persist LastSeen.
+		host.Status.LastSeen = metav1.Now()
+		if err := h.client.Status().Update(ctx, host); err != nil {
+			fmt.Println("Failed to update Host status:", err)
 		}
 	})
```
ricochet: `Status().Update()` requires a matching `resourceVersion`. If the `ConditionedReconciler` adds a finalizer between `CreateOrUpdate` and this call, the update gets a 409 Conflict. `MergeFrom` patch avoids this by sending only the diff.

```suggestion
		// Status is a separate subresource; c.Update() (used by CreateOrUpdate)
		// silently ignores status fields. Use Status().Patch() to persist
		// LastSeen without conflicting with concurrent metadata changes (e.g.
		// the ConditionedReconciler adding a finalizer between CreateOrUpdate
		// and this call).
		base := host.DeepCopy()
		host.Status.LastSeen = metav1.Now()
		if err := h.client.Status().Patch(ctx, host, client.MergeFrom(base)); err != nil {
			fmt.Println("Failed to update Host status:", err)
		}
```

runtime-operator/internal/controller/runtime/host_pod_controller.go
```diff
@@ -0,0 +1,134 @@
...
+		For(&corev1.Pod{}, builder.WithPredicates(
+			// Only enqueue Pods that carry the HostPodLabel — avoids processing
+			// every Pod in the cluster.
+			predicate.NewPredicateFuncs(func(obj client.Object) bool {
+				if r.Namespace != "" && obj.GetNamespace() != r.Namespace {
+					return false
+				}
+				_, ok := obj.GetLabels()[HostPodLabel]
+				return ok
+			}),
```
ricochet, line 128: The namespace check here is redundant — the Pod cache is already restricted to the operator namespace via `ByObject` in `cmd/main.go`. When `Namespace` is empty, all namespaces are watched (by design), so the predicate filter wouldn't help either. Simplify to just the label check:

```suggestion
			// Only enqueue Pods that carry the HostPodLabel — avoids processing
			// every Pod in the namespace. Namespace scoping is handled by the
			// cache (ByObject in cmd/main.go), so no namespace check is needed here.
			predicate.NewPredicateFuncs(func(obj client.Object) bool {
				_, ok := obj.GetLabels()[HostPodLabel]
				return ok
			}),
```

PR #5012, opened 2026-03-31
feat: upgrade to wasmtime 43

ricochet:
- Workspace Cargo.toml - Bumped wasmtime deps from 42 to 43:
 - wasmtime, wasmtime-wasi, wasmtime-wasi-io, wasmtime-wasi-http all to version 43
- wasi-gfx crates - Updated git rev to latest which includes
  wasmtime 43 and wgpu 29 support (no crates.io releases exist; still git deps)
- wasmtime-wasi-http restructuring - Major module reorganization in v43

### DIFF
--- a/crates/wash-runtime/src/sockets/udp.rs
+++ b/crates/wash-runtime/src/sockets/udp.rs
@@ -492,3 +492,191 @@ impl UdpSocket {
         }
     }
 }
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+mod tests {
+    use super::*;
+    use crate::sockets::WasiSocketsCtx;
+    use cap_net_ext::AddressFamily;
+
+    fn make_ipv4_socket() -> NetworkUdpSocket {
+        let ctx = WasiSocketsCtx::default();
+        NetworkUdpSocket::new(&ctx, AddressFamily::Ipv4).unwrap()
+    }
+
+    fn bind_socket(socket: &mut NetworkUdpSocket) {
+        let addr: std::net::SocketAddr = "127.0.0.1:0".parse().unwrap();
+        socket.bind(addr).unwrap();
+        socket.finish_bind().unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_new_socket_default_state() {
+        let socket = make_ipv4_socket();
+        assert!(!socket.is_bound());
+        assert!(!socket.is_connected());
+    }
+
+    #[tokio::test]
+    async fn test_bind_and_finish_bind() {
+        let mut socket = make_ipv4_socket();
+        let addr: std::net::SocketAddr = "127.0.0.1:0".parse().unwrap();
+
+        socket.bind(addr).unwrap();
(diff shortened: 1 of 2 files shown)


crates/wash-runtime/src/sockets/host_udp.rs
```diff
@@ -129,25 +129,6 @@ impl udp::HostUdpSocket for WasiSocketsCtxView<'_> {
...
-        // - To leave our socket instance in a consistent state in case the
-        //   connect fails.
-        // - When reconnecting to a different address, Linux sometimes fails
-        //   if there isn't a disconnect in between.
-
-        // Step #1: Disconnect
-        if socket.is_connected() {
-            let mut loopback = self
-                .ctx
-                .loopback
```
vados-cosmonic, line 142: Did we ever have something explicit to test this?  A bit worried that this was load bearing (especially on linux in particular)

crates/wash-runtime/src/sockets/udp.rs
```diff
@@ -139,28 +138,26 @@ impl NetworkUdpSocket {
             _ => return Err(ErrorCode::InvalidState),
         }
 
-        // We disconnect & (re)connect in two distinct steps for two reasons:
-        // - To leave our socket instance in a consistent state in case the
-        //   connect fails.
-        // - When reconnecting to a different address, Linux sometimes fails
-        //   if there isn't a disconnect in between.
```
vados-cosmonic, line 146: Same note about whether this is load bearing -- if it is I think this comment is very useful

crates/wash-runtime/src/sockets/host_udp.rs
```diff
@@ -164,6 +145,15 @@ impl udp::HostUdpSocket for WasiSocketsCtxView<'_> {
             socket
                 .connect(connect_addr, &mut loopback)
                 .map_err(super::network::socket_error_from_util)?;
+        } else if socket.is_connected() {
+            let mut loopback = self
+                .ctx
+                .loopback
+                .lock()
+                .map_err(|e| SocketError::trap(wasmtime::format_err!("{e}")))?;
```
vados-cosmonic, line 153:
```suggestion
                .map_err(|e| SocketError::trap(wasmtime::format_err!("failed to lock loopback: {e}")))?;
```

What do you think about a descriptive error here?

crates/wash-runtime/src/sockets/udp.rs
ricochet: The new udp_connect() in util.rs handles the Linux case and also has a comment explaining. This matches upstream wasmtime 43 exactly.

crates/wash-runtime/src/sockets/host_udp.rs
ricochet: This is a good callout. We were mostly betting on upstream correctness and tests here, but it is much safer for us to run our own suite. I added unit tests with complete coverage for UDP.

PR #5013, opened 2026-03-31
ci(wash): add winget package

ricochet:
This takes the artifact we're already building, zips it (required for winget) and uploads it to winget repo

### DIFF
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -208,11 +208,25 @@ jobs:
           ${{ matrix.buildCommand }} --release --target ${{ matrix.target }}
           cp target/${{ matrix.target }}/release/${{ matrix.bin }} ${{ matrix.artifact }}
 
+      - name: Create Windows zip archive
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          Compress-Archive -Path "target/${{ matrix.target }}/release/${{ matrix.bin }}" -DestinationPath "${{ matrix.artifact }}.zip"
+
       - name: Upload artifact
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/wash.yml
```diff
@@ -208,13 +208,27 @@ jobs:
           ${{ matrix.buildCommand }} --release --target ${{ matrix.target }}
           cp target/${{ matrix.target }}/release/${{ matrix.bin }} ${{ matrix.artifact }}
 
+      - name: Create Windows zip archive
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          Compress-Archive -Path "target/${{ matrix.target }}/release/${{ matrix.bin }}" -DestinationPath "${{ matrix.artifact }}.zip"
```
vados-cosmonic, line 215: A great follow up would be some sort of sanity check for this artifact -- a quick `winget install` ?

ricochet: Lil chicken and egg. We need to get this one published for the first time then we can add a winget install check.

vados-cosmonic: Ah so I was hoping we could do a local install first, if that's possible -- i.e. with the zip file we're creating here. That said, if it's not easy, this is definitely worth doing in a follow up, not here

[EDIT] Asking AI, it looks like winget doesn't really support instalilng from zip? I guess you have to unzip and then run msiexec against the `.msi`/`.exe` inside? The Windows ecosystem never ceases to amaze me.

ricochet: It's baffling yeah. I also tried a bit of searching and essentially it'd be just validating the exe we just zipped and tested with. I'd much rather do something that does more to validate the package install itself works with winget.

ricochet: Added validation as a follow-up in #5017

PR #5014, opened 2026-04-01
feat(http): add OutgoingHandler trait for customizable outgoing request handling

### LINKED ISSUE #4978: [FEATURE] Simplify `outgoing_request` configuration for custom send logic
Currently, `HttpServer` implements `outgoing_request` as follows:

```rust
fn outgoing_request(
    &self,
    workload_id: &str,
    request: hyper::Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
    config: wasmtime_wasi_http::types::OutgoingRequestConfig,
) -> wasmtime_wasi_http::HttpResult<wasmtime_wasi_http::types::HostFutureIncomingResponse> {
    self.router
        .allow_outgoing_request(workload_id, &request, &config)
        .map_err(|e| {
            wasmtime_wasi_http::HttpError::trap(anyhow::anyhow!(
                "request not allowed: {}",
                e
            ))
        })?;

    if is_grpc_request(&request) {
        Ok(send_grpc_request(request, config))
    } else {
        Ok(wasmtime_wasi_http::types::default_send_request(
            request, config,
        ))
    }
}
```

This implementation directly relies on default_send_request, which makes it difficult to customize the underlying HTTP client behavior.

I propose refactoring HttpServer to delegate outgoing request handling to a dedicated abstraction:

```rust
pub struct HttpServer<T: Router, O: OutgoingHandler = WasiOutgoingHandler> {
    router: Arc<T>,
    outgoing_handler: O,
    addr: SocketAddr,
    workload_handles: WorkloadHandles,
    shutdown_tx: Arc<RwLock<Option<mpsc::Sender<()>>>>,
    tls_acceptor: Option<TlsAcceptor>,
    listener: Arc<tokio::sync::Mutex<Option<TcpListener>>>,
}

pub trait OutgoingHandler {
    fn send(
        &self,
        request: hyper::Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
        config: wasmtime_wasi_http::types::OutgoingRequestConfig,
    ) -> wasmtime_wasi_http::types::HostFutureIncomingResponse;
}
```

This change would allow injecting a custom implementation of `OutgoingHandler`, enabling more flexible control over outgoing HTTP behavior.

The primary issue I am facing is the need to configure custom TLS certificates for HTTPS requests. With the current implementation based on `default_send_request`, there is no straightforward way to provide a custom root CA store or client certificates. Introducing an `OutgoingHandler` abstraction would make this possible without tightly coupling HttpServer to a specific transport implementation.

Aditya1404Sal:
### Related Issue/PR
Closes #4978

### Why it's needed

The current `outgoing_request` implementation is tightly coupled to `default_send_request`, which limits flexibility in how outgoing http requests are handled.

This makes it difficult to support use cases like custom TLS configuration or alternative transport behavior.

By introducing the `OutgoingHandler` abstraction, outgoing request handling becomes pluggable, enabling custom egress logic without modifying the HTTP server itself.

### DIFF
--- a/crates/wash-runtime/README.md
+++ b/crates/wash-runtime/README.md
@@ -39,7 +39,7 @@ async fn main() -> anyhow::Result<()> {
 
     // Configure plugins
     let http_router = DynamicRouter::default();
-    let http_handler = HttpServer::new(http_router, WasiOutgoingHandler, "127.0.0.1:8080".parse()?).await?;
+    let http_handler = HttpServer::new(http_router, "127.0.0.1:8080".parse()?).await?;
     let wasi_config_plugin = DynamicConfig::default();
 
     // Build and start the host
--- a/crates/wash-runtime/README.md
+++ b/crates/wash-runtime/README.md
@@ -24,7 +24,7 @@ use std::collections::HashMap;
 use wash_runtime::{
     engine::Engine,
     host::{HostBuilder, HostApi,
-      http::{HttpServer, DynamicRouter},
+      http::{HttpServer, DynamicRouter, WasiOutgoingHandler},
   },
     plugin::{
         wasi_config::DynamicConfig,
@@ -39,7 +39,7 @@ async fn main() -> anyhow::Result<()> {
 
     // Configure plugins
     let http_router = DynamicRouter::default();
-    let http_handler = HttpServer::new(http_router, "127.0.0.1:8080".parse()?).await?;
+    let http_handler = HttpServer::new(http_router, WasiOutgoingHandler, "127.0.0.1:8080".parse()?).await?;
     let wasi_config_plugin = DynamicConfig::default();
 
     // Build and start the host
--- a/crates/wash-runtime/README.md
+++ b/crates/wash-runtime/README.md
@@ -24,7 +24,7 @@ use std::collections::HashMap;
 use wash_runtime::{
     engine::Engine,
     host::{HostBuilder, HostApi,
-      http::{HttpServer, DynamicRouter, WasiOutgoingHandler},
+      http::{HttpServer, DynamicRouter},
   },
     plugin::{
         wasi_config::DynamicConfig,
--- a/crates/wash-runtime/src/host/http.rs
+++ b/crates/wash-runtime/src/host/http.rs
@@ -486,7 +486,7 @@ impl<T: Router, O: OutgoingHandler> std::fmt::Debug for HttpServer<T, O> {
     }
 }
 
-impl<T: Router, O: OutgoingHandler> HttpServer<T, O> {
+impl<T: Router> HttpServer<T, WasiOutgoingHandler> {
     /// Creates a new HTTP server that eagerly binds to the specified address.
     ///
     /// The socket is bound immediately so the port is reserved. Use port `0`
@@ -499,8 +499,82 @@ impl<T: Router, O: OutgoingHandler> HttpServer<T, O> {
     /// # Arguments
     /// * `router` - The router implementation for handling requests
     /// * `addr` - The socket address to bind to
-    pub async fn new(router: T, outgoing_handler: O, addr: SocketAddr) -> anyhow::Result<Self> {
+    pub async fn new(router: T, addr: SocketAddr) -> anyhow::Result<Self> {
         crate::init_crypto();
+        let listener = TcpListener::bind(addr).await?;
+        let addr = listener.local_addr()?;
+        Ok(Self {
+            router: Arc::new(router),
+            outgoing_handler: WasiOutgoingHandler,
+            addr,
+            workload_handles: Arc::default(),
+            shutdown_tx: Arc::new(RwLock::new(None)),
+            tls_acceptor: None,
+            listener: Arc::new(tokio::sync::Mutex::new(Some(listener))),
+            meters: Default::default(),
+        })
+    }
+
+    /// Creates a new HTTPS server with TLS support.
+    ///
+    /// # Arguments
+    /// * `router` - The router implementation for handling requests
+    /// * `addr` - The socket address to bind to
+    /// * `cert_path` - Path to the TLS certificate file
+    /// * `key_path` - Path to the private key file
+    /// * `ca_path` - Optional path to CA certificate for mutual TLS
+    ///
+    /// # Returns
+    /// A new `HttpServer` instance configured for HTTPS connections.
+    ///
+    /// # Errors
+    /// Returns an error if the TLS configuration cannot be loaded.
+    pub async fn new_with_tls(
+        router: T,
+        addr: SocketAddr,
+        cert_path: &Path,
+        key_path: &Path,
+        ca_path: Option<&Path>,
+    ) -> anyhow::Result<Self> {
+        let tls_config = load_tls_config(cert_path, key_path, ca_path).await?;
+        let tls_acceptor = TlsAcceptor::from(Arc::new(tls_config));
+
+        let listener = TcpListener::bind(addr).await?;
+        let addr = listener.local_addr()?;
+        Ok(Self {
+            router: Arc::new(router),
+            outgoing_handler: WasiOutgoingHandler,
+            addr,
+            workload_handles: Arc::default(),
+            shutdown_tx: Arc::new(RwLock::new(None)),
+            tls_acceptor: Some(tls_acceptor),
+            listener: Arc::new(tokio::sync::Mutex::new(Some(listener))),
+            meters: Default::default(),
+        })
+    }
+}
+
+impl<T: Router, O: OutgoingHandler> HttpServer<T, O> {
+    /// Returns the actual bound address (useful when binding to port 0).
+    pub fn addr(&self) -> SocketAddr {
+        self.addr
+    }
+
+    /// Creates a new HTTP server with a custom outgoing request handler.
+    ///
+    /// Use this when you need to customise egress behaviour (e.g. custom TLS
+    /// roots, proxying, or mock handlers in tests). For the common case use
+    /// [`new`](HttpServer::new) instead, which defaults to [`WasiOutgoingHandler`].
+    ///
+    /// # Arguments
+    /// * `router` - The router implementation for handling requests
+    /// * `outgoing_handler` - Custom handler for outgoing HTTP requests
+    /// * `addr` - The socket address to bind to
+    pub async fn new_with_outgoing_handler(
+        router: T,
+        outgoing_handler: O,
+        addr: SocketAddr,
+    ) -> anyhow::Result<Self> {
         let listener = TcpListener::bind(addr).await?;
         let addr = listener.local_addr()?;
         Ok(Self {
@@ -515,30 +589,25 @@ impl<T: Router, O: OutgoingHandler> HttpServer<T, O> {
         })
     }
 
-    /// Returns the actual bound address (useful when binding to port 0).
-    pub fn addr(&self) -> SocketAddr {
-        self.addr
-    }
-
-    /// Creates a new HTTPS server with TLS support.
+    /// Creates a new HTTPS server with TLS support and a custom outgoing request handler.
+    ///
+    /// Use this when you need both TLS and custom egress behaviour. For the
+    /// common case use [`new_with_tls`](HttpServer::new_with_tls) instead.
     ///
     /// Side effect: installs the process-level rustls crypto provider via
     /// [`crate::init_crypto`] (idempotent).
     ///
     /// # Arguments
     /// * `router` - The router implementation for handling requests
-    /// * `outgoing_handler` - The handler responsible for outgoing HTTP requests
+    /// * `outgoing_handler` - Custom handler for outgoing HTTP requests
     /// * `addr` - The socket address to bind to
     /// * `cert_path` - Path to the TLS certificate file
     /// * `key_path` - Path to the private key file
     /// * `ca_path` - Optional path to CA certificate for mutual TLS
     ///
-    /// # Returns
-    /// A new `HttpServer` instance configured for HTTPS connections.
-    ///
     /// # Errors
     /// Returns an error if the TLS configuration cannot be loaded.
-    pub async fn new_with_tls(
+    pub async fn new_with_tls_and_outgoing_handler(
         router: T,
         outgoing_handler: O,
         addr: SocketAddr,
--- a/crates/wash-runtime/src/host/http.rs
+++ b/crates/wash-runtime/src/host/http.rs
@@ -142,6 +142,16 @@ pub trait Router: Send + Sync + 'static {
         _allowed_hosts: &[String],
     ) -> anyhow::Result<()>;
 
+    /// Determine if a P3 outgoing request is allowed.
+    #[cfg(feature = "wasip3")]
+    fn allow_outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        request: &hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        _options: &Option<wasmtime_wasi_http::p3::RequestOptions>,
+        allowed_hosts: &[String],
+    ) -> anyhow::Result<()>;
+
     /// Pick a workload ID based on the incoming request.
     ///
     /// On failure, the returned [`RouteError`] determines the HTTP status
@@ -251,6 +261,17 @@ impl Router for DynamicRouter {
         check_allowed_hosts(request, allowed_hosts)
     }
 
+    #[cfg(feature = "wasip3")]
+    fn allow_outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        request: &hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        _options: &Option<wasmtime_wasi_http::p3::RequestOptions>,
+        allowed_hosts: &[String],
+    ) -> anyhow::Result<()> {
+        check_allowed_hosts(request, allowed_hosts)
+    }
+
     /// Pick a workload ID based on the incoming request
     fn route_incoming_request(
         &self,
@@ -283,15 +304,41 @@ impl Router for DynamicRouter {
     }
 }
 
-/// Trait defining the behavior for handling outgoing HTTP requests
-/// Allows for custom egress logic such as transport, TLS configuration, or protocol handling
-/// Use this trait to override the default HTTP client behavior in the HTTP server
+/// Trait defining the behavior for handling outgoing HTTP requests.
+///
+/// Allows for custom egress logic such as transport, TLS configuration, or
+/// protocol handling. Use this trait to override the default HTTP client
+/// behavior in the HTTP server. A single implementor can serve both P2 and
+/// P3 components — [`send`](Self::send) handles P2 outgoing requests and
+/// [`send_p3`](Self::send_p3) handles P3 outgoing requests, with a sensible
+/// default that defers to `wasmtime_wasi_http::p3::default_send_request`.
 pub trait OutgoingHandler: Send + Sync + 'static {
+    /// Send an outgoing request from a P2 component.
     fn send(
         &self,
         _request: hyper::Request<wasmtime_wasi_http::p2::body::HyperOutgoingBody>,
         _config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
     ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>;
+
+    /// Send an outgoing request from a P3 component.
+    ///
+    /// The default implementation defers to
+    /// `wasmtime_wasi_http::p3::default_send_request`. Override this when your
+    /// custom egress policy applies to P3 components as well.
+    #[cfg(feature = "wasip3")]
+    fn send_p3(
+        &self,
+        request: hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        _fut: crate::host::http_p3::P3RequestErrorFuture,
+    ) -> crate::host::http_p3::P3SendFuture {
+        Box::new(async move {
+            use http_body_util::BodyExt;
+            let (res, io) = wasmtime_wasi_http::p3::default_send_request(request, options).await?;
+            let io: crate::host::http_p3::P3RequestErrorFuture = Box::new(io);
+            Ok((res.map(BodyExt::boxed_unsync), io))
+        })
+    }
 }
 
 #[derive(Default)]
@@ -358,6 +405,17 @@ impl Router for DevRouter {
         Ok(())
     }
 
+    #[cfg(feature = "wasip3")]
+    fn allow_outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        _request: &hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        _options: &Option<wasmtime_wasi_http::p3::RequestOptions>,
+        _allowed_hosts: &[String],
+    ) -> anyhow::Result<()> {
+        Ok(())
+    }
+
     /// Pick a workload ID based on the incoming request
     fn route_incoming_request(
         &self,
@@ -407,6 +465,17 @@ pub trait HostHandler: Send + Sync + 'static {
         config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
         allowed_hosts: &[String],
     ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>;
+
+    /// Handle an outgoing HTTP request from a P3 workload.
+    #[cfg(feature = "wasip3")]
+    fn outgoing_request_p3(
+        &self,
+        workload_id: &str,
+        request: hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        fut: crate::host::http_p3::P3RequestErrorFuture,
+        allowed_hosts: &[String],
+    ) -> crate::host::http_p3::P3SendFuture;
 }
 
 impl std::fmt::Debug for dyn HostHandler {
@@ -455,6 +524,23 @@ impl HostHandler for NullServer {
         Err(wasmtime_wasi_http::p2::HttpError::trap(
             wasmtime::format_err!("http client not available"),
         ))
+    }
+
+    #[cfg(feature = "wasip3")]
+    fn outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        _request: hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        _options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        _fut: crate::host::http_p3::P3RequestErrorFuture,
+        _allowed_hosts: &[String],
+    ) -> crate::host::http_p3::P3SendFuture {
+        use wasmtime_wasi_http::p3::bindings::http::types::ErrorCode;
+        Box::new(async move {
+            Err(wasmtime_wasi::TrappableError::from(
+                ErrorCode::InternalError(Some("http client not available".to_string())),
+            ))
+        })
     }
 }
 
@@ -542,8 +628,11 @@ impl<T: Router> HttpServerBuilder<T, WasiOutgoingHandler> {
             tls: None,
         }
     }
+}
 
+impl<T: Router, O: OutgoingHandler> HttpServerBuilder<T, O> {
     /// Set a custom [`OutgoingHandler`], changing the builder's handler type.
+    /// The same handler serves both P2 and P3 outgoing requests.
     pub fn outgoing_handler<O2: OutgoingHandler>(self, handler: O2) -> HttpServerBuilder<T, O2> {
         HttpServerBuilder {
             router: self.router,
@@ -552,9 +641,7 @@ impl<T: Router> HttpServerBuilder<T, WasiOutgoingHandler> {
             tls: self.tls,
         }
     }
-}
 
-impl<T: Router, O: OutgoingHandler> HttpServerBuilder<T, O> {
     /// Enable TLS with the given certificate, key, and optional CA paths.
     pub fn tls(mut self, cert_path: &Path, key_path: &Path, ca_path: Option<&Path>) -> Self {
         self.tls = Some(TlsConfig {
@@ -737,6 +824,30 @@ impl<T: Router, O: OutgoingHandler> HostHandler for HttpServer<T, O> {
             ));
         }
         self.outgoing_handler.send(request, config)
+    }
+
+    #[cfg(feature = "wasip3")]
+    fn outgoing_request_p3(
+        &self,
+        workload_id: &str,
+        request: hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        fut: crate::host::http_p3::P3RequestErrorFuture,
+        allowed_hosts: &[String],
+    ) -> crate::host::http_p3::P3SendFuture {
+        if let Err(e) =
+            self.router
+                .allow_outgoing_request_p3(workload_id, &request, &options, allowed_hosts)
+        {
+            warn!(workload_id = %workload_id, err = %e, "P3 outgoing request denied by allowed_hosts policy");
+            use wasmtime_wasi_http::p3::bindings::http::types::ErrorCode;
+            return Box::new(async move {
+                Err(wasmtime_wasi::TrappableError::from(
+                    ErrorCode::HttpRequestDenied,
+                ))
+            });
+        }
+        self.outgoing_handler.send_p3(request, options, fut)
     }
 }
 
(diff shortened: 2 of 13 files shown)


ricochet:
We might need a builder. Worth considering what this looks like with p3 in the mix: https://github.com/wasmCloud/wasmCloud/pull/5022

crates/wash-runtime/src/host/http.rs
```diff
@@ -440,64 +568,124 @@ pub struct HttpServer<T: Router> {
...
+        self.tls = Some(TlsConfig {
+            cert_path: cert_path.to_path_buf(),
+            key_path: key_path.to_path_buf(),
+            ca_path: ca_path.map(Path::to_path_buf),
+        });
+        self
+    }
+
+    /// Bind to the address and build the [`HttpServer`].
+    pub async fn build(self) -> anyhow::Result<HttpServer<T, O>> {
```
MendyBerger, line 691: Shouldn't `init_crypto` be called in here?

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +304,63 @@ impl Router for DynamicRouter {
...
+}
+
+#[derive(Default)]
+pub struct WasiOutgoingHandler;
+
+impl OutgoingHandler for WasiOutgoingHandler {
+    fn send(
+        &self,
+        request: hyper::Request<wasmtime_wasi_http::p2::body::HyperOutgoingBody>,
+        config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
```
MendyBerger, line 338: Probably want to get the workload_id here

crates/wash-runtime/src/host/http_p3.rs
```diff
@@ -12,7 +17,22 @@ use wasmtime::component::InstancePre;
 use wasmtime_wasi_http::p3::bindings::ServicePre;
 use wasmtime_wasi_http::p3::bindings::http::types::ErrorCode;
 
-type P3Body = http_body_util::combinators::UnsyncBoxBody<bytes::Bytes, ErrorCode>;
+/// Body type used on the P3 outgoing path (also used as the return-body type
+/// for [`handle_component_request_p3`]).
+pub type P3OutgoingBody = http_body_util::combinators::UnsyncBoxBody<bytes::Bytes, ErrorCode>;
```
MendyBerger: Why the rename? This is not only used for outgoing

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +304,63 @@ impl Router for DynamicRouter {
...
+            use http_body_util::BodyExt;
+            let (res, io) = wasmtime_wasi_http::p3::default_send_request(request, options).await?;
+            let io: crate::host::http_p3::P3RequestErrorFuture = Box::new(io);
+            Ok((res.map(BodyExt::boxed_unsync), io))
+        })
+    }
+}
+
+#[derive(Default)]
+pub struct WasiOutgoingHandler;
```
MendyBerger: This is more than just wasi + grpc. Maybe just call it something like `DefaultOutgoingHandler`?

crates/wash-runtime/src/host/http.rs
```diff
@@ -370,6 +459,17 @@ pub trait HostHandler: Send + Sync + 'static {
...
+    /// Handle an outgoing HTTP request from a P3 workload.
+    #[cfg(feature = "wasip3")]
+    fn outgoing_request_p3(
+        &self,
+        workload_id: &str,
+        request: hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        fut: crate::host::http_p3::P3RequestErrorFuture,
+        allowed_hosts: &[String],
+    ) -> crate::host::http_p3::P3SendFuture;
```
MendyBerger: This is a new required method on an existing trait, making it a breaking change. Is there a way we can avoid it?

crates/wash-runtime/src/host/http.rs
```diff
@@ -440,64 +568,124 @@ pub struct HttpServer<T: Router> {
...
+        HttpServerBuilder {
+            router: self.router,
+            outgoing_handler: handler,
+            addr: self.addr,
+            tls: self.tls,
+        }
+    }
+
+    /// Enable TLS with the given certificate, key, and optional CA paths.
+    pub fn tls(mut self, cert_path: &Path, key_path: &Path, ca_path: Option<&Path>) -> Self {
```
MendyBerger: There are multiple `path`s here so easy to get confused. Maybe the args can go in a struct so it's harder to get confused?

crates/wash-runtime/src/host/http.rs
MendyBerger: Could use a doc comment

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +304,63 @@ impl Router for DynamicRouter {
...
+    ///
+    /// The default implementation defers to
+    /// `wasmtime_wasi_http::p3::default_send_request`. Override this when your
+    /// custom egress policy applies to P3 components as well.
+    #[cfg(feature = "wasip3")]
+    fn send_p3(
+        &self,
+        request: hyper::Request<crate::host::http_p3::P3OutgoingBody>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        _fut: crate::host::http_p3::P3RequestErrorFuture,
```
MendyBerger: Please add comment why it's okay to ignore `_fut` in this specific case

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +304,63 @@ impl Router for DynamicRouter {
...
+pub struct WasiOutgoingHandler;
+
+impl OutgoingHandler for WasiOutgoingHandler {
+    fn send(
+        &self,
+        request: hyper::Request<wasmtime_wasi_http::p2::body::HyperOutgoingBody>,
+        config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
+    ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>
+    {
+        if is_grpc_request(&request) {
```
MendyBerger: Worth noting that custom `OutgoingHandler` users will lose grpc. Are we cool with that?

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +304,63 @@ impl Router for DynamicRouter {
...
+/// behavior in the HTTP server. A single implementor can serve both P2 and
+/// P3 components — [`send`](Self::send) handles P2 outgoing requests and
+/// [`send_p3`](Self::send_p3) handles P3 outgoing requests, with a sensible
+/// default that defers to `wasmtime_wasi_http::p3::default_send_request`.
+pub trait OutgoingHandler: Send + Sync + 'static {
+    /// Send an outgoing request from a P2 component.
+    fn send(
+        &self,
+        _request: hyper::Request<wasmtime_wasi_http::p2::body::HyperOutgoingBody>,
+        _config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
```
MendyBerger: Probably want to get the workload_id here

MendyBerger: No need to lead with underscore here

crates/wash-runtime/README.md
```diff
@@ -24,7 +24,7 @@ use std::collections::HashMap;
 use wash_runtime::{
     engine::Engine,
     host::{HostBuilder, HostApi,
-      http::{HttpServer, DynamicRouter},
+      http::{HttpServer, DynamicRouter, WasiOutgoingHandler},
```
MendyBerger: What's this about?

MendyBerger (changes requested):
Hey @Aditya1404Sal thanks for the PR!

crates/wash-runtime/README.md
Aditya1404Sal: Ah a stray change from a past commit, I'll remove it

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: Great idea!

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: That's right, misplaced it in the initialising new() & new_with_tls()

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: Yes, I think I can avoid that

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: will pack it into their struct

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: Hmm, probably best to make that configurable somehow, let me take a look

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: P2 & P3 types are distinct, we'll have to keep the method but I could add the default impl making it non-breaking at the very least

crates/wash-runtime/src/host/http.rs
```diff
@@ -370,6 +456,24 @@ pub trait HostHandler: Send + Sync + 'static {
...
+    #[cfg(feature = "wasip3")]
+    fn outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        request: hyper::Request<crate::host::http_p3::P3Body>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        // Response-side body-error sink: unused here because hyper's response
+        // body already reports body errors through its `Stream` impl.
+        _fut: crate::host::http_p3::P3RequestErrorFuture,
+        _allowed_hosts: &[String],
```
ricochet: _allowed_hosts not checked

crates/wash-runtime/src/host/http.rs
```diff
@@ -142,6 +142,16 @@ pub trait Router: Send + Sync + 'static {
...
 
+    /// Determine if a P3 outgoing request is allowed.
+    #[cfg(feature = "wasip3")]
+    fn allow_outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        request: &hyper::Request<crate::host::http_p3::P3Body>,
+        _options: &Option<wasmtime_wasi_http::p3::RequestOptions>,
+        allowed_hosts: &[String],
+    ) -> anyhow::Result<()>;
```
ricochet:
```suggestion
    ) -> anyhow::Result<()> {
        check_allowed_hosts(request, allowed_hosts)
    }
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +304,60 @@ impl Router for DynamicRouter {
...
+        config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
+    ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>;
+
+    #[cfg(feature = "wasip3")]
+    fn send_request_p3(
+        &self,
+        _workload_id: &str,
+        request: hyper::Request<crate::host::http_p3::P3Body>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        _fut: crate::host::http_p3::P3RequestErrorFuture,
```
ricochet:
```suggestion
        fut: crate::host::http_p3::P3RequestErrorFuture,
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +304,60 @@ impl Router for DynamicRouter {
...
+        &self,
+        workload_id: &str,
+        request: hyper::Request<wasmtime_wasi_http::p2::body::HyperOutgoingBody>,
+        config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
+    ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>;
+
+    #[cfg(feature = "wasip3")]
+    fn send_request_p3(
+        &self,
+        _workload_id: &str,
```
ricochet:
```suggestion
        workload_id: &str,
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -370,6 +456,33 @@ pub trait HostHandler: Send + Sync + 'static {
         config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
         allowed_hosts: &[String],
     ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>;
+
```
ricochet, line 466:
```suggestion
    /// Handle a P3 outgoing request, enforcing `allowed_hosts` policy and
    /// delegating transport to [`wasmtime_wasi_http::p3::default_send_request`].
    ///
    /// Override to apply custom egress logic (e.g. alternate transports or
    /// per-workload TLS configuration) while still honouring the allowlist via
    /// [`check_allowed_hosts`].
```

ricochet: add doc comment

crates/wash-runtime/src/host/http.rs
ricochet: _workload_id → workload_id, _fut → fut in the trait declaration

 trait methods should never have underscores. OK in default concrete implementations.

crates/wash-runtime/src/host/http.rs
ricochet: Add a default impl calling `check_allowed_hosts` to make it a non-breaking change for existing Router implementors.

ricochet (changes requested):

ricochet:
@if0ne I know you were looking for something similar for your use-case. Does this change work for you?

if0ne:
> @if0ne I know you were looking for something similar for your use-case. Does this change work for you?

Yes, these changes helped me out

crates/wash-runtime/src/host/http.rs
```diff
@@ -425,6 +546,9 @@ impl HostHandler for NullServer {
             wasmtime::format_err!("http client not available"),
         ))
     }
+    // P3 outgoing requests fall through to the trait default
+    // (wasmtime_wasi_http::p3::default_send_request). NullServer has nothing
+    // workload-specific to enforce.
```
ricochet:
```suggestion
    #[cfg(feature = "wasip3")]
    fn outgoing_request_p3(
        &self,
        _workload_id: &str,
        _request: hyper::Request<crate::host::http_p3::P3Body>,
        _options: Option<wasmtime_wasi_http::p3::RequestOptions>,
        _fut: crate::host::http_p3::P3RequestErrorFuture,
        _allowed_hosts: &[String],
    ) -> crate::host::http_p3::P3SendFuture {
        use wasmtime_wasi_http::p3::bindings::http::types::ErrorCode;
        Box::new(async {
            Err(wasmtime_wasi::TrappableError::from(ErrorCode::InternalError(
                Some("http client not available".to_string()),
            )))
        })
    }
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -446,91 +581,153 @@ pub struct HttpServer<T: Router> {
...
-        let addr = listener.local_addr()?;
-        Ok(Self {
-            router: Arc::new(router),
+/// TLS configuration for [`HttpServerBuilder::tls`] / [`HttpServer::new_with_tls`].
+#[derive(Debug, Clone)]
+pub struct TlsConfig {
+    pub cert_path: std::path::PathBuf,
+    pub key_path: std::path::PathBuf,
+    pub ca_path: Option<std::path::PathBuf>,
+}
```
ricochet, line 611:
```suggestion
pub struct TlsConfig {
    cert_path: std::path::PathBuf,
    key_path: std::path::PathBuf,
    ca_path: Option<std::path::PathBuf>,
}
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +306,60 @@ impl Router for DynamicRouter {
...
+/// handled by the runtime before this trait is called.
+pub trait OutgoingHandler: Send + Sync + 'static {
+    fn send_request(
+        &self,
+        workload_id: &str,
+        request: hyper::Request<wasmtime_wasi_http::p2::body::HyperOutgoingBody>,
+        config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
+    ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>;
+
+    #[cfg(feature = "wasip3")]
```
ricochet, line 318:
```suggestion
    /// Send a P3 outgoing HTTP request for the given `workload_id`.
    ///
    /// `fut` is a future provided by the WASI runtime to communicate
    /// request-side processing errors back to the guest (for example, a
    /// connection reset that occurs while the request body is being uploaded,
    /// before or while the response arrives). Most implementations can ignore
    /// it (`_fut`). It is provided so that custom transports with out-of-band
    /// error channels can still deliver upload errors to the component after
    /// the response has been returned.
    #[cfg(feature = "wasip3")]
```

crates/wash-runtime/src/host/http_p3.rs
```diff
@@ -3,6 +3,11 @@
 //! This module provides the HTTP request handling path for components that
 //! target WASIP3's `wasi:http/handler` interface. It uses wasmtime-wasi-http's
 //! P3 `ServicePre`/`Service` to invoke the component.
+//!
+//! It also exposes the type aliases used at the P3 outgoing-request boundary
+//! ([`P3Body`], [`P3RequestErrorFuture`], [`P3SendFuture`]). The
+//! outgoing-request egress policy itself lives on the unified
+//! [`crate::host::http::OutgoingHandler`] trait via its `send_p3` method.
```
ricochet:
```suggestion
//! [`crate::host::http::OutgoingHandler`] trait via its `send_request_p3` method.
```

crates/wash-runtime/src/host/http_p3.rs
```diff
@@ -12,7 +17,22 @@ use wasmtime::component::InstancePre;
...
+pub type P3RequestErrorFuture = Box<dyn std::future::Future<Output = Result<(), ErrorCode>> + Send>;
+
+/// Result type returned by the inner future of a P3 outgoing send: a response
+/// paired with an I/O future for response-body errors.
+pub type P3SendResult = Result<
+    (hyper::Response<P3Body>, P3RequestErrorFuture),
+    wasmtime_wasi::TrappableError<ErrorCode>,
+>;
+
+/// Future returned by [`crate::host::http::OutgoingHandler::send_p3`].
```
ricochet:
```suggestion
/// Future returned by [`crate::host::http::OutgoingHandler::send_request_p3`].
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -142,6 +142,18 @@ pub trait Router: Send + Sync + 'static {
         _allowed_hosts: &[String],
     ) -> anyhow::Result<()>;
 
+    /// Determine if a P3 outgoing request is allowed.
+    #[cfg(feature = "wasip3")]
+    fn allow_outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        request: &hyper::Request<crate::host::http_p3::P3Body>,
+        _options: &Option<wasmtime_wasi_http::p3::RequestOptions>,
```
ricochet:
```suggestion
        _options: Option<wasmtime_wasi_http::p3::RequestOptions>,
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -251,6 +263,17 @@ impl Router for DynamicRouter {
...
+        allowed_hosts: &[String],
+    ) -> anyhow::Result<()> {
+        check_allowed_hosts(request, allowed_hosts)
+    }
+
```
ricochet:
```suggestion
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -283,6 +306,60 @@ impl Router for DynamicRouter {
     }
 }
 
+/// Trait for custom outgoing HTTP egress. gRPC requests (P2 and P3) are
+/// handled by the runtime before this trait is called.
+pub trait OutgoingHandler: Send + Sync + 'static {
+    fn send_request(
```
ricochet, line 302:
```suggestion
    /// Send a P2 outgoing HTTP request for the given `workload_id`.
    fn send_request(
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -327,6 +404,17 @@ impl Router for DevRouter {
         Ok(())
     }
 
+    #[cfg(feature = "wasip3")]
+    fn allow_outgoing_request_p3(
+        &self,
+        _workload_id: &str,
+        _request: &hyper::Request<crate::host::http_p3::P3Body>,
+        _options: &Option<wasmtime_wasi_http::p3::RequestOptions>,
```
ricochet:
```suggestion
        _options: Option<wasmtime_wasi_http::p3::RequestOptions>,
```

`&Option<T>` as a parameter is a Rust anti-pattern. The outer & adds indirection without benefit: callers must write &options instead of options, and the function can't do anything with a `&Option<T>` that it couldn't do with `Option<&T>`

crates/wash-runtime/src/host/http.rs
```diff
@@ -376,6 +464,39 @@ pub trait HostHandler: Send + Sync + 'static {
...
+    /// Handle a P3 outgoing request, enforcing `allowed_hosts` policy and
+    /// delegating transport to [`wasmtime_wasi_http::p3::default_send_request`].
+    ///
+    /// Override to apply custom egress logic (e.g. alternate transports or
+    /// per-workload TLS configuration) while still honouring the allowlist via
+    /// [`check_allowed_hosts`].
+    #[cfg(feature = "wasip3")]
+    fn outgoing_request_p3(
+        &self,
+        _workload_id: &str,
```
ricochet:
```suggestion
        workload_id: &str,
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -376,6 +464,39 @@ pub trait HostHandler: Send + Sync + 'static {
...
+        request: hyper::Request<crate::host::http_p3::P3Body>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        // Response-side body-error sink: unused here because hyper's response
+        // body already reports body errors through its `Stream` impl.
+        _fut: crate::host::http_p3::P3RequestErrorFuture,
+        allowed_hosts: &[String],
+    ) -> crate::host::http_p3::P3SendFuture {
+        if let Err(e) = check_allowed_hosts(&request, allowed_hosts) {
+            use wasmtime_wasi_http::p3::bindings::http::types::ErrorCode;
+            warn!(err = %e, "outgoing request denied by allowed_hosts policy");
```
ricochet:
```suggestion
            warn!(workload_id = %workload_id, err = %e, "outgoing request denied by allowed_hosts policy");
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -638,14 +835,39 @@ impl<T: Router> HostHandler for HttpServer<T> {
...
+        &self,
+        workload_id: &str,
+        request: hyper::Request<crate::host::http_p3::P3Body>,
+        options: Option<wasmtime_wasi_http::p3::RequestOptions>,
+        fut: crate::host::http_p3::P3RequestErrorFuture,
+        allowed_hosts: &[String],
+    ) -> crate::host::http_p3::P3SendFuture {
+        if let Err(e) =
+            self.router
+                .allow_outgoing_request_p3(workload_id, &request, &options, allowed_hosts)
```
ricochet:
```suggestion
                .allow_outgoing_request_p3(workload_id, &request, options, allowed_hosts)
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -1179,6 +1401,133 @@ async fn send_grpc_request_handler(
...
+    use tokio::net::TcpStream;
+    use tokio::time::timeout;
+    use wasmtime_wasi_http::p3::bindings::http::types::ErrorCode;
+
+    let connect_timeout = options
+        .and_then(|o| o.connect_timeout)
+        .unwrap_or(Duration::from_secs(600));
+    let first_byte_timeout = options
+        .and_then(|o| o.first_byte_timeout)
+        .unwrap_or(Duration::from_secs(600));
```
ricochet, line 1440:
```suggestion
        .unwrap_or(Duration::from_secs(600));
    let between_bytes_timeout = options
        .and_then(|o| o.between_bytes_timeout)
        .unwrap_or(Duration::from_secs(600));
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -1179,6 +1401,133 @@ async fn send_grpc_request_handler(
...
+
+    let resp = timeout(first_byte_timeout, sender.send_request(request))
+        .await
+        .map_err(|_| ErrorCode::ConnectionReadTimeout)?
+        .map_err(ErrorCode::from_hyper_request_error)?
+        .map(|body| {
+            use http_body_util::BodyExt;
+            body.map_err(ErrorCode::from_hyper_request_error)
+                .boxed_unsync()
+        });
```
ricochet, line 1582:
```suggestion
    let resp = timeout(first_byte_timeout, sender.send_request(request))
        .await
        .map_err(|_| ErrorCode::ConnectionReadTimeout)?
        .map_err(ErrorCode::from_hyper_request_error)?
        .map(|body| {
            use std::pin::Pin;
            use std::task::{Context, Poll, ready};
            use http_body_util::BodyExt;

            struct TimedBody {
                inner: hyper::body::Incoming,
                interval: tokio::time::Interval,
            }
            impl hyper::body::Body for TimedBody {
                type Data = bytes::Bytes;
                type Error = ErrorCode;
                fn poll_frame(
                    mut self: Pin<&mut Self>,
                    cx: &mut Context<'_>,
                ) -> Poll<Option<Result<hyper::body::Frame<Self::Data>, Self::Error>>> {
                    match Pin::new(&mut self.as_mut().inner).poll_frame(cx) {
                        Poll::Ready(None) => Poll::Ready(None),
                        Poll::Ready(Some(Err(err))) => Poll::Ready(Some(Err(
                            ErrorCode::from_hyper_request_error(err),
                        ))),
                        Poll::Ready(Some(Ok(frame))) => {
                            self.interval.reset();
                            Poll::Ready(Some(Ok(frame)))
                        }
                        Poll::Pending => {
                            ready!(self.interval.poll_tick(cx));
                            Poll::Ready(Some(Err(ErrorCode::ConnectionReadTimeout)))
                        }
                    }
                }
                fn is_end_stream(&self) -> bool {
                    self.inner.is_end_stream()
                }
                fn size_hint(&self) -> hyper::body::SizeHint {
                    self.inner.size_hint()
                }
            }

            let mut interval = tokio::time::interval(between_bytes_timeout);
            interval.reset();
            TimedBody { inner: body, interval }.boxed_unsync()
        });

```

crates/wash-runtime/src/host/http.rs
ricochet, line 1602: * custom_outgoing_handler_is_invoked: verifies that HttpServer actually calls your custom handler instead of silently falling back to the default. Without this test, someone could accidentally short-circuit the delegation in outgoing_request (e.g. by returning early before the self.outgoing_handler.send_request(...) line) and no existing test would catch it.
* grpc_requests_bypass_outgoing_handler: verifies that gRPC requests (identified by Content-Type: application/grpc) must go to send_grpc_request directly, not through the OutgoingHandler. This is architecturally important because gRPC requires HTTP/2, which send_grpc_request enforces — a custom OutgoingHandler might use HTTP/1.1 or an alternate transport that would break gRPC. The test uses the same spy handler and asserts it is not called.
* null_server_denies_p3_outgoing_request: verifies the bug fix  from relying on default impl. NullServer is the placeholder used when there is no HTTP server. Its P2 outgoing_request returns "http client not available". The explicit P3 override added in the previous session must behave the same way. Without this test, someone could delete the override (thinking the trait default is equivalent), reverting the bug where P3 requests through NullServer would actually be sent.

```suggestion
    fn dummy_config() -> OutgoingRequestConfig {
        OutgoingRequestConfig {
            use_tls: false,
            connect_timeout: Duration::from_secs(30),
            first_byte_timeout: Duration::from_secs(30),
            between_bytes_timeout: Duration::from_secs(30),
        }
    }

    // --- OutgoingHandler delegation tests ---

    struct SpyHandler {
        called: Arc<std::sync::atomic::AtomicBool>,
    }

    impl OutgoingHandler for SpyHandler {
        fn send_request(
            &self,
            _workload_id: &str,
            _request: hyper::Request<HyperOutgoingBody>,
            _config: OutgoingRequestConfig,
        ) -> wasmtime_wasi_http::p2::HttpResult<
            wasmtime_wasi_http::p2::types::HostFutureIncomingResponse,
        > {
            self.called
                .store(true, std::sync::atomic::Ordering::SeqCst);
            Err(wasmtime_wasi_http::p2::HttpError::trap(
                wasmtime::format_err!("spy: no real request"),
            ))
        }

        #[cfg(feature = "wasip3")]
        fn send_request_p3(
            &self,
            _workload_id: &str,
            _request: hyper::Request<crate::host::http_p3::P3Body>,
            _options: Option<wasmtime_wasi_http::p3::RequestOptions>,
            _fut: crate::host::http_p3::P3RequestErrorFuture,
        ) -> crate::host::http_p3::P3SendFuture {
            unimplemented!("spy does not implement P3")
        }
    }

    #[tokio::test]
    async fn custom_outgoing_handler_is_invoked() {
        let called = Arc::new(std::sync::atomic::AtomicBool::new(false));
        let server = HttpServer::builder(DevRouter::default(), "127.0.0.1:0".parse().unwrap())
            .outgoing_handler(SpyHandler { called: called.clone() })
            .build()
            .await
            .unwrap();

        let request = build_request("http://example.com/");
        let _ = server.outgoing_request("test-workload", request, dummy_config(), &[]);

        assert!(called.load(std::sync::atomic::Ordering::SeqCst));
    }

    /// gRPC requests must bypass the OutgoingHandler and go directly to
    /// send_grpc_request, so the spy must NOT be called.
    #[tokio::test]
    async fn grpc_requests_bypass_outgoing_handler() {
        let called = Arc::new(std::sync::atomic::AtomicBool::new(false));
        let server = HttpServer::builder(DevRouter::default(), "127.0.0.1:0".parse().unwrap())
            .outgoing_handler(SpyHandler { called: called.clone() })
            .build()
            .await
            .unwrap();

        let request = hyper::Request::builder()
            .uri("http://example.com/")
            .header(hyper::header::CONTENT_TYPE, "application/grpc")
            .body(HyperOutgoingBody::default())
            .unwrap();
        let _ = server.outgoing_request("test-workload", request, dummy_config(), &[]);

        assert!(!called.load(std::sync::atomic::Ordering::SeqCst));
    }

    /// NullServer must deny P3 outgoing requests with an internal error,
    /// matching its P2 behaviour of returning "http client not available".
    #[cfg(feature = "wasip3")]
    #[tokio::test]
    async fn null_server_denies_p3_outgoing_request() {
        use crate::host::http_p3::{P3Body, P3RequestErrorFuture};
        use http_body_util::BodyExt;

        let server = NullServer::default();
        let body: P3Body = http_body_util::Empty::new()
            .map_err(|never| match never {})
            .boxed_unsync();
        let request = hyper::Request::builder()
            .uri("http://example.com/")
            .body(body)
            .unwrap();
        let fut: P3RequestErrorFuture = Box::new(async { Ok(()) });

        let result =
            Box::into_pin(server.outgoing_request_p3("test", request, None, fut, &[])).await;

        assert!(
            result.is_err(),
            "NullServer P3 outgoing request should return an error"
        );
    }
```

ricochet, line 1591:
```suggestion
    use wasmtime_wasi_http::p2::body::HyperOutgoingBody;
    use wasmtime_wasi_http::p2::types::OutgoingRequestConfig;
```

crates/wash-runtime/src/host/http.rs
ricochet: The Router trait defines allow_outgoing_request_p3 with a default body that calls check_allowed_hosts(request, allowed_hosts)

ricochet (changes requested):
This is close!

crates/wash-runtime/src/host/http.rs
```diff
@@ -1179,10 +1414,182 @@ async fn send_grpc_request_handler(
...
+        let root_cert_store = rustls::RootCertStore {
+            roots: webpki_roots::TLS_SERVER_ROOTS.into(),
+        };
+        let mut config = rustls::ClientConfig::builder()
+            .with_root_certificates(root_cert_store)
+            .with_no_client_auth();
+        config.alpn_protocols = vec![b"h2".to_vec()];
+
+        let connector = tokio_rustls::TlsConnector::from(Arc::new(config));
+        let host = authority.split(':').next().unwrap_or(&authority);
```
ricochet, line 1475:
```suggestion
        let host = authority.split(':').next().unwrap_or_default();
```

ricochet (approved):

PR #5027, opened 2026-04-03
Add endpointslice support for services

jfleitz:
## Summary

This PR replaces the runtime-gateway with native Kubernetes Service routing for wasmCloud workloads. The runtime-operator now manages EndpointSlices for user-defined Services, pointing directly to the host pod IPs running each workload. This lets workloads participate in standard Kubernetes networking — meaning any Gateway API controller (Traefik, Envoy, etc.) can provide host-based or path-based routing without any wasmCloud-specific ingress infrastructure.

### How it works

1. You create a headless Kubernetes Service and a WorkloadDeployment that references it via `spec.service.name`.
2. The operator watches for ready workloads and builds an EndpointSlice for each referenced Service, populated with the host pod IPs and HTTP ports.
3. DNS aliases (e.g. `my-svc`, `my-svc.default`, `my-svc.default.svc`) are automatically registered with the wash-runtime's HTTP router so requests arriving via cluster DNS are matched to the correct workload.
4. As workloads scale up/down or move between hosts, the EndpointSlice is updated automatically.

### Changes

- **New controller: `WorkloadRouteReconciler`** — reconciles EndpointSlices keyed by `namespace/service-name`, deduplicates endpoints by pod IP, and sets the Service as owner for garbage collection.
- **`WorkloadReconciler`** — injects `host-aliases` into the HTTP HostInterface config when `kubernetesService` is set.
- **`WorkloadDeploymentReconciler`** — propagates `spec.service` into spawned Workload specs.
- **wash-runtime `DynamicRouter`** — now tracks multiple hostnames (primary + aliases) per workload for bind/unbind.
- **CRDs** — added `kubernetesService` field to `WorkloadSpec` and `service` field to `WorkloadDeploymentSpec`.
- **RBAC** — added `get/list/watch` on Services and full CRUD on EndpointSlices.
- **Helm values** — deprecated the `gateway` section; routing is now handled via EndpointSlices.

### WorkloadDeployment with Service reference

```yaml
apiVersion: v1
kind: Service
metadata:
  name: workload-a
  namespace: default
spec:
  # Headless: the operator manages the EndpointSlice directly.
  clusterIP: None
  ports:
    - name: http
      port: 9191
      protocol: TCP
---
kind: WorkloadDeployment
metadata:
  name: workload-a
  namespace: default
spec:
  deployPolicy: RollingUpdate
  replicas: 1
  service:
    name: workload-a
  template:
    spec:
      hostSelector:
        hostgroup: default
      components:
        - name: hello-world-routing
          image: ghcr.io/cosmonic-labs/hello-world-routing:latest
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: workload-a.example.com
        - namespace: wasi
          package: config
          interfaces:
            - store
          config:
            name: workload-a
```

### Gateway and HTTPRoute (host-based routing)

Since workloads are now backed by standard Kubernetes Services, you can use the Gateway API to configure ingress routing — no wasmCloud-specific gateway required.

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: wasmcloud-gateway
  namespace: default
spec:
  gatewayClassName: traefik
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      hostname: "*.example.com"
      allowedRoutes:
        namespaces:
          from: Same
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: workload-a
  namespace: default
spec:
  parentRefs:
    - name: wasmcloud-gateway
  hostnames:
    - workload-a.example.com
  rules:
    - backendRefs:
        - name: workload-a
          port: 9191
---
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: workload-b
  namespace: default
spec:
  parentRefs:
    - name: wasmcloud-gateway
  hostnames:
    - workload-b.example.com
  rules:
    - backendRefs:
        - name: workload-b
          port: 9191
```

### HTTPRoute (path-based routing)

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: path-routing
  namespace: default
spec:
  parentRefs:
    - name: wasmcloud-gateway
  hostnames:
    - api.example.com
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /workload-a
      filters:
        - type: URLRewrite
          urlRewrite:
            hostname: workload-a.example.com
            path:
              type: ReplacePrefixMatch
              replacePrefixMatch: /
      backendRefs:
        - name: workload-a
          port: 9191
    - matches:
        - path:
            type: PathPrefix
            value: /workload-b
      filters:
        - type: URLRewrite
          urlRewrite:
            hostname: workload-b.example.com
            path:
              type: ReplacePrefixMatch
              replacePrefixMatch: /
      backendRefs:
        - name: workload-b
          port: 9191
```

## Test plan

- [x] Tested on a Kind cluster configured with Traefik as the Gateway API controller
- [x] Workloads were created, deleted, and scaled — EndpointSlices were verified to be gracefully created, updated, and deleted in response
- [x] Confirmed the runtime-gateway was not installed; all routing was handled via Kubernetes Services and EndpointSlices managed by the runtime-operator
- [x] Configured a Gateway and HTTPRoutes for both host-based and path-based routing, verified requests were correctly routed to the expected workloads

Fixes #5025 


charts/runtime-operator/crds/runtime.wasmcloud.dev_workloads.yaml
```diff
@@ -267,6 +267,22 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              kubernetesService:
```
vados-cosmonic: Nit, but how painful would it be to have a `k8s` object with `service` inside of it? `kubernetesService` is fine of course, but it's a bit long and I get the feeling this isn't the last bit of state we're going to want to stuff under to control Kubernetes functionality -- I think the top level could get cluttered pretty quickly.

charts/runtime-operator/templates/operator/clusterrole.yaml
```diff
@@ -56,4 +56,24 @@ rules:
...
+    resources:
+    - services
+    verbs:
+    - get
+    - list
+    - watch
+  - apiGroups:
+    - discovery.k8s.io
+    resources:
+    - endpointslices
```
vados-cosmonic, line 72: One thing that's interesting here is that we are setting up a cluster role, but really only need same-namespace permissions for endpoint slices in the operator. 

It feels like we could/should either: 

1. Focus on same-namespace for permissions (`RoleBinding`)
2. Add native multi-namespace management

I *might* lean towards (2) just because of the prevalence of control plane & tenant plane separation for internal teams (i.e. many setups will be same-namespace if the team that owns the namespace are also the admin, but there's the idea of a "platform team" as well that has a centralized namespace w/ more control), BUT I also like the idea of getting same-namespace working first and then expanding from there (especially if we have this configuration tidily tucked away under something like `k8s.service.name` and only have to add `k8s.service.namespace`)

charts/runtime-operator/values.yaml
```diff
@@ -60,7 +60,11 @@ operator:
   watchNamespaces: []
 
 gateway:
-  enabled: true
+  # -- DEPRECATED: runtime-gateway has been removed. HTTP routing is now handled
+  # by the runtime-operator via EndpointSlices 
+  # Set service.name in your WorkloadDeployment spec to enable automatic
+  # EndpointSlice management.
+  enabled: false
```
vados-cosmonic: Is this a breaking change as far as helm releases go?

I'm not sure exactly what the procedure is here, but just how surprised would people be by this? Should we enable but have the deprecation & "will be changed in future version" type message?

crates/wash-runtime/src/host/http.rs
```diff
@@ -115,34 +117,57 @@ impl Router for DynamicRouter {
...
             .config
             .get("host")
             .cloned()
             .context("No host header found")?;
 
+        // Collect primary hostname plus any DNS aliases injected by the operator.
+        // Aliases are a comma-separated list of Service DNS names (e.g.
+        // "my-svc,my-svc.default,my-svc.default.svc,my-svc.default.svc.cluster.local")
+        // that allow cluster-internal callers to reach this workload via Service DNS.
+        let mut all_hosts = vec![primary_host.clone()];
```
vados-cosmonic: Do we need the second `clone()` here? I might be missing somewhere else we use `primary_host`.

crates/wash-runtime/src/host/http.rs
```diff
@@ -115,34 +117,57 @@ impl Router for DynamicRouter {
...
+        if let Some(aliases) = http_iface.config.get("host-aliases") {
+            all_hosts.extend(
+                aliases
+                    .split(',')
+                    .map(|s| s.trim().to_string())
+                    .filter(|s| !s.is_empty()),
```
vados-cosmonic: Can we do some sort of check for valid hostnames here? I'm not a fan of adding more deps but

https://crates.io/crates/hostname-validator

Was hoping `url` did it, but [it looks like they don't](https://github.com/servo/rust-url/issues/913)

Really conflicted here, but it's up to you. also, I wonder if we want to do stuff like trim `http://` etc if present, it seems very likely that people are going to put in the wrong values there and be confused when things don't work. Maybe that's just over-engineering though and decent docs/examples will ensure they don't.

Then again, that makes the case for hostname validity checking -- non-encoded `:` for example shouldn't be allowed (we can emit warnings, etc)...

crates/wash-runtime/src/host/http.rs
```diff
@@ -115,34 +117,57 @@ impl Router for DynamicRouter {
...
-            let mut host_lock = self.host_to_workload.write().await;
-            if let Some(workload_set) = host_lock.get_mut(&host_header) {
-                workload_set.remove(workload_id);
-                if workload_set.is_empty() {
-                    host_lock.remove(&host_header);
+        let hostnames = {
+            let mut wth_lock = self.workload_to_host.write().await;
+            wth_lock.remove(workload_id)
+        };
+        if let Some(hostnames) = hostnames {
```
vados-cosmonic, line 183: Thanks for removing the nested lock call here -- one of the biggest problems with having two locks that need to be updated like this is ensuring lock/unlock order is consistent and it's not possible to deadlock... This should at least help us avoid this for a bit.

runtime-operator/config/crd/bases/runtime.wasmcloud.dev_workloadreplicasets.yaml
```diff
@@ -280,6 +280,22 @@ spec:
                         additionalProperties:
                           type: string
                         type: object
+                      kubernetesService:
+                        description: |-
+                          KubernetesService references an existing Kubernetes Service that the
```
vados-cosmonic:
```suggestion
                          kubernetesService references an existing Kubernetes Service that the
```

Is this supposed to match the name of the option/variable? Can we use backticks like `kubernetesService` ?

Avoiding making this comment in other places just in case this is totally normal/expected

runtime-operator/internal/controller/runtime/workload_controller.go
```diff
@@ -118,6 +118,71 @@ func (r *WorkloadReconciler) findFreeHost(ctx context.Context, workload *runtime
...
+}
+
+// injectServiceDNSAliases adds host-aliases to wasi:http/incoming-handler
+// HostInterfaces so the wash-runtime DynamicRouter accepts requests arriving
+// via Kubernetes Service DNS.
+func injectServiceDNSAliases(hostInterfaces []*runtimev2.WitInterface, svcName, namespace string) {
+	aliases := strings.Join([]string{
+		svcName,
+		fmt.Sprintf("%s.%s", svcName, namespace),
+		fmt.Sprintf("%s.%s.svc", svcName, namespace),
```
vados-cosmonic, line 160: Is this enough, or do we also need the `svc.cluster.local` ? It depends on what the client is passing through, and it *could* pass pass through w/ the `svc.cluster.local` attached.

That said, after chatting with AI a bit, we may actually get *nothing* on the host end for an internal request like this if there is no host header attached to the request. The pod will get an IP address resolved and we're going to simply receive traffic from some pod IP (assuming we're there is no proxy in between the internal pod in question and the service).

runtime-operator/internal/controller/runtime/workload_controller.go
```diff
@@ -118,6 +118,71 @@ func (r *WorkloadReconciler) findFreeHost(ctx context.Context, workload *runtime
...
+			}
+		}
+		if !hasIncomingHandler {
+			continue
+		}
+		if hi.Config == nil {
+			hi.Config = make(map[string]string)
+		}
+		hi.Config["host-aliases"] = aliases
+		hostInterfaces[i] = hi
```
vados-cosmonic: Could probably move this work up into the actual loop IMO?

runtime-operator/internal/controller/runtime/workload_route_controller.go
```diff
@@ -0,0 +1,302 @@
...
+		host, err := r.findHostByID(ctx, workload.Status.HostID)
+		if err != nil {
+			log.V(1).Info("host not found for workload, skipping", "hostID", workload.Status.HostID, "workload", workload.Name)
+			continue
+		}
+		if host.Hostname == "" || host.HTTPPort == 0 {
+			continue
+		}
+
+		podIP := host.Hostname
```
vados-cosmonic, line 90: `podIP` ends up being the hostname, is this name a little misleading?

runtime-operator/internal/controller/runtime/workload_route_controller.go
```diff
@@ -0,0 +1,302 @@
...
+	); err != nil {
+		return ctrl.Result{}, fmt.Errorf("listing workloads for service %s: %w", serviceName, err)
+	}
+
+	// Deduplicate endpoints by pod IP. Multiple workloads on the same host
+	// result in a single endpoint because the host's DynamicRouter handles
+	// internal workload selection.
+	type endpointInfo struct {
+		podIP    string
+		httpPort int32
```
vados-cosmonic, line 65:
```suggestion
		httpPort int16
```

runtime-operator/internal/controller/runtime/workload_route_controller.go
```diff
@@ -0,0 +1,302 @@
...
+		// No ready workloads for this service; delete our EndpointSlice if it exists.
+		existing := &discoveryv1.EndpointSlice{}
+		err := r.Get(ctx, types.NamespacedName{Namespace: namespace, Name: sliceName}, existing)
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.Delete(ctx, existing); err != nil && !errors.IsNotFound(err) {
```
vados-cosmonic: After asking AI, we don't need to do the additional get here to do delete, something like this:

```go
slice := &discoveryv1.EndpointSlice{
	ObjectMeta: metav1.ObjectMeta{
		Namespace: namespace,
		Name:      sliceName,
	},
}

if err := r.Delete(ctx, slice); err != nil && !apierrors.IsNotFound(err) {
	return ctrl.Result{}, fmt.Errorf("deleting empty EndpointSlice: %w", err)
}

log.Info("deleted EndpointSlice (no ready endpoints)", "endpointSlice", sliceName)
return ctrl.Result{}, nil
```

runtime-operator/internal/controller/runtime/workload_route_controller.go
```diff
@@ -0,0 +1,302 @@
...
+	// if they unexpectedly differ, use the first value and log a warning.
+	var port int32
+	for _, ep := range endpoints {
+		if port == 0 {
+			port = ep.httpPort
+		} else if port != ep.httpPort {
+			log.Info("warning: hosts have different HTTPPort values for this service, using first", "port", port)
+			break
+		}
+	}
```
vados-cosmonic: Seems like we could separate the check for non-matching http ports from setting to the first endpoint port value. 

Also, it would be nice to say *which* host(s) (do we have enough information?) had the incorrect port.

runtime-operator/internal/controller/runtime/workload_route_controller.go
```diff
@@ -0,0 +1,302 @@
...
+	if err := r.Update(ctx, existing); err != nil {
+		return ctrl.Result{}, fmt.Errorf("updating EndpointSlice: %w", err)
+	}
+	log.Info("updated EndpointSlice", "endpointSlice", sliceName, "endpoints", len(endpoints))
+
+	return ctrl.Result{}, nil
+}
+
+// findHostByID looks up a Host CRD by its HostID field using a field index.
+func (r *WorkloadRouteReconciler) findHostByID(ctx context.Context, hostID string) (*runtimev1alpha1.Host, error) {
```
vados-cosmonic: Question from earlier but worth repeating here -- is it reasonable to watch hosts actually and keep some mutex + map around that makes this available for lookup @ reconcile time? Is that overkill?

runtime-operator/internal/controller/runtime/workload_route_controller.go
```diff
@@ -0,0 +1,302 @@
...
+			{NamespacedName: types.NamespacedName{
+				Namespace: workload.Namespace,
+				Name:      workload.Spec.KubernetesService.Name,
+			}},
+		}
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Watches(&runtimev1alpha1.Workload{}, handler.EnqueueRequestsFromMapFunc(workloadToServiceRequest)).
+		Named("workload-Route").
```
vados-cosmonic: NIT: is this capitalization normal? could it be `workload-route`?

vados-cosmonic (commented):
This looks good to me mostly -- the comments I had were mostly small, and a few config organization changes that might be too painful to add anyway -- definitely happy to get something moving quickly that we can experiment with.

What do you think about gating this behavior behind a configurable in the helm chart? This feels like a new way to operate  -- essentially a sort of "direct service management" feature? "use endpoint slices"s? not sure what to call it. Could definitely understand this being the default going forward and just making it the default now though.

I ask that mostly because I'm wondering if we could add a test that ensured the new configuration worked (and check the old one didn't break). Not sure if we have in-repo tests that are extensive enough for this.

charts/runtime-operator/crds/runtime.wasmcloud.dev_workloads.yaml
jfleitz: Great point! I changed to have under a `kubernetes.service` property.

charts/runtime-operator/values.yaml
jfleitz: You are right, as this should not impact any current release. I updated the comment and have `enabled: true`

runtime-operator/internal/controller/runtime/workload_route_controller.go
jfleitz: Great catch! There are some other controller managers that had the upper-cased first name, and some that were all lower-case. I updated all of them to be all lower-case for standardization.

charts/runtime-operator/templates/operator/clusterrole.yaml
jfleitz: We have had this issue with other permissions for the Operator, and completely agree we should be as less permissive as possible. We did document for the other ClusterRoles on what is required at the ClusterRole scope vs the Role/Namespace scope. I did make a comment in the helm chart, and will get this updated in the Docs once pushed [here](https://wasmcloud.com/docs/kubernetes-operator/operator-manual/roles-and-rolebindings/#restricting-access-of-the-operator-to-namespaces)

crates/wash-runtime/src/host/http.rs
jfleitz: I added a validation to make sure each host name added is RFC 1123 compliant. This allows for some flexibility to what aliases are added, and also would make documentation easier (i.e. we just say any host must be valid for RFC 1123 and link to the spec).

runtime-operator/internal/controller/runtime/workload_route_controller.go
jfleitz: Added a comment. The Host's IP address is stored in the Hostname field. A better approach would be to just add another field to the Host CRD for HostIP, but then it would just be the same value as Hostname anyway.

runtime-operator/internal/controller/runtime/workload_route_controller.go
jfleitz: Keeping this as int32, as for some reason the Kubernetes api spec has port as int32 vs int16, to avoid casting in places it is used.

runtime-operator/internal/controller/runtime/workload_route_controller.go
jfleitz: I removed the check. This should be safe, since the HostGroup is actually a Deployment, and on the Deployment you set the HostPort (so all Hosts must have the same port). I switched to using the first port in the endpoints. This will help with performance too, when someone scales up the number of hosts in a HostGroup, and we don't traverse everyone "just because".

vados-cosmonic: Nice thanks, that seems reasonable!

runtime-operator/internal/controller/runtime/workload_route_controller.go
vados-cosmonic: Yup, weird that it is that way, but thanks for the explanation

runtime-operator/internal/controller/runtime/workload_route_controller.go
vados-cosmonic: Ah but can we literally call this something else? It would be really good if we had a name that accurately reflected what this was.

Is this some sort of thing that is outside our control or is fossilized in some way?

crates/wash-runtime/src/host/http.rs
vados-cosmonic: Yeah that makes sense!

vados-cosmonic (approved):
Hey @jfleitz so I think github did it's thing and hid a bunch of my comments -- there are a few that are still unresolved!

Don't want to keep this up for much longer though, since nothing is show-stopping. 

Also, it would be great to file a follow up issue for adding a test on the Rust side. I looked and I don't think we have a test that uses kind via `testcontainers` or anything like that to test the k8s stuff end to end. I thought we used to have a test like that, but clearly not (it might also be way harder/more annoying than we thought).

runtime-operator/config/crd/bases/runtime.wasmcloud.dev_workloadreplicasets.yaml
jfleitz: This was generated by KubeBuilder from the comment block in the type struct. This was also changed to be under a [KubernetesSpec property](https://github.com/wasmCloud/wasmCloud/pull/5027/changes#diff-29bcfacbaa1d5922cfe33c64e53fe527f26d9f890e4e392f12a14a2d066ee245R175) property as well, so this shouldn't apply any more.

runtime-operator/internal/controller/runtime/workload_controller.go
jfleitz: Yeah I intentionally made it just `service name`.`namespace` and `service name`.`namespace`.`svc`. The reason being is that the cluster name can be different (`svc.cluster.local` is the default, but can be set to something different on configuring k8s), and there is not programmatic way to get the actual cluster domain. `service name`.`namespace` will ensure uniqueness since Kubernetes enforces already that you can't have more than one service with the same name in the same namespace.

runtime-operator/internal/controller/runtime/workload_controller.go
jfleitz: Good call!

runtime-operator/internal/controller/runtime/workload_route_controller.go
jfleitz: More fossilized. [--host-name=status.podIP](https://github.com/wasmCloud/wasmCloud/blob/33fa1990f9d4fbc85cac2a2b5f7f8be7b4fab83e/charts/runtime-operator/templates/runtime/deployment.yaml#L41-L48)

I agree it is a bit misleading. I could add an optional argument to have `--pod-ip` for the wash-runtime, and then add to the heartbeat etc. an `ip_address` field. It would literally be the same value though as host-name.

ricochet (approved):

PR #5031, opened 2026-04-08
chore: examples as templates via symlinks

vados-cosmonic:
This commit adds a symlink for the `http-hello-world` example under the templates directory so it can be used as an template via `wash`.

The following command was tested on linux:

```
wash new https://github.com/vados-cosmonic/wasmCloud.git \
    --name http-hello-world \
    --subfolder templates/http-hello-world
```

Other OSes *should* also be able to work with this solution, but it's worth trying out in a few other places before we take this. That said, the approach *does* work and I think avoids the problem of copying the codebases to ensure they exist as examples and templates.

### DIFF
--- a/templates/http-hello-world
+++ b/templates/http-hello-world
@@ -1,0 +1,1 @@
+../examples/http-hello-world


ricochet:
By default, `core.symlinks` is false on Windows Git, so clones will get a plain text file containing the path. Examples should be comprehensive demonstrations and templates should be written for scaffolding. They shouldn't be intermixed. I'd argue hello-worlds are very poor examples but that's a separate issue.

If someone is intent on using an example as a starting point, they can just as easily point to it as a subfolder: 

```bash
      --name http-hello-world \
      --subfolder examples/http-hello-world
```

vados-cosmonic:
Ahhh that's a really good point. Rough that this won't work on :window: .

The context here is actually [this issue](https://github.com/wasmCloud/wasmcloud.com/pull/1116) -- basically we were using an example as a template in the guide, and it seems like this example in particular was good enough to be a template. 

I like the idea of examples-as-templates (we have certainly been here before, in wasmCloud v1, and we may have even tried symlinks before too)... but I'm happy to take some more time to think about this -- @ericgregory already wisely merged the original fix so we're at least out of the woods for now

PR #5032, opened 2026-04-08
feat(ci): ensure templates install

vados-cosmonic:
This commit adds CI checks that ensure that `wash` loads available templates properly after being built.

~~Before merging this we should of course change away from [my fork](https://github.com/vados-cosmonic/wasmcloud).~~

### DIFF
--- a/.github/workflows/wash-e2e.yml
+++ b/.github/workflows/wash-e2e.yml
@@ -1,0 +1,55 @@
+name: wash-e2e
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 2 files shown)


vados-cosmonic:
Hey @ricochet / @jfleitz thinking that we should have a simple E2E test here ensuring that all our templates load properly. thoughts? Right now I think running only on PRs and making sure we don't break the templates with new code going in is enough (rather than running on every single push/release/etc)

ricochet:
This is now covered by https://github.com/wasmCloud/wasmCloud/pull/5151

PR #5037, opened 2026-04-10
Helm updates for TLS, Annotations, and Labels

### LINKED ISSUE #5036: Add additional configuration options to the Helm chart for wasmCloud v2
For installations of wasmCloud that use Istio, or any other service mesh, it would make the installation easier to have a couple more options in the Helm chart vs using Kustomize.

1. Enable TLS (default) and create Certificates:
Add a `global.tls.enabled` flag to the runtime-operator Helm chart that controls whether TLS is configured on NATS connections and whether TLS certificates are generated. When set to `false`, all components connect to NATS over plaintext and no certificate secrets are created. The current behavior (TLS enabled with self-signed certificates) remains the default. This enables deployments where a service mesh such as Istio handles mTLS at the network layer, making application-level TLS on NATS redundant.

2. Rename the nats service port to `tcp-nats`: This will allow for a Service mesh to know that it is to handle this as TCP vs HTTP.

3. Provide the ability to specify the http port used for the Host runtime: This is in the values.yaml file, however it is hard-coded in places as `port 9191` in the Helm chart.
 
4. For each deployment, add the ability to specify pod labels and annotations: This will allow for additional flexibility to allow for external operators (such as Istio) to react to a scheduled pod, such as adding a sidecar.

jfleitz:
Fixes #5036 

### New values.yaml property

```yaml
global:
  tls:
    # -- Enable TLS for NATS connections and certificate generation.
    # Set to false when a service mesh (e.g. Istio) provides mTLS.
    enabled: true
```

This is separate from (and takes precedence over) `global.certificates.generate`. The relationship:

| `global.tls.enabled` | `global.certificates.generate` | Behavior |
|---|---|---|
| `true` (default) | `true` (default) | Self-signed certs generated, TLS args passed to all components, NATS TLS block present. **Current behavior.** |
| `true` | `false` | No certs generated (user supplies their own secrets), TLS args still passed, NATS TLS block present. **Existing BYOC flow.** |
| `false` | (ignored) | No certs generated, no TLS args passed, NATS runs plaintext. **New flow.** |

### Pod Labels and Annotations for the Nats, Operator and Runtime deployments
In `values.yaml`, you can now specify  labels and annotations that should be applied to each deployment independently. Example:
```yaml
operator:
  podLabels:
    sidecar.istio.io/inject: "true"
  podAnnotations:
    proxy.istio.io/config: '{"holdApplicationUntilProxyStarts": true}'
```
### Additional changes
- Nats service port renamed from `nats` to `tcp-nats`. This protocal prefix allows for service mesh's to know not to treat this port as HTTP.
- `runtime.hostGroups[].http.port` is used as the value for the http port in the host. Previously it was hard-coded.

## Testing
Tested locally with Kind+Istio enabled. Values.yaml file that was used:
```yaml
global:
  # disabling tls, since Istio is used
  tls:
    enabled: false
gateway:
  enabled: false

nats:
  podLabels:
    sidecar.istio.io/inject: "true"
  podAnnotations:
    proxy.istio.io/config: '{"holdApplicationUntilProxyStarts": true}'

operator:
  podLabels:
    sidecar.istio.io/inject: "true"
  podAnnotations:
    proxy.istio.io/config: '{"holdApplicationUntilProxyStarts": true}'
  image:
    registry: ""
    repository: controller
    tag: latest
    pull_policy: Never
  watchNamespaces: ["default"]

runtime:
  podLabels:
    sidecar.istio.io/inject: "true"
  podAnnotations:
    proxy.istio.io/config: '{"holdApplicationUntilProxyStarts": true}'
  image:
    registry: ""
    repository: wash
    tag: local
    # testing with local image
    pull_policy: Never
  hostGroups:
    - name: default
      replicas: 3
      service:
        type: ClusterIP
      webgpu:
        enabled: false
      http:
        enabled: true
        port: 8080
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m" # 0.25 CPU cores
        limits:
          memory: "512Mi"
          cpu: "500m" # 0.5 CPU cores
```

### DIFF
--- a/charts/runtime-operator/templates/operator/clusterrole.yaml
+++ b/charts/runtime-operator/templates/operator/clusterrole.yaml
@@ -64,8 +64,10 @@ rules:
     - get
     - list
     - watch
-  # EndpointSlice permission can be managed in a Role vs ClusterRole since WorkloadDeployments are deployed to a specific namespace.
-  # See https://wasmcloud.com/docs/kubernetes-operator/operator-manual/roles-and-rolebindings/#restricting-access-of-the-operator-to-namespaces for more details.
+  {{- if not .Values.operator.watchNamespaces }}
+  # EndpointSlice permission can be managed in per Namespace since WorkloadDeployments if desired.
+  # See https://wasmcloud.com/docs/kubernetes-operator/operator-manual/roles-and-rolebindings/#restricting-access-of-the-operator-to-namespaces
+  # for more details.
   - apiGroups:
     - discovery.k8s.io
     resources:
@@ -78,4 +80,5 @@ rules:
     - patch
     - update
     - watch
+  {{- end }}
 {{- end }}
--- a/charts/runtime-operator/templates/operator/clusterrole.yaml
+++ b/charts/runtime-operator/templates/operator/clusterrole.yaml
@@ -65,8 +65,9 @@ rules:
     - list
     - watch
   {{- if not .Values.operator.watchNamespaces }}
-  # EndpointSlice permission can be managed in per Namespace since WorkloadDeployments if desired.
-  # See https://wasmcloud.com/docs/kubernetes-operator/operator-manual/roles-and-rolebindings/#restricting-access-of-the-operator-to-namespaces
+  # If the operator is watching all namespaces, it needs permissions to manage EndpointSlices across the cluster. If it's only watching specific
+  # namespaces, a more limited Role and RoleBinding will be created in each namespace instead, and this ClusterRole will not include permissions
+  # for EndpointSlices. See https://wasmcloud.com/docs/kubernetes-operator/operator-manual/roles-and-rolebindings/#restricting-access-of-the-operator-to-namespaces
   # for more details.
   - apiGroups:
     - discovery.k8s.io
--- a/charts/runtime-operator/templates/runtime/deployment.yaml
+++ b/charts/runtime-operator/templates/runtime/deployment.yaml
@@ -38,7 +38,6 @@ spec:
           secret:
             secretName: {{ $top.Values.global.certificates.dataCertSecretName }}
         {{- end }}
-            # update this
       serviceAccountName: {{ include "runtime.serviceAccountName" $top }}
       {{- include "runtime-operator.imagePullSecrets" $top | nindent 6 }}
       containers:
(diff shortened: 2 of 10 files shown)


charts/runtime-operator/templates/operator/clusterrole.yaml
```diff
@@ -64,8 +64,10 @@ rules:
     - get
     - list
     - watch
-  # EndpointSlice permission can be managed in a Role vs ClusterRole since WorkloadDeployments are deployed to a specific namespace.
-  # See https://wasmcloud.com/docs/kubernetes-operator/operator-manual/roles-and-rolebindings/#restricting-access-of-the-operator-to-namespaces for more details.
+  {{- if not .Values.operator.watchNamespaces }}
+  # EndpointSlice permission can be managed in per Namespace since WorkloadDeployments if desired.
```
ricochet: This comment should be fixed

charts/runtime-operator/templates/runtime/deployment.yaml
```diff
@@ -21,17 +21,23 @@ spec:
...
           emptyDir: {}
+        {{- if $top.Values.global.tls.enabled }}
         - name: runtime-cert
           secret:
             secretName: {{ $top.Values.global.certificates.runtimeCertSecretName }}
         - name: data-cert
           secret:
             secretName: {{ $top.Values.global.certificates.dataCertSecretName }}
+        {{- end }}
             # update this
```
ricochet: should this comment be removed?

charts/runtime-operator/values.yaml
```diff
@@ -6,6 +6,10 @@ global:
     pullSecrets: []
   nameOverride: ""
   fullnameOverride: ""
+  tls:
+    # -- Enable TLS for NATS connections and certificate generation.
+    # Set to false when a service mesh (e.g. Istio) provides mTLS.
+    enabled: true
   certificates:
     # -- Enable automatic generation of self-signed TLS certificates for the operator and runtime hosts
```
ricochet:
```suggestion
    # -- Enable automatic generation of self-signed TLS certificates for the operator and runtime hosts.
    # Ignored when global.tls.enabled is false.
```

charts/runtime-operator/values.yaml
ricochet: This is technically a breaking change for the default where it now actually uses this value. 

I wonder if we should set this to 9191 to avoid any unexpected breakage.

jfleitz: Great catch! I had it as 80 for testing, and forgot to change it back.

ricochet (approved):

PR #5048, opened 2026-04-14
Update TLS support for both wash dev and wash host

### LINKED ISSUE #5047: Unify TLS support for `wash dev` and `wash host`
`wash dev` and `wash host` handle TLS inconsistently:

1. `wash dev` supports TLS configuration via `.wash/config.yaml` but panicked at runtime due to a missing rustls `CryptoProvider`.
2. `wash host` had no TLS support for its HTTP server at all, and always used `wash_runtime::host::http::HttpServer::new()`. Additionally, the `rcgen` dependency used for certificate generation relied on the `ring` crypto backend, while the rest of the TLS stack (`rustls`, `async-nats`, `tonic`) used `aws-lc-rs`, creating a split crypto dependency that blocks future FIPS compliance.
3. There should be a test added for testing TLS with `wash dev` and `wash host` as well.

jfleitz:
## Unify TLS support across `wash dev` and `wash host`, standardize on `aws-lc-rs`

Fixes #5047

Fixed bug in [Helm chart](https://github.com/wasmCloud/wasmCloud/pull/5048/changes#diff-bd198a38cdc987e309dfbd6a84035cd06b169714155104173ffb5ac040bc84ccR1) where if you provide a list of namespaces to create a RoleBinding for EndpointSlices, a yaml formatting error will occur

Adds `hostGroups.wasip3.enabled` property in the helm chart to allow for enabling wasip3 support for the hosts in the group. 

### Summary

- Fixed `wash dev` panicking when TLS was configured due to missing rustls `CryptoProvider` initialization
- Added `--tls-cert-path`, `--tls-key-path`, and `--tls-ca-path` CLI flags to `wash host` so both commands now support TLS for their HTTP servers
- Removed `ring` crypto backend dependency and standardized on `aws-lc-rs` across the entire TLS stack (`rustls`, `rcgen`, `async-nats`, `tonic`), enabling a future path to FIPS compliance via `aws-lc-rs/fips`
- Upgraded `rcgen` from 0.13 to 0.14.7 to pick up bug fixes in the previous version
- Added TLS integration tests that generate certificates at test time and validate HTTPS requests end-to-end

### Changes

**`crates/wash/src/cli/dev.rs`**
- Added `rustls::crypto::aws_lc_rs::default_provider().install_default()` to fix the runtime panic when TLS config was present

**`crates/wash/src/cli/host.rs`**
- Added `--tls-cert-path`, `--tls-key-path`, `--tls-ca-path` CLI flags with mutual requirement validation via clap
- Updated HTTP server setup to use `HttpServer::new_with_tls()` when TLS flags are provided

**`crates/wash/Cargo.toml`**
- Upgraded `rcgen` from 0.13 to 0.14.7
- Switched `rcgen` crypto backend from `ring` to `aws_lc_rs`

**`crates/wash-runtime/Cargo.toml`**
- Added `rcgen` 0.14.7 as a dev-dependency with `aws_lc_rs` backend

**`crates/wash-runtime/tests/integration_http_tls.rs`**
- New integration test file with two tests:
  - `test_https_request_succeeds` — generates self-signed certs, starts a TLS host, deploys `http-counter`, verifies HTTPS requests return correct responses
  - `test_https_rejects_plain_http` — confirms plain HTTP connections to the TLS server are rejected

### Test plan

- [x] `cargo test -p wash-runtime --test integration_http_tls` — both new TLS tests pass
- [x] `cargo test -p wash-runtime --test integration_http_counter` — existing HTTP tests unaffected
- [x] `cargo check -p wash` — compiles cleanly
- [x] Manual: `wash dev` with TLS config in `.wash/config.yaml` serves HTTPS and responds to `curl -k https://localhost:8443/`
- [x] Manual: `wash host --http-addr 0.0.0.0:8443 --tls-cert-path ... --tls-key-path ...` serves HTTPS

### DIFF
--- a/crates/wash/Cargo.toml
+++ b/crates/wash/Cargo.toml
@@ -75,7 +75,7 @@ pbjson-build = { workspace = true, default-features = true }
 bytes = { version = "1", default-features = false }
 http = { version = "1.3.1", default-features = false }
 http-body-util = { version = "0.1.3", default-features = false }
-rcgen = { version = "0.13", default-features = false, features = ["crypto", "ring", "pem"] }
+rcgen = { version = "0.14.7", default-features = false, features = ["crypto", "aws_lc_rs", "pem"] }
 scopeguard = { version = "1.2", default-features = false }
 tempfile = { version = "3.0", default-features = false }
 tokio = { version = "1.45.1", default-features = false, features = ["full"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,6 +178,45 @@ dependencies = [
 ]
 
 [[package]]
+name = "asn1-rs"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56624a96882bb8c26d61312ae18cb45868e5a9992ea73c58e45c3101e56a1e60"
+dependencies = [
+ "asn1-rs-derive",
+ "asn1-rs-impl",
+ "displaydoc",
+ "nom",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 2.0.18",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs-derive"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+ "synstructure",
+]
+
+[[package]]
+name = "asn1-rs-impl"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "astral-tokio-tar"
 version = "0.5.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1555,6 +1594,20 @@ dependencies = [
  "flagset",
  "pem-rfc7468",
  "zeroize",
+]
+
+[[package]]
+name = "der-parser"
+version = "10.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07da5016415d5a3c4dd39b11ed26f915f52fc4e0dc197d87908bc916e51bc1a6"
+dependencies = [
+ "asn1-rs",
+ "displaydoc",
+ "nom",
+ "num-bigint",
+ "num-traits",
+ "rusticata-macros",
 ]
 
 [[package]]
@@ -3819,6 +3872,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "oid-registry"
+version = "0.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "12f40cff3dde1b6087cc5d5f5d4d65712f34016a03ed60e9c08dcc392736b5b7"
+dependencies = [
+ "asn1-rs",
+]
+
+[[package]]
 name = "olpc-cjson"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4938,14 +5000,15 @@ dependencies = [
 
 [[package]]
 name = "rcgen"
-version = "0.13.2"
+version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "75e669e5202259b5314d1ea5397316ad400819437857b90861765f24c4cf80a2"
+checksum = "10b99e0098aa4082912d4c649628623db6aba77335e4f4569ff5083a6448b32e"
 dependencies = [
+ "aws-lc-rs",
  "pem",
- "ring",
  "rustls-pki-types",
  "time",
+ "x509-parser",
  "yasna",
 ]
 
@@ -5179,6 +5242,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cfcb3a22ef46e85b45de6ee7e79d063319ebb6594faafcf1c225ea92ab6e9b92"
 dependencies = [
  "semver",
+]
+
+[[package]]
+name = "rusticata-macros"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "faf0c4a6ece9950b9abdb62b1cfcf2a68b3b67a10ba445b3bb85be2a293d0632"
+dependencies = [
+ "nom",
 ]
 
 [[package]]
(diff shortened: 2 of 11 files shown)


jfleitz:
For testing within a Kind cluster, the following `values.yaml` was used:
```yaml
global:
  # disabling tls, since Istio is used
  tls:
    enabled: true
gateway:
  enabled: false


operator:
  image:
    pull_policy: Always

runtime:
  image:
    registry: ""
    repository: "wash"
    tag: "local"
    pull_policy: Never
  hostGroups:
    - name: default
      replicas: 3
      service:
        type: ClusterIP
      webgpu:
        enabled: false
      http:
        enabled: true
        port: 8443
        tls:
          enabled: true
          certificate:
            secretName: "byoc-cert"
            generate:
              enabled: false
              validity: 365
              commonName: "example.com"              
              domains:
                - "localhost"
                - "*.example.com"
                - "hostgroup-default.default.svc.cluster.local"
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m" # 0.25 CPU cores
        limits:
          memory: "512Mi"
          cpu: "500m" # 0.5 CPU cores
```
The byoc.cert:
```
kind: Secret
apiVersion: v1
type: kubernetes.io/tls
metadata:
  name: byoc-cert
  namespace: default
data:
  ca.crt: ...
  tls.crt: ...
  tls.key: ...
```
Deployed Workload:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: workload-a
  namespace: default
spec:
  type: NodePort
  ports:
    - name: http
      port: 8443
      protocol: TCP
      nodePort: 30951
---
kind: WorkloadDeployment
metadata:
  name: workload-a
  namespace: default
spec:
  deployPolicy: RollingUpdate
  replicas: 1

  # Reference an existing Kubernetes Service. The runtime-operator will
  # maintain an EndpointSlice for this Service pointing to the host pod IPs
  # that are running this workload. Apply service.yaml before this manifest.
  kubernetes:
    service:
      name: workload-a

  template:
    spec:
      hostSelector:
        hostgroup: default

      components:
        - name: hello-world-routing
          image: ghcr.io/cosmonic-labs/hello-world-routing:latest

      hostInterfaces:
        # HTTP interface: specifies the hostname that routes to this workload.
        # The operator also registers Service DNS aliases (hello-world-routing,
        # hello-world-routing.default.svc.cluster.local, etc.) automatically.
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: example.com

        # Config interface: key-value pairs exposed to the component at runtime
        # via wasi:config/store. The "name" key is read by the component and
        # included in the greeting response.
        - namespace: wasi
          package: config
          interfaces:
            - store
          config:
            name: workload-a
```

Once deployed, a curl was performed and confirmed TLS worked:
```bash
curl -k -v  -H "Host: example.com" https://localhost:8443

* Host localhost:8443 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8443...
* connect to ::1 port 8443 from ::1 port 51715 failed: Connection refused
*   Trying 127.0.0.1:8443...
* Connected to localhost (127.0.0.1) port 8443
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
* (304) (IN), TLS handshake, Server hello (2):
* (304) (IN), TLS handshake, Unknown (8):
* (304) (IN), TLS handshake, Certificate (11):
* (304) (IN), TLS handshake, CERT verify (15):
* (304) (IN), TLS handshake, Finished (20):
* (304) (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / AEAD-CHACHA20-POLY1305-SHA256 / [blank] / UNDEF
* ALPN: server accepted h2
* Server certificate:
*  subject: CN=example.com
*  start date: Apr 15 19:07:12 2026 GMT
*  expire date: Apr 15 19:07:12 2027 GMT
*  issuer: CN=wasmcloud CA
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://localhost:8443/
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: example.com]
* [HTTP/2] [1] [:path: /]
* [HTTP/2] [1] [user-agent: curl/8.7.1]
* [HTTP/2] [1] [accept: */*]
> GET / HTTP/2
> Host: example.com
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/2 200 
< date: Wed, 15 Apr 2026 20:00:25 GMT
< 
Hello, workload-a!
* Connection #0 to host localhost left intact
```

Once tested, `tls.certificate.secretName` was set to "", and `tls.certificate.generate.enable` was set to `true`. Test was redone, and had the same response.

crates/wash/Cargo.toml
```diff
@@ -75,7 +75,7 @@ pbjson-build = { workspace = true, default-features = true }
 bytes = { version = "1", default-features = false }
 http = { version = "1.3.1", default-features = false }
 http-body-util = { version = "0.1.3", default-features = false }
-rcgen = { version = "0.13", default-features = false, features = ["crypto", "ring", "pem"] }
+rcgen = { version = "0.14.7", default-features = false, features = ["crypto", "aws_lc_rs", "pem"] }
```
vados-cosmonic, line 78: Did we have to get off `ring`? IIRC `ring` is a bit more portable and can be statically built more easily.

Is it possible to support both backends via a feature flag or is that more trouble than it's worth? 

@ricochet may have some opinions here

vados-cosmonic (approved):
This looks fine to me -- had one note about `ring` vs `aws_lc_rs`, it feels like we should allow access to either via a feature flag, but maybe that's much complexity here. 

This [Reddit comment](https://www.reddit.com/r/rust/comments/1de13y6/ring_vs_awslcrs/l89pbmc/) makes a *really* good case for `aws-lc-rs`:

> From rustls point of view, I would add:
>
>    - aws-lc has had significant formal verification (see https://github.com/aws/aws-lc#formal-verification for a start); I think it is one of the leading cryptography libraries in this measure.
>    - earlier this year we shipped https://crates.io/crates/rustls-post-quantum which requires aws-lc-rs.
>    - shortly we will ship Encrypted ClientHello support, again only usable out of the box with aws-lc-rs.


Also a [good (but short) thread on Rust-lang](https://users.rust-lang.org/t/why-did-rustls-choose-aws-lc-rs-to-replace-ring-as-its-default-cryptography-library/134559/3) forums about this

Another good comment [here on Reddit](https://www.reddit.com/r/rust/comments/1od195h/the_crate_ring_we_dont_recommend_that_third/nkszoin/)

Either way it'd be *nice* if we could maintain optionality here (happy to default to `aws-lc-rs` of course!) but maybe it's just not something we need until someone in particular notes that they need `ring` support due to other deps, etc.

crates/wash/Cargo.toml
vados-cosmonic: Ah so I just realized that we're [using `rustls` with `aws_lc_rs` ](https://github.com/wasmCloud/wasmCloud/blob/main/Cargo.toml#L84) -- going to resolve this, and let's get this in, thanks for adding a test!

PR #5059, opened 2026-04-17
fix(test): add shared p2 WIT deps infrastructure

ricochet:
Extracts duplicated wasi package.wit files from individual fixture wit/deps/ directories into a shared p2-wit-deps/ directory (mirroring the existing p3-wit-deps/ pattern). build.rs resolves shared deps into each fixture's wit/deps/ at build time so the source tree carries one copy per package.

- Unify p2/p3 fixture build paths into a single `build_fixtures` function driven by `FixtureKind::{P2, P3}`
- Add `http-handler-p2` fixture: minimal wasi:http/incoming-handler that returns a static body, used by upcoming benchmarks
- Migrate http-counter to shared deps (drop checked-in wit/deps, bump blobstore's wasi:io refs from 0.2.1 to 0.2.2)
- Fix clippy::redundant_locals in integration_p2_p3_matrix test


vados-cosmonic (approved):
LGTM 🚀

PR #5064, opened 2026-04-20
feat: Expand DevRouter and DynamicRouter Tests

### LINKED ISSUE #4945: [IMPROVEMENT] expand router tests
Expand routing test coverage:
  - DynamicRouter with multiple host headers
  - DevRouter with rapid workload changes
  - Concurrent routing scenarios

Aditya1404Sal:
Resolves #4945 

## Summary
Adds integration tests for DynamicRouter (multi-host + aliases, concurrent load, race-with-unbind) and DevRouter (last-wins, selective unbind, concurrent reads during swap).

## Test plan
- `cargo test -p wash-runtime --test integration_router_dynamic`
-  `cargo test -p wash-runtime --test integration_router_dev`

### DIFF
--- a/crates/wash-runtime/tests/integration_router_dev.rs
+++ b/crates/wash-runtime/tests/integration_router_dev.rs
@@ -1,0 +1,379 @@
+//! Integration tests for `DevRouter` — the development-mode router that
+//! dispatches every request to the last resolved workload.
+//!
+//! Covers:
+//! - `last_workload_id` is updated on each `on_workload_resolved`, proved by
+//!   distinguishable response bodies from two simultaneously-registered
+//!   components (http-counter vs. http-handler-p2).
+//! - `on_workload_unbind` clears `last_workload_id` only when the stopped id
+//!   matches the currently-bound workload; stopping any other workload is a
+//!   no-op for routing.
+//! - Concurrent reads racing a workload swap all resolve to a response (no
+//!   hangs, no panics from `try_lock` contention).
+
+#![allow(clippy::unwrap_used, clippy::expect_used)]
+
+use anyhow::{Context, Result};
+use futures::future::join_all;
+use std::{collections::HashMap, sync::Arc, time::Duration};
+use tokio::time::timeout;
+
+use wash_runtime::{
+    engine::Engine,
+    host::{
+        HostApi, HostBuilder,
+        http::{DevRouter, HttpServer},
+    },
+    plugin::{
+        wasi_blobstore::InMemoryBlobstore, wasi_config::DynamicConfig,
+        wasi_keyvalue::InMemoryKeyValue, wasi_logging::TracingLogger,
+    },
+    types::{Component, LocalResources, Workload, WorkloadStartRequest, WorkloadStopRequest},
+    wit::WitInterface,
+};
+
+const HTTP_COUNTER_WASM: &[u8] = include_bytes!("wasm/http_counter.wasm");
+const HTTP_HANDLER_P2_WASM: &[u8] = include_bytes!("wasm/http_handler_p2.wasm");
+
+fn http_counter_host_interfaces(http_host_config: &str) -> Vec<WitInterface> {
+    vec![
+        WitInterface {
+            namespace: "wasi".to_string(),
+            package: "http".to_string(),
+            interfaces: ["incoming-handler".to_string()].into_iter().collect(),
+            version: Some(semver::Version::parse("0.2.2").unwrap()),
+            config: {
+                let mut c = HashMap::new();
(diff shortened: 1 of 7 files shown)


crates/wash-runtime/tests/integration_router_dev.rs
ricochet, line 1: `http_counter_host_interfaces` / `http_counter_workload_request` / `start_host_with_*` are now duplicated across four integration files (`integration_http_counter.rs`, `integration_http_tls.rs`, and the two new ones). Not blocking, but a tests/common/ module would be nice to have.

Aditya1404Sal: I'll assimilate them

crates/wash-runtime/tests/integration_router_dev.rs
```diff
@@ -0,0 +1,379 @@
...
+    .await
+    .context("request timed out")?
+    .context("request failed")?;
+    let status = response.status();
+    let body = response.text().await.unwrap_or_default();
+    Ok((status, body))
+}
+
+fn is_unrouted(status: reqwest::StatusCode) -> bool {
+    !status.is_success()
```
ricochet: We should assert exactly the correct status code in case we run into regressions. 

```suggestion
    status == reqwest::StatusCode::BAD_REQUEST
```

While looking at this, I actually think the 400 here is debatable. I don't expect this PR to fix this but does highlight why it's handy to surface the error code.  IMO 404 (no handler for this host) or 503 (service unavailable) would be more conventional.

```bash
  ┌────────┬───────────────────────────────┬─────────────────────────────────────────────┐
  │ Status │            Meaning            │           Should this test pass?            │
  ├────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ 2xx    │ Component ran and responded   │ No — this is the opposite of "unrouted"     │
  ├────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ 3xx    │ Redirect                      │ Shouldn't happen, but would pass spuriously │
  ├────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ 400    │ route_incoming_request failed │ Yes — the intended signal                   │
  ├────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ 404    │ Routed, but handle missing    │ Maybe — different bug                       │
  ├────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ 500    │ Component panicked/errored    │ No — unrelated failure                      │
  └────────┴───────────────────────────────┴─────────────────────────────────────────────┘
```

ricochet (approved):

Aditya1404Sal:
getting the helpers sorted right now, sorry for the delay!!

crates/wash-runtime/tests/integration_router_dynamic.rs
```diff
@@ -0,0 +1,218 @@
...
+            resp.map(|r| r.status().is_success()).unwrap_or(false)
+        }));
+    }
+
+    let successful = join_all(handles)
+        .await
+        .into_iter()
+        .filter(|r| r.as_ref().copied().unwrap_or(false))
+        .count();
+    // 18/20 requests accounting 2 reqs for flakes
```
ricochet: either (a) treat 503 as a retry-able transient and retry once, then assert ==N/N succeed, or (b) keep the threshold but assert that every failure is specifically a 503 (not 4xx, not connection-refused). As written, a routing regression that returns 200 for the wrong workload could hide here

crates/wash-runtime/tests/integration_router_dynamic.rs
```diff
@@ -0,0 +1,218 @@
...
+    assert!(
+        successful >= 18,
+        "expected >=18 concurrent multi-host requests to succeed, got {successful}"
+    );
+
+    Ok(())
+}
+
+/// 50 concurrent requests against a single fixed host must mostly succeed
+/// and complete quickly, the try_read must not deadlock or starve under stres.
```
ricochet:
```suggestion
/// and complete quickly, the try_read must not deadlock or starve under stress.
```

ricochet:
We could add these as a separate issue but noticed a couple more cases worth covering: 
  - test for RouteError::MissingHost (HTTP/1.0 request without Host header which should return 400).
  - test asserting DynamicRouter rejects invalid hostnames in host-aliases (the silent filter at crates/wash-runtime/src/host/http.rs:204). A test that registers "valid.local,not a hostname" and confirms only valid.local routes.
  - test for multiple workloads bound to the same host (the HashSet<String> value in host_to_workload exists for this.

ricochet (approved):
I think there's probably more we can iterate on here, but let's land this for now as it's a great basis for router tests.

PR #5067, opened 2026-04-20
fix(chart): disable gateway in values.local.yaml and route hello-world via Service

ericgregory:
The Runtime Gateway is deprecated in 2.0.3, and HTTP traffic is now routed to workloads by the operator via EndpointSlices tied to standard Kubernetes Services. This PR brings the local-dev overlay and the `http-hello-world` example in line with that pattern, so users can follow the quickstart without needing `--set gateway.enabled=false` on every install.

- `charts/runtime-operator/values.local.yaml`: set `gateway.enabled: false` and drop the gateway Service block. The host HTTP listener on port 80 is still preserved.
- `examples/http-hello-world/manifests/workloaddeployment.yaml`: add a `NodePort` Service named `hello-world` (nodePort 30950) and reference it from the WorkloadDeployment via `spec.template.spec.kubernetes.service.name`. Traffic on host port 80 (mapped by `deploy/kind/kind-config.yaml` to container 30950) now reaches the component via the operator-managed EndpointSlice.

### DIFF
--- a/charts/runtime-operator/values.local.yaml
+++ b/charts/runtime-operator/values.local.yaml
@@ -1,8 +1,24 @@
+# The Runtime Gateway is deprecated as of wasmCloud 2.0.3. HTTP traffic to
+# workloads is now routed by the operator via EndpointSlices tied to standard
+# Kubernetes Services (see a WorkloadDeployment's
+# spec.template.spec.kubernetes.service.name field).
+#
+# The gateway is disabled here so its NodePort Service does not collide with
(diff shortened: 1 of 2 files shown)


vados-cosmonic (approved):
LGTM 🚀 

I think we might want a note around the gateway and why it's disabled by default/unnecessary going forward. 

Often it's also nice to include an example of what it looks like when configured so people can just uncomment if they need it it (i.e. an example of the alternate configuration). That said, if we have docs on how to configure the gateway we can just link to them instead!

ricochet (approved):

PR #5075, opened 2026-04-23
fix(runtime-gateway): strip port from Host header during workload lookup

RoopeKuisma:
## Problem

Per RFC 7230 §5.4, HTTP/1.1 clients must include the port in the `Host` header
when it is non-standard (e.g. `Host: localhost:8080`). Workload hostnames are
registered as bare RFC 1123 names (no port allowed by the CRD validator), so
the exact map lookup always missed, returning **503 Service Unavailable** for
every browser or standard `curl` request.

The bug exists in **two places**:
1. The `runtime-gateway` (Go) — the user-facing symptom most operators hit first.
2. The `wash-runtime` host (Rust) — the same logic, hidden behind the gateway
   until users migrate to direct `EndpointSlice`-based routing.

**Workaround (before this fix):**
```
curl -H 'Host: localhost' http://localhost:8080/   # 200 OK
curl              http://localhost:8080/            # 503
```

### `runtime-gateway/tracker.go` (Go)

After an exact-match miss, retry the lookup with the port stripped using
`net.SplitHostPort`. Exact-match registrations still take precedence.
`net.SplitHostPort` handles bracketed IPv6 (`[::1]:8080`) correctly.

### `crates/wash-runtime/src/host/http.rs` (Rust)

Same logic in `DynamicRouter::route_incoming_request`. Extracted into a
testable `lookup_workload_set` helper with a small `strip_port` utility that:
- handles bracketed IPv6 (`[::1]:8080` → `[::1]`)
- refuses to truncate unbracketed IPv6 (e.g. `fe80::1`) where the split
  would be ambiguous

The `Host` header is **not** rewritten before forwarding to the backend —
components see exactly what the client sent, matching nginx/traefik/envoy
default behaviour. (An earlier draft also rewrote `Host` in `proxy.go`; that
was reverted as redundant once the host-side lookup became port-tolerant.)

## Tests

**Go (`runtime-gateway/tracker_test.go`)** — 3 cases:
- Port-stripped lookup resolves correctly
- Exact `host:port` registration takes precedence over stripped fallback
- Unknown host still hits `InvalidHostname` fallback with the original header

**Rust (`crates/wash-runtime/src/host/http.rs`)** — 6 cases against the
extracted helpers:
- Port-stripped lookup
- Exact-match precedence
- Unknown host returns `None`
- Bracketed IPv6 lookup
- Unbracketed IPv6 is not mangled
- `strip_port` table-driven cases (IPv4/IPv6/no-port)

### DIFF
--- a/crates/wash-runtime/src/host/http.rs
+++ b/crates/wash-runtime/src/host/http.rs
@@ -217,8 +217,8 @@ impl Router for DynamicRouter {
                 .and_then(|h| h.to_str().ok())
                 .or_else(|| req.uri().authority().map(|a| a.as_str()))
                 .context("no Host header or :authority in request")?;
-            let Some(workload_set) = lock.get(workload_host) else {
-                anyhow::bail!("no workload bound to host header: {}", workload_host);
+            let Some(workload_set) = lookup_workload_set(&lock, workload_host) else {
+                anyhow::bail!("no workload bound to host header: {workload_host}");
             };
 
             let workload_id = workload_set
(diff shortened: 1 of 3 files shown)


RoopeKuisma:
I migrated to EndpointSlices and found that http.rs needs port stripping logic too

vados-cosmonic:
Hey @RoopeKuisma thanks for the hard work here! It seems like we actually had this bug on the gateway *and* the new `EndpointSlice` based approach. At least it was consistent :sweat_smile: I guess. 

I started off the tests, but it looks like the [DCO check](https://github.com/wasmCloud/wasmCloud/pull/5075/checks?check_run_id=72685558466) is failing -- would you mind taking a look?

RoopeKuisma:
Fixed DCO thanks for notifying me @vados-cosmonic 
I also removed the repro files

vados-cosmonic:
Hey @jfleitz @ricochet what do you think? I'm leaning towards us trying to prevent ports in hostnames as early as possible rather than trying to strip them out but not sure

RoopeKuisma:
closing this as I've understood now that ports are usually stripped by the LB/Proxy. apologies for the noise

PR #5079, opened 2026-04-24
chore: upgrade to wasmtime 44

ricochet:
Introduces the map Val type and one issue we ran into while upgrading:

wasmtime 44 upgraded its internal rustls dependency from 0.22 to 0.23. Under 0.22 with only the ring feature enabled, rustls auto-selected a provider; under 0.23 with both ring and aws-lc-rs features enabled in the unified tree, the caller must install one explicitly.

Wasmtime PR that changed this: [bytecodealliance/wasmtime#12837 — Update ](https://github.com/bytecodealliance/wasmtime/pull/12837) security update (RUSTSEC-2026-0049 in rustls-webpki) and it claimed to "use ring as the crypto backend to preserve existing dependencies". In wasmCloud's Cargo feature-unified tree, aws-lc-rs also gets pulled in (e.g. via async-nats), producing the ambiguity that triggers a runtime panic (caught in our integration tests).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -338,7 +338,7 @@ dependencies = [
  "thiserror 1.0.69",
  "time",
  "tokio",
- "tokio-rustls 0.26.4",
+ "tokio-rustls",
  "tokio-stream",
  "tokio-util",
  "tokio-websockets",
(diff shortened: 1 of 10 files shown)


vados-cosmonic (approved):
LGTM 🚀

jfleitz (approved):

PR #5080, opened 2026-04-24
fix: preserve wash dev postgres config

### LINKED ISSUE #4935: wasmcloud postgres configuration failure
Disclaimer: this bug is found and patched by claude opus 4.6.

## Issue Summary

The `wash dev` command in wasmCloud v2 (main branch as of 2026-03-01) fails to load `postgres_url` from the project's `.wash/config.yaml` file due to a bug in how `figment` configuration merging is used.

## Impact

- The `wasmcloud:postgres` plugin is not registered when running `wash dev`
- Components that import `wasmcloud:postgres/query` fail to start with error: "component imports instance wasmcloud:postgres/query@0.1.1-draft, but a matching implementation was not found in the linker"
- Users cannot use PostgreSQL database capabilities in local development

## Root Cause

The bug is in `crates/wash/src/cli/dev.rs` and `crates/wash/src/config.rs`:

### Problem 1: Config::default() serialization

In `crates/wash/src/config.rs`:

```rust
impl Default for Config {
    fn default() -> Self {
        Config {
            version: Some(env!("CARGO_PKG_VERSION").to_string()),  // ← This causes the bug
            build: None,
            new: None,
            dev: None,
            wit: None,
        }
    }
}
```

When `figment::providers::Serialized::defaults(Config::default())` is called, figment serializes **all fields** of the struct, including `dev: None`, because `version` is `Some(...)`. This serialized `dev: null` then **overrides** the `dev` section from the project's `.wash/config.yaml` file during the merge.

### Problem 2: CLI args override file config

In `crates/wash/src/cli/dev.rs`:

```rust
let config = load_config(
    &ctx.user_config_path(),
    Some(project_dir),
    Some(Config {
        dev: None,  // ← This gets serialized and overrides file config
        wit: Some(WitConfig { skip_fetch: true, ..Default::default() }),
        ..Default::default()  // ← Includes version: Some(...), triggering serialization
    }),
)
```

The `Config { dev: None, ..Default::default() }` passed as CLI args gets serialized by figment and merged **after** the file config, causing the file's `dev.postgres_url` to be clobbered.

### Why address and host_interfaces work but postgres_url doesn't

This is inconsistent behavior in figment's merge logic. The `address` and `host_interfaces` fields from the file config survive the merge, but `postgres_url` specifically gets clobbered. This may be related to how figment handles nested optional fields during serialization.

## Reproduction

1. Create a project with `.wash/config.yaml`:
```yaml
dev:
  address: 127.0.0.1:8090
  postgres_url: postgres://user:pass@localhost:5432
  host_interfaces:
    - namespace: wasmcloud
      package: postgres
      interfaces: [query, types]
      version: 0.1.1-draft
      config:
        database: mydb
```

2. Run `wash dev` in the project directory

3. Observe that:
   - HTTP server listens on the correct `address` (127.0.0.1:8090) ✅
   - `host_interfaces` are present in the workload ✅
   - But `postgres_url` is **not** picked up ❌
   - The postgres plugin is **not** registered ❌

### Option 1: Fix Config::default() (minimal change)

Change `Config::default()` to not set `version`:

```rust
impl Default for Config {
    fn default() -> Self {
        Config {
            version: None,  // ← Changed from Some(...)
            build: None,
            new: None,
            dev: None,
            wit: None,
        }
    }
}
```

This prevents figment from serializing all fields, allowing the file config to be preserved.

### Option 2: Add environment variable override (better UX)

Add support for a `WASH_POSTGRES_URL` environment variable that always takes priority:

```rust
let mut dev_config = config.dev();

// Environment variable always overrides config file
if let Ok(url) = std::env::var("WASH_POSTGRES_URL") {
    dev_config.postgres_url = Some(url);
}
```

This provides a clean way to pass credentials without modifying tracked config files.

### Option 3: Don't pass dev: None in CLI args

Instead of passing `Config { dev: None, ... }`, only pass the fields that actually need to be set:

```rust
let config = load_config(
    &ctx.user_config_path(),
    Some(project_dir),
    Some(Config {
        wit: Some(WitConfig { skip_fetch: true, ..Default::default() }),
        ..Config {
            version: None,
            build: None,
            new: None,
            dev: None,  // Don't override this
            wit: None,
        }
    }),
)
```

## Workaround (Temporary)

Until this is fixed upstream, users can patch the wash source code:

1. **Patch `config.rs`**: Change `version: Some(...)` to `version: None` in `Config::default()`

2. **Patch `dev.rs`**: Add env var override after `let dev_config = config.dev();`:

```rust
let dev_config = config.dev();
let dev_config = {
    let mut dc = dev_config;
    // WASH_POSTGRES_URL env var always takes priority
    if let Ok(url) = std::env::var("WASH_POSTGRES_URL") {
        dc.postgres_url = Some(url);
    } else if dc.postgres_url.is_none() {
        // Fallback: read from project .wash/config.yaml directly
        let cfg_path = project_dir.join(".wash").join("config.yaml");
        if let Ok(raw) = std::fs::read_to_string(&cfg_path) {
            if let Ok(val) = serde_yaml_ng::from_str::<serde_yaml_ng::Value>(&raw) {
                if let Some(url) = val.get("dev").and_then(|d| d.get("postgres_url")).and_then(|v| v.as_str()) {
                    dc.postgres_url = Some(url.to_string());
                }
            }
        }
    }
    dc
};
```

## Related

- PR wasmCloud/wash#285: Added postgres plugin support to wash v2 (merged Feb 12, 2026)
- The postgres plugin code itself works correctly; the issue is purely in config loading

## Environment

- wash version: 2.0.0-rc.7 (main branch as of 2026-03-01)
- OS: Linux
- Rust: 1.93.1

## Testing

To verify the fix works:

1. Apply the patch
2. Rebuild wash: `cargo build --release -p wash`
3. Run `wash dev` in a project with `postgres_url` in `.wash/config.yaml`
4. Verify the postgres plugin is registered (check debug logs or component startup)
5. Verify components using `wasmcloud:postgres/query` start successfully

officialasishkumar:
Fixes #4935

## Summary
- keep the `wash dev` WIT skip-fetch overlay from carrying unrelated config defaults
- preserve project `dev.postgres_url` and related dev config while applying that overlay
- allow `WASH_POSTGRES_URL` to override the dev postgres URL, matching the host command environment variable

## Testing
- `CARGO_NET_GIT_FETCH_WITH_CLI=true cargo test -p wash load_config_preserves_local_dev_config_when_cli_override_disables_wit_fetch`
- `CARGO_NET_GIT_FETCH_WITH_CLI=true cargo clippy -p wash --features wasip3`

### DIFF
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -18,7 +18,7 @@ use wash_runtime::{
 
 use crate::{
     cli::{CliCommand, CliContext, CommandOutput, component_build::build_dev_component},
-    config::{Config, load_config},
+    config::{Config, DevConfig, load_config},
     wit::WitConfig,
 };
 
@@ -40,6 +40,7 @@ impl CliCommand for DevCommand {
             &ctx.user_config_path(),
             Some(project_dir),
             Some(Config {
+                version: None,
                 dev: None,
                 wit: Some(WitConfig {
                     skip_fetch: true,
@@ -50,7 +51,8 @@ impl CliCommand for DevCommand {
         )
         .context("failed to load config for development")?;
 
-        let dev_config = config.dev();
+        let mut dev_config = config.dev();
+        apply_dev_env_overrides(&mut dev_config);
         let http_addr = dev_config
             .address
             .clone()
@@ -404,6 +406,14 @@ async fn create_workload(host: &Host, config: &Config, bytes: Bytes) -> anyhow::
         service,
         volumes,
     })
+}
+
+fn apply_dev_env_overrides(dev_config: &mut DevConfig) {
+    if let Ok(postgres_url) = std::env::var("WASH_POSTGRES_URL")
+        && !postgres_url.is_empty()
+    {
+        dev_config.postgres_url = Some(postgres_url);
+    }
 }
 
 /// Reload the component in the host, stopping the previous workload if needed
--- a/crates/wash/src/config.rs
+++ b/crates/wash/src/config.rs
@@ -349,3 +349,62 @@ pub async fn generate_default_config(path: &Path, force: bool) -> Result<()> {
     info!(config_path = %path.display(), "Generated default configuration");
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use std::fs;
+
+    use super::*;
+
+    #[test]
+    fn load_config_preserves_local_dev_config_when_cli_override_disables_wit_fetch() -> Result<()> {
+        let project_dir = tempfile::tempdir()?;
+        let wash_dir = project_dir.path().join(CONFIG_DIR_NAME);
+        fs::create_dir_all(&wash_dir)?;
+        fs::write(
+            wash_dir.join(CONFIG_FILE_NAME),
+            r#"
+dev:
+  address: 127.0.0.1:8090
+  postgres_url: postgres://user:pass@localhost:5432
+  host_interfaces:
+    - namespace: wasmcloud
+      package: postgres
+      interfaces: [query, types]
+      version: 0.1.1-draft
+      config:
+        database: mydb
+"#,
+        )?;
+
+        let config = load_config(
+            &project_dir.path().join("missing-global-config.yaml"),
+            Some(project_dir.path()),
+            Some(Config {
+                version: None,
+                dev: None,
+                wit: Some(WitConfig {
+                    skip_fetch: true,
+                    ..Default::default()
+                }),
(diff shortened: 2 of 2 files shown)


ricochet:
I think this will also fix the same issue generically: https://github.com/wasmCloud/wasmCloud/pull/5117

vados-cosmonic:
Given that https://github.com/wasmCloud/wasmCloud/pull/5117 has merged, I'm closing this for now, but please let us know if it didn't solve your issue!

PR #5084, opened 2026-04-28
docs(templates): reorganize Rust templates and add HTTP handler set

ericgregory:
- Move `examples/http-hello-world` to `templates/http-hello-world`
- Move and rename `examples/key-value` to `templates/http-kv-handler` (Cargo package and WIT world renamed for parity with TypeScript naming; default `BACKEND` corrected to `in_memory` to match the README)
- Add `templates/http-client` for outgoing HTTP via `wstd::http::Client`
- Add `templates/http-handler` for HTTP routing with `axum` via `wstd-axum`
- Update `templates/README.md` and root `README.md` to reference the new paths

Brings Rust template coverage closer to parity with the TypeScript template set. Each template was verified end-to-end with `wash dev` against `wash 2.0.4`.


templates/http-kv-handler/.wash/config.yaml
ricochet, line 1: Chatted with Eric on this one offline. We're thinking we should rename some of these with the `-service` postfix to `-handler` to disambiguate between `wasi:http/service` worlds and our own `Service` type in `WorkloadDeployment`.

ericgregory: Naming is now updated to disambiguate!

ricochet (approved):

PR #5085, opened 2026-04-28
fix: explicit fixture building

dman-os:
A variation on #4998. 

Using build.rs for building fixtures can be a bit tricky as there's no easy heuristic for knowing when not to build. We want to avoid building the fixtures unless someone is developing wasmCloud directly. The most reliable way to indicate that like feature flags or env vars require extra flags on the invocation. Might as well be a separate command.

- Adds a new xtask crate to house the fixture building script which now need to explicitly be built

Note that this looses the change detection recompile features that a build.rs script offers but an incremental build-test-fixtures takes like 3 seconds for all fixtures when there's no work to be done.


ricochet:
This makes sense to me. I'd like to hear from other maintainers if xtask is the preferred tool for the job. I've heard others liking Justfiles

Aditya1404Sal:
> This makes sense to me. I'd like to hear from other maintainers if xtask is the preferred tool for the job. I've heard others liking Justfiles

I like justfiles but for this particular usecase xtask seems better 👍🏼

dman-os:
I'm a just fan of just and it'd def improve the latency and snappiness of this tool. This comes at the cost of having to wrangle sh script and also requiring wasm-tools in the env. Will re-write the fixture building if that's acceptable to ya'll.

Aditya1404Sal:
closing this over preference for #5089

PR #5086, opened 2026-04-28
Fix workload readiness for NATS subs

### LINKED ISSUE #5074: [BUG] `wasmcloud:messaging/handler` subscriptions silently fail to land on NATS when deployed via `runtime-operator` chart
**Version:** `runtime-operator` chart `2.0.4`, host image `ghcr.io/wasmcloud/wash:2.0.4`, operator `ghcr.io/wasmcloud/runtime-operator:2.0.4`, gateway `ghcr.io/wasmcloud/runtime-gateway:2.0.4`. Also reproduces with `wash-runtime` pinned to tag `v2.0.3` and `v2.0.4`.

**Summary:** A component that exports `wasmcloud:messaging/handler@0.2.0` and declares `config.subscriptions` in a `WorkloadDeployment` reaches `Ready=True`, the engine logs `bind_plugins` with the expected subscription config, but no NATS subscription is ever observable on the shared bus. Messages published by a sibling component reach NATS correctly, yet the handler never receives them. The same components work end to end under `wash dev`, which wires `InMemoryMessaging` instead of `NatsMessaging`.

### Repro

Build the `templates/http-api-with-distributed-workloads` template (either the Rust variant at [`wasmCloud/wasmCloud/templates/http-api-with-distributed-workloads`](https://github.com/wasmCloud/wasmCloud/tree/main/templates/http-api-with-distributed-workloads) or the TypeScript variant at [`wasmCloud/typescript/templates/http-api-with-distributed-workloads`](https://github.com/wasmCloud/typescript/tree/main/templates/http-api-with-distributed-workloads)), then push the two resulting components to a registry the cluster can pull from:

```bash
# Rust variant
  --name http-api-with-distributed-workloads \
  --subfolder templates/http-api-with-distributed-workloads
cd http-api-with-distributed-workloads
wash build

wash oci push <REGISTRY>/http-api-with-distributed-workloads/http-api:0.1.0 \
  target/wasm32-wasip2/release/http_api.wasm
wash oci push <REGISTRY>/http-api-with-distributed-workloads/task-worker:0.1.0 \
  target/wasm32-wasip2/release/task_worker.wasm
```

Kind cluster from [`deploy/kind/kind-config.yaml`](https://github.com/wasmCloud/wasmCloud/blob/main/deploy/kind/kind-config.yaml):

```bash
kind create cluster --config deploy/kind/kind-config.yaml

helm install wasmcloud oci://ghcr.io/wasmcloud/charts/runtime-operator \
  --version 2.0.4 \
  --namespace wasmcloud --create-namespace \
  --set global.tls.enabled=false \
  --set gateway.service.type=NodePort --set gateway.service.nodePort=30950 \
  --set runtime.image.tag=2.0.4 \
  --set operator.image.tag=2.0.4 \
  --set gateway.image.tag=2.0.4
```

Apply two `WorkloadDeployment`s, one for the HTTP component and one for the task-worker component. Full manifest:

```yaml
---
kind: WorkloadDeployment
metadata:
  name: ref-http
spec:
  replicas: 1
  template:
    spec:
      hostInterfaces:
        - namespace: wasi
          package: http
          config:
            host: ref.localhost.direct
      components:
        - name: http-api
          image: <REGISTRY>/http-api-with-distributed-workloads/http-api:0.1.0
---
kind: WorkloadDeployment
metadata:
  name: ref-task
spec:
  replicas: 1
  template:
    spec:
      hostInterfaces:
        - namespace: wasmcloud
          package: messaging
          version: "0.2.0"
          interfaces: [handler]
          config:
            subscriptions: "tasks.task-worker"
      components:
        - name: task-worker
          image: <REGISTRY>/http-api-with-distributed-workloads/task-worker:0.1.0
```

Both `WorkloadDeployment`s and both child `Workload` CRs reach `Ready=True`.

```bash
curl -sS -m 15 -H 'Host: ref.localhost.direct' \
  -X POST -H 'Content-Type: application/json' \
  -d '{"worker":"task-worker","payload":"Hello World"}' \
  http://localhost/task
# => request timed out (or HTTP 504 from the runtime-gateway)
```

Under `wash dev` with the same component sources, this returns `H3110 W0r1d`.

### Isolation

Three probes with `natsio/nats-box` on the same NATS service the host connects to (`nats://nats.wasmcloud.svc:4222`).

Probe 1, confirm publish reaches NATS:

```bash
kubectl run nats-sub -n wasmcloud --image=natsio/nats-box:latest --restart=Never \
  --command -- sh -c 'nats --server=nats://nats:4222 sub "tasks.*" --raw'
# Trigger the curl above. Subscriber log:
# {"worker":"task-worker","payload":"Hello World"}
```

The HTTP component's outbound publish reaches the shared bus.

Probe 2, confirm the task-worker component is not subscribed:

The external `nats-box` subscriber sees one delivery per request (itself, on the wildcard). If the task-worker were subscribed, the host would log an `incoming_wasmcloud_message` span. None ever fires.

Probe 3, confirm the HTTP component's reply path works when any subscriber replies:

```bash
kubectl run nats-reply -n wasmcloud --image=natsio/nats-box:latest --restart=Never \
  --command -- sh -c 'nats --server=nats://nats:4222 reply "tasks.task-worker" "MOCK_REPLY"'
# Trigger the curl above. HTTP response body: MOCK_REPLY
```

The round trip completes end to end the moment any external subscriber replies on `tasks.task-worker`. This rules out the HTTP component, the reply correlation, NATS, and networking. The remaining unaccounted step is the task-worker component's inbound NATS subscription.

### Host debug logs (`RUST_LOG=info,wash_runtime=debug`)

```
INFO  Host started host_id=... version="0.2.0"
INFO  Host provides interfaces count=10 interfaces=[..., "wasmcloud:messaging/handler@0.2.0"]
INFO  Starting workload workload.name="ref-task-..."
DEBUG successfully initialized workload component
DEBUG bind_plugins: binding plugin to workload plugin_id="wasmcloud-messaging"
      interfaces={WitInterface { namespace: "wasmcloud", package: "messaging",
        interfaces: {"handler"}, version: Some(0.2.0),
        config: {"subscriptions": "tasks.task-worker"}, name: None }}
DEBUG bind_plugins: binding plugin to workload item plugin_id="wasmcloud-messaging"
      component_id="..." interfaces={..., config: {"subscriptions": "tasks.task-worker"}, ...}
DEBUG link_components: import not found in component exports, assuming host-provided
      name="wasmcloud:messaging/consumer@0.2.0"
... (silence for task-worker for the remainder of its lifetime) ...
WARN  request{subject=tasks.task-worker}: failed to send request: request timed out
```

`on_workload_item_bind` fires and stores `subscriptions=["tasks.task-worker"]` in the tracker at `crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs:167`. No log appears for `on_workload_resolved` because that function has no observability on the success path. Either it is never called, or the `self.client.subscribe(...)` call at `nats.rs:206` completes without registering a subscription visible on the bus, or the spawned `tokio::spawn` handler loop at `nats.rs:225` exits before the first `messages.next()` poll.

The `InMemoryMessaging` plugin used by `wash dev` works for the same components with the same manifest shape. Confirmed against the Rust template HEAD: `POST /task` returns `H3110 W0r1d`. Confirmed against the TypeScript variant: same payload, same response. Messaging only fails on the k8s path, where `NatsMessaging` is the wired plugin.

### Suggested next step

The success path of `on_workload_resolved` at `crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs:178-230` is completely silent. Seven targeted `debug!` calls (entry, after tracker read, after `instantiate_pre`, before/after each `subscribe`, inside the spawned loop on enter and exit) would localize the break to one of four candidate causes: (a) `on_workload_resolved` never called, (b) `self.client.subscribe` fails or hangs, (c) the spawned handler loop exits immediately, (d) the tracker's `ComponentData` is overwritten between bind and resolve. Happy to open a PR adding the logs if that helps.

### Environment

Darwin 25.4.0 arm64, Docker 28.3.3, kind 0.27.0, kubectl 1.33.0, helm 3.17.3, kind node `kindest/node:v1.32.2`. Wash 2.0.4 for local `wash dev` comparisons.

ricochet:
Fixes #5074. A `WorkloadDeployment` exporting `wasmcloud:messaging/handler` could reach `Ready=True` with no NATS subscription on the bus, so requests timed out with "no responders".

The plugin called `client.flush()` after `subscribe()`, but `flush()` only drains the local TCP buffer — NATS hadn't yet processed the SUB. Under TLS or any data-path latency the race opened wide.

The fix subscribes to a fresh inbox, publishes a sentinel byte, and awaits the round-trip. NATS processes per-connection commands in order, so the sentinel's arrival proves every prior SUB is active.

I broke this down into commits that you might want to review one at a time:
- **fix(messaging): force NATS server sync before resolve returns**
- **test(messaging): add integration and fault-injection coverage**
- **test(runtime-operator): add kind e2e for NATS messaging round-trip**
- **ci(wash): gate canary-v2 on runtime-operator e2e**

## Test plan

- [x] `cargo test -p wash-runtime` — unit tests pass without Docker
- [x] `NATS_INTEGRATION_TESTS=1 cargo test -p wash-runtime` — integration + fault-injection suites pass
- [x] `make test-e2e MESSAGING_E2E_IMAGE=… BUILD_RUNTIME_IMAGE=true` — kind suite passes
- [x] CI's `runtime-operator-e2e` goes green on this PR

### DIFF
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -12,6 +12,9 @@ on:
       - "Cargo.*"
       - "rust-toolchain.toml"
       - "rustfmt.toml"
+      - "runtime-operator/test/e2e/**"
+      - "charts/runtime-operator/**"
+      - "deploy/kind/kind-config.yaml"
   push:
     branches:
       - main
@@ -126,7 +129,66 @@ jobs:
         run: |
           cargo machete
 
-  canary:
+  # Integration gate: build the wash image from this commit, install it
+  # alongside the runtime-operator chart in a kind cluster, and run the
+  # operator's e2e suite (including the messaging round-trip from #5074)
+  # against it. The canary tag only moves if this passes. Also runs on
+  # PRs that touch wash, the chart, or the e2e suite, so regressions are
+  # caught before merge rather than after.
+  runtime-operator-e2e:
+    if: github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/main')
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: ./.github/actions/setup-go
+      - name: Build wash runtime image
+        run: docker build -t localhost/wasmcloud-wash:e2e .
+      - name: Build runtime-operator image
+        working-directory: runtime-operator
+        run: make docker-build IMG=localhost/runtime-operator:e2e
+      - uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+        with:
+          cluster_name: wasmcloud-e2e
+          config: deploy/kind/kind-config.yaml
+      - name: Run runtime-operator e2e against locally-built wash
+        working-directory: runtime-operator
+        run: make test-e2e
+        env:
+          # Operator and wash images are already built above; the suite
+          # just needs to load them into kind.
+          SKIP_IMAGE_BUILD: "true"
+          BUILD_RUNTIME_IMAGE: "true"
+          SKIP_RUNTIME_BUILD: "true"
+          # Exercise the messaging spec — this is the regression test for
+          # #5074 and is the whole reason the gate exists.
+          MESSAGING_E2E_IMAGE: ghcr.io/wasmcloud/components/messaging-echo-rust:0.1.0
+          PROMETHEUS_INSTALL_SKIP: "true"
+          CERT_MANAGER_INSTALL_SKIP: "true"
+          KIND_CLUSTER_NAME: wasmcloud-e2e
+      - name: Dump diagnostics
+        if: failure()
+        run: |
+          echo "=== Pod Status ==="
+          kubectl get pods -n wasmcloud-system -o wide || true
+          echo "=== Events ==="
+          kubectl get events -n wasmcloud-system --sort-by=.lastTimestamp || true
+          echo "=== Operator Logs ==="
+          kubectl logs -n wasmcloud-system -l wasmcloud.com/name=runtime-operator --tail=200 || true
+          echo "=== Host Logs ==="
+          kubectl logs -n wasmcloud-system -l wasmcloud.com/name=hostgroup --tail=300 || true
+          echo "=== CRD Resources ==="
+          kubectl get hosts.runtime.wasmcloud.dev -A -o yaml || true
+          kubectl get workloaddeployments.runtime.wasmcloud.dev -A -o yaml || true
+          kubectl get workloads.runtime.wasmcloud.dev -A -o yaml || true
+
+  # The canary docker image is built in parallel with check/lint/e2e to
+  # avoid serializing the slow cargo build behind the slow tests. With
+  # `push: false`, the multi-arch digests are still pushed to ghcr (the
+  # workflow always does that for caching), but the human-visible
+  # `canary-v2` tag is *not* created here — that happens in
+  # `canary-publish` below, gated on every test job passing.
+  canary-build:
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     permissions:
       contents: write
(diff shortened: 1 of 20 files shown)


jfleitz (approved):
LGTM. I also downloaded and tested locally with a example WorkloadDeployment without an issue.

ericgregory:
I tested with several components and everything worked perfectly 🚀

ericgregory (approved):

PR #5089, opened 2026-04-29
fix: sidecar crate to build fixtures

dman-os:
Another variation on #4998 and #5085 implementing @Aditya1404Sal [workaround](https://github.com/wasmCloud/wasmCloud/pull/4998#issuecomment-4338814001).

- Moves fixture building into a sidecar crate's build.rs that's only added as a dev/test dependency to `wash-runtime`.


crates/wash-runtime/tests/integration_cron_service.rs
```diff
@@ -1,113 +0,0 @@
-//! Integration test for cron-service component
```
Aditya1404Sal, line 1: was this removed by mistake?

dman-os: Oof, idk how that happend. Let me amend.

crates/wash-runtime/tests/common/build.rs
```diff
@@ -0,0 +1,260 @@
+// Build scripts commonly use expect() since panics produce clear compile-time errors
```
Aditya1404Sal, line 1: I don't think this sidecar crate should exist in /tests/fixtures/* 
I think you could probably modify the /tests/common crate and make it a workspacce package and add the build.rs there while keeping the common helpers for the integration tests intact

Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "3"
 default-members = ["crates/wash"]
 members = [
     "crates/wash-runtime",
+    "crates/wash-runtime/tests/fixtures/build-fixtures/",
```
Aditya1404Sal: This is not necessary

```suggestion
```

Cargo.toml
```diff
@@ -118,6 +119,7 @@ wasmtime-wasi-io = { version = "44", default-features = false }
 wasmtime-wasi-http = { version = "44", default-features = false }
 wit-component = { version = "0.247.0", default-features = false }
 wash-runtime = { path = "crates/wash-runtime", default-features = false }
+build-fixtures = { path = "crates/wash-runtime/tests/fixtures/build-fixtures/" }
```
Aditya1404Sal:
```suggestion
```
Instead of the root Cargo.toml, please make such change to the `wash-runtime/Cargo.toml`
```rust
[dev-dependencies]
common = { package = "wash-runtime-test-helper", path = "tests/common" }
```
This is with reference to the above review comment, ie: to add the fixture building to the `tests/common` crate.

Aditya1404Sal (changes requested):
Nice job :)

Aditya1404Sal:
Also @dman-os could you resolve the merge conflicts?

crates/wash/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![doc = include_str!("../../../README.md")]
+#![recursion_limit = "256"]
```
Aditya1404Sal, line 2: Did you get an error from the compiler to bump this up ?

dman-os: Yes, on a nightly version though.

Aditya1404Sal: If the error doesn't surface on the pinned toolchain rustc, then maybe it's better if we don't pre-emptively set this now. on subsequent version bump we can change this.

ricochet: We run nightly with our fmt so likely many run with nightly by default. 256 is still a reasonable limit so personally OK with setting this noww.

crates/wash-runtime/build.rs
```diff
@@ -287,7 +67,6 @@ fn main() {
     // can locate fixture artifacts regardless of how it was invoked.
     println!("cargo:rustc-env=WORKSPACE_ROOT={}", workspace_dir.display());
```
Aditya1404Sal: The tests themselves don't use the `WORKSPACE_ROOT` for locating the fixtures since the path is hardcoded to be `wasm/*.wasm` so this should be safe to remove in both this build file and the one in `tests/common`

crates/wash-runtime/tests/common/build.rs
```diff
@@ -0,0 +1,257 @@
...
+    build_fixtures(workspace_dir, P3_FIXTURES, FixtureKind::P3, &[])
+        .expect("failed to build P3 fixtures");
+}
+
+fn main() {
+    let workspace_dir = workspace_dir().expect("failed to get workspace dir");
+
+    // Export WORKSPACE_ROOT so runtime code (`env!("WORKSPACE_ROOT")`)
+    // can locate fixture artifacts regardless of how it was invoked.
+    println!("cargo:rustc-env=WORKSPACE_ROOT={}", workspace_dir.display());
```
Aditya1404Sal: Not required since WORKSPACE_ROOT is unused and IMO arguably the fixture inference is better with the current way of path inference (`wasm/*.wasm`) 
```suggestion
```

Aditya1404Sal (changes requested):
This is really close!! Sorry for the back and forth

ricochet:
I went on a little adventure trying to pre-warm macros for CodeQL in https://github.com/wasmCloud/wasmCloud/pull/5168 

This change will run into the same class of challenges I tried working around there and we'll either have to keep expanding out the hacks, or I think we should reconsider the approach in #5085. Basically the whole class of env-inheritance bugs (CARGO_TARGET_DIR, RUSTFLAGS, CARGO_BUILD_*) goes away when you're not nesting cargo in build.rs. It's a bummer to have to do more than `cargo test` but I think doing anything else is fighting the system.

crates/wash-runtime/tests/common/build.rs
dman-os: Uhh, I'm not sure what to replace `workspace_dir` with. You'll need to navigate using CARGO_MANIFEST_DIR one way or another. There's no analogue to `include_bytes!` where it's like `comptime_full_path!` that I know of.

dman-os: Oh, do you mean to only remove the unused rustc-env?

dman-os:
@ricochet yeah, cargo's philosophy is to do *the thing* and nothing more. It's designed to be extended with `cargo-subcommand` programs though and I usually can't go without things like https://nexte.st/. Also, almost always, a cross platform scripting layer on top like xtask, just, or my latest preferrence, a folder of `chmod +x` ts/deno files [[^](https://github.com/dman-os/townframe/blob/main/x/build-daybook-http.ts)] that I put in the `$PATH`.

dman-os:
Superseded by #5220.

PR #5097, opened 2026-04-30
ci: disable 1.9 dependabot

ricochet:
Since we haven't re-enabled/rebuilt the old CI for 1.9, I think it's better to leave the branch as-is rather than roll the deps forward for now.


jfleitz (approved):
LGTM

PR #5098, opened 2026-05-01
ci: automated release train

### LINKED ISSUE #5071: Proposal: Adopt a train release model for wasmCloud
Today wasmCloud releases are ad hoc. I'd like to propose moving to a train release model with a regular, predictable cadence.

## Motivation

- Predictability for users planning upgrades and for maintainers planning work into a release
- Reduced coordination cost, no more one-off "are we cutting a release?" threads
- Automation-first, if releases are scheduled, they must be automated.

## Proposal

- Cadence: every two weeks, on Tuesday
  - Why Tuesday: avoids Monday catch-up (people returning from the weekend, CI/infra hiccups surfacing), avoids Friday/pre-weekend cuts (no one wants
to debug a bad release on Saturday), and leaves Wed–Thu as a buffer for any follow-up patch release within the same working week.
- No skip conditions (like if release lands on a US holiday). The train leaves on schedule. If something is broken, we revert. Predictability is the whole point.
- Push-button & fully automated release process
- Patch releases still allowed at any time for critical fixes; the cadence is a floor, not a ceiling.
- Single release version across the unit. For simplicity, we keep one version (vX.Y.Z) across wash, runtime-gateway, and runtime-operator. This is the same as today. We haven't started publishing the wash-runtime crate but when we do, it will be part of this version.

## Alternative cadence considered

Monthly, aligned to the day after Wasmtime's release. Benefits: easier to pick up Wasmtime updates on a fresh cycle. Tradeoff: slower feedback loop for users.

## Implementation

Docs to update if accepted:
- [RELEASE_RUNBOOK.md](https://github.com/wasmCloud/wasmCloud/blob/main/RELEASE_RUNBOOK.md): add a "Release cadence" section; note the no-skip policy; clarify that patch releases remain on-demand
- [GOVERNANCE.md](https://github.com/wasmCloud/wasmCloud/blob/main/GOVERNANCE.md): the "different release cycles" line (≈L21) and the "release timing" reference (≈L142) should reflect the scheduled cadence for this repo
- [CONTRIBUTING.md](https://github.com/wasmCloud/wasmCloud/blob/main/CONTRIBUTING.md): brief pointer to the cadence so contributors know when their merged change will ship
- [ROADMAP.md](https://github.com/wasmCloud/wasmCloud/blob/main/ROADMAP.md) — link to the release cadence so roadmap items can be planned against train dates

## CI approach

The existing release workflows (wash.yml, runtime-gateway.yml, runtime-operator.yml, charts.yml) are already tag-triggered on v*. They stay as-is. The train model only automates the pre-tag steps.

New `.github/workflows/release-train.yml`:

1. Trigger: weekly schedule cron at Tuesday 16:00 UTC, with a biweekly gate (ISO week parity). workflow_dispatch bypasses the gate for manual runs and patch releases.
2. Version bump: always patch by default. Minor/major bumps are done manually by a maintainer via workflow_dispatch input (or done ahead of time and committed at the time of feature addition). We can iterate toward conventional-commit-driven auto-detection later if we want nicer changelogs.
3. PR route (not direct push): the job opens a release/vX.Y.Z branch, edits the root Cargo.toml, runs cargo check to update Cargo.lock, updates chart, and opens a PR labeled release.
4. CI must pass. If the version-bump PR's CI fails, the train fails loudly (Slack notification, job marked failed). We fix forward the same
day and re-run; we do not skip the train.
5. Auto-merge on green. Once required checks pass, the PR auto-merges.
6. Tag on merge. A second workflow listens for merged PRs labeled release-train and pushes vX.Y.Z from main. Existing tag-triggered workflows take over from there.
7. Slack notifications to #release-bot:
  - Train started (PR opened)
  - Train failed (CI red, human intervention needed)
  - Release published

ricochet:
Closes #5071. Biweekly Tuesday cron opens a version-bump PR; release-tag.yml
promotes the canary OCI digest as vX.Y.Z (build-once, no rebuild), attests
artifacts with SLSA provenance, and pushes the immutable tag. wash-runtime
co-versioned and published via crates.io OIDC trusted publishing.

RC support via bump=rc.

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -1,0 +1,372 @@
+name: release-tag
+
+# Once the release-train PR has merged AND the canary builds on the merge
+# commit have succeeded, this workflow:
+#   1. Promotes (build-once) each canary container image: retags
+#      `ghcr.io/wasmcloud/<image>:sha-<merge-sha>` as `vX.Y.Z`. Same
+#      manifest digest, no rebuild — the bytes that passed canary CI are
+#      the bytes users pull at the release tag.
+#   2. Re-attests the promoted manifest digest so SLSA provenance covers
+#      the release tag (and not just the sha-prefixed canary).
+#   3. Downloads the canary-binaries artifacts produced by wash.yml on the
+#      merge commit, attests them, and creates the GitHub Release directly
+#      with those exact bytes attached (no rebuild on tag).
+#   4. Pushes the immutable annotated `vX.Y.Z` git tag at the merge commit.
+#      That tag triggers the remaining tag-listening workflows (charts
+#      release, runtime-operator go-module-tag, wash-runtime crates.io
+#      publish, wash homebrew + winget updates).
+#
+# Why workflow_run on release-train:
+#   - Avoids running on every closed PR repo-wide.
+#   - release-train.yml waits for the PR to actually merge before exiting,
+#     so by the time this workflow fires the merge commit is on main.
+#
+# Idempotency:
+#   - If a vX.Y.Z OCI tag already exists at GHCR, the retag step skips it
+#     (immutability — never overwrite a published release).
+#   - If the git tag already exists on origin, the tag step is a no-op.
+#   - Re-running this workflow is safe: each step is idempotent.
+
+on:
+  workflow_run:
+    workflows: [release-train]
+    types: [completed]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-tag-${{ github.event.workflow_run.id }}
+  cancel-in-progress: false
+
+jobs:
+  promote:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    permissions:
+      contents: write
+      packages: write
+      # For re-attesting the promoted release manifest digest.
+      id-token: write
+      attestations: write
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: main
+          # PAT so the tag push triggers the downstream tag-listening workflows.
+          token: ${{ secrets.WASMCLOUD_BOT_PAT }}
+          persist-credentials: true
+          fetch-depth: 0
+
+      - name: Configure git
+        run: |
+          git config user.name "wasmcloud-bot"
+          git config user.email "wasmcloud-bot@users.noreply.github.com"
+
+      - name: Verify HEAD is a release commit and read version
+        id: ver
+        run: |
+          set -euo pipefail
+          echo "STAGE=verify-head" >> "${GITHUB_ENV}"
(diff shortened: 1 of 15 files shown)


crates/wash-runtime/Cargo.toml
Aditya1404Sal, line 25:
```suggestion
exclude = ["tests/fixtures/**/*"]
```
Does it make sense publishing them (or even /tests in general) ?

ricochet: This is a good callout. We should use an include list here like https://github.com/hyperium/hyper/blob/master/Cargo.toml#L16

.github/workflows/wash.yml
```diff
@@ -226,15 +229,16 @@ jobs:
...
-        uses: docker/setup-buildx-action@v4
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
       - name: Docker meta
         id: meta
-        uses: docker/metadata-action@v6
+        uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
         with:
           images: ghcr.io/wasmcloud/wash
           tags: |
             type=raw,value=canary-v2
```
jfleitz: This isn't really related to this PR change, but maybe we can now get rid of the `canary-v2` tag and just use `canary` now?

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,372 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.WASMCLOUD_BOT_PAT }}
+          MERGE_SHA: ${{ steps.ver.outputs.sha }}
+        run: |
+          set -euo pipefail
+          echo "STAGE=wait-for-canary" >> "${GITHUB_ENV}"
+          # Each of these workflows publishes a sha-prefixed canary image
+          # (wash, runtime-gateway, runtime-operator) or a chart canary that
+          # we'll repackage on tag (charts).
+          REQUIRED=("wash" "runtime-operator" "runtime-gateway" "charts")
```
jfleitz, line 111: Do we need to add "wash-runtime" as well? Just in case the build fails for wash-runtime itself, we wouldn't want to ship everything else?

jfleitz (commented):
Question left on release-tag.yml to see if we need to add wash-runtime on the required list, otherwise everything else looks good

.github/workflows/wash.yml
ricochet: Agreed it's worth doing. It touches a handful of refs but while we're making this larger cut over, now might be the best time.

.github/workflows/release-tag.yml
ricochet: wash-runtime.yml triggers on `push: tags`, so adding it to REQUIRED would deadlock.

The underlying concern is real though. If cargo publish fails post-tag, we've already shipped containers + the GH Release + the Go module tag, and
crates.io is left behind. I'm going to opt to have a dry-run mode in `release-train.yml`.

jfleitz (dismissed):

jfleitz (approved):

PR #5113, opened 2026-05-04
feat(wash-runtime): add wasi:tls support for p3

### LINKED ISSUE #5076: Add wasi:tls (0.3.0-draft) support to wash-runtime behind a flag
## Background

wasmtime 44 shipped an initial host implementation of the [wasi:tls](https://github.com/WebAssembly/wasi-tls) proposal ([#12834](https://github.com/bytecodealliance/wasmtime/pull/12834)) in the new wasmtime-wasi-tls crate, with both p2 and p3 bindings
([#12896](https://github.com/bytecodealliance/wasmtime/pull/12896)). We should evaluate wiring this into wash-runtime so components can open outbound TLS connections without a host-mediated HTTP client.

The crate targets wasi:tls@0.3.0-draft. It's a new, externally-developed interface: components call into a guest-visible Connector resource that lazily wraps a plain bytestream (from wasi:sockets or similar) in a TLS session.

## Today

We already use rustls server-side and client-side inside the host:
- crates/wash-runtime/src/host/http.rs: TlsAcceptor/TlsConnector for the inbound HTTP server and the outbound HTTP client
- crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs: rustls client config for tokio-postgres

None of this is exposed to guest components. Guests that want TLS today have to go through wasi:http (where the host terminates TLS).

## Proposed work

- Add wasmtime-wasi-tls = "44" as an optional workspace dep behind a new wasi-tls feature on wash-runtime. Default backend = rustls (matches our current stack).
- Extend the host context (SharedCtx) with a WasiTlsCtx field and implement WasiTlsView on the view type used by the linker.
- Wire wasmtime_wasi_tls::p2::add_to_linker with LinkOptions::tls(true) for the p2 path. Gate the call behind the wasi-tls feature and the existing per-workload policy (should TLS be opt-in per workload?).
- Wire wasmtime_wasi_tls::p3::add_to_linker behind the existing wasip3 feature.
- Register wasi:tls/*@0.3.0-draft in the host interface manifest (companion to the sockets entry at crates/wash-runtime/src/host/mod.rs:474).
- Test plan:
  - Unit: a guest component that opens a wasi:sockets TCP stream, wraps it with wasi:tls, and round-trips an HTTPS GET against a local rustls echo server.
  - Policy: confirm a workload without tls granted gets a link-time (not runtime) error.

## Design questions

1. Per-workload policy. wasi:sockets is already gated per workload; TLS should compose with that. Do we treat wasi:tls as a separate capability or bundle it with outbound network access?
2. Cert verification. WasiTlsCtxBuilder accepts a provider. Do we expose root-store configuration (custom CAs, mTLS client cert) per workload, or ship a single host-wide config first and extend later?

## Risks / unknowns

- wasi:tls is 0.3.0-draft. The WIT can still shift. Our bindings may change with every wasmtime bump.
- Resource lifetime around Connector and deferred handshake (Shared<Deferred<TlsStream>>): our custom sockets layer does some resource bookkeeping and we will need to confirm the TLS stream plays well with our p2 stream handling in crates/wash-runtime/src/sockets/p2_tcp.rs.
- wasmtime-wasi-tls pulls tokio + a TLS backend; verify no version conflicts with our existing rustls 0.23 / tokio-rustls 0.26 pins.

References

- wasmtime 44 release notes: added wasi:tls 0.3.0-draft
- [wasmtime PR #12834 ](https://github.com/bytecodealliance/wasmtime/pull/12834): initial p3 implementation
- [wasmtime PR #12896](https://github.com/bytecodealliance/wasmtime/pull/12896): unified WasiTlsView across p2 and p3
- [WebAssembly/wasi-tls](https://github.com/WebAssembly/wasi-tls): proposal repo
- wasmtime-wasi-tls example wiring: see the rustdoc on crates/wasi-tls/src/lib.rs

Aditya1404Sal:
Fixes #5076

(Note: LOC changes are highly deceptive because this PR ~~includes un-shared wit deps in the 2 new fixtures that were un-gitignored)~~

Now includes the respective wits in the shared wit directories

## Summary
Adds wasi:tls support to wash-runtime for ~~both WASIP2 (@0.2.0-draft) and~~ WASIP3 (@0.3.0-draft), gated behind a new wasi-tls cargo feature. Guest components can now establish TLS connections over wasi:sockets, with a pluggable provider so embedders can swap TLS backends, install custom trust stores (corporate CAs, certificate pinning), or integrate HSM-backed keys.

## Changes
Engine wiring (feat: add `wasi:tls@0.3.0-draft` support)

New `wasi-tls` feature flag pulling in `wasmtime-wasi-tls` with the p2, p3, and rustls features.
~~`wasmtime-wasi-tls::p2::add_to_linker` (with tls(true) link option) and~~ `p3::add_to_linker` registered in the engine.
~~`wasi:tls/types@0.2.0-draft`~~ and `wasi:tls/client,types@0.3.0-draft` advertised in the host's known-imports set so workloads can declare them.
Pluggable provider + tests (test: add ~~P2/~~P3 wasi:tls round-trip integration tests)

`EngineBuilder::with_tls_provider`, plumbed through UnresolvedWorkload → ResolvedWorkload → CtxBuilder so each component context picks up the override. None keeps the default rustls + system trust roots.

### ~~Two~~ One end-to-end fixture:
~~- `tls-echo-client-p2` — synchronous `wasi:sockets` TCP + `wasi:tls/types::ClientHandshake`, polling-based.~~
- `tls-echo-client-p3` — async `TcpSocket` + P3 streams + `wasi:tls/client::Connector`.
tests/integration_tls_socket.rs starts an in-process rustls echo server on a random loopback port, injects its self-signed cert via a custom TlsProvider, and verifies a PING round-trip for both fixtures.
Shared TLS test helpers extracted to tests/common/tls.rs (echo server, TestTlsProvider, engine builders).

### Shared WIT deps for TLS
~~build.rs gains a P3_SKIP_SHARED_WIT mechanism so tls-echo-client-p3 can manage its own WIT deps (it needs wasi:tls@0.3.0-draft alongside the standard P3 packages).~~
Following feedback from Bailey, the TLS WIT now lives in the shared WIT dirs so future fixtures can import it without copy-pasting:

P3 :  `wasi-tls-0.3.0-draft/` is a regular sibling under `p3-wit-deps/`, picked up by the default flat copy. No fixture-side changes needed beyond declaring the import in world.wit.

~~P2 :  has a wrinkle: `wasi:tls@0.2.0-draft` hard-pins `wasi:io@0.2.6` (`use wasi:io/streams@0.2.6.{...}`), which transitively requires the 0.2.6 versions of cli/clocks/sockets/random/filesystem. Putting those at the top level of `p2-wit-deps/` would force every other P2 fixture (currently on the 0.2.2 graph) to disambiguate. To avoid that, the TLS WIT plus its 0.2.6 dep set lives under `p2-wit-deps/tls-bundle/`. build.rs gains a per-fixture `P2_EXTRA_BUNDLES` opt-in so TLS-using fixtures pull this bundle on top of (or instead of) the default flat copy. New P2 fixtures that want TLS just add themselves to that list.~~

`copy_shared_wit_deps` now copies every `*.wit` in a package dir, not just `package.wit`, so multi-file packages like `wasi-tls@0.3.0-draft` (which ships client.wit + types.wit + world.wit from upstream) work without flattening.

> (Note: `wasi:tls@0.3.0-draft` should probably be published as a package to actually resolve this)

## Testing
Run the integration suite:

```bash
cargo test -p wash-runtime \
  --features "wash-runtime/wasi-tls,wash-runtime/wasip3" \
  --test integration_tls_socket
```
~~Both `test_p2_tls_tcp_round_trip` and ~~`test_p3_tls_tcp_round_trip` exercise:

- TCP connect to a non-loopback echo server (loopback is intercepted by the runtime)
- TLS handshake against a self-signed cert via the custom `TestTlsProvider`
- Plaintext `PING\r\n` write through the encrypted stream
- Receipt verified via oneshot channel on the server side


ricochet:
This is looking good to me and should already be mapped to what we added last week for `wash_runtime::init_crypto() in `crates/wash-runtime/src/lib.rs:32-43` 👍

crates/wash-runtime/tests/integration_tls_socket.rs
```diff
@@ -0,0 +1,120 @@
...
+        echo_client_workload_request("tls-echo-client-p3", TLS_ECHO_CLIENT_P3_WASM, echo_addr);
+
+    timeout(Duration::from_secs(15), host.workload_start(req))
+        .await
+        .context("P3 tls round-trip test timed out")?
+        .context("workload_start failed")?;
+
+    tokio::time::sleep(Duration::from_secs(3)).await;
+
+    Ok(())
```
ricochet, line 105: assert WorkloadState after a poll/wait for completion

check response.workload_status.workload_state == WorkloadState::Running (and not Error) immediately after workload_start returns

ricochet: Even better: the echo server could assert it received PING\r\n by surfacing the result through a oneshot channel.

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -367,6 +424,15 @@ impl CtxBuilder {
             component_id: self.component_id,
             http: WasiHttpCtx::new(),
             sockets: self.sockets.unwrap_or_default(),
+            #[cfg(feature = "wasi-tls")]
+            tls: {
+                crate::init_crypto();
```
ricochet: This is called every CtxBuilder::build()

init_crypto() uses a Once guard so it's safe, but calling it on every component invocation is unnecessary overhead. 

It should be called once at EngineBuilder::build() time (same pattern as HTTP's HttpServer::new). The init_crypto() call here should be moved to Engine::build().

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -367,6 +424,15 @@ impl CtxBuilder {
...
+                let mut builder = WasiTlsCtxBuilder::new();
+                if let Some(provider) = self.tls_provider {
+                    builder = builder.provider(Box::new(ArcTlsProvider(provider)));
```
ricochet: Should emit a warning when wasi-tls is enabled but no provider is set

ricochet: so operators know they're using the default from wasmtime-wasi-tls

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -230,14 +242,23 @@ pub mod workload;
...
     #[cfg(feature = "wasip3")]
     wasip3: bool,
+    /// TLS provider override for `wasi:tls` client connections.
+    #[cfg(feature = "wasi-tls")]
+    pub(crate) tls_provider: Option<Arc<dyn wasmtime_wasi_tls::TlsProvider>>,
+}
+
+impl std::fmt::Debug for Engine {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Engine").finish_non_exhaustive()
```
ricochet: Manual Debug impls lose field visibility

Engine { .. } with no fields makes {:?} output useless for debugging. Since only the tls_provider field blocks derive(Debug), consider adding #[debug(skip)] via the derive_more crate (if already in tree) or implementing it to include the non-TLS fields

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -481,6 +481,19 @@ pub struct ResolvedWorkload {
...
     /// The requested host [`WitInterface`]s to resolve this workload
     host_interfaces: Vec<WitInterface>,
+    /// TLS provider override for `wasi:tls` client connections in this workload.
+    #[cfg(feature = "wasi-tls")]
+    tls_provider: Option<Arc<dyn wasmtime_wasi_tls::TlsProvider>>,
+}
+
+impl std::fmt::Debug for ResolvedWorkload {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ResolvedWorkload")
```
ricochet: Manual Debug impls lose field visibility see other comment

crates/wash-runtime/build.rs
```diff
@@ -108,11 +108,17 @@ fn componentize(core_module: &[u8], adapter: &[u8]) -> Vec<u8> {
...
+/// `extra_bundles` is a per-fixture list of named subdirectories under
+/// the kind's shared dir to pull into the fixture in addition to (or
+/// instead of) the default flat copy. Used for version-pinned bundles
+/// like `tls-bundle/` whose contents conflict with the default deps.
 fn build_fixtures(
     workspace_dir: &Path,
     fixtures: &[&str],
     kind: FixtureKind,
     skip_shared_wit: &[&str],
+    extra_bundles: &[(&str, &[&str])],
```
ricochet: I don't understand the need for the extra_bundles. The TLS world doesn't have dependencies so there shouldn't be a conflict.

Aditya1404Sal: `wasi:tls@0.2.0-draft` hard-pins `wasi:io@0.2.6` (`use wasi:io/streams@0.2.6.{...}`), which transitively requires the 0.2.6 versions of cli/clocks/sockets/random/filesystem. Putting those at the top level of `p2-wit-deps/` would force every other P2 fixture (currently on the 0.2.2 graph) to disambiguate. To avoid that, the TLS WIT plus its 0.2.6 dep set lives under `p2-wit-deps/tls-bundle/`.

ricochet: Ahh P2 specific. Since wasi-tls is very, very unlikely to stabilize for P2, wdyt about making it p3 only?

Aditya1404Sal: Honestly I am also inclined to making it p3 only, better to phase out the unstable p2 as well.

Aditya1404Sal:
@ricochet  gonna transition wasi-tls impl to a wasip3 only addition.

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -15,9 +15,37 @@ use wasmtime::component::ResourceTable;
...
+                    Output = Result<
+                        Box<dyn wasmtime_wasi_tls::TlsStream>,
+                        wasmtime_wasi_tls::Error,
+                    >,
+                > + Send,
+        >,
+    > {
+        self.0.connect(server_name, transport)
+    }
+}
```
ricochet, line 59:
```suggestion
#[cfg(feature = "wasi-tls")]
struct ArcTlsProvider(Arc<dyn wasmtime_wasi_tls::TlsProvider>);

/// Concrete return type of [`wasmtime_wasi_tls::TlsProvider::connect`]. The
/// upstream alias (`BoxFutureTlsStream`) is `pub(crate)`, so we re-declare an
/// equivalent here to keep the impl signature readable.
#[cfg(feature = "wasi-tls")]
type TlsConnectFuture = std::pin::Pin<
    Box<
        dyn std::future::Future<
                Output = Result<Box<dyn wasmtime_wasi_tls::TlsStream>, wasmtime_wasi_tls::Error>,
            > + Send,
    >,
>;

#[cfg(feature = "wasi-tls")]
impl wasmtime_wasi_tls::TlsProvider for ArcTlsProvider {
    fn connect(
        &self,
        server_name: String,
        transport: Box<dyn wasmtime_wasi_tls::TlsTransport>,
    ) -> TlsConnectFuture {
        self.0.connect(server_name, transport)
    }
}
```

crates/wash-runtime/src/host/mod.rs
```diff
@@ -483,6 +483,8 @@ impl Host {
             "wasi:filesystem/preopens,types@0.2.0".into(),
             "wasi:random/insecure-seed,insecure,random@0.2.0".into(),
             "wasi:sockets/instance-network,ip-name-lookup,network,tcp-create-socket,tcp,udp-create-socket,udp@0.2.0".into(),
+            #[cfg(all(feature = "wasi-tls", feature = "wasip3"))]
```
ricochet:
```suggestion
            #[cfg(feature = "wasi-tls")]
```

crates/wash-runtime/tests/common/tls.rs
```diff
@@ -0,0 +1,240 @@
...
+            tokio::spawn(async move {
+                let mut tls_stream = match acceptor.accept(stream).await {
+                    Ok(s) => s,
+                    Err(_) => return,
+                };
+                let mut buf = vec![0u8; 256];
+                let mut received = Vec::new();
+                loop {
+                    match tls_stream.read(&mut buf).await {
+                        Ok(0) => break,
```
ricochet:
```suggestion
                        Ok(0) => return,
```

Subtle correctness fix: with break, an EOF before \r\n falls through to ping_slot.take().send(received) and fires the oneshot with a partial buffer — a malformed client would falsely pass the test. With return, no PING means no oneshot fire and the test correctly times out.

crates/wash-runtime/build.rs
```diff
@@ -190,13 +195,18 @@ fn copy_shared_wit_deps(shared_wit_dir: &Path, fixture_dir: &Path) -> anyhow::Re
         if !entry.file_type()?.is_dir() {
             continue;
         }
-        let pkg_wit = entry.path().join("package.wit");
-        if !pkg_wit.exists() {
+        let wit_files: Vec<_> = fs::read_dir(entry.path())?
+            .filter_map(|f| f.ok())
```
ricochet:
```suggestion
            .collect::<Result<Vec<_>, _>>()?
            .into_iter()
```

ricochet: We should make sure build failures propagate, e.g. if fs::read_dir returns an Err for an entry, file vanished between the parent listing and our open, fs corruption, etc that we fail.

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -230,14 +234,31 @@ pub mod workload;
...
 pub struct Engine {
     // wasmtime engine
     pub(crate) inner: wasmtime::Engine,
     pub(crate) cache: Cache<CacheKey, CacheValue>,
     /// Whether WASIP3 support is enabled for this engine.
     #[cfg(feature = "wasip3")]
     wasip3: bool,
+    /// TLS provider override for `wasi:tls` client connections.
+    #[cfg(feature = "wasi-tls")]
+    pub(crate) tls_provider: Option<Arc<dyn wasmtime_wasi_tls::TlsProvider>>,
```
ricochet:
```suggestion
    pub(crate) tls_provider: Option<SharedTlsProvider>,
```

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -230,14 +234,31 @@ pub mod workload;
...
+impl std::fmt::Debug for Engine {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut s = f.debug_struct("Engine");
+        #[cfg(feature = "wasip3")]
+        s.field("wasip3", &self.wasip3);
+        #[cfg(feature = "wasi-tls")]
+        s.field(
+            "tls_provider",
+            &self.tls_provider.as_ref().map(|_| "<TlsProvider>"),
+        );
```
ricochet:
```suggestion
        s.field("tls_provider", &self.tls_provider.is_some());
```

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -381,14 +402,23 @@ impl Engine {
...
             host_interfaces,
-        ))
+        );
+
+        #[cfg(feature = "wasi-tls")]
+        let workload = if let Some(provider) = self.tls_provider.clone() {
+            workload.with_tls_provider(provider)
+        } else {
+            workload
+        };
```
ricochet:
```suggestion
        let workload = workload.maybe_with_tls_provider(self.tls_provider.clone());
```

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -594,6 +624,9 @@ pub struct EngineBuilder {
     fuel_consumption: bool,
     #[cfg(feature = "wasip3")]
     wasip3: bool,
+    /// Optional TLS provider override for wasi:tls client connections.
+    #[cfg(feature = "wasi-tls")]
+    tls_provider: Option<Arc<dyn wasmtime_wasi_tls::TlsProvider>>,
```
ricochet:
```suggestion
    tls_provider: Option<SharedTlsProvider>,
```

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -654,6 +687,17 @@ impl EngineBuilder {
...
         self
     }
+
+    /// Override the TLS provider used for `wasi:tls` client connections.
+    ///
+    /// Use this to plug in an alternative TLS backend, install a custom root
+    /// certificate store (corporate CAs, certificate pinning), or integrate
+    /// with HSM-backed key material.
+    #[cfg(feature = "wasi-tls")]
+    pub fn with_tls_provider(mut self, provider: Arc<dyn wasmtime_wasi_tls::TlsProvider>) -> Self {
```
ricochet:
```suggestion
    pub fn with_tls_provider(mut self, provider: SharedTlsProvider) -> Self {
```

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -669,6 +713,9 @@ impl EngineBuilder {
     /// # Errors
     /// Returns an error if the wasmtime engine creation fails.
     pub fn build(mut self) -> anyhow::Result<Engine> {
+        #[cfg(feature = "wasi-tls")]
+        crate::init_crypto();
```
ricochet:
```suggestion
        {
            crate::init_crypto();
            if self.tls_provider.is_none() {
                tracing::warn!(
                    "wasi-tls is enabled but no TLS provider was set; \
                     falling back to wasmtime-wasi-tls default — \
                     set one via EngineBuilder::with_tls_provider"
                );
            }
        }
```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -15,9 +15,38 @@ use wasmtime::component::ResourceTable;
...
 use wasmtime_wasi_http::WasiHttpCtx;
 use wasmtime_wasi_http::p2::{WasiHttpCtxView, WasiHttpHooks, WasiHttpView};
+#[cfg(feature = "wasi-tls")]
+use wasmtime_wasi_tls::{WasiTlsCtx, WasiTlsCtxBuilder, WasiTlsCtxView, WasiTlsView};
 
 use crate::plugin::HostPlugin;
 
+/// Adapts a shareable `Arc<dyn TlsProvider>` into the owned `Box<dyn TlsProvider>`
+/// that `WasiTlsCtxBuilder::provider` requires. The same `Arc` can back many
+/// per-component contexts without re-creating the underlying provider.
```
ricochet:
```suggestion
/// Adapts a shareable [`super::SharedTlsProvider`] into the owned
/// `Box<dyn TlsProvider>` that `WasiTlsCtxBuilder::provider` requires. The
/// same `Arc` can back many per-component contexts without re-creating the
/// underlying provider.
```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -15,9 +15,38 @@ use wasmtime::component::ResourceTable;
...
+#[cfg(feature = "wasi-tls")]
+struct ArcTlsProvider(Arc<dyn wasmtime_wasi_tls::TlsProvider>);
```
ricochet:
```suggestion
struct ArcTlsProvider(super::SharedTlsProvider);

```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -15,9 +15,38 @@ use wasmtime::component::ResourceTable;
...
+/// upstream alias (`BoxFutureTlsStream`) is `pub(crate)`, so we re-declare an
+/// equivalent here to keep the impl signature readable.
+#[cfg(feature = "wasi-tls")]
+type TlsConnectFuture = std::pin::Pin<
+    Box<
+        dyn std::future::Future<
+                Output = Result<Box<dyn wasmtime_wasi_tls::TlsStream>, wasmtime_wasi_tls::Error>,
+            > + Send,
+    >,
+>;
```
ricochet, line 48:
```suggestion
>;

```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -289,6 +331,9 @@ pub struct CtxBuilder {
     plugins: HashMap<&'static str, Arc<dyn HostPlugin + Send + Sync>>,
     http_handler: Option<Arc<dyn crate::host::http::HostHandler>>,
     allowed_hosts: Arc<[String]>,
+    /// TLS provider override for `wasi:tls` client connections.
+    #[cfg(feature = "wasi-tls")]
+    tls_provider: Option<Arc<dyn wasmtime_wasi_tls::TlsProvider>>,
```
ricochet:
```suggestion
    tls_provider: Option<super::SharedTlsProvider>,
```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -302,9 +347,22 @@ impl CtxBuilder {
...
         }
     }
 
+    /// Override the TLS provider used for `wasi:tls` client connections.
+    ///
+    /// Use this to plug in an alternative TLS backend, install a custom root
+    /// certificate store (corporate CAs, certificate pinning), or integrate
+    /// with HSM-backed key material.
+    #[cfg(feature = "wasi-tls")]
+    pub fn with_tls_provider(mut self, provider: Arc<dyn wasmtime_wasi_tls::TlsProvider>) -> Self {
```
ricochet:
```suggestion
    pub fn with_tls_provider(mut self, provider: super::SharedTlsProvider) -> Self {
```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -367,6 +425,20 @@ impl CtxBuilder {
             component_id: self.component_id,
             http: WasiHttpCtx::new(),
             sockets: self.sockets.unwrap_or_default(),
+            #[cfg(feature = "wasi-tls")]
+            tls: {
```
ricochet, line 440:
```suggestion
            tls: {
                // EngineBuilder::build already warns once if no provider was
                // set; just fall through to the wasmtime-wasi-tls default
                // here when none is provided.
```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -367,6 +425,20 @@ impl CtxBuilder {
...
+                let mut builder = WasiTlsCtxBuilder::new();
+                if let Some(provider) = self.tls_provider {
+                    builder = builder.provider(Box::new(ArcTlsProvider(provider)));
+                } else {
+                    tracing::warn!(
+                        "wasi-tls is enabled but no TLS provider was set; \
+                         falling back to wasmtime-wasi-tls default — \
+                         set one via EngineBuilder::with_tls_provider"
+                    );
```
ricochet:
```suggestion

```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -481,6 +481,34 @@ pub struct ResolvedWorkload {
     service: Option<WorkloadService>,
     /// The requested host [`WitInterface`]s to resolve this workload
     host_interfaces: Vec<WitInterface>,
+    /// TLS provider override for `wasi:tls` client connections in this workload.
+    #[cfg(feature = "wasi-tls")]
+    tls_provider: Option<Arc<dyn wasmtime_wasi_tls::TlsProvider>>,
```
ricochet:
```suggestion
    tls_provider: Option<super::SharedTlsProvider>,
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -481,6 +481,34 @@ pub struct ResolvedWorkload {
...
+                    .components
+                    .try_read()
+                    .map(|g| g.len() as isize)
+                    .unwrap_or(-1),
+            );
+        #[cfg(feature = "wasi-tls")]
+        s.field(
+            "tls_provider",
+            &self.tls_provider.as_ref().map(|_| "<TlsProvider>"),
+        );
```
ricochet:
```suggestion
        s.field("tls_provider", &self.tls_provider.is_some());
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1292,6 +1325,9 @@ pub struct UnresolvedWorkload {
     service: Option<WorkloadService>,
     /// All [`WorkloadComponent`]s in the workload
     components: HashMap<Arc<str>, WorkloadComponent>,
+    /// TLS provider override for `wasi:tls` client connections in this workload.
+    #[cfg(feature = "wasi-tls")]
+    tls_provider: Option<Arc<dyn wasmtime_wasi_tls::TlsProvider>>,
```
ricochet:
```suggestion
    tls_provider: Option<super::SharedTlsProvider>,
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1329,9 +1365,22 @@ impl UnresolvedWorkload {
...
         }
     }
 
+    /// Override the TLS provider used for `wasi:tls` client connections in this workload.
+    ///
+    /// Use this to plug in an alternative TLS backend, install a custom root
+    /// certificate store (corporate CAs, certificate pinning), or integrate
+    /// with HSM-backed key material.
+    #[cfg(feature = "wasi-tls")]
+    pub fn with_tls_provider(mut self, provider: Arc<dyn wasmtime_wasi_tls::TlsProvider>) -> Self {
```
ricochet:
```suggestion
    pub fn with_tls_provider(mut self, provider: super::SharedTlsProvider) -> Self {
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1329,9 +1365,22 @@ impl UnresolvedWorkload {
...
+        self.tls_provider = Some(provider);
+        self
+    }
+
```
ricochet, line 1382:
```suggestion

    /// Apply an optional TLS provider override. No-op when `None`.
    #[cfg(feature = "wasi-tls")]
    pub fn maybe_with_tls_provider(self, provider: Option<super::SharedTlsProvider>) -> Self {
        match provider {
            Some(p) => self.with_tls_provider(p),
            None => self,
        }
    }

```

crates/wash-runtime/tests/common/tls.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+/// Wrap `cert_der` into an `Arc<dyn TlsProvider>` that trusts only that cert.
+fn test_tls_provider(cert_der: &[u8]) -> Result<Arc<dyn TlsProvider>> {
+    let client_config = Arc::new(client_config_with_cert(cert_der)?);
+    Ok(Arc::new(TestTlsProvider { client_config }) as Arc<dyn TlsProvider>)
+}
+
+/// Build an engine with P3 and a custom TLS provider that trusts `cert_der`.
+#[cfg(feature = "wasip3")]
```
ricochet:
```suggestion

```

crates/wash-runtime/tests/common/tls.rs
```diff
@@ -0,0 +1,240 @@
...
+        .with_wasip3(true)
+        .with_tls_provider(test_tls_provider(cert_der)?)
+        .build()
+        .context("failed to build P3+TLS engine")
+}
+
+/// Install the default `aws-lc-rs` rustls crypto provider exactly once.
+///
+/// Required when both `aws-lc-rs` and `ring` features are enabled (no
+/// unambiguous default). Safe to call repeatedly.
```
ricochet:
```suggestion
/// `EngineBuilder::build` already runs `wash_runtime::init_crypto` when
/// `wasi-tls` is enabled, but the test's TLS *server* (rustls `ServerConfig`,
/// rcgen cert generation) needs a provider installed *before* the engine is
/// built. Tests call this up-front so both sides agree on the provider; the
/// install is idempotent and safe to call repeatedly.
```

crates/wash-runtime/tests/common/tls.rs
```diff
@@ -0,0 +1,240 @@
...
+pub fn install_default_crypto_provider() {
+    let _ = tokio_rustls::rustls::crypto::aws_lc_rs::default_provider().install_default();
```
ricochet:
```suggestion
    wash_runtime::init_crypto();
```

crates/wash-runtime/tests/integration_tls_socket.rs
```diff
@@ -0,0 +1,108 @@
...
+use wash_runtime::{
+    host::{HostApi, HostBuilder},
+    types::{LocalResources, Service, Workload, WorkloadStartRequest, WorkloadState},
+};
+
+mod common;
+use common::tls::{EchoServer, install_default_crypto_provider, start_tls_echo_server};
+
+#[cfg(feature = "wasip3")]
+use common::tls::engine_with_p3_and_tls;
```
ricochet:
```suggestion
use common::tls::{
    EchoServer, engine_with_p3_and_tls, install_default_crypto_provider, start_tls_echo_server,
};
```

crates/wash-runtime/tests/integration_tls_socket.rs
```diff
@@ -0,0 +1,108 @@
...
+
+#[cfg(feature = "wasip3")]
```
ricochet:
```suggestion

```

crates/wash-runtime/src/engine/mod.rs
ricochet, line 234:
```suggestion

/// A shareable handle to a `wasi:tls` provider. Kept as an `Arc` so the same
/// provider can back many per-component contexts without re-creating it.
#[cfg(feature = "wasi-tls")]
pub(crate) type SharedTlsProvider = Arc<dyn wasmtime_wasi_tls::TlsProvider>;

/// The core WebAssembly engine for executing components and workloads.
```

crates/wash-runtime/src/engine/mod.rs
ricochet:
```suggestion
    pub(crate) tls_provider: Option<SharedTlsProvider>,
```

crates/wash-runtime/src/engine/mod.rs
ricochet:
```suggestion
        s.field("tls_provider", &self.tls_provider.is_some());
```

crates/wash-runtime/src/engine/mod.rs
ricochet:
```suggestion
        let workload = workload.maybe_with_tls_provider(self.tls_provider.clone());
```

crates/wash-runtime/src/engine/mod.rs
ricochet:
```suggestion
    tls_provider: Option<SharedTlsProvider>,
```

crates/wash-runtime/src/engine/mod.rs
ricochet:
```suggestion
    pub fn with_tls_provider(mut self, provider: SharedTlsProvider) -> Self {
```

crates/wash-runtime/src/engine/value.rs
```diff
@@ -270,6 +270,9 @@ mod tests {
     use crate::engine::ctx::Ctx;
 
     fn make_store() -> wasmtime::Store<SharedCtx> {
+        #[cfg(feature = "wasi-tls")]
+        crate::init_crypto();
```
ricochet, line 274:
```suggestion
        #[cfg(feature = "wasi-tls")]
        {
            crate::init_crypto();
            if self.tls_provider.is_none() {
                tracing::warn!(
                    "wasi-tls is enabled but no TLS provider was set; \
                     falling back to wasmtime-wasi-tls default — \
                     set one via EngineBuilder::with_tls_provider"
                );
            }
        }
```

ricochet: Sorry for the back and forth on this one. I needed to pull this down and walk through. 

CtxBuilder::build() runs per component, not per engine. So a workload with 50 components produces 50 identical warning lines on every start.

crates/wash-runtime/src/engine/mod.rs
ricochet: Lets alias this. Its a complex type with a number of call sites.

ricochet (dismissed):
We'll want to carefully rebase on main before merging now that the OutgoingHandler change is in.

These are all small nits.

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -15,9 +15,40 @@ use wasmtime::component::ResourceTable;
...
+use wasmtime_wasi_tls::{WasiTlsCtx, WasiTlsCtxBuilder, WasiTlsCtxView, WasiTlsView};
 
 use crate::plugin::HostPlugin;
 
+/// Adapts a shareable [`super::SharedTlsProvider`] into the owned
+/// `Box<dyn TlsProvider>` that `WasiTlsCtxBuilder::provider` requires. The
+/// same `Arc` can back many per-component contexts without re-creating the
+/// underlying provider.
+#[cfg(feature = "wasi-tls")]
+struct ArcTlsProvider(super::SharedTlsProvider);
```
vados-cosmonic: Why is this called literally `Arc`? Can we not implement the relevant traits on `SharedTlsProvider` ?

vados-cosmonic (commented):
Mostly looks great to me (I only have one nit!) -- thanks for working on this @Aditya1404Sal

crates/wash-runtime/src/engine/ctx.rs
Aditya1404Sal: The orphan rule prevents implementing TlsProvider (a foreign trait) directly on `Arc<dyn TlsProvider>` (foreign type) in this crate, so the newtype is unavoidable. Renamed it to `SharedProvider` to drop the implementation detail from the name though.

vados-cosmonic: Hey thanks -- can we remove the type alias for `SharedTlsProvider`? Something like:

```rust
struct SharedTlsProvider(Arc<dyn TlsProvider>)
```

It just doesn't seem like we need to spend any time holding a `Arc<dyn TlsProvider>` without wrapping it in the `SharedTlsProvider` immediately (and moving that around as necessary).

ricochet (approved):

PR #5117, opened 2026-05-04
feat(wash): workload env/config/secrets

### LINKED ISSUE #5088: Add workload-level env, secrets, and config to `.wash/config.yaml`
## Summary

`.wash/config.yaml` today describes the *local dev environment* (TLS paths, plugin backends, etc.) but not the *workload*.

This adds a small `workload:` section covering only those gaps, plus top-level `configs:` and `secrets:` blocks for the values they reference. Field names mirror `WorkloadDeployment.spec.template.spec.components[].localResources` (`runtime-operator/api/runtime/v1alpha1/workload_types.go:69`) so the same shape exports cleanly to a Kubernetes `WorkloadDeployment`.

## Motivation

Env vars, secrets, and runtime config are first-class on `WorkloadDeployment` (via `LocalResources` / `ConfigLayer`, `workload_types.go:52-80`) but unrepresentable in `.wash/config.yaml`. The closest existing field is `BuildConfig.env` (`crates/wash/src/config.rs:108`), which is build-time only.

## Goals

1. Add a `workload:` section for env vars, secrets, runtime config, and allowed hosts on the component under development.
2. Add top-level `configs:` and `secrets:` blocks where each named entry pulls from inline literals, local files, or shell env vars. Make it so that these could round-trip 1:1 to a `ConfigMap` / `Secret`.

## Non-goals

- **Pass-through of dev-loop fields.** Host interfaces, volumes, and the Wasm service file already work via `dev.host_interfaces`, `dev.volumes`, `dev.service_file`. No parallel `workload.*` copies. Bringing the rest of the `WorkloadDeployment` surface across is also out of scope.
- **Multi-environment composition.** No profiles, overlays, or environment inheritance in the file. Users who need a second local shape pass `--config <path>` (loader already supports it, `crates/wash/src/config.rs:282`). Production multi-environment lives in GitOps / Helm / the K8s manifests.
- **A `wash export workload` CLI** — worth a follow-up issue; the schema is being shaped to make it trivial.

## Prior art

- **Docker Compose** has top-level `configs:` and `secrets:` blocks where each entry is a *source* (`file:`, `environment:`, `external:`); services reference them by name. The shape proposed here is the same pattern with sources adapted for local dev.
- **Kustomize** (`secretGenerator` / `configMapGenerator`) is the closest match for the source-types subschema specifically. Each generator entry accepts `literals: ["KEY=VALUE"]`, `envs: [path/to/.env]`, and `files: [key=path]`, combines them additively, and emits a `Secret` / `ConfigMap` (no base64 needed — Kustomize handles encoding). Field names worth comparing against ours (see Open Questions).
- **Tilt** is the closest analog in spirit: a dev-loop tool for Kubernetes apps that points at existing manifests rather than redefining them.

### 1. `workload:` section

The component is implicit: it's the project itself, built per `build.component_path`.

```yaml
workload:
  environment:
    config:                     # inline plain values
      LOG_LEVEL: debug
    configFrom:                 # named refs to top-level `configs:`
      - name: app-settings
    secretFrom:                 # named refs to top-level `secrets:`
      - name: db-credentials
  config:                       # wasi:config/runtime values
    feature.flags: "v2,beta"
  allowedHosts:
    - https://api.example.com
```

### 2. Top-level `configs:` and `secrets:` blocks

`configFrom` / `secretFrom` reference named ConfigMaps/Secrets in Kubernetes. The local equivalent: each named entry is a *source*.

```yaml
configs:
  app-settings:
    inline:
      THEME: dark
      FEATURE_X: "true"
  feature-flags:
    file: ./config/flags.env

secrets:
  db-credentials:
    fromEnv: [DB_PASSWORD, DB_USER]      # pull from the developer's shell
  kv-credentials:
    file: ~/.config/wash/kv.secret.yaml  # gitignored local file
```

Resolution rules:

- `inline` — map of literal values.
- `file` — path relative to project dir; format inferred by extension (`.env`, `.yaml`, `.json`).
- `fromEnv` — list of env var names; missing vars are an error unless marked `optional: true`.
- Multiple sources in one entry merge last-wins (matches K8s, `workload_types.go:54-56`).

## Mapping to `WorkloadDeployment`

| `.wash/config.yaml` | `WorkloadDeployment` field |
|---|---|
| `workload.environment` | `spec.template.spec.components[<this>].localResources.environment` |
| `workload.config` | `….localResources.config` |
| `workload.allowedHosts` | `….localResources.allowedHosts` |
| `configs:` / `secrets:` | Synthesized `ConfigMap` / `Secret` resources |
| existing `dev.*` fields | (none) — local-host concerns, not exported |

## Backward compatibility

Strictly additive. No existing field changes meaning, no deprecations.

## Security considerations

`configs:` and `secrets:` are designed for the local dev loop, but the failure modes are well-known. Some to keep in mind:

1. **Canonicalize every `file:` path; reject paths that escape the project root or follow untrusted symlinks.** This is exactly the CVE-2025-62725 class — Compose joined attacker-controlled path strings with its cache directory without canonicalization, allowing arbitrary host file writes even from read-only commands like `docker compose ps`.
2. **Refuse to load secret source files with group- or world-readable mode** (require 0600/0400). Mirrors the Kubernetes `defaultMode: 0400` push after years of 644-default complaints on Secret volumes (k8s #28317, #129043).
3. **Logging hygiene**: secret values never appear in logs, traces, or error messages — error paths reference the secret by name only. Particularly relevant given `wash dev`'s OpenTelemetry hook (`dev.wasi_otel`). Same class as CVE-2025-3911 / CVE-2025-6587 (Docker Desktop env vars in diagnostic logs).
4. **Warn when a secret `file:` path resolves inside the repo working tree**. Mass `.env` exfiltration campaigns (Unit 42 reported ~110,000 compromised domains) are exactly this class.
5. **Don't persist decoded secret values to disk.** Resolve `secrets:` fresh per `wash dev` invocation; hold values in process memory only. Vercel's April 2026 incident exposed customers' non-sensitive environment variables because they were stored in a form that decrypted to plaintext at rest — once an attacker reached the storage layer via a compromised employee account, enumeration was straightforward. The local laptop is the same threat model: assume the disk is recoverable and don't write a cache the user didn't ask for. Fly's architecture is the positive analog (encrypt-only API; decryption only at the host that needs the value, with a short-lived token).
6. **Defer remote/OCI sources for `configs:` / `secrets:`.** The instant we add them, we inherit the full CVE-2025-62725 surface (annotation trust, cache-directory confinement, signature verification).

## Implementation sketch

- New `crates/wash/src/config/workload.rs` with `WorkloadConfig` and `ConfigLayer`. Field names match Go CRD JSON tags so YAML keys are identical.
- New `crates/wash/src/config/sources.rs` with `ConfigSource` / `SecretSource` enums and a resolver returning `HashMap<String, String>`.
- `wash dev`: resolve `configs:` / `secrets:`, then pass `workload.environment` as env vars, `workload.config` through wasi:config/runtime, and `workload.allowedHosts` to the outbound HTTP plugin's allowlist.
- JSON Schema under `crates/wash/schemas/` for editor autocomplete.

## Open questions

1. **Placement: top-level `workload:` vs nested `dev.workload:`.** Top-level mirrors the CRD shape and reads cleanly for future consumers (`wash run`, `wash export workload`), but implies wash owns the canonical workload definition. Nested keeps everything local-dev-scoped and matches the framing in this proposal. I could go either way here and the schema works either way; just add or strip a `dev.` prefix.
2. **Section name**: `workload:` vs `deployment:` vs `spec:`. Lean: `workload:`.
3. **Source field names**. Three reasonable choices, no clear winner:
   - **Current (`inline` / `file` / `fromEnv`)** — distinguishes "values written here" vs "values from elsewhere" cleanly; wash specific.
   - **Kustomize-aligned (`literals` / `envs` / `files` + a wash-specific `fromEnv`)** — muscle memory for K8s users, but `literals` takes a list of `"KEY=VALUE"` strings rather than a YAML map, which is uglier.
   - **Kubernetes-aligned (`data` / `file` / `fromEnv`)** — `data:` matches the field name on the resource these export to (`ConfigMap.data`, `Secret.stringData`); makes round-trip visible in syntax.
4. **Secret file format**: `.env` only first. Just note I considered sops behind a flag.
5. **`fromEnv` ergonomics**: support `fromEnv: { DB_PASSWORD: APP_PASSWORD }` for renaming the var between shell and component, or keep it list-only? Map form adds parser complexity but means devs don't have to rename their shell vars to match what the component expects.

## Acceptance criteria

- A project sets env vars in `.wash/config.yaml` and `wash dev` exposes them to the running component.
- A project references a secret from the developer's shell (`fromEnv`) without committing it.
- The `workload:` / `configs:` / `secrets:` shape exports cleanly to the matching `localResources` paths on a `WorkloadDeployment`.
- All existing `.wash/config.yaml` files in `examples/` keep working unchanged.

### LINKED ISSUE #5099: ability to configure component_path when using wash dev
Currently wash dev doesn't implement all the options that wash build has which makes it pretty limiting.

I saw there was already a ticket for the env option.
I also really miss the component_path option because now like in rust you can set a different build command.
But then it often can't find the build component in rust because like the component is in debug mode instead of released mode it is in a different folder.

ricochet:
Adds a `workload:` section to .wash/config.yaml with `environment`
(inline + configFrom + secretFrom), opaque `config`, and outbound
`allowedHosts`. Top-level `configs:` and `secrets:` carry named
sources of three kinds: `inline`, `file:` (.env format), and
`fromEnv:`. Field shape mirrors WorkloadDeployment.localResources
so the same file can round-trip to a Kubernetes manifest later.

Resolution lives in `crates/wash/src/workload.rs`. Secrets get a
stricter posture: file paths canonicalized and confined to the
project dir (CVE-2025-62725 class), 0600/0400 mode required on
Unix, gitignore-aware warning when a secret file resolves inside
the repo working tree. Errors and traces reference sources by name
only - resolved values never appear in log output.

`wash dev` now resolves the workload before deploy and applies the
result to the dev component's LocalResources. The wasi:config
plugin is constructed with copy_environment=true, so workload env
vars surface via both wasi:cli/env and wasi:config/store::get.
`workload.config` is injected into the wasi:config interface entry
when at least one component in the workload imports it, sidecars
included.

create_workload is split into a thin I/O wrapper plus a pure
build_workload that the unit tests exercise end-to-end. Also fixes
a pre-existing gap by honoring `dev.component_path` (lets the dev
build command produce its artifact in a different folder than the
release build).

Fixes #5088
Fixes #5099

### DIFF
--- a/crates/wash-runtime/src/plugin/wasi_otel/mod.rs
+++ b/crates/wash-runtime/src/plugin/wasi_otel/mod.rs
@@ -32,6 +32,12 @@ use crate::wit::{WitInterface, WitWorld};
 
 const WASI_OTEL_ID: &str = "wasi-otel";
 
+/// OTel gRPC default per the OTLP/gRPC spec. Matches what
+/// `opentelemetry_otlp::SpanExporter::builder().with_tonic()` falls back to
+/// when no `OTEL_EXPORTER_OTLP_*_ENDPOINT` is set; duplicated here only so
+/// the log line at plugin start reflects what the exporter actually used.
+const DEFAULT_OTLP_GRPC_ENDPOINT: &str = "http://localhost:4317";
+
 mod bindings {
     wasmtime::component::bindgen!({
         world: "otel",
@@ -41,17 +47,33 @@ mod bindings {
 
 use bindings::wasi::otel::tracing::{SpanContext as WitSpanContext, TraceFlags as WitTraceFlags};
 
-/// Plugin configuration
-#[derive(Clone, Debug)]
+/// Configuration for the [`WasiOtel`] plugin.
+///
+/// Construct via [`WasiOtelConfig::builder`] (or [`Default::default`]) to
+/// stay forward-compatible with new fields.
+///
+/// # Examples
+///
+/// ```
+/// use wash_runtime::plugin::wasi_otel::WasiOtelConfig;
+///
+/// let cfg = WasiOtelConfig::builder()
+///     .service_name("my-service")
+///     .build();
+/// assert_eq!(cfg.service_name, "my-service");
+/// ```
+#[derive(Clone, Debug, bon::Builder)]
+#[non_exhaustive]
 pub struct WasiOtelConfig {
+    /// `service.name` resource attribute attached to all exported spans,
+    /// metrics, and logs. Defaults to the plugin id (`wasi-otel`).
+    #[builder(default = WASI_OTEL_ID.to_string(), into)]
     pub service_name: String,
 }
 
 impl Default for WasiOtelConfig {
     fn default() -> Self {
-        Self {
-            service_name: "wasi-otel".to_string(),
-        }
+        Self::builder().build()
     }
 }
 
@@ -104,11 +126,12 @@ impl HostPlugin for WasiOtel {
     async fn start(&self) -> anyhow::Result<()> {
         // The exporter resolves its endpoint from `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`,
         // then `OTEL_EXPORTER_OTLP_ENDPOINT`, falling back to the OTel gRPC default
-        // (`http://localhost:4317`). Protocol is fixed to gRPC because we use
-        // `with_tonic()` below.
+        // ([`DEFAULT_OTLP_GRPC_ENDPOINT`]). Protocol is fixed to gRPC because we use
+        // `with_tonic()` below; tracking richer per-target endpoint configuration as a
+        // follow-up to this PR (see TODO below).
         let endpoint = std::env::var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")
             .or_else(|_| std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT"))
-            .unwrap_or_else(|_| "http://localhost:4317".to_string());
+            .unwrap_or_else(|_| DEFAULT_OTLP_GRPC_ENDPOINT.to_string());
         tracing::info!(
             endpoint = %endpoint,
             protocol = "grpc",
--- a/crates/wash-runtime/src/plugin/wasi_otel/mod.rs
+++ b/crates/wash-runtime/src/plugin/wasi_otel/mod.rs
@@ -44,21 +44,13 @@ use bindings::wasi::otel::tracing::{SpanContext as WitSpanContext, TraceFlags as
 /// Plugin configuration
 #[derive(Clone, Debug)]
 pub struct WasiOtelConfig {
-    pub endpoint: String,
-    pub protocol: String,
     pub service_name: String,
-    pub propagate_context: bool,
-    pub batch_timeout_ms: u64,
 }
 
 impl Default for WasiOtelConfig {
     fn default() -> Self {
         Self {
-            endpoint: String::new(),
-            protocol: String::new(),
             service_name: "wasi-otel".to_string(),
-            propagate_context: true,
-            batch_timeout_ms: 5000,
         }
     }
 }
@@ -110,14 +102,21 @@ impl HostPlugin for WasiOtel {
     }
 
     async fn start(&self) -> anyhow::Result<()> {
+        // The exporter resolves its endpoint from `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`,
+        // then `OTEL_EXPORTER_OTLP_ENDPOINT`, falling back to the OTel gRPC default
+        // (`http://localhost:4317`). Protocol is fixed to gRPC because we use
+        // `with_tonic()` below.
+        let endpoint = std::env::var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")
+            .or_else(|_| std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT"))
+            .unwrap_or_else(|_| "http://localhost:4317".to_string());
         tracing::info!(
-            endpoint = %self.config.endpoint,
-            protocol = %self.config.protocol,
+            endpoint = %endpoint,
+            protocol = "grpc",
             "Starting WASI OTel plugin"
         );
 
-        // TODO: Add configurable endpoints/protocols to use. This would be beneficial for when you want to have Host otel go to Platform engineering teams,
-        // And Workload otel go to a different backend for application monitoring.
+        // TODO: thread per-target endpoints (host vs workload) through `WasiOtelConfig`
+        // so platform telemetry and application telemetry can ship to different backends.
 
         // set up the grpc span exporter
         let span_exporter = SpanExporter::builder()
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -81,8 +81,11 @@ impl CliCommand for DevCommand {
         // `wasi:config/store::get`, so values from `workload.environment` are
         // visible to the component as both env vars and wasi:config entries
         // without further plumbing.
-        host_builder =
-            host_builder.with_plugin(Arc::new(plugin::wasi_config::DynamicConfig::new(true)))?;
+        host_builder = host_builder.with_plugin(Arc::new(
+            plugin::wasi_config::DynamicConfig::builder()
+                .copy_environment(true)
+                .build(),
+        ))?;
 
         // Enable wasmcloud:messaging
         host_builder = host_builder.with_plugin(Arc::new(
@@ -616,10 +619,6 @@ mod tests {
         workload.components.iter().find(|c| c.name == name)
     }
 
-    // ---------------------------------------------------------------------
-    // Tests for `build_workload` used by `wash dev`
-    // ---------------------------------------------------------------------
-
     #[test]
     fn build_workload_lands_workload_values_on_dev_component_only() {
         // The dev component (when not running as a service) is the one
@@ -628,7 +627,7 @@ mod tests {
         let resolved = ResolvedWorkload {
             environment: HashMap::from([("LOG".into(), "debug".into())]),
             config: HashMap::from([("flag".into(), "on".into())]),
-            allowed_hosts: vec!["https://api.example.com".into()],
+            allowed_hosts: vec!["https://api.example.com".parse().unwrap()],
         };
         let dev_cfg = DevConfig {
             components: vec![dev_component_named("sidecar-a")],
@@ -654,7 +653,7 @@ mod tests {
         assert_eq!(dev.local_resources.config.get("flag").unwrap(), "on");
         assert_eq!(
             &dev.local_resources.allowed_hosts[..],
-            &["https://api.example.com".to_string()]
+            &["https://api.example.com".parse().unwrap()]
         );
 
         let sidecar = find_component(&workload, "sidecar-a").unwrap();
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -56,6 +56,7 @@ impl CliCommand for DevCommand {
         .context("failed to load config for development")?;
 
         let dev_config = config.dev();
+
         let http_addr = dev_config
             .address
             .clone()
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 use anyhow::{Context as _, bail, ensure};
 use bytes::Bytes;
@@ -14,12 +17,14 @@ use wash_runtime::{
         Component, HostPathVolume, LocalResources, Service, Volume, VolumeMount, VolumeType,
         Workload, WorkloadStartRequest, WorkloadState, WorkloadStopRequest,
     },
+    wit::WitInterface,
 };
 
 use crate::{
     cli::{CliCommand, CliContext, CommandOutput, component_build::build_dev_component},
     config::{Config, load_config},
     wit::WitConfig,
+    workload::{ResolvedWorkload, resolve_workload},
 };
 
 /// Start a development server for a Wasm component
@@ -70,9 +75,13 @@ impl CliCommand for DevCommand {
             .with_engine(engine)
             .with_meters(Meters::new(ctx.enable_meters()));
 
-        // Enable wasi config
+        // Enable wasi config. `copy_environment = true` causes the plugin to
+        // surface every entry of `LocalResources.environment` via
+        // `wasi:config/store::get`, so values from `workload.environment` are
+        // visible to the component as both env vars and wasi:config entries
+        // without further plumbing.
         host_builder =
-            host_builder.with_plugin(Arc::new(plugin::wasi_config::DynamicConfig::default()))?;
+            host_builder.with_plugin(Arc::new(plugin::wasi_config::DynamicConfig::new(true)))?;
 
         // Enable wasmcloud:messaging
         host_builder = host_builder.with_plugin(Arc::new(
@@ -233,7 +242,14 @@ impl CliCommand for DevCommand {
         let wasm_bytes = tokio::fs::read(&build_result.component_path)
             .await
             .context("failed to read component file")?;
-        let workload = create_workload(&host, &config, wasm_bytes.into()).await?;
+        // Resolve workload-level env / config / allowed_hosts now (before
+        // running the component) so that bad config or missing secrets fail
+        // before we deploy. Pass project_dir as the repo root for the
+        // gitignored-secret warning.
+        let resolved_workload = resolve_workload(&config, project_dir, Some(project_dir))
+            .context("failed to resolve workload-level configuration")?;
+        let workload =
+            create_workload(&host, &config, wasm_bytes.into(), &resolved_workload).await?;
         // Running workload ID for reloads
         let workload_id = reload_component(host.clone(), &workload, None).await?;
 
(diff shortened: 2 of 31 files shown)


crates/wash-runtime/src/plugin/wasi_otel/mod.rs
```diff
@@ -44,21 +44,13 @@ use bindings::wasi::otel::tracing::{SpanContext as WitSpanContext, TraceFlags as
 /// Plugin configuration
 #[derive(Clone, Debug)]
 pub struct WasiOtelConfig {
-    pub endpoint: String,
-    pub protocol: String,
     pub service_name: String,
```
vados-cosmonic, line 71: these are breaking changes and we probably don't care so much since it's internal, but we should make this a builder, probably

crates/wash-runtime/src/plugin/wasi_otel/mod.rs
```diff
@@ -110,14 +102,21 @@ impl HostPlugin for WasiOtel {
     }
 
     async fn start(&self) -> anyhow::Result<()> {
+        // The exporter resolves its endpoint from `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`,
+        // then `OTEL_EXPORTER_OTLP_ENDPOINT`, falling back to the OTel gRPC default
+        // (`http://localhost:4317`). Protocol is fixed to gRPC because we use
+        // `with_tonic()` below.
+        let endpoint = std::env::var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")
+            .or_else(|_| std::env::var("OTEL_EXPORTER_OTLP_ENDPOINT"))
+            .unwrap_or_else(|_| "http://localhost:4317".to_string());
```
vados-cosmonic: NIT: probably should make this a constant somewhere

crates/wash-runtime/src/plugin/wasi_otel/mod.rs
```diff
@@ -110,14 +102,21 @@ impl HostPlugin for WasiOtel {
...
         tracing::info!(
-            endpoint = %self.config.endpoint,
-            protocol = %self.config.protocol,
+            endpoint = %endpoint,
+            protocol = "grpc",
```
vados-cosmonic, line 137: did we remove the ability to do either? I'm all for making sure we put gRPC if that's all we support -- would probably want a tiny bit of documentation somewhere (and encouraging people who absolutely need/want JSON to file a ticket or something)

crates/wash-runtime/src/plugin/wasi_otel/mod.rs
```diff
@@ -110,14 +102,21 @@ impl HostPlugin for WasiOtel {
...
             "Starting WASI OTel plugin"
         );
 
-        // TODO: Add configurable endpoints/protocols to use. This would be beneficial for when you want to have Host otel go to Platform engineering teams,
-        // And Workload otel go to a different backend for application monitoring.
+        // TODO: thread per-target endpoints (host vs workload) through `WasiOtelConfig`
+        // so platform telemetry and application telemetry can ship to different backends.
```
vados-cosmonic, line 142: IMO if you have the context for this, might as well make it a ticket with as much context as possible so someone can pick it up. 

If I'm understanding this correctly this is basically a completely new feature -- i.e. being able to take two different otel configs for platform vs application, but there's a lot of design space there IMO (maybe I'm thinking of it as being too complex)

crates/wash/src/config.rs
```diff
@@ -44,6 +44,20 @@ pub struct Config {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub workload: Option<WorkloadConfig>,
+
+    /// Named ConfigMap-equivalent sources referenced by `workload.environment.configFrom`.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub configs: HashMap<String, ConfigSource>,
+
+    /// Named Secret-equivalent sources referenced by `workload.environment.secretFrom`.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub secrets: HashMap<String, ConfigSource>,
```
vados-cosmonic:
```suggestion
    pub secret_sources: HashMap<String, ConfigSource>,
```

Should we call these `_sources` instead? Same goes for `configs`

Also, I assume `SecretSource` doesn't exist, but I wonder if they'll diverge at some point. 

Also, we should use `BTreeMap` quite often over `HashMap` IIRC to maintain `Ord`, etc. I feel like we've run into this before and generally using `HashMap`s would break down when we were serializing... But I can't remember the exact case/reason why -- do you remember? It's fine if it's absolutely not an issue at this point (and IMO we can change it in the future without too much fuss easily I think)

crates/wash/src/config.rs
```diff
@@ -115,6 +132,67 @@ pub struct BuildConfig {
     pub component_path: Option<PathBuf>,
 }
 
+/// Workload-level configuration that mirrors the `localResources` shape of a
+/// `WorkloadDeployment` component. Currently consumed by `wash dev`; the same
+/// shape is intended to round-trip to a Kubernetes `WorkloadDeployment`.
```
vados-cosmonic:
```suggestion
/// Workload-level configuration that mirrors the `localResources` shape of a
/// `WorkloadDeployment` component. 
/// 
/// Currently consumed by `wash dev`; the same shape is intended to round-trip 
/// to a Kubernetes `WorkloadDeployment`.
```

crates/wash/src/config.rs
```diff
@@ -115,6 +132,67 @@ pub struct BuildConfig {
...
+    /// values with named references to top-level `configs:` and `secrets:`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub environment: Option<EnvironmentLayer>,
+    /// Opaque key-value config delivered to the component (e.g. wasi:config/store).
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub config: HashMap<String, String>,
+    /// Outbound HTTP allowlist. Empty means unrestricted (matches runtime
+    /// behavior in `crates/wash-runtime/src/host/http.rs`).
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub allowed_hosts: Vec<String>,
```
vados-cosmonic: I wish we could hold something better than a `String` here

crates/wash/src/config.rs
```diff
@@ -115,6 +132,67 @@ pub struct BuildConfig {
...
+}
+
+/// One layer of environment variables. Inline values are written directly;
+/// `configFrom` / `secretFrom` reference named entries in the top-level
+/// `configs:` / `secrets:` blocks by name. On key conflicts later entries win,
+/// in order: inline → configFrom → secretFrom (matches K8s envFrom semantics).
```
vados-cosmonic:
```suggestion
/// One layer of environment variables. 
///
/// Inline values are written directly;
/// `configFrom` / `secretFrom` reference named entries in the top-level
/// `configs:` / `secrets:` blocks by name. On key conflicts later entries win,
/// in order: inline → configFrom → secretFrom (matches K8s `envFrom` semantics).
```

crates/wash/src/config.rs
```diff
@@ -115,6 +132,67 @@ pub struct BuildConfig {
...
+    pub config_from: Vec<String>,
+    /// Names of entries in the top-level `secrets:` block.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub secret_from: Vec<String>,
+}
+
+/// A source of key-value pairs for a `configs:` or `secrets:` entry. Multiple
+/// fields can be set on a single entry.. They merge last-wins in the order
+/// `inline` → `file` → `fromEnv` (matches K8s ConfigMap merge semantics).
+///
```
vados-cosmonic:
```suggestion
/// A source of key-value pairs for a `configs:` or `secrets:` entry. 
///
/// Multiple fields can be set on a single entry. They merge last-wins in the order
/// `inline` → `file` → `fromEnv` (matches K8s ConfigMap merge semantics).
///
```

crates/wash/src/config.rs
```diff
@@ -115,6 +132,67 @@ pub struct BuildConfig {
...
+/// Distinct types are not used for config vs secret because the schema is
+/// identical; the difference is the security posture applied at resolve time
+/// (see [`crate::workload::resolve_workload`]).
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ConfigSource {
+    /// Literal key-value entries.
```
vados-cosmonic:
```suggestion
    /// Literal key-value entries supplied inline
```

crates/wash/src/config.rs
```diff
@@ -115,6 +132,67 @@ pub struct BuildConfig {
     pub component_path: Option<PathBuf>,
 }
 
+/// Workload-level configuration that mirrors the `localResources` shape of a
+/// `WorkloadDeployment` component. Currently consumed by `wash dev`; the same
+/// shape is intended to round-trip to a Kubernetes `WorkloadDeployment`.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct WorkloadConfig {
```
vados-cosmonic, line 210: We may want to use builders & `#[non_exhaustive]` here -- [`bon`](https://crates.io/crates/bon) ? [`typed-builder`](https://crates.io/crates/typed-builder) ?

crates/wash/src/main.rs
```diff
@@ -147,6 +147,18 @@ async fn main() {
 
     let (mut stdout, mut stderr) = (Box::new(std::io::stdout()), Box::new(std::io::stderr()));
 
+    // For `wash dev`, export `dev.environment` from the project config into the
+    // wash process before observability initializes. The tracing subscriber reads
+    // `RUST_LOG` and the OTel exporter checks for `OTEL_*` once during
+    // `initialize_observability`. Neither honors changes made later, so values
+    // set inside `DevCommand::handle()` arrive too late.
+    if matches!(global_args.command, Some(WashCliCommand::Dev(_))) {
+        wash::config::apply_dev_environment(
```
vados-cosmonic, line 156: I wonder if we can make this explicit instead of relying on the unsafe to set ENV vars -- i.e. being able to add the relevant modified environment to the `DevCommand` before it gets run

crates/wash/src/config.rs
```diff
@@ -333,6 +426,48 @@ pub fn local_config_path(project_dir: &Path) -> PathBuf {
...
+    };
+
+    let Ok(config) = load_config::<Config>(&global_config_path, Some(project_dir), None) else {
+        return;
+    };
+
+    for (key, value) in &config.dev().environment {
+        // SAFETY: see function-level docs.
+        unsafe { std::env::set_var(key, value) };
+    }
```
vados-cosmonic, line 677: Can we avoid this in favor of returning a set of the generated key/value pairs? I left a note at the usage point, but basically can we inject the keys/values into the command itself before we attempt to handle it?

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+        "secret source `file:` resolves inside the repo working tree and is not gitignored; add it to .gitignore"
+    );
+}
+
+/// Returns `Some(true)` if `git check-ignore` reports `path` as ignored from
+/// `repo_root`, `Some(false)` if it reports it as not ignored, and `None` if
+/// git is unavailable or the directory isn't a git repository. Wraps a
+/// blocking subprocess; only called in the secret-resolve warning path so
+/// the cost is negligible.
+fn is_gitignored(repo_root: &Path, path: &Path) -> Option<bool> {
```
vados-cosmonic: It's not clear what the caller shoudl do if they get a `None` -- Should this be a `Result` as in "this really shouldn't fail" ?

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+    Ok(out)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::config::EnvironmentLayer;
+    use tempfile::TempDir;
+
+    fn write(path: &Path, contents: &str) {
```
vados-cosmonic: not sure we need this IMO

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+        Some(1) => Some(false),
+        _ => None,
+    }
+}
+
+/// Minimal `.env` parser: `KEY=VALUE` per line, `#` line comments, blank lines
+/// ignored. No quoting, escapes, or variable expansion — keep this surface
+/// small and predictable.
+fn parse_env_file(path: &Path) -> Result<HashMap<String, String>> {
+    let contents = std::fs::read_to_string(path)
```
vados-cosmonic: We can use a buffered reader here IMO

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+            path.display(),
+            mode
+        );
+    }
+    Ok(())
+}
+
+#[cfg(not(unix))]
+fn check_secret_file_perms(_path: &Path) -> Result<()> {
+    // No portable equivalent on Windows; rely on the user's NTFS ACLs.
```
vados-cosmonic: Should we warn here that this is a no-op? Maybe at trace level or something if it would get super noisy?

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+#[cfg(unix)]
+fn check_secret_file_perms(path: &Path) -> Result<()> {
+    use std::os::unix::fs::PermissionsExt;
+
+    let meta = std::fs::metadata(path)
+        .with_context(|| format!("could not stat secret file: {}", path.display()))?;
+    let mode = meta.permissions().mode() & 0o777;
+    // Allow only owner-read or owner-read-write. Anything group/world readable
+    // is rejected — matches Kubernetes Secret volume defaultMode 0400.
+    if mode != 0o600 && mode != 0o400 {
```
vados-cosmonic: Is this [TOCTOU](https://doc.rust-lang.org/std/fs/index.html#time-of-check-to-time-of-use-toctou)-safe? This would be worse if it was async but I think there's a chance here. 

I think to make this safe you'd need to use the more generic open and keep it open while you check?

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+    }
+
+    Ok(canonical)
+}
+
+fn expand_tilde(path: &Path) -> PathBuf {
+    let Some(s) = path.to_str() else {
+        return path.to_path_buf();
+    };
+    let home = std::env::var_os("HOME").map(PathBuf::from);
```
vados-cosmonic: https://docs.rs/etcetera/0.11.0/etcetera/fn.home_dir.html could be preferable here since we have it pulled in

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+            );
+        };
+
+        let resolved = resolve_source(name, source, project_dir, secret, repo_root)?;
+        out.extend(resolved);
+    }
+    Ok(())
+}
+
+fn resolve_source(
```
vados-cosmonic: this feels like maybe something that should be on `ConfigSource`... But regardless a comment would be nice here.

I'm not a huge fan of `ConfigSource` and `SecretSource` being the same object here even if they're currently exactly the same

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+    )?;
+
+    Ok(ResolvedWorkload {
+        environment,
+        config: workload.config.clone(),
+        allowed_hosts: workload.allowed_hosts.clone(),
+    })
+}
+
+fn resolve_environment(
```
vados-cosmonic, line 101: Comment here would be nice on the goal -- IMO this should be on `WorkloadConfig` maybe

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+        /* secret */ true,
+        repo_root,
+        &mut out,
+    )
+    .context("failed to resolve workload.environment.secretFrom")?;
+
+    Ok(out)
+}
+
+fn apply_refs(
```
vados-cosmonic: Comment would be nice here

crates/wash/src/workload.rs
```diff
@@ -0,0 +1,661 @@
...
+    out: &mut HashMap<String, String>,
+) -> Result<()> {
+    for name in refs {
+        let Some(source) = catalog.get(name) else {
+            let kind = if secret { "secret" } else { "config" };
+            bail!(
+                "workload references {} '{}' which is not defined in the top-level `{}s:` block",
+                kind,
+                name,
+                kind
```
vados-cosmonic:
```suggestion
                "workload references {kind} '{name}' which is not defined in the top-level `{kind}s:` block",
```

crates/wash/src/cli/dev.rs
```diff
@@ -70,9 +76,13 @@ impl CliCommand for DevCommand {
...
 
-        // Enable wasi config
+        // Enable wasi config. `copy_environment = true` causes the plugin to
+        // surface every entry of `LocalResources.environment` via
+        // `wasi:config/store::get`, so values from `workload.environment` are
+        // visible to the component as both env vars and wasi:config entries
+        // without further plumbing.
         host_builder =
-            host_builder.with_plugin(Arc::new(plugin::wasi_config::DynamicConfig::default()))?;
+            host_builder.with_plugin(Arc::new(plugin::wasi_config::DynamicConfig::new(true)))?;
```
vados-cosmonic: NIT: DynamicConfig::new(true) is pretty unergonomic IMO... Builder here or if it's pub (and we're OK with it being pub, which we probably shouldn't be) we could do `DyanmicConfig { copy_environment: true, ..Default::default() }` ?

crates/wash/src/cli/dev.rs
```diff
@@ -269,173 +286,251 @@ impl CliCommand for DevCommand {
...
-/// - `config`: The overall Wash configuration
-/// - `bytes`: The bytes of the component under development
-async fn create_workload(host: &Host, config: &Config, bytes: Bytes) -> anyhow::Result<Workload> {
+/// Loaded inputs for [`build_workload`]. Bundles a component's raw bytes
+/// with its already-extracted WIT interface set so the pure assembly step
+/// has everything it needs without further I/O or host calls.
+struct LoadedComponent {
+    name: String,
+    bytes: Bytes,
+    interfaces: HashSet<WitInterface>,
```
vados-cosmonic, line 300: `BTreeSet`? for Ord?

crates/wash/src/cli/dev.rs
```diff
@@ -466,3 +561,456 @@ async fn reload_component(
...
+    }
+
+    /// Find a component by name in a built Workload.
+    fn find_component<'a>(workload: &'a Workload, name: &str) -> Option<&'a Component> {
+        workload.components.iter().find(|c| c.name == name)
+    }
+
+    // ---------------------------------------------------------------------
+    // Tests for `build_workload` used by `wash dev`
+    // ---------------------------------------------------------------------
```
vados-cosmonic: we can remove this probably

vados-cosmonic (commented):
Did a quick round of review -- I personally am OK with this landing and us iterating on it rather than being stuck in review too long! 

In particular I think we can iterate a bit if we land this under flag for early experimentation (but honestly maybe that's too much extra ceremony).

Also, I kind of like the idea of `workload::config::X` and `workload::secrets::Y` for interacting with different things related to workloads

crates/wash-runtime/src/plugin/wasi_otel/mod.rs
ricochet: Yeah but they were never plumbed through. Taking a look at a builder

crates/wash-runtime/src/plugin/wasi_otel/mod.rs
ricochet: We realistically only ever supported gRPC

crates/wash/src/config.rs
ricochet: Yeah including _sources does seem like a better name. I was copying Docker Compose directly as it has top-level `configs` and `secrets` and felt like a closer analog. I don't like that k8s doesn't have symmetry with ConfigMapSource and no source equivalent for secrets. Does that change your opinion on naming? I could go either way.

ack on `BTreeMap`. No memory on the serialization problem

crates/wash/src/config.rs
ricochet: I don't think so but I might be misunderstanding what you're suggesting. In this case we're not the ones reading the envs, it's an external library and I suspect that the list of config knobs used by external deps will always be non-exhaustive/unknown. I do think it's a best practice for us to be explicit about our own first-party plugin config knobs (aka for known OTEL knobs, put those in our otel plugin builder) rather than what's happening today with implicit handling by the third-party library setting config by reading envs.

crates/wash/src/workload.rs
ricochet: Yeah that is a foot-gun waiting to happen.

crates/wash-runtime/src/plugin/wasi_otel/mod.rs
ricochet: https://github.com/wasmCloud/wasmCloud/issues/5162

crates/wash/src/main.rs
ricochet: Certainly easy to do for host component plugins, but I couldn't find a way that would work for host plugin third party libs. Will continue convo below

crates/wash/src/workload.rs
ricochet: hardened, thanks for this

crates/wash/src/workload.rs
ricochet: Opting for a single warn since I'm afraid traces won't be visible to most users.

jfleitz (dismissed):

vados-cosmonic (dismissed):

crates/wash/src/config.rs
vados-cosmonic: Ah OK gotcha -- I'm not sure there isn't a way to pass the env more directly but if this is OTEL stuff I do remember they were quite unreasonable in how it was configured. We can hopefully revisit this in the future.

Issue #5116, opened 2026-05-04
CI builds wash-image more than once


ricochet:
Looking into potential CI speedups I noticed that we could extract a dedicated wash-image job that does a single-arch docker build → docker save → upload-artifact once. Both runtime-operator-e2e (download →  docker load → kind load docker-image) and canary-build consume it. canary-build would need to be reworked away from docker-build-push.yml though, since that workflow does its own buildx push-by-digest.

Cloud-Architect-Emma:
@ricochet, thanks for raising the issue, I'd like to help fix this issue. Can it be assigned to me ?

suyogkumawat:
Thanks @ricochet for raising the issue. I'd like to help fix this issue. Could it be assigned to me?

ricochet:
Thanks for the offers of help folks! @dutchaiagency made a pass at it already and I think it's a good start on it.

PR #5121, opened 2026-05-05
ci: reuse wash runtime image build

dutchaiagency:
﻿Refs #5116.

## Summary
- add a shared `wash-image` job that builds `localhost/wasmcloud-wash:e2e`, saves it, and uploads it once
- update `runtime-operator-e2e` to download/load that artifact before the existing kind/e2e path loads it into the cluster
- replace the `docker-build-push.yml` canary image path with a gated publish from the same saved image after check/lint/e2e pass

## Maintainer note
This follows the issue's single-arch artifact direction. One contract to confirm: `wash:sha-<full-sha>` and `wash:canary` now come from the same single-arch image exercised by e2e, instead of the previous multi-arch digest artifacts assembled by `docker-build-push.yml`. If the release path must keep a multi-arch wash image, I can revise this to preserve the arm64 digest path while still reusing the amd64 artifact for e2e.

## Validation
- `actionlint .github/workflows/wash.yml`
- `python -c "import yaml; yaml.safe_load(open('.github/workflows/wash.yml', encoding='utf-8'))"`
- `git diff --check`

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -29,6 +29,11 @@ on:
         description: "Build and Push Image. Default=false"
         required: false
         default: false
+      push-by-digest:
+        type: boolean
+        description: "Push image by digest to the registry (required for manifest assembly). Set false to build locally without pushing, e.g. on pull-request runs where the token lacks org-package write access. Default=true"
+        required: false
+        default: true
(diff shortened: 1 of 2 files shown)


ricochet:
We do want to preserve multi-arch

ricochet (approved):

PR #5125, opened 2026-05-06
feat: artifact precompile pipeline (MVP)

Mees-Molenaar:
## Summary

> **Status: feedback request, not merge-ready.** The MVP works end-to-end on kind, but some work is still open (see "Open work" below). However, before going on I would like a sanity check on the overall direction of this PR.     

As discussed in [#wasmcloud-precompile Slack thread](https://wasmcloud.slack.com/archives/CS38R7N9Y/p1776699871041139).

Adds an opt-in precompile pipeline so a wasmCloud host can skip compilation at workload start and load a precompiled `.cwasm` produced by a separate worker. The pipeline is disabled unless the operator is started with `--precompile-target` and `--precompile-wasmtime-version`.

## End-to-end flow

1. A new `PrecompileReconciler` spawns a Kubernetes `Job` per `Artifact`
2. The `Job` runs the `wash-precompile` binary. This binary pulls the OCI image, compiles it to `.cwasm` and writes it to a configured store (`nats://` for production, `file://` for dev/test).
3. When the `Job` is finished, the operator records the results in `Artifact.status.precompiled`
4. The `WorkloadDeploymentReconciler` waits for the precompilation to be complete and sends the `PrecompiledURL` via the workload proto to the wasmCloud Host.
5. The host loads the `.cwasm` if it was passed as `PrecompiledURL`. Otherwise the host will use the existing pull + compile path.

## Open work

* [x] Re-precompile on `Artifact.spec.image` change
* [ ] Cleanup unused `.cwasm`
* [ ] Persistent storage NATS
* [ ] NATS using TLS
* [ ] Wasmtime config matching between host and precompiler

Looking forward to your thoughts.
Mees


charts/runtime-operator/values.yaml
```diff
@@ -72,6 +72,28 @@ operator:
...
+      pull_policy: Always
+      # -- Defaults to the chart AppVersion when empty.
+      tag: ""
+    # -- Target triple precompiled bytes are produced for. Must match the
+    # runtime hosts' architecture/OS.
+    target: ""
+    # -- Wasmtime version the worker links against. Must match the host's
+    # wasmtime version exactly — cwasm bytes are version-specific.
+    # No default; required when precompile.enabled is true.
+    wasmtimeVersion: ""
```
thomas9911, line 92: why not infer this from the installed wasmtime version?

crates/wash-precompile/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+mod pull;
+
+#[derive(Parser, Debug)]
+#[command(name = "wash-precompile")]
+struct Args {
+    /// OCI reference to the source component
+    #[arg(long)]
+    image: String,
+
+    /// URL where the precompiled .cwasm bytes will be written
```
Aditya1404Sal, line 15:
```suggestion
    /// URL where the precompiled .cwasm bytes will be written
    ///
    /// Supported schemes:
    /// - file:///path/to/component.cwasm
    /// - nats://<bucket>/<key> (requires NATS_URL env var)
```

crates/wash-precompile/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[arg(long)]
+    output: String,
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+    let output_url = Url::parse(&args.output)?;
```
Aditya1404Sal, line 23:
```suggestion
    let output_url = Url::parse(&args.output)
    .with_context(|| format!("invalid output URL '{}'", args.output))?;
```
better to have a descriptive error

crates/wash-precompile/README.md
```diff
@@ -0,0 +1,62 @@
...
+## Run locally against a `file://` output
+
+```bash
+cargo run -p wash-precompile -- \
+    --image ghcr.io/wasmcloud/components/http-hello-world-rust:0.1.0 \
+    --output file:///tmp/out.cwasm
+```
+
+For `nats://`, set `NATS_URL`. For pulling from a private registry, set `DOCKER_CONFIG` pointing at a directory containing a `config.json` (the same format `kubectl create secret docker-registry` produces).
+
```
Aditya1404Sal, line 53: Could we add a short section clarifying the semantics of `nats://` outputs?

Right now it’s not obvious that:

- `NATS_URL` configures the NATS server connection
-  `nats://<bucket>/<key>` identifies the JetStream object store destination

An example mapping of bucket/key from the URL would probably help too, since the scheme is behaving more like an object-store locator (`s3://bucket/key` I think) than a direct connection URL.

charts/runtime-operator/values.local.yaml
```diff
@@ -17,6 +17,16 @@
...
 # See https://wasmcloud.com/docs/kubernetes-operator/operator-manual/helm-values
 # for the full gateway configuration reference.
+global:
+  tls:
+    enabled: false
+
+operator:
+  precompile:
+    enabled: true
+    wasmtimeVersion: "43"
```
Aditya1404Sal, line 27: probably should match the host's version `44`

crates/wash-runtime/tests/integration_fetch_precompiled.rs
```diff
@@ -0,0 +1,119 @@
...
+    let nats_url = format!("nats://127.0.0.1:{port}");
+
+    Ok((container, nats_url))
+}
+
+fn precompile_minimal_component() -> Result<Vec<u8>> {
+    let mut config = wasmtime::Config::new();
+    config.wasm_component_model(true);
+    let engine = wasmtime::Engine::new(&config)?;
+    let wasm = wat::parse_str("(component)")?;
```
Aditya1404Sal, line 86: I think we'd prefer using an actual component like the http-hello-world just so we can validate that the rest of the flow executes too.

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -505,6 +505,14 @@ impl Engine {
         }
     }
 
+    #[allow(unsafe_code)]
```
Aditya1404Sal, line 508: I know the `deserialize` has been pointed out in the slack threads, so it's more of a call back. 
perhaps an upstream modification to wasmtime? Signing/Attestation or a checksum of sorts

crates/wash-precompile/src/precompile.rs
```diff
@@ -0,0 +1,26 @@
...
+use wasmtime::{Config, Engine};
+
+pub fn compile(wasm_bytes: &[u8]) -> Result<Vec<u8>> {
+    let mut config = Config::new();
+    config.wasm_component_model(true);
+    let engine = Engine::new(&config)
+        .map_err(|e| anyhow::anyhow!("Error setting up wasmtime engine: {e}"))?;
+    let cwasm = engine
+        .precompile_component(wasm_bytes)
+        .map_err(|e| anyhow::anyhow!("Error precompiling wasm component: {e}"))?;
```
Aditya1404Sal, line 11: As a compatibility step I think It makes sense to store some additional metadata like a hash https://docs.wasmtime.dev/api/wasmtime/struct.Engine.html#method.precompile_compatibility_hash

crates/wash-runtime/src/washlet/mod.rs
```diff
@@ -373,40 +374,85 @@ async fn workload_start(
...
-                pool_size: component.pool_size,
-                max_invocations: component.max_invocations,
-            })
+            if !component.precompiled_url.is_empty() {
+                tracing::info!(
+                    component = %component.name,
+                    url = %component.precompiled_url,
+                    "using precompiled component"
+                );
+                let bytes = match fetch(&component.precompiled_url).await {
```
Aditya1404Sal, line 383: Since cwasm bytes are only valid for compatible engine configs, this fetch path should probably return both the bytes and attached metadata (at least a stored precompile_compatibility_hash). Then, before setting `is_precompiled: true`, we can compare that metadata against the current host engine and probably fall back to pulling/compiling the original wasm when it does not match

See comment in precompiler.rs for the precompile hash context.
ref: https://docs.wasmtime.dev/api/wasmtime/struct.Engine.html#method.precompile_compatibility_hash

Aditya1404Sal (changes requested):

PR #5127, opened 2026-05-06
feat(examples/otel-config): demo workload env/config/secrets

ricochet:
Builds on #5117 to make this
example a complete walkthrough of the new `.wash/config.yaml`
schema: an `app-defaults` config-source feeds non-sensitive runtime
knobs from `config/defaults.env`, an `upstream-credentials`
secret-source pulls `UPSTREAM_API_TOKEN` from the developer's shell
via `fromEnv`, `workload.environment.{configFrom,secretFrom}`
references both, and `workload.allowedHosts` constrains the outbound
call. `workload.config` continues to carry the `otel.resource.*`
identity attributes.

- New cached `AppConfig` reads `OUTBOUND_HOST`, `request_timeout_ms`,
  and `UPSTREAM_API_TOKEN` from `wasi:config/store::get_all` once on
  the first request. Defaults (`example.com`, 5s) apply when keys
  are absent.
- `make_outgoing_request` uses the configured URL, applies the
  configured timeout to first-byte and between-bytes, and attaches
  `Authorization: Bearer <token>` when the secret is present.
  Missing-token paths still work so the example is exercisable
  without setting `UPSTREAM_API_TOKEN`.
- New `log_runtime_config_once` emits a single OTel log line at
  startup listing every `wasi:config` key visible to the component.
  Keys only since secret values land in this same map under
  `copy_environment = true` and the schema's contract is that values
  are never logged.
- Outgoing HTTP span name is now low-cardinality `"outgoing http"`;
  the host moves to the `url.full` attribute.


examples/otel-config/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+        &trace_id,
+        &request_span_id,
+        flags,
+    );
+
+    export_metrics(count, response_len);
+    Ok(count.to_string())
+}
+
+fn export_metrics(count: u64, response_len: usize) {
```
vados-cosmonic, line 230: Can we move all the helpers to other files? It would be nice if `lib.rs` was really short and just the "handler" code.

examples/otel-config/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+            &format!("wasi:config keys: {}", keys.join(", ")),
+            trace_id,
+            span_id,
+            flags,
+        );
+    });
+}
+
+#[wstd::http_server]
+async fn main(_request: Request<Body>) -> Result<Response<Body>, wstd::http::Error> {
```
vados-cosmonic, line 110: Do we have to call this `main`? can we call it `handle_incoming_request` or `serve_request` or something like that?

vados-cosmonic (approved):
LGTM :rocket:

Let's get this in ASAP so we can have an OTEL example and then iterate from there.

vados-cosmonic:
One of the iterations I think we can do is adding our own integration for this into the `opentelemetry` Rust machinery, so we can use the usual `#![instrument]` and other things -- we just need a custom exporter IIRC. 

That said, I don't think that work is trivial/as easy as it sounds, so definitely happy to go from working version to working version and add it later

examples/otel-config/src/lib.rs
ricochet: Going to take this on in a follow-up PR while updating to rc2.

examples/otel-config/src/lib.rs
ricochet: Going to take this on in a follow-up PR while updating to rc2.

ricochet: ack learned something. The wstd::http_server proc macro hard-requires the function be named main:

```rust
  if name != "main" {
      return quote_spanned! { input.sig.ident.span()=>
          compile_error!("only `async fn main` can be used for #[wstd::http_server]");
      }
      ...
  }
```

vados-cosmonic: yeah figured that might be the case... thanks for checking

PR #5128, opened 2026-05-06
Add Host location and namespace deployment support

### LINKED ISSUE #5096: wasmCloud Kubernetes Namespace-scoped `Host` CRD
## Summary

Change the `Host` CRD from `scope: Cluster` to `scope: Namespaced`. With the default Helm values (`operator.allowSharedHosts: true`, no `WorkloadDeployment.Spec.HostNamespace` set), this is transparent — wasmCloud schedules and operates exactly as it does today. Multi-tenant operators get a single switch (`allowSharedHosts: false`) and an optional per-workload field (`Spec.HostNamespace`) to lock scheduling to a namespace.

## Motivation

Today the `Host` CRD lives at the cluster level even though every wasmCloud host already runs in a Kubernetes Pod inside a specific namespace. Aligning the CRD with the deployment topology unlocks three things:

1. **Native Kubernetes management of host groups.** Host groups are deployed as per-namespace Deployments, but their `Host` CRD objects sit at cluster scope. Namespacing them lets operators manage host pools the same way they manage every other workload — `kubectl get hosts -n team-a` returns just team-a's hosts; quota, NetworkPolicy, and audit scoping all apply naturally.
2. **Standard Kubernetes tenancy patterns for workloads.** Workloads can be scheduled to hosts within their own namespace, giving each team an isolated host pool. The existing host-group + host-selector pattern keeps working; namespacing makes the tenancy boundary explicit at the API level rather than relying purely on label conventions.
3. **No `ClusterRole` / `ClusterRoleBinding` requirement.** Many enterprise environments forbid or heavily restrict `ClusterRole`s. With a namespaced `Host` CRD, the user-facing `host-admin` / `host-editor` / `host-viewer` roles are generated as namespaced `Role`s, and a namespace admin can grant host visibility without cluster-admin involvement.

## Goals

- Ship `Host` unconditionally as `scope: Namespaced` — one shape, no Helm switch, no parallel CRDs.
- Preserve today's scheduling behavior with the default Helm values so existing installs continue to work after upgrade without manifest changes.
- Add `operator.allowSharedHosts` (default `true`) to give multi-tenant operators a single lever that locks scheduling to the workload's own namespace.
- Add `WorkloadDeployment.Spec.HostNamespace` (optional) so a workload can explicitly target a different namespace's host pool, gated by `allowSharedHosts`.
- Generate the user-facing host roles as namespaced `Role`s, eliminating the need for `ClusterRole`/`ClusterRoleBinding` to grant host visibility.

## Non-Goals

- **Multi-namespace targeting on a single `WorkloadDeployment`.** `Spec.HostNamespace` is a single namespace name. A label-based `HostNamespaceSelector` that matches many namespaces at once is intentionally out of scope and will only be considered if a concrete use case appears.

## Overview

Today every `Host` CRD lives at the cluster level. A wasmCloud host pod
runs in some namespace, but the `Host` object the operator creates from its NATS heartbeat has no namespace — it is named after `req.FriendlyName` and is visible to every tenant of the cluster. That model has two consequences worth changing:

1. **Tenant isolation.** A `WorkloadDeployment` in namespace `team-a` can be scheduled onto a host running for `team-b`. Tenant isolation today is accomplished by creating different host groups per tenant and ensuring the `hostSelector`s match accordingly — a label convention, not an API guarantee.
2. **Coarse RBAC.** Granting "see your team's hosts" requires a `ClusterRole` binding rather than a namespaced `Role`.

The proposal: make `Host` unconditionally `Namespaced`, and govern cross-namespace scheduling with a single operator-level toggle (`allowSharedHosts`) plus an optional per-workload override (`WorkloadDeployment.Spec.HostNamespace`).

### Behavior matrix

| `allowSharedHosts` | `WorkloadDeployment.Spec.HostNamespace` | Scheduling target |
|---|---|---|
| `true` (default) | unset | Any host that matches the `HostSelector` (cluster-wide — preserves today's behavior) |
| `true` | set, equals `workload.Namespace` | Hosts in `workload.Namespace` |
| `true` | set, different namespace | Hosts in `HostNamespace` |
| `false` | unset | Hosts in `workload.Namespace` |
| `false` | set, equals `workload.Namespace` | Hosts in `workload.Namespace` |
| `false` | set, different namespace | **Rejected** — workload not scheduled, `Schedulable=False` condition surfaces the violation |

The default of `true` + unset preserves the operational model of today's installs. Multi-tenant operators flip `allowSharedHosts: false` to enforce isolation; workloads that need to share an "infra" host pool opt in explicitly via `Spec.HostNamespace`.

### Operator

The operator currently uses one `ClusterRole` for everything ([`charts/runtime-operator/templates/operator/clusterrole.yaml`](charts/runtime-operator/templates/operator/clusterrole.yaml)). When `operator.watchNamespaces` is non-empty, hosts (following the same pattern already used for endpointslices) move out of the `ClusterRole` and into a per-namespace `Role`:

```yaml
# templates/operator/role.yaml — extend existing per-namespace Role
- apiGroups: [runtime.wasmcloud.dev]
  resources: [hosts, hosts/status, hosts/finalizers]
  verbs: [create, delete, get, list, patch, update, watch]
```

When `watchNamespaces` is empty (cluster-wide watch) the rule stays in the `ClusterRole`.

### User-facing roles

[`runtime-operator/config/rbac/runtime_host_*_role.yaml`](runtime-operator/config/rbac/) —`admin`, `editor`, `viewer` — are generated by kubebuilder from the `+kubebuilder:resource:scope` marker. Switching the marker to `Namespaced`
regenerates them as `Role` resources instead of `ClusterRole`. This is the desired end state: a namespace admin can grant `host-viewer` to a tenant without cluster-admin involvement, and clusters that forbid `ClusterRole`s can adopt the operator without exception.

### Host Pods

No change. Host pods do not read or write the `Host` CRD — only the operator does.

## Performance Impact

| Path | Cluster scope (today) | Namespaced scope (proposed) | Notes |
|---|---|---|---|
| `findFreeHost` (scheduling) | Lists all Hosts in cluster | Default (`allowSharedHosts=true`, no `HostNamespace`): cluster-wide list, **identical cost to today**. With `HostNamespace` set, or `allowSharedHosts=false`: list narrowed to one namespace. | Same cost as today in default mode; **faster** in the per-namespace modes — list size shrinks proportionally to the per-namespace fan-out. |
| `findHostByID` (routing) | Cluster-wide indexed `Get` | Cluster-wide indexed `Get` (HostIDs are globally unique UUIDs) | Indexed; O(1) either way. |
| `HostPodReconciler.deleteHostForIP` | Cluster-wide indexed list by `hostnameFieldIndex` | Namespace-scoped indexed list | Slightly faster; podIP collisions across namespaces no longer matter. |
| `HostReconciler.finalize` (delete-host fan-out) | Cluster-wide list of Workloads | **Unchanged** — workloads remain cluster-wide because HostID is the binding key | No regression. |
| `hostStatusUpdater` (heartbeat ingestion) | One Host CreateOrUpdate per heartbeat | Same, plus one cache-served Pod `Get` to resolve namespace | The Pod `Get` is cache-served by the operator's existing Pod informer (already populated for `HostPodReconciler`), so it does not hit the API server. |
| Operator memory footprint | One global Host informer cache | One Host cache per watched namespace, smaller in aggregate when only a subset is watched | Net reduction for watched-namespaces installs. |
| Heartbeat throughput | NATS-bound; not CRD-bound | Unchanged | Single round-trip CreateOrUpdate either way. |

**Headline:** the default mode costs exactly what today's Cluster-scope costs. Operators that opt into per-namespace scheduling (via `allowSharedHosts=false` or `Spec.HostNamespace`) get cheaper scheduling proportional to per-namespace host count. The only universal new cost is one cache-served Pod `Get` per heartbeat to resolve the host's namespace, which does not touch the API server and is dwarfed by the existing CreateOrUpdate round-trip.

jfleitz:
# Add Host environment + namespace deployment support

## Summary
Fixes #5096 

Lets the runtime-operator chart deploy host pods into tenant namespaces with
properly scoped RBAC, ServiceAccounts, and TLS certificates, and introduces a
new `environment` field on the host heartbeat that the operator records on
each `Host` CRD and the scheduler uses to enforce per-tenant placement of
WorkloadDeployments. Together these collapse the operator's blast radius from
a `ClusterRole` down to a namespaced `Role` and unblock multi-tenant
installs.

### Bug Fix
- Adds an Image Tag helper to the Helm chart to format the Chart's AppVersion to `vX.Y.Z` for the container images.

### 1. `Host` CRD moves to the operator's namespace — no more `ClusterRole` for host CRUD

The `Host` CRD is now `Namespaced` (was `Cluster`) and the operator
unconditionally creates every `Host` object in its own namespace, regardless
of where the underlying host pod runs.

- The operator's `Host` CRUD permissions move from the chart's `ClusterRole`
  ([clusterrole.yaml](charts/runtime-operator/templates/operator/clusterrole.yaml))
  into the namespaced `Role`
  ([role.yaml](charts/runtime-operator/templates/operator/role.yaml)). The
  operator can no longer write `Host` resources outside its own namespace.
- `WorkloadReconciler.findFreeHost`, `WorkloadRouteReconciler.findHostByID`,
  and `HostPodReconciler.deleteHostForIP` are all scoped to the operator
  namespace via `client.InNamespace(r.OperatorNamespace)`.
- The `Namespace` `get/list/watch` cluster-wide rule the earlier draft
  required (for validating heartbeat-claimed namespaces) is dropped.

### 2. Host pods can be deployed to separate namespaces from the chart

`runtime.hostGroups[].namespace` was already accepted by the chart but only
the host Deployment itself was rendered into the override namespace. The
chart now also renders, per unique tenant namespace:

- `wasmcloud-runtime-tls` and `wasmcloud-data-tls` Secrets so kubelet can
  mount them locally on host pods
  ([certificates.yaml](charts/runtime-operator/templates/certificates.yaml)).
- The runtime `ServiceAccount`
  ([runtime/serviceaccount.yaml](charts/runtime-operator/templates/runtime/serviceaccount.yaml)).
- A Pod-only `Role` + `RoleBinding` for the operator's ServiceAccount,
  rendered from
  [host-pod-role.yaml](charts/runtime-operator/templates/operator/host-pod-role.yaml),
  granting `pods get/list/patch/watch` so `HostPodReconciler` can manage
  finalizers on host pods that live outside the operator's own namespace.

A new `runtime-operator.hostNamespaces` Helm helper unions
`operator.hostNamespaces` with every distinct
`runtime.hostGroups[].namespace`, so the per-namespace RBAC and the
operator's `-host-namespaces` flag are derived automatically — no need to
keep the two values in sync by hand.

The CA + every signed cert is also generated **once** (via `lookup` +
`buildCustomCert` reuse on upgrade, fresh `genCA` on first install) and
the same key material is replicated into each host-pod namespace, so a
host pod in `team-a` and a host pod in `team-b` trust the same CA.

### 3. New `environment` field on the host heartbeat

`HostHeartbeat.environment` (proto field 18) records where the host
reports itself running. In-cluster host pods populate it from the downward
API (`metadata.namespace` → `WASMCLOUD_HOST_ENVIRONMENT` →
`--environment` flag); out-of-cluster hosts pass `--environment` to
`wash host` directly.

- Proto:
  [proto/wasmcloud/runtime/v2/host_heartbeat.proto](proto/wasmcloud/runtime/v2/host_heartbeat.proto).
- wash-runtime: `HostHeartbeat.environment`, `HostBuilder::with_environment`,
  `Host::environment()` accessor; populated in the heartbeat construction
  and mapped through the proto `From` impl.
- wash CLI: `--environment <ENVIRONMENT>` (also picks up
  `WASMCLOUD_HOST_ENVIRONMENT`).
- Operator: stored verbatim on `Host.spec.environment` — no validation
  against cluster state, which means external hosts and in-cluster hosts
  travel the same code path.

The Helm chart wires the env var from the downward API on every host
Deployment, so existing in-cluster installs get the right value with no
user-facing config change.

### 4. WorkloadDeployments can target a specific environment

`WorkloadDeployment.Spec.template.spec.environment` is matched against
`Host.spec.environment` at scheduling time, alongside the existing
`hostSelector` label match. Behavior is governed by
`operator.allowSharedHosts`:

| `allowSharedHosts` | `Spec.environment` | Scheduling target |
|---|---|---|
| `true` (default) | unset | Any matching host (legacy behavior) |
| `true` | set, equals workload namespace | Hosts whose `Environment` matches that namespace |
| `true` | set, different value | Hosts whose `Environment` matches that value (cross-tenant allowed) |
| `false` | unset | Hosts whose `Environment` matches the workload's own namespace |
| `false` | set, equals workload namespace | Hosts whose `Environment` matches |
| `false` | set, different value | **Rejected** with `CrossEnvironmentSchedulingDenied` warning Event + `Schedulable=False` condition |

A new `hostEnvironmentIndex` field index (`spec.environment`) on `Host`
keeps the filtered list at O(matching) rather than O(all hosts).

## Testing

A local Kind cluster was used for both scenarios. The values files and
test workloads below are not checked in; reproduce the test by saving the
following YAML and running the commands in each scenario.

### Test fixture — `values.cluster.yaml`

Single `default` hostGroup deployed into the chart's release namespace
(3 replicas, all sharing the operator's namespace).

```yaml
gateway:
  enabled: false

operator:
  image:
    tag: "local"
    repository: "runtime-operator"
    registry: ""
    pull_policy: Never

runtime:
  image:
    registry: ""
    repository: wash
    pull_policy: Never
    tag: "local"
  hostGroups:
    - name: default
      replicas: 3
      service:
        type: ClusterIP
      webgpu:
        enabled: false
      http:
        enabled: true
        port: 80
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
          memory: "512Mi"
          cpu: "500m"
```

### Test fixture — `values.namespace.yaml`

Two `default` hostGroups split across `namespace-a` and `namespace-b`,
plus a `default-gpu` hostGroup in the release namespace. Exercises the
multi-namespace RBAC + per-namespace TLS Secret + per-namespace
ServiceAccount rendering.

```yaml
gateway:
  enabled: false

operator:
  image:
    tag: "local"
    repository: "runtime-operator"
    registry: ""
    pull_policy: Never

runtime:
  image:
    registry: ""
    repository: wash
    pull_policy: Never
    tag: "local"
  hostGroups:
    - name: default
      namespace: "namespace-a"
      replicas: 1
      service:
        type: ClusterIP
      webgpu:
        enabled: false
      http:
        enabled: true
        port: 80
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
          memory: "512Mi"
          cpu: "500m"
    - name: default
      namespace: "namespace-b"
      replicas: 1
      service:
        type: ClusterIP
      webgpu:
        enabled: false
      http:
        enabled: true
        port: 80
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
          memory: "512Mi"
          cpu: "500m"
    - name: default-gpu
      replicas: 1
      service:
        type: ClusterIP
      webgpu:
        enabled: true
      http:
        enabled: true
        port: 80
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
          memory: "512Mi"
          cpu: "500m"
```

### Test fixture — `testworkload.yaml`

Four WorkloadDeployments covering every branch of the
`allowSharedHosts × Spec.environment` scheduling matrix: no environment
set, same-namespace environment, and two cross-namespace combinations.
Used unchanged in both scenarios.

```yaml
kind: WorkloadDeployment
metadata:
  name: no-environment-world
  namespace: namespace-a
spec:
  replicas: 1
  template:
    spec:
      hostSelector:
        hostgroup: default
      components:
        - name: hello-world
          image: ghcr.io/wasmcloud/components/hello-world:0.1.0
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: localhost
---
kind: WorkloadDeployment
metadata:
  name: environment-world
  namespace: namespace-a
spec:
  replicas: 1
  template:
    spec:
      environment: namespace-a
      hostSelector:
        hostgroup: default
      components:
        - name: hello-world
          image: ghcr.io/wasmcloud/components/hello-world:0.1.0
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: localhost
---
kind: WorkloadDeployment
metadata:
  name: cross-environment-world
  namespace: namespace-b
spec:
  replicas: 1
  template:
    spec:
      environment: namespace-a
      hostSelector:
        hostgroup: default
      components:
        - name: hello-world
          image: ghcr.io/wasmcloud/components/hello-world:0.1.0
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: localhost
---
kind: WorkloadDeployment
metadata:
  name: cross-environment-world
  namespace: namespace-a
spec:
  replicas: 1
  template:
    spec:
      environment: namespace-b
      hostSelector:
        hostgroup: default
      components:
        - name: hello-world
          image: ghcr.io/wasmcloud/components/hello-world:0.1.0
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: localhost
```

### Scenario A — host pods in the chart's release namespace

```sh
helm upgrade --install -f values.cluster.yaml wasmcloud ./charts/runtime-operator
kubectl apply -f testworkload.yaml
kubectl get workloads -A -o wide
```

Using `values.cluster.yaml` above (single `default` hostGroup, 3 replicas,
no `namespace` override) and the shared `testworkload.yaml`. The
`ENVIRONMENT` column on each `Workload` reflected the host's actual
environment; cross-tenant deployments scheduled successfully because
`allowSharedHosts` defaults to `true`.

### Scenario B — host pods in two separate tenant namespaces

```sh
helm upgrade --install -f values.namespace.yaml wasmcloud ./charts/runtime-operator
kubectl apply -f testworkload.yaml
kubectl get workloads -A -o wide
```

Using `values.namespace.yaml` above. The chart rendered:

- `wasmcloud-runtime-tls` and `wasmcloud-data-tls` in both
  `namespace-a` and `namespace-b` (and the release namespace), all signed
  by the same CA.
- The runtime `ServiceAccount` in each tenant namespace.
- `wasmcloud-runtime-operator-host-pod` Role + RoleBinding in each tenant
  namespace.
- The operator was started with `-host-namespaces=namespace-a,namespace-b`,
  derived automatically from the helper.

The same `testworkload.yaml` covered the four scheduling cases. The
`ENVIRONMENT` column on each `Workload` correctly showed `namespace-a` or
`namespace-b` based on which host the scheduler picked, and
cross-environment deployments routed onto the matching tenant's hosts.

## Test plan

- [x] `helm upgrade --install` cleanly on Kind with
      `testing/cluster/values.cluster.yaml`; host pods reach `Ready`,
      `kubectl get workloads -A -o wide` shows `ENVIRONMENT` populated.
- [x] `helm upgrade --install` cleanly on Kind with
      `testing/host\ namespaces/values.namespace.yaml`; host pods in both
      `namespace-a` and `namespace-b` reach `Ready`; per-namespace TLS
      Secrets, ServiceAccounts, and Pod RBAC render in each tenant
      namespace.
- [x] `kubectl apply -f testing/host\ namespaces/testworkload.yaml`
      schedules each of the four WorkloadDeployment cases correctly:
      no-environment, same-namespace-environment, and two cross-namespace
      cases (with `allowSharedHosts=true` permitting the cross-tenant ones).
- [x] Re-run `helm upgrade` against an existing release; TLS Secrets are
      reused (operator/host pods do not need to roll) and no
      `Error: UPGRADE FAILED: ...buildCustomCert` is emitted.
- [x] `go test ./...` and `cargo test -p wash-runtime --lib` green.
- [x] `wash host --help` shows `--environment` and accepts
      `WASMCLOUD_HOST_ENVIRONMENT`.


jfleitz:
FYI, the runtime-operator/test-e2e step is going to fail until the host:canary image supports passing the `environment` flag (This PR adds that). Since the wash / runtime-operator-e2e (pull_request) uses a local build of the wash runtime, it does pass, and therefore confirms that the runtime-operator/test-e2e step will pass once merged to main.

runtime-operator/internal/controller/runtime/workload_controller.go
```diff
@@ -90,32 +115,94 @@ func (r *WorkloadReconciler) reconcileHostSelection(ctx context.Context, workloa
...
 	return condition.ErrSkipReconciliation()
 }
 
-func (r *WorkloadReconciler) findFreeHost(ctx context.Context, workload *runtimev1alpha1.Workload) (string, error) {
+func (r *WorkloadReconciler) findFreeHost(ctx context.Context, workload *runtimev1alpha1.Workload) (*runtimev1alpha1.Host, error) {
 	hostList := runtimev1alpha1.HostList{}
 
-	if err := r.List(ctx, &hostList,
+	// Every Host object lives in the operator's namespace. Tenant
+	// isolation is enforced by an indexed match on Host.Location, not by
```
ricochet: Environment

ricochet:
> FYI, the runtime-operator/test-e2e step is going to fail until the host:canary image supports passing the `environment` flag (This PR adds that). Since the wash / runtime-operator-e2e (pull_request) uses a local build of the wash runtime, it does pass, and therefore confirms that the runtime-operator/test-e2e step will pass once merged to main.

Added the canary label so that we push canaries from this change and allow testing in CI.

ricochet (approved):

PR #5139, opened 2026-05-11
ci: add status gate pattern to all path-filtered workflows

ricochet:
## Summary

Required status checks that use trigger-level paths: filters stay Pending (never fire) when a PR touches unrelated files, forcing maintainers to use bypass. This PR implements the status gate pattern across all 8 CI workflows to fix that.

## What changed

Every workflow that had a paths: filter on its pull_request trigger now uses:

1. changes job (if: github.event_name == 'pull_request'): runs dorny/paths-filter via the GitHub API to detect whether relevant files changed. Skipped entirely on push/tag/dispatch events.
2. Work jobs (if: always() && (needs.changes.result == 'skipped' || needs.changes.outputs.relevant == 'true')): run unconditionally on push/tag/dispatch; on PRs, run only when changes reports relevant files touched.
3. gate job (if: always()): the single required check. Always reports a definitive pass or fail, never stays Pending.

### Why dorny/paths-filter instead of git diff

I started with a custom bash approach which required fetch-depth: 0 to reliably find the merge base for three-dot diffs. dorny/paths-filter uses the GitHub Checks API directly for PRsso it has a cleaner glob syntax matching the original paths: format, and battle-tested across thousands of repos.

### References

- dorny/paths-filter: https://github.com/dorny/paths-filter
- GitHub docs — skipped vs. pending required checks:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -69,7 +69,6 @@ jobs:
       - cargo-audit
     if: always()
     runs-on: ubuntu-latest
-    permissions: {}
     steps:
       - name: Require all jobs to pass or be skipped
         env:
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -4,10 +4,6 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-      - "**/Cargo.toml"
-      - "**/Cargo.lock"
-      - ".github/workflows/audit.yml"
   push:
     branches:
       - main
@@ -24,13 +20,38 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
+  changes:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      relevant: ${{ steps.filter.outputs.relevant }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+        id: filter
+        with:
+          filters: |
+            relevant:
+              - '**/Cargo.toml'
+              - '**/Cargo.lock'
+              - '.github/workflows/audit.yml'
+
   cargo-audit:
+    needs:
+      - changes
+    if: always() && (needs.changes.result == 'skipped' || needs.changes.outputs.relevant == 'true')
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    permissions:
+      contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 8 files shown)


.github/workflows/audit.yml
```diff
@@ -42,3 +61,17 @@ jobs:
           tool: cargo-audit
       - name: Run cargo audit
         run: cargo audit
+
+  gate:
+    needs: [cargo-audit]
+    if: always()
+    runs-on: ubuntu-latest
+    permissions: {}
```
vados-cosmonic: we could invert this and have `permissions: {}` and the top, and then we would be secure by default

(could absolutely break other steps that aren't explicit though!)

.github/workflows/proto.yml
```diff
@@ -49,3 +70,17 @@ jobs:
         uses: ./.github/actions/no-diff
         with:
           fixup-command: "buf generate"
+
+  gate:
+    needs: [lint, check-generated]
```
vados-cosmonic: NIT: would like to make these into YAML lists like the below:

```yaml
needs:
  - lint
  - check-generated
```

Going to make for better diffs later, IMO

.github/workflows/proto.yml
```diff
@@ -20,7 +17,29 @@ permissions:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+        id: filter
+        with:
+          filters: |
+            relevant:
+              - 'proto/**'
+              - '.github/workflows/proto.yml'
```
vados-cosmonic, line 41: Should we include `buf.yaml` and friends here? Is the workflow that we can update the protos independently of all the individual codebases (that would be fine too -- if someone changes proto/buf settings and always makes sure to commit the updated proto). Feels like it's possible to update buf settings and not re-generate proto?

Maybe this isn't an issue in practice though.

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but NITs

.github/workflows/proto.yml
ricochet: Added!

.github/workflows/proto.yml
ricochet: Agreed, worth it for the diffs

.github/workflows/audit.yml
ricochet: We actually follow the empty `permission: {}` in most other workflows. Updated to do this here as well.

.github/workflows/audit.yml
```diff
@@ -42,3 +63,18 @@ jobs:
           tool: cargo-audit
       - name: Run cargo audit
         run: cargo audit
+
+  gate:
+    needs:
+      - cargo-audit
+    if: always()
+    runs-on: ubuntu-latest
+    permissions: {}
```
vados-cosmonic:
```suggestion
```

.github/workflows/charts.yml
```diff
@@ -27,9 +46,25 @@ jobs:
       - name: Render
         run: make helm-render
 
+  gate:
+    needs:
+      - test
+    if: always()
+    runs-on: ubuntu-latest
+    permissions: {}
```
vados-cosmonic:
```suggestion
```

.github/workflows/lint-workflows.yml
```diff
@@ -54,3 +77,19 @@ jobs:
...
           # referenced action's tag still resolves to its pinned SHA).
           GH_TOKEN: ${{ github.token }}
+
+  gate:
+    needs:
+      - actionlint
+      - zizmor
+    if: always()
+    runs-on: ubuntu-latest
+    permissions: {}
```
vados-cosmonic:
```suggestion
```

.github/workflows/proto.yml
```diff
@@ -49,3 +76,19 @@ jobs:
...
         with:
           fixup-command: "buf generate"
+
+  gate:
+    needs:
+      - lint
+      - check-generated
+    if: always()
+    runs-on: ubuntu-latest
+    permissions: {}
```
vados-cosmonic: this spot should probably be updated

vados-cosmonic (approved):

Issue #4962, opened 2025-08-13
Add cleanup subcommand to config command


brooksmtownsend:
## Description

The `wash config` command should include a `cleanup` subcommand to help users clean up wash-related directories and cached data when needed.

## Location

`crates/wash/src/cli/config.rs:25` in the `ConfigCommand` enum

## Current Behavior

```rust
#[derive(Debug, Parser)]
pub enum ConfigCommand {
    // ... existing subcommands  
    // TODO(GFI): cleanup config command, to clean the dirs we use
}
```

Users currently have no easy way to clean up wash's data/cache directories.

## Proposed Solution

Add a `cleanup` subcommand:

```rust
#[derive(Debug, Parser)]
pub enum ConfigCommand {
    // ... existing subcommands
    /// Clean up wash directories and cached data
    Cleanup {
        /// Remove cache directory
        #[clap(long)]
        cache: bool,
        /// Remove data directory  
        #[clap(long)]
        data: bool,
        /// Remove all wash directories (cache + data)
        #[clap(long)]
        all: bool,
        /// Show what would be removed without actually deleting
        #[clap(long)]
        dry_run: bool,
    },
}
```

## Cleanup Targets

The cleanup command should be able to remove:

1. **Cache directory**: Temporary files, downloads, build caches
2. **Data directory**: Plugin data, component artifacts
3. **Generated configs**: Remove generated configuration files
4. **Logs**: Old log files if any are stored
5. **Temporary files**: Any temporary wash-related files

## Example Usage

```bash
# Clean only cache
wash config cleanup --cache

# Clean data directory
wash config cleanup --data

# Clean everything
wash config cleanup --all

# Preview what would be cleaned
wash config cleanup --all --dry-run
```

## Safety Features

- **Confirmation prompts**: Ask for confirmation before removing important data
- **Dry run mode**: Show what would be removed without actually doing it
- **Selective cleanup**: Allow users to choose what to clean
- **Backup suggestions**: Warn about data that might be important to backup

## Benefits

- **Disk space recovery**: Easy way to reclaim space used by wash
- **Troubleshooting**: Reset wash state when debugging issues
- **Clean slate**: Start fresh with wash configuration and data

vados-cosmonic:
The project has changed so drastically that this idea is probably not reasonable to pursue in this form -- closing this as we're focusing effort now on wasmCloud v2.

If you think this definitely still applies towards v2, please feel free to open a new issue!

PR #5140, opened 2026-05-11
 feat(wash): extend config command with init formats, cleanup, and validation

### LINKED ISSUE #27: Add validate subcommand to config command
## Description

The `wash config` command should include a `validate` subcommand to help users verify their configuration files are correctly formatted and contain valid values.

## Location

`crates/wash/src/cli/config.rs:24` in the `ConfigCommand` enum

## Current Behavior

```rust
#[derive(Debug, Parser)]
pub enum ConfigCommand {
    // ... existing subcommands
    // TODO(GFI): validate config command
}
```

The config command currently only supports `init`, `info`, and `show` subcommands.

## Proposed Solution

Add a `validate` subcommand:

```rust
#[derive(Debug, Parser)]
pub enum ConfigCommand {
    // ... existing subcommands
    /// Validate the current configuration
    Validate {
        /// Path to specific config file to validate (optional)
        #[clap(long)]
        file: Option<PathBuf>,
        /// Validate project config instead of global config
        #[clap(long)]
        project: bool,
    },
}
```

## Validation Features

The validate command should check:

1. **JSON syntax**: Ensure the config file is valid JSON
2. **Schema validation**: Verify all fields match expected config structure
3. **Value validation**: Check that URLs, paths, and other values are valid
4. **Template validation**: Verify template URLs are accessible
5. **Dependency conflicts**: Check for conflicting settings

## Example Usage

```bash
# Validate global config
wash config validate

# Validate specific config file
wash config validate --file .wash/config.json

# Validate project config
wash config validate --project
```

## Benefits

- **Error prevention**: Catch config issues before they cause runtime problems
- **Better UX**: Clear feedback about config problems and suggestions for fixes
- **Development aid**: Helpful for config file authors and CI/CD pipelines

if0ne:
- `wash config init` now accepts `--format yaml|yml|json|toml` (default: `yaml`) to write the config in the chosen serialization format. `--example` populates every section with illustrative values. `--global` is removed; the command always writes to the project-local `.wash/` directory.
- New `wash config cleanup` removes the wash cache and/or data directories (`--cache`, `--data`, `--all`). `--dry-run` prints what would be removed with per-directory sizes (via `fs_extra`) formatted with `humansize`.
- New `wash config validate` checks the loaded configuration for hard errors: socket-address syntax, URL scheme correctness (redis/nats/postgres), TLS cert+key pairing, WebGPU on Windows, empty build command, malformed WIT registry URLs, and local WIT source paths that don't exist on disk. Remote sources (HTTP, Git, OCI) are skipped. Errors from all sections are collected before returning so the user sees every issue at once. Pass `--file` to validate a specific file instead of the merged project config.

Closes https://github.com/wasmCloud/wash/issues/27 https://github.com/wasmCloud/wasmCloud/issues/4962

I wanted to add example config generation, saw TODOs for the validation and cleanup commands, and decided to add those too


ricochet (commented):
Design-wise LGTM. Note this is touching some of the same areas as #5117 but I don't see any conflicts.

I wish we removed `--global` before cutting v2. Since it is a breaking change, could/should we keep the flag as a hidden alias that prints a deprecation message?

crates/wash/src/config.rs
```diff
@@ -333,19 +454,115 @@ pub fn local_config_path(project_dir: &Path) -> PathBuf {
...
+            postgres_url: Some("postgres://user:pass@127.0.0.1:5432".to_string()),
+            ..Default::default()
+        }),
+        new: None,
+        wit: Some(WitConfig {
+            registries: vec![],
+            skip_fetch: false,
+            wit_dir: Some(PathBuf::from("wit")),
+            sources: HashMap::from_iter([
+                ("example:local".to_string(), "./local/wit".to_string()),
```
ricochet: Running `wash config init --example && wash config validate` will fail with the local file path: `./local/wit`

crates/wash/src/config.rs
ricochet: this func is dead code now

Cargo.toml
```diff
@@ -58,6 +58,8 @@ etcetera = { version = "0.10.0", default-features = false }
 figment = { version = "0.10.19", default-features = false }
 futures = { version = "0.3", default-features = false }
 flate2 = { version = "1.0", default-features = false }
+fs_extra = { version = "1.3" }
```
ricochet: This isn't actively maintained: https://github.com/webdesus/fs_extra

crates/wash/src/config.rs
```diff
@@ -333,19 +454,115 @@ pub fn local_config_path(project_dir: &Path) -> PathBuf {
...
+    match url::Url::parse(value) {
+        Ok(u) if expected.contains(&u.scheme()) => {}
+        Ok(u) => errors.push(format!(
+            "{field} '{value}' has scheme '{}', expected one of: {}",
+            u.scheme(),
+            expected.join(", ")
+        )),
+        Err(e) => errors.push(format!("{field} '{value}' is not a valid URL: {e}")),
+    }
+}
```
ricochet, line 562: BuildConfig::validate, DevConfig::validate, and WitConfig::validate are unit-testable without any async setup. 

Coverage for the error cases would be most helpful: empty build command, TLS cert without key, invalid socket address, and bad URL scheme.

crates/wash/src/cli/config.rs
```diff
@@ -45,68 +98,253 @@ impl CliCommand for ConfigCommand {
...
+
+/// Sum the byte size of all files under `path` using `fs_extra::dir::get_size`.
+///
+/// Runs the synchronous walk on a blocking task so the async runtime is not stalled.
+async fn dir_size(path: &Path) -> anyhow::Result<u64> {
+    let owned = path.to_path_buf();
+    tokio::task::spawn_blocking(move || fs_extra::dir::get_size(&owned))
+        .await
+        .context("directory size task panicked")?
+        .with_context(|| format!("failed to compute size for {}", path.display()))
```
ricochet: To drop fs_extra, I think this is all that's needed: 

```suggestion
/// Sum the byte size of all files under `path`.
///
/// Runs the synchronous walk on a blocking task so the async runtime is not stalled.
async fn dir_size(path: &Path) -> anyhow::Result<u64> {
    let owned = path.to_path_buf();
    tokio::task::spawn_blocking(move || {
        let mut total = 0u64;
        let mut stack = vec![owned];
        while let Some(dir) = stack.pop() {
            for entry in std::fs::read_dir(&dir)
                .with_context(|| format!("failed to read directory {}", dir.display()))?
            {
                let entry = entry
                    .with_context(|| format!("failed to read entry in {}", dir.display()))?;
                let meta = entry
                    .metadata()
                    .with_context(|| format!("failed to stat {}", entry.path().display()))?;
                if meta.is_dir() {
                    stack.push(entry.path());
                } else {
                    total = total.saturating_add(meta.len());
                }
            }
        }
        Ok(total)
    })
    .await
    .context("directory size task panicked")?
```

ricochet (changes requested):

crates/wash/src/config.rs
ricochet: Maybe for simplicity, just remove the local path as an example

crates/wash/src/cli/config.rs
```diff
@@ -45,68 +102,284 @@ impl CliCommand for ConfigCommand {
...
+                    "source": source_label,
+                    "errors": e.to_string().lines().collect::<Vec<_>>(),
+                    "valid": false,
+                })),
+            ))
+        }
+    }
+}
+
+/// Sum the byte size of all files under `path` using `fs_extra::dir::get_size`.
```
ricochet:
```suggestion
/// Sum the byte size of all files under `path`.
```

ricochet (dismissed):
Just the doc comment nit

ricochet (approved):

PR #5143, opened 2026-05-11
ci(examples): lint, build, and publish Rust examples

### LINKED ISSUE #4937: [CI] publish OCI artifacts of examples
Our examples should be published to OCI so that they can be used in docs/examples without requiring building from source.

ricochet:
## Summary

**New `examples` workflow** that lints, builds, and OCI-publishes every Rust example. Per-example filters mean a PR touching one example only rebuilds that example.

Examples are not tests and are not meant to be used for test coverage. We do not want examples to build on every PR. Right now we only smoke test that the examples build and pass code style.

This repo will only ever contain Rust component examples. For other languages, they will be in a language specific repo.

## Changes

- Two new composite actions: `setup-rust-example` (nightly + stable Rust + wash, called once per job) and `build-rust-example` (per-`.wash/config.yaml`: `wash wit fetch`, `cargo +nightly fmt --check`, `cargo clippy --locked -- -D warnings`,
  `wash build`, and on main pushes `wash oci push` to `ghcr.io/wasmcloud/components/<name>:<Cargo-package-version>`. This is skipped if that tag already exists, so bumping the Cargo version in a PR is the explicit release
  signal).
- **Coverage job** in the examples workflow enforces that every directory with both `.wash/config.yaml` and `Cargo.toml` has a CI job and a `README.md`.
- **Cleanup**: removed `examples/go-hello-world` (had no CI, doesn't belong in Rust-only examples), added `examples/README.md` as an index pointing to wasmcloud.com/docs/examples for other languages, added
`--locked` to every example's `.wash/config.yaml` build command, fixed clippy warnings in blobby and otel-http to pass `-D warnings`.
- **Dependabot**: the four example Cargo workspaces join the existing `all-cargo` group; both new composite actions join `all-github-actions`

Follows the project's status-gate pattern: pull_request fires
unconditionally, dorny/paths-filter detects per-example changes inside
the workflow, and a single `gate` job is the required check. This is why there are individual examples jobs rather than a matrix. For PRs where examples were not edited but we want to run the examples CI, add label `examples-canary`.

## Test plan

- [ ] PR touching unrelated files (docs only): example jobs are skipped, `gate` is green — no Pending checks block merge
- [ ] PR touching `examples/blobby/`: only the `blobby` job runs; other example jobs are skipped; coverage + gate pass
- [ ] PR touching `.github/actions/build-rust-example/` or `setup-rust-example/`: all four example jobs rerun (filter is wired)
- [ ] PR adds `examples/foo/.wash/config.yaml` without a corresponding job: coverage check fails with the missing path printed
- [ ] PR removes a `README.md` from an existing example: coverage check fails with the missing path printed
- [ ] Merge to main with the same Cargo version as the last release: publish step logs `<image>:<version> already exists in registry, skipping push` and exits 0
- [ ] Bump `version` in `examples/blobby/Cargo.toml` and merge: `ghcr.io/wasmcloud/components/blobby:<new-version>` appears on ghcr
- [ ] Local: `cargo +nightly fmt --check` and `cargo clippy --target wasm32-wasip2 --locked -- -D warnings` pass in every example workspace, and `wash build` succeeds

Fixes #4937


.github/actions/setup-rust-example/action.yml
```diff
@@ -0,0 +1,18 @@
...
+  example: nightly Rust with rustfmt (for `cargo +nightly fmt --check`),
+  stable Rust with clippy + rustfmt + wasm32-wasip2 (for clippy and the
+  build), and the wash CLI (for `wash wit fetch`, `wash build`, and
+  `wash oci push`). Designed to be called once per job; subsequent
+  build-rust-example invocations in the same job reuse this setup.
+
+runs:
+  using: composite
+  steps:
+    - uses: dtolnay/rust-toolchain@3c5f7ea28cd621ae0bf5283f0e981fb97b8a7af9 # master
```
vados-cosmonic, line 13: ~~pinned version here?~~

I guess there's only the current `@stable` ? I think dependabot should be able to handle tracking the moving branch ref as well.

.github/actions/build-rust-example/action.yml
```diff
@@ -0,0 +1,94 @@
...
+  working-directory:
+    description: "Directory containing the example's .wash/config.yaml."
+    required: true
+  image:
+    description: >
+      OCI image to push to, without a tag (e.g. ghcr.io/wasmcloud/components/blobby).
+      On pushes to main, publishes a single tag whose value is the Cargo
+      package version. Skips the push if that tag already exists in the
+      registry, so bump-the-version is the explicit signal to release.
+    required: true
```
vados-cosmonic, line 22: Feels like we could make this *not* required, and avoid doing if not present? Is that worth doing to make this more reusable?

(probably not)

examples/blobby/.wash/config.yaml
```diff
@@ -1,7 +1,7 @@
 version: 2.0.0
 
 build:
-  command: cargo build --target wasm32-wasip2 --release
+  command: cargo build --target wasm32-wasip2 --release --locked
```
vados-cosmonic, line 4: Not using `--locked` should help us catch non-semver breaking changes -- did this break?

vados-cosmonic (approved):
LGTM 🚀

examples/blobby/.wash/config.yaml
ricochet: Didn't break but was flagged as not being reproducible for something we will vet in CI

vados-cosmonic: yeah I think that depends on what we're trying to get out of CI runs... I could go either way here -- the fact is that if we update the template, then actually blobby will always get built via `--locked`, so this problem sort of solves itself. 

But if we live in a world where people are writing their own build commands they almost certainly don't use `--locked` by default (IMO this happens only when something goes wrong) then we get some small value out of making sure the most obvious build command works. 

That said, this is probably fine -- we have really cut down on our use of custom crates that we maintain (and could potentially break) so probably fine.

.github/actions/setup-rust-example/action.yml
ricochet: Yeah we have to use the sha pin or our action linters will flag. I'll replace this with shell and remove the dep

.github/actions/build-rust-example/action.yml
ricochet: I like the semantic of forcing callers to name their package which I think in turn makes for better examples.

PR #5149, opened 2026-05-13
ci(wit): validate and publish wasmCloud WIT interfaces

ricochet:
Builds each wit/<pkg>/ with `wash wit build`, embeds provenance via `wasm-tools metadata add`, and on push to main publishes to ghcr.io/wasmcloud/interfaces/<name>:<version> with gh attestation. Stable versions are immutable. Re-pushing an existing tag is a no-op.

I took the existing interfaces wasmcloud:messaging and wasmcloud:secrets and added them to the project root.

A new workflow opens weekly PRs to bump pinned wasm-tools since dependabot doesn't cover it.


.github/workflows/wit-tools-bump.yml
```diff
@@ -0,0 +1,92 @@
...
+          fi
+          sed -i -E "s/^(  WASM_TOOLS_VERSION: )'[^']+'/\\1'${LATEST}'/" "${FILE}"
+          {
+            echo "changed=true"
+            echo "body=- wasm-tools \`${CURRENT}\` → \`${LATEST}\` ([release notes](https://github.com/bytecodealliance/wasm-tools/releases/tag/v${LATEST}))"
+          } >> "${GITHUB_OUTPUT}"
+
+      - name: Open / update PR
+        if: steps.update.outputs.changed == 'true'
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
```
vados-cosmonic, line 72: TIL this works with `schedule` and doesn't require a PAT

.github/workflows/wit.yml
```diff
@@ -0,0 +1,327 @@
...
+        # so that a *new* wit/<pkg>/wkg.lock (e.g. a contributor added a
+        # brand-new package and forgot to commit its lock) also fails CI —
+        # `git diff` only reports changes against tracked files and would
+        # silently miss the untracked-lock case.
+        run: |
+          set -euo pipefail
+          PORCELAIN=$(git status --porcelain -- 'wit/*/wkg.lock')
+          if [[ -n "${PORCELAIN}" ]]; then
+            echo "${PORCELAIN}" >&2
+            echo "wit/<pkg>/wkg.lock is out of sync (untracked or modified) — commit the lock file(s)" >&2
```
vados-cosmonic:
```suggestion
            echo "one or more wkg.lock files are out of sync (untracked or modified) — commit the lock file(s)" >&2
```
Not sure `<pkg>` helps here (but the important output should be above)

.github/workflows/wit.yml
```diff
@@ -0,0 +1,327 @@
...
+            exit 1
+          fi
+      - name: Build publish matrix from built wasms
```
vados-cosmonic, line 182: spacing between different steps would be nice

.github/workflows/wit.yml
```diff
@@ -0,0 +1,327 @@
...
+          echo "{\"include\":[${joined}]}"
+      - name: Upload built wasm bundle
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: wit-build
+          path: .wit-build/*.wasm
+          if-no-files-found: error
+          retention-days: 7
+
+  gate:
```
vados-cosmonic: Let's keep the gate step consistently at the bottom?

Interesting that we gate only on validation and not publish... Skipped steps work right, so `needs` could be just `publish`?

vados-cosmonic (dismissed):
LGTM 🚀 

So I think we might be starting to do too much bash in scripts. I dunno if we want to introduce `xtask` or use JS or Go or something, but I'm sure hoping these scripts don't break often (or at least LLMs will always be good at modifying/updating bash!)

That said, this code seems like the kind that will not change for months if not years if it works just fine!

ricochet:
> LGTM 🚀
> 
> So I think we might be starting to do too much bash in scripts. I dunno if we want to introduce `xtask` or use JS or Go or something, but I'm sure hoping these scripts don't break often (or at least LLMs will always be good at modifying/updating bash!)
> 
> That said, this code seems like the kind that will not change for months if not years if it works just fine!

Yeah I waffled on this one, but like I did for the WASI repo, I think we should lean into JS for actions logic. I have another incoming PR for benching and for that I'll look at xtask since we're essentially building on two different crates (callgrind and criterion).

vados-cosmonic (approved):

PR #5151, opened 2026-05-14
ci(templates): add templates workflow

ricochet:
Adds a `templates` workflow that scaffolds every project under templates/ via `wash new file://$GITHUB_WORKSPACE` (local git clone, no network round-trip for the template content) and runs the result through the same fmt + clippy + build pipeline as examples by reusing `./.github/actions/build-rust-example`. The action's `image` input is now optional. Omit it to skip the OCI publish steps, which is how templates opt out of publishing.

examples.yml is rewritten so that logic between templates and examples is shared.

Both have one matrix job and now adding a new template or example is now one line.


.github/workflows/templates.yml
```diff
@@ -0,0 +1,178 @@
...
+          # and dies with `not a terminal` when stdin isn't a TTY. Wrap
+          # in script(1) to supply a PTY, and feed a bounded handful of
+          # y's so the prompt (if any) auto-accepts.
+          #
+          # Bounded matters: `yes | script …` deadlocks at job timeout
+          # because yes keeps producing after wash exits — script never
+          # sees EOF on its stdin and never returns. `printf` is finite;
+          # the pipe closes once it's drained, script exits with the
+          # child's status.
+          printf 'y\ny\ny\n' | script -qec "wash new 'file://${WORKSPACE}' --subfolder '${SUBFOLDER}' --name '${TEMPLATE}' --non-interactive" /dev/null
```
ricochet, line 152: https://github.com/wasmCloud/wasmCloud/issues/5152

.github/scripts/coverage-matrix.mjs
```diff
@@ -0,0 +1,96 @@
...
+//                     '["Cargo.toml"]' restricts coverage to Rust
+//                     projects. Default: '[]'.
+//   REQUIRE_README    Optional 'true': when set, every top-level
+//                     subdirectory of DISCOVERY_ROOT that contains an
+//                     enrolled item must have README.md at its root.
+
+import { existsSync, readdirSync } from 'node:fs';
+import { join } from 'node:path';
+
+const items = JSON.parse(process.env.ITEMS);
```
vados-cosmonic: Since modules should have top level `await` at this point, let's wrap all the actual functionality in a `async function main()` and at the bottom we can have `await main();`

IMO we should also check for missing env and error loudly

.github/scripts/coverage-matrix.mjs
```diff
@@ -0,0 +1,96 @@
...
+    missingReadme.forEach((d) => errors.push(`  ${d}/README.md`));
+  }
+}
+
+if (errors.length) {
+  errors.forEach((e) => console.error(e));
+  process.exit(1);
+}
+
+console.log(`coverage OK: ${onDisk.length} dirs under ${root}/ enrolled (${items.length} items)`);
```
vados-cosmonic: Should this be `console.error` if we want to print it to stderr as a diagnostic msg?

vados-cosmonic (dismissed):
LGTM 🚀 

This is a really interesting approach we haven't done before, interested to see how we like it! Personally like how more configurable things are and should be easier to modify as code if we need to.

.github/scripts/coverage-matrix.mjs
ricochet: I had to do a little learning here, thank you for calling this out. In case this helps others:

- In node.js scripts, console.log is for humans and console.error is for diagnostics, this feedback is about following idiomatic stream-semantics.
- console.error looks exactly the same as console.log in GitHub actions logs. I actually assumed these would look like red error messages but that only happens if you include `::error::`

jfleitz (approved):

.github/scripts/coverage-matrix.mjs
vados-cosmonic: Yeah honestly `console.error` mapping in this way catches a lot of people off guard. Normally in NodeJS scripts I'll use `process.stderr.write(...)` because it's so counter intuitive/seems like an error state rather than just write-to-stderr.

PR #5153, opened 2026-05-14
bench infra

ricochet:
## Summary

Stand up the wash-runtime benchmarking pipeline end-to-end: a dedicated Hetzner bench host, deterministic instruction-count benches in addition to the existing wall-clock ones, and a GitHub-driven workflow that publishes a self-updating trend timeline for wasmCloud/arewefastyet.

Three commits on top of `main`:

- `feat: add iai-callgrind bench and bench-tools crate`: adds the deterministic CPU-instruction bench (less immune to runner noise) plus an internal Rust CLI that parses criterion + iai output into the pipeline's JSONL / step-summary / comparison-delta formats
- `feat: bench host provisioning scripts`: operator-side automation from Hetzner rescue → ready-to-register runner. Two phases (installimage chroot + Ansible playbook), an idempotent runner installer, AWS bootstrap one-shot, and a 1Password env helper
- `ci: bench pipeline and trend-tracking workflows`: the three workflows (trends pipeline, pair-bench compare, monthly upstream version check), three plain-ESM `.mjs` scripts in `.github/scripts/`, the runbook at `scripts/bench/README.md`, CODEOWNERS, and the actionlint config for the self-hosted runner labels

Sensitive host identity (IP / IPv6 / hostname) stays in 1Password and is templated everywhere in the repo as `<WASMCLOUD_BENCH_*>`. AWS identifiers flow through repo secrets / OIDC; no credentials touch the bench host.

## Test plan

- [x] `cargo test -p bench-tools` passes locally
- [x] `cargo bench -p wash-runtime --features wasip3 --bench iai_callgrind --no-run` compiles
- [x] `ansible-playbook scripts/bench/ansible/provision.yml --check` is clean (or only shows kernel-cmdline drift)
- [ ] After merge: dispatch `bench` for `http_invoke` against `main` — verify step summary renders, S3 artifacts upload, `history.json` updates, CloudFront invalidation fires
- [ ] Dispatch `bench` for `iai_callgrind` — verify the iai instruction-count row lands in `history.json` alongside the criterion row (same `(bench, group, param)`, different `metric`)
- [ ] Dispatch `bench-compare` with `ref_a=main`, `ref_b=<this PR's head sha>` for `iai_callgrind` — verify delta table renders in the step summary
- [ ] First scheduled `bench-host-checks` run (or `workflow_dispatch`) — verify a tracking issue opens if pinned `RUNNER_VERSION` is behind upstream, and auto-closes after the bump

### DIFF
--- a/.github/scripts/bench-push-results.mjs
+++ b/.github/scripts/bench-push-results.mjs
@@ -1,0 +1,238 @@
+#!/usr/bin/env node
+// Upload one bench run's artifacts to S3, then update the public
+// history.json aggregate and invalidate CloudFront.
+//
+// Heavy JSON construction + history.json read-merge-write live here in
+// JS where they're cleaner than jq-in-bash; the actual S3/CloudFront
+// calls shell out to the `aws` CLI (already on the bench host's PATH).
+//
+// Per-run layout (private; only the bench role can read):
+//   s3://${WASMCLOUD_BENCH_S3_BUCKET}/runs/<date>/<short-sha>/<run-id>/<bench>/
+//     ├─ criterion.tar.zst   raw criterion data
+//     ├─ iai.tar.zst         raw iai-callgrind data (when applicable)
+//     ├─ results.jsonl       one JSON row per (group, param, metric)
+//     ├─ metadata.json       run-level facts
+//     └─ run.log             cargo bench stdout/stderr
+//
+// Aggregate (publicly readable through CloudFront):
+//   s3://${WASMCLOUD_BENCH_S3_BUCKET}/history.json
+//     - JSON array of every (group, param, metric) row from every run,
+//       deduped on (sha, bench, group, param, run_attempt, metric),
+//       sorted by timestamp.
+//     - Cache-Control: max-age=60.
+//     - CloudFront invalidation issued after each push.
+//
+// Reads (required):
+//   WASMCLOUD_BENCH_NAME                 bench whose output we're uploading
+//   WASMCLOUD_BENCH_S3_BUCKET            target bucket
+//   WASMCLOUD_BENCH_CF_DISTRIBUTION_ID   CloudFront distribution to invalidate
+//
+// Reads (optional):
+//   CARGO_TARGET_DIR        default /var/lib/bench/target
+//   GITHUB_RUN_ID           default "local"
+//   GITHUB_RUN_ATTEMPT      default "1"
+//   GITHUB_REF_NAME         default `git rev-parse --abbrev-ref HEAD`
+//   GITHUB_ACTOR/_EVENT_NAME/_WORKFLOW/_SERVER_URL/_REPOSITORY  — for metadata.json
+
+import { execFileSync, spawnSync } from 'node:child_process';
+import { existsSync, mkdtempSync, readFileSync, rmSync, unlinkSync, writeFileSync } from 'node:fs';
+import { hostname, tmpdir } from 'node:os';
+import { basename, join } from 'node:path';
+
+// rustup is in $HOME/.cargo/bin (not on the global PATH unless the user
+// shell sources $HOME/.cargo/env). Prepend it here so the `cargo run -p
+// bench-tools` call below resolves without us having to wrap it in `sh
+// -c '. ~/.cargo/env && …'`.
+process.env.PATH = `${process.env.HOME}/.cargo/bin:${process.env.PATH ?? ''}`;
+
+const bench = required('WASMCLOUD_BENCH_NAME');
+const bucket = required('WASMCLOUD_BENCH_S3_BUCKET');
+const distId = required('WASMCLOUD_BENCH_CF_DISTRIBUTION_ID');
+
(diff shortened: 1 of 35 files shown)


.github/scripts/bench-push-results.mjs
```diff
@@ -0,0 +1,238 @@
...
+  '--no-progress',
+  '--content-type',
+  'application/json',
+  '--cache-control',
+  'public, max-age=60',
+  histOut,
+  `s3://${bucket}/history.json`,
+]);
+
+// 8. Invalidate CloudFront so the next request hits a fresh edge cache.
```
jfleitz, line 189: Should we check if the s3 cp command was successful before invalidating? I guess maybe not an issue, since if it did error out, the cache would just refresh to the last successful run.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,161 @@
...
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: bench-${{ matrix.bench }}-${{ github.run_id }}
+          # criterion benches write to target/criterion/, iai-callgrind benches
+          # (iai_callgrind) write to target/iai/. A genuinely-empty run is a
+          # bug worth flagging loudly — fail rather than warn.
+          path: |
+            ${{ env.CARGO_TARGET_DIR }}/criterion/
+            ${{ env.CARGO_TARGET_DIR }}/iai/
+          retention-days: 90
```
jfleitz, line 148: Maybe bump to 100? Just thinking since this is ran once a month, it would guarantee we would have a history of 3 runs all of the time.

jfleitz (approved):
Overall looks good and will provide some much needed benchmarking stats!

scripts/bench/aws/setup-aws.sh
vados-cosmonic, line 1: I know it sucks but it would be great if we could make the on-machine stuff we run not *all* bash. 

It works as-is (and I'm happy to merge as is), but if we're an ansible task away from and apt install of Golang or NodeJS or whatever, it might be worth it if we can turn everything we run on-machine to something other than bash.

SUPER low priority, but a thought

vados-cosmonic (approved):
This is an absolutely massive PR, but mostly looks good to me :) 

Want to see more Node/Golang over bash, but happy to have this as a first version. 

IMO in the future we could also completely automate this with some Pulumi/Terraform and it can be "one click", but *anything* is better than nothing so do not want to hold this PR up at all.

I'd also love a local mode so someone locally could easily run this. 

Also, Packer might be a great tool to use in such a new version (and possibly be helpful across local, AWS and Hetzner)

ricochet:
> This is an absolutely massive PR, but mostly looks good to me :)
> 
> Want to see more Node/Golang over bash, but happy to have this as a first version.
> 
> IMO in the future we could also completely automate this with some Pulumi/Terraform and it can be "one click", but _anything_ is better than nothing so do not want to hold this PR up at all.
> 
> I'd also love a local mode so someone locally could easily run this.
> 
> Also, Packer might be a great tool to use in such a new version (and possibly be helpful across local, AWS and Hetzner)

yeah I've been whittling away with ansible (og was all bash to get something working). Packer I debated but felt like gross overkill for mostly off-the-shelf ubuntu lts. Would consider it again once we're looking at more than one box.

I started with terraform for hetzner baremetal (not cloud) but its super awkward and not worth it. The infra part for S3 is already pre-provisioned (terraform for the resources cosmonic is donating is in our terraform ops repo).

Not sure what you mean by local mode? All of the benches run locally with cargo bench.

scripts/bench/aws/setup-aws.sh
ricochet: We could delete this script. I left it here as a form of doc, but the chances of someone opting to reproduce this part of the bench pipeline is very unlikely. The actual bucket we use is managed in Cosmonic's own ops repo via terraform (we're paying the bill and managing the account).

vados-cosmonic: Hmnn whether we should delete it is a good question -- seems like a fork:

1. People use it and get set up easily running manually
2. People use it and get incredibly annoyed if it doesn't work

.github/scripts/bench-push-results.mjs
ricochet: Good instinct, and the good news is that it's already covered.  Unfortunately just not obvious. The run() is execFileSync which throws on non-zero exit, so if the s3 cp of history.json fails the script aborts before we hit the invalidate. CloudFront keeps serving the previous good copy until the next successful run. I'll add a comment to make that explicit. Since I don't have other changes for this one, I'm going to roll that into a follow-pr that adds k6.

.github/workflows/bench.yml
ricochet: Cadence is actually on every release so we're talking a minimum of 1 every 2 weeks (so 6 in the current 90 days window but potentially being higher like 10ish if including patch spot releases). These files are nice when diving deep, but for historical diffs not as useful (e.g. the svg charts of each run not as handy as the summary stats).

PR #5154, opened 2026-05-14
fix(wash): honor --non-interactive in wash new

### LINKED ISSUE #5152: [BUG] wash new --non-interactive does not suppress the new.command confirmation prompt
## Summary

`wash new --non-interactive` errors with `failed to read user confirmation / not a terminal` in non-TTY environments (CI, scripts) when the target template's `.wash/config.yaml` declares a `new.command`. The `--non-interactive` flag's help reads:

> Run in non-interactive mode (skip terminal checks for host exec). Automatically enabled when stdin is not a TTY

so I expected it to cover all prompts, but it only covers wash-host-exec confirmations. The post-scaffold `new.command` prompt still requires a TTY and aborts the run.

## Reproduction

Using any template that has `new.command` set (e.g. wasmCloud/wasmCloud's `templates/http-api-with-distributed-workloads`, which has `new.command: cargo fetch`):

```bash
  --subfolder templates/http-api-with-distributed-workloads \
  --name demo \
  --non-interactive < /dev/null
```

### Actual

```bash
INFO Creating new project 'demo' from git repository: …
INFO cloning git repository …
INFO Successfully cloned repository …
INFO extracting subfolder subfolder=…
INFO successfully extracted subfolder …
failed to read user confirmation

Caused by:
    0: IO error: not a terminal
    1: not a terminal
```

Exit status non-zero; the project directory is created, but new.command is never reached.

### Expected

With `--non-interactive` set (or with stdin not a TTY, which the help says auto-enables non-interactive mode), I expect wash to take the prompt's default answer and continue. No error, new.command runs, project ready.

### Why this matters

This blocks fully automated use of wash new like in CI pipelines that scaffold and build templates can't pass through wash new without either (a) faking a TTY via script(1) + piping y\n, or (b) skipping the new.command post-step and re-implementing it manually. Both are workarounds for behavior that `--non-interactive` already promises.

Concretely, wasmCloud/wasmCloud is adding a CI workflow that scaffolds every Rust template under templates/ and runs the build pipeline; this is the failure mode that drove us to wrap wash new in script(1). Even more fun is that some of these required multiple 'y' prompts so the current workaround looks like this:

```bash
printf 'y\ny\ny\n' | script -qec "wash new … --non-interactive" /dev/null
```

### Suggested fix

When `--non-interactive` is set (or stdin is not a TTY and the flag auto-enables), the new.command confirmation should take its default answer (should be yes, the user passed the flag, they've opted in) and run, rather than attempting to read a confirmation. Same handling should apply to any other interactive prompts wash adds in the future.

ricochet:
`wash new` against a template with `new.command` errored with
"failed to read user confirmation / not a terminal" when stdin
wasn't a TTY, even with `--non-interactive` set. The flag was
wired into `CliContext` but `request_confirmation` ignored it.

`request_confirmation` now returns the default answer (`true`)
when non-interactive, covering this prompt and any future ones
added through the same helper. Update the `--non-interactive`
help text to reflect the broader scope.

Extract `resolve_non_interactive(flag, stdin_is_tty)` from
`main.rs` and `run_new_command(new_cmd, working_dir)` from
`NewCommand::handle` so the fix and the existing post-scaffold
failure path are unit-testable. Adds tests covering: the
non-interactive bypass, and the zero/non-zero exit paths.

Fixes #5152

### DIFF
--- a/crates/wash/src/cli/mod.rs
+++ b/crates/wash/src/cli/mod.rs
@@ -486,14 +486,51 @@ impl CliContext {
         )
     }
 
+    /// Prompt the user for a yes/no confirmation, returning the default answer (`true`)
+    /// when running in non-interactive mode.
     pub fn request_confirmation<S>(&self, prompt: S) -> anyhow::Result<bool>
     where
         S: Into<String>,
     {
+        if self.non_interactive {
+            return Ok(true);
+        }
+
         Confirm::with_theme(&ColorfulTheme::default())
             .with_prompt(prompt)
             .default(true)
             .interact()
             .context("failed to read user confirmation")
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn request_confirmation_returns_true_in_non_interactive_mode() {
+        // CliContext::builder().build() mutates the process cwd; restore it before
+        // the tempdir drops so other parallel tests that read std::env::current_dir()
+        // don't see a deleted path.
+        let orig_cwd = std::env::current_dir().expect("get cwd");
+        let _restore = scopeguard::guard(orig_cwd, |c| {
+            let _ = std::env::set_current_dir(c);
+        });
+
+        let tempdir = tempfile::tempdir().unwrap();
+        let ctx = CliContext::builder()
+            .non_interactive(true)
+            .project_dir(tempdir.path().to_path_buf())
+            .build()
+            .await
+            .unwrap();
+
+        // Must not touch stdin: in CI / piped contexts the dialoguer call would
+        // error with "not a terminal", which is exactly the bug fix we're locking in.
+        assert!(
+            ctx.request_confirmation("would prompt the user")
+                .expect("non-interactive path should not read stdin")
+        );
     }
 }
--- a/crates/wash/src/cli/new.rs
+++ b/crates/wash/src/cli/new.rs
@@ -1,5 +1,6 @@
 //! CLI command for creating new component projects from git repositories
 
+use std::path::Path;
 use std::process::Stdio;
 
 use anyhow::{Context, bail};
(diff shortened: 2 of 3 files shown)


jfleitz (dismissed):

crates/wash/src/cli/mod.rs
```diff
@@ -486,14 +486,59 @@ impl CliContext {
...
             .interact()
             .context("failed to read user confirmation")
     }
 }
+
+/// Resolve whether wash should run in non-interactive mode.
+///
+/// Non-interactive is enabled when the user explicitly passes `--non-interactive`,
+/// or when stdin is not a TTY (e.g., piped input, CI, redirected from `/dev/null`).
+pub fn resolve_non_interactive(flag: bool, stdin_is_tty: bool) -> bool {
```
vados-cosmonic:
```suggestion
pub(crate) fn resolve_non_interactive(flag: bool, stdin_is_tty: bool) -> bool {
```

I'm not even sure this needs to... exist? it's just an or right?

crates/wash/src/cli/new.rs
```diff
@@ -151,3 +122,66 @@ impl NewCommand {
         }
     }
 }
+
+/// Execute a template's `new.command` shell command in `working_dir`.
+/// Returns an error if the command exits non-zero.
+async fn run_new_command(new_cmd: &str, working_dir: &Path) -> anyhow::Result<()> {
```
vados-cosmonic, line 128: Thanks for breaking this out!

vados-cosmonic (dismissed):
LGTM 🚀

crates/wash/src/cli/mod.rs
ricochet: Yes, good call. I had it in a couple places before refactoring out new_command and the very silly unit tests that I initially generated, not anymore.

vados-cosmonic (approved):

PR #5155, opened 2026-05-14
fix(runtime-operator): WorkloadRouteReconciler writes pod IP, not OS …

CharlesV1234:
## What

Per-workload Services created by `WorkloadRouteReconciler` get
zero-endpoint EndpointSlices on every default Kubernetes install.
Operator log every ~17s:

    creating EndpointSlice: ... Invalid value: "<hostname>":
    must be a valid IPv4 address

Symptom: traffic to a Service's ClusterIP times out;
`runtime-gateway` remains the only working ingress path.

## Why

`EndpointSlice.AddressType = IPv4` (set by the reconciler at
`workload_route_controller.go`) requires `Endpoint.Addresses`
entries to parse as IPv4. The reconciler plugs `host.Hostname`
in directly, which `host_controller.go` populates verbatim
from the heartbeat's `req.Hostname` — i.e. whatever the host
process reports via `os.Hostname()`. In every default
deployment template that's the pod name, never a pod IP.

The HostPodLabel doc-comment in `host_pod_controller.go` says
the contract is `Host.Hostname == pod.Status.PodIP`. That's
a host-process-side responsibility this PR doesn't try to fix.
What it adds is operator-side resilience: when `host.Hostname`
doesn't parse as an IP, look up the matching Pod by hostname
and use its `Status.PodIP`.

## Prior discussion

No prior issue. First report.

## Verification

* `make lint` → `0 issues`
* `make test` (non-e2e) → all packages green
* Built `docker build .` locally, swapped into a chart-v2.1.0
  install, watched per-workload EndpointSlices populate with
  the real Pod IP; `curl http://core-banking-platform:9191/`
  to the Service ClusterIP went from HTTP 000 (timeout, no
  endpoints) to HTTP 401 (gateway responded).

## Tests added

`workload_route_controller_test.go` (new) — 3 fake-client
tests covering the regression pin, the IP-direct backward-
compat path, and the skip-not-fail behaviour when no Pod
matches.

### DIFF
--- a/runtime-operator/internal/controller/runtime/workload_route_controller.go
+++ b/runtime-operator/internal/controller/runtime/workload_route_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"hash/fnv"
+	"net"
 	"reflect"
 	"sort"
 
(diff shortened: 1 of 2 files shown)


jfleitz (approved):
This is a great catch, and curious on how you are installing wasmCloud @CharlesV1234 ? In the Helm chart, the Host Name is always set to the [IP address](https://github.com/wasmCloud/wasmCloud/blob/276d00c7df728727bb149a50816235c75cba01db/charts/runtime-operator/templates/runtime/deployment.yaml#L66). However, it should always be checked to see if the host name is a valid IP format instead of assuming. Looking at the way the lookup occurs, it doesn't look like additional RBAC permissions would be needed, since the Operator already has Pod permissions to the Host namespaces.

CharlesV1234:
Thanks  @jfleitz , and good point on the chart side.

**Install path**: this is on a kustomize-style overlay (`kubectl apply -f` against the operator manifests + a small set of patches), not the Helm chart. The reason we don't pick up the chart's IP-assigning behaviour is that we deliberately pin the host pod's OS hostname via a pod-spec patch:

```yaml
spec:
  template:
    spec:
      hostname: wasmcloud-host
```

That pin is itself a workaround for the *legacy* `runtime-gateway` path — it proxies HTTP to `<host.Hostname>:<httpPort>` verbatim, and Kubernetes' default pod hostname (= the pod name like `hostgroup-default-66c55d6586-5pjgb`) isn't cluster-DNS-resolvable, so the proxy hard-fails with `dial tcp: lookup … no such host`. Pinning gives DNS something to resolve.

The reason this PR matters for us specifically: we *want* to retire `runtime-gateway` and move to the canonical post-2.1 path (per-workload Services + EndpointSlices, as described in the 2.1 release notes). At that point the hostname-pin goes away with it — but the EndpointSlice path is exactly what `WorkloadRouteReconciler` populates, and today it's broken without this fix. So this isn't just a custom-install edge case; it's load-bearing for the migration off `runtime-gateway`. We tried the `runtime-gateway`-retirement upgrade against chart v2.1.0 and rolled it back specifically because of this bug.

The bug also surfaces on at least two install patterns:

1. Custom manifests (this case), where `hostname` is intentionally a non-IP label.
2. Any default-Kubernetes setup that doesn't inject an IP via downward API and lets the pod name leak into `Host.Hostname` (the original repro on a fresh kind cluster without the chart).

PR #5171, opened 2026-05-19
feat(runtime): support wasip3 cross-component streams via dynamic linker

Aditya1404Sal:
## Summary

wash supports **dynamic linking**: a component can import an interface that's satisfied at runtime by another component (a "peer"/linked export) rather than by the host. That worked for plain synchronous wasip2 calls, but fell apart the moment wasip3's async component-model types entered the picture. This PR makes wasip3 `stream<T>`, `future<T>`, `error-context`, and resource handles flow correctly across the dynamic linker — including calling resource methods on a resource owned by another component — and reworks instance lifetime so the pooling allocator doesn't run out of slots under concurrent load.

## Background

Under wasip3 the component model gains first-class async: async functions, plus the `stream<T>`, `future<T>`, and `error-context` types. These aren't plain values — at the canonical-ABI level they're **handles** into a specific store's resource tables, the same way `own<T>` / `borrow<T>` resource handles are.

wash's existing cross-component bridge lowered call params, invoked the peer's export, and lifted the results using wasmtime's *synchronous* typed-func path. Against a wasip3 component that path has two problems:

1. It bails with **"async not supported"** as soon as a linked import or export references `stream` / `future` / `error-context`, or is an async function.
2. It tries to *copy* values across the call boundary. You can copy a plain value, but you can't copy a handle — copying a `stream` / `future` / resource handle out of its store's table produces a dangling reference. The same underlying object has to be referenced **by identity** on both sides.

So linked calls involving any of those types either errored outright or were unsound.

### 1. Concurrent canonical-ABI bridge

Switch the cross-component bridge to wasmtime's concurrent APIs — `func_new_concurrent` to register the host-side shim, `call_concurrent` / `run_concurrent` to drive the peer, and `Accessor` / `accessor.with(...)` for scoped store access inside the concurrent closure. This is the path that supports async exports and the async value types, so linked calls that touch `stream` / `future` / `error-context` — or that are simply async functions — now run instead of being rejected with "async not supported."

### 2. Identity-passing of `stream` / `future` / `error-context`

At the canonical-ABI level a `stream` / `future` / `error-context` isn't a value you can copy — it's a handle into a specific store's tables. In lift/lower we special-case those `Val` variants so the handle is passed through **by identity** rather than re-created. A `stream<T>` produced inside peer B and returned to caller A keeps referring to the same underlying stream, so data written on one side is observed on the other. Plain values continue to be lifted/lowered normally.

### 3. Cross-component resource handles and method calls

Resource handles now cross the dynamic linker, and methods can be called on a resource from a component other than the one that implements it.

When component A receives a `resource` produced by component B, the handle is carried across the linker **by identity** as a host `ResourceAny` (via `lower_with_type`) rather than copied. Calling a method on that handle from A then dispatches **back through the linker into B**, where the resource's implementation actually lives — so resource methods work transparently across the component boundary.

### 4. Per-store exporter-instance cache

To call peer B's export you need a live `Instance` of B in the store. The natural move — `pre.instantiate_async(&mut store)` — doesn't work here:

- The concurrent bridge hands us a **synchronous** `accessor.with(|access| ...)` closure. You can't `.await` inside it, so you can't instantiate there.
- By the time the p2/p3 HTTP handler has dispatched into the component, the store is already in **async-required** mode, so a synchronous instantiate is out too.

The fix is to **pre-instantiate** each linked exporter into the store *before* dispatch and cache the resulting `Instance` keyed by `(component_id, store_id)`. Inside the sync concurrent closure we just look the instance up. The cache is cleared on store teardown. (The shared-store path is intentionally **error-on-miss** — see design notes.)

### 5. Ephemeral stores for plain-value async linked calls

The cache above keeps each linked instance alive for as long as its store lives — and a request store lives for the whole response. With the pooling allocator every instance holds a fixed core-instance slot, so under concurrent P3 load long-lived request stores pin slots and you exhaust capacity.

For the common case where a linked call's signature is **entirely plain values** — no `own` / `borrow` / `stream` / `future` / `error-context` anywhere in params or results — the callee doesn't need to share the request store's tables at all. So we run it in a **short-lived store** that is created, instantiated, called, and dropped within that single call, reclaiming its pooling slots immediately instead of at end-of-response.

The decision is made per call via `func_is_ephemeral_safe` / `type_is_ephemeral_safe`. Any signature carrying a handle stays on the shared request store, because those handles are bound to that store and can't be moved to a throwaway one.

## Test coverage

Each capability is exercised end-to-end through fixture components linked at runtime:

- **`integration_p3_streams`** — three cases: `test_p3_cross_component_stream_to_http` sends a `stream<u8>` from one peer through another to the HTTP response and checks it reassembles byte-for-byte; `test_p3_cross_component_stream_streams_incrementally` paces the producer and times arrivals at the client, proving the stream stays *concurrent across the linker* (first chunk `< last/2`) rather than being buffered at the boundary — the byte-for-byte test above passes even on a buffered path, so this is the one that actually pins down the concurrency; `test_p3_incoming_handler_streams_incrementally` runs the same timing check on a single self-contained handler.
- **`integration_p3_resources`** — the `res-producer-p3` / `res-caller-p3` / `res-sink-p3` trio: the producer mints a `token` and implements `token.greet()`; the caller hands that token to the sink; the sink calls `token.greet()`, which routes **back to the producer**; the test asserts the round-trip body `"sink:hello world"`. If the handle were copied instead of passed by identity, `greet()` would read the wrong resource or fail.
- **`integration_p3_ephemeral`** — *new.* The stream and resource fixtures all carry handles, so they only ever take the shared-store path and would never touch the ephemeral path. This adds an `ephemeral-callee-p3` / `ephemeral-caller-p3` pair with a plain-value async signature and asserts the call goes through, giving the ephemeral path real regression coverage.

## Design notes & limitations

- **Error-on-miss, not lazy-instantiate.** The shared-store path errors if a linked exporter is missing from the cache rather than instantiating on demand, because the wasmtime fork in use doesn't expose an accessor-side instantiate, and pre-instantiation already covers every legitimate call site. This keeps a missing-instance situation loud instead of silently slow.
- **Ephemeral is plain-value-only by construction.** It is never selected for handle-bearing signatures, so it can't accidentally orphan a resource / stream / future. When in doubt, it falls back to the shared store.
- **Pooling-allocator interaction.** The whole instance-lifetime story (cache + ephemeral stores) exists specifically because of the pooling allocator's fixed per-instance slots; with the on-demand allocator the slot pressure this addresses wouldn't arise, but pooling is what production hosts run.

## Upstream dependencies

- **Wasmtime 46**, via the `release-46.0.0-implements` fork patch — the concurrent canonical-ABI APIs this PR relies on come from there.
- Builds on upstream making **wasip3 the default**: the old `wasip3` Cargo feature is gone, and the P3 WIT deps are pinned to `wasi:*@0.3.0`.

## Test plan

- [ ] `cargo build -p wash` from the repo root succeeds (wasip3 is on by default now).
- [ ] `cargo test -p wash-runtime` passes — in particular:
  - [ ] `integration_p3_streams::test_p3_cross_component_stream_to_http` (cross-component `stream<u8>` → HTTP body == `"abcdefghijklmnop"`).
  - [ ] `integration_p3_streams::test_p3_cross_component_stream_streams_incrementally` (paced producer; first chunk < last/2 → stream stays concurrent across the linker).
  - [ ] `integration_p3_streams::test_p3_incoming_handler_streams_incrementally` (first chunk < last/2 → streamed, not buffered).
  - [ ] `integration_p3_resources::test_p3_resource_handle_crosses_linker` (resource handle + cross-component `greet()` → `"sink:hello world"`).
  - [ ] `integration_p3_ephemeral` (plain-value async linked call takes the ephemeral path).
- [ ] No regression in `cargo test -p wash`.

### DIFF
--- a/crates/wash-runtime/src/engine/ctx.rs
+++ b/crates/wash-runtime/src/engine/ctx.rs
@@ -100,51 +100,77 @@ impl SharedCtx {
     }
 }
 
-pub struct AccessorActiveCtxGuard<'a> {
+/// RAII guard that points [`SharedCtx::active_ctx`] at a linked component for
+/// the duration of a cross-component call, restoring the previous component on
+/// drop.
+///
+/// TODO(concurrency): the single `active_ctx` slot with LIFO save/restore is
+/// only correct for *sequential* linked calls. `func_new_concurrent` allows
+/// several in flight on one shared store, and their save/restore interleaves —
+/// A sets B; C (concurrent) saves B, sets D; A restores B; C restores B — so
+/// the original ctx is lost and `active_ctx` is wrong even after both finish
+/// (and mid-flight, across awaits, it need not match the running call). The
+/// single-call fixtures issue one at a time so they don't exercise this; a real
+/// fix needs per-call ctx scoping rather than a shared slot. Tracked follow-up.
+pub(crate) struct AccessorActiveCtxGuard<'a> {
     accessor: &'a Accessor<SharedCtx>,
-    previous: Arc<str>,
+    previous_component_id: Arc<str>,
 }
 
 impl<'a> AccessorActiveCtxGuard<'a> {
-    pub fn new(accessor: &'a Accessor<SharedCtx>, id: &Arc<str>) -> wasmtime::Result<Self> {
-        let previous = accessor.with(|mut access| -> wasmtime::Result<_> {
-            let previous = access.data_mut().active_ctx.component_id.clone();
+    pub(crate) fn new(accessor: &'a Accessor<SharedCtx>, id: &Arc<str>) -> wasmtime::Result<Self> {
+        let previous_component_id = accessor.with(|mut access| -> wasmtime::Result<_> {
+            let previous_component_id = access.data_mut().active_ctx.component_id.clone();
             access.data_mut().set_active_ctx(id)?;
-            Ok(previous)
+            Ok(previous_component_id)
         })?;
 
-        Ok(Self { accessor, previous })
+        Ok(Self {
+            accessor,
+            previous_component_id,
+        })
     }
 }
 
 impl Drop for AccessorActiveCtxGuard<'_> {
     fn drop(&mut self) {
-        let _ = self
-            .accessor
-            .with(|mut access| access.data_mut().set_active_ctx(&self.previous));
+        let _ = self.accessor.with(|mut access| {
+            access
+                .data_mut()
+                .set_active_ctx(&self.previous_component_id)
+        });
     }
 }
 
-pub struct StoreActiveCtxGuard<'a> {
+pub(crate) struct StoreActiveCtxGuard<'a> {
     store: StoreContextMut<'a, SharedCtx>,
-    previous: Arc<str>,
+    previous_component_id: Arc<str>,
 }
 
 impl<'a> StoreActiveCtxGuard<'a> {
-    pub fn new(mut store: StoreContextMut<'a, SharedCtx>, id: &Arc<str>) -> wasmtime::Result<Self> {
-        let previous = store.data().active_ctx.component_id.clone();
+    pub(crate) fn new(
+        mut store: StoreContextMut<'a, SharedCtx>,
+        id: &Arc<str>,
+    ) -> wasmtime::Result<Self> {
+        let previous_component_id = store.data().active_ctx.component_id.clone();
         store.data_mut().set_active_ctx(id)?;
-        Ok(Self { store, previous })
+        Ok(Self {
+            store,
+            previous_component_id,
+        })
     }
 
-    pub fn store_mut(&mut self) -> &mut StoreContextMut<'a, SharedCtx> {
+    pub(crate) fn store_mut(&mut self) -> &mut StoreContextMut<'a, SharedCtx> {
         &mut self.store
     }
 }
 
 impl Drop for StoreActiveCtxGuard<'_> {
     fn drop(&mut self) {
-        let _ = self.store.data_mut().set_active_ctx(&self.previous);
+        let _ = self
+            .store
+            .data_mut()
+            .set_active_ctx(&self.previous_component_id);
     }
 }
 
@@ -192,7 +218,7 @@ pub struct Ctx {
     /// Unique identifier for this component context. This is a [uuid::Uuid::new_v4] string.
     pub id: Arc<str>,
     /// Unique identifier shared by all component contexts in the same store.
-    pub store_id: String,
+    pub store_id: Arc<str>,
     /// The unique identifier for the workload component this instance belongs to
     pub component_id: Arc<str>,
     /// The unique identifier for the workload this component belongs to
@@ -407,7 +433,7 @@ impl wasmtime_wasi_http::p3::WasiHttpHooks for CtxHttpHooksP3 {
 /// Helper struct to build a [`Ctx`] with a builder pattern
 pub struct CtxBuilder {
     id: Arc<str>,
-    store_id: String,
+    store_id: Arc<str>,
     workload_id: Arc<str>,
     component_id: Arc<str>,
     ctx: Option<WasiCtx>,
@@ -424,7 +450,7 @@ impl CtxBuilder {
     pub fn new(workload_id: impl Into<Arc<str>>, component_id: impl Into<Arc<str>>) -> Self {
         Self {
             id: uuid::Uuid::new_v4().to_string().into(),
-            store_id: uuid::Uuid::new_v4().to_string(),
+            store_id: uuid::Uuid::new_v4().to_string().into(),
             component_id: component_id.into(),
             workload_id: workload_id.into(),
             ctx: None,
--- a/crates/wash-runtime/src/engine/ctx.rs
+++ b/crates/wash-runtime/src/engine/ctx.rs
@@ -138,6 +138,8 @@ impl<'a> DerefMut for ActiveCtx<'a> {
 pub struct Ctx {
     /// Unique identifier for this component context. This is a [uuid::Uuid::new_v4] string.
     pub id: Arc<str>,
+    /// Unique identifier shared by all component contexts in the same store.
+    pub store_id: String,
     /// The unique identifier for the workload component this instance belongs to
     pub component_id: Arc<str>,
     /// The unique identifier for the workload this component belongs to
@@ -352,6 +354,7 @@ impl wasmtime_wasi_http::p3::WasiHttpHooks for CtxHttpHooksP3 {
 /// Helper struct to build a [`Ctx`] with a builder pattern
 pub struct CtxBuilder {
     id: Arc<str>,
+    store_id: String,
     workload_id: Arc<str>,
     component_id: Arc<str>,
     ctx: Option<WasiCtx>,
@@ -368,6 +371,7 @@ impl CtxBuilder {
     pub fn new(workload_id: impl Into<Arc<str>>, component_id: impl Into<Arc<str>>) -> Self {
         Self {
             id: uuid::Uuid::new_v4().to_string().into(),
+            store_id: uuid::Uuid::new_v4().to_string(),
             component_id: component_id.into(),
             workload_id: workload_id.into(),
             ctx: None,
@@ -444,6 +448,7 @@ impl CtxBuilder {
 
         Ctx {
             id: self.id,
+            store_id: self.store_id,
             ctx: self.ctx.unwrap_or_else(|| {
                 WasiCtxBuilder::new()
                     .args(&["main.wasm"])
--- a/crates/wash-runtime/src/engine/ctx.rs
+++ b/crates/wash-runtime/src/engine/ctx.rs
@@ -11,7 +11,8 @@ use std::{
     sync::Arc,
 };
 
-use wasmtime::component::ResourceTable;
+use wasmtime::StoreContextMut;
+use wasmtime::component::{Accessor, Instance, ResourceTable};
 use wasmtime_wasi::{WasiCtx, WasiCtxBuilder, WasiCtxView, WasiView};
 use wasmtime_wasi_http::WasiHttpCtx;
 use wasmtime_wasi_http::p2::{WasiHttpCtxView, WasiHttpHooks, WasiHttpView};
@@ -67,6 +68,15 @@ pub struct SharedCtx {
     pub table: wasmtime::component::ResourceTable,
     /// Contexts for linked components
     pub contexts: HashMap<Arc<str>, Ctx>,
+    /// Pre-instantiated linked-component instances, keyed by component id.
+    ///
+    /// Owned by the store: every store created by the workload's store factory
+    /// is born with its linked components instantiated here, and they are
+    /// reclaimed automatically when the store is dropped (no external cache,
+    /// guards, or completion hooks). The concurrent cross-component call path
+    /// reads from here under `accessor.with`, where it cannot instantiate on
+    /// demand.
+    pub exporter_instances: HashMap<Arc<str>, Instance>,
 }
 
 impl SharedCtx {
@@ -75,6 +85,7 @@ impl SharedCtx {
             active_ctx: context,
             table: ResourceTable::new(),
             contexts: Default::default(),
+            exporter_instances: Default::default(),
         }
     }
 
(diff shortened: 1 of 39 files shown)


Aditya1404Sal:
Blocked till wasmtime releases v46.0

Aditya1404Sal:
Will be closing this as it's supposed to be now opened against wasmtime-46

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -95,6 +100,54 @@ impl SharedCtx {
     }
 }
 
+pub struct AccessorActiveCtxGuard<'a> {
+    accessor: &'a Accessor<SharedCtx>,
+    previous: Arc<str>,
```
vados-cosmonic: Can we call this something more specific? maybe `previous_component_id` ?

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -95,6 +100,54 @@ impl SharedCtx {
...
+    }
+}
+
+pub struct StoreActiveCtxGuard<'a> {
+    store: StoreContextMut<'a, SharedCtx>,
+    previous: Arc<str>,
+}
+
+impl<'a> StoreActiveCtxGuard<'a> {
+    pub fn new(mut store: StoreContextMut<'a, SharedCtx>, id: &Arc<str>) -> wasmtime::Result<Self> {
```
vados-cosmonic: Can these functions be `pub(crate)` ? Maybe I'm not seeing where they're used outside of `wash-runtime`

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -352,6 +407,7 @@ impl wasmtime_wasi_http::p3::WasiHttpHooks for CtxHttpHooksP3 {
 /// Helper struct to build a [`Ctx`] with a builder pattern
 pub struct CtxBuilder {
     id: Arc<str>,
+    store_id: String,
```
vados-cosmonic: Can we use an `Arc<str>` like the other IDs?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
     Vec<String>,
 );
 
-/// Shared slot holding the most recently instantiated component, keyed by its id.
-/// Used while resolving inter-component imports.
-type SharedInstanceSlot = Arc<RwLock<Option<(Arc<str>, Instance)>>>;
+/// Lower dynamic-linker call params, using each param's declared type so host
+/// resource handles cross the linker by identity (see [`lower_with_type`]).
+fn lower_params(
```
vados-cosmonic: NIT: can we move lowering/lifting helper functions to another file?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
...
+struct ResolvedVolumeMount {
+    host_path: PathBuf,
+    mount_path: String,
+    dir_perms: DirPerms,
+    file_perms: FilePerms,
+}
+
+impl ResolvedVolumeMount {
+    async fn from_mount(host_path: &PathBuf, mount: &VolumeMount) -> anyhow::Result<Self> {
+        let host_path = tokio::fs::canonicalize(host_path).await?;
```
vados-cosmonic: It's probably worth adding a `.with_context` here that includes the path that was attempting to be canonicalized -- IIRC the error messages for tokio::fs are really bad, just like stds, and you can't tell what file you were originally looking for, etc.

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
...
+        if component.metadata.resolved_volume_mounts.is_empty() {
+            component.metadata.resolved_volume_mounts = resolved_volume_mounts;
+        }
+    }
+
+    Ok(())
+}
+
+#[derive(Clone)]
+struct ComponentCtxTemplate {
```
vados-cosmonic: Some documentation would be nice -- what is the template used for? What other pieces of code are related?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
...
+    metadata: &WorkloadMetadata,
+    tls_provider: Option<SharedTlsProvider>,
+) -> ComponentCtxTemplate {
+    let mut template = ComponentCtxTemplate::from_metadata(metadata);
+    template.tls_provider = tls_provider;
+    template
+}
+
+#[derive(Clone)]
+struct EphemeralLinkedCall {
```
vados-cosmonic: Some documentation on this struct would be great -- where it's likely to be used, circumstances, the mechanics of a cross-component call and where this struct fits in

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
...
+    let call_export_name = inv.export_name.clone();
+    let call_pre = inv.pre.clone();
+    let func_idx = inv.func_idx;
+    let call_result = tokio::task::spawn(async move {
+        let instance = call_pre.instantiate_async(&mut store).await?;
+        store
+            .run_concurrent(async move |accessor| {
+                let func = accessor.with(|mut access| -> wasmtime::Result<_> {
+                    instance.get_func(&mut access, func_idx).ok_or_else(|| {
+                        wasmtime::format_err!(
```
vados-cosmonic: IIRC wasmtime errors support `.context` as well BTW

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
...
+    .map_err(|e| wasmtime::format_err!("ephemeral linked call task failed: {e}"));
+    let call_result = call_result??;
+
+    for (i, v) in call_result.into_iter().enumerate() {
+        *results
+            .get_mut(i)
+            .ok_or_else(|| wasmtime::format_err!("result index out of bounds"))? = v;
+    }
+
+    trace!(
```
vados-cosmonic: Can we add this trace before the call starts as well?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
...
+        name = %inv.import_name,
+        fn_name = %inv.export_name,
+        ?results,
+        "invoked ephemeral dynamic export"
+    );
+
+    Ok(())
+}
+
+async fn invoke_shared_store_linked_export(
```
vados-cosmonic: Can we put all these invocation helpers in another module?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -41,9 +43,585 @@ type BoundPluginWithInterfaces = (
...
+            lift_results(&mut access.as_context_mut(), results_buf, results)
+        })?;
+
+        Ok(())
+    }
+    .await;
+
+    call?;
+
+    trace!(name = %inv.import_name, fn_name = %inv.export_name, "invoked dynamic export");
```
vados-cosmonic: Let's trace before as well

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -644,6 +1236,51 @@ impl ResolvedWorkload {
         &self.host_interfaces
     }
 
+    async fn resolve_component_volume_mounts(
```
vados-cosmonic, line 671: Can we move all workload code related to volumes to a separate module?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -644,6 +1236,51 @@ impl ResolvedWorkload {
...
+        {
+            service.metadata.resolve_volume_mounts().await?;
+        }
+        Ok(())
+    }
+
+    fn component_ctx_template(&self, metadata: &WorkloadMetadata) -> ComponentCtxTemplate {
+        #[cfg(feature = "wasi-tls")]
+        {
+            let mut template = ComponentCtxTemplate::from_metadata(metadata);
```
vados-cosmonic: Why didn't we call the `with_tls(...)` fn here? IMO this pattern is actually better (i.e. adding a tls provider only when in teh feature is present. I think you can move the cfg stanza further into this block -- like moving an if statement inside.

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -889,120 +1558,69 @@ impl ResolvedWorkload {
...
+                                                let inv = inv.clone();
+                                                Box::new(async move {
+                                                    invoke_linked_sync_export(
+                                                        store, params, results, &inv,
+                                                    )
+                                                    .await
+                                                })
+                                            },
+                                        )
+                                        .map_err(|e| e.context("failed to create async func"))?;
```
```suggestion
                                        .map_err(|e| e.context("failed to wrap sync func in async func"))?;
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1104,124 +1722,92 @@ impl ResolvedWorkload {
...
+        // the store factory pre-instantiates linked components, which takes a
+        // `components` write lock (via `instantiate_pre`) — holding the read
+        // lock across that call would deadlock.
+        let metadata = {
+            let components = self.components.read().await;
+            components
+                .get(&component_id)
+                .context("component ID not found in workload")?
+                .metadata
+                .clone()
```
vados-cosmonic, line 1173: If we're going to clone the metadata out, maybe we should do it while we have the lock above?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1957,6 +2543,19 @@ fn topological_sort_components(
     Ok(result)
 }
 
+fn expand_link_closure(
```
vados-cosmonic, line 1965: This looks like it flattens the direct links list w/ one level, is that right? Could we add some comments to that effect or maybe change the the name of this fn?

crates/wash-runtime/tests/fixtures/ephemeral-callee-p3/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+    });
+}
+
+use bindings::exports::wasmcloud::ephemeral_test::compute::Guest;
+
+struct Component;
+
+impl Guest for Component {
+    async fn run(n: u32) -> u32 {
+        // Deterministic plain-value transform the caller can assert on.
```
vados-cosmonic: I think you can remove this comment

crates/wash-runtime/tests/integration_p3_resources.rs
```diff
@@ -0,0 +1,91 @@
+//! Integration test for passing a guest resource handle across the WASIP3
+//! dynamic linker.
+//!
+//! `res-caller-p3` (the HTTP entrypoint) gets a `token` resource from
```
vados-cosmonic: Can we rewrite this section in a point by point (numbered/bulleted list) of what is supposed to happen? This is a great place to put documentation on exactly how we expect these calls to work.

vados-cosmonic (dismissed):
LGTM 🚀 

Mostly this looks good to me, great work here @Aditya1404Sal -- great that we're covering this under test as well. I think mostly if we can improve the organization of stuff and do a little more documenting this code can stay easy to maintain into the future

crates/wash-runtime/src/engine/workload.rs
Aditya1404Sal: Good call -- I made it eager instead. Components now canonicalize their volume mounts once in resolve() (right after linking), same as the service does at startup, so by the time `new_store` runs the mounts are always resolved. 
That let me drop the lazy resolve from `new_store` and collapse it to a single lock with the clone right under it. Bonus: a bad volume path now fails at deploy time instead of silently on the first request :)

ricochet:
I have another refactor that I will aim to add on top of this which is largely complementary with support for p3 types on our dynamic linker. The goal is to prevent deadlocks on the Service while components may block_on their requests.

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -95,6 +100,68 @@ impl SharedCtx {
     }
 }
 
+pub(crate) struct AccessorActiveCtxGuard<'a> {
```
ricochet, line 115: The active_ctx swap appears to be racy. 

This is LIFO and holds for sequential calls, but func_new_concurrent permits multiple linked calls in-flight on the same shared store.


call A saves prev=comp0, sets B; call C (a second concurrent linked import from the same guest) saves prev=B, sets D; A drops → restores B; C drops → restores B. comp0 lost and active_ctx is wrong even after both complete.

The single fixtures issue one call at a time so they won't catch this.

I'm ok if we doc this now and then immediately follow-up with a fix

Aditya1404Sal: gonna doc this as tracked for fix-up

ricochet: cool I think I have a fix, so will put up a follow-on pr

Aditya1404Sal: sounds good 🚀

ricochet (approved):
1. make active_ctx switch concurrency safe
2. invoke_ephemeral_linked_export cancellation safe
3. Ctx::store_id is dead
4. (Nit) `#![recursion_limit = "512"]` in `wash/src/lib.rs` is presumably for the deeply-recursive carries_cross_store_handle/bindgen types. Comment would be handy
5. shared-store concurrent path for streams doesn't have a timeout like the other calls (30s)

PR #5183, opened 2026-05-20
ci(release): detect release via Cargo.toml version

ricochet:
Removes the `startsWith(commit_msg, 'release: v')` convention as the
"this is a release" signal. The version-of-truth is now
`[workspace.package].version` in Cargo.toml; a release is detected by
comparing HEAD's value against HEAD~1's. Same observable behavior for
release-train's automatic flow (it bumps Cargo.toml before merging),
but the commit subject is no longer load-bearing — any subject the
contributor uses works.

wash.yml:
- New `release-detect` job, runs early. Outputs `is_release` and
  `version`. Reads `Cargo.toml` at HEAD and HEAD~1, marks `is_release=true`
  when they differ (or when the workflow_dispatch input `is_release` is
  true).
- `canary-binaries` gates on `needs.release-detect.outputs.is_release ==
  'true'` instead of `startsWith(...)`.
- `gate`'s release-commit check uses the same output (env var
  `IS_RELEASE` replaces `COMMIT_MESSAGE`).
- New `workflow_dispatch` trigger with optional `is_release` input. The
  manual recovery path becomes
    gh workflow run wash.yml --ref main -f is_release=true
  for cases where Cargo.toml is already at the target version and we
  need to re-build the binaries without bumping again (e.g., recovering
  a release after a partial release-tag run).

release-tag.yml verify-head:
- Manual-sha path unchanged (already reads Cargo.toml directly).
- Automatic path: drop the subject regex; require HEAD to bump the
  workspace version vs HEAD~1, matching wash.yml's release-detect.

Header docs and the canary-binaries comment block updated to describe
the new detection mechanism. release-train.yml is unchanged. Its
commits still happen to start with `release: v…` for human readability,
but nothing in CI checks for it any more.

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -24,16 +24,15 @@ name: release-tag
 #   - release-train.yml waits for the PR to actually merge before exiting,
 #     so by the time this workflow fires the merge commit is on main.
 #
-# workflow_dispatch is a recovery hatch: if release-train.yml fails late
-# (e.g. DCO bounces and a human finishes the merge manually), the train
-# run's conclusion is `failure` and this workflow stays skipped. Dispatch
-# manually with `sha=<merge-sha>` once canary builds on that SHA are green
-# — when sha is passed, verify-head reads the version from Cargo.toml at
-# that sha and skips the commit-subject match, so the recovery path works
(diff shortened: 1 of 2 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5187, opened 2026-05-21
ci(runtime-operator): guard RBAC drift

ricochet:
Three checks that would have caught the missing `*/finalizers` rule fixed in https://github.com/wasmCloud/wasmCloud/pull/5186

* Enable OwnerReferencesPermissionEnforcement in the kind e2e cluster so the existing WorkloadReplicaSet-creation assertion exercises the same admission check that surfaces blockOwnerDeletion RBAC gaps in production (GKE/AKS/OpenShift). NodeRestriction is re-listed because `--enable-admission-plugins` overrides kubeadm's default set.
* New `manifests` job runs `make manifests` and fails on diff via the existing no-diff action — catches +kubebuilder:rbac markers that were edited without regenerating, and markers placed where controller-gen silently skips them.
* New runtime-operator-rbac-parity.mjs renders the chart in two modes (defaults and watchNamespaces=[ns-a], hostNamespaces=[ns-b]), walks RoleBindings/ClusterRoleBindings to find roles bound to the operator SA, and asserts every (apiGroup, resource, verb) tuple in config/rbac/role.yaml is granted by at least one. Scope-agnostic by design — the chart's deliberate cluster-wide → namespaced re-mapping is not what this script polices.

### DIFF
--- a/.github/scripts/runtime-operator-rbac-parity.mjs
+++ b/.github/scripts/runtime-operator-rbac-parity.mjs
@@ -1,0 +1,174 @@
+#!/usr/bin/env node
+// Verifies that the runtime-operator Helm chart grants every (apiGroup,
+// resource, verb) tuple that controller-gen emits into
+// runtime-operator/config/rbac/role.yaml.
+//
+// The chart is hand-mapped from role.yaml — it splits the single generated
+// ClusterRole across a structural layout of (Cluster)Role + (Cluster)RoleBinding
+// (e.g. pods move to a namespaced Role in the operator's namespace and per-host
+// namespaces; configmaps/secrets/services/events move to a ClusterRole bound
+// per-namespace when operator.watchNamespaces is set). Because the mapping is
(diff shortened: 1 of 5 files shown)


ricochet:
Correctly failed and caught the services missing finalizers: 
```bash
  2026-05-21T18:20:02Z	ERROR	Reconciler error	{"controller": "workload-route", "namespace": "wasmcloud-system", "name": "hello-workload", "reconcileID": "b2cc1402-bf8c-41e9-8323-eac2069fd1e8", "error": "creating EndpointSlice: endpointslices.discovery.k8s.io \"hello-workload-318a658f\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
```

https://github.com/wasmCloud/wasmCloud/actions/runs/26244482441/job/77240039739?pr=5187

vados-cosmonic (approved):
LGTM 🚀

PR #5188, opened 2026-05-21
feat(wash-runtime): namespace host plugin spans by WIT package

### LINKED ISSUE #5185: OTEL prefix spans with plugin/package name
The `wasi:blobstore` and `wasi:keyvalue` host plugins in `wash-runtime` emit spans whose names are the bare Rust function names (or, for keyvalue, no `#[instrument]` at all). Downstream observability like Tempo/Grafana/Perses dashboards that want to surface "blobstore activity" or "keyvalue activity" don't have a great way to select for them without hard-coding every function name.

## Repro

Run a workload that uses `wasi:blobstore` (e.g. `blobby`). With OTLP traces flowing into Tempo, the span names visible are:

```text
container_exists, create_container, delete_container, get_container,
copy_object, move_object, get_data, write_data, list_objects, delete_object,
delete_objects, has_object, object_info, clear, new_outgoing_value,
outgoing_value_write_body, finish, incoming_value_consume_sync,
incoming_value_consume_async
```

No `blobstore.` (or any other) prefix. The names also collide with anything else that happens to be called `get_data`, `clear`, `finish`, etc.

A natural TraceQL filter like `{ name =~ "blobstore.*" }` matches nothing.

`wasi:keyvalue` is even less observable, `plugin/wasi_keyvalue/nats.rs` doesn't use `#[instrument]` on its bucket impl at all, so per-operation spans don't appear in Tempo unless something else creates them.

## Proposed fix

Add `name = "<namespace>.<package>.<fn>"` to each `#[instrument]` macro in the plugin impls, and add `#[instrument]` to the keyvalue impls where missing.

### blobstore (`crates/wash-runtime/src/plugin/wasi_blobstore/nats.rs`)

```diff
-    #[instrument(skip(self))]
+    #[instrument(name = "wasi.blobstore.create_container", skip(self))]
     async fn create_container(...)

-    #[instrument(skip(self))]
+    #[instrument(name = "wasi.blobstore.get_container", skip(self))]
     async fn get_container(...)

   # ... and so on for: delete_container, container_exists, copy_object,
   # move_object, get_data, write_data, list_objects, delete_object,
   # delete_objects, has_object, object_info, clear,
   # new_outgoing_value, outgoing_value_write_body, finish,
   # incoming_value_consume_sync, incoming_value_consume_async
```

## Why not make it look like wit?

"wasi:.*" needs careful regex handling around `:` in some engines; prometheus label values are fine but anything that round-trips through other systems (curl URL-encoding in dashboards, JSON escaping, search syntax in some tools) gets noisier. Dots are the semantic  convention for OTEL. Using `/` for before function names, is rare enough that some log/trace pipelines treat it as a path separator and will mangle.

ricochet:
Plugin handler spans were named with just the bare Rust fn (create_container,
get_data, finish, ...), which collided with anything else of the same name
and gave Tempo/Grafana no prefix to filter on. Add an explicit
`name = "<namespace>.<package>.<fn>"` to every #[instrument] on a WIT-bound
handler so dashboards can select e.g. `{ name =~ "wasi.blobstore.*" }`.

- wasi:blobstore (nats, filesystem, in_memory): rename existing spans,
  add instruments to the handlers that were missing them
  (object_info, new_outgoing_value, outgoing_value_write_body in
  filesystem; all 19 handlers in in_memory)
- wasi:keyvalue (nats, redis, in_memory): add #[instrument] across
  store/atomics/batch handlers — previously had none on the non-fs
  backends, so per-op spans never appeared in Tempo at all
- wasi:config, wasi:logging: rename existing spans; add to logging.log
- wasmcloud:messaging, wasmcloud:postgres: rename existing spans

Also fixes an unrelated bug in wasi_keyvalue/nats.rs where `set`
recorded the OTel operation counter as "get".

Adds tests/integration_blobstore_span_names.rs which drives the
http-blobstore fixture end-to-end, captures every span via a custom
tracing Layer, and asserts the expected wasi.blobstore.* spans are
emitted with the right shape — guards against future drift.

Closes #5185

### DIFF
--- a/crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
+++ b/crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
@@ -88,7 +88,7 @@ impl FilesystemBlobstore {
 
 // Implementation for the main blobstore interface
 impl<'a> bindings::wasi::blobstore::blobstore::Host for ActiveCtx<'a> {
-    #[instrument(skip(self))]
+    #[instrument(name = "wasi.blobstore.create_container", skip(self))]
     async fn create_container(
         &mut self,
         name: ContainerName,
@@ -113,7 +113,7 @@ impl<'a> bindings::wasi::blobstore::blobstore::Host for ActiveCtx<'a> {
         Ok(Ok(resource))
     }
 
-    #[instrument(skip(self))]
+    #[instrument(name = "wasi.blobstore.get_container", skip(self))]
     async fn get_container(
         &mut self,
         name: ContainerName,
@@ -139,7 +139,7 @@ impl<'a> bindings::wasi::blobstore::blobstore::Host for ActiveCtx<'a> {
         Ok(Ok(resource))
     }
 
-    #[instrument(skip(self))]
+    #[instrument(name = "wasi.blobstore.delete_container", skip(self))]
     async fn delete_container(
         &mut self,
         name: ContainerName,
@@ -162,7 +162,7 @@ impl<'a> bindings::wasi::blobstore::blobstore::Host for ActiveCtx<'a> {
         Ok(Ok(()))
     }
 
-    #[instrument(skip(self))]
+    #[instrument(name = "wasi.blobstore.container_exists", skip(self))]
     async fn container_exists(
         &mut self,
         name: ContainerName,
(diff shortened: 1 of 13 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Thanks or fixing this -- I guess this is just something people will have to do for themselves to make the traces a bit easier to understand. Far in the future we can probably figure out a way to make this automatic, but IMO it's pretty easy to understand as-is.

PR #5191, opened 2026-05-21
ci(release-tag): pin previous_tag_name when generating release notes

ricochet:
`generate_release_notes: true` on action-gh-release delegated to GitHub's
/releases/generate-notes API without specifying which tag to compute the
diff from. The auto-pick walks every tag namespace in the repo
(wash-cli-v…, runtime-operator/v…, control-interface-v…, plus the
wasmCloud release v… series) and picked a tag well before v2.0.6 as
"previous" for v2.2.0, producing a body with ~84 PRs instead of the ~44
actually merged since v2.1.0.

Add `.github/scripts/generate-release-notes.mjs` (node built-ins only):
read the wasmCloud release tags via `git tag --list 'v[0-9]*.[0-9]*.[0-9]*'
--sort=-version:refname`, find the predecessor of the current tag, and
call generate-notes with `previous_tag_name` pinned.

Tag-type policy:
- GA `vX.Y.Z`: previous = immediate semver predecessor (or empty,
  first-release-ever → API auto-detect).
- RC `vX.Y.Z-rc.N`: previous = `vX.Y.Z-rc.M` (same X.Y.Z base) if
  present in the list; otherwise empty body. Lets a monotonic RC
  chain (rc.1 → rc.2 → rc.3) ship notes showing what changed
  between RCs.
- Anything else with `-` (alpha, beta, draft, prealpha, …): empty
  body. Auto-generated diffs against a prior GA are noise on an
  in-progress release.

action-gh-release now takes `body_path: release-notes.md` instead of
`generate_release_notes: true`.

### DIFF
--- a/.github/scripts/generate-release-notes.mjs
+++ b/.github/scripts/generate-release-notes.mjs
@@ -1,0 +1,127 @@
+#!/usr/bin/env node
+// Generates the body for a GitHub Release.
+//
+// GitHub's /releases/generate-notes API picks `previous_tag_name` by
+// walking every tag in the repo when not given an explicit one — that
+// includes the non-wasmcloud tag namespaces (wash-cli-v…,
+// runtime-operator/v…, control-interface-v…) and ends up picking the
+// wrong predecessor for the wasmCloud release `v…` series. Compute
+// the predecessor here from the release namespace only and pin it.
+//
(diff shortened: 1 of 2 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5195, opened 2026-05-22
chore: wasmtime 44.0.2 security patch

ricochet:
Caught in cargo-audit CI


vados-cosmonic (approved):

vados-cosmonic:
Also, should we update Cargo.toml to force use of the newer version? Is it possible that resolution still picks the older dep?

ricochet:
I don't think so. Cargo.lock records wasmtime-wasi v44.0.2 (and siblings) and Cargo always honors the lockfile when present. For a fresh resolution (no lockfile), Cargo's default is "newest compatible"

PR #5196, opened 2026-05-22
ci: reclaim space

ricochet:
Ubuntu runners reporting out of space


vados-cosmonic (approved):
LGTM 🚀 

TIL you could do this, very interesting. Feels like this should be everywhere

PR #5200, opened 2026-05-25
feat: more flexible hosts creation in the runtime-operator helm chart

gabrielegiammatteo:
# Summary

I did minor modifications to the runtime-operator Helm chart to make the creation of hosts deployments more flexible:
- add possibilty to enable the `otel` plugin and configure the default endpoint through Helm values
- add `extraArgs` and `extraEnv` Helm values to pass custom args and environment variables to the host container

### DIFF
--- a/charts/runtime-operator/templates/runtime/deployment.yaml
+++ b/charts/runtime-operator/templates/runtime/deployment.yaml
@@ -62,6 +62,10 @@ spec:
             - name: OTEL_EXPORTER_OTLP_ENDPOINT
               value: {{ .otel.endpoint }}
             {{- end }}
+            {{- range .extraEnv }}
+            - name: {{ .name }}
+              value: {{ .value | quote }}
+            {{- end }}
           args:
             {{- if .logLevel }}
             - "--log-level={{ .logLevel }}"
@@ -98,6 +102,9 @@ spec:
             {{- end }}
             {{- if and (.otel) (.otel.enabled) }}
             - "--wasi-otel"
+            {{- end }}
+            {{- range .extraArgs }}
+            - "{{ . }}"
             {{- end }}
           imagePullPolicy: {{ $top.Values.runtime.image.pull_policy }}
           volumeMounts:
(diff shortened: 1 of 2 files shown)


charts/runtime-operator/templates/runtime/deployment.yaml
```diff
@@ -58,6 +58,18 @@ spec:
...
                   fieldPath: metadata.namespace
+            {{- if and (and (.otel) (.otel.enabled)) (.otel.endpoint) }}
+            - name: OTEL_EXPORTER_OTLP_ENDPOINT
+              value: {{ .otel.endpoint }}
+            {{- end }}
+            {{- if and (and (.otel) (.otel.enabled)) (.otel.protocol) }}
+            - name: OTEL_EXPORTER_OTLP_PROTOCOL
+              value: {{ .otel.protocol }}
+            {{- end }}
+            {{- range .extraEnv }}
```
jfleitz, line 69: If you can check out this other PR, it uses `envFrom`, which is a more k8s naming convention than `extraEnv`. https://github.com/wasmCloud/wasmCloud/pull/5219/changes#diff-89d9b218a1361f0ab75df21341a07213bab39ede2c2bb23884e8e1ad6de9c536R70-R77

Once that is merged, you could then use it instead of adding the additional range.

jfleitz:
FYI, this functionality will be supported with this [PR](https://github.com/wasmCloud/wasmCloud/pull/5246). Instead of adding an additional flag for metrics, you will now be able to pass any arguments to the host that is necessary in your `values.yaml`. This will allow for additional features to be turned off/on, which supports customized hosts too.

gabrielegiammatteo:
> FYI, this functionality will be supported with this [PR](https://github.com/wasmCloud/wasmCloud/pull/5246). Instead of adding an additional flag for metrics, you will now be able to pass any arguments to the host that is necessary in your `values.yaml`. This will allow for additional features to be turned off/on, which supports customized hosts too.

great, thanks!

PR #5204, opened 2026-05-27
test(wash-runtime): stub egress in allowed_hosts integration

ricochet:
`test_star_any_permits_all` was flaking when the real outbound call to
www.example.com exceeded the 10s per-request timeout: reqwest gave up,
the test failed, and the still-running guest then panicked writing to the
reset connection.

Install a test-local `FakeOutgoingHandler` on the `HttpServer` so
policy-allowed requests get a synthetic 200 OK without leaving the test
process. The runtime evaluates `allowed_hosts` before invoking the
handler, so denied requests still short-circuit with `HttpRequestDenied`
and the fixture still maps them to 403 — only the upstream is faked.

Tightens the allowed-case asserts from `200 || 502` to exact `200` now
that egress is deterministic.

### DIFF
--- a/crates/wash-runtime/Cargo.toml
+++ b/crates/wash-runtime/Cargo.toml
@@ -151,9 +151,14 @@ wit-component = { workspace = true, default-features = false }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["full"] }
+tracing = { workspace = true }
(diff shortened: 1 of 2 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5218, opened 2026-05-30
docs: fix stale wash project structure in contributing guide

immanuwell:
Why
`CONTRIBUTING.md` points folks to `src/main.rs`, but this repo has no root `src/`. The real CLI crate is under `crates/wash/`, so new contributors get sent to a ghost path. Kinda confusing tbh.

What
Update the project structure snippet so it matches the current workspace layout.

Repro on `main`
1. `sed -n '37,52p' CONTRIBUTING.md`
2. `test -d src || echo missing`
3. `ls crates/wash/src/main.rs`

Result: docs say `src/main.rs`, actual entrypoint is `crates/wash/src/main.rs`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,21 +46,46 @@ cargo test --bin wash
 
 ## Project Structure
 
-This repository is a workspace. The `wash` CLI lives in `crates/wash`:
+This repository is a workspace. The `wash` CLI lives in `crates/wash` and the core Wasm runtime it depends on lives in `crates/wash-runtime`:
 
 ```text
 crates/
-├── bench-tools/        # Benchmarking utilities
-├── wash/               # wash CLI crate
+├── bench-tools/            # Benchmarking utilities
+├── wash/                   # wash CLI crate
 │   └── src/
-│       ├── cli/        # CLI structs and command handling
+│       ├── cli/            # CLI structs and command handling
 │       │   ├── mod.rs
 │       │   └── <subcommand>.rs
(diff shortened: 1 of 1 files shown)


Aditya1404Sal (approved):
Thanks for pointing this out!

CONTRIBUTING.md
```diff
@@ -36,19 +36,21 @@ cargo test --bin wash
...
+├── wash/               # wash CLI crate
+│   └── src/
+│       ├── cli/        # CLI structs and command handling
+│       │   ├── mod.rs
+│       │   └── <subcommand>.rs
+│       ├── lib.rs      # Module exports
+│       ├── main.rs     # The wash binary entrypoint
+│       ├── config.rs   # Configuration management
+│       └── new.rs      # Project creation functionality
+└── wash-runtime/       # Runtime support used by wash
```
Aditya1404Sal: would be amazing if the subdirectories and main modules inside wash-runtime/src/ could be expanded and explanatory comments could highlight the context behind them.

Aditya1404Sal (changes requested):

CONTRIBUTING.md
Aditya1404Sal: since this contributing guide has a much wider scope than just the wash cli, we don't want to exclude the core runtime :)

immanuwell:
@Aditya1404Sal done

Aditya1404Sal (approved):
Thanks! LGTM🚀

PR #5219, opened 2026-05-31
runtime-operator chart: optional hostGroups passthrough fields

CharlesV1234:
Four `{{- with }}`-guarded optional fields on each `hostGroups[]` entry, letting chart users wire host-side configuration declaratively without forking the chart or post-rendering the release:

```text
hostGroups[].env (+ defensive envFrom) → host container env, appended
  after the operator-managed WASMCLOUD_HOST_* vars
hostGroups[].volumes               → pod spec.volumes
hostGroups[].volumeMounts          → host container volumeMounts
hostGroups[].ports                 → host container ports
                                     (e.g. metrics scrape port — do
                                     NOT re-list http; the chart
                                     already renders it from
                                     `.http.port`)
```

All four fields are optional. With the fields unset, the rendered Deployment is byte-identical to pre-patch:

```bash
helm template charts/runtime-operator --set global.tls.enabled=false
# vs pre-patch baseline → empty diff (verified locally)
```

The volumes / volumeMounts / ports passthroughs render unconditionally (outside the `global.tls.enabled` guard above them) so they survive a TLS-off install. The env / envFrom passthroughs render after the two operator-managed vars, so user vars can reference them but cannot collide.

Use cases the four close together:

- ConfigMap / Secret mounts for host-side config files
- secretKeyRef env for DB credentials, OIDC client secrets
- Prometheus metrics port
- Custom CA / docker auth config for OCI image pulls

No values.yaml defaults, no values.schema.json (the chart doesn't carry one today), no behaviour change for any existing user. Reviewers can verify locally with:

```bash
helm template charts/runtime-operator --set global.tls.enabled=false > /tmp/before.yaml
# apply this patch
helm template charts/runtime-operator --set global.tls.enabled=false > /tmp/after.yaml
diff /tmp/before.yaml /tmp/after.yaml
# expect: empty diff
```

### DIFF
--- a/charts/runtime-operator/templates/runtime/deployment.yaml
+++ b/charts/runtime-operator/templates/runtime/deployment.yaml
@@ -44,6 +44,12 @@ spec:
             secretName: {{ if and (.http.tls.certificate) (.http.tls.certificate.secretName) }}{{ .http.tls.certificate.secretName }}{{ else }}hostgroup-{{ .name }}-http-tls{{ end }}
         {{- end }}
         {{- end }}
+        {{- /* Optional pod volumes (e.g. ConfigMap / Secret mounts). Rendered
+               unconditionally — outside the global.tls.enabled guard above —
+               so they survive a TLS-off install. */}}
+        {{- with .volumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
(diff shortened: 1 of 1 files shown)


jfleitz (approved):
Thanks @CharlesV1234! These changes do provide VolumeMounts to the host pods, which can be very useful in some cases.

PR #5220, opened 2026-06-01
refactor(wash-runtime): build wasm test fixtures via xtask

ricochet:
Move wasm test-fixture compilation out of `crates/wash-runtime/build.rs` into a new root-level xtask crate, invoked explicitly as `cargo xtask build-fixtures`.

Building fixtures from build.rs meant `cargo test`/`cargo build` spawned nested `cargo build` invocations ("cargo building cargo"), with the jobserver/lock contention and rebuild storms that come with it. Pulling the work into an explicitly-invoked binary keeps fixture compilation off the normal build path entirely, so build.rs is now pure-Rust proto codegen.

This is currently the problem blocking our benchmark suite.

This builds directly on @dman-os's work in **#5085** (and the earlier **#4998**), which introduced the `xtask`-based "explicit fixture building" approach. Huge thanks to @dman-os for working through the design and the prototype, this PR is co-authored with them.

There's also a sibling approach in **#5089**, which keeps fixture building in a `build.rs` but moves it into a sidecar crate added as a dev/test dependency. That's a clean way to stop fixtures rebuilding on plain `cargo build`, but it *relocates* the nested-`cargo` call rather than removing it.

- Add /xtask (excluded from default-members) with a build-fixtures task; one batched `cargo build` per WASI target rather than one spawn per fixture, failing loud on any error. Drops an ambient CARGO_TARGET_DIR so staging finds the artifacts on the bench hosts.
- Reduce build.rs to proto codegen (306 -> 78 lines);
- Run `cargo xtask build-fixtures` before tests in wash.yml and codeql.yml, and in scripts/bench/run-bench.sh
- Document the flow in CONTRIBUTING, the wash-runtime README, and a new tests/fixtures/README.

I opted to start a fresh PR based on latest main primarily because:
- #5085 is closed, and both it and #5089 live on @dman-os's fork branches. Rather than force-push over someone else's PR, I opted to attribute to them as `Co-authored-by`.
- #5085 was branched off an older `main` and predates a few changes that are now in tree (the multi-file WIT-dep copy needed for split packages like `wasi-tls@0.3.0-draft`, two newer fixtures, etc.), so a fresh implementation picks those up cleanly instead of reintroducing the old behavior.

## How it differs from #5085

- Crate at root `/xtask` with `src/main.rs` layout (the cargo-xtask convention), vs `crates/xtask` with files at the crate root. This seems to be the more common convention. Documented in `matklad/cargo-xtask` `README.md` and I referenced [rust-analyzer](https://github.com/rust-lang/rust-analyzer/tree/master/xtask) as the main example.
- Dependency-light: `anyhow` + `clap` + `wit-component` + the adapter. No `tokio` runtime, `dotenv-flow`, or tracing subscriber for what is synchronous file I/O.
- One batched `cargo build` per target (2 total). We could opt to do something more complex and do 20 per-fixture spawns. This was derived from the minimal deps design choice but I'm up for revisiting later if it becomes a problem.
- Wires up `codeql.yml` and the bench scripts (which also consume the fixtures).

## Tradeoff

Fixtures no longer rebuild automatically on source change. You re-run `cargo xtask build-fixtures` after editing a fixture (an incremental is practically a no-op and finishes in ~seconds). CI and the bench scripts do this for you; the dev flow is
documented in the updated CONTRIBUTING.

## Testing

- `cargo xtask build-fixtures` stages all 20 fixtures; P3 components validate with `wasm-tools validate --features all`.
- `cargo build -p wash-runtime` (proto codegen) and all wash-runtime integration test binaries compile.
- `cargo test -p wash-runtime --features wasip3 --test integration_wasip3` passes (exercises P3 component instantiation against the v45 adapter).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,4 @@
+# `cargo xtask <task>` runs the repo task runner at /xtask. See the
+# cargo-xtask pattern: https://github.com/matklad/cargo-xtask
+[alias]
+xtask = "run --package xtask --"
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -75,14 +75,19 @@ jobs:
           # fails to expand tracing::*, anyhow::*, and std macros like
           # vec!/format!, which tanks the "calls with resolved target"
           # quality metric. We deliberately do NOT set CARGO_TARGET_DIR as
-          # a job env: that env var propagates into any child `cargo build`
-          # invoked from a build script (e.g. wash-runtime's fixture
-          # builds), redirecting their outputs and breaking their
-          # post-build artifact lookups.
+          # a job env: that env var propagates into the child `cargo build`
+          # that `cargo xtask build-fixtures` runs in the nested fixtures
+          # workspace, redirecting its output and breaking xtask's
+          # post-build artifact lookup.
           config: |
             extraction:
               rust:
                 cargo_target_dir: ${{ github.workspace }}/target
+
+      # `cargo build --all-targets` compiles the wash-runtime test/bench
+      # targets, which `include_bytes!` the wasm fixtures. Build them first.
+      - name: Build test fixtures
+        run: cargo xtask build-fixtures
 
       - name: Build workspace
         run: cargo build --workspace --all-targets --locked
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -161,6 +161,11 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Setup Rust
         uses: ./.github/actions/setup-rust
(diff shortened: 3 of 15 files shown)


ricochet:
CodeQL error is unrelated and is https://github.com/wasmCloud/wasmCloud/pull/5221

ricochet:
Triggered bench workflow with this PR and confirmed that this fixes the issue we encountered in the bench workflow: https://github.com/wasmCloud/wasmCloud/actions/runs/26785806789

```bash
PR_SHA=$(gh pr view 5220 -R wasmCloud/wasmCloud --json headRefOid -q .headRefOid)
gh workflow run bench.yml -R wasmCloud/wasmCloud --ref $PR_BRANCH -f bench=http_invoke -f ref=$PR_SHA
```

Aditya1404Sal (approved):

PR #5222, opened 2026-06-01
refactor(wash-runtime): harden p3 sockets

ricochet:
Address review feedback from #5022 (tracked in #5028) and bring the
p3 socket imp in line with the upstream wasmtime wasi-sockets.

- name lookup: resolve domains with tokio::net::lookup_host instead of
  spawn_blocking + to_socket_addrs (lookup_host already runs getaddrinfo
  on a blocking task), parse the host before the capability check, and
  convert addresses via the From<IpAddr> impl wasmtime provides. All
  matching the upstream implementation
- tcp: return an owned Arc from tcp_listener_arc and listen_p3 so callers
  drop redundant clone()/unwrap() panic paths
- tcp: make the loopback TcpConn send/receive halves Option so each stream
  is taken exactly once; a second take reports InvalidState instead of
  swapping in a throwaway unbounded channel (receive) or minting a fresh
  semaphore (send), matching the take-once semantics of network sockets
- document why p3_error_code_from_util maps Unknown to Other(None)
- doc cleanups on the p3 TcpSocket helpers
- add a unit test covering loopback take-once behavior.

Closes #5028

### DIFF
--- a/crates/wash-runtime/src/sockets/host_ip_name_lookup_p3.rs
+++ b/crates/wash-runtime/src/sockets/host_ip_name_lookup_p3.rs
@@ -13,47 +13,37 @@ impl HostWithStore for WasiSockets {
         store: &Accessor<U, Self>,
         name: String,
     ) -> wasmtime::Result<Result<Vec<types::IpAddress>, ErrorCode>> {
-        let allow_lookup =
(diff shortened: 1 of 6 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5223, opened 2026-06-01
Fix In-memory blobstore default

### LINKED ISSUE #5126: BUG: In-memory dev plugin for blobstore
In-memory blobstore is currently unusable in wash dev. `InMemoryBlobstore` derives `Default`, which leaves `max_object_size` at 0, and any non-zero write traps with "write beyond capacity of MemoryOutputPipe".

ricochet:
InMemoryBlobstore derived Default, leaving max_object_size at 0, so the in-memory blobstore used by wash dev made a zero-capacity MemoryOutputPipe and any non-empty write trapped with "write beyond capacity of MemoryOutputPipe". Replaced the derive with a manual Default that delegates to `new(None)` (1MB), matching the explicit constructor.

Verified with wash dev on examples/blobby: POST/GET round-trips now succeed (28B, 256KB) where the released wash traps.

Fixes #5126

### DIFF
--- a/crates/wash-runtime/src/plugin/wasi_blobstore/in_memory.rs
+++ b/crates/wash-runtime/src/plugin/wasi_blobstore/in_memory.rs
@@ -84,12 +84,18 @@ pub struct StreamObjectNamesHandle {
 }
 
 /// Memory-based blobstore plugin
-#[derive(Clone, Default)]
+#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM :rocket:

NIT: 1MB may be a bit low (5MiB/10MiB) IMO is more reasonable, and the value should likely be a constant, or even tunable.

PR #5224, opened 2026-06-01
chore: test dynamic router removes aliases

### LINKED ISSUE #5034: [Tests] Add wash-runtime integration test for host aliases in DynamicRouter
## Summary
Add test for DynamicRoute host-alias change introduced in #5027 . The change adds support for `host-aliases` — a comma-separated list of additional hostnames (typically Kubernetes Service DNS names) that are registered alongside the primary `host` config value. This allows cluster-internal callers to reach a workload via any of its Service DNS variants (e.g. `my-svc`, `my-svc.default`) registered by the runtime-operator, but can also be used for additional aliases as well.

Currently, there is no integration test in the `wash-runtime` testcontainer that exercises this feature. The existing HTTP integration tests (`integration_http_counter.rs`, `integration_http_allowed_hosts.rs`, etc.) only configure a single `host` value in the `wasi:http/incoming-handler` interface config and route requests using `DevRouter`, which bypasses host-header-based routing entirely.

## What needs to be tested

1. **Alias registration**: When a workload is started with both a `host` and `host-aliases` config on the `wasi:http/incoming-handler` interface, the `DynamicRouter` should route incoming requests matching any of the aliases to that workload.

2. **Routing by Host header**: Sending HTTP requests with `Host` headers set to each alias should all resolve to the same workload and return valid responses.

3. **Alias cleanup on unbind**: When a workload is stopped, all aliases (not just the primary hostname) should be removed from the router. A subsequent request to any former alias should fail with "no workload bound to host header".

4. **Multiple workloads with distinct aliases**: Two workloads with different primary hosts and different aliases should be independently routable. A request to workload A's alias should never reach workload B.

5. **Invalid alias handling**: Aliases that are not valid RFC 1123 hostnames should be silently filtered out and not registered in the router.

## Approach

The test should use `DynamicRouter` with `HttpServer`, start a workload that has `host-aliases` configured, and make HTTP requests using each alias as the `Host` header. This mirrors the real Kubernetes path where the operator injects `host-aliases` into the workload spec via `injectServiceDNSAliases`.

## Acceptance criteria

- [ ] New integration test file `integration_http_host_aliases.rs` (or added to existing HTTP test file)
- [ ] Test uses `DynamicRouter` for host-header-based routing
- [ ] Verifies requests routed correctly via primary host and all aliases
- [ ] Verifies aliases are cleaned up when workload is unbound
- [ ] Verifies invalid aliases are ignored

ricochet:
Final test for complete coverage, closes #5034

### DIFF
--- a/crates/wash-runtime/tests/integration_router_dynamic.rs
+++ b/crates/wash-runtime/tests/integration_router_dynamic.rs
@@ -17,6 +17,8 @@
 //! - HTTP/1.0 request without a Host header returns 400 (RouteError::MissingHost).
 //! - Invalid hostnames in host-aliases are silently filtered; only valid ones route.
 //! - Two workloads bound to the same host both serve requests (HashSet routing).
+//! - Stopping a workload removes all of its routes — primary host and every
+//!   alias — so requests to any former alias return 404 after unbind.
 
 #![allow(clippy::unwrap_used, clippy::expect_used)]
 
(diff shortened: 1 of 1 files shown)


crates/wash-runtime/tests/integration_router_dynamic.rs
```diff
@@ -355,3 +357,41 @@ async fn test_dynamic_router_unknown_host_returns_404() -> Result<()> {
...
 }
+
+/// Stopping a workload must remove *all* of its routes — the primary host and
+/// every alias — not just the primary hostname. After unbind, a request to any
+/// former alias must map to `RouteError::NoWorkloadForHost` (404).
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_dynamic_router_aliases_removed_on_unbind() -> Result<()> {
+    let (addr, host) = start_host_with_dynamic_router("127.0.0.1:0").await?;
+
+    let req = http_counter_request("primary.local", Some("alias-one.local,alias-two.local"));
```
Aditya1404Sal:
```suggestion

let primary = "primary.local";
let aliases = ["alias-one.local", "alias-two.local"];

let req = http_counter_request(
    primary,
    Some(&aliases.join(",")),
);

```

crates/wash-runtime/tests/integration_router_dynamic.rs
```diff
@@ -355,3 +357,41 @@ async fn test_dynamic_router_unknown_host_returns_404() -> Result<()> {
...
+    let client = reqwest::Client::new();
+    let hostnames = ["primary.local", "alias-one.local", "alias-two.local"];
+
+    // Sanity check: the primary host and both aliases route while bound.
+    for hostname in hostnames {
+        assert!(
+            get_status(&client, addr, hostname).await?.is_success(),
+            "{hostname} should route while the workload is bound"
+        );
+    }
```
Aditya1404Sal:
```suggestion
   for hostname in std::iter::once(primary).chain(aliases) {
    assert!(
        get_status(&client, addr, hostname).await?.is_success(),
        "{hostname} should route while the workload is bound"
    );
}
```

crates/wash-runtime/tests/integration_router_dynamic.rs
```diff
@@ -355,3 +357,41 @@ async fn test_dynamic_router_unknown_host_returns_404() -> Result<()> {
...
+
+    // After unbind, every former route — primary and aliases — must 404.
+    for hostname in hostnames {
+        let status = get_status(&client, addr, hostname).await?;
+        assert_eq!(
+            status,
+            reqwest::StatusCode::NOT_FOUND,
+            "{hostname} must be removed on unbind and return 404, got {status}"
+        );
+    }
```
Aditya1404Sal:
```suggestion
   for hostname in std::iter::once(primary).chain(aliases) {
    let status = get_status(&client, addr, hostname).await?;

    assert_eq!(
        status,
        reqwest::StatusCode::NOT_FOUND,
        "{hostname} must be removed on unbind and return 404, got {status}"
    );
}
```

Aditya1404Sal (dismissed):
Just a couple of non-blocking nits from me, otherwise looks good.

PR #5225, opened 2026-06-02
chore(lint): pin Rust to 1.96.0 and make clippy --all-targets clean

ricochet:
Pin rust-toolchain.toml to an explicit stable (1.96.0) instead of a
floating "stable", so clippy is deterministic.

MSRV (1.91.0) is unchanged.

Add clippy.toml with allow-{unwrap,expect,panic,indexing-slicing}-in-tests
so test code passes the workspace's denied restriction lints under
--all-targets, without weakening the production `src/` lint surface (CI's
src-only clippy is unaffected).

Fix the genuine lints the newer clippy flags across src, tests, and benches:
uninlined_format_args, needless_collect, manual is_none, char::is_digit ->
is_ascii_digit, nonminimal_bool/op_ref/single_match, and doc-list
formatting. Refactor helper-level unwraps where clean (Version::new, ?,
poison-tolerant lock, buf.get(..n)) and drop per-file test allows now
covered by clippy.toml; keep them only where module-level helpers still
need them, and make tests/common self-contained.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,9 @@
+# Test code is allowed to unwrap/expect/panic/index freely. The workspace
+# denies these restriction lints (see [workspace.lints.clippy] in Cargo.toml)
+# so production `src/` stays panic-free, but failing fast is fine in tests.
+# These knobs only relax code clippy identifies as a test, so `cargo clippy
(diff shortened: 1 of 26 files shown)


Aditya1404Sal (approved):
LGTM

PR #5229, opened 2026-06-02
fix: create bucket on open

### LINKED ISSUE #5228: `wasi:keyvalue` NATS backend `open` does not auto-create the JetStream KV bucket, unlike the blobstore backend
The NATS backends for `wasi:keyvalue` and `wasi:blobstore` are inconsistent about bucket creation:

* **blobstore**: `create-container` auto-provisions the JetStream object store if it doesn't exist
* **keyvalue**: `open` is get-only and fails if the JetStream KV bucket doesn't already exists. There is no way for a guest to create a KV bucket.

Furthermore, the `in_memory` and `filesystem` implementations both do create the bucket if it doesn't exist.

The expected behaviour would be one of these:
1. Auto-create on open (like `in_memory` and `filesystem`)
2. Add `create_key_value` just like the `wasi:blobstore` implementation

Mees-Molenaar:
Makes the interface consistent with the in_memory and filesystem implementations

Closes #5228

### DIFF
--- a/crates/wash-runtime/src/plugin/wasi_keyvalue/nats.rs
+++ b/crates/wash-runtime/src/plugin/wasi_keyvalue/nats.rs
@@ -75,6 +75,32 @@ impl NatsKeyValue {
         )];
         self.metrics.operations_total.add(1, &attributes);
     }
+
+    /// Get the JetStream KV bucket, creating it if it doesn't exist yet.
+    async fn get_or_create_bucket(
+        &self,
+        identifier: &str,
+    ) -> Result<async_nats::jetstream::kv::Store, String> {
+        match self.client.get_key_value(identifier).await {
+            Ok(kv) => {
+                tracing::debug!("Opened existing bucket in JetStream");
+                Ok(kv)
+            }
+            Err(e) => {
+                tracing::debug!("Bucket not found in JetStream({identifier}): {e}; creating it");
+                self.client
+                    .create_key_value(async_nats::jetstream::kv::Config {
+                        bucket: identifier.to_string(),
+                        ..Default::default()
+                    })
+                    .await
+                    .map_err(|e| {
+                        tracing::error!("Failed to create bucket in JetStream({identifier}): {e}");
+                        "failed to create keyvalue bucket in JetStream".to_string()
+                    })
+            }
+        }
+    }
 }
 
 // Implementation for the store interface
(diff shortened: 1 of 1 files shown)


crates/wash-runtime/src/plugin/wasi_keyvalue/nats.rs
```diff
@@ -75,6 +75,32 @@ impl NatsKeyValue {
...
+    async fn get_or_create_bucket(
+        &self,
+        identifier: &str,
+    ) -> Result<async_nats::jetstream::kv::Store, String> {
+        match self.client.get_key_value(identifier).await {
+            Ok(kv) => {
+                tracing::debug!("Opened existing bucket in JetStream");
+                Ok(kv)
+            }
+            Err(e) => {
```
jfleitz: Hello @Mees-Molenaar , I like the idea of making this consistent with the other backends, but I am concerned on creating the bucket if any error is returned. What if the bucket exists, and you don't have permission or there is a read error? I think the error returned should be checked if it is a "bucket not found" type error. 

Other thought is on race condition...If you have several requests coming in, the first request creates the bucket and then other ones would get an error trying to create the bucket?

Mees-Molenaar: Thanks for the feedback. I have digged into your concerns and simplified the code based on the findings.
For your first concern, indeed trying to create the bucket on any error (permission, connection, Jetstream-disabled) was incorrect.

As for your second concern (race condition), I verified that the `create_key_value` is idempotent for an identical bucket config. If you try to create a bucket that already exists, it returns the existing bucket rather than an error (verified also in an integration test). So concurrent `opens()` will all create safely.

As for the permissions, the current setup relies on the fact that the Host ownes the NATS connection and shares the permissions across all workloads. I documented inline that if per-workload credentials are ever introduced, it is clear that this should be handled accordingly.

Let me know if this addresses your concerns or if you'd like me to take a different approach.

crates/wash-runtime/src/plugin/wasi_keyvalue/nats.rs
```diff
@@ -75,6 +75,41 @@ impl NatsKeyValue {
...
+        self.client
+            .create_key_value(async_nats::jetstream::kv::Config {
+                bucket: identifier.to_string(),
+                ..Default::default()
+            })
+            .await
+            .map_err(|e| {
+                tracing::error!("Failed to open keyvalue bucket in JetStream({identifier}): {e}");
+                "failed to open keyvalue bucket in JetStream".to_string()
+            })
```
Aditya1404Sal, line 115:
```suggestion
           .map_err(|e| {
    tracing::error!(
        error = ?e,
        bucket = %identifier,
        "Failed to open keyvalue bucket in JetStream"
    );

    format!("failed to open keyvalue bucket in JetStream({identifier}): {e}")
})
```

Nit: the doc comment mentions that genuine failures are surfaced rather than masked. 

Would it make sense to preserve the underlying error in the returned Err as well (like, include {e} in the message)? We currently log the original error via tracing, but callers only receive a generic string.

Mees-Molenaar: Adjusted it

Aditya1404Sal (approved):
Looks good to me 🚀

PR #5234, opened 2026-06-03
ci(bench): wait for host load to settle

ricochet:
The release matrix runs http_invoke and gungraun back-to-back on the
single bench host. The pre-flight 1-min loadavg guard (MAX_LOAD1 = 1.0)
fired on the *previous* bench's residual load — the 1-min average decays
over ~60s. The second bench failed before it started rather than
because the host was actually busy.

Poll loadavg for up to 240s and pass as soon as it drops below the
threshold; a genuinely busy host never settles and still fails.

### DIFF
--- a/.github/scripts/bench-preflight.mjs
+++ b/.github/scripts/bench-preflight.mjs
@@ -16,12 +16,18 @@
 
 import { execFileSync, spawnSync } from 'node:child_process';
 import { existsSync, mkdirSync, readFileSync, statfsSync } from 'node:fs';
-import { hostname } from 'node:os';
+import { hostname, loadavg, platform } from 'node:os';
 
 const EXPECTED_NPROC = 6;
 const EXPECTED_GOVERNOR = 'performance';
 const MIN_FREE_BYTES = 5 * 1024 ** 3; // 5 GiB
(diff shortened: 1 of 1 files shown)


.github/scripts/bench-preflight.mjs
```diff
@@ -43,6 +49,15 @@ function runStdout(cmd, args = []) {
   return execFileSync(cmd, args, { encoding: 'utf8' }).trim();
 }
 
+// Block the synchronous pre-flight for `ms` without spawning a subprocess.
+function sleepSync(ms) {
+  Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, ms);
+}
+
+function readLoad1() {
+  return parseFloat(readTrim('/proc/loadavg').split(/\s+/)[0]);
```
vados-cosmonic: So this works, but TIL node has [`os.loadavg()`](https://nodejs.org/api/os.html#osloadavg)

We also might want to throw real hard if the OS is detected to be windows

vados-cosmonic (dismissed):
LGTM 🚀 

Looks like it works, one nit about using a builtin I found

PR #5235, opened 2026-06-03
Have container health use NATs connectivity status

jfleitz:
- Adds a Health port to the runtime-operator deployment with a Health and Liveness probe configured
- Health of runtime-operator uses the underlying NATS connection status. This allows for automatic recovery if/when the NATS instance is unreachable (patch rollout for example).
- Adds an integration test to ensure that the runtime-operator can recover from a NATS restart.

Tested with a local Kind cluster using WasmCloud 2.3.0 host, and locally-build runtime-operator.

### DIFF
--- a/runtime-operator/cmd/main_test.go
+++ b/runtime-operator/cmd/main_test.go
@@ -1,19 +1,3 @@
-/*
-Copyright 2024.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


runtime-operator/cmd/main_test.go
```diff
@@ -0,0 +1,91 @@
+/*
+Copyright 2024.
```
ricochet: We haven't been adding apache headers

ricochet (approved):

PR #5236, opened 2026-06-05
feat: build wash for s390x-unknown-linux-gnu

LiamRandall:
Quick build for IBM s390 series processors.

Tested on an Ubuntu host, successfully ran complex app.


vados-cosmonic (approved):
This looks good to me! We should probably set up some basic testing for the binary but this is a good start.

DCO needs fixing!

Issue #4942, opened 2026-01-13
[wash-runtime] Crate API Refinements


lxfontes:
In wasmCloud/wash#218  we can see a few patterns emerging with the amount of code copying across plugins & inconsistencies in Ctx field vs method access.

This issue is about:
- Identifying any helper function / api refactor that would improve quality of life for wash-runtime crate consumers
- Ensure expected behaviour correctness
- Proper data ownership ( references vs owned )
- Clarify  identifiers ( workload_id, component_id, context_id, etc )

For api refactor: all plugins have code like this  to find their `hostInterface` entry in workload / component binding:

```
 // Find the "wasi:config/store" interface, if present
        let Some(interface) = interfaces.iter().find(|i| {
            i.namespace == "wasi" && i.package == "config" && i.interfaces.contains("store")
        }) else {
            // Log a warning if the requested interfaces are not wasi:config/store
            tracing::warn!(
                "WasiConfig plugin requested for non-wasi:config/store interface(s): {:?}",
                interfaces
            );
            return Ok(());
        };
```

This could be an `Option<WitInterface>` passed to the function from the runtime.

For data ownership: we have one `WorkloadMetadata` per `WorkloadComponent` when it should be one `WorkloadMetadata` per `Workload` and references in each `WorkloadComponent`.

These are a couple examples of the refinements expected to be addressed by this issue.

**The boundary**

Within `wash-runtime` we have 3 APIs:
- `engine` API: low level API. depends only on wasmtime.
- `host` API: entrypoint for most crate consumers, uses `engine`
- `washlet` API: tightly coupled to `runtime-operator` & protobuf, uses `host`. End User API / Behaviour contract.

Changes to `engine` are 👍 .
Changes to `host` are 👍  as long as we are not introducing significant downstream changes to `washlet` or **any** change to protobuf api.
Changes to `washlet` are not in scope due to blast radius.

vados-cosmonic:
Closing this out -- we've had quite a few changes and the codebase has changed quite a bit (also given that some work was merged already), I think more targeted issues here will be better going forward.

PR #5237, opened 2026-06-05
refactor(wash-runtime): improve host plugin API ergonomics

if0ne:
Related #4942 

Hi there!

Refactors the `wash-runtime` host plugin API to reduce boilerplate at plugin call sites and give plugin authors a cleaner, less error-prone interface

1. `Ctx::try_get_plugin` / `get_plugin`

Replaced the old fallible accessor:

```rust
pub fn get_plugin<T: HostPlugin>(&self, id: &str) -> Option<Arc<T>>
```

with a pair:

- `try_get_plugin::<T>(id) -> wasmtime::Result<Arc<T>>` - returns a descriptive error (plugin missing / type mismatch) that composes with `?`.
- `get_plugin::<T>(id) -> Arc<T>` - infallible convenience wrapper that panics on miss.

This collapses the repeated `let Some(plugin) = self.get_plugin(...) else { return Ok(Err(...)) }` guard blocks across all plugins into a single `let plugin = self.try_get_plugin::<T>(ID)?;`, removing ~200 lines of duplicated error-handling.

2. **`WitInterfaces<'a>`** (`plugin/mod.rs`)

Introduces a borrowed wrapper around `&HashSet<WitInterface>` with `get()` / `contains()` lookup helpers, replacing the owned `HashSet<WitInterface>` previously passed by value to the `on_workload_bind`, `on_workload_item_bind`, and `on_workload_unbind` trait methods. Avoids cloning the interface set per plugin and gives binding code a readable `interfaces.contains("wasi", "blobstore", &[])` API.

3. **`Ctx.id: Arc<str>`**

Changed from an owned string to `Arc<str>` for cheaper cloning, consistent with the other id fields on `Ctx`.

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -41,6 +41,10 @@ type BoundPluginWithInterfaces = (
     Vec<String>,
 );
 
+/// Shared slot holding the most recently instantiated component, keyed by its id.
+/// Used while resolving inter-component imports.
+type SharedInstanceSlot = Arc<RwLock<Option<(Arc<str>, Instance)>>>;
+
 /// Metadata associated with components and services within a workload.
 #[derive(Clone)]
 pub struct WorkloadMetadata {
@@ -814,7 +818,7 @@ impl ResolvedWorkload {
         let ty = component.component_type();
         let imports: Vec<_> = ty.imports(component.engine()).collect();
 
-        let instance: Arc<RwLock<Option<(Arc<str>, Instance)>>> = Arc::default();
+        let instance: SharedInstanceSlot = Arc::default();
         for (import_name, import_item) in imports.into_iter() {
             match import_item {
                 ComponentItem::ComponentInstance(import_instance_ty) => {
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -814,7 +814,7 @@ impl ResolvedWorkload {
         let ty = component.component_type();
         let imports: Vec<_> = ty.imports(component.engine()).collect();
 
-        let instance: Arc<RwLock<Option<(String, Instance)>>> = Arc::default();
+        let instance: Arc<RwLock<Option<(Arc<str>, Instance)>>> = Arc::default();
         for (import_name, import_item) in imports.into_iter() {
             match import_item {
                 ComponentItem::ComponentInstance(import_instance_ty) => {
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -8,7 +8,10 @@ use std::{
     time::Duration,
 };
 
-use crate::sockets::{self, SocketAddrUse, loopback};
+use crate::{
+    plugin::WitInterfaces,
+    sockets::{self, SocketAddrUse, loopback},
+};
 use anyhow::{Context as _, bail, ensure};
 use tokio::{sync::RwLock, task::JoinHandle, time::timeout};
 use tracing::{Instrument, debug, error, info, instrument, trace, warn};
@@ -1274,7 +1277,15 @@ impl ResolvedWorkload {
                         .cloned()
                         .collect::<std::collections::HashSet<_>>();
 
-                    if let Err(e) = plugin.on_workload_unbind(self.id(), bound_interfaces).await {
+                    if let Err(e) = plugin
+                        .on_workload_unbind(
+                            self.id(),
+                            WitInterfaces {
+                                inner: &bound_interfaces,
+                            },
+                        )
+                        .await
+                    {
                         warn!(
                             plugin_id,
                             component_id = component.id(),
@@ -1530,7 +1541,12 @@ impl UnresolvedWorkload {
 
                 // Call on_workload_bind with the workload and all matched interfaces
                 if let Err(e) = p
-                    .on_workload_bind(self, plugin_matched_interfaces.clone())
+                    .on_workload_bind(
+                        self,
+                        WitInterfaces {
+                            inner: &plugin_matched_interfaces,
+                        },
+                    )
                     .instrument(bind_span)
                     .await
                 {
@@ -1548,7 +1564,12 @@ impl UnresolvedWorkload {
                             "calling on_workload_unbind for cleanup after bind failure"
                         );
                         if let Err(cleanup_err) = bound_plugin
-                            .on_workload_unbind(self.id(), bound_interfaces.clone())
+                            .on_workload_unbind(
+                                self.id(),
+                                WitInterfaces {
+                                    inner: bound_interfaces,
+                                },
+                            )
                             .await
                         {
                             warn!(
@@ -1592,7 +1613,12 @@ impl UnresolvedWorkload {
                         plugin_id = plugin_id,
                     );
                     if let Err(e) = p
-                        .on_workload_item_bind(&mut workload_item, matching_interfaces.clone())
+                        .on_workload_item_bind(
+                            &mut workload_item,
+                            WitInterfaces {
+                                inner: &matching_interfaces,
+                            },
+                        )
                         .instrument(item_bind_span)
                         .await
                     {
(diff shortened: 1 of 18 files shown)


crates/wash-runtime/src/plugin/mod.rs
```diff
@@ -55,6 +55,33 @@ pub mod wasmcloud_messaging;
 #[cfg(all(feature = "wasi-webgpu", not(target_os = "windows")))]
 pub mod wasi_webgpu;
 
+/// A wrapper around a [`std::collections::HashSet`] of [`crate::wit::WitInterface`] that provides convenience methods for looking up interfaces by namespace, package, and set of interfaces.
+#[derive(Debug)]
+pub struct WitInterfaces<'a> {
+    pub inner: &'a std::collections::HashSet<crate::wit::WitInterface>,
+}
+
+impl<'a> WitInterfaces<'a> {
```
ricochet: nit: I think it'd be better to for inner to be private since the reason it was added was to have folks interact with lookup helpers, e.g. `get`/`contains`.

```suggestion
    inner: &'a std::collections::HashSet<crate::wit::WitInterface>,
}

impl<'a> WitInterfaces<'a> {
    /// Wraps a borrowed set of [`crate::wit::WitInterface`]s.
    pub fn new(inner: &'a std::collections::HashSet<crate::wit::WitInterface>) -> Self {
        Self { inner }
    }

    /// Returns an iterator over the wrapped [`crate::wit::WitInterface`]s.
    pub fn iter(&self) -> impl Iterator<Item = &crate::wit::WitInterface> {
        self.inner.iter()
    }

```

crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
```diff
@@ -738,14 +726,11 @@ impl HostPlugin for FilesystemBlobstore {
...
         &self,
         component_handle: &mut WorkloadItem<'a>,
-        interfaces: std::collections::HashSet<crate::wit::WitInterface>,
+        interfaces: WitInterfaces<'_>,
     ) -> anyhow::Result<()> {
-        let Some(_interface) = interfaces
-            .iter()
-            .find(|i| i.namespace == "wasi" && i.package == "blobstore")
-        else {
+        if !interfaces.contains("wasi", "package", &[]) {
```
ricochet:
```suggestion
        if !interfaces.contains("wasi", "blobstore", &[]) {
```

crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
```diff
@@ -773,14 +758,11 @@ impl HostPlugin for FilesystemBlobstore {
...
         &self,
         workload: &crate::engine::workload::UnresolvedWorkload,
-        host_interfaces: std::collections::HashSet<crate::wit::WitInterface>,
+        host_interfaces: WitInterfaces<'_>,
     ) -> anyhow::Result<()> {
-        let Some(_interface) = host_interfaces
-            .iter()
-            .find(|i| i.namespace == "wasi" && i.package == "blobstore")
-        else {
+        if !host_interfaces.contains("wasi", "package", &[]) {
```
ricochet:
```suggestion
        if !host_interfaces.contains("wasi", "blobstore", &[]) {
```

crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
```diff
@@ -470,19 +454,17 @@ impl HostPlugin for WasmcloudPostgres {
...
     ) -> anyhow::Result<()> {
-        let database = match interfaces
-            .iter()
-            .find(|i| i.namespace == "wasmcloud" && i.package == "postgres")
-        {
-            Some(i) => match i.config.get("database").cloned() {
-                Some(db) => db,
-                None => bail!("wasmcloud:postgres requires a 'database' config parameter"),
-            },
+        let database = match interfaces.get("wasmcloud", "postgres", &["database"]) {
```
ricochet:
```suggestion
        let database = match interfaces.get("wasmcloud", "postgres", &[]) {
```

crates/wash-runtime/src/engine/workload.rs
ricochet, line 2288: Given the package mismatch for filesystem blobstore and postgres, we obviously are missing test coverage for those. A contrived way to catch that type of regression is by rebinding:


```suggestion
    /// Every built-in plugin must actually bind the interface it advertises in its
    /// `world()`. The bind hooks pick which interfaces to serve by matching
    /// namespace / package / interface-name strings against the requested set, so a
    /// wrong literal (a package typo, or a config key passed where an interface name
    /// is expected) makes the plugin silently no-op instead of registering its host
    /// functions. This guards that whole class of regression: feed each plugin an
    /// interface it is designed to serve, then assert the first bind registers host
    /// functions. Because a second bind of the same interface on the same linker
    /// conflicts, a plugin that no-op'd (never touched the linker) is caught by the
    /// second bind wrongly succeeding.
    #[cfg(all(
        feature = "wasmcloud-postgres",
        feature = "wasi-blobstore",
        feature = "wasi-keyvalue"
    ))]
    #[tokio::test]
    async fn builtin_plugins_bind_their_declared_interfaces() {
        use crate::plugin::wasi_blobstore::{FilesystemBlobstore, InMemoryBlobstore};
        use crate::plugin::wasi_keyvalue::InMemoryKeyValue;
        use crate::plugin::wasmcloud_postgres::WasmcloudPostgres;

        // wasmcloud:postgres carries the target database as interface config, not as
        // an interface name; the plugin reads it after matching on package.
        let mut postgres_iface =
            WitInterface::from("wasmcloud:postgres/types,query,prepared@0.1.1-draft");
        postgres_iface
            .config
            .insert("database".to_string(), "testdb".to_string());

        let blobstore_iface =
            WitInterface::from("wasi:blobstore/blobstore,container,types@0.2.0-draft");
        let keyvalue_iface = WitInterface::from("wasi:keyvalue/store,atomics,batch@0.2.0-draft");

        let cases: Vec<(Arc<dyn HostPlugin>, WitInterface)> = vec![
            (
                Arc::new(WasmcloudPostgres::new("postgres://user:pass@localhost:5432/db").unwrap())
                    as Arc<dyn HostPlugin>,
                postgres_iface,
            ),
            (
                Arc::new(FilesystemBlobstore::new(std::env::temp_dir())) as Arc<dyn HostPlugin>,
                blobstore_iface.clone(),
            ),
            (
                Arc::new(InMemoryBlobstore::new(None)) as Arc<dyn HostPlugin>,
                blobstore_iface,
            ),
            (
                Arc::new(InMemoryKeyValue::new()) as Arc<dyn HostPlugin>,
                keyvalue_iface,
            ),
        ];

        // A minimal empty component is enough: the bind hooks only mutate the linker.
        let build_component = || {
            let engine = wasmtime::Engine::default();
            let linker = Linker::new(&engine);
            // Minimal empty component: magic "\0asm" + component version + layer.
            let component =
                Component::new(&engine, [0x00, 0x61, 0x73, 0x6d, 0x0d, 0x00, 0x01, 0x00]).unwrap();
            WorkloadComponent::new(
                "workload-bind".to_string(),
                "test-workload-bind".to_string(),
                "test-namespace".to_string(),
                "test-component".to_string(),
                component,
                linker,
                Vec::new(),
                LocalResources::default(),
                Arc::default(),
                #[cfg(feature = "wasip3")]
                false,
            )
        };

        for (plugin, iface) in cases {
            let id = plugin.id();
            let set: HashSet<WitInterface> = std::iter::once(iface).collect();

            let mut component = build_component();
            let mut item = WorkloadItem::Component(&mut component);

            plugin
                .on_workload_item_bind(&mut item, WitInterfaces::new(&set))
                .await
                .unwrap_or_else(|e| {
                    panic!("plugin `{id}` failed to bind its declared interface: {e}")
                });

            // The first bind registered the plugin's host functions, so binding the
            // same interface again on the same linker must conflict. If the plugin
            // silently no-op'd (e.g. a matcher typo), it never touched the linker and
            // this second bind wrongly succeeds.
            let rebind = plugin
                .on_workload_item_bind(&mut item, WitInterfaces::new(&set))
                .await;
            assert!(
                rebind.is_err(),
                "plugin `{id}` silently no-op'd on its declared interface instead of \
                 registering host functions"
            );
        }
    }

    /// Tests basic plugin binding with one plugin and one component.
```

ricochet (changes requested):
This is a nice improvement

ricochet (approved):

PR #5241, opened 2026-06-08
fix(wash): scope new.command to template config

immanuwell:
`wash new` was reading merged user and project config when it decided whether to run `new.command`.
That means a user-level `new.command` can leak into any scaffold, even if the template has no `.wash/config` at all. Kinda sneaky.

This keeps `new.command` scoped to the scaffolded template config only. Template hooks still work, global hooks stop bleeding in.

Repro on `main`:
1. Put this in a user config file:
   `new.command: sh -c 'echo fired >/tmp/wash-new-hook'`
2. Make a git template repo with no `.wash/config`.
3. Run `wash new <repo> --name demo` and accept the prompt.
4. `/tmp/wash-new-hook` gets created, even though the template never asked for it.

Checks:
`cargo test -p wash cli::new -- --nocapture`

Related area: #5154

### DIFF
--- a/crates/wash/src/cli/new.rs
+++ b/crates/wash/src/cli/new.rs
@@ -11,7 +11,7 @@ use tracing::{info, instrument};
 
 use crate::{
     cli::{CliCommand, CliContext, CommandOutput},
-    config::{Config, load_config},
+    config::{load_config_from_file, locate_project_config},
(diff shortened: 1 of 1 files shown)


ricochet (approved):
This is a good catch. Initially I debated if it was really a bug, but the deciding factor for me is what `new.command` is. It's a template-authored setup hook, and not a user preference.

PR #5243, opened 2026-06-09
multiple subscription workloads

ricochet:
Builds on the work in #5242. Marking as draft until #5242 lands.

Make `wash dev` hosts support workloads with multiple components and messaging subscriptions.

Both the In-memory messaging plugin and NATS plugin now reads: `subscriptions` from per-component config (host-interface fallback) and tracks handlers by world export.
Subscription parsing is shared between both backends.

We do this by adding a single `dev.data_nats_url` that points blobstore, keyvalue, and messaging at NATS at once (mirroring the production host's `--data-nats-url`), with the existing per-plugin URLs kept as overrides (although I'm leaning towwards deprecating these). Note this also gives wash dev a blobstore NATS backend for the first time.

The `templates/http-api-with-distributed-workloads` template is extended to exercise this with two subscriptions.

There is one runtime behavior change that I'd appreciate extra eyes on... we now allow multiple components to export the same host-invoked interface. Workload linking previously rejected any interface exported by more than one component, which blocked workloads with multiple components each with messaging handlers in one workload (`another component already implements the interface`). The uniqueness check is now deferred to the import side: duplicate exports are fine for host-invoked interfaces like `wasmcloud:messaging/handler` and `wasi:http/incoming-handler` that no component imports; genuinely ambiguous component-to-component imports still error.

Another behavior change for the dev host that I think is correct, but worth calling out: when `data_nats_url` points at a NATS without JetStream, messaging works (core subjects) but blobstore and keyvalue ops fail at runtime. This matches `wash host` and k8s, but a potential footgun. It's very edge case. The defaults are in-memory so the scenario where it might surprise folks is if they didn't set a file path or some other config and therefore are getting in-memory plugins, then they set a data_nats_url thinking they only want this for messaging.

What didn't change:  Single-component dev workloads behave as they did before. No subscriptions then they will receive all messages.

This change is broken down into several commits to make it easier to review:
- **feat(dev): nats plugin with multi-component subs**
- **fix(runtime): allow multiple components to export the same host-invoked interface**
- **feat(dev): subject-based routing for in-memory messaging**
- **feat(templates): add task-reverse worker and branded UI**

### DIFF
--- a/crates/wash/src/config.rs
+++ b/crates/wash/src/config.rs
@@ -422,6 +422,16 @@ pub struct DevConfig {
     #[serde(default)]
     pub wasi_webgpu: bool,
 
+    /// Shared NATS connection URL for the data-plane plugins
+    /// blobstore, keyvalue, and messaging. Mirroring the `wash host`
+    /// with `--data-nats-url`. When set, all three use NATS unless a per-plugin
+    /// URL below overrides it (e.g. `wasi_keyvalue_redis_url`,
+    /// `wasi_keyvalue_path`, `wasi_keyvalue_nats_url`,
+    /// `wasmcloud_messaging_nats_url`, `wasi_blobstore_path`).
+    /// Example: nats://127.0.0.1:4222
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub data_nats_url: Option<String>,
+
     /// Optional Redis connection URL for the WASI keyvalue plugin.
     /// Example: redis://127.0.0.1:6379
     /// When set, takes precedence over wasi_keyvalue_path.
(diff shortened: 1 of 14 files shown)


crates/wash/src/config.rs
```diff
@@ -400,12 +442,21 @@ pub struct DevConfig {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub wasi_keyvalue_nats_url: Option<String>,
 
+    /// Optional NATS connection URL for the wasmcloud:messaging plugin.
+    /// Overrides `data_nats_url` for messaging.
+    /// Example: nats://127.0.0.1:4222
+    /// When set (or when `data_nats_url` is set), components subscribe to real
+    /// NATS subjects; otherwise an in-memory backend is used.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub wasmcloud_messaging_nats_url: Option<String>,
```
jfleitz: Can we remove this and favor the data_messaging_url instead? This way we don't have too many options for Nats configuring.

jfleitz (dismissed):

crates/wash/src/config.rs
ricochet: yeah let's just have `data_nats_url`. I think the previous knobs we keep but could eye them for deprecation later.

jfleitz (approved):

PR #5244, opened 2026-06-09
Add Autoscaling support for WorkloadDeployments

### LINKED ISSUE #4939: Autoscaling WorkloadDeployments
Hello, it would be nice to be able to autoscale WorkloadDeployments with k8s builtin HPA. I was unaware of whether it is possible, so I tried to do it, but with no success. I am not an expert on k8s autoscaling so mostly debugged my deployments with the help of LLM and search engines. Here is what I learned:

- In order to autoscale custom resources, one must define [https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource](`/scale` subresource).
- There was no such subresource defined in WorkloadDeployment CR, so I added it there myself like this:
```yaml
scale:
  specReplicasPath: .spec.replicas
  statusReplicasPath: .status.replicas
  labelSelectorPath: .status.labelSelector
```
- With this addition I was able to scale workloaddeployment with `kubectl scale workloaddeployment <name> --replicas 3`
- Now I tried to deploy an HPA to do autoscaling, after adding the scale subresource, the HPA resource found the workloaddeployment, but describing the HPA resource revealed:
```
Metrics:                                              ( current / target )
  "http_requests_per_second" (target average value):  <unknown> / 10
Min replicas:                                         1
Max replicas:                                         100
...
Conditions:
  Type           Status  Reason             Message
  ----           ------  ------             -------
  AbleToScale    True    SucceededGetScale  the HPA controller was able to get the target's current scale
  ScalingActive  False   InvalidSelector    the HPA target's scale is missing a selector
Events:
  Type     Reason                        Age   From                       Message
  ----     ------                        ----  ----                       -------
  Warning  SelectorRequired              4s    horizontal-pod-autoscaler  selector is required
  Warning  FailedComputeMetricsReplicas  4s    horizontal-pod-autoscaler  selector is required
```
- Next I tried to scale it with KEDA, but it uses HPA under the hood and the HPA resource created by KEDA had exactly same event complaining that selector is required.
- Then I found this [https://github.com/kedacore/keda/issues/5898](issue) stating that the `/scale` needs to be implemented correctly for it to work, so I concluded that autoscaling is not possible right now.

I was running my test on k0s cluster using nginx ingress controller and prometheus to create `http_requests_per_second` metric, so the used metric was external. I have tested it with regular container deployments, and it was working fine with them. Testing was done on ubuntu 24.04 lts (x86_64) and versions used in the test are:
```sh
export CONTAINERD_VERSION_STRING=2.2.0-2~ubuntu.24.04~noble
export K0S_VERSION=v1.32.11+k0s.0
export KEDA_VERSION=2.18.3
export WC_VERSION=v2.0.0-rc.6 # wasmcloud
```
Because there is no images for wasmcloud rc.6, I cloned the repo and built the images myself.

Finally, here is the manifest I used to deploy the workload and HPA:
```yaml
kind: WorkloadDeployment
metadata:
  name: wash-go-server
spec:
  replicas: 1
  template:
    spec:
      hostSelector:
        hostgroup: default
      components:
        - name: go-server
          image: 10.164.178.1:5000/wash-go-server:0.1.0
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: nuhakala.com
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: go-ingress
  annotations:
    "nginx.org/proxy-pass-headers": "Host"
spec:
  ingressClassName: nginx
  rules:
  - host: nuhakala.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: runtime-gateway
            port:
              number: 80
---
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: go-deploy-hpa-workload
  namespace: default
spec:
  scaleTargetRef:
    kind: WorkloadDeployment
    name: wash-go-server
  minReplicas: 1
  maxReplicas: 100
  metrics:
  - type: External
    external:
      metric:
        name: http_requests_per_second
      target:
        type: Value
        averageValue: 10
  behavior:
    scaleDown:
      stabilizationWindowSeconds: 60
```

jfleitz:
Closes #4939 

Makes `WorkloadDeployment` autoscalable by the Kubernetes Horizontal Pod
Autoscaler — and, by extension, KEDA (which builds on HPA) — by implementing
the `/scale` subresource correctly. `kubectl scale workloaddeployment <name>
--replicas=N` now works too.

HPA scales any resource that exposes a `/scale` subresource with three
JSONPaths: `specReplicasPath`, `statusReplicasPath`, and `labelSelectorPath`.
`WorkloadDeployment` didn't define the subresource at all, and the two things
needed to make it valid were missing:

1. **No selector.** HPA requires `scale.status.selector` to be a non-empty,
   parseable label selector; without it scaling never activates.
2. **`statusReplicasPath` had no integer to point at.** `.status.replicas` is
   an object (`ReplicaSetStatus`), but HPA needs a scalar replica count.

### `/scale` subresource

- Added the kubebuilder scale marker to `WorkloadDeployment`

### New status fields

Added two flat fields to `WorkloadDeploymentStatus`:

- `currentReplicas` (`int32`) — the running replica count HPA reads via
  `statusReplicasPath` (mirrors `.status.replicas.current` as a scalar).
- `selector` (`string`) — the serialized label selector HPA reads via
  `labelSelectorPath`.

### Selector wiring

- Defined a managed label `runtime.wasmcloud.dev/workload-deployment=<name>`.
- The controller stamps this label onto the `WorkloadReplicaSet`s a deployment
  owns, so the selector truthfully matches the deployment's workloads.
- `reconcileScale` populates `status.currentReplicas` and `status.selector` on
  every reconcile — including the not-yet-available path — so HPA never
  observes a missing selector mid-deploy.

### `spec.replicas` defaults to 1

Matches native `Deployment` semantics. Without a value, HPA's `GET /scale`
errors on the missing `specReplicasPath`, and the workload would otherwise
scale to 0 when `replicas` is omitted.

### Testing
Tested with a local kind cluster using KEDA to scale the WorkloadDeployment using an OTEL metric from the Wasm component. Verified that scaling the number of components as well as EndpointSlices worked as expected.

### DIFF
--- a/charts/runtime-operator/crds/runtime.wasmcloud.dev_workloaddeployments.yaml
+++ b/charts/runtime-operator/crds/runtime.wasmcloud.dev_workloaddeployments.yaml
@@ -98,6 +98,7 @@ spec:
                     type: object
                 type: object
               replicas:
+                default: 1
                 format: int32
                 type: integer
               template:
@@ -150,7 +151,26 @@ spec:
                                 will be made available to a workload component.
                               properties:
                                 allowedHosts:
+                                  description: |-
+                                    AllowedHosts is the outbound egress allowlist for this component.
+
+                                    Each entry must match one of:
+                                      - "*"                       (allow all)
+                                      - "host[:port]"             e.g. "example.com" or "example.com:8443"
+                                      - "scheme://host[:port][/]" e.g. "https://api.example.com" or "https://api.example.com/"
+                                      - "*.suffix[:port]"         e.g. "*.example.com" or "*.example.com:8443"
+                                      - "scheme://*.suffix[:port][/]" e.g. "https://*.example.com"
+
+                                    This is a hosts policy, not a URL policy: entries must not include a
+                                    path (beyond bare `/`), query string, or fragment. The wildcard must
+                                    be `*.<rest>` (leading dot required); a bare `*foo` is rejected.
+
+                                    Empty or absent allowedHosts denies all outgoing requests
+                                    (fail-closed). To opt into unrestricted egress, set `allowedHosts:
+                                    ["*"]` explicitly. Final validation runs in the runtime. This regex
+                                    is an admission-time guard, not the source of truth.
                                   items:
+                                    pattern: ^\*$|^([A-Za-z][A-Za-z0-9+.-]*://)(\*\.)?[A-Za-z0-9]([A-Za-z0-9-]{0,61}[A-Za-z0-9])?(\.[A-Za-z0-9]([A-Za-z0-9-]{0,61}[A-Za-z0-9])?)*(:[0-9]{1,5})?/?$|^(\*\.)?[A-Za-z0-9]([A-Za-z0-9-]{0,61}[A-Za-z0-9])?(\.[A-Za-z0-9]([A-Za-z0-9-]{0,61}[A-Za-z0-9])?)*(:[0-9]{1,5})?$
                                     type: string
                                   type: array
                                 config:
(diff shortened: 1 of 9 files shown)


jfleitz:
Retested this with [https://github.com/wasmCloud/wasmCloud/pull/5246](https://github.com/wasmCloud/wasmCloud/pull/5246), and worked as expected with KEDA driving HPA using Envoy metrics.

ricochet (approved):
This lgtm. Wondering if there's a way to include a basic version of this in our e2e tests.

jfleitz:
> This lgtm. Wondering if there's a way to include a basic version of this in our e2e tests.

Added this as a new [issue](https://github.com/wasmCloud/wasmCloud/issues/5249), since this will take a little effort on best approach (to not boggle down our CI/CD more)

PR #5254, opened 2026-06-16
feat(wash-runtime): make engine config composable via WasmProposal

### LINKED ISSUE #5023: [FEATURE] Configurable engine flags and builder
The `EngineBuilder` currently hardcodes wasmtime `Config` settings internally, making it difficult to extend with new WebAssembly proposals and feature flags. As we add support for more features like WASIP3, the number of engine configuration knobs is growing and the interactions between them are getting trickier.

When adding the wasip3 flag to the engine builder ([#5022 discussion](https://github.com/wasmCloud/wasmCloud/pull/5022#discussion_r3030873356)), we had to manually wire wasm_component_model_async into the config when the flag is enabled. This pattern of "feature flag X implies wasmtime config settings Y, Z, W" is only going to grow as wasmtime stabilizes more proposals.

Looking at wasmtime's [stability tiers](https://docs.wasmtime.dev/stability-tiers.html?highlight=exception-handling#tier-2), there are several Tier 2+ features that users or plugins may want to opt into:

- wasm_component_model_async (needed for WASIP3)
- wasm_gc / wasm_gc_types
- wasm_exception_handling
- wasm_wide_arithmetic

Each of these has implications for the pooling allocator config, linker setup, and feature interactions.

## Proposal

Make the engine configuration more composable and transparent:

1. **Expose commonly needed wasmtime Config flags** as first-class EngineBuilder methods rather than requiring with_config() (which is all-or-nothing and mutually exclusive with other builder options).
2. **Document which flags are implied by higher-level features** e.g., with_wasip3(true) enables component_model_async + async_builtins + async_stackful. Users shouldn't need to read the source to know what's being turned on.
3. **Consider a feature/proposal enum** for enabling bundles of related settings: 
```rust
engine.with_wasm_proposal(WasmProposal::ComponentModelAsync)
      .with_wasm_proposal(WasmProposal::Threads)
```

Current Pain Points
- `with_config()` and `with_max_instances()` / `with_pooling_allocator()` are mutually exclusive, you can't use a custom config and pooling settings together
- Feature interactions are implicit (e.g., WASIP3 silently enables three separate config flags)
- No way to enable a single additional wasm proposal without replacing the entire config
- Pooling allocator config (PoolingAllocationConfig) is not externally configurable

Aditya1404Sal:
Closes #5023 
- Layer builder settings (pooling, fuel, proposals) on top of any with_config() base instead of erroring when combined
- Add WasmProposal enum + with_wasm_proposal() to enable individual wasm proposals without replacing the whole config
- Add with_pooling_config() for external PoolingAllocationConfig
- Only apply fuel consumption when explicitly set
- Document that with_wasip3() implies component-model-async

### DIFF
--- a/crates/wash-runtime/Cargo.toml
+++ b/crates/wash-runtime/Cargo.toml
@@ -82,7 +82,7 @@ uuid = { workspace = true, features = ["v4"] }
 tempfile = { workspace = true }
 tokio = { workspace = true, features = ["sync", "net", "macros"] }
 tracing = { workspace = true }
-wasmtime = { workspace = true, features = ["anyhow", "component-model", "component-model-async", "cranelift", "pooling-allocator", "gc", "gc-null"] }
+wasmtime = { workspace = true, features = ["anyhow", "component-model", "component-model-async", "cranelift", "pooling-allocator", "gc", "gc-null", "threads"] }
 wasmtime-wasi = { workspace = true }
 wasmtime-wasi-io = { workspace = true }
 wasmtime-wasi-http = { workspace = true, features = ["default-send-request", "p2"] }
--- a/crates/wash-runtime/src/engine/mod.rs
+++ b/crates/wash-runtime/src/engine/mod.rs
@@ -604,19 +604,95 @@ impl Engine {
     }
 }
 
+/// A wasmtime WebAssembly proposal that can be opted into on the engine.
+///
+/// Each variant maps to one or more `wasmtime::Config` feature flags. Use it
+/// with [`EngineBuilder::with_wasm_proposal`] to enable a bundle of related
+/// settings without having to replace the entire config via
+/// [`EngineBuilder::with_config`].
+///
+/// ```no_run
+/// # use wash_runtime::engine::{Engine, WasmProposal};
+/// # fn example() -> anyhow::Result<()> {
+/// let engine = Engine::builder()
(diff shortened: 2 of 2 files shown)


if0ne:
Nice improvement. Perhaps we should consider [bitflags](https://docs.rs/bitflags/latest/bitflags/) for WasmProposal?

ricochet:
Thinking though bitflags vs the enum approach, it's really a choice between compiler-enforced completeness of apply() (enum) versus A | B composition + mirroring wasmtime's own WasmFeatures.

Mirroring wasmtime's idiom has nice pros for extensibility and maintainability in that we won't have to introduce changes for end-users to take advantage of new flags. This downside can be mitigated in the enum approach by adding something like:

```rust
  pub fn with_config_fn(
      mut self,
      f: impl FnOnce(&mut wasmtime::Config) + Send + Sync + 'static,
  ) -> Self {
      self.config_fn = Some(Box::new(f));
      self
  }

...

  Engine::builder()
      .with_pooling_allocator(true)             // still works
      .with_wasm_proposal(WasmProposal::Gc)     // still works
      .with_config_fn(|cfg| { cfg.wasm_brand_new_thing(true); })  // anything else
      .build()?;
```

I like us having our own layer here since we likely will have interactions with our own feature flags.

ricochet (approved):

PR #5257, opened 2026-06-16
fix: bump wasmtime to 45.0.2 for RUSTSEC-2026-0182

ricochet:
Resolves the WASIp1 fd_renumber leak (RUSTSEC-2026-0182) in
wasmtime-wasi by updating the wasmtime crate family from 45.0.0
to the patched 45.0.2 release.


ricochet:
#5256 won the race

PR #5264, opened 2026-06-17
fix(wash-runtime): stream P3 HTTP responses and release timed-out gRPC bodies

ricochet:
Rework the P3 wasi:http handler to stream response bodies instead of
buffering them with collect(). The component now runs in a task that owns
the store and drives run_concurrent, forwarding frames over a bounded
channel (backpressure instead of unbounded buffering). The store and
therefore the guest stays alive until the body is fully drained, and the
task is aborted via AbortOnDropHandle when the response body is dropped, so
a disconnected client cancels the guest promptly (there is no epoch/
wall-clock backstop otherwise).

into_http's finish future now reports the body-delivery outcome back to the
guest (Ok on clean delivery, ConnectionTerminated if the client goes away)
instead of signalling immediate success.

Fix a connection leak in the P3 gRPC outbound path: the between-bytes
timeout body kept its inner hyper::body::Incoming alive after surfacing
ConnectionReadTimeout, pinning the HTTP/2 stream/TCP connection until the
guest dropped its handle. The inner body is now held in an Option and
dropped eagerly on timeout, end-of-stream, or error.

Add outbound-HTTP client span instrumentation (P2 default + gRPC, P3),
recording method, url.full, server.address/port and response status, and
marking otel.status_code=ERROR on transport failures and allowed-hosts
denials. Adopt opentelemetry-semantic-conventions constants on the incoming
request span and record http.response.body.size via a body wrapper that
tallies streamed bytes.

Tests:
- TimedBody drops its inner body (and connection) on between-bytes timeout
- MeteredBody counts data-frame bytes and excludes trailers
- ChannelBody streams frames incrementally and aborts the task on drop

### DIFF
--- a/crates/wash-runtime/src/host/http.rs
+++ b/crates/wash-runtime/src/host/http.rs
@@ -38,6 +38,10 @@ use hyper_util::{
     server::conn::auto,
 };
 use opentelemetry::{KeyValue, context::FutureExt};
+use opentelemetry_semantic_conventions::attribute::{
+    HTTP_REQUEST_METHOD, HTTP_RESPONSE_BODY_SIZE, HTTP_RESPONSE_STATUS_CODE, OTEL_STATUS_CODE,
+    RPC_GRPC_STATUS_CODE, SERVER_ADDRESS, SERVER_PORT, URL_FULL, URL_PATH,
+};
 use tokio::net::TcpListener;
 use tokio::task::JoinHandle;
 use tracing::{Instrument, debug, error, info, instrument, warn};
@@ -339,9 +343,23 @@ impl OutgoingHandler for DefaultOutgoingHandler {
         config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
     ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>
     {
-        Ok(wasmtime_wasi_http::p2::default_send_request(
-            request, config,
-        ))
+        // Spawn the default handler ourselves (rather than calling
+        // `default_send_request`) so the request can be wrapped in a client
+        // span and the response status recorded once it arrives.
+        let span = outbound_client_span(request.method(), request.uri());
+        let handle = wasmtime_wasi::runtime::spawn(
+            async move {
+                let result =
+                    wasmtime_wasi_http::p2::default_send_request_handler(request, config).await;
+                match &result {
+                    Ok(incoming) => record_outbound_status(incoming.resp.status()),
+                    Err(_) => record_outbound_error(),
+                }
+                Ok(result)
+            }
+            .instrument(span),
+        );
+        Ok(HostFutureIncomingResponse::pending(handle))
     }
     #[cfg(feature = "wasip3")]
     fn send_request_p3(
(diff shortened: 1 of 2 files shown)


crates/wash-runtime/src/host/http.rs
```diff
@@ -1071,12 +1118,132 @@ async fn handle_http_request<T: Router>(
...
+    span.record(HTTP_RESPONSE_STATUS_CODE, status.as_u16());
+    if status.as_u16() >= 400 {
+        span.record("otel.status_code", "ERROR");
+    }
+}
+
+/// Mark the current outbound span as failed when the request never produced a
+/// response (allowed-hosts denial, connection failure, transport error, …).
+fn record_outbound_error() {
+    tracing::Span::current().record("otel.status_code", "ERROR");
```
vados-cosmonic: NIT: We could probably use [`OTEL_STATUS_CODE`](https://docs.rs/opentelemetry-semantic-conventions/0.32.0/opentelemetry_semantic_conventions/attribute/constant.OTEL_STATUS_CODE.html) here and above

crates/wash-runtime/src/host/http.rs
```diff
@@ -1071,12 +1118,132 @@ async fn handle_http_request<T: Router>(
...
+    }
+    span
+}
+
+/// Record an outbound response status on the current span. Per the HTTP client
+/// semconv, both 4xx and 5xx flip `otel.status_code` to `ERROR` for client
+/// spans (unlike server spans, where 4xx stays UNSET).
+fn record_outbound_status(status: hyper::StatusCode) {
+    let span = tracing::Span::current();
+    span.record(HTTP_RESPONSE_STATUS_CODE, status.as_u16());
```
vados-cosmonic, line 1162: When we're doing a GRPC request, should we be using [`RPC_GRPC_STATUS_CODE`](https://docs.rs/opentelemetry-semantic-conventions/0.32.0/opentelemetry_semantic_conventions/attribute/constant.RPC_GRPC_STATUS_CODE.html) ?

crates/wash-runtime/src/host/http_p3.rs
```diff
@@ -57,65 +90,217 @@ pub async fn handle_component_request_p3(
...
+                        // `wasi:http/types.response#new`). Resolve it once we have
+                        // finished forwarding the body to the client, so a guest
+                        // that awaits the result learns whether delivery succeeded.
+                        let (finish_tx, finish_rx) =
+                            tokio::sync::oneshot::channel::<Result<(), ErrorCode>>();
+                        let http_response = match store.with(|s| {
+                            response.into_http(s, async move {
+                                // A dropped sender (task aborted mid-stream) is
+                                // treated as "no error to report".
+                                finish_rx.await.unwrap_or(Ok(()))
```
vados-cosmonic: This seems like a place to add a debug/trace level message

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but NITs

crates/wash-runtime/src/host/http.rs
ricochet: shoot not sure how I missed that one

ricochet:
FYI filed https://github.com/wasmCloud/wasmCloud/issues/5266 as another best practice we could consider.

vados-cosmonic (approved):

PR #5268, opened 2026-06-18
feat(wash-runtime): implements feature flag

ricochet:
Add the host-side routing that lets a standalone plugin and a multiplexed
(named-instance) plugin serve the same package without conflict..

An unnamed import is served by the regular plugin's default instance, while a named (implements ..) import is served by a plugin advertising
supports_named_instances(). This is so that we can keep implements behind a
feature flag.

Behavior is unchanged when only one kind of plugin is present.

### DIFF
--- a/crates/wash-runtime/src/engine/mod.rs
+++ b/crates/wash-runtime/src/engine/mod.rs
@@ -625,6 +625,14 @@ pub enum WasmProposal {
     /// Component model async ABI (`stream`/`future`/`error-context` types).
     /// Enables `wasm_component_model_async`. Required for WASIP3.
     ComponentModelAsync,
+    /// Component model `(implements ..)` named imports, letting a component
+    /// import the same interface multiple times under distinct names so host
+    /// plugins can route each independently. Enables
+    /// `wasm_component_model_implements`. Requires the backported wasmtime
+    /// support, so it is only available with the `wasm_component_model_implements`
+    /// crate feature.
+    #[cfg(feature = "wasm_component_model_implements")]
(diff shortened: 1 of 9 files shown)


crates/wash-runtime/src/engine/mod.rs
```diff
@@ -729,6 +729,13 @@ impl EngineBuilder {
         } else {
             let mut cfg = wasmtime::Config::default();
 
+            // Accept components that import an interface multiple times via the
+            // component-model `(implements ..)` annotation, so host plugins can
+            // route each named import (e.g. two `wasi:keyvalue/store` imports
+            // backed by redis vs NATS) independently.
+            #[cfg(feature = "implements")]
```
vados-cosmonic: could we have a better name for this? `cm_implements` or something? or `wasmtime_implements`?

vados-cosmonic (approved):
LGTM 🚀

crates/wash-runtime/src/engine/mod.rs
ricochet: I like cm_implements

Aditya1404Sal (approved):
LGTM ! (Just fyi : the wasmtime proposal coupler for the engine in #5254 should superceed the bare `cfg.wasm_component_model_implements` later on :)

crates/wash-runtime/src/engine/mod.rs
ricochet: Going to opt for `wasm_component_model_implements` to match upstream

PR #5270, opened 2026-06-19
Fix quickstart: Update wasi:http/incoming-handler version

teeberg:
The Rust quickstart requires version 0.2.9 of the `wasi:http/incoming-handler`. See https://github.com/wasmCloud/wasmcloud.com/pull/1203 for more details.


ricochet:
Thank you for jumping in! Another PR around the same time also fixed this https://github.com/wasmCloud/wasmCloud/pull/5278

PR #5271, opened 2026-06-19
fix(template): align Rust http-hello-world WIT and dep versions

ericgregory:
The Rust `http-hello-world` template was internally inconsistent once a user reached the [docs/quickstart "Add persistent storage" step](https://wasmcloud.com/docs/quickstart/develop-a-webassembly-component/#add-persistent-storage). `wstd 0.6.6`'s transitive `wasip2 1.0.3+wasi-0.2.9` produces a `wasi-http 0.2.9` export, but the template's `wit/world.wit` declared `0.2.2`. The stock template builds because `wstd::http_server` doesn't run through `wit_bindgen`, but the moment the quickstart adds `wit_bindgen::generate!`, the encoder rejects the component.

This PR makes the following changes:

- `wit/world.wit`: bump to `wasi:http/incoming-handler@0.2.9` with a comment documenting the coupling
- `Cargo.toml`: pin `wstd = "=0.6.6"` and add `wasip2 = "=1.0.3"` as a direct dep (previously transitive with `^1.0`, so any new `wasip2` patch could silently break the template)
- `Cargo.lock` / `wkg.lock`: regenerated to reflect the pins

Note:

- wasmcloud.com PR #1203 bumped the doc to 0.2.9 to work around this. With the template now consistent, the doc's `world.wit` diff for the persistent-storage step shows only the keyvalue imports as additions, not the (previously hidden) handler version bump
- A companion PR is needed in `wasmCloud/typescript` to pin `@bytecodealliance/jco-std` similarly — TS jco-std 0.1.x only ships 0.2.6 adapters, so the doc's TS code block still needs 0.2.6 (a follow-up docs PR will split the per-language version)


ericgregory:
Superseded by #5278

PR #5272, opened 2026-06-19
feat(examples): add http-hello-world-persistent-storage

ericgregory:
Adds the Rust HTTP component as it stands at the end of the wasmcloud.com quickstart's [Add persistent storage](https://wasmcloud.com/docs/quickstart/develop-a-webassembly-component#add-persistent-storage) step, so we can test and catch any regressions.

Note:

`wasi:http/incoming-handler` is intentionally **omitted** from `wit/world.wit` because `wstd`'s `#[http_server]` proc macro generates that export internally. Leaving it declared in the world while also calling `wit_bindgen::generate!` (the pattern the current quickstart prose shows) makes the component encoder fail with *"failed to find export of interface `wasi:http/incoming-handler@...` function `handle`"* — the macro and the manual binding generation race over the same export symbol. Same pattern `examples/blobby` already uses; documented in the example's README. Docs follow-up to be filed against wasmcloud.com.


vados-cosmonic:
Quick question... should we just maybe add the quickstart lvl 2 code as an example? A script that mostly copies in the file contents (and that we have to change when the example changes) seems almost identical to just having the code in-tree.

ericgregory (commented):
Thanks, this is definitely a lot cleaner as a regular example. Rewrote to that end!

vados-cosmonic (approved):
LGTM 🚀

PR #5279, opened 2026-06-23
feat(wash-runtime): multiplex wasi:keyvalue via `(implements ..)` imports

ricochet:
Let a component import the same interface multiple times under distinct
labels and route each to its own backend. A component that imports
`wasi:keyvalue/store` twice (e.g. `import team-a: …; import team-b: …;`)
gets each label bound to a separate backend (in-memory/filesystem/redis/
nats), and the per-import id routes every call.

Core pieces:
- plugin/multiplex.rs: generic `Multiplexer<Id>` + `BackendProvider<Id>`
  with a connection pool (single-flight via OnceCell, LRU eviction, and
  per-backend-type caps) and a background idle reaper.
- wasi_keyvalue/multiplexed{.rs,/}: the multiplexed keyvalue plugin
  (wasmtime `named_imports` bindgen) threading the backend id through
  store/atomics/batch and the bucket resource.
- wasi_keyvalue/fs_store.rs: shared `FsKvStore` so the standalone
  `FilesystemKeyValue` and the multiplexed filesystem backend share one
  implementation (also fixes blocking IO on async paths and increment
  overflow).
- engine/workload.rs: surface `(implements ..)` imports as named
  `WitInterface`s — identity from the `implements` annotation, label as
  the routing name — so the plugin dispatcher binds them. Without this the
  multiplexed plugin was never bound for real-guest implements imports.

Testing & toolchain:
- keyvalue-implements fixture: a real guest emitting labeled external
  `wasi:keyvalue/store` imports, with an end-to-end isolation test; plus
  coexistence (standalone + multiplexed on one linker) and a redis/nats
  multiplexed test. Bump fixtures wit-bindgen to 0.58.
- setup-rust now installs a newer standalone `wasm-component-ld` (via
  cargo-binstall): the toolchain-bundled linker lags the crate and can't
  encode `(implements ..)` imports.
- Gate Docker-backed integration tests (nats/oci) behind `#[ignore]`,
  run in a Linux-only CI leg.

### DIFF
--- a/crates/wash-runtime/src/plugin/wasi_keyvalue/multiplexed.rs
+++ b/crates/wash-runtime/src/plugin/wasi_keyvalue/multiplexed.rs
@@ -1,0 +1,940 @@
+//! # Multiplexed WASI KeyValue (implements-routed)
+//!
+//! Binds `wasi:keyvalue` via the component-model `(implements ..)` /
+//! `named_imports` mechanism so a single component can import the same
+//! interface multiple times and have each import backed by a *different*
+//! backend (e.g. one `wasi:keyvalue/store` import backed by redis and another
+//! by NATS).
+//!
+//! Each named import is resolved by a `lookup` closure to a [`KvBackend`]
+//! (the wasmtime "implements id"); the embedder-chosen id is threaded into
+//! every host method, including the `bucket` resource methods.
+
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+
+use bytes::{Buf, Bytes};
+use futures::stream::{FuturesOrdered, StreamExt};
+use redis::AsyncCommands;
+use tokio::sync::RwLock;
+use wasmtime::component::Resource;
+
+use crate::engine::ctx::{SharedCtx, extract_active_ctx};
+use crate::engine::workload::WorkloadItem;
+use crate::plugin::multiplex::{BackendProvider, Multiplexer};
+use crate::plugin::{HostPlugin, WitInterfaces};
+use crate::wit::{WitInterface, WitWorld};
+
+mod bindings {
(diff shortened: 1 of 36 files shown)


vados-cosmonic (dismissed):
Hey so this mostly looks good to me!

I was unclear on why we removed the skipped test stuff in favor of --run-ignored, but I assume those are still under test via CI so we should be OK there

ricochet:
> Hey so this mostly looks good to me!
> 
> I was unclear on why we removed the skipped test stuff in favor of --run-ignored, but I assume those are still under test via CI so we should be OK there

Yep, running in CI for ubuntu. Basically I added a handful of new integration tests and ran into cases where not all of them were opt-ed into in the CI. It was also awkward to do while I was running locally to test and `--run-ignored` seemed to be the most common approach folks used with testcontainer integration tests. If macos and windows github runners actually supported nested virtualization, I wouldn't even have them ignored, just run them all the time.

crates/wash-runtime/src/plugin/wasi_keyvalue/multiplexed.rs
```diff
@@ -0,0 +1,940 @@
...
+    fn backend_type(&self) -> &'static str {
+        DEFAULT_BACKEND
+    }
+
+    async fn instantiate(&self, _config: &HashMap<String, String>) -> anyhow::Result<KvId> {
+        Ok(Arc::new(InMemoryBackend::new()))
+    }
+}
+
+// ---- redis backend ---------------------------------------------------------
```
vados-cosmonic, line 372: NIT: this should be in another file, but IMO it's worth doing in a follow up

vados-cosmonic (approved):
LGTM 🚀

PR #5286, opened 2026-06-24
feat(wit): add async wasmcloud:keyvalue and wasmcloud:blobstore

ricochet:
Introduce wasmCloud-namespaced, async-native WIT packages for key-value and
blob storage, laid out like wit/secrets and wit/messaging:

  wit/keyvalue/wit/world.wit   wasmcloud:keyvalue@0.1.0
  wit/blobstore/wit/world.wit  wasmcloud:blobstore@0.1.0

Every backend-touching operation is an `async func`, so a component can
have many in-flight storage operations without the wasi:io/poll ceremony
required under wasip2. Blob bodies use the native component-model
`stream<u8>` instead of wasi:io streams, which drops the incoming-value,
outgoing-value, and stream-object-names resources and removes the
wasi:io dependency entirely (wkg.lock has no packages).

These intentionally live in the wasmcloud namespace rather than waiting
on wasi:keyvalue/wasi:blobstore to gain async support upstream.
Our goal is to align with wasi upstream once we have solid working
implementation feedback.


jfleitz (dismissed):

.github/workflows/wit.yml
```diff
@@ -96,17 +103,29 @@ jobs:
...
           version: ${{ env.WASM_TOOLS_VERSION }}
 
-      - uses: wasmcloud/setup-wash-action@47756fa211cc82f392c57e39504f43ef46fc0b3e # v2.1.0
+      - name: Setup wkg
+        # Pinned wkg binary (no published setup action). Used instead of
+        # `wash wit build` because wash's bundled wit-parser predates wasip3
+        # async syntax; see the WKG_VERSION note above. Installed into a temp
+        # dir added to PATH rather than /usr/local/bin, which is not reliably
+        # writable without sudo across runner images.
+        run: |
```
vados-cosmonic, line 116: Did taikie-e/cache-install-action not work here? 

And why are we using `wkg` instead of `wash wit build` ? Is `wash` behind in the version it's using?

vados-cosmonic (dismissed):

.github/workflows/wit.yml
ricochet: Yeah wash behind

ricochet: I can swap us over to binstall. Would that be better?

vados-cosmonic: binstall would be if wkg is in there yeah that's fine -- honestly the ideal fix is to update wash BUT IMO not worth blocking on this.

Would be fine with a merge as-is too, if we leave a TODO that we should move back to `wash wit build` when it get updatetd

ricochet: cool dropped in a TODO

vados-cosmonic (approved):

PR #5289, opened 2026-06-26
feat(wash-runtime): multiplex wasmcloud:messaging/consumer

ricochet:
Bind `wasmcloud:messaging/consumer` through the component-model
`(implements ..)` / `named_imports` mechanism so a single component can
import the consumer interface multiple times and route each import's
outbound `publish`/`request` to a different backend (e.g. one import
publishing to NATS cluster A and another to cluster B). The consumer
interface has no resources, so the resolved backend (`MsgBackend`) is
threaded directly into each host method; `types` is record-only and bound
via the regular path. The `handler` export is unaffected by import
multiplexing and stays served by the standalone messaging plugins.

- crates/wash-runtime: add `wasmcloud_messaging::multiplexed`
- crates/wash host/dev: register `MultiplexedMessaging`
- integration test against 2 nats clusters to show tenant isolation

### DIFF
--- a/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
+++ b/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
@@ -1,7 +1,14 @@
 mod in_memory;
+#[cfg(feature = "wasm_component_model_implements")]
+mod multiplexed;
 mod nats;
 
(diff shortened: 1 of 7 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5291, opened 2026-06-26
feat(wash-runtime): multiplex wasmcloud:postgres via `(implements ..)` imports

ricochet:
Route per-import `wasmcloud:postgres` named imports to isolated, per-credential
connection pools through multiplexer core. A component can import
`wasmcloud:postgres/query` multiple times under distinct labels and have each
backed by a different postgres connection credential or database.

- Add `PgId` (the implements id): a deadpool pool plus its own prepared-statement
  table, Arc-backed so binds the multiplexer routes to the same URL share one
  pool and prepared statements drop with the connection.
- Register a `PostgresProvider` (pool_key = URL) on a `Multiplexer<PgId>`; split
  `on_workload_item_bind` into the legacy unnamed per-component-database path and
  the named per-credential path. `types` is linked once for both.

Fix `WitWorld::includes_bidirectional`: it bound to the first import matching
namespace+package, so a package split across multiple imports (postgres `types`
unnamed + `query` under labels) matched order-dependently against the imports
HashSet, intermittently failing to bind the plugin. Now match each interface
across all imports/exports.

### DIFF
--- a/crates/wash-runtime/src/wit.rs
+++ b/crates/wash-runtime/src/wit.rs
@@ -50,46 +50,34 @@ impl WitWorld {
     /// different than [`WitWorld::includes`] because it considers that in one
     /// [`WitInterface`] there may be both imports and exports.
     pub fn includes_bidirectional(&self, interface: &WitInterface) -> bool {
-        let import_match = self.imports.iter().find(|i| {
+        // Same namespace+package, and (when both specify one) the same version.
+        // Name is intentionally ignored here: this answers "does the world use
+        // this namespace:package interface", regardless of the `(implements ..)`
+        // label — label routing is resolved later during plugin binding.
+        let matches_pkg = |other: &WitInterface| {
             if let Some(v) = &interface.version
-                && let Some(ov) = &i.version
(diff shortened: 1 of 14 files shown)


crates/wash-runtime/src/wit.rs
```diff
@@ -50,46 +50,34 @@ impl WitWorld {
     /// different than [`WitWorld::includes`] because it considers that in one
     /// [`WitInterface`] there may be both imports and exports.
     pub fn includes_bidirectional(&self, interface: &WitInterface) -> bool {
-        let import_match = self.imports.iter().find(|i| {
+        // Same namespace+package, and (when both specify one) the same version.
+        // Name is intentionally ignored here: this answers "does the world use
+        // this namespace:package interface", regardless of the `(implements ..)`
+        // label — label routing is resolved later during plugin binding.
+        let matches_pkg = |other: &WitInterface| {
```
vados-cosmonic: this should probably be a method on `WitInterface` ?

vados-cosmonic (dismissed):
LGTM 🚀

PR #5295, opened 2026-06-28
fix: support multiline world declarations in wash wit add

immanuwell:
No linked issue, couldnt find a direct match.

`wash wit add` fails on valid `world.wit` files when the `world` opening brace is on the next line. It bails with `Could not find a world block in world.wit to add the import`, kinda rough.

This makes the import insertion logic handle both brace styles, keeps new imports after the last existing import, and adds regression tests for the helper plus the real command path.

Repro:
1. Create `wit/world.wit` with:
```wit
package test:component@0.1.0;

world example
{
}
```
2. Run `wash wit add wasi:logging/logging@0.1.0-draft`
3. Before this fix it errors with `Could not find a world block in world.wit to add the import`
4. With this fix the import is added inside the `world` block and the command keeps going

### DIFF
--- a/crates/wash/src/cli/wit.rs
+++ b/crates/wash/src/cli/wit.rs
@@ -281,6 +281,83 @@ async fn find_world_wit_file(wit_dir: &std::path::Path) -> Result<std::path::Pat
     )
 }
 
+fn insert_import_into_world(content: &str, import_line: &str) -> Result<String> {
+    let lines: Vec<&str> = content.lines().collect();
+    let mut new_lines = Vec::new();
+    let mut inserted = false;
+    let mut pending_world = false;
+    let mut world_indent = String::new();
+
+    for (i, line) in lines.iter().enumerate() {
+        let trimmed = line.trim();
+
+        if trimmed.starts_with("world ") {
+            pending_world = true;
+            if trimmed.contains('{') {
+                pending_world = false;
+                let next_indent = lines
+                    .get(i + 1)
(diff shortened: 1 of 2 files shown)


crates/wash/src/cli/wit.rs
```diff
@@ -281,6 +281,83 @@ async fn find_world_wit_file(wit_dir: &std::path::Path) -> Result<std::path::Pat
...
+    let mut world_indent = String::new();
+
+    for (i, line) in lines.iter().enumerate() {
+        let trimmed = line.trim();
+
+        if trimmed.starts_with("world ") {
+            pending_world = true;
+            if trimmed.contains('{') {
+                pending_world = false;
+                let next_indent = lines
```
ricochet, line 298: could be a helper since used twice, but this is good

crates/wash/tests/integration_wit.rs
```diff
@@ -107,6 +107,40 @@ async fn test_add_fetch_clean_workflow() -> Result<()> {
     Ok(())
 }
 
+/// Test 1b: Add dependency when the world opening brace is on the next line
+/// Verifies `wash wit add` handles valid multi-line world declarations
+#[tokio::test]
```
ricochet, line 112: I think we can pair this one down to call ` insert_import_into_world(original_content, &import_line)` directly.  As a side benefit it would no longer need a tempdir or async file I/O, and can be converted to a plain #[test] and run in-memory.

ricochet (approved):
just left two nits. OK not to do those

PR #5296, opened 2026-06-29
feat(wit)!: share keyvalue `bucket` via a `types` interface

ricochet:
Move the `bucket` resource (and `error`, `key-response`, `set-options`) out of `wasmcloud:keyvalue/store` into a new shared `types` interface. `store`, `atomics`, `cas`, `batch`, and `watcher` now `use types.{bucket, error}`, and `store` keeps only `open`.

Why: with `(implements ..)` routing, a guest that imports a labeled (multiplexed) `store` could not also use `cas`/`atomics`/`batch`. Those interfaces `use store.{bucket}`, which resolves to the *unlabeled* `store`. Defining `bucket` in a `use`d `types` interface makes the type identical across all of them. This is the same shape `wasmcloud:blobstore` already has (its resource lives in the `container` interface).

Also:
- Split the keyvalue package into one file per interface
  (types/store/atomics/cas/batch/watcher) plus world.wit for the worlds.
- Vendor wasmcloud:keyvalue (restructured) and wasmcloud:blobstore into
  wash-runtime (wit/deps, tests/fixtures/p3-wit-deps) so bindgen! and the P3
  fixtures see the new shape; the async-keyvalue world now imports `types`.

Since we haven't released with this yet. My intention is to yank the current 0.1.0 oci wit so that when this merges, it will be republished.


Aditya1404Sal (approved):
LGTM 🚀

PR #5298, opened 2026-06-29
Only update LastSeen if no change to metadata

jfleitz:
## Desciption
A wasmCloud host currently sends a heartbeat message every 15 seconds to the runtime-operator that includes it's Friendly Name, Namespace, HostID etc. This metadata is rarely changed once the hosts is registered. 

The heartbeat is also used to report when the host was last seen so that the operator can determine if Workloads need to be rescheduled or if the host is ready to be scheduled to.

This change checked the cached Host meta data in Kubernetes with the heartbeat metadata. If it is the same, then only an update to the `Status.LastSeen` sub-resource is completed. Other than lessening the update call to kube-api, this also prevents the operator from triggering a Host update that would fire the reconciler off again as well.

## Testing
Tested locally following the wasmCloud installation steps and a local operator container image.
Three hosts were initially created and scaled to 5 to ensure that no error was reported by the operator and the `Host` custom resource objects were created as well.


runtime-operator/internal/controller/runtime/host_controller.go
```diff
@@ -226,15 +230,30 @@ func (h *hostStatusUpdater) Start(ctx context.Context) error {
...
-			log.Error(err, "failed to apply Host resource", "host", req.FriendlyName, "hostID", req.Id)
+		// Read the current state from the cache to see if the values in the heartbeat changed.
+		existing := &runtimev1alpha1.Host{}
+		getErr := h.client.Get(ctx,
+			types.NamespacedName{Name: req.FriendlyName, Namespace: h.operatorNamespace},
+			existing)
+		switch {
+		case apierrors.IsNotFound(getErr):
+			// fall through: the apply below creates the Host
+		case getErr != nil:
```
ricochet:
```suggestion
		if getErr != nil && !apierrors.IsNotFound(getErr) {
```

ricochet (dismissed):

ricochet (approved):

PR #5309, opened 2026-07-02
fix: default backends for wasmcloud:{keyvalue,blobstore}

ricochet:
Adds support for default backends for wasmcloud:{keyvalue,blobstore}.
Multiplexed capability plugins previously required an `(implements ..)`
label on an import to bind a backend; a plain (unlabeled) import of
`wasmcloud:blobstore` or `wasmcloud:keyvalue/store` had no backend and
failed to instantiate. Now any multiplexed plugin serves a plain import
from the workload's default (the unnamed `""`) route.

- multiplex: add `defaults` map + `set_default`/`default_for` (+ unit test)
- wasi_blobstore: move the local default map onto the shared `Multiplexer`
- wasi_keyvalue: add the standard `store` binding + labeled/plain gating
- add `blobstore-default-p3` / `keyvalue-default-p3` real-guest fixtures and
  in-memory P3 integration tests proving a plain import round-trips through
  a default backend with no label

### DIFF
--- a/crates/wash-runtime/src/plugin/multiplex.rs
+++ b/crates/wash-runtime/src/plugin/multiplex.rs
@@ -98,6 +98,12 @@ pub struct Multiplexer<Id> {
     /// Whether the background idle-reaper task has been spawned yet (spawned
     /// lazily on the first pooled bind, when a tokio runtime is guaranteed).
     reaper_started: AtomicBool,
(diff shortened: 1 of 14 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5314, opened 2026-07-02
Route messages to Services

ricochet:
Builds on #5312

This is specifically for routing messages to long-running Services in addition to handling inbound http.

### DIFF
--- a/crates/wash-runtime/src/engine/linked_call.rs
+++ b/crates/wash-runtime/src/engine/linked_call.rs
@@ -35,6 +35,8 @@ use wasmtime_wasi::WasiCtxBuilder;
 #[cfg(feature = "wasi-tls")]
 use crate::engine::ctx::SharedTlsProvider;
 use crate::engine::ctx::{AccessorActiveCtxGuard, Ctx, SharedCtx, StoreActiveCtxGuard};
+use crate::engine::relocate::{self, Relocated, bridgeable_element_type};
+use crate::engine::stream_pump::Done;
 use crate::engine::value::{carries_cross_store_handle, lift_results, lower_params};
 use crate::engine::volumes::{ResolvedVolumeMount, resolve_component_volume_mounts_in_map};
 use crate::engine::workload::{WorkloadComponent, WorkloadMetadata};
@@ -141,6 +143,15 @@ pub(crate) struct EphemeralLinkedCall {
     pub(crate) linked_component_ids: Vec<Arc<str>>,
     #[cfg(feature = "wasi-tls")]
     pub(crate) tls_provider: Option<SharedTlsProvider>,
+    /// When `true`, the signature carries a bridgeable `stream<T>` handle, so
+    /// args/results are relocated across the boundary (see [`relocate`]) rather
+    /// than copied. When `false`, this is a plain-value call and the params are
+    /// copied directly.
+    pub(crate) relocate: bool,
+    /// Param/result types, captured at link time, used to drive the relocation
+    /// pass (empty for the plain-value path).
+    pub(crate) param_tys: Arc<[Type]>,
+    pub(crate) result_tys: Arc<[Type]>,
 }
 
 fn type_is_ephemeral_safe(ty: &Type) -> bool {
@@ -150,6 +161,42 @@ fn type_is_ephemeral_safe(ty: &Type) -> bool {
 pub(crate) fn func_is_ephemeral_safe(func_ty: &ComponentFunc) -> bool {
     func_ty.params().all(|(_, ty)| type_is_ephemeral_safe(&ty))
         && func_ty.results().all(|ty| type_is_ephemeral_safe(&ty))
+}
+
+/// Whether a type can cross an ephemeral-store boundary via [`relocate`]: it is
+/// handle-free, or every handle it contains is a `stream<T>` of a relocatable
+/// element type (nested anywhere in aggregates). `future`/`resource`/
+/// `error-context` handles are not relocatable.
+fn type_is_bridge_safe(ty: &Type) -> bool {
+    if !carries_cross_store_handle(ty) {
+        return true;
+    }
+    match ty {
+        Type::Stream(st) => st.ty().is_some_and(|e| bridgeable_element_type(&e)),
+        Type::Future(ft) => ft.ty().is_some_and(|e| bridgeable_element_type(&e)),
+        Type::List(t) => type_is_bridge_safe(&t.ty()),
+        Type::Option(t) => type_is_bridge_safe(&t.ty()),
+        Type::Tuple(t) => t.types().all(|t| type_is_bridge_safe(&t)),
+        Type::Record(t) => t.fields().all(|f| type_is_bridge_safe(&f.ty)),
+        Type::Variant(t) => t
+            .cases()
+            .all(|c| c.ty.is_none_or(|t| type_is_bridge_safe(&t))),
+        Type::Result(t) => {
+            t.ok().is_none_or(|t| type_is_bridge_safe(&t))
+                && t.err().is_none_or(|t| type_is_bridge_safe(&t))
+        }
+        Type::Map(t) => type_is_bridge_safe(&t.key()) && type_is_bridge_safe(&t.value()),
+        // future / resource (own/borrow) / error-context: not relocatable.
+        _ => false,
+    }
+}
+
+/// Whether every param/result of `func_ty` is [`type_is_bridge_safe`], so a call
+/// carrying a `stream<T>` can still run in an ephemeral store (with relocation)
+/// instead of being pinned to the shared store.
+pub(crate) fn func_is_bridge_safe(func_ty: &ComponentFunc) -> bool {
+    func_ty.params().all(|(_, ty)| type_is_bridge_safe(&ty))
+        && func_ty.results().all(|ty| type_is_bridge_safe(&ty))
 }
 
 async fn build_ctx_from_template(
@@ -379,7 +426,7 @@ pub(crate) async fn invoke_linked_async_export(
     inv: &LinkedExportInvocation,
 ) -> wasmtime::Result<()> {
     if let Some(ephemeral_call) = &inv.ephemeral_call {
-        invoke_ephemeral_linked_export(params, results, inv, ephemeral_call).await
+        invoke_ephemeral_linked_export(accessor, params, results, inv, ephemeral_call).await
     } else {
         invoke_shared_store_linked_export(accessor, params, results, inv).await
     }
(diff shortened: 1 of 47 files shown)


crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -141,6 +143,15 @@ pub(crate) struct EphemeralLinkedCall {
     pub(crate) linked_component_ids: Vec<Arc<str>>,
     #[cfg(feature = "wasi-tls")]
     pub(crate) tls_provider: Option<SharedTlsProvider>,
+    /// When `true`, the signature carries a bridgeable `stream<T>` handle, so
+    /// args/results are relocated across the boundary (see [`relocate`]) rather
+    /// than copied. When `false`, this is a plain-value call and the params are
+    /// copied directly.
+    pub(crate) relocate: bool,
```
vados-cosmonic, line 150: Sounds like we should make this an enum

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -141,6 +143,15 @@ pub(crate) struct EphemeralLinkedCall {
...
+    /// Param/result types, captured at link time, used to drive the relocation
+    /// pass (empty for the plain-value path).
+    pub(crate) param_tys: Arc<[Type]>,
+    pub(crate) result_tys: Arc<[Type]>,
```
vados-cosmonic, line 154: Could these be combined? Do we ever have one without the other?

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -152,6 +163,42 @@ pub(crate) fn func_is_ephemeral_safe(func_ty: &ComponentFunc) -> bool {
         && func_ty.results().all(|ty| type_is_ephemeral_safe(&ty))
 }
 
+/// Whether a type can cross an ephemeral-store boundary via [`relocate`]: it is
+/// handle-free, or every handle it contains is a `stream<T>` of a relocatable
+/// element type (nested anywhere in aggregates). `future`/`resource`/
+/// `error-context` handles are not relocatable.
```
vados-cosmonic, line 169: Can we make the first line a good summary, then rework this description into a list?

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -397,9 +444,185 @@ impl<T> Drop for AbortOnDrop<T> {
...
+    }
+}
+
+/// Run a `stream`-carrying async linked call in an ephemeral store, relocating
+/// args/results across the boundary (see [`relocate`]).
+///
+/// Args are extracted in the caller store, so each source stream begins pumping
+/// under the caller's long-lived runtime; the call then runs in a throwaway
+/// store, where result streams are extracted before the store is torn down. The
+/// store-driving task is **detached** (not [`AbortOnDrop`]): it must outlive
```
vados-cosmonic, line 470:
```suggestion
/// store-driving task is **detached** (leaked after initial [`AbortOnDrop`] wrapping): it must outlive
```

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -397,9 +444,185 @@ impl<T> Drop for AbortOnDrop<T> {
...
+    params: &[Val],
+    results: &mut [Val],
+    inv: &LinkedExportInvocation,
+    ephemeral_call: &Arc<EphemeralLinkedCall>,
+) -> wasmtime::Result<()> {
+    // Extract args in the caller store: source-stream pumps run under the
+    // caller's (long-lived) runtime, so their drain signals are dropped here.
+    let param_tys = Arc::clone(&ephemeral_call.param_tys);
+    let args = accessor.with(|mut access| -> wasmtime::Result<Vec<Relocated>> {
+        let mut dones: Vec<Done> = Vec::new();
```
vados-cosmonic, line 485: Do we have any expectation of how many dones we should get? Seems like params.len() + results.len()?

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -397,9 +444,185 @@ impl<T> Drop for AbortOnDrop<T> {
...
+            let v = relocate::inject(access.as_context_mut(), r)?;
+            *results.get_mut(i).context("result index out of bounds")? = v;
+        }
+        Ok(())
+    })?;
+
+    trace!(
+        name = %inv.import_name,
+        fn_name = %inv.export_name,
+        "invoked relocated ephemeral dynamic export"
```
vados-cosmonic, line 617:
```suggestion
        "successfully invoked relocated ephemeral dynamic export"
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -647,6 +667,77 @@ impl ResolvedWorkload {
         }
     }
 
+    /// Execute a p3 service that also exports a host-invoked handler (today
+    /// `wasi:http/handler`): one instance co-drives `cli/run` and the handler
```
vados-cosmonic, line 671:
```suggestion
    /// `wasi:http/handler` and `wasmcloud:messaging`): one instance co-drives `cli/run` and the handler
```

crates/wash-runtime/src/host/reactor.rs
```diff
@@ -0,0 +1,497 @@
+//! Reactor: a long-lived service instance that co-drives `wasi:cli/run`
```
vados-cosmonic, line 1: Yeah we probably gotta call this something different, given that reactor is quite taken -- that said, naming is obviously quite hard... Some suggestions:

- `HostService` ?
- `IngressService` ?
- `HostTriggeredService` ?


Taking a step back though, why can't this be unified with a "normal" service? I'm not sure this is a separate concept entirely.

crates/wash-runtime/src/host/reactor.rs
```diff
@@ -0,0 +1,497 @@
...
+/// A host-invoked handler export the Reactor serves, carrying the receiver end
+/// of its delivery channel. The paired sender is handed to the host-side ingress
+/// (the HTTP server, the messaging subscriber, ...) so it can deliver
+/// invocations to this live instance.
+pub enum Ingress {
+    /// `wasi:http/handler@0.3` — the HTTP server delivers requests here.
+    Http(tokio::sync::mpsc::Receiver<ServiceHttpJob>),
+    /// `wasmcloud:messaging/handler@0.2.0` — the messaging subscriber delivers
+    /// received messages here.
+    Messaging(tokio::sync::mpsc::Receiver<MessagingJob>),
```
vados-cosmonic, line 64: Let's split these impls up into modules below here, and also make this non-exhaustive

crates/wash-runtime/src/host/reactor.rs
vados-cosmonic: IMO if "trigger" is the right word, then we need to just call it that -- maybe `TriggerService`

crates/wash-runtime/src/host/reactor.rs
```diff
@@ -0,0 +1,497 @@
...
+                    });
+                }
+            }
+        }
+    }
+}
+
+/// A running service instance co-driving `cli/run` and its host-invoked handler
+/// exports on one instance.
+pub struct Reactor {
```
vados-cosmonic, line 157: IMO non-exhaustive with no pub members

crates/wash-runtime/src/host/reactor.rs
```diff
@@ -0,0 +1,497 @@
...
+/// Lift the `result<_, string>` returned by `handle-message`.
+fn lift_handle_result(v: Option<&Val>) -> Result<(), String> {
+    match v {
+        Some(Val::Result(Ok(_))) => Ok(()),
+        Some(Val::Result(Err(Some(boxed)))) => match &**boxed {
+            Val::String(s) => Err(s.clone()),
+            other => Err(format!("{other:?}")),
+        },
+        Some(Val::Result(Err(None))) => Err(String::new()),
+        other => Err(format!("unexpected handle-message result: {other:?}")),
```
vados-cosmonic, line 293: IMO we can have this be reusable, it's not related to handle-message specifically

crates/wash-runtime/tests/common/mod.rs
```diff
@@ -264,6 +264,25 @@ pub async fn start_host_with_p3(addr: &str) -> Result<(std::net::SocketAddr, imp
     Ok((bound_addr, host))
 }
 
+/// Like [`start_host_with_p3`] but also returns the [`HttpServer`], so a test can
+/// drive host-side ingress hooks directly (e.g. deliver a message to a reactor's
+/// messaging handler via `deliver_reactor_message`).
+pub async fn start_host_with_p3_handler(
```
vados-cosmonic, line 270: NIT: `start_host_with_p3_http_handler` ?

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -397,9 +444,185 @@ impl<T> Drop for AbortOnDrop<T> {
...
+                                )
+                            })?;
+                        let mut arg_vals = Vec::with_capacity(args.len());
+                        for a in args {
+                            arg_vals.push(relocate::inject(access.as_context_mut(), a)?);
+                        }
+                        Ok((func, arg_vals))
+                    })?;
+                    let mut results_buf = vec![Val::Bool(false); result_tys.len()];
+                    const CALL_TIMEOUT: Duration = Duration::from_secs(600);
```
vados-cosmonic, line 547: Maybe this should be configurable with a default from the jump here? We can cross this bridge when we get there though.

crates/wash-runtime/src/engine/relocate.rs
vados-cosmonic, line 1: We should put this under some module like `store` or `store_interop` or something

vados-cosmonic (approved):
LGTM 🚀 

Mostly light suggestions, but this is fantastic, this very nicely bridges component-facing services and host... ~~triggers~~ reactors

crates/wash-runtime/src/engine/linked_call.rs
ricochet: It's one per stream-typed param + one per stream-typed result. So params.len()+results.len() is the upper bound.

crates/wash-runtime/src/host/reactor.rs
ricochet: The only difference from Services as we have today is that these have P3 exports. I'm ok with TriggerService

crates/wash-runtime/src/engine/linked_call.rs
vados-cosmonic: Ah OK yeah since we know what size this is we can probably use `Vec::with_capacity` like you did further down IIRC

ricochet:
Incorporated review feedback into stacked prs starting with #5332

This PR specifically is replaced by #5336

PR #5315, opened 2026-07-02
feat: wasmcloud:postgres async

ricochet:
This creates a new version of wasmCloud postgres interface with async interfaces. The existing 0.1.1-draft surface is unchanged. A component may import either.

The biggest change is taking the same pattern from wasi:http@0.3:

```wit
query: async func(query: string, params: list<pg-value>) -> result<tuple<list<column-name>, stream<row>, future<result<_, error>>>, error>;
```

This shape allows us to stream rows and surface errors.
The outer `result<…, error>` surfaces immediate failure (bad SQL, connection dead).
`future<result<_, error>>`: the trailer reports an error that happens mid-stream. A bare `stream<row>` has nowhere to put a late error without it.

Additionally this changes to one non-exhaustive `error` variant with a structured `db-error`.
Instead of stringly unexpected(string), they carry a db-error record with the Postgres SQLSTATE code, severity, message, detail.


crates/wash-runtime/src/plugin/wasmcloud_postgres/async_p3.rs
```diff
@@ -0,0 +1,514 @@
...
+use super::PgId;
+#[cfg(feature = "wasm_component_model_implements")]
+use crate::plugin::multiplex::Multiplexer;
+
+/// How many fetched rows may sit buffered between the fetch task and the guest
+/// before the task blocks — bounds host memory and applies backpressure.
+const ROW_CHANNEL_CAPACITY: usize = 16;
+
+#[cfg(not(feature = "wasm_component_model_implements"))]
+mod bindings {
```
vados-cosmonic: We should probably start putting `bindings` in it's own file (`bindings.rs`)

crates/wash-runtime/src/plugin/wasmcloud_postgres/async_p3.rs
```diff
@@ -0,0 +1,514 @@
...
+        },
+    });
+}
+
+use bindings::wasmcloud::postgres0_2_0::types::{DbError, Error, PgValue, Row};
+use bindings::wasmcloud::postgres0_2_0::{prepared, query, types};
+
+/// The completion `future<result<_, error>>` end returned alongside a query's
+/// row stream.
+type ResultFuture = FutureReader<Result<(), Error>>;
```
vados-cosmonic: Maybe we should call it a `CompletionFuture` ?

crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
```diff
@@ -543,55 +576,105 @@ impl HostPlugin for WasmcloudPostgres {
...
-            self.component_databases
-                .write()
-                .await
-                .insert(component_id, database);
-            bindings::wasmcloud::postgres::query::add_to_linker::<_, SharedCtx>(
-                linker,
-                extract_active_ctx,
-            )?;
-            bindings::wasmcloud::postgres::prepared::add_to_linker::<_, SharedCtx>(
+        // ── sync `0.1.1-draft` (wasip2) ──────────────────────────────────────
```
vados-cosmonic:
```suggestion
```

crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
```diff
@@ -543,55 +576,105 @@ impl HostPlugin for WasmcloudPostgres {
...
-                extract_active_ctx,
-            )?;
-            bindings::named_imports::wasmcloud::postgres::prepared::add_to_linker::<_, SharedCtx>(
-                linker,
-                &component,
-                |name| self.mux.resolve(&registry, name),
-                extract_active_ctx,
-            )?;
+        // ── async `0.2.0` (wasip3) ───────────────────────────────────────────
+        if !pg_async.is_empty() {
```
vados-cosmonic, line 646:
```suggestion
        if pg_async.is_some() {
```

vados-cosmonic (dismissed):
LGTM 🚀 

Some nits but in general the updates to the interface and new plugin look good!

jfleitz (approved):

crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
ricochet: pg_async is a Vec, and Vec has no .is_some()

PR #5317, opened 2026-07-03
feat(runtime): Implement and enable epoch-interruption for invocation cancellation

Aditya1404Sal:


ricochet:
I'm excited about this. We'll want to also use this to yield for long-running/compute-heavy components that should periodically yield the async executor instead of starving it, so this scope should expand to be more than just cancellation IMO

1. Registrations leak. Register inserts into the map but only cancel removes. Needs an unregister-on-completion path and/or reaping in on_workload_unbind.
2. `registrations` is one map shared across all workloads, keyed by a guest-supplied string. Any component importing control can cancel(id) an invocation registered by a different workload if it knows the id. I think we should instead of the host own the cancellation identity (e.g. scoped by workload_id).
3. Needs the flag enabled in CI. FYI I've been opting to have a default-only test and all-the-flags test rather than re-running the tests multiple times for each flag. 
4. Epoch interruption only traps while the guest is executing wasm. A guest parked in a host call (a long wait-for, blocking host function, or awaiting host I/O) won't observe the trap until it re-enters wasm. It seems like we should additionally design for cancellation dropping the invocation/store future while relying on the epoch interruption for CPU-bound guests.

Aditya1404Sal:
> registrations is one map shared across all workloads, keyed by a guest-supplied string. Any component importing control can cancel(id) an invocation registered by a different workload if it knows the id. I think we should instead of the host own the cancellation identity (e.g. scoped by workload_id).

My concern with that is, since registrations live in-memory, how do we preserve request to workload routing? In the case of Autoscaling there will be multiple replicas and due to that, even if the host we're hitting, has that registration stored -- it won't trigger the cancellation just because that request was routed with a different workload-id.

How does the host know which request is part of what workload

ricochet:
The abort itself is host-local by nature. A running invocation and its task handle live in one process, so no replica can cancel a task it isn't running. Replicating registrations wouldn't help; the registration belongs on the host that owns the invocation. So the multi-replica case is a routing problem, not a state-sharing one: deliver the cancel to the host that owns the invocation, then it does the local abort.

"How does the host know which request is part of what workload?" It knows because it did the dispatch.

Aditya1404Sal:
> The abort itself is host-local by nature. A running invocation and its task handle live in one process, so no replica can cancel a task it isn't running. Replicating registrations wouldn't help; the registration belongs on the host that owns the invocation. So the multi-replica case is a routing problem, not a state-sharing one: deliver the cancel to the host that owns the invocation, then it does the local abort.
> 
> 
> 
> "How does the host know which request is part of what workload?" It knows because it did the dispatch.

Ah It's making sense to me now, I'll scope the registration per workload id

ricochet:
Actually I started a sketch of this since I also need something like this
with host components. I think we'll have a good bit of overlap.

Aditya1404Sal:
will continue this now since #5335 got merged

PR #5324, opened 2026-07-09
feat(wash): resolve WIT and wkg config

ricochet:
wash could not fetch WIT packages from a user's own OCI registry:

- The `wit.sources` OCI branch was dead code: detect_source_type only
  classifies a value as OCI when it contains a '/', but the value was
  then parsed as a bare registry authority, which rejects '/'. Every OCI
  source failed with "invalid uri character". Replace the conversion with
  oci_registry_mapping, which splits host from repository path and encodes
  any leading path as an OCI namespace prefix (RegistryMapping::Custom),
  so both `ghcr.io/ns/pkg` and deep paths resolve. Also drop the misuse
  of the target `@version` as part of the namespace-registry label.

- wash ignored the standard wkg config. load_config's (None, None) arm
  was an empty TODO, so `~/.config/wasm-pkg/config.toml` namespace
  registries were never read, hence "no registry configured for
  namespace" even when plain wkg worked. Load the global wkg config as a
  fallback, matching wkg, so WKG_CONFIG_FILE is no longer required.

- `wit.registries` was validated but never applied. Now wired up as OCI
  basic-auth (username + token) per registry host, add a `username` field,
  drop the bogus `https://wasm.pkg` default, and accept bare hosts as
  well as full URLs.

### DIFF
--- a/crates/wash/src/wit.rs
+++ b/crates/wash/src/wit.rs
@@ -53,7 +53,7 @@ impl WitConfig {
                     reg.url
                 ));
             }
-            if let Err(err) = reg.credentials() {
+            if let Err(err) = reg.basic_auth_credentials() {
                 errors.push(err.to_string());
             }
         }
@@ -113,7 +113,7 @@ impl WitRegistry {
 
     /// Basic-auth credentials for this entry, or `None` when unset. Errors when exactly one of
     /// `username`/`token` is set, since basic auth requires both.
-    fn credentials(&self) -> Result<Option<(&str, &str)>> {
+    fn basic_auth_credentials(&self) -> Result<Option<(&str, &str)>> {
         match (self.username.as_deref(), self.token.as_deref()) {
             (Some(username), Some(token)) => Ok(Some((username, token))),
             (None, None) => Ok(None),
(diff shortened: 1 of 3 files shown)


crates/wash/src/wit.rs
```diff
@@ -83,26 +89,60 @@ impl WitConfig {
...
 
+impl WitRegistry {
+    /// The registry host this entry configures.
+    fn registry(&self) -> Result<Registry> {
+        registry_authority(&self.url)
+    }
+
+    /// Basic-auth credentials for this entry, or `None` when unset. Errors when exactly one of
+    /// `username`/`token` is set, since basic auth requires both.
+    fn credentials(&self) -> Result<Option<(&str, &str)>> {
```
vados-cosmonic:
```suggestion
    fn basic_auth_credentials(&self) -> Result<Option<(&str, &str)>> {
```

crates/wash/src/wit.rs
```diff
@@ -83,26 +89,60 @@ impl WitConfig {
...
+                self.url
+            ),
+        }
+    }
+}
+
+/// Extract a wkg [`Registry`] (host`[:port]`) from a `wit.registries` `url`, accepting both a
+/// bare authority (`ghcr.io`) and a full URL (`https://ghcr.io`).
+fn registry_authority(url: &str) -> Result<Registry> {
+    if let Ok(parsed) = Url::parse(url)
```
vados-cosmonic: Did [`Url::authority`](https://docs.rs/url/2.5.8/url/struct.Url.html#method.authority) not work here?

vados-cosmonic (dismissed):
LGTM 🚀

crates/wash/src/wit.rs
ricochet: It does, refactored

vados-cosmonic (approved):

PR #5325, opened 2026-07-09
fix(wash-runtime): recognize P3 wasi:http/handler as host-served HTTP

ricochet:
A workload declaring the unified WASI P3 `wasi:http/handler` interface
failed to deploy: bind_plugins only stripped the P2 `incoming-handler`/
`outgoing-handler` names from the host-served set, so `handler` fell
through to plugin matching and failed with "no plugins found for
requested interfaces".

Add WitInterface::is_incoming_http_handler(), covering both the P2
incoming-handler and the unified P3 handler names, and use it in the
listener gate and DynamicRouter; include `handler` in the bind-time
host-served filter. Request dispatch was already export-driven via
targets_wasip3_http(), so only the binding/routing gates were stale.

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -1416,8 +1416,11 @@ impl UnresolvedWorkload {
         // This tracks which interfaces each component still needs to be bound
         let mut unmatched_interfaces: HashMap<IdFlavor, HashSet<WitInterface>> = HashMap::new();
         let host_interfaces = {
-            // filter out Plugins fulfilled by host
-            let http_iface = WitInterface::from("wasi:http/incoming-handler,outgoing-handler");
+            // filter out Plugins fulfilled by host. `handler` is the unified
+            // WASI P3 http interface (covers both incoming and outgoing);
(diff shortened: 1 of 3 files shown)


vados-cosmonic (dismissed):
LGTM 🚀 

Super low priority but tests that use a 2.x interface (usually 0.2.2) but actually export a 0.3.0 interface should probably also be updated at some point -- do we want it to be an error if a workload has an interface that isn't what it's exporting?

Maybe the right move is actually just removing the version from the test interfaces since the version doesn't really matter as much as we think

vados-cosmonic (approved):

ricochet:
> Super low priority but tests that use a 2.x interface (usually 0.2.2) but actually export a 0.3.0 interface should probably also be updated at some point -- do we want it to be an error if a workload has an interface that isn't what it's exporting?
> 
> Maybe the right move is actually just removing the version from the test interfaces since the version doesn't really matter as much as we think

Removing version from tests that aren't meant for testing version matching makes sense to me. That'd also help greatly reduce churn.

On the error question, I think a deploy-time warning might be better to start since we likely have working manifests in the wild that will break. I'm going to start with a separate PR for the tests refactor

PR #5329, opened 2026-07-13
BugFix: NATS keys > batch size returns a valid cursor

jfleitz:
## Description
While load testing a configuration that uses NATS subjects for communicating between WorkloadDeployments, one of the WorkloadDeployments stopped processing due to no cursor being returned when there was >1000 keys in the NATS subject. This occurred when spamming the requests in to create a high volume.

This happens when there is an error with unwrapping the key from the stream, which the error would then impact the "skip cursor count" and shifting the page by 1.

## Fix
This fix moves the "skip cursor count" tracking to after unwrapping the key value. This ensures that if there was an error, the stream processing while loop returns immediately, and thus preserving the rest of the stream values (vs just eating them and now offset by the error count), so that the next call will have the cursor at the next location. 

## Testing
This was tested with 4 wasm workloads using NATS to communicate processing of incoming request. I load script was used to push a total 5000 requests in at 300 concurrently. Prior to the change the load script would hang until timeout and would have values "missing in pipeline" reported.

Test architecture used:
```
HTTP ?message=red,green,blue
        │
        ▼
    gateway (WorkloadDeployment w/k8s Service)
        │  publish
        ▼
    NATS subject1
        │
        ▼
    subject1-handler (WorkloadDeployment)
        │  stores values in KV bucket "nats-example" under generated keys
        │  publishes the keys to subject2
        ▼
    NATS subject2
        │
        ▼
    subject2-handler (WorkloadDeployment)
        │  retrieves the values by key from the KV bucket
        │  publishes the resolved key=value pairs
        ▼
    NATS processed
        │
        ▼
    monitor (WorkloadDeployment w/k8s Service)
           stores every message in KV bucket "processed-log"
           serves http://monitor.localhost.cosmonic.sh/ — a live page
           listing the messages (polls /messages JSON every 2s)
```


ricochet (approved):

jfleitz:
This one is ready for review. My pause on if this is a performance hit on nats messaging with [skipping message](https://github.com/wasmCloud/wasmCloud/pull/5329/changes#diff-c068b86837ec45de1d92fc11e4ef01067590d9033727db8e66bd1e4174f6c440R57-R60) to get to the right cursor location. Since the current version does do a `.skip(cursor_count)` [today](https://github.com/wasmCloud/wasmCloud/pull/5329/changes#diff-c068b86837ec45de1d92fc11e4ef01067590d9033727db8e66bd1e4174f6c440L263), but can silently stop processing if there is an error with unwrapping. So even if there was an performance hit, preventing the messaging to silently fail is more important.

PR #5330, opened 2026-07-13
fix(wash-runtime): version-tolerant matching for messaging

ricochet:
The in-memory and NATS messaging plugins decided whether to track a
component as a subscriber via HashSet::contains with a WitInterface
pinned to handler@0.2.0. A component built against handler@0.2.1, or one
whose handler export merged into a wider entry, was silently never
subscribed even though the bindgen call path resolves exports
semver-compatibly within 0.2.x.

Add exports_messaging_handler() to the messaging module, matching via
WitInterface::contains with a versionless probe, and use it in both
backends.

Also drop versions from test interface declarations where version
matching is not the subject under test, so tests survive fixture
rebuilds. This makes test_missing_interface_fails real: its declared
blobstore@0.2.0 never matched the fixture's 0.2.0-draft import, so
binding silently succeeded and nothing was asserted; versionless it
matches, binding fails, and the test now asserts the error.

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -2323,20 +2323,15 @@ mod tests {
     /// Verifies that `on_workload_bind` is called before `on_workload_item_bind`.
     #[tokio::test]
     async fn test_single_plugin_single_component() {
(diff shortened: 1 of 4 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5332, opened 2026-07-14
feat(wash-runtime): trigger service

ricochet:
A p3 service that exports both wasi:cli/run@0.3 and wasi:http/handler@0.3 is now driven on a single instance under one Store::run_concurrent: cli/run runs as a background task while each inbound HTTP request is served as a concurrent per-invocation task on the same instance. This allows the long-running work and the handlers to share in-memory state, instead of the host instantiating a component per request.

The TriggerService is deliberately general: each host-invoked export is an Ingress carrying the receiver of its delivery channel, prepared into a binding view before `run_concurrent` and served by spawning one task per invocation. Adding another host-invoked interface later is a new Ingress variant plus a serve arm.

The HTTP server keeps a ServiceHandlers registry (workload id -> delivery channel): a resolved trigger-service workload registers its sender (HostHandler::on_service_http_resolved), inbound requests for it are delivered to the live instance, and stop/unbind drops the registration. Responses stream incrementally, the head is delivered as soon as the handler returns it, body frames follow through a small bounded channel that back-pressures the guest, with a configurable overall timeout backstop (WASH_HTTP_RESPONSE_TIMEOUT_SECS, the first entry in the new env-tunable timeouts module).

Supervision: on a fault (e.g. a guest trap in cli/run) the driver re-instantiates into the same store with fresh ingress channels, re-registering their senders, within the service's existing max_restarts budget.

The svc-counter fixture exports both cli/run (a tick loop) and http/handler (a counter); the integration test proves one instance co-drives both: cli_ticks grows across requests while http_calls increments exactly once per request on
the same instance.

This is the first in a series of stacked PRs.

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -315,6 +315,18 @@ impl WorkloadService {
         Ok(command)
     }
 
+    /// Pre-instantiate the raw component, leaving binding-view construction
+    /// (e.g. both cli `Command` and http `Service`) to the caller. Used when a
+    /// p3 service also serves HTTP and must drive both exports on one instance.
+    pub fn pre_instantiate_raw(
+        &self,
+    ) -> anyhow::Result<wasmtime::component::InstancePre<SharedCtx>> {
+        Ok(self
+            .metadata
+            .linker
+            .instantiate_pre(&self.metadata.component)?)
+    }
+
     /// Whether or not the service is currently running.
     pub fn is_running(&self) -> bool {
         self.handle.is_some()
@@ -535,6 +547,26 @@ impl std::fmt::Debug for ResolvedWorkload {
     }
 }
 
+/// Build a trigger service's host-invoked ingresses, returning them alongside the
+/// paired senders to register with the host-side ingresses. Called once per
+/// incarnation (start and each restart) so a restarted service gets fresh
+/// channels whose senders replace the stale registrations.
+#[allow(clippy::type_complexity)]
+fn build_trigger_ingresses(
+    serves_http: bool,
+) -> (
+    Vec<crate::host::trigger_service::Ingress>,
+    Option<tokio::sync::mpsc::Sender<crate::host::http::ServiceHttpJob>>,
+) {
+    let mut ingresses = Vec::new();
+    let http_tx = serves_http.then(|| {
+        let (tx, rx) = tokio::sync::mpsc::channel(256);
+        ingresses.push(crate::host::trigger_service::Ingress::Http(rx));
+        tx
+    });
+    (ingresses, http_tx)
+}
+
 impl ResolvedWorkload {
     /// Executes the service, if present, and returns whether it was run.
     #[instrument(name="execute_service", skip_all, fields(workload.id = self.id.as_ref(), workload.name = self.name.as_ref(), workload.namespace = self.namespace.as_ref()))]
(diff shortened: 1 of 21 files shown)


crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1297,6 +1427,15 @@ impl ResolvedWorkload {
             }
         }
 
+        // A trigger service registered its HTTP handler at start
+        // (`execute_trigger_service`); drop that registration on stop so it no
+        // longer receives host-invoked deliveries on a torn-down instance.
+        if self.service.is_some()
+            && let Err(e) = self.http_handler.on_service_http_unbind(self.id()).await
+        {
+            tracing::warn!(workload.id = %self.id(), err = %e, "failed to unbind service HTTP handler, continuing");
```
```suggestion
            tracing::error!(workload.id = %self.id(), err = %e, "failed to unbind service HTTP handler, continuing");
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -837,10 +887,29 @@ impl<T: Router, O: OutgoingHandler> HostHandler for HttpServer<T, O> {
...
         Ok(())
     }
 
+    async fn on_service_http_resolved(
+        &self,
+        workload_id: &str,
+        sender: tokio::sync::mpsc::Sender<ServiceHttpJob>,
+    ) -> anyhow::Result<()> {
+        self.router.on_service_http_resolved(workload_id).await?;
+        self.service_handlers
```
vados-cosmonic: Is it possible to have this fire twice in a row (i.e. overwrite some mapping)? 

I *think* based on what I've read so far the answer is no (we should have unbind happen before this), we might want to preserve that as a invariant here so we know if that's ever the case.

crates/wash-runtime/src/timeouts.rs
```diff
@@ -0,0 +1,46 @@
...
+                );
+                default_secs
+            }
+        },
+        Err(_) => default_secs,
+    };
+    Duration::from_secs(secs)
+}
+
+macro_rules! timeout {
```
vados-cosmonic: Le's leave a longer comment here about what kind of functions this generates and the mapping, and what this *should* be used for. 

i.e. `timeout::http_response()` is generated, and it takes some env var and attempts to look it up, etc.

I also wonder if we should not be assuming `std::env::var` here in favor of something more generic (i.e. `impl FnOnce() -> String`), but that might be over-engineering it at this point, even if there's a minor benefit in testability.

vados-cosmonic (dismissed):
LGTM 🚀 

I think we may need to start factoring some of the magic numbers hanging around but I think that can be guided by usage (once we run into one that is too small or should be tweakable for perf, etc)

crates/wash-runtime/src/timeouts.rs
ricochet: I think adding an abstraction here is probably too early since it was pretty easy to get working with tests as-is, but definitely open to refactoring later. Expanded out the doc here

ricochet: I think I'd opt for doing the same thing as crates/wash/src/config.rs with figment.

crates/wash-runtime/src/host/http.rs
ricochet: Yep, expanded docs here. A re-resolve without unbind is the expected restart path. Stop is the only path that unbinds.

PR #5333, opened 2026-07-14
feat(wash-runtime): messaging ingress for trigger services

ricochet:
Adds wasmcloud:messaging/handler@0.2.0 as a second TriggerService ingress: a p3 service exporting the p2 handle-message alongside cli/run (and optionally wasi:http/handler) has messages delivered to its ONE long-lived instance, invoked via the dynamic call_concurrent path instead of a fresh instance per message. The host keeps a MessagingHandlers registry mirroring the HTTP one (register on resolve, swap on restart, drop on stop/unbind), with deliver_trigger_service_message / has_trigger_service_messaging as the delivery API for host-side messaging ingresses.

Trap handling: a handler Err(string) is an ordinary application outcome, reported to the deliverer only. A guest TRAP, however, leaves the shared instance unenterable for every later invocation, so MessagingTask reports it and then faults the driver out of run_concurrent. The service supervisor restarts within max_restarts and now rebuilds a FRESH store per incarnation (via a captured ServiceStoreRecipe): a trapped store can never re-enter any component instance, so reusing it would leave every restarted incarnation permanently broken. The plain p3 cli/run restart loop (execute_service_p3) had the same latent same-store-reuse pattern and now rebuilds from the same recipe on every restart, which also stops stale instances accumulating in a reused store across clean restarts.

The msg-counter fixture co-exports cli/run + http/handler + the messaging handler; tests cover count persistence across deliveries on one instance, the HTTP ingress observing the messaging ingress's state (multi-ingress on one instance), deregistration on stop, and trap -> restart -> re-subscribe with the per-instance count reset to prove the incarnation is fresh.

Second PR in series of stacked PRs and builds on #5332

### DIFF
--- a/crates/wash-runtime/src/host/http.rs
+++ b/crates/wash-runtime/src/host/http.rs
@@ -27,6 +27,7 @@ use std::{
 };
 
 use crate::host::allowed_hosts::AllowedHost;
+use crate::host::trigger_service::{BrokerMessage, MessagingJob};
 use crate::{engine::ctx::SharedCtx, observability::Meters};
 use crate::{engine::workload::ResolvedWorkload, observability::FuelConsumptionMeter};
 use anyhow::{Context, ensure};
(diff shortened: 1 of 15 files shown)


crates/wash-runtime/src/host/http.rs
```diff
@@ -910,6 +949,55 @@ impl<T: Router, O: OutgoingHandler> HostHandler for HttpServer<T, O> {
         Ok(())
     }
 
+    async fn on_trigger_service_messaging_resolved(
```
vados-cosmonic, line 958: Almost certainly something for a follow up, but maybe this shouldn't be called the `HttpServer` any more? Your guess is as good as mine on what the name should be though (`Ingress{Server, Runner, ...}`)

vados-cosmonic (dismissed):
LGTM 🚀 

This is very clean on top of the other PR

crates/wash-runtime/src/host/http.rs
ricochet: I like Ingress, but yup will opt to do that as a follow-up

PR #5334, opened 2026-07-14
feat(wash-runtime): cross-store streams and futures

ricochet:
Extends the ephemeral linked-call path so a call whose signature carries p3 stream<T>/future<T> handles (nested anywhere in aggregates) can still run in a throwaway store instead of being pinned to the caller's shared store: each argument is extracted in the caller store into a store-agnostic Relocated tree and injected into the callee store, and results come back the same way. Handle-free values/subtrees are copied wholesale; each stream/future becomes a
live, no-buffering pump (a bounded channel that back-pressures the source), so a body streams incrementally rather than buffering at the boundary. `resource` and `error-context` handles are not relocatable between ephemeral stores; a signature carrying either stays on the shared-store path.

Classification happens at link time (func_is_bridge_safe) and only for a p3-service workload, where it matters: a long-lived service store must never be pinned or frozen by a stream-carrying backend call. The ephemeral store driving a relocated call is detached once results are in hand so it can keep feeding result streams while the caller drains them (bounded by a drain timeout), and it self-terminates if the caller drops its readers. Call timeouts move to the env-tunable timeouts module and error messages report the actual configured duration.

The bridge-backend/bridge-service fixtures and integration suite cover: handle-free calls, backend statelessness, streams in both directions, a non-u8 element type, a stream nested in a variant, futures, empty streams both directions and the payload-less variant case, ~4 MiB through the bounded channel (backpressure), early reader drop, client cancellation mid-call, and service responsiveness while a backend busy-spins in its own store.

Third PR in a series of stacked PRs for the work towards Host Component Plugins.

### DIFF
--- a/crates/wash-runtime/src/engine/linked_call.rs
+++ b/crates/wash-runtime/src/engine/linked_call.rs
@@ -19,7 +19,6 @@
 
 use std::collections::HashMap;
 use std::sync::Arc;
-use std::time::Duration;
 
 use tokio::sync::RwLock;
 use tokio::time::timeout;
@@ -35,6 +34,8 @@ use wasmtime_wasi::WasiCtxBuilder;
 #[cfg(feature = "wasi-tls")]
 use crate::engine::ctx::SharedTlsProvider;
 use crate::engine::ctx::{AccessorActiveCtxGuard, Ctx, SharedCtx, StoreActiveCtxGuard};
+use crate::engine::store::relocate::{self, Relocated, bridgeable_element_type};
+use crate::engine::store::stream_pump::Done;
 use crate::engine::value::{carries_cross_store_handle, lift_results, lower_params};
 use crate::engine::volumes::{ResolvedVolumeMount, resolve_component_volume_mounts_in_map};
 use crate::engine::workload::{WorkloadComponent, WorkloadMetadata};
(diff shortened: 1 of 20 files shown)


crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -141,6 +142,23 @@ pub(crate) struct EphemeralLinkedCall {
...
+    pub(crate) mode: EphemeralCallMode,
+}
+
+/// How an ephemeral linked call transfers its args/results across the store
+/// boundary, decided by the signature classification at link time.
+#[derive(Clone)]
+pub(crate) enum EphemeralCallMode {
+    /// Handle-free call: params/results are copied directly.
+    PlainValue,
+    /// The signature carries a bridgeable `stream<T>`, so args/results are
```
```suggestion
    /// The signature carries a bridgeable `stream<T>` or `future<T>`, so args/results are
```

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -397,9 +459,214 @@ impl<T> Drop for AbortOnDrop<T> {
...
+                &mut dones,
+            )?);
+        }
+        Ok(out)
+    })?;
+
+    let (ready_tx, ready_rx) =
+        futures::channel::oneshot::channel::<wasmtime::Result<Vec<Relocated>>>();
+    let ephemeral_call = Arc::clone(ephemeral_call);
+    let call_pre = inv.pre.clone();
```
```suggestion
    let caller_pre = inv.pre.clone();
```

crates/wash-runtime/src/timeouts.rs
```diff
@@ -40,6 +40,19 @@ macro_rules! timeout {
     };
 }
 
+timeout! {
```
vados-cosmonic: This macro should be expanded to take multiple forms, probably and be named something like `declare_timeouts`?

crates/wash-runtime/tests/fixtures/bridge-service/wit/world.wit
```diff
@@ -0,0 +1,10 @@
+package wasmcloud:bridgesvc@0.1.0;
```
vados-cosmonic, line 1: ?? Did I miss where this gets used?

vados-cosmonic (dismissed):
LGTM 🚀 

Pretty gnarly code here, great job

crates/wash-runtime/tests/fixtures/bridge-service/wit/world.wit
vados-cosmonic: NVM I see where it's auto pulled in!

crates/wash-runtime/src/engine/linked_call.rs
ricochet: opting for callee here since this is from the export

vados-cosmonic: Ah yeah I definitely thought this was the caller but that makes sense (the caller should already be running as it did the call)

PR #5335, opened 2026-07-14
feat(wash-runtime): host component plugins

ricochet:
A host component plugin is a WebAssembly component that provides a host capability (a workload's import) from its own long-lived, supervised store, rather than as Rust code in the caller's store. ComponentHostPlugin flows through the existing bind_plugins machinery. The world() derived from the component's exports, start() spins up the persistent store as a TriggerService with a new Ingress::Capability, and on_workload_item_bind installs func_new_concurrent routing shims (not add_to_linker) forwarding each call across the store boundary to the persistent instance via the relocation layer.

Resources cross as proxies: own<r> returns become an opaque proxy in the caller (the plugin store keeps the real in a per-store ResourceRegistry), borrow<r> and method calls route to the real resource, and dropping the proxy frees it. The driver steps out of run_concurrent to run the destructor, which needs top-level store access, and resumes the preserved in-flight tasks.

Overlap with #5317: Per-caller partitioning via a `wasmcloud:host/identity` import: a plugin resolves the calling workload from its own root guest task (async_call_stack), so ambient caller identity is exact under concurrent, interleaved calls. Cooperative per-invocation cancellation via wasmcloud:host/cancel (current-job / request-cancel / is-cancelled), backed by a per-incarnation JobRegistry; wasmtime cannot hard-cancel a guest call_concurrent subtask short of dropping the whole store, so request-cancel marks the job and the guest unwinds itself. A job may be cancelled only by a requester from its owner's workload, and RAII guards retire per-call state safely under guest-task-id reuse.

Reliability: one supervised store per plugin, rebuilt fresh on a fault within a bounded, backoff-paced restart budget, swapping the capability channel and job registry per incarnation (lock-free reads via arc-swap); a non-blocking atomic in-flight ceiling bounds concurrency and re-entrant depth (self-imports are wired back to the plugin itself) with no bounded-pool deadlock; error-context values are rejected with a pointer at plain error types.

Canonical wasmcloud:host WIT lives at wit/host/ (with its wkg.lock), pinned to every vendored fixture copy by a CI parity check. Everything is behind the new `host-component-plugins` feature.

The kv-plugin/plugin-caller fixtures and integration suite cover: handle-free round-trips, sharing and per-caller partitioning across genuinely distinct workloads (host-header routed), partitioning under concurrent interleaving, both identity halves, streams and futures across the persistent-store boundary, resource round-trip/isolation/exactly-once reclamation under churn, re-entrancy and the in-flight ceiling, no head-of-line blocking, trap recovery including in-flight calls failing promptly, restart-budget exhaustion, and cancellation (baseline, truncation, tenant isolation, stale-job refusal).

Closes #5018

4th in a series of stacked PRs.


wit/host/wit/cancel.wit
```diff
@@ -0,0 +1,24 @@
...
+
+    /// The id of the invocation the caller is currently running under, or `0` if
+    /// none. A guest stashes this so another caller can later cancel it.
+    current-job: func() -> job-id;
+
+    /// Ask the invocation with `job` to cancel; returns whether the request was
+    /// accepted (not whether it has stopped). Accepted only for a caller in the
+    /// same workload as the job's owner. The target observes the request (via
+    /// `is-cancelled`, or a dropped stream reader) and unwinds itself.
+    request-cancel: func(job: job-id) -> bool;
```
vados-cosmonic, line 19: Two thoughts: 

* We could have `list<job-id>` here and save ourselves some calls
* Is there any case where someone would want to *wait* for a cancellation of one or more in-flight invocations?

crates/wash-runtime/Cargo.toml
```diff
@@ -47,6 +47,14 @@ wasi-blobstore = []
...
 # Component-model `(implements ..)` / `named_imports` multiplexing
 wasm_component_model_implements = []
+# Host component plugins: a component providing a host capability from its own
+# long-lived, supervised store, serving cross-store capability calls. Gates the
+# `ComponentHostPlugin` adapter, the trigger service's `Capability` ingress, and
+# the per-call job registry. The cross-store relocation layer it builds on
+# (`engine::store::{relocate,resource_bridge,stream_pump}`) is shared with the
+# ephemeral bridge and stays compiled; its resource-drop lifecycle is dormant
+# unless a plugin store activates it.
+host-component-plugins = []
```
vados-cosmonic, line 53:
```suggestion
# Host component plugins: a component providing a host capability from its own
# long-lived, supervised store, serving cross-store capability calls. 
host-component-plugins = []
```
This probably isn't the right place for this amount of detail -- maybe add something like "see: <some module/struct/etc>"

crates/wash-runtime/tests/fixtures/Cargo.toml
```diff
@@ -43,6 +43,8 @@ members = [
     "msg-counter",
     "bridge-backend",
     "bridge-service",
+    "kv-plugin",
+    "plugin-caller",
```
vados-cosmonic:
```suggestion
    "kv-plugin-caller",
```

crates/wash-runtime/tests/fixtures/.gitignore
```diff
@@ -11,6 +11,8 @@
 !msg-counter/wit/deps/
 !bridge-backend/wit/deps/
 !bridge-service/wit/deps/
+!kv-plugin/wit/deps/
+!plugin-caller/wit/deps/
```
vados-cosmonic:
```suggestion
!kv-plugin-caller/wit/deps/
```

crates/wash-runtime/src/timeouts.rs
```diff
@@ -57,3 +57,20 @@ timeout! {
     /// Max wall-clock for a trigger service to produce an HTTP response.
     http_response = ("WASH_HTTP_RESPONSE_TIMEOUT_SECS", 600)
 }
+timeout! {
```
vados-cosmonic: Same note I left here about `declare_timeouts!` -- I think we could collapse all these

crates/wash-runtime/src/host/trigger_service/capability.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub(super) in_flight: InFlightGuard,
+    /// Retires this call's registry job (and its task binding) when the call
+    /// completes or the task is otherwise dropped (e.g. store teardown). Held
+    /// across the `.await` so the retire runs however the task ends.
+    pub(super) job_guard: JobGuard,
+}
+
+/// Decrements a plugin store's in-flight capability-call counter on drop, so a
+/// slot is reclaimed whether the task completes normally or is cancelled.
+pub(super) struct InFlightGuard(Arc<AtomicUsize>);
```
vados-cosmonic, line 117: Any reason in particular we're not using a [Semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html) here? This looks lighter and definitely isn't obviously wrong or anything, but feels like we're reinventing here

vados-cosmonic (dismissed):
LGTM 🚀 

This was a big one but everything looks just fine -- cross-store calls for resources is awesome to see.

crates/wash-runtime/src/host/trigger_service/capability.rs
ricochet: Addressed this question in the docs for MAX_INFLIGHT_CAPABILITY_CALLS since its a good one and will probably come up again. 

This is deliberately not a [`tokio::sync::Semaphore`]: A semaphore `acquire().await` would let an ancestor hold a permit while suspended on a re-entrant sub-call that can never acquire one — a self-deadlock. Now a `try_acquire` semaphore would behave like this counter but carries a waiter queue and fairness machinery we wouldn't use, so a plain atomic is better.

wit/host/wit/cancel.wit
ricochet: We want cancellation to be cooperative and wasmtime can't hard-cancel a guest call_concurrent subtask (yet). This is the main reason request-cancel returns acceptance, not completion, and there's no reliable "await cancelled" signal to offer. Batch list<job-id> is a reasonable ergonomic add (each cancel is just a cheap registry mark), so I think right now I'd prefer leaving it and revisiting once wasmtime lands subtask cancellation.

crates/wash-runtime/src/host/trigger_service/capability.rs
vados-cosmonic: Ah thanks, that's a good point, thanks!

.github/scripts/wit-host-parity.mjs
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env node
```
ricochet: after refactoring how we vendor wit, I'm going to remove this

jfleitz (approved):

PR #5336, opened 2026-07-14
feat(wash-runtime): route inbound messages to long-lived trigger services

ricochet:
The messaging backends (in-memory and NATS) only tracked handler *components* and gave each message a fresh per-message instance. A long-lived trigger service that exports wasmcloud:messaging/handler@0.2.0 was never subscribed, so published messages never reached it. Track handler services alongside components and deliver to a running service when one is registered, so a message lands on the service's single long-lived instance (preserving its
in-memory state) instead of a throwaway one.

- Generalize WorkloadTracker::add_component to take &WorkloadMetadata; both
  WorkloadComponent and WorkloadService deref to it, so existing
  &WorkloadComponent callers coerce unchanged. Drop the WorkloadItem::Component
  guard in both backends so a handler service is set up as a subscriber too.
- A service has no per-component instance to pre-instantiate, so MessagingPre
  becomes optional: instantiate_pre failure is not fatal for a service (its
  receive loop delivers to the service); only components keep a MessagingPre
  for per-message work. A workload with neither a service nor a pre drops the
  message with a warning instead of panicking.
- In both receive loops, when has_trigger_service_messaging holds, forward the
  broker message via deliver_trigger_service_message (the API added with the
  messaging ingress) and skip per-message instantiation. This restores the
  ResolvedWorkload::http_handler() accessor as its intended consumer.
- Add InMemoryMessaging::publish so a host or test can inject a message through
  the normal subscriber routing without a component-side consumer.publish.

`test_trigger_service_messaging_via_in_memory_backend` publishes through the in-memory backend and reads the count over the service's HTTP handler, proving the message was handled on the co-driven service instance (0 -> 1) rather than a fresh instance per message.

### DIFF
--- a/crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
+++ b/crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
@@ -9,7 +9,7 @@ use crate::wit::{WitInterface, WitWorld};
 use anyhow::Context;
 use opentelemetry::KeyValue;
 use tokio::sync::{Notify, RwLock, oneshot};
-use tracing::{Instrument, debug, instrument, warn};
+use tracing::{Instrument, debug, instrument, trace, warn};
 
 const PLUGIN_MESSAGING_MEMORY_ID: &str = "wasmcloud-messaging-memory";
 const MAX_QUEUE_SIZE: usize = 10000;
@@ -131,6 +131,28 @@ impl InMemoryMessaging {
             tracker: Arc::new(RwLock::new(WorkloadTracker::default())),
             meters: Default::default(),
         }
+    }
+
+    /// Route a message to a workload's subscribers, exactly as an inbound publish
+    /// would: it is enqueued to every component whose subscriptions match
+    /// `subject`, then processed by that component's receive loop. Lets a host or
+    /// test inject a message without a component-side `consumer.publish`.
+    pub async fn publish(
+        &self,
+        workload_id: &str,
+        subject: &str,
+        body: Vec<u8>,
+    ) -> Result<(), String> {
+        route_to_subscribers(
+            self,
+            workload_id,
+            &types::BrokerMessage {
+                subject: subject.to_string(),
+                reply_to: None,
+                body,
+            },
+        )
+        .await
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -351,11 +372,20 @@ impl HostPlugin for InMemoryMessaging {
...
+        // pre-instantiate; its receive loop delivers to the running service
+        // instead. Only components get a `MessagingPre` for per-message work.
+        let pre = match workload.instantiate_pre(component_id).await {
+            Ok(instance_pre) => Some(
+                bindings::MessagingPre::new(instance_pre)
+                    .map_err(anyhow::Error::from)
+                    .context("failed to instantiate messaging pre")?,
+            ),
+            Err(e) => {
+                debug!(component_id, error = %e, "no per-message instance (long-lived service); messages delivered to the service");
```
vados-cosmonic: I think this is going to get really noisy maybe -- trace? There's a message down below IMO that would basically convey the same info

vados-cosmonic (dismissed):
LGTM 🚀

crates/wash-runtime/tests/integration_trigger_service_messaging.rs
```diff
@@ -221,3 +228,116 @@ async fn test_trigger_service_restarts_and_resubscribes_on_fault() -> Result<()>
...
+    anyhow::ensure!(resp.status().is_success(), "GET /count: {}", resp.status());
+    let body = resp.text().await?;
+    Ok(common::json_u64_field(&body, "count"))
+}
+
+/// End-to-end: a message published through the in-memory messaging backend is
+/// delivered to the long-lived trigger service (not a fresh per-message
+/// instance), so the count observed over the service's HTTP handler advances.
+#[tokio::test]
+async fn test_trigger_service_messaging_via_in_memory_backend() -> Result<()> {
```
vados-cosmonic, line 291: can we get a nats test here too via `testcontainers`? Not in this file of course, but would be nice to have

vados-cosmonic (approved):

PR #5348, opened 2026-07-15
P3 host plugin loader

ricochet:
#5335 added the `ComponentHostPlugin` adapter, a WebAssembly component that provides a host capability from its own supervised store but nothing could actually *load* one outside of tests. This PR adds the production loading path across all three host front-ends, plus an end-to-end test.

A host component plugin is loaded exactly like a built-in Rust plugin (`HostBuilder::with_plugin`, started at `Host::start`); the only new work is turning a **declaration** (an id + a place to fetch the wasm) into a registered plugin.

**Runtime primitive** (`wash-runtime`)
- `ComponentPluginSpec` / `PluginSource` — a plain, always-compiled declaration (id + OCI-image-or-file source + optional `max_restarts` / digest pin), with a `FromStr` for the CLI grammar.
- `load_component_plugin` (gated on `host-component-plugins`) — fetches the wasm (`oci::pull_component` or `tokio::fs::read`), verifies an optional digest pin, and builds the `ComponentHostPlugin`. The caller registers it with `with_plugin` before `Host::start`.

**`wash host`** (cluster / operator path)
- `--host-plugin` / `WASH_HOST_PLUGINS` (repeatable). Comma-separated `key=value`: `id=<name>,image=<ref>[,pull=…][,max-restarts=N][,digest=…]` or `id=<name>,file=<path>`.

**`wash dev`** (local path)
- `dev.host_plugins` config entries (`file` or `image` source).

**Chart** (operator path)
- Per-host-group `runtime.hostGroups[].hostPlugins` values, rendered into `--host-plugin` args on the host Deployment.

```yaml
# wash dev — .wash/config.yaml
dev:
  host_plugins:
    - id: acme-kv
      file: ./build/kv_plugin.wasm
      maxRestarts: 3

# chart — values.yaml
runtime:
  hostGroups:
    - name: default
      hostPlugins:
        - id: acme-kv
          image: ghcr.io/acme/kv-host:1.0.0
          pullPolicy: ifNotPresent
```

## Design decisions

- **Host-operator controlled only.** A host component plugin runs with the host's capability surface and serves *every* workload that imports its interface — a privileged install. So it's declared in host config; **nothing in a `WorkloadStartRequest` can register a host-global provider.**
- **Static at host start.** There is no dynamic-plugin API (`HostApi` is only heartbeat/workload_start/status/stop) and the registry is fixed after `build()`. Loading resolves once at boot. Dynamic load/unload on a running host is noted as future work (needs a mutable registry, workload re-bind, and an `Arc<str>` id).
- **All-features-only, and it fails loud.** The loader is gated on `host-component-plugins` (not in release builds). The spec/flag/config types are *always* compiled, so a plugin declared against a release host produces a clear "requires the feature" error rather than a silent drop — important for the operator/env path.
- **`Host` CRD is status-only.** It's created from the host's heartbeat (`HostPodReconciler` only bridges pod↔Host lifecycle); host pods are Helm-templated per host-group. So the declarative host-config seam is the host-group values, which is where `hostPlugins` lives.
- **`&'static str` id** is interned via a bounded `Box::leak` (plugins load once at startup and live for the process); flagged to revisit if dynamic loading lands.

## Testing

- Unit: `ComponentPluginSpec` grammar parsing (valid/invalid).
- Integration: `load_component_plugin` from a file builds a plugin with the right id/world; digest-pin-on-file and missing-file error paths; and an opt-in OCI-source test (`HOST_PLUGIN_OCI_REF`, self-skips in CI).
- Config: `HostPluginConfig::to_spec` validation + `dev.host_plugins` YAML round-trip.
- **e2e** (`runtime-operator/test/e2e/host_plugin_test.go`): a ginkgo/kind spec that loads a plugin onto a hostGroup via `hostPlugins`, deploys a workload importing the plugin's `acme:kv/store`, and asserts a value written via `/set` reads back via `/get` — proving the CLI/chart loading path and cross-store bind work in a real cluster.

**Verified end-to-end locally in kind** against a `registry:2` on the kind network: built the all-features host image, pushed the two fixtures with `wash oci push`, and ran the ginkgo spec — the host pulled the plugin from OCI, loaded it, an `acme:kv` workload scheduled and reached Ready, and `/set`→`/get` round-tripped a value through the plugin's store. `SUCCESS! -- 1 Passed`. The gated OCI-source `load_component_plugin` test and a direct `wash host --host-plugin …` run were also verified against the same registry. In CI this stays gated on the two fixtures being published (`HOST_PLUGIN_E2E_IMAGE`, `HOST_PLUGIN_CALLER_E2E_IMAGE`).

## Follow-up to activate the e2e

The e2e **self-skips until two fixtures are published** (same bootstrap the implements/messaging specs used), because host plugins load at host startup. A non-pullable image would fail host boot and break the all-features leg. To turn it on:

1. Publish `crates/wash-runtime/tests/fixtures/kv-plugin` and `kv-plugin-caller` as OCI images.
2. Fill the empty `host_plugin_image` / `host_plugin_caller_image` matrix values in `wash.yml` (all-features leg).

### DIFF
--- a/charts/runtime-operator/templates/runtime/deployment.yaml
+++ b/charts/runtime-operator/templates/runtime/deployment.yaml
@@ -133,6 +133,11 @@ spec:
             {{- range .wasmProposals }}
             - "--wasm-proposal={{ . }}"
             {{- end }}
+            {{- /* Host component plugins: one --host-plugin flag per entry, each
+                   rendered into the CLI's `key=value` grammar. */}}
+            {{- range .hostPlugins }}
+            - "--host-plugin=id={{ .id }}{{ if .image }},image={{ .image }}{{ if .pullPolicy }},pull={{ .pullPolicy }}{{ end }}{{ if .digest }},digest={{ .digest }}{{ end }}{{ else if .file }},file={{ .file }}{{ end }}{{ if .maxRestarts }},max-restarts={{ .maxRestarts }}{{ end }}"
+            {{- end }}
             {{- /* Chart-wide runtime.extraArgs applies to every host group; per-host-group
                    .extraArgs is appended after it. */}}
             {{- with $top.Values.runtime.extraArgs }}
--- a/charts/runtime-operator/values.yaml
+++ b/charts/runtime-operator/values.yaml
@@ -355,6 +355,28 @@ runtime:
       # component-model-async, gc, exception-handling, wide-arithmetic,
       # threads, tail-call.
       wasmProposals: []
+      # -- Host component plugins for this host group: WebAssembly components
+      # that provide a host capability, each served from its own supervised
+      # store to every workload that imports its interface. Rendered as
+      # `--host-plugin=...` args.
+      #
+      # Requires a wash image built with the `host-component-plugins` feature
+      # (not enabled in release images yet); a host handed a plugin it cannot
+      # build fails to start.
+      #
+      # Each entry needs an `id` plus exactly one of `image` (OCI reference) or
+      # `file` (path in the container). Optional: `pullPolicy`
+      # (always | ifNotPresent | never, image only), `maxRestarts`, `digest`
+      # (image only).
+      #
+      # Example:
+      #
+      #   hostPlugins:
+      #     - id: acme-kv
+      #       image: ghcr.io/acme/kv-host:1.0.0
+      #       pullPolicy: ifNotPresent
+      #       maxRestarts: 3
+      hostPlugins: []
       http:
         enabled: true
         port: 9191
--- a/crates/wash-runtime/src/plugin/component_plugin_spec.rs
+++ b/crates/wash-runtime/src/plugin/component_plugin_spec.rs
@@ -1,0 +1,195 @@
+//! Declarative spec for a host component plugin: a host-unique id plus where to
+//! fetch its wasm from.
+//!
+//! A spec is produced from a `wash host --host-plugin` flag string (via
+//! [`FromStr`]) or converted from a `wash dev` config entry, then resolved to a
+//! running plugin by [`super::component_host::load_component_plugin`]. That
+//! loader is gated on the `host-component-plugins` feature; this spec type is
+//! always compiled so the CLI can accept a plugin declaration and fail with a
+//! clear error on a build that lacks the feature, rather than silently dropping
+//! it.
+
+use std::path::PathBuf;
+use std::str::FromStr;
+
+use anyhow::{Context as _, anyhow, bail, ensure};
+
+use crate::oci::OciPullPolicy;
+
+/// Where a host component plugin's wasm bytes come from.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PluginSource {
+    /// Pulled from an OCI registry by image reference.
+    Oci {
+        image: String,
+        pull_policy: OciPullPolicy,
+    },
+    /// Read from a local file path.
+    File(PathBuf),
+}
+
+/// A host component plugin to load: a host-unique id, a source for its wasm, and
+/// optional supervision/integrity settings.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ComponentPluginSpec {
+    /// Host-unique plugin id. Collides loudly with an existing plugin's id at
+    /// registration time (`HostBuilder::with_plugin` dedupes).
+    pub id: String,
+    pub source: PluginSource,
+    /// Supervised driver restarts before the plugin is declared dead. `None`
+    /// uses the loader default.
+    pub max_restarts: Option<u32>,
+    /// Optional OCI digest to pin for supply-chain integrity. Only meaningful
+    /// for [`PluginSource::Oci`]; the loader rejects it on a file source.
+    pub expected_digest: Option<String>,
+}
+
+impl ComponentPluginSpec {
+    /// Build a spec from an id and source with default supervision/integrity
+    /// settings (no restart-cap override, no digest pin).
+    pub fn from_plugin_source(id: impl Into<String>, source: PluginSource) -> Self {
+        Self {
+            id: id.into(),
+            source,
+            max_restarts: None,
+            expected_digest: None,
+        }
+    }
+}
+
+/// Parse a `wash host --host-plugin` value: a comma-separated list of
+/// `key=value` fields. Required: `id`, and exactly one of `image` / `file`.
+/// Optional: `pull` (image only), `max-restarts`, `digest` (image only).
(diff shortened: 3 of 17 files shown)


charts/runtime-operator/templates/runtime/deployment.yaml
```diff
@@ -133,6 +133,11 @@ spec:
             {{- range .wasmProposals }}
             - "--wasm-proposal={{ . }}"
             {{- end }}
+            {{- /* Host component plugins: one --host-plugin flag per entry, each
+                   rendered into the CLI's `key=value` grammar. */}}
+            {{- range .hostPlugins }}
+            - "--host-plugin=id={{ .id }}{{ if .image }},image={{ .image }}{{ if .pullPolicy }},pull={{ .pullPolicy }}{{ end }}{{ if .digest }},digest={{ .digest }}{{ end }}{{ else if .file }},file={{ .file }}{{ end }}{{ if .maxRestarts }},max-restarts={{ .maxRestarts }}{{ end }}"
```
ricochet, line 139: FYI I opted to have this be plain image refs and not Artifacts. 

Host is a status-only CRD (heartbeat-derived, OSArch, Version, LastSeen…), and there's no HostGroup CRD. Today these are just helm-templated into the host Deployment.

So if we want this in the future we'd essentially create a HostGroup CRD (or a HostSpec) whose hostPlugins[] carry artifactFrom references instead of inline images. Then the host reconciler:

1. watches HostGroup + the referenced plugin Artifacts
2. for each plugin, waits for its Artifact to reach Published
3. renders the resolved Status.ArtifactURL into the host Deployment's --host-plugin args
4. owns the host Deployment, so there's a single owner and re-rendering on a new artifact rolls the pod cleanly.

AKA this is something we would add on top if we wanted it.

charts/runtime-operator/values.yaml
```diff
@@ -355,6 +355,22 @@ runtime:
       # component-model-async, gc, exception-handling, wide-arithmetic,
       # threads, tail-call.
       wasmProposals: []
+      # -- Host component plugins for this host group: WebAssembly components
```
vados-cosmonic, line 358: Let's space his out a bit?

crates/wash-runtime/src/plugin/component_plugin_spec.rs
```diff
@@ -0,0 +1,222 @@
...
+use std::path::PathBuf;
+use std::str::FromStr;
+
+use anyhow::{Context as _, anyhow, bail, ensure};
+
+use crate::oci::OciPullPolicy;
+
+/// Where a host component plugin's wasm bytes come from.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PluginSource {
```
vados-cosmonic, line 21: Hmnn weird that this isn't a bit more generic across all the wasms we load? 

Fine to do in a follow up, but surely there's some coalescing that could happen here.

crates/wash-runtime/src/plugin/component_plugin_spec.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// uses the loader default.
+    pub max_restarts: Option<u32>,
+    /// Optional OCI digest to pin for supply-chain integrity. Only meaningful
+    /// for [`PluginSource::Oci`]; the loader rejects it on a file source.
+    pub expected_digest: Option<String>,
+}
+
+/// Parse a pull policy name: `always`, `ifNotPresent`, or `never`
+/// (case- and separator-insensitive).
+pub fn parse_pull_policy(s: &str) -> anyhow::Result<OciPullPolicy> {
```
vados-cosmonic: Could this be a `FromStr` impl?

crates/wash-runtime/src/plugin/component_plugin_spec.rs
```diff
@@ -0,0 +1,222 @@
...
+        "always" => Ok(OciPullPolicy::Always),
+        "ifnotpresent" => Ok(OciPullPolicy::IfNotPresent),
+        "never" => Ok(OciPullPolicy::Never),
+        other => bail!("invalid pull policy {other:?}; expected always|ifNotPresent|never"),
+    }
+}
+
+impl ComponentPluginSpec {
+    /// Build an OCI-sourced spec.
+    pub fn oci(
```
vados-cosmonic:
```suggestion
    pub fn from_oci(
```

Should these be combined into `from_plugin_source(` ?

crates/wash/src/cli/host.rs
```diff
@@ -126,6 +126,77 @@ pub struct HostCommand {
         value_delimiter = ','
     )]
     pub wasm_proposals: Vec<WasmProposal>,
+
+    /// Load a host component plugin: a WebAssembly component that provides a
```
vados-cosmonic: We need to break up this description so the short description shows up?

crates/wash/src/cli/host.rs
```diff
@@ -126,6 +126,77 @@ pub struct HostCommand {
...
+    )]
+    pub host_plugin_registry_user: Option<String>,
+
+    /// Password paired with `--host-plugin-registry-user` /
+    /// `WASH_HOST_PLUGIN_REGISTRY_USER`.
+    #[cfg(feature = "host-component-plugins")]
+    #[arg(
+        long = "host-plugin-registry-password",
+        env = "WASH_HOST_PLUGIN_REGISTRY_PASSWORD",
+        hide_env_values = true
```
vados-cosmonic: I think we can require that registry-user is provided here w/ `requires` and vice versa

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but nits!

crates/wash-runtime/src/plugin/component_plugin_spec.rs
ricochet: This is a great suggestion but to do it right, I'd want to refactor workload components, wash oci, wash dev where right now they all resolve image/file sources their own way. I'll do this one as a quick follow

crates/wash-runtime/src/plugin/component_plugin_spec.rs
ricochet: Replaced oci()+file() with a single from_plugin_source(id, source)

crates/wash/src/cli/host.rs
ricochet: Added clap requires both ways (mutually required); dropped the runtime warn-on-partial from the helper

jfleitz (approved):

vados-cosmonic (approved):

Issue #5338, opened 2026-07-14
Sockets: allow `allow-ip-name-lookup` in wasmtime


jamesstocktonj1:
# Description
When trying to call `resolve-addresses` in the `wasi:sockets/ip-name-lookup@0.2.6` we get the following error:

```sh
$ wash dev
...
$ curl localhost:8000
Resolve Addresses Error: permanent-resolve-failure
```

You can run the same wasm binary within wasmtime with the following feature enabled:
```sh
$ wasmtime serve -S allow-ip-name-lookup, ...
...
$ curl localhost:8000
Success
```

Can we enable this feature within wasmCloud, possibly behind a similar feature flag?

ricochet:
Fixed by #5404

PR #5362, opened 2026-07-20
feat: allow IP name lookup in `wasi:sockets` interface

jamesstocktonj1:
# Description
- Allow `resolve-addresses` call in the `wasi:sockets/ip-name-lookup` interface
  - Enabled by `--allow-ip-name-lookup` in `wash host`
  - Enabled by `dev.allow_ip_name_lookup` bool in `.wash/config.yaml`
- https://github.com/wasmCloud/wasmCloud/issues/5338

## Testing

Tested using a simple function calling `resolve-addresses` with both wasi p2 (0.2.6) and p3 (0.3.0) interfaces:

### Before
```sh
$ wash dev
...
$ curl localhost:8000
Resolve Addresses Error: permanent-resolve-failure
```

### After 
```sh
$ wash dev
...
$ curl localhost:8000
Success
```


Aditya1404Sal (commented):
Hey @jamesstocktonj1, this looks great :D -- I like that it's host-controlled at the engine level, which matches the wasmtime convention of configuring the engine to permit `wasi:sockets/ip-name-lookup` rather than the component asking for it.

I do want to point out that you've already got the engine-level piece -- ie. `Engine::allow_ip_name_lookup` + `with_allow_ip_name_lookup`. The only bit I'd change is the last hop: rather than copying it onto LocalResources at init and reading it back in `build_ctx_from_template`, you could thread the engine's value straight through `ComponentCtxTemplate` the way `tls_provider` already does. That would let you delete the `LocalResources.allow_ip_name_lookup field`, the init-time overwrites, and the washlet default, so nobody's gonna (well, atleast be tempted to) configure it per-component later.

ah also as a bonus: LocalResources isn't #[non exhaustive], so not adding a field to it avoids source-breaking every literal constructor which is about 13 test files this PR currently touches. (It's not a semver break (since wash-runtime is publish = false)), but it's a fair bit of churn on a commonly constructed type which the engine route can sidestep entirely :)

Again, awesome work here so far!

jamesstocktonj1:
Thanks @Aditya1404Sal for the feedback! Added the field to `ComponentCtxTemplate` and looks a bit neater now.

ricochet:
Right now this grants DNS to every workload on the host. It looks like you may have started with localResources which I think is the better direction. 

- We already have allowedHosts and it is an established egress policy lives in localResources and travels with the workload. DNS resolution is the same category of grant.
- A host-wide flag has grant-creep semantics. On a multi-workload wash host, one workload needing DNS means every workload gets DNS. DNS is a classic exfiltration channel precisely because it works even when direct egress is blocked.
- Scheduling portability. WorkloadConfig is designed to round-trip to a Kubernetes WorkloadDeployment. If the capability lives on the host, a workload's behavior depends on which host it lands on and there's no declarative way to express "this workload needs to resolve names." Per-workload, the need is visible to admission control, statically auditable, and satisfiable on any host.
- Wasmtime's CLI does have a host-level -S allow-ip-name-lookup, but that's a single-tenant runner where host and workload are the same trust domain. Platforms that schedule other people's workloads (Kubernetes NetworkPolicy being the obvious analog) put network policy on the workload.

ricochet:
Worked on an alternative approach with finer-grain control: https://github.com/wasmCloud/wasmCloud/pull/5404

jamesstocktonj1:
Thanks, this looks good to me!

Closing in favour of https://github.com/wasmCloud/wasmCloud/pull/5404

PR #5365, opened 2026-07-20
Fix dropping span links and trace state in wasi-otel

ricochet:
Fix dropping span links and trace state in wasi-otel

on-end counted span_data.links into link_count but never attached them to the exported span, so guest-recorded links (producer/consumer correlation, fan-in batches) were lost. Build them into the SpanBuilder via with_links.

otel_span_context_to_wit hardcoded trace_state to an empty list, dropping W3C tracestate when the host hands its span context back to the guest over outer-span-context. Render the OTel TraceState and split it back into the WIT list-of-tuples form.


vados-cosmonic (approved):
LGTM :rocket:

PR #5367, opened 2026-07-20
fix(blobstore): report created-at for fs blobstore

ricochet:
The filesystem blobstore backend hardcoded created-at to 0 in both container info and object-info, so guests saw an epoch-0 timestamp even though the fs metadata (already read for object size) carries the real creation time. Derive created-at from the entry's fs metadata, falling back to last-modified where a creation time isn't recorded. Matches the in-memory backend, which already reports seconds since the Unix epoch.


vados-cosmonic (approved):
LGTM 🚀

Issue #5351, opened 2026-07-16
Service file interfaces not included in host interface resolution during wash dev


george-lim:
### Summary

When using `wash dev` with a `service_file`, host plugins (e.g. `wasi:keyvalue`) are never bound to the service component. The service fails at runtime with "service did not properly execute" because its imports can't be satisfied.

### Root cause

In `crates/wash/src/cli/dev.rs`, `create_workload` calls `host.intersect_interfaces()` on the main dev component and each sidecar component to extract their WIT interfaces, but never on the `service_file`. When `build_workload` assembles `all_component_interfaces` for `build_workload_host_interfaces`, the service's imports are missing. Downstream, `bind_plugins` in `workload.rs` never sees that the service needs capabilities like `wasi:keyvalue/store`, so it never binds the corresponding plugin.

The plugin binding logic already supports service components — the issue is that the service file's interfaces are never extracted and included in the host interface list.

### Fix

1. Call `host.intersect_interfaces()` on the service file bytes in `create_workload`
2. Pass the extracted interfaces to `build_workload`
3. Include them in `all_component_interfaces` before calling `build_workload_host_interfaces`

### Reproduction

1. Create a project with a main HTTP handler component and a `service_file` component that imports `wasi:keyvalue/store`
2. Configure `wasi_keyvalue_path` in `.wash/config.toml`
3. Run `wash dev`
4. Observe WARN service did not properly execute — the service traps because keyvalue is never linked

Removing the `wasi:keyvalue` import from the service makes it start successfully, confirming the plugin binding is the issue.

### Branch

I have a branch that I've verified working. If you're good with it I can make a PR: https://github.com/george-lim/wasmCloud/tree/expose-interfaces-to-service

ricochet:
Thank you for reporting! Your fix looks correct to me.

ricochet:
Fixed in #5358

PR #5368, opened 2026-07-20
fix(wash): skip unused service file

ricochet:
When `dev.service` is true the dev component itself becomes the workload's
service and any `dev.service_file` is ignored by `build_workload`. Gate the
service-file read and interface extraction on `!dev.service` so the file is
neither read nor folded into the workload host interfaces when it won't be
deployed.

Also add a regression test (for #5351) asserting a service file's imports
reach `host_interfaces`, so host plugins bind to the service component.

### DIFF
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -446,8 +446,12 @@ async fn create_workload(
         });
     }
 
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5370, opened 2026-07-21
fix(ci): update protoc setup

ricochet:
arduino/setup-protoc was abandoned and last release on Node 20. GH switched over to node 24 this morning and that's why main CI is down.

This adds our own protoc setup action and a way to keep this updated. We were stale on 29, so this also updates the version to the latest.


.github/actions/setup-protoc/action.yml
```diff
@@ -0,0 +1,64 @@
...
+  version resolution or token and runs on any Node version the runner defaults
+  to. Requires the repository to be checked out first.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install protoc (Linux/macOS)
+      if: runner.os != 'Windows'
+      shell: bash
+      run: |
```
jfleitz: To get passed the zizmor flag, you can add `# zizmor: ignore[github-env]`:
`run: | # zizmor: ignore[github-env]`.

This block looks safe, since `RUNNER_TEMP` is used for the filename, and then used to GITHUB_PATH, so no injection is possible.

jfleitz (dismissed):
LGTM. Left comment for not having zizmor mis-flag a security issue.

.github/actions/setup-protoc/action.yml
ricochet: Fixed!

PR #5373, opened 2026-07-21
chore(wash-runtime): upgrade to wasmtime 47

ricochet:
Move the workspace off the git-pinned wasmtime 46 fork onto the stock crates.io wasmtime 47.0.1 release. The resource-implements feature (bytecodealliance/wasmtime#13666) that the ricochet fork backported now ships in 47, so both [patch] blocks and the fork branch are retired. wasi-webgpu-wasmtime moves to 0.2 (its wasmtime-47 release).

Adapt to 47 API changes:
- wasi:sockets p3 HostTcpSocketWithStore::listen is now `async fn` (and the impl now requires `T: Send`).
- Accessor::spawn now returns `Result<JoinHandle>`; log spawn failures in the trigger service ingress loops instead of dropping them.
- StoreContextMut::async_call_stack now returns `Result`; treat an error as "no caller task" in the component_host identity/cancel host funcs.

Port the socket_addr_check-on-implicit-bind fix (wasmtime#13677) to the loopback-aware fork: p3 `listen` and the first UDP `send-to` implicitly bind an unbound socket, so run the host's socket_addr_check against the implicit bind address (TcpBind / UdpBind) first. Without it a guest could implicitly bind to an address (e.g. all-interfaces 0.0.0.0) the network policy would otherwise deny.

wasmtime 47 enables the GC (and its wasm_function_references prerequisite), exception-handling, and wide-arithmetic proposals by default, so the WasmProposal::{Gc,ExceptionHandling,WideArithmetic} arms were re-asserting defaults. Collapse them into a documented no-op and update the variant docs.

The variants and FromStr stay unchanged so `--wasm-proposal gc` etc. still parse (accepted for compatibility). Threads and TailCall keep setting their flags since their defaults are conditional (the `threads` crate feature and non-Winch respectively); ComponentModelAsync and implements are default-off and still required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -665,14 +665,14 @@ dependencies = [
 
 [[package]]
 name = "cap-fs-ext"
-version = "3.4.5"
+version = "4.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d5528f85b1e134ae811704e41ef80930f56e795923f866813255bc342cc20654"
+checksum = "d78e5a3368ae89b7cb68186411452b4b9fac8b41be9c19bf3f47c2d2c8e36e6b"
 dependencies = [
- "cap-primitives",
- "cap-std",
- "io-lifetimes",
- "windows-sys 0.59.0",
+ "cap-primitives 4.0.2",
+ "cap-std 4.0.2",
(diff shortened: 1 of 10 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5374, opened 2026-07-21
fix(wash-runtime): round-robin across service replicas per host

ricochet:
N service replicas on one host served like one, from two defects:

1. Service-only workloads never registered their hostname. A p3 trigger
   service reaches routing via on_service_http_resolved, not
   on_workload_resolved, and DynamicRouter left the Router hook a default
   no-op — so a service behind a hostname router 404'd. Hostnames now flow
   through HostHandler/Router::on_service_http_resolved (derived once via
   http_ingress_hostnames) and DynamicRouter registers them like component
   workloads.

2. route_incoming_request picked HashSet::iter().next(), pinning every
   request for a hostname to one arbitrary replica. It now strides a relaxed
   atomic cursor over an ordered (BTreeSet) replica set for round-robin, with
   selection split into select_workload() and an empty-set guard.

Registering service hostnames means they must be cleaned on stop: the
service-only stop path calls on_service_http_unbind but not
on_workload_unbind, so that hook now also drops the router registration
(idempotent) — otherwise a stopped replica lingers in rotation and 404s.

Adds unit tests for round-robin distribution, service hostname
registration, unbind cleanup, and http_ingress_hostnames extraction, plus
an integration test asserting 8 requests across 4 replicas land 2 each.

Caught by @LiamRandall. I took his fix and added tests. I ran our benches and this has no material impact on perf.

### DIFF
--- a/crates/wash-runtime/src/host/http.rs
+++ b/crates/wash-runtime/src/host/http.rs
@@ -22,9 +22,11 @@ use std::{
     collections::{BTreeSet, HashMap},
     net::SocketAddr,
     path::Path,
-    sync::{Arc, atomic::AtomicUsize},
+    sync::Arc,
     time::Duration,
 };
+
+use arc_swap::ArcSwap;
 
 use crate::host::allowed_hosts::AllowedHost;
 use crate::host::trigger_service::{BrokerMessage, MessagingJob};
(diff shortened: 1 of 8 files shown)


jfleitz (dismissed):

crates/wash-runtime/src/host/http.rs
```diff
@@ -178,10 +219,64 @@ pub trait Router: Send + Sync + 'static {
...
+            .map_err(|_| RouteError::Unavailable)?;
+        let Some(workload_set) = lock.get(host) else {
+            return Err(RouteError::NoWorkloadForHost(host.to_string()));
+        };
+        // An entry can exist but be empty; treat that as "no workload bound"
+        // (same 404) and, importantly, avoid the `% 0` below.
+        if workload_set.is_empty() {
+            return Err(RouteError::NoWorkloadForHost(host.to_string()));
+        }
+        let idx = self.rr.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```
vados-cosmonic: I think we might be better off just doing PRNG here rather than trying to use an atomic sharding mechanism. It works, but shouldn't be incredibly different from a decent PRNG and the PRNG won't require cross-core coordination.

This would change the nature of the tests that you had to make them more probabilistic but should be much faster when it comes to requests coming in at scale.

vados-cosmonic (dismissed):
LGTM 🚀

LiamRandall:
For services, I've documented a few warnings to keep replica instances set to 1 until we have time to think through a per-workload routing policy (sticky, round robin, etc)

https://github.com/wasmCloud/wasmcloud.com/commit/6b3893ecf00c1107f0ea012541efcd50138c74ff

vados-cosmonic (approved):
LGTM 🚀

crates/wash-runtime/src/host/http.rs
ricochet: This is a great suggestion, I iterated on this a bit and got to a fully lock-free path with spread.

PR #5377, opened 2026-07-21
fix(wash-runtime): honor WASH_PLUGIN_STOP_TIMEOUT_SECS in Host::stop

ericgregory:
Host::stop's hardcoded 3-second per-plugin timeout fired before ComponentHostPlugin::stop's env-tunable supervisor wait (default 5s), dropping the stop future and detaching a wedged supervisor task instead of aborting it — derive the cap from the same knob plus a one-second grace so the abort path always runs.

### DIFF
--- a/crates/wash-runtime/src/host/mod.rs
+++ b/crates/wash-runtime/src/host/mod.rs
@@ -345,9 +345,10 @@ impl Host {
 
     /// Stop the host and shut down all plugins.
     ///
-    /// Attempts to gracefully stop all plugins with a 3-second timeout
-    /// for each. Errors are logged but don't prevent other plugins from
-    /// being stopped.
+    /// Attempts to gracefully stop all plugins, allowing each the
+    /// `WASH_PLUGIN_STOP_TIMEOUT_SECS` budget plus a one-second grace.
+    /// Errors are logged but don't prevent other plugins from being
+    /// stopped.
     ///
     /// # Returns
     /// Ok if the shutdown process completes (even with plugin errors).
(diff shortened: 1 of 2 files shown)


crates/wash-runtime/src/host/mod.rs
```diff
@@ -1014,6 +1029,54 @@ mod tests {
     use super::*;
     use crate::types::Component;
 
+    /// `Host::stop`'s per-plugin timeout must outlast the plugin-stop budget.
+    /// A host component plugin's `stop()` waits the full budget for its
+    /// supervisor and only then runs its abort-and-cleanup tail; if the outer
+    /// timeout fired first it would drop that future mid-wait, detaching the
+    /// supervisor's JoinHandle instead of aborting it. The mock below mirrors
+    /// that shape: sleep the budget, then record that the tail ran.
+    #[tokio::test(start_paused = true)]
```
vados-cosmonic, line 1038: super nit: Is starting w/ the runtime's clock paused strictly necessary?

vados-cosmonic (approved):
LGTM 🚀

crates/wash-runtime/src/host/mod.rs
ericgregory: Not strictly necessary, but I think worthwhile - it's there so the test doesn't spend a real 5 second (by default) plugin_stop() sleeping. With the paused clock tokio auto-advances the timers and the test finishes in ~40ms.

vados-cosmonic: Thanks, TIL about this feature!

crates/wash-runtime/src/timeouts.rs
```diff
@@ -62,9 +62,10 @@ declare_timeouts! {
     shared_store_call = ("WASH_SHARED_STORE_CALL_TIMEOUT_SECS", 30);
     /// Max wall-clock for a trigger service to produce an HTTP response.
     http_response = ("WASH_HTTP_RESPONSE_TIMEOUT_SECS", 600);
-    /// How long `stop()` waits for a host component plugin's supervisor to exit
-    /// before aborting it.
-    #[cfg(feature = "host-component-plugins")]
```
ricochet, line 67: Not sure we want to drop the flag here but also totally fine for this field

ricochet (approved):

PR #5379, opened 2026-07-22
fix(wash-runtime): reject duplicate workload IDs

vados-cosmonic:
One thing I've realized is that we currently allow workloads to come in with the *same* exact workload_id, and when they do we simply overwrite the existing one. 

We probably want to require that workload IDs are distinct and do any overwriting (with gracious shutdown of the previous workload) explicitly. In favor of that/avoiding this subtle behavior, this PR rejects duplicate workload IDs by default.

### DIFF
--- a/crates/wash-runtime/src/host/mod.rs
+++ b/crates/wash-runtime/src/host/mod.rs
@@ -664,9 +664,12 @@ impl HostApi for Host {
             if workloads.contains_key(&request.workload_id) {
                 return Ok(WorkloadStartResponse {
                     workload_status: WorkloadStatus {
-                        workload_id: request.workload_id,
+                        workload_id: request.workload_id.clone(),
                         workload_state: WorkloadState::Error,
-                        message: "Workload ID already exists".to_string(),
+                        message: format!(
+                            "Workload ID [{}] already exists (the exising workload must be stopped to reuse the ID)",
+                            request.workload_id
+                        ),
                     },
                 });
             }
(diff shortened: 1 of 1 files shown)


crates/wash-runtime/src/host/mod.rs
```diff
@@ -656,11 +656,22 @@ impl HostApi for Host {
         &self,
         request: WorkloadStartRequest,
     ) -> anyhow::Result<WorkloadStartResponse> {
-        // Store the workload with initial state
-        self.workloads
-            .write()
-            .await
-            .insert(request.workload_id.clone(), HostWorkload::Starting);
+        // Reserve the workload ID while holding the write lock so concurrent
```
ricochet, line 659: When a workload fails to start, we're now stuck in an Error-state that blocks ID reuse. I think we want to have an error in this case so that folks at least know they need to stop and start again: e.g. "workload ID {id} already exists (state: error); stop it first to reuse the ID".

An alternative would be to treat HostWorkload::Error as reclaimable and replace it in-place (nothing live to leak), reserving rejection for Starting/Running/Stopping. So it'd act like a retry for reconcilers

ricochet (dismissed):
Nice tight fix. Glad you spied this 👍

crates/wash-runtime/src/host/mod.rs
vados-cosmonic: Yeah that's a good point -- I'll improve the error message there

ricochet (dismissed):

ricochet (approved):

PR #5381, opened 2026-07-22
[FEAT] plugin workload lifecycle

ricochet:
A host component plugin may now export `wasmcloud:host/workload-lifecycle@0.1.0` to react to the workloads that call it. The host delivers `on-workload-bind`, carrying the workload's identity, its service and component ids, and each matched interface instance with its typed version and manifest config, before any capability call from that workload is served, and a best-effort `on-workload-unbind` when the workload goes away.

This lets a plugin provision and reclaim per-workload state **eagerly** rather than lazily on first call: validate interface config at deploy time so a misconfigured workload fails to deploy instead of failing on its first request, allocate per-tenant resources up front, and free them deterministically. Later capability calls correlate back to that state through the existing `wasmcloud:host/identity` import.

A plugin that does not export the interface is entirely unaffected. It never hears about workload lifecycle and its capability calls behave exactly as before.

## The interface

```wit
interface workload-lifecycle {
    record workload-info {
        id: workload-id,
        name: string,
        namespace: string,
        service: option<component-id>,
        components: list<component-id>,
        interfaces: list<interface-binding>,
    }

    on-workload-bind: async func(workload: workload-info) -> result<_, string>;
    on-workload-unbind: async func(id: workload-id);
}
```

`interface-binding` carries the namespace, package, matched interface names, an optional typed `version` (major/minor/patch/pre/build), an optional instance label, and the interface-level config from the workload manifest.

`service` and `components` are reported separately. A service is the workload's stateful item and binds to a plugin through the same path a component does, but conflating the two would leave a plugin unable to tell which id is which, and the id in `service` is the same one `identity.get-component-id` reports for capability calls the service itself makes.

## Behavior

**Reserved namespace.** `wasmcloud:host` exports are reserved: only the lifecycle interface is accepted, it never enters the plugin's world, and no workload import can ever match or call it. Both hook signatures are shape-checked at registration (bind takes the `workload-info` record and returns `result<_, string>`; unbind takes the workload id), so a mismatched plugin fails at construction with a clear message rather than on first deploy.

**Bind rejection fails the deploy.** A bind the guest rejects with a `result` error fails the workload deploy with the guest's message surfaced. The plugin rolls back its bound-workload record and fires a best-effort unbind, since the engine only unbinds *previously*-bound plugins on a bind failure.

**Replay after a supervised restart.** A restart rebuilds the plugin store and loses all guest state while bound workloads keep running, so each fresh incarnation replays `on-workload-bind` for every workload still bound, and the TriggerService completes the replay before serving queued capability calls.

**Quarantine and eviction.** Replay is serial with a per-workload marker recorded before each bind: a guest trap short-circuits `run_concurrent` and drops the serving future before its handler runs, so a replay fault can only be attributed by the marker, which survives the store in the incarnation's registry. A bind that crash-loops the shared store is struck on each attributed fault and, past a strike ceiling (below the plugin restart budget, reset after a healthy uptime), evicted aka dropped from the replay set so the store recovers for every other tenant and reported to the host so the evicted workload's scheduling health becomes failed. Attributed replay faults do not consume the plugin-wide restart budget.

**Bounded lifecycle calls.** Lifecycle calls are bounded end-to-end (channel send plus reply) by `WASH_PLUGIN_LIFECYCLE_CALL_TIMEOUT_SECS` (default 30s, overridable per plugin) so a full channel or a hung hook cannot wedge a deploy or a stop. A bind that overruns the budget fails the deploy *at the budget* rather than waiting the hook out. The hook is uncancellable, so because it keeps running, the rollback unbind is deferred until it returns and scoped to the same incarnation, guaranteeing bind-then-unbind ordering so state the hook provisions late is still reclaimed.

## Engine-side fixes

Two pre-existing cleanup gaps this work surfaced, fixed for all plugins rather than just component plugins:

- `unbind_all_plugins` now unbinds plugins bound to the workload's **service** item; previously only components were visited, so a service's plugin bindings leaked on workload stop.
- The deploy-failure path for unsatisfiable interfaces now unbinds already-bound plugins, matching what the other bind-failure paths already did.

## Fixtures and tests

The `kv-plugin` fixture exports the interface end-to-end with eager provision at bind, identity-keyed lookup at call, reclaim at unbind and verifies the ambient identity contract inside hooks, and exposes captured binds through new `bound-config` / `bind-info` / `lifecycle-log` store functions. A dedicated `badlifecycle` fixture exercises the registration-time signature check, and a new `kv-plugin-service` fixture deploys as a workload **service** (exports `wasi:cli/run`, imports the plugin capability) to cover the service bind path, which previously had no coverage at all.

Covered by the `integration_component_plugin_lifecycle` suite with typed delivery, replay ordering, quarantine and eviction with failed health, deferred rollback unbind, and service-vs-component reporting plus the existing component-host plugin tests. The service test asserts that the id delivered as `workload-info.service` is the same id `identity.get-component-id` reports for the service's own capability calls, which is the property that makes the field useful to a plugin.

## Notes for reviewers

The feature is behind the `host-component-plugins` cargo feature, which is not on by default.

`workload-info` is a WIT `record`, so adding fields to it later is a breaking change for compiled plugins. 

Two timing-sensitive tests set an explicit `digest` on their workload requests. `digest` is used for nothing but the engine's compiled-component cache, and without it every `workload_start` re-runs cranelift over the fixture so putting an unbounded, machine-dependent cost inside a window meant to measure only the plugin's lifecycle behavior. This was a source of CI flakiness. Separately, and not addressed here: digest-less deploys recompiling on every start is a real inefficiency on the bytes-supplied path, since content-hashing would be far cheaper than recompiling.

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -1629,6 +1629,37 @@ impl ResolvedWorkload {
             }
         }
 
+        // The service item records plugin bindings just like a component;
+        // unbind them the same way so a plugin bound only to the service is
+        // not left tracking a stopped workload.
+        if let Some(service) = self.service.as_ref()
+            && let Some(plugins) = service.plugins()
+        {
+            for (plugin_id, plugin) in plugins.iter() {
+                let world = service.world();
+                let plugin_world = plugin.world();
+                let bound_interfaces = world
+                    .imports
+                    .iter()
+                    .filter(|import| plugin_world.imports.contains(import))
+                    .cloned()
+                    .collect::<std::collections::HashSet<_>>();
+
+                if let Err(e) = plugin
+                    .on_workload_unbind(self.id(), WitInterfaces::new(&bound_interfaces))
+                    .await
+                {
+                    warn!(
+                        plugin_id,
+                        service_id = service.id(),
+                        workload_id = self.id.as_ref(),
+                        error = ?e,
+                        "failed to unbind plugin from service, continuing cleanup"
+                    );
+                }
+            }
+        }
+
         // A trigger service registered its HTTP/messaging handlers at start
         // (`execute_trigger_service`); drop those registrations on stop so it no
         // longer receives host-invoked deliveries on a torn-down instance.
@@ -2063,6 +2094,27 @@ impl UnresolvedWorkload {
                     interfaces = ?unmatched,
                     "no plugins found for requested interfaces"
                 );
+                // The same rollback the bind-failure paths perform: without it
+                // every successfully bound plugin keeps tracking a workload
+                // that never deploys.
+                for (bound_plugin, bound_interfaces, _) in
+                    bound_plugins_with_interfaces.iter().rev()
+                {
+                    debug!(
+                        plugin_id = bound_plugin.id(),
+                        "calling on_workload_unbind for cleanup after unmatched interfaces"
+                    );
+                    if let Err(cleanup_err) = bound_plugin
+                        .on_workload_unbind(self.id(), WitInterfaces::new(bound_interfaces))
+                        .await
+                    {
+                        warn!(
+                            plugin_id = bound_plugin.id(),
+                            error = ?cleanup_err,
+                            "failed to cleanup plugin after unmatched interfaces"
+                        );
+                    }
+                }
                 bail!(
                     "workload component {component_id} requested interfaces that are not available on this host: {unmatched:?}",
                 )
@@ -2217,6 +2269,19 @@ impl UnresolvedWorkload {
     /// Gets the namespace of the workload
     pub fn namespace(&self) -> &str {
         &self.namespace
+    }
+
+    /// Id of this workload's service, if it has one.
+    pub fn service_id(&self) -> Option<Arc<str>> {
+        self.service.as_ref().map(|s| Arc::from(s.id()))
+    }
+
+    /// Ids of this workload's components, sorted (deterministic despite the
+    /// backing map). Excludes the service — see [`Self::service_id`].
+    pub fn component_ids(&self) -> Vec<Arc<str>> {
+        let mut components: Vec<Arc<str>> = self.components.keys().cloned().collect();
+        components.sort();
+        components
     }
 
     /// Retrieves the interface configuration for a given WIT interface, if it exists.
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -1,7 +1,7 @@
 //! This module is primarily concerned with converting an [`UnresolvedWorkload`] into a [`ResolvedWorkload`] by
 //! resolving all components and their dependencies.
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     ops::{Deref, DerefMut},
     path::PathBuf,
     sync::Arc,
@@ -526,7 +526,7 @@ pub struct ResolvedWorkload {
     namespace: Arc<str>,
     /// All components in the workload. This is behind a `RwLock` to support mutable
     /// access to the component linkers.
-    components: Arc<RwLock<HashMap<Arc<str>, WorkloadComponent>>>,
+    components: Arc<RwLock<BTreeMap<Arc<str>, WorkloadComponent>>>,
     /// The HTTP handler for outgoing HTTP requests
     http_handler: Arc<dyn crate::host::http::HostHandler>,
     /// An optional service component that runs once to completion or for the duration of the workload
@@ -887,7 +887,7 @@ impl ResolvedWorkload {
         }
     }
 
-    pub fn components(&self) -> Arc<RwLock<HashMap<Arc<str>, WorkloadComponent>>> {
+    pub fn components(&self) -> Arc<RwLock<BTreeMap<Arc<str>, WorkloadComponent>>> {
         self.components.clone()
     }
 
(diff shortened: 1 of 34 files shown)


wit/host/wit/lifecycle.wit
```diff
@@ -0,0 +1,117 @@
...
+        /// platform's configuration calls the same instance.
+        ///
+        /// Two names, one contract: `name` above and `external-id` here are the
+        /// two names an instance can carry, and both refer to the one interface
+        /// contract described by the rest of this record. An external-id is not
+        /// an identity — it carries no uniqueness requirement, and a binding may
+        /// declare either name, both, or neither.
+        external-id: option<string>,
+        /// Interface-level configuration from the workload manifest.
+        config: list<config-entry>,
```
vados-cosmonic: map? Not super necessary and can definitely be a follow up if you want

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -2191,6 +2243,19 @@ impl UnresolvedWorkload {
         &self.namespace
     }
 
+    /// Id of this workload's service, if it has one.
+    pub fn service_id(&self) -> Option<Arc<str>> {
+        self.service.as_ref().map(|s| Arc::from(s.id()))
+    }
+
+    /// Ids of this workload's components, sorted (deterministic despite the
+    /// backing map). Excludes the service — see [`Self::service_id`].
```
vados-cosmonic: NIT: maybe the backing map should be a BTreeMap

crates/wash-runtime/src/plugin/component_host/mod.rs
```diff
@@ -300,11 +523,122 @@ impl HostPlugin for ComponentHostPlugin {
...
+                    self.id
+                ));
+            }
+        };
+
+        // The bind returned (rejection, malformed reply, or a delivery error
+        // after the hook ran): the hook is no longer in flight, so roll back
+        // immediately with a best-effort unbind.
+        if let Err(e) = self.remove_and_unbind(lifecycle, &workload_id).await {
+            debug!(id = self.id, %workload_id, err = %e, "post-failure unbind not delivered");
```
vados-cosmonic:
```suggestion
            warn!(id = self.id, %workload_id, err = %e, "post-failure unbind not delivered");
```

crates/wash-runtime/src/plugin/component_host/mod.rs
```diff
@@ -315,18 +649,73 @@ impl HostPlugin for ComponentHostPlugin {
...
-            add_capabilities_to_linker(linker, &self.state, exported)?;
+            if let Err(e) = add_capabilities_to_linker(linker, &self.state, exported) {
+                // The engine's bind-failure cleanup only unbinds plugins whose
+                // item binds ALL succeeded — a plugin failing its own item bind
+                // is not yet on that list — so roll back the bind delivered in
+                // `on_workload_bind` ourselves.
+                if let Some(lifecycle) = &self.lifecycle
+                    && let Err(unbind_err) = self.remove_and_unbind(lifecycle, &workload_id).await
+                {
+                    debug!(
```
vados-cosmonic:
```suggestion
                    warn!(
```

crates/wash-runtime/src/plugin/component_host/mod.rs
```diff
@@ -396,7 +805,400 @@ fn build_plugin_linker(
...
+}
+
+/// Pull the two lifecycle hooks out of the plugin's introspected
+/// `wasmcloud:host/workload-lifecycle` export. Presence, arity, and the
+/// top-level shape of each signature are checked here so a mismatched plugin
+/// fails at registration with a clear message rather than on its first
+/// workload deploy. Deep per-field types (every field of `workload-info`)
+/// still surface as a per-delivery typecheck error — the fixed shape below
+/// catches the mistakes worth catching cheaply.
+fn lifecycle_funcs(id: &str, mut export: ExportedInterface) -> anyhow::Result<LifecycleFuncs> {
```
vados-cosmonic: We should move this function and all the lifecycle related stuff to a module named `lifecycle` at some point.

crates/wash-runtime/tests/fixtures/badlifecycle/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+struct Component;
+
+impl OpsGuest for Component {
+    fn ping() -> u32 {
+        1
+    }
+}
+
+impl LifecycleGuest for Component {
+    fn on_workload_bind(_id: String) -> u32 {
```
vados-cosmonic, line 28: This is supposed to return a `Result<(), String>`? I think 0 will lift into the right thing, but curious how this is working

vados-cosmonic (dismissed):
LGTM 🚀

wit/host/wit/lifecycle.wit
ricochet: yeah easier to convert to a map now

crates/wash-runtime/src/engine/workload.rs
ricochet: Yep definitely. Annoyed at myself for missing this one.

crates/wash-runtime/src/plugin/component_host/mod.rs
ricochet: better to do it now

crates/wash-runtime/tests/fixtures/badlifecycle/src/lib.rs
ricochet: I can see how this is weird. This is a test of passing a malformed copy of the interface wit. This is what wit-bindgen produces with (incorrect) wit. I wanted a test for the parts of the signature mismatching. Updated the doc comment to hopefully make this clearer.

ricochet: But for how is it working? Well the point is that at registration it doesn't :)

crates/wash-runtime/src/engine/workload.rs
vados-cosmonic: I think we have this all over the place -- IMO it mostly didn't matter (and map was the right choice, I'm not sure we've actually been doing any sorting until now)

crates/wash-runtime/tests/fixtures/badlifecycle/src/lib.rs
vados-cosmonic: Ah, good point, I didn't quite grasp the test here

PR #5382, opened 2026-07-22
fix: use consumer groups by default for NATS

vados-cosmonic:

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,12 +216,33 @@ cargo build
 cargo build --workspace
 ```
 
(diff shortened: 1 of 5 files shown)


jfleitz (approved):
This is great and I think much needed, as this solves the scaling issue for nats processing, where you want the message to be processed by one (any) available listener vs broadcasting to all listeners.

This feels like a "default change" to how NATS messaging works, so it should probably be called out? If this gets in for the next release, we are already bumping the minor revision, so that fits well.

vados-cosmonic:
Yeah this is definitely quite the change -- we don't know that anyone was depending on this functionality but it certainly is a big difference from how it worked before.This probably deserves a docs change.

ricochet (approved):

PR #5383, opened 2026-07-22
feat(wash-runtime): add service HTTP benchmarks

ricochet:
Criterion bench suite (service_http) for the service HTTP plane. A long-lived p3 instance co-driving wasi:cli/run with wasi:http/handler, served through the ingress channel instead of per-request instantiation:

- cold start: host build + service workload start + first request (teardown runs per iteration but outside the timed window)
- hot invocation: steady-state latency against the warm instance
- throughput: RPS with 32 concurrent clients against the one instance
- service -> component: the service forwards each request through its wasi:http/client import to an HTTP component workload on a second host (allowed-hosts policy -> egress -> TCP -> router -> per-request instance), measured against both P2 and P3 backends

Adds the svc-http-proxy fixture: a p3 trigger service that proxies a request carrying an `x-backend` header to that authority and answers directly without one. A malformed header is an error, never a silent fallback to the faster direct path.

The gungraun bench gains a matching `service` group (hot_service, service_to_component, cold_service) for instruction-count regression tracking in CI.

Shared bench plumbing moves to benches/common: host/workload builders, request helpers with a 30s bound so a wedged host fails loudly, and a BenchHost handle whose shutdown stops the workload before the host — dropping a host does not abort a running service driver, and leaked instances accumulate until the process aborts.

Run locally: `cargo bench -p wash-runtime --bench service_http`

### DIFF
--- a/crates/wash-runtime/benches/gungraun.rs
+++ b/crates/wash-runtime/benches/gungraun.rs
@@ -10,6 +10,12 @@
 //! component). A cold-path measurement is included to track startup-cost
 //! drift (component compile + host build + first request).
 //!
+//! The `service` group is the instruction-count counterpart of the
+//! `service_http` wall-clock bench: a request served by a long-lived
+//! `svc-http-proxy` service instance (direct), the same request routed
+//! through the service's `wasi:http/client` import to a per-request HTTP
+//! component on a second host (p2/p3), and the service's cold start.
+//!
 //! Requires the `gungraun-runner` binary on `PATH` and `valgrind`
 //! installed:
 //! ```text
@@ -23,43 +29,20 @@ mod common;
 
 use std::{collections::HashMap, hint::black_box, sync::Arc};
 
-use common::Flavor;
+use common::{
+    BenchHost, DIRECT_BODY, Flavor, bench_client, checked_request, engine, http_host_interfaces,
+    service_request, start_backend_host, start_service_host,
+};
 use gungraun::{library_benchmark, library_benchmark_group, main};
 use tokio::runtime::Runtime;
 
 use wash_runtime::{
-    engine::Engine,
     host::{
         HostApi, HostBuilder,
         http::{DevRouter, HttpServer},
     },
     types::{Component, LocalResources, Workload, WorkloadStartRequest},
-    wit::WitInterface,
 };
-
-fn flavor_host_header(flavor: Flavor) -> &'static str {
-    match flavor {
-        Flavor::P2 => "bench-p2",
-        Flavor::P3 => "bench-p3",
-    }
-}
-
-fn engine() -> Engine {
-    Engine::builder().build().expect("failed to build engine")
-}
-
-fn http_host_interfaces(host: &str) -> Vec<WitInterface> {
-    let mut config = HashMap::new();
-    config.insert("host".to_string(), host.to_string());
-    vec![WitInterface {
-        namespace: "wasi".to_string(),
-        package: "http".to_string(),
-        interfaces: ["incoming-handler".to_string()].into_iter().collect(),
-        version: Some(semver::Version::parse("0.2.2").unwrap()),
-        config,
-        name: None,
-    }]
-}
 
 /// Warm host bundled with the runtime that owns it. The host's drop impl
 /// must run inside its tokio runtime, so the two have to die together.
@@ -101,7 +84,7 @@ fn setup_warm(flavor: Flavor) -> Warm {
                     pool_size: 0,
                     max_invocations: 0,
                 }],
-                host_interfaces: http_host_interfaces(flavor_host_header(flavor)),
+                host_interfaces: http_host_interfaces(flavor.host_header()),
                 volumes: vec![],
             },
         };
@@ -116,7 +99,7 @@ fn setup_warm(flavor: Flavor) -> Warm {
         // Warmup primes any one-time lazy state and validates the fixture.
         let warmup = client
             .get(format!("http://{addr}/"))
-            .header("HOST", flavor_host_header(flavor))
+            .header("HOST", flavor.host_header())
             .send()
             .await
             .unwrap();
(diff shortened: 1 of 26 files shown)


crates/wash-runtime/benches/gungraun.rs
```diff
@@ -187,9 +170,114 @@ fn cold_invocation(flavor: Flavor) -> Warm {
...
+        let service = start_service_host().await.unwrap();
+        let (backend, expected_body) = match backend {
+            Some(flavor) => (
+                Some(start_backend_host(flavor).await.unwrap()),
+                flavor.expected_body(),
+            ),
+            None => (None, DIRECT_BODY),
+        };
+        let client = bench_client();
+        checked_request(
```
vados-cosmonic, line 199: NIT: is one request warm enough?

crates/wash-runtime/benches/service_http.rs
```diff
@@ -0,0 +1,284 @@
...
+    start_backend_host, start_service_host,
+};
+
+/// Cold start: build the host, start the service workload, serve + validate
+/// the first request. The scale-from-zero cost of a service. Shutdown runs
+/// per iteration but outside the timed window - it keeps hundreds of
+/// iterations from accumulating live service instances without folding the
+/// stop path into the startup numbers.
+fn bench_cold(c: &mut Criterion) {
+    let rt = Runtime::new().expect("tokio runtime");
```
vados-cosmonic:
```suggestion
    let rt = Runtime::new().expect("cold bench runtime");
```

crates/wash-runtime/benches/service_http.rs
```diff
@@ -0,0 +1,284 @@
...
+            total
+        });
+    });
+    group.finish();
+}
+
+/// Hot latency of the direct path: client -> HTTP server -> service ingress
+/// channel -> live instance handler -> streamed response.
+fn bench_hot_direct(c: &mut Criterion) {
+    let rt = Runtime::new().expect("tokio runtime");
```
vados-cosmonic:
```suggestion
    let rt = Runtime::new().expect("hot bench runtime");
```

crates/wash-runtime/benches/service_http.rs
```diff
@@ -0,0 +1,284 @@
...
+fn bench_service_to_component(c: &mut Criterion) {
+    /// Timed requests per pacing cycle.
+    const BURST: u64 = 64;
+    /// Untimed pause between bursts. (BURST + warmers) / PAUSE stays well
+    /// under the ephemeral-port drain rate.
+    const PAUSE: Duration = Duration::from_millis(200);
+    /// Untimed requests absorbing the post-pause wake-up cost.
+    const WARMERS: usize = 2;
+
+    let rt = Runtime::new().expect("tokio runtime");
```
vados-cosmonic:
```suggestion
    let rt = Runtime::new().expect("svc->component runtime");
```

crates/wash-runtime/tests/fixtures/svc-http-proxy/src/lib.rs
```diff
@@ -0,0 +1,96 @@
+//! A p3 proxy service: `cli/run` parks forever while `http/handler` serves
```
vados-cosmonic:
```suggestion
//! A p3 proxy trigger service: `cli/run` parks forever while `http/handler` serves
```

crates/wash-runtime/tests/fixtures/svc-http-proxy/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+            monotonic_clock::wait_for(u64::MAX).await;
+        }
+    }
+}
+
+impl HttpGuest for Component {
+    async fn handle(request: Request) -> Result<Response, ErrorCode> {
+        match request
+            .get_headers()
+            .get(&"x-backend".to_string())
```
vados-cosmonic: I think this is might be an allocation in the hot path, especially for P3 I think we can avoid this probably with a `OnceLock` or `smol_str` or something

crates/wash-runtime/tests/fixtures/svc-http-proxy/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+            .into_iter()
```
vados-cosmonic: Lazy ask: Do we *need* to `into_iter()` here? I also wonder if we can move the into_iter() and next() into the `Some(....)` branch. Probably unnecessary and optimized out but

crates/wash-runtime/tests/fixtures/svc-http-proxy/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+                let authority = String::from_utf8(value)
+                    .map_err(|_| internal("x-backend header is not UTF-8"))?;
+                proxy(&authority).await
+            }
+            None => Ok(direct_response()),
+        }
+    }
+}
+
+fn internal(msg: &str) -> ErrorCode {
```
vados-cosmonic:
```suggestion
fn internal_err(msg: &str) -> ErrorCode {
```

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but nits, you can take or leave the runtime naming IMO, was just wondering why bother to name them if they're all the same name, but it really doesn't matter because the tests are named well.

crates/wash-runtime/benches/gungraun.rs
ricochet: It should be since we only have one replica therefore one 1 instance, but it doesn't hurt to go ahead and make a couple more passes just to make sure this never is a problem.

crates/wash-runtime/tests/fixtures/svc-http-proxy/src/lib.rs
ricochet: Yep, fixed 👍

crates/wash-runtime/tests/fixtures/svc-http-proxy/src/lib.rs
ricochet: went ahead and made this change a few other optimizations here. Ideally this is a good base now we can build from with the helpers.

vados-cosmonic (approved):
LGTM 🚀 

It's great to have some baseline numbers here!

PR #5384, opened 2026-07-22
fix(ci): handle binstall fallback

ricochet:
We ran into a case of binstall falling back to cargo install, but cargo install not supporting us passing in an install-path. This change now does the install cp in a separate step.


vados-cosmonic (dismissed):
LGTM :rocket:

NIT: What about moving the install portion to something like https://github.com/taiki-e/cache-cargo-install-action (should be cached much of the time)? 

SUPER NIT: I think a wildcard (`*`) could cover us more simply, though detecting windows is fine too

vados-cosmonic (approved):

PR #5385, opened 2026-07-23
Blocks s3 blobstore plugin

bitwoded-TRON:


Aditya1404Sal:
Thank you for this but this isn't how the blobstore plugin is designed to be implemented.

PR #5387, opened 2026-07-24
Use first export wasi:http component in manifest

### LINKED ISSUE #5322: Bug: Inconsistent deployment of http components in `wash dev`
# Description
If you have two wasm components, both exporting the `wasi:http` type. It can lead to inconsistent deployment depending on which component is loaded first.

## Scenario
Given the following `.config/wash.yaml` file:

```yaml
version: 2.5.1

build:
  command: exit 0
  component_path: build/something.wasm

dev:
  components:
    - name: other
      file: build/other.wasm
```

#### Outcome 1
```sh
$ wash dev
...
$ curl localhost:8000
Hello, from something.wasm!
```

#### Outcome 2
```sh
$ wash dev
...
$ curl localhost:8000
Hello, from other.wasm!
```

## Expected
Since `something.wasm` is defined as the primary component, this should link the http server to this component, ignoring the `wasi:http` export of `other.wasm`.

jfleitz:
Fixes #5322 

## Description

Adds a generic `select_exporter` function that will return the first component in manifest order for the export chosen. This could have been just for `wasi:http` only, but thought a generic function made more sense to support future types that must ensure only one component is used.

## Testing

Tested with `wash dev` with a component configured like Issue #5322 . Only the primary component was called, and a warning was logged with `multiple components carry an export the host dispatches to only one component; routing to the first in manifest order and ignoring the rest export="wasi:http" selected="wash-dev-component" ignored=["sidecar"]`

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -1650,6 +1650,9 @@ pub struct UnresolvedWorkload {
     service: Option<WorkloadService>,
     /// All [`WorkloadComponent`]s in the workload
     components: HashMap<Arc<str>, WorkloadComponent>,
+    /// Component IDs in manifest order. Used to pick the first in the manifest
+    /// whenever the host can dispatch an export to only one component
+    component_order: Vec<Arc<str>>,
     /// TLS provider override for `wasi:tls` client connections in this workload.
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5391, opened 2026-07-25
feat(runtime): warm-instance pooling (pool_size / max_invocations)

ricochet:
Implements the `// TODO: Implement pooling and instance limits` in `engine/workload.rs`. `Component.pool_size` and `Component.max_invocations` have been on the proto, the CRD and `types::Component` all along, and were hardcoded to `0` in the runtime — set them today and nothing happens.

A component that sets `pool_size > 0` opts out of per-call instantiation: after a call returns cleanly its store is parked, and the next call reuses it skipping the context build, the `Store` allocation and the instantiation, and letting the guest keep whatever it cached in linear memory (a connection pool, a lazily built runtime, parsed config).

Both paths that build a store per call are covered:

* **linked calls** — a plain-value async call between components (`invoke_ephemeral_plain`)
* **HTTP dispatch (p3)** — a component reached over `wasi:http/handler@0.3`

Also plumbed through `wash dev`, so `.wash/config.yaml` can set `poolSize` / `maxInvocations` per component.

## Measured

Trivial work in every case, so these are the shapes' own costs (`integration_topology_overhead`, included):

| topology | peak req/s |
|---|---|
| component, cold | 19,900 |
| component, warm | **60,100** |
| service (one instance) | 34,500 |
| service + linked component, cold | 11,000 |
| service + linked component, warm | 22,100 |

Warm components overtake a single service because `pool_size` buys real parallelism: each warm instance is an independent store, where a service is one instance every request shares. Per-request cost read at concurrency 1: a store build + instantiation is ~44 µs, an ephemeral store for a linked call ~53 µs, and a warm linked call ~9 µs.

## Why it is opt-in

`Component` is defined as having ephemeral state, and this keeps that as the default. `pool_size: 0` (and unset, which arrives as `-1`) behaves exactly as before.

## The rules that bound an instance's life

* `max_invocations` retires an instance after that many calls (0 = unlimited).
* An instance is parked **only** after a call that returned cleanly. A trap, a
  timeout, a host error, or a caller cancelled mid-call all retire the store
  instead — a guest interrupted partway through may hold locks or half-written
  buffers. This is correct by construction rather than by checking: the store
  travels into the spawned task and back out with the result, so the
  `AbortOnDrop` guard drops it rather than parking it. On the HTTP path the
  release happens only once the response body has been fully forwarded
  (`run_concurrent` does not resolve until then).
* Bursts past `pool_size` are served from fresh stores rather than queued, so
  pooling never adds latency it did not save.

## Linked components share the store's lifetime

A store holds more than the one component: whatever it links to is instantiated into the same store and lives exactly as long as it does. So a store is parked only when **every** component in it has opted in.

## A service needs no `wasi:cli/run`

Folded in here because the topology comparison above needs a service fixture, and the simplest one has no `cli/run`.

`wasi:cli/run` is what a service uses to drive long-running work of its own — an
accept loop, a poller. It is *not* what keeps a service instance alive: the host
instantiates the service once and serves each host-invoked export as a
concurrent task on it, co-driving `cli/run` only when present
(`run_trigger_driver` already treats `Command::new` as optional). That was true
but untested for services — host component plugins covered the optional-`cli/run`
path already (`kv-plugin` exports none), but nothing covered a *service*
without one.

New `svc-no-run` fixture: exports `wasi:http/handler@0.3` and nothing else, and
reports a per-instance request count. Five sequential requests return
`1,2,3,4,5` — one long-lived instance served them all. It mirrors the existing
`svc-counter` fixture minus its `cli/run` tick loop, so the pair isolates
exactly what `cli/run` buys. The same fixture backs the HTTP-dispatch pooling
tests below.

## Tests

`integration_warm_instances` reads a counter out of the guest's linear memory,
so the response says directly whether an instance was reused:

* linked call, unpooled → `1,1,1,1,1`; pooled → `1,2,3,4,5`;
  `max_invocations: 3` → `1,2,3,1,2,3,1`
* HTTP dispatch, same three shapes
* caller opted in with an unpooled callee → stays cold (`1,1,1,1`); both opted
  in → warm (`1,2,3,4`)
* 3,000 sequential requests through one warm instance still serve correctly,
  which a store leaking a resource-table entry per request would not

## Not covered

**The p2 HTTP path (`wasi:http/incoming-handler@0.2`) still instantiates per
request.** Its store is owned by a detached task that outlives the response
head, so recovering it for reuse needs a restructure this PR does not do. Happy
to follow up.

A pooled instance freezes the context it was built with (environment, config,
resolved volume mounts) for its lifetime; `max_invocations` bounds how stale
that can get. Documented on `InstancePool`.

A service that owns a **loopback listener** still needs `cli/run`. Moving an
accept loop into a task spawned from the HTTP handler does not work: the spawned
task is driven (its first poll runs, and outbound sockets from such a task
complete fine), but the virtualized-loopback accept stream never delivers a
connection. Separate runtime gap, not addressed here.

## Reviewer note

`test_bridge_backend_is_stateless` carried `pool_size: 1` while asserting the
backend was stateless — inert boilerplate until now. It says `0` explicitly.
Worth knowing that `pool_size: 1` appears as boilerplate in ~25 test files; the
rest pass because their guests are stateless or are not reached via a pooled
path, but the values now mean something.

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -22,6 +22,7 @@ use crate::engine::ctx::SharedTlsProvider;
 use crate::{
     engine::{
         ctx::SharedCtx,
+        instance_pool::InstancePool,
         linked_call::{
             ComponentCtxTemplate, EphemeralCallMode, EphemeralLinkedCall, LinkedExportInvocation,
             func_is_bridge_safe, func_is_ephemeral_safe, invoke_linked_async_export,
@@ -345,15 +346,19 @@ pub struct WorkloadComponent {
     name: Arc<str>,
     /// The [`WorkloadMetadata`] for this component
     pub(crate) metadata: WorkloadMetadata,
-    /// The number of warm instances to keep for this component
-    pool_size: usize,
-    /// The maximum number of concurrent invocations allowed for this component
-    max_invocations: usize,
+    /// Instances kept warm between ephemeral linked calls. Shared by every
+    /// clone of this component, so all importers draw on the one warm set.
+    pub(crate) instances: Arc<InstancePool>,
 }
 
 impl WorkloadComponent {
     /// Create a new [`WorkloadComponent`] with the given workload ID,
     /// wasmtime [`Component`], [`Linker`], volume mounts, and instance limits.
+    ///
+    /// `pool_size` and `max_invocations` are the component's warm-instance
+    /// limits; both are `-1` when unset. `pool_size` of zero (or unset) keeps
+    /// the default behavior of a fresh store per linked call — see
+    /// [`InstancePool`] for what opting in changes.
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         workload_id: impl Into<Arc<str>>,
@@ -365,6 +370,8 @@ impl WorkloadComponent {
         volume_mounts: Vec<(PathBuf, VolumeMount)>,
         local_resources: LocalResources,
         loopback: Arc<std::sync::Mutex<loopback::Network>>,
+        pool_size: i32,
+        max_invocations: i32,
     ) -> Self {
         Self {
             metadata: WorkloadMetadata {
@@ -382,9 +389,7 @@ impl WorkloadComponent {
                 linked_components: Default::default(),
             },
             name: component_name.into(),
-            // TODO: Implement pooling and instance limits
-            pool_size: 0,
-            max_invocations: 0,
+            instances: Arc::new(InstancePool::new(pool_size, max_invocations)),
         }
     }
 
@@ -458,8 +463,7 @@ impl std::fmt::Debug for WorkloadComponent {
             .field("id", &self.metadata.id.as_ref())
             .field("workload_id", &self.metadata.workload_id.as_ref())
             .field("volume_mounts", &self.metadata.volume_mounts)
-            .field("pool_size", &self.pool_size)
-            .field("max_invocations", &self.max_invocations)
+            .field("warm_instances", &self.instances.enabled())
             .finish()
     }
 }
@@ -1557,6 +1561,12 @@ impl ResolvedWorkload {
         );
 
         for component in self.components.read().await.values() {
+            // Warm instances hold guest resources (sockets, open files) for as
+            // long as they stay parked, so release them with the rest of the
+            // workload's teardown. Calls still in flight own their own stores
+            // and are unaffected.
+            component.instances.clear();
+
             if let Some(plugins) = component.plugins() {
                 for (plugin_id, plugin) in plugins.iter() {
                     trace!(
@@ -2533,6 +2543,8 @@ mod tests {
             Vec::new(),
             local_resources,
             Arc::default(),
+            0,
+            0,
         )
     }
 
@@ -2594,6 +2606,8 @@ mod tests {
             Vec::new(),
             local_resources,
             Arc::default(),
+            0,
+            0,
         )
     }
 
@@ -2685,6 +2699,8 @@ mod tests {
                 Vec::new(),
                 LocalResources::default(),
                 Arc::default(),
+                0,
+                0,
             )
         };
         for (plugin, iface) in cases {
@@ -2944,6 +2960,8 @@ mod tests {
             Vec::new(),
             LocalResources::default(),
             Arc::default(),
+            0,
+            0,
         );
 
         let mut http_interface = WitInterface::from("wasi:http/handler");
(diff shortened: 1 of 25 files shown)


crates/wash/src/cli/dev.rs
```diff
@@ -419,6 +419,10 @@ struct LoadedComponent {
     bytes: Bytes,
     interfaces: HashSet<WitInterface>,
     workload: ResolvedWorkload,
+    /// Warm-instance limits from `dev.components[].poolSize` /
+    /// `maxInvocations`. `-1` where the config left them unset.
+    pool_size: i32,
```
vados-cosmonic: Hmnn that's a weird option, why not an option or an enum properly capturing the state?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1462,6 +1466,25 @@ impl ResolvedWorkload {
         .await
     }
 
+    /// The pool a request store for `component_id` may be parked in, or `None`
+    /// when it must be built and dropped per request.
+    ///
+    /// The component's linked components are instantiated into the same store
+    /// by [`Self::new_store`] and live exactly as long as it does, so they have
+    /// to have opted in too — see [`instance_pool::poolable`].
+    pub(crate) async fn instance_pool_for(
```
```suggestion
    pub(crate) async fn instance_pool_for_component(
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -458,8 +463,7 @@ impl std::fmt::Debug for WorkloadComponent {
             .field("id", &self.metadata.id.as_ref())
             .field("workload_id", &self.metadata.workload_id.as_ref())
             .field("volume_mounts", &self.metadata.volume_mounts)
-            .field("pool_size", &self.pool_size)
-            .field("max_invocations", &self.max_invocations)
+            .field("warm_instances", &self.instances.enabled())
```
```suggestion
            .field("has_warm_instances", &self.instances.enabled())
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -345,15 +346,19 @@ pub struct WorkloadComponent {
...
+    /// clone of this component, so all importers draw on the one warm set.
+    pub(crate) instances: Arc<InstancePool>,
 }
 
 impl WorkloadComponent {
     /// Create a new [`WorkloadComponent`] with the given workload ID,
     /// wasmtime [`Component`], [`Linker`], volume mounts, and instance limits.
+    ///
+    /// `pool_size` and `max_invocations` are the component's warm-instance
+    /// limits; both are `-1` when unset. `pool_size` of zero (or unset) keeps
```
vados-cosmonic: Can we use an option or another enum that represents the state explicitly here?

crates/wash-runtime/src/host/http.rs
```diff
@@ -1623,12 +1623,25 @@ async fn invoke_component_handler(
...
+        // Reuse a warm instance when the component has opted in, otherwise
+        // build and instantiate a store for this request alone. A burst past
+        // `pool_size` is served cold rather than queued.
+        let pool = workload_handle.instance_pool_for(component_id).await;
+        let warm = match pool.as_ref().and_then(|pool| pool.checkout()) {
+            Some(warm) => warm,
+            None => {
+                let mut store = workload_handle.new_store(component_id).await?;
+                let instance = instance_pre.instantiate_async(&mut store).await?;
+                crate::engine::instance_pool::WarmInstance {
```
vados-cosmonic: NIT: Hmnn slightly against calling this a "warm" instance if it's not (and shouldn't be). 

What if we had an enum that encapsulated both of these cases? Just spitballing but:

```
enum ExecutingInstance {
    Pool(...),
    EphemeralSingle(...),
}
```

Maybe over-engineering and fine with doing this in a follow up if you agree because it would be a tidy PR I think

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but Nits, and glad to see the perf improvement!

crates/wash/src/cli/dev.rs
ricochet: Ok swapped this over to an InstancePolicy enum:

pub enum InstancePolicy {
    Ephemeral,
    Warm { pool_size: NonZeroUsize, max_invocations: Option<NonZeroUsize> },
}

 WorkloadComponent::new takes the policy instead of two i32s

crates/wash-runtime/src/engine/workload.rs
ricochet: yep see comment above for dev.rs

```rust
pub enum InstancePolicy {
    Ephemeral,
    Warm { pool_size: NonZeroUsize, max_invocations: Option<NonZeroUsize> },
}
```

crates/wash-runtime/src/host/http.rs
ricochet: Changed to ComponentInstance mainly since it's still a reactor and might not be "executing" per se.

jfleitz (approved):
This is great and will be a big performance gain! LGTM, especially since this is a ln opt in feature to do warm instances with the default pool side of zero.

PR #5393, opened 2026-07-27
feat(runtime): surface component-model external-id

### LINKED ISSUE #5392: [FEAT] Surfacing `external-id` in wasmCloud
Design for adopting the component-model `external-id` attribute ([Explainer §import-and-export-definitions](https://github.com/WebAssembly/component-model/blob/c940736c/design/mvp/Explainer.md#import-and-export-definitions),
[wasm-tools#2555](https://github.com/bytecodealliance/wasm-tools/pull/2555), [wasm-tools#2558](https://github.com/bytecodealliance/wasm-tools/pull/2558)) across the wash-runtime, `wash dev`, and the runtime operator.

## What `external-id` is

An optional attribute on component imports and exports, sibling to `implements` and gated by the same 🏷️ feature:

```ebnf
import   ::= (import <externnamelit> <attribute>* bind-id(<externtype>))
attribute ::= (implements <interfacenamelit>) | (external-id <name>) | ...
```

It is a **free-form Unicode string, ignored by type checking**. Names identify imports/exports; attributes tell the host *how* to satisfy them. `external-id` closes the expressivity gap between the restricted `externname` syntax and whatever identifier scheme a platform already uses. The spec's motivating example is exactly wasmCloud's multi-backend problem:

```wat
(component
  (import "users"
    (implements "wasi:keyvalue/store") (external-id "user-db-prod:region-a")
    (instance ...))
  (import "catalog"
    (implements "wasi:keyvalue/store") (external-id "catalog-db-prod:region-a")
    (instance ...)))
```

In WIT (wit-parser ≥ 0.253):

```wit
world inventory {
  @external-id("user-db-prod:region-a")
  import users: wasi:keyvalue/store@0.2.0-draft;

  @external-id("catalog-db-prod:region-a")
  import catalog: wasi:keyvalue/store@0.2.0-draft;

  @external-id("inventory.example.com")
  export wasi:http/incoming-handler@0.2.2;
}
```

It applies to plain interface imports/exports too, not only `implements`-labeled ones, and to individual funcs and types.

Toolchain surface in wasm-tools 0.253.0:

- `wit_parser::Resolve::external_id_value(&WorldKey, &WorldItem) -> Option<String>`
- `external_id: Option<String>` on `WorldItem::Interface`, `Function`, `TypeDef`
- `wit_component::WitPrinter` prints `@external-id(...)` automatically
- wasmparser: new `ComponentNameOpt::ExternalId` binary name-option tag, validated behind the `cm-implements` feature (same gate as `implements`)

## Two names and a contract, and what wasmCloud does today

Every import has exactly one *name*. On a plain interface import that name must be the interface name itself (`wasi:keyvalue/store@0.2.0`), so it plays two roles at once: what the import is called and what contract it satisfies. The `implements` attribute splits those roles. The import gets a single developer-chosen name (the **implements name**, e.g. `users`, which is what wasmCloud routes on). `external-id` adds a new piece of metadata:

| Metadata | Role | Who authors it | Example | wasmCloud today |
|---|---|---|---|---|
| import name (the implements name, when labeled) | the component's name for the dependency; routing key | component author | `users` | `WitInterface.name`, multiplex registry key |
| interface contract — the name itself, or the `implements` value when labeled | scopes which plugins/bindings are candidates | the WIT package | `wasi:keyvalue/store` | `WitInterface` namespace/package/interfaces |
| `external-id` | the platform's name for the concrete resource; binding selector | platform / deploy pipeline | `user-db-prod:region-a` |  |

Today the platform-side resource name is authored out-of-band, keyed on the implements name:

- **Operator**: `HostInterface{Name, Config}` on the Workload CRD (`runtime-operator/api/runtime/v1alpha1/workload_types.go:145`) → proto `WitInterface.name` (field 6, `proto/wasmcloud/runtime/v2/wit_interface.proto:15`) → runtime `WitInterface.name` (`crates/wash-runtime/src/wit.rs:141`).
- **Runtime**: `WorkloadComponent::world()` reads `import_item.implements` from wasmtime reflection and sets `iface.name = import label` (`crates/wash-runtime/src/engine/workload.rs:200`); the multiplex core keys backends by that name (`crates/wash-runtime/src/plugin/multiplex.rs:291`) and resolves import label → backend at call time (`multiplex.rs:304`).
- **wash dev**: `.wash/config.yaml` `dev.host_interfaces` entries (`crates/wash/src/config.rs:412`) provide the same `name` + `config` pairs.

Consequence: component author and operator must coordinate on **implements names**, which are source-language-idiomatic, per-component, and invisible to the platform's own naming. `external-id` lets the artifact itself carry the platform-native name, so a binding can be expressed once, platform-side, and match every component that references that resource or need no binding at all.

Essentially platform operators (or the automated processes deploying to that platform), may inject metadata into the external-id annotation.

## Design principles

1. **Precedence: deploy-time config > binary `external-id` > conventions.** The operator always can override what a binary declares; a binary's external-id always beats a bare default. Build-time metadata never removes operator control.
2. **`external-id` is identification, not authorization.** Matching an external-id never grants access the operator didn't configure. For imports this is already true structurally: guests today pass arbitrary identifiers to `store::open(identifier)` within a configured backend, so treating an external-id as a resource identifier inside a bound backend grants no new authority. Export-side identities (e.g. hostnames) *are* new authority and need operator opt-in (§7).
3. **Selection and identification are separate axes.** The implements name and external-id serve different functions and never compete: a binding (matched by name, by external-id, or the default) picks the backend, and the external-id always travels to that backend as the resource's platform name. Two distinct bindings matching one import is operator misconfiguration, rejected explicitly and not resolved by ranking the attributes.
4. **One feature gate.** Reuse `wasm_component_model_implements` (crate feature + `WasmProposal::WasmComponentModelImplements`). The binary validation of external-id is gated behind the same `cm-implements` wasmparser feature upstream, so they ship together.
5. **Attributes are droppable.** Because attributes are ignored by type  checking, tooling may inject or rewrite external-ids post-build (a  deploy-pipeline step can stamp `user-db-prod:region-a` into a genericartifact without recompiling). The design should not assume external-ids are only authored in source WIT.

## Dependencies

- Not yet in a wasmtime release.
- wasm-pkg-core / wasm-pkg-client need updating.
- wit-bindgen-core 0.51 is on wit-parser 0.244. Guest toolchains need awit-bindgen on wit-parser ≥ 0.253 to build worlds that use `@external-id`

## Runtime (`wash-runtime`)

Add to `crates/wash-runtime/src/wit.rs` next to `name`:

```rust
/// Component-model `external-id` attribute: the platform-native identity of
/// the external resource this interface binds to.
pub external_id: Option<String>,
```

Treatment mirrors `config`, not `name`:
- **Excluded from `Hash`/identity** and from `instance()` (the merge key stays
  `namespace:package[/name]@version`; the implements name already
  disambiguates world items).
- `merge()` unions it (error on conflicting non-equal values for the same
  instance key).
- `contains()` gains one rule: a host-supplied interface **with**
  `external_id` set matches only a request with an equal `external_id`
  (it's a selector); one **without** matches anything (today's wildcard
  behavior, unchanged).

**Binding selection** answers "which backend serves this import"; **resource identification** answers "which resource within that backend". The implements name is a binding-selection key; the external-id is primarily the resource's platform name, and secondarily a binding-selection key for platform-authored bindings.

**Selection.** Within the import's interface contract (`ns:pkg` scope), an import with implements name `L` and external-id `X` matches:

- a `host_interfaces` entry with `name == L` (exactly today's behavior), or
- an entry with `external_id == X`, or
- an entry declaring both, in which case both must match (composition), or
- failing all of those, the default (unkeyed) entry for `ns:pkg`.

**Identification.** Whichever entry is selected, the external-id always travels with the import to the backend (it is on the `WitInterface` handed to the plugin at bind time). Both attributes pass through, never one or the other. Whether it parameterizes the backend follows principle 1: explicit binding `config` wins; where config doesn't pin the resource aka most notably the default route, the external-id is the resource identifier (§5.4).

Mechanically, in `plugin/multiplex.rs`: ambiguity detection and entry selection happen once at `on_workload_item_bind` time, where the plugin has the component's imports (with `name` and `external_id`) and the matched `host_interfaces`. The registry is then keyed by import name exactly as today (`build_registry_with`, `:273`), so the bindgen resolve callback (`|name| self.mux.resolve(&registry, name)` — `wasi_keyvalue/multiplexed.rs:398`) stays a plain lookup with no call-time precedence logic.

The built-in multiplexed plugins interpret an import's external-id as the
resource identifier where binding config doesn't pin one — the spec's "host
implementation dynamically parameterized by store names" — most notably on
the default route:

- `wasi:keyvalue` — default bucket for `store::open` / the store identity for
  that import.
- `wasi:blobstore` — container (namespaced under the backend's configured
  root).
- `wasmcloud:messaging` — subject/queue prefix.
- `wasmcloud:postgres` — database name within the configured server.

Each builtin documents its interpretation. This is not a new capability
grant — guests already choose identifiers via `open()` within whatever backend
the operator bound.

## `wash dev` and the wash CLI

`dev.host_interfaces` entries (already deserialized straight into `WitInterface`, `crates/wash/src/config.rs:412`) gain the same optional `external-id` key:

```yaml
dev:
  host_interfaces:
    - namespace: wasi
      package: keyvalue
      interfaces: [store]
      external-id: "user-db-prod:region-a"
      config:
        backend: redis
        url: redis://localhost:6379
```

ricochet:
Adds the `external-id` attribute alongside `implements`

Implements name the component's import/dependency, the external-id is the reference config/name/id of the platform. A binding matches by name, by external-id, by both, or is the unkeyed default.

Requires wasmparser 0.253 to read the attribute at all, so the workspace patches wasmtime to a 47 backport until 48 ships it.

Fixes #5392


ricochet:
Note depending on the order of landing, https://github.com/wasmCloud/wasmCloud/pull/5381/ may need to surface external-id in the `workload-lifecycle` interface.

PR #5395, opened 2026-07-27
feat: async wasmcloud:secrets + secrets host component plugin

ricochet:
Depends on #5381 

This change revs `wit/secrets` from sync `@1.0.0` to async **`@2.0.0`** (`get`/`reveal` are `async func`), adds a test to excercise it and an example for how this can be used in a guest component (oci-reg with basic auth).

For Host Component Plugins, they require async API's to work. A host component plugin serves capability calls across its store boundary, which is inherently async. Right now the sync interface cannot be plugin-served (the cross-store shim only type-matches async imports). I did try hacking together basically a wasm shim composed in with an async wrapper but that added overhead.

The test shows how secrets can be delivered per-workload at `on-workload-bind` (operator sources it from `secretFrom`) and correlated to the caller via `wasmcloud:host/identity`.

### DIFF
--- a/crates/wash-runtime/tests/common/mod.rs
+++ b/crates/wash-runtime/tests/common/mod.rs
@@ -168,6 +168,36 @@ pub fn kv_plugin_caller_host_interfaces_with_config(
     ]
 }
 
+/// The `wasmcloud:secrets` capability (store + reveal), provided by the
+/// `secrets-host` host component plugin.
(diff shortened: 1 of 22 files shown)


vados-cosmonic (dismissed):
LGTM 🚀 

I assume we'll have a reusable secret host component somewhere that's not a fixture soon

PR #5397, opened 2026-07-27
gungraun host component plugin bench

ricochet:
Depends on #5381

Adds `gungraun_plugin`, a cachegrind bench covering a host component plugin's surface: the relocation paths a capability call can take, the `wasmcloud:host/workload-lifecycle` hooks, and bringing an incarnation up.

## Why instruction counts and not wall-clock

A criterion version of these measurements was written first and thrown away.

Driven over HTTP, a plugin-backed request (152.6 µs [149.4, 155.9]) and a plugin-free one (164.9 µs [141.2, 193.7]) land inside each other's confidence intervals. The cross-store hop is smaller than the harness's noise floor. Both arms also topped out around 19–20k rps, which is the in-process load generator's HTTP ceiling rather than any plugin limit. Worse, criterion's confidence intervals describe sampling variance *within* one run, not variance *between* runs.

Cachegrind counts are deterministic and have neither failure mode.

## What it measures

**`capability_call`**: one HTTP request per arm, every arm against the same warm host and the same `kv-plugin-caller` workload, differing only in which route it drives. So an arm minus `no_plugin` is that mechanism end to end, and an arm minus `get` is what the mechanism costs above a plain handle-free call.

| arm | mechanism |
| --- | --- |
| `no_plugin` | `http-handler-p3`: HTTP and component-invoke cost with no plugin involvement — the baseline |
| `get` | handle-free primitives across the store boundary, the relocation fast path |
| `resource` | a cross-store `own<bucket>` proxy: open it, `set`/`get` through borrow methods, drop it. Exercises the resource bridge and the drop flush, which steps out of `run_concurrent` to run the guest destructor |
| `stream_in` | caller → plugin `stream<u8>` |
| `stream_out` | plugin → caller `stream<u8>` |
| `future` | plugin → caller `future<u64>`, resolved across the boundary |
| `reentrant` | the plugin's self-import: it calls its own capability, re-entering the store it is already running in |

These are distinct paths through `relocate` and the resource bridge. The plugin is loaded on the host in *every* arm including `no_plugin`, so the delta is the mechanism rather than the cost of having a plugin present.

## bench-tools fix

`bench-tools` classified benches by an exact `== "gungraun"` match in three places — `summary`, `jsonl`, and `delta`. Any second gungraun-harness bench would have been routed down the criterion path, where it finds no output and reports nothing: a silent empty result, not an error. All three now share one prefix predicate in `callgrind.rs`, matching how the shell scripts already key off the target name.

`run-bench.sh` and `compare-bench.sh` are widened from exact-match to the same `gungraun*` prefix, so the new bench inherits isolated-CPU pinning and the one-run-per-ref comparison regime rather than being silently treated as a wall-clock bench.

## Running it

```sh
cargo xtask build-fixtures
cargo bench -p wash-runtime --features host-component-plugins --bench gungraun_plugin
```

`run-bench.sh` adds the feature flag itself, so a local `./scripts/bench/run-bench.sh gungraun_plugin` behaves identically to CI.

### DIFF
--- a/.github/workflows/bench-compare.yml
+++ b/.github/workflows/bench-compare.yml
@@ -40,6 +40,7 @@ on:
           - gungraun
           - wasmtime_baseline
           - wasmtime_serve
+          - gungraun_plugin
       ref_a:
         description: baseline ref (branch, tag, sha, or #PR)
         type: string
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -13,6 +13,7 @@ on:
           - gungraun
           - wasmtime_baseline
           - wasmtime_serve
+          - gungraun_plugin
       ref:
         description: git ref to bench (branch, tag, or sha; default = the workflow ref)
         type: string
--- a/crates/bench-tools/src/callgrind.rs
+++ b/crates/bench-tools/src/callgrind.rs
@@ -165,3 +165,16 @@ pub fn read_ir(path: &Path) -> Result<Option<u64>> {
 pub fn dir_from_target(target_dir: &Path) -> PathBuf {
     target_dir.join("gungraun")
 }
+
+/// Whether `bench` is an instruction-count bench rather than a wall-clock one.
+///
+/// Every gungraun-harness bench target is named `gungraun*` (`gungraun`,
+/// `gungraun_plugin`, ...) and writes to `$CARGO_TARGET_DIR/gungraun/` instead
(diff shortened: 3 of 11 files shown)


vados-cosmonic (dismissed):
LGTM 🚀 

Seems like what's mostly changed here is the rbase to include the new lifecycle code (gungraun stuff looks fine to me from the last time I looked at it)

PR #5398, opened 2026-07-28
feat: concurrent pooled instances

### LINKED ISSUE #5396: feat(runtime): serve concurrent calls on each pooled instance
`pool_size` (in #5391) keeps component instances warm, but each pooled instance serves **one call at a time**. We can drive pooled components the same as how we do Services and Host Component plugins with concurrent tasks, times `pool_size`. This is especially useful for workloads that are I/O bound awaiting network calls. 

## Proposed shape

`InstancePool` stops parking values and starts owning drivers.

```
InstancePool
 ├── policy: InstancePolicy       // pool_size, max_invocations, max_concurrency
 └── drivers: [InstanceDriver; <= pool_size]

InstanceDriver
 ├── tx: mpsc::Sender<InstanceJob>
 ├── in_flight: Arc<AtomicUsize>  // admission
 ├── served: Arc<AtomicUsize>     // against max_invocations
 └── task: JoinHandle             // owns the Store, runs run_concurrent
```

Each driver is `run_trigger_driver` with the ingress specialized to "a call on this component":

```rust
store.run_concurrent(async |accessor| {
    while let Some(job) = rx.recv().await {
        accessor.spawn(InstanceTask { job, in_flight: guard, .. })?;
    }
}).await
```

### Routing and admission

A call picks a driver instead of taking a store:

1. Choose the live driver with the lowest `in_flight`
2. If it is under `max_concurrency`, claim a slot and send the job.
3. If every driver is saturated and `drivers.len() < pool_size`, start another.
4. If the pool is at `pool_size` and all are saturated, fall back to a cold one-shot store.

## Prior art in-tree

Most of this machinery exists:

- `trigger_service::run_trigger_driver` — long-lived `run_concurrent`, mpsc ingress, `accessor.spawn` per request (`trigger_service/mod.rs:201`).
- `trigger_service::capability` — the same *plus* the two pieces the HTTP ingress lacks: a non-blocking admission ceiling (`InFlightGuard` over an `AtomicUsize`, `MAX_INFLIGHT_CAPABILITY_CALLS`) and per-call cancellation via `JobRegistry`/`JobGuard`, using `call_concurrent` so calls interleave.
- `plugin::component_host` — supervises such a driver and rebuilds it after a fault, with a restart budget.

So this is generalizing the trigger-service model from one singleton instance to
a supervised pool of N, with routing in front.

`InstancePolicy` was made `#[non_exhaustive]` in #5391 specifically so `max_concurrency` can be added without a breaking change, and `InstancePolicy::from_component` takes the whole component so a new limit needs no signature change.

## Acceptance

The existing benches cannot show this: every fixture returns immediately, so per-instance concurrency has nothing to overlap. Needs a fixture whose handler awaits (monotonic clock, or an outbound call), then:

> `pool_size: 1`, concurrency 8, handler awaiting 10 ms — serial today
> (~80 ms per batch), ~10 ms once concurrent.

## Open questions

- **Per-call fuel/metering** on a shared instance. Already unresolved on the service path (`let _ = &fuel_meter`), so this inherits rather than creates the problem. Worth deciding how to do this for both cases.
- **Cancellation.** We will need something like the `JobRegistry`. Note the known limit: wasmtime cannot host-cancel a guest `call_concurrent` subtask.
- **Ordering.** Nothing guarantees arrival order today either, but concurrency makes interleaving observable to guests that keep state. Worth documenting.

ricochet:
Fixes #5396 

For the benchmark in this PR, this resulted in a ~13.6x improvement.


crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -659,70 +660,83 @@ async fn invoke_ephemeral_relocated(
...
+                trace!(
+                    name = %inv.import_name,
+                    fn_name = %inv.export_name,
+                    ?results,
+                    "invoked ephemeral dynamic export"
+                );
+                return Ok(());
+            }
+            // Every warm instance was busy; run it in a store of its own.
+            Err(_declined) => {}
```
vados-cosmonic: tracing this might be useful

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -659,70 +660,83 @@ async fn invoke_ephemeral_relocated(
...
         }
-    };
+    }
+
+    let mut store = new_ephemeral_store(ephemeral_call)
+        .await
+        .map_err(|e| wasmtime::format_err!("{e:#}"))?;
```
```suggestion
        .map_err(|e| wasmtime::format_err!("new ephemeral store creation failed: {e:#}"))?;
```

crates/wash-runtime/src/washlet/mod.rs
```diff
@@ -366,6 +366,23 @@ fn image_pull_secret_to_oci_config(
...
+/// The instance limits a component declared on the wire, carried through
+/// verbatim so the runtime decodes them once (see
+/// [`crate::engine::InstancePolicy`]).
+///
+/// A named seam rather than three field reads inline: `max_concurrency` was
+/// added to `types::Component` before it existed on the wire, and this
+/// conversion silently pinned it to 1 — the knob was unreachable from a
+/// deployed workload while looking wired everywhere else. Extracting it makes
+/// the wire path testable without an image pull.
+fn instance_limits(component: &types::v2::Component) -> (i32, i32, i32) {
```
vados-cosmonic: Not sure this is necessary...

crates/wash-runtime/src/types.rs
```diff
@@ -67,6 +67,15 @@ pub struct Component {
...
     pub max_invocations: i32,
+    /// How many calls one warm instance may serve at the same time.
+    ///
+    /// Unset or below `1` means one, which is what a component gets without
+    /// asking: a warm instance serves a single call at a time, exactly as an
+    /// unpooled one does. Raising it lets an instance overlap calls while it is
+    /// awaiting I/O, and is only safe for a guest that yields rather than
+    /// blocks — a guest driving its own executor with `block_on` must stay at
+    /// one.
+    pub max_concurrency: i32,
```
vados-cosmonic, line 84: We almost certainly want stuff like this and pool size and max invocations to be `u32`s... But I assume this is here because of serialization? But then again I don't see deserialize?

proto/wasmcloud/runtime/v2/workload.proto
```diff
@@ -93,6 +93,14 @@ message Component {
...
   string name = 6;
   ImagePullPolicy image_pull_policy = 7;
+  // How many calls one warm instance may serve at the same time.
+  //
+  // Unset or below 1 means one, which is what a component gets without
+  // asking: an instance serves a single call at a time. Raising it lets an
+  // instance overlap calls while it awaits I/O, and is only safe for a guest
+  // that yields rather than blocks -- a guest driving its own executor with a
+  // blocking call must stay at one. Only meaningful alongside pool_size.
+  sint32 max_concurrency = 8;
```
vados-cosmonic, line 108: why not a uint here? are we forced into making this a signed int for some reason?

crates/wash-runtime/src/engine/instance_driver.rs
```diff
@@ -0,0 +1,462 @@
...
+            func.call_concurrent(accessor, &params, &mut results),
+        )
+        .await
+        {
+            Ok(Ok(())) => Ok(results),
+            // A host error mid-call leaves the guest in an indeterminate
+            // state; a trap will also fault the whole driver, but retiring
+            // here covers the errors that do not.
+            Ok(Err(e)) => {
+                self.slot.retire_instance();
```
vados-cosmonic, line 182: debug/warn here would be useful

crates/wash-runtime/src/engine/instance_pool.rs
```diff
@@ -153,101 +178,135 @@ impl InstancePolicy {
...
-            trace_retire("warm set full", instance.invocations);
+
+        // Reap instances that have drained or whose store faulted, so a
+        // retired one frees its place in the pool.
+        drivers.retain(|d| !(d.is_gone() || d.is_retired() && d.in_flight() == 0));
+
+        // Least-busy first: spread calls rather than filling one instance, so
+        // a trap takes down as little as possible.
+        let mut candidates: Vec<Arc<InstanceDriver>> = drivers.iter().map(Arc::clone).collect();
+        candidates.sort_by_key(|d| d.in_flight());
```
vados-cosmonic: I think this is a lot of work to do in this area, but IMO we can try to tighten this up later. Doing essentially GC on the pool every time seems over-eager, and doing a sort on candidates to find the in flight ones seems like it could be amortized.

crates/wash-runtime/src/engine/instance_pool.rs
```diff
@@ -153,101 +178,135 @@ impl InstancePolicy {
...
+        drop(drivers);
+        driver.try_send(job)
     }
 
-    /// How many instances are parked right now. Test/observability only.
-    #[cfg(test)]
-    pub(crate) fn idle_len(&self) -> usize {
-        self.idle.lock().map(|idle| idle.len()).unwrap_or(0)
+    /// Whether this component keeps instances warm at all.
+    pub(crate) fn enabled(&self) -> bool {
```
vados-cosmonic:
```suggestion
    pub(crate) fn warms_instances(&self) -> bool {
```

NIT: can we come up with something more descriptive here?

vados-cosmonic (dismissed):
LGTM 🚀 

Mostly nits and follow ups, one concern about i32s though but I'm probably missing some context on why we need to use them over u32s.

crates/wash-runtime/src/types.rs
ricochet: Yep, they're signed because they mirror the Kubernetes CRD fields. The proto mirrors the CRD and the Rust type mirrors the proto. It's an anti-pattern in k8s to use unsigned integers.

https://raw.githubusercontent.com/kubernetes/community/master/contributors/devel/sig-architecture/api-conventions.md

crates/wash-runtime/src/washlet/mod.rs
ricochet: yeah goofy.  Replaced with component_from_wire

proto/wasmcloud/runtime/v2/workload.proto
ricochet: see above that we're starting from a k8s field. I don't want to do conversion in the operator.

crates/wash-runtime/src/engine/instance_pool.rs
ricochet: yeah good call. Renamed to warms_instances() which matches InstancePolicy::keeps_instances_warm

crates/wash-runtime/src/engine/instance_pool.rs
ricochet: Definitely plenty of options here. I mainly wanted to get the API surfaced in this pass. 

I did go ahead and improve this a bit by no longer cloning every Arc into a fresh Vec per call. It sorts the pool in place, so a near-sorted vec makes the next sort near-linear.

ricochet:
The failing e2e test is also failing on main so unrelated to this change (I'm chasing it down now).

crates/wash-runtime/src/engine/instance_pool.rs
```diff
@@ -34,9 +39,9 @@
 //!    that can get.
 //!  * A task the guest spawned but did not await lives on the store's
 //!    concurrent state. Dropping the store per call used to discard it; a
-//!    parked store carries it to the next call, where it resumes alongside
-//!    that call's work. A guest that spawns background work and relies on it
-//!    being torn down with the call should not be pooled.
+//!    warm instance carries it forward, where it resumes alongside later
+//!    calls. A guest that spawns background work and relies on it being torn
+//!    down with the call should not be pooled.
```
jfleitz, line 44: Apologies, but couldn't find and just wanted to make sure, but what about where a warm instance is created up to max allowed from a scaling scenario....and then all is served. Is the connection pool automatically cleaned up after a given time (scaled down to min?) or do the new instances just stay around?  Or is it that max_concurrency > 1 is only used when a current instance is blocked, so it adds another instance (if less than max_currency), then on complete it auto-cleans up? Just wondering if there needs to be some sort of check based on a time/timeout...maybe not.

ricochet: pool_size is our ceiling for number of instances. They are created lazily and not reaped. max_concurrency controls how many calls one instance serves simultaneously aka max_concurrency never adds instances.

The only things that ever end a warm instance include: max_invocations, a trap, a timeout/host-error retirement, or workload stop (clear()). We could add a time-based reaper, but warm instance pooling landed in 2.6 and is separate from using the pool for concurrent calls.

ricochet: I filed https://github.com/wasmCloud/wasmCloud/issues/5430. It's a great callout and easy to add.

jfleitz: Ahh..got it. Thank you!

jfleitz (approved):

PR #5399, opened 2026-07-28
Component local routing / same-host support

### LINKED ISSUE #5369: [Feature] Same-host routing between components
When two components are scheduled on the same host, add the ability to call the other component without having to exit the host to the ingress / router. This will have performance gains since it eliminates the network hop. Since scheduled workloads can be scheduled for serving path-based routing vs just FQDN, an ingress name / prefix will need to be added to the configuration of scheduled Workloads if needed.

Since this feature will not be exiting the host back to the network:
- Disabled by default
- Runtime feature flag to enable
- May not want to enable due to network controls (you need to know what you are doing)
 - Should have in-memory map to route

jfleitz:
Fixes #5369

## Description

This adds an option to enable local host routing, where a component calling another component on the same host can bypass an egress call and just call in-memory to the component. 

This is accomplished by adding a check in the outbound http handler to check for a local route, if configured: `allowedHosts` → gRPC → local routing → network. 

## How to configure

**Note::** Since `wash dev` only has one component, and no "internal routing to components" supported, a warning is thrown if localRouting is configured. This is to catch any malformed local route config up-front vs deployed in a `wash host`.

### Enable Local Routing on Host

Similar to how `allowedHostLoopback` + `--allow-host-loopback` work, two keys are added for enabling this feature:

| Key | Who sets it | What it means |
| --- | --- | --- |
| `--http-local-routing` (`WASH_HTTP_LOCAL_ROUTING`), chart `runtime.hostGroups[].http.localRouting` | host operator | this host is *willing* to short-circuit |
| `config.localRoute` on a workload's `wasi:http/incoming-handler` interface | workload author | this workload *offers* these routes in-memory |

**

Example:

```yaml
runtime:
  hostGroups:
    - name: default
      replicas: 1 # both workloads must land on the SAME host
      http:
        enabled: true
        port: 9191
        localRouting: true # <<-- enable here
```

Or, running a host directly:

```shell
wash host --http-addr 0.0.0.0:9191 --http-local-routing
```

### Configure Local Route on Workload

Add `localRoute` to the Workload's config to allow for this component to be called directly. `localRoute` is also not reachable from outside of the host, including if someone was to add to a Host header request.`localRoute` is comma-separated and takes two forms:

| Entry | Serves |
| --- | --- |
| `svc.internal` | every path on that hostname |
| `svc.internal/hello` | `/hello` and below, on that hostname |

Example callee Workload:
```yaml
kind: WorkloadDeployment
metadata:
  name: callee
spec:
  replicas: 1
  template:
    spec:
      hostSelector:
        hostgroup: default # <<-- must be the same the same hostGroup that has localRoute enabled
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: functiona.example.com # <<-- What is reachable from the rest of the cluster via k8s
            localRoute: functiona.internal/hello # <<-- allow for local routing and for this domain/path
      components:
        - name: callee
          image: <your-registry>/callee:0.1.0
```

Example caller workload:

```yaml
kind: WorkloadDeployment
metadata:
  name: caller
spec:
  replicas: 1
  template:
    spec:
      hostSelector:
        hostgroup: default #<<-- Must be on same host
      hostInterfaces:
        - namespace: wasi
          package: http
          interfaces:
            - incoming-handler
          config:
            host: hello.example.com
      components:
        - name: caller
          image: <your-registry>/caller:0.1.0
          localResources:
            # Authority-level policy — no path, even though the route is
            # path-scoped. Enforced before local routing is consulted.
            allowedHosts:
              - functiona.internal #<<--still need to list the localRoute domain, since this is the first gate
            environment:
              config:
                CALLEE_URL: http://functiona.internal/hello #<<--example for doing a localRoute call with path
```

## What resolves where

Given the callee above:

| Caller requests | Result |
| --- | --- |
| `http://functiona.internal/hello` | in-memory → callee |
| `http://functiona.internal/hello/items` | in-memory → callee |
| `http://functiona.internal/` | network (no route claims `/`) |
| `http://functiona.internal/hello-world` | network (not a segment-boundary prefix) |
| `http://functiona.internal:9187/hello` | network (non-default port) |
| `http://functiona.example.com/hello` | network (published, not offered locally) |

And from outside the host, `Host: functiona.internal` gets a 404 — the name
exists only in the host's in-memory table.

### More forms

```yaml
# whole hostname, any path
localRoute: svc.internal

# several entries, mixing both forms
localRoute: svc.internal, api.internal/v1, api.internal/v2

# reachable both from the network and in-memory
host: svc.example.com
localRoute: svc.example.com
```

## Considerations before use

- Since Network Policies and other ingress configurations can dictate whether a component can be called, whether to use this feature should be thought through before enabling on a Workload.

- Since `localRoute` adds a "short circuit" to an outbound http request by matching a domain or domain + path, it is possible for a `localRoute` to intercept a call that the workload doesn't own. For example, a workload declaring `localRoute: api.stripe.com` intercepts a co-located workload's calls to Stripe.

- Though this is in-memory, there is no tls on a localRoute path
- There is no check whether multiple workloads have the same `localRoute` configured. This was intentional since scaling the same workload, or a migration/upgrade of a component could change names.

---

## Observability

Locally dispatched requests record `wasmcloud.http.route = "local"` on the outbound client span. 

---

## Testing

- `wash-runtime` lib suite: 547 passing.
- `integration_http_local_routing`: 9 tests, covering both `localRoute` forms, both halves of the two-key gate, the scope partition in both directions, segment-boundary matching, path scoping, and `allowedHosts` ordering.
- Unit coverage for the parser (URL-shaped, ported, bare-path and invalid-host entries all refused), longest-prefix-wins, unbind at both scopes, replica spread, case-insensitivity, the readiness pre-check, and quota exhaustion and recovery.
- One test pins the shipped example's exact config to real router behaviour, so the docs cannot drift from the code.

---

## Known gaps

Tracked, not addressed here:

- **No ownership validation** on a claimed hostname. This is the impersonation risk above; it is currently mitigated by documentation and by the feature being off by default. An allowlist, or restricting `localRoute` to names the
  workload also publishes, would close it. 
- **No loop guard.** `A → B → A` example recursion call does not deadlock, but it grows until something upstream times out.
- **P3 request-side errors** are not propagated on the local path: a g uest whose request body fails mid-stream is told the request side succeeded.
- example under `examples/local-ingress/` used for testing on a kind cluster with wasmCloud deployed w/changed


charts/runtime-operator/values.yaml
```diff
@@ -503,6 +503,31 @@ runtime:
       http:
         enabled: true
         port: 9191
+        # -- Enable same-host local routing: an outgoing HTTP request matching a
+        # co-located workload's `localRoute` config is served in-memory by that
```
vados-cosmonic, line 507:
```suggestion
        # co-located workload's `localRoute` config is served in-process by that
```

MEGA NIT: "in-process"? "in-host"?

charts/runtime-operator/values.yaml
```diff
@@ -503,6 +503,31 @@ runtime:
...
+        # `https://` request to a claimed name is handed over in plaintext with
+        # the certificate never checked, and `allowedHosts` does not help
+        # because the caller legitimately lists the name it means to reach.
+        # Enable only on a hostgroup whose workloads are equally trusted — not
+        # on a shared multi-tenant hostgroup.
+        #
+        # Locally routed calls also bypass anything on the network path (ingress
+        # auth, rate limits, mesh mTLS, NetworkPolicy). The caller's
+        # `allowedHosts` egress policy is still enforced first.
+        localRouting: false
```
vados-cosmonic, line 530: MEGA NIT: Should we maybe call this `insecureLocalRouting` or `localBypassRouting` or `insecureDirectRoute`? Is there any way to convey that you basically give up most of the security auth, ratelimits, etc by using this?

The thing is, `localRouting` definitely portrays that routing will happen locally, but at the *very* least it's local to the *host*, so it makes me think of `hostLocalRouting`. 

Maybe not worth the change here, to be honest, as long as we explain local routing really well everywhere.

crates/wash-runtime/src/host/trigger_service/http.rs
```diff
@@ -47,7 +47,7 @@ impl hyper::body::Body for ChannelBody {
 /// re-registers) a fresh instance. See `test_trigger_service_http_restarts_on_fault`.
 pub(crate) struct HttpTask {
     pub(crate) service: Arc<Service>,
-    pub(crate) req: hyper::Request<hyper::body::Incoming>,
+    pub(crate) req: hyper::Request<wasmtime_wasi_http::p2::body::HyperIncomingBody>,
```
vados-cosmonic, line 50: This makes me a bit nervous -- regular hyper body incoming is more general, I wonder why this is now scoped to P2 in particular (I would also assume we use this same HttpTask for P3?)

examples/local-ingress/README.md
```diff
@@ -0,0 +1,319 @@
...
+the callee **in-memory** — it never touches the network, kube-proxy, or any
+ingress in between.
+
+- [`caller/`](./caller/) — HTTP handler that makes an outgoing request to the
+  callee and reports the result. Target resolution: `?url=` query param →
+  `CALLEE_URL` env var → `http://callee.internal/hello`.
+- [`callee/`](./callee/) — HTTP handler that answers with a greeting echoing
+  the path and `Host` header it received.
+
+Same-host routing takes **two keys**, and neither works alone:
```
vados-cosmonic, line 15:
```suggestion
Same-host routing requires setting two pieces of configuration:
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -80,39 +80,227 @@ fn is_valid_hostname(host: &str) -> bool {
...
 }
 
-/// Collect the ingress hostnames a workload's HTTP handler serves on: the
-/// `host` config plus any comma-separated `host-aliases`, keeping only entries
-/// that are valid RFC 1123 hostnames.
+/// Normalize a configured `path` prefix into the form the routing table stores:
+/// no trailing slash, and a leading slash supplied if the author omitted one.
+/// An absent, empty, or `"/"` value normalizes to the empty prefix, which is a
+/// catch-all matching every path.
+fn normalize_path_prefix(path: &str) -> String {
```
vados-cosmonic, line 87: Wonder if it's worth adding a type to represent this...

like a 

```rust
struct NormalizedUrlPath(String)
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -80,39 +80,227 @@ fn is_valid_hostname(host: &str) -> bool {
...
+        Self::new(host, path_prefix, RouteScope::Local)
+    }
+
+    fn new(host: impl AsRef<str>, path_prefix: impl AsRef<str>, scope: RouteScope) -> Self {
+        Self {
+            // Hostnames are case-insensitive (RFC 1123) and the host serves one
+            // HTTP port, so neither case nor port carries routing information.
+            // Normalizing here means every lookup can normalize the same way and
+            // the two sides are guaranteed to agree.
+            host: split_host_port(host.as_ref()).0.to_ascii_lowercase(),
```
vados-cosmonic, line 171: can we use `url::Url` here? Obviously slower than just splitting, but might cover more edge cases

crates/wash-runtime/src/host/http.rs
```diff
@@ -80,39 +80,227 @@ fn is_valid_hostname(host: &str) -> bool {
...
+pub fn parse_local_route(entry: &str) -> Option<IngressRoute> {
+    let entry = entry.trim();
+    if entry.is_empty() {
+        return None;
+    }
+    // A URL is the likeliest thing to write here, and splitting one on `/`
+    // yields a scheme as the "hostname" (`http://svc/x` -> host `http`, path
+    // `/svc/x`) which is a perfectly valid-looking route that matches nothing.
+    // Refuse it outright rather than registering the misparse.
+    if entry.contains("://") {
```
vados-cosmonic, line 199: I think using `url::Url` would save us some headaches here, though it may be slower

crates/wash-runtime/src/host/http.rs
```diff
@@ -336,59 +613,66 @@ impl Router for DynamicRouter {
...
+            if let Some(routes) = table.workload_to_host.remove(workload_id) {
+                for route in &routes {
+                    let Some(buckets) = table.host_to_workload.get_mut(&route.host) else {
+                        continue;
+                    };
+                    if let Some(bucket) = buckets
+                        .iter_mut()
+                        .find(|b| b.prefix == route.path_prefix && b.scope == route.scope)
+                    {
+                        bucket.workloads.remove(workload_id);
```
vados-cosmonic, line 664: Is it possible for a workload to be unbound but still running and then try to use this lookup?

crates/wash-runtime/src/host/http.rs
```diff
@@ -233,75 +446,139 @@ pub struct DynamicRouter {
...
+                        buckets.push(PathBucket {
+                            prefix: route.path_prefix.clone(),
+                            scope: route.scope,
+                            workloads: BTreeSet::from([workload_id.to_string()]),
+                        });
+                        // Longest prefix first, so the most specific route that
+                        // matches is the first one found. The catch-all (empty
+                        // prefix) sorts last, which is what makes it a fallback
+                        // rather than a shadow over every scoped route.
+                        buckets.sort_by(|a, b| {
```
vados-cosmonic, line 508: Should we make this the ord impl for `PathBucket` ? If we did that would allow a `BtreeSet` of them to stay sorted right?

vados-cosmonic (commented):
Haven't had LLMs take a look yet (I assume you have), but did a quick review, nothing but nits from me.

vados-cosmonic:
OK, so also I gave :robot:  a pass at it, and it found more things that I didn't:

1. [P1, availability] There is no local-hop/recursion guard
2. [P1, correctness] Replica selection happens before readiness, so one unready claimant causes intermittent network fallback; unbind also has a TOCTOU failure
3. [P1, P3 correctness/resource control] Local P3 drops request-side errors and ignores `between_bytes_timeout`
4. [P2, correctness/security] Local dispatch accepts non-HTTP schemes, and malformed route declarations fail open to network egress
5. [P2, DX/testing] The shipped Kind example cannot enable this branch's feature as written and its image instructions contradict the manifest
  (Kind example might have to be another commit after we publish stuff? But also, I also think our tests shouldn't pass if the kind example is broken?)
6. [P2, test coverage] All new runtime integration coverage is P2-only
 (Yeah we really should have a P3 test in here) 

I think (4) might be partially fixed by using `url::Url` more... at the very least it can more easily validate the scheme.

Either way I think these should be pretty easy to throw an :robot: at and replicate the issue, but I can share more if necessary!

charts/runtime-operator/values.yaml
jfleitz: Changing to `localBypassRouting`, as I feel that implies everything skipped and not just security related (like rate limiting that may be in place)

crates/wash-runtime/src/host/http.rs
jfleitz: Changed to NormalizedPathPrefix struct

crates/wash-runtime/src/host/http.rs
jfleitz: The local routes are removed before the handles, so it has less of a chance versus the typical incoming request (how wasmCloud works today). I feel this is ok like it is to match with how the unbind logic works with the incoming routes, but I am up for a separate PR to keep the change consistent for both. It is a small window, so maybe that is why we haven't had any reported issues.

crates/wash-runtime/src/host/trigger_service/http.rs
jfleitz: This is a great callout. Fixing this so that it is consistent with caller to callee.

PR #5402, opened 2026-07-28
refactor(runtime): one list for multiplexed plugins

ricochet:
`wash host`, `wash dev`, and downstream embedders each wrote out the same multiplexed plugin registrations by hand. The list is load-bearing, not decorative: every single-backend plugin leaves `HostPlugin::supports_named_instances()` at false, so without it a workload declaring two entries of one namespace:package under distinct names fails to bind and `(implements ..)` does not work at all.

Build the set once in `plugin::multiplexed_plugins()` and register it through `with_multiplexed_plugins()` on both builders, so adding a multiplexed plugin reaches every embedder at once. Postgres stays at the call sites: whether it registers, and whether as `multiplex_only`, depends on host configuration.

### DIFF
--- a/crates/wash-runtime/src/host/mod.rs
+++ b/crates/wash-runtime/src/host/mod.rs
@@ -984,6 +984,19 @@ impl HostBuilder {
         Ok(self)
     }
 
+    /// Registers the multiplexed plugin set from
+    /// [`crate::plugin::multiplexed_plugins`], which is what makes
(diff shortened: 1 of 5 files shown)


vados-cosmonic (approved):

vados-cosmonic:
I think just about any change to multiplexed plugins will mean a breaking change (so maybe it shouldn't necessarily be upstream), but the improvement in DX is almost certainly worth it.

PR #5404, opened 2026-07-28
feat: workload ip name lookup

ricochet:
Introduce new capability: allowIpNameLookup list on LocalResources

This change scopes wasi:sockets/ip-name-lookup to a name allowlist for each workload. It is enforced where the sockets ctx is built from the component's ctx template, so the p2 stream resolver and the p3 async resolver both use it.

An undeclared or empty list denies every lookup with permanent-resolver-failure. Entries accept `*` and `*.suffix` wildcard, an exact name, or a literal IP address. Matching per name rather than per interface keeps resolution from being a channel for carrying data off the host in attacker-chosen labels, which a single on/off switch would leave open.


crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -250,6 +250,7 @@ async fn build_ctx_from_template(
             })
         }),
         loopback: Arc::clone(&template.loopback),
+        allowed_names: Arc::clone(&template.local_resources.allow_ip_name_lookup),
```
```suggestion
        allowed_ip_names: Arc::clone(&template.local_resources.allow_ip_name_lookup),
```

crates/wash-runtime/src/host/allowed_names.rs
```diff
@@ -0,0 +1,395 @@
...
+/// # Errors
+///
+/// Parsing via [`FromStr`] returns an error when the input:
+///
+/// - is empty (after trimming),
+/// - is a wildcard not of the form `*.<rest>`,
+/// - carries a scheme, port, path, query, or fragment,
+/// - is not a syntactically valid domain name or IP address.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AllowedName {
```
vados-cosmonic: probably should be non exhaustive, one example we might need in the future is subnets

crates/wash-runtime/tests/fixtures/http-ip-name-lookup/src/lib.rs
vados-cosmonic, line 1: This could technically be combined into one component with the other one right? With one endpoint for p2/p3?

runtime-operator/internal/controller/runtime/workload_controller.go
```diff
@@ -215,6 +215,7 @@ func materializeLocalResources(ctx context.Context, c client.Client, namespace s
 	}
 
 	lr.AllowedHosts = spec.AllowedHosts
+	lr.AllowIpNameLookup = spec.AllowIPNameLookup
```
vados-cosmonic: How do you feel about `AllowedIpNameLookups` here? Sounds like a boolean here but is actually multiple things

vados-cosmonic (dismissed):
LGTM 🚀 

Some minor nits around naming

runtime-operator/internal/controller/runtime/workload_controller.go
ricochet: Like it and all the other renames. Working through them now

crates/wash-runtime/tests/fixtures/http-ip-name-lookup/src/lib.rs
ricochet: Not today. We short-circuit the http handler and don't support two different http backends at once.

jfleitz (dismissed):

jfleitz (approved):

PR #5410, opened 2026-07-29
Host component imports for wasi:config and http

jfleitz:
## Description

Host component plugins do not have the same level of trust as native host plugins due to the native plugin runs as host code vs the host component running in its own wasmtime store. This is designed since it supports the component model approach.

### wash-runtime
This change adds supports for allowing a host component to import configuration settings through the `wasi:config` interface `wasi_config` host plugin, as well as providing egress support by passing the host's http handler.

#### on the wasi_config change:
When a wasm Workload uses the `wasi_config` host interface, the config values are all put together in `on_workload_item_bind`.  There isn't a current setup of a **plugin using another plugin** (i.e. `on_component_plugin_bind` ?), so this approach needs to be reviewed carefully, as this does give the ability for a host component to read any environment variable through the wasi_config host plugin.

### wash
- **fix(wash): union interface names when merging same-package WIT imports into host_interfaces**

### DIFF
--- a/crates/wash-runtime/src/engine/mod.rs
+++ b/crates/wash-runtime/src/engine/mod.rs
@@ -998,6 +998,17 @@ pub fn imports_wasi_http(component: &Component) -> bool {
         .any(|(import, _item)| import.starts_with("wasi:http"))
 }
 
+/// Whether a component imports any `wasi:config` interface. Used to decide
+/// whether a host component plugin's linker needs the `wasi:config/store`
+/// host functions.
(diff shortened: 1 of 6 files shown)


ricochet:
To support this well, we want to make sure we can set policy configuration on the host components, have a secure way to pass secrets to the host component in addition to config, and bind all runtime native plugins. We should still aim to avoid potential recursive bombs like host components depending on other host components. Bringing all of that into scope supersedes this PR (but good start) with #5411

PR #5411, opened 2026-07-30
host component plugins with native bindings

ricochet:
Host component plugins can now import native (non-component) host capabilities (`config`, `secrets`, `keyvalue`, etc) resolved at load time against the host's registered natives, delivered their own bind-time config through the same `on-workload-bind` path a real workload's `secretFrom`-sourced config goes through.

- **Native-import resolution** (`ComponentHostPlugin::new`): a plugin's imports resolve against the host's native plugin registry; its own `allowed_hosts`/`allowed_ip_name_lookups` gate its own `wasi:http`/DNS egress the same way a workload's `LocalResources` do. Reserved `wasmcloud:host` exports are matched by interface name, not the exact versioned string, so a patch bump elsewhere in that package can't break an already-built plugin. There's more layers of changes I'd like to do for this, but saving that for a future versioning handling refactor.
- **`wash host` config sourcing**: adds a `dir:` source kind for config/secrets which is the shape a Kubernetes ConfigMap/Secret volume mount projects (filename=key, file content=value) alongside the existing `file:` single-blob source, plus `HostConfig`/per-plugin `config`/`configFrom`/`secretFrom` wiring through `wash dev` and the runtime-operator chart.
- **Labeled `wasmcloud:secrets/secret` imports**: a new nullary-`get` interface so a component imports one labeled instance per secret it needs (`import db-password: wasmcloud:secrets/secret;`), and the host resolves each label against the plugin's own bind-time config synchronously, from the component's import type, *before* it's ever instantiated. A missing label fails construction immediately, naming it (unlike the existing dynamic `store`/`reveal` path) where a missing key is only discovered when `store.get` returns `not-found`.
- **`http-egress-plugin` fixture**: a host component plugin exercising its own `wasi:http/outgoing-handler` + `wasi:sockets/ip-name-lookup` egress policy end to end.
- **WIT publish-drift fix**: `wasmcloud:host` (`workload-lifecycle`, added after the published `0.1.0` tag, never pushed) and `wasmcloud:keyvalue` (the multiplex refactor moved `bucket` from `store` into a new `types` interface, also never pushed) had both drifted past their last published versions with nothing catching it, because `wit.yml`'s publish step treats a stable version as immutable and silently skips it if the tag already exists. Bumps `host` to `0.1.1` and `keyvalue` to `0.2.0` (the latter is a breaking shape change, hence a minor bump), and adds a validate-time drift guard that pulls the published wasm for any stable version already at the destination and fails the job if its WIT differs from what was just built. Verified against the real registry (correctly flags the pre-fix `wasmcloud:host@0.1.0`, reports no drift for `wasmcloud:messaging@0.2.0`).
- **CI**: adds a reusable `oci-push-with-digest` action and a `component-matrix` workflow.

## Test plan

- [x] new coverage for labeled-secret resolution, missing-label construction failure, and http-egress fixture behavior
- [x] `check-wit-drift.mjs` manually verified against the live registry (see commit message for `fix(wit): ...`)

This is probably easiest reviewed commit by commit. I broke it down into the logical units. I'm game to separate this into other PR's if that's preferred, but the biggest parts of this are tests.


jfleitz (dismissed):
This LGTM. This is a lot of changes, so I did pull a copy down and ran locally to test with a custom Host Component Secret plugin and was able to use it effectively.

jfleitz (approved):

PR #5413, opened 2026-07-30
feat(messaging): async wasmcloud:messaging@0.3.0

### LINKED ISSUE #5341: async version of wasmcloud:messaging
Like `wasmcloud:blobstore` and `wasmcloud:keyvalue` asyncify `wasmcloud:messaging`.

Aditya1404Sal:
Closes #5341

Makes `wasmcloud:messaging` async, bringing it in line with blobstore, keyvalue and postgres.

`publish`, `request` and the guest-side `handle-message` are now `async func`s, so a component can keep several messages in flight instead of stalling its whole instance on every request/reply.

This ships as **`@0.3.0`** -- sync to async is a breaking ABI change, so `@0.2.0` is left completely untouched and both are served side by side. A workload picks a surface with the version on its `wasmcloud:messaging` host-interface entry; an entry with no version still resolves to `@0.2.0`. Nothing that exists today has to change.

Both directions are covered: the multiplexed plugin binds the async `consumer` import (labeled and unlabeled), and the NATS/in-memory plugins plus the trigger service deliver to an async `handle-message` export.

Verified with a real guest fixture that awaits a reply from inside its async handler -- round trip, concurrent delivery, and a `@0.2.0` guest alongside a `@0.3.0` one on the same host. The operator's kind-based e2e now runs its NATS round trip against both revisions.


crates/wash-runtime/src/host/trigger_service/mod.rs
```diff
@@ -103,15 +103,23 @@ impl Ingress {
...
-                // dynamically (there is no accessor-driven p3 messaging binding).
-                let iface = instance
-                    .get_export(&mut *store, None, MESSAGING_HANDLER)
-                    .with_context(|| format!("service is missing {MESSAGING_HANDLER} export"))?
-                    .1;
+                // Look up the `handle-message` export up front; it's invoked
+                // dynamically, which serves either handler revision. Prefer the
+                // async `@0.3.0` interface and fall back to `@0.2.0`, so a
+                // service built before the async revision still resolves.
+                let (handler, iface) = MESSAGING_HANDLERS
```
vados-cosmonic: It's possible for a component to export *both* messaging handlers at the same time, no?

We might want to show some sort of warning here if we're choosing one or the other and ignoring the other one. Alternatively, there's the option to keep both and call whichever handler is available, or one that matches the trigger service's interface.

There's a bit more subtlety to think through here/make an explicit decision on.

Also, a test for this case would be good, as a regression test of what the behavior should be.

wit/messaging/wit/types.wit
```diff
@@ -0,0 +1,41 @@
...
+/// Types common to message broker interactions.
+///
+/// Unlike `wasmcloud:messaging@0.2.0`, every broker operation in this package is an `async func`.
+/// A single component can keep many publishes and requests in flight at once, so an inbound HTTP
+/// request no longer stalls the whole instance while a `request` waits on its reply.
+interface types {
+  /// A message sent to or received from a broker
+  record broker-message {
+    subject: string,
+    body: list<u8>,
```
vados-cosmonic: Do you think we could benefit from making the body a `stream<u8>` here?

wit/messaging/wit/handler.wit
```diff
@@ -0,0 +1,13 @@
...
+/// The consuming side of a broker: implemented by components that subscribe.
+interface handler {
+  use types.{broker-message, error};
+
+  /// Callback invoked when a message is received from a subscription.
+  ///
+  /// As an `async func` this may await other operations — including
+  /// `consumer.publish` to reply on `msg.reply-to` — while the host continues to
+  /// deliver further messages to the same instance concurrently.
+  handle-message: async func(msg: broker-message) -> result<_, error>;
```
vados-cosmonic: This might be a good time to break apart the kinds of errors that can occur -- a `handle-message-error` might not be a bad thing to have, with errors that only apply to handling messages (rather than everything in one)

wit/messaging/wit/consumer.wit
```diff
@@ -0,0 +1,20 @@
...
+  /// outstanding concurrently; awaiting one does not block the instance.
+  ///
+  /// Returns `error::timeout` if no reply arrives within `timeout-ms`.
+  request: async func(subject: string, body: list<u8>, timeout-ms: u32) -> result<broker-message, error>;
+
+  /// Publish a message to a subject without awaiting a response.
+  ///
+  /// Resolving indicates the broker accepted the message, not that any subscriber
+  /// received it.
+  publish: async func(msg: broker-message) -> result<_, error>;
```
vados-cosmonic, line 26: Same note here about `error` being overly broad

wit/messaging/wit/consumer.wit
```diff
@@ -0,0 +1,20 @@
...
+interface consumer {
+  use types.{broker-message, error};
+
+  /// Perform a request operation on a subject, resolving once the reply arrives.
+  ///
+  /// Because this is an `async func`, a component may have many requests
+  /// outstanding concurrently; awaiting one does not block the instance.
+  ///
+  /// Returns `error::timeout` if no reply arrives within `timeout-ms`.
+  request: async func(subject: string, body: list<u8>, timeout-ms: u32) -> result<broker-message, error>;
```
vados-cosmonic: What do you think about having `body` be a `stream<u8>` here? We don't have it yet, but we will eventually get a `stream.forward` intrinsic that will make it easier to forward streams, but removing the requirement of holding the body in it's totality in memory would be beneficial I think.

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -114,6 +154,110 @@ async fn route_to_subscribers(
...
+    let pending_requests = {
+        let lock = plugin.tracker.read().await;
+        match lock.get_workload_data(workload_id) {
+            Some(data) => data.pending_requests.clone(),
+            None => wasmtime::bail!("workload state not found"),
+        }
+    };
+
+    // Generate a unique reply-to subject
+    let reply_to = format!("_INBOX.{}", uuid::Uuid::new_v4());
```
vados-cosmonic: What do you think about a v7 here to include the time axis -- we lose some randomness but the chances of collision in the same millisecond are still astronomically low

vados-cosmonic (commented):
Thanks for the work here @Aditya1404Sal this looks *almost* perfect!

I just had a few comments on the new interface and an edge case I think we *might* have some day, let me know if anything doesn't make sense.

vados-cosmonic:
Hey also @ricochet you briefly discussed trigger services being only p3? Does that mean we shouldn't actually bother with supporting messaging 0.2.0 for the messaging trigger service?

I think I may have misheard but that might affect this PR too if I didn't mishear

wit/messaging/wit/consumer.wit
Aditya1404Sal: Sounds reasonable, changed it!

wit/messaging/wit/consumer.wit
Aditya1404Sal: Thanks for pointing this out & You're right, So i ended up splitting the handler side -- there's a handle-message-error now (reject, retry, other), since a handler can't really raise broker errors about a message it's already holding, all it can pretty much say is whether redelievery would help

But here for publish / request though I left the shared error on purpose. When I went through the cases, almost all of them apply to both -- publish hits : subject-invalid, message-too-large, broker-unavailable the same way request does ( sinc NATS classifier actually produces those for publishes ). The only request-specific one is the timeout, and I've scoped its docs to request. 

Plus based on your suggestion of making body a stream, publish might legitimately want a timeout too -- like waiting on a hypothetical caller to finish producing the body (so personally I don't think this slight deviation is worth stripping it out into distinct types)

but if you would rather prefer seperate publish-error/request-error, I understand and I'd be happy to do it.

crates/wash-runtime/src/host/trigger_service/mod.rs
Aditya1404Sal: keeping the service ingress p3-only, per-message components still support p2 -- dual exports now warn that the @0.2.0 handler is ignored, and there's a regression test (messaging-dual-handler fixture) pinning that @0.3.0 is the one invoked

Aditya1404Sal:
Hey @vados-cosmonic -- Factored in the changes you requested, also left a comment around the broad error being alright for request/publish. Let me know if there's any more changes you think are necessary.

vados-cosmonic:
Hey @Aditya1404Sal thank you! Would you mind resolving all the conversations that you addressed before I do a second pass? I think I'm basically going to need to do another fresh pass since it's been a little bit and I don't have all this code "paged in" so to speak!

Aditya1404Sal:
> Hey @Aditya1404Sal thank you! Would you mind resolving all the conversations that you addressed before I do a second pass? I think I'm basically going to need to do another fresh pass since it's been a little bit and I don't have all this code "paged in" so to speak!

Apologies for not doing that before, resolved all but one ;)

wit/messaging/wit/consumer.wit
vados-cosmonic: OK, that's reasonable -- I generally like having errors be separate per operation so that we know exactly what operation we're dealing with all the time *but* as you said the cases are actually identical right now, I think that makes a lot of sense.

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
```diff
@@ -146,21 +186,133 @@ impl<'a> Host for ActiveCtx<'a> {
...
+        body: wasmtime::component::StreamReader<u8>,
+        timeout_ms: u32,
+    ) -> wasmtime::Result<Result<AsyncBrokerMessage, AsyncMsgError>> {
+        let plugin =
+            accessor.with(|mut a| a.get().try_get_plugin::<NatsMessaging>(PLUGIN_MESSAGING_ID))?;
+
+        // Drain the body first (see `collect_body`); `timeout-ms` starts only
+        // once the body is fully consumed — the caller controls how fast it
+        // arrives, so that wait is not counted against the broker.
+        let body = match super::collect_body(accessor, body).await? {
```
vados-cosmonic, line 212: So this is interesting -- if we collect the whole body here, I'm not sure there's much point in streaming since the plugin still has to hold the whole body in memory. 

I'm not sure reading slowly being counted against the broker is worth having to hold the whole body in memory if we can avoid it.

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
```diff
@@ -146,21 +186,133 @@ impl<'a> Host for ActiveCtx<'a> {
...
+/// `ActiveCtx`. Errors are classified into [`MsgError`] and lowered into the WIT
+/// `error` variant. Note this differs from the sync `publish` above, which
+/// *traps* the guest on a publish failure; the async surface reports it as an
+/// ordinary `result` error, which is what the WIT says it is.
+impl<T: 'static + Send> HostWithStoreP3<T> for SharedCtx {
+    async fn request(
+        accessor: &Accessor<T, Self>,
+        subject: String,
+        body: wasmtime::component::StreamReader<u8>,
+        timeout_ms: u32,
```
vados-cosmonic: Can the timeout be optional?

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
```diff
@@ -146,21 +186,133 @@ impl<'a> Host for ActiveCtx<'a> {
...
+    ) -> wasmtime::Result<Result<(), AsyncMsgError>> {
+        let plugin =
+            accessor.with(|mut a| a.get().try_get_plugin::<NatsMessaging>(PLUGIN_MESSAGING_ID))?;
+
+        let AsyncBrokerMessage {
+            subject,
+            body,
+            reply_to,
+        } = msg;
+        let body = match super::collect_body(accessor, body).await? {
```
vados-cosmonic, line 257: here we don't have a timeout but we still collect the whole body -- We should be able to pass the stream through, no?

crates/wash-runtime/tests/fixtures/messaging-dual-handler/wit/deps/wasi-clocks-0.3.0/package.wit
vados-cosmonic, line 1: So, I might be wrong here, but IIRC we avoid checking in deps now in favor of `wkg.toml` overrides? An example would be [`stream-consumer-p3`](https://github.com/wasmCloud/wasmCloud/blob/main/crates/wash-runtime/tests/fixtures/stream-consumer-p3/wkg.toml) -- could we do that here?

I saw the note that we can't do this for wasmcloud:messaging (which is also a bit weird.... we should be able to override a specific version via `wkg`??), but is this also the case for wasi:clocks and stuff?

vados-cosmonic (commented):
Thanks for the fixes, this looks better -- a note or two about the streaming/whether we should be collecting and WIT inclusion

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
Aditya1404Sal: I may be wrong here, but I did go looking for a way to stream instead of collect and hit a dead end and it's not even a choice we get to make honestly, core NATS just has no streaming shape to it. The PUB frame carries the payload length before the payload, and the client api is `Bytes` (No async stream reader in it's traits) both of the directions even an inbound Message.payload is a fully assembled buffer, so there's nothing to hand increments to. 
I also checked the latest async-nats (0.50) in case something had landed; but still Bytes everywhere. The only streaming thing in NATS is the obvious jetstream object store, but I don't think that's what you meant, but nats-wise it's kinda the last way to do message streaming for the object blob payload.

So the question is whether we keep `stream<u8>` for a future backend that can actually stream bodies, or downshift to `list<u8>` (although both still default to being lifted in-memory and frankly won't lighten the contract)

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
Aditya1404Sal: Same as the request path, the timeout is not the reason we collect. This is being passed at `client.publish(subject, payload)`, and that wants the whole payload as one `Bytes`. There's no streaming publish in the client, therefore nothing on the other side to pass a stream to.

crates/wash-runtime/tests/fixtures/messaging-dual-handler/wit/deps/wasi-clocks-0.3.0/package.wit
Aditya1404Sal: Fair, and I tried a fair few ways to get there, wkg just can't express two versions of one package. Versioned override keys are a parse error in wkg.toml, wit.sources accepts them but discards the version so they collapse onto one entry,

`wash wit fetch` wipes `wit/deps` and rebuilds from overrides + registry rather than merging, so committing just the messaging deps and letting `wasi fetch` doesn't survive either. That's why the wasi deps (clocks and stuff) ride along.

vados-cosmonic: Yeah, this is unfortunate -- :( Thanks for looking into this @Aditya1404Sal. 

I wonder if this is actually worth filing against the repo upstream on `wasm-pkg-tools`.

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
vados-cosmonic: > So the question is whether we keep stream<u8> for a future backend that can actually stream bodies, or downshift to list<u8> (although both still default to being lifted in-memory and frankly won't lighten the contract)

Yeah so definitely keep the stream interface and then when we have something that doesn't have to hold stuff in memory! Thanks for looking into this, it's a bit unfortunate that `async-nats` doesn't have this functionality.

vados-cosmonic (approved):
LGTM 🚀 

Thanks @Aditya1404Sal ! If you wouldn't mind clearing up some of the test failures, we can get this merged. 

I think we also need a Go code maintainer so @ricochet / @jfleitz if you wouldn't mind taking a look

xtask/src/main.rs
```diff
@@ -96,7 +96,15 @@ const P2_FIXTURES: &[&str] = &[
     "postgres-implements",
 ];
 
+/// Fixtures whose `wit/deps` is committed rather than fetched, built with
```
ricochet, line 99: for whatever reason, llm's love trying to add this idiom back. For new versions, have wkg.toml refer to them with relative paths. For defining it's own wit, again local paths work. 

I'm tempted to have a comment here that says exactly this (it's happened to me like 5 times now) :P

ricochet (approved):

PR #5414, opened 2026-07-31
preserve component id

jfleitz:
Two issues when testing with a custom secret host plugin:
- Preserves the component ID vs having a random UUID so that at bind time the plugin's config can be loaded.
- The dedupe logic kept only the first entry per namespace:package and dropped the rest, so a component importing e.g. both wasmcloud:secrets/store and wasmcloud:secrets/reveal lost whichever instance iterated second. Same-package instances now merge their interface names into the existing entry, preserving any config/name/version from the first declaration


ricochet:
I actually just folded these into the current PR (plus additional test coverage). Thank you for trying it out and the report!

PR #5420, opened 2026-08-03
test(wash): add e2e k8s host test

vados-cosmonic:
This commit adds an E2E test that ensures from the Rust perspective that the E2E flow of running in k8s (installing with Helm, etc), works. 

Right now the test isn't run by default but that's something we can add (maybe to release PRs) -- right now it runs locally and is relatively easy to test out (see the updated `CONTRIBUTING.md`), though the build can be long if the host has to be built from scratch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,28 @@ cargo test --lib
 cargo test --bin wash
 ```
 
+The Rust Kubernetes host smoke test is ignored by default because it builds a
+host image and creates a complete kind cluster, and generally takes a long time
+to run.
+
+The smoke test requires:
+
+* A `docker`-compatible daemon
+* `helm`
+* `kubectl`
+
+During the test we download the pinned kind/Kubernetes and chart images to create and use
+a temporary kubeconfig in a `testcontainers`-managed, DinD `kind` cluster.
+
+To run the smoke test:
+
+```bash
+# (optional) if the host image is already built with tag 'localhost/wasmcloud-wash:rust-k8s-e2e', you can skip that build
+# export TEST_WASH_INT_K8S_SKIP_HOST_IMAGE_BUILD=true
+
+cargo test -p wash --test k8s_host_serves_http_workload -- --include-ignored --nocapture
+```
+
 ## Project Structure
 
 This repository is a workspace. The `wash` CLI lives in `crates/wash` and the core Wasm runtime it depends on lives in `crates/wash-runtime`:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6092,6 +6092,7 @@ dependencies = [
  "toml 0.8.23",
  "tonic-prost-build",
  "tracing",
+ "tracing-subscriber",
  "url",
  "uuid",
  "wash-runtime",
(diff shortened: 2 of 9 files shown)


ricochet:
I'm confused and maybe I'm missing something? All of the test cases are already covered in kind and integrated into CI: https://github.com/wasmCloud/wasmCloud/blob/main/runtime-operator/test/e2e/e2e_test.go

Go is a much better fit for automating kind, the go operator, and working with the k8s SDK in general. You can argue against ginkgo and gomega (I didn't pick them), but it's all there and already integrated into CI. So if the goal is to rustify everything, I'm not a fan for places where the ecosystem is already largely go. There's also random problems I see here like the cluster leaks on panic.

```text
┌──────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                          Rust test step                          │                                               Already in Go e2e                                                │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ kind cluster w/ OwnerReferencesPermissionEnforcement + 30950     │ deploy/kind/kind-config.yaml (via make test-e2e)                                                               │
│ mapping                                                          │                                                                                                                │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ build host image from local tree, kind load                      │ e2e_suite_test.go BeforeSuite, BUILD_RUNTIME_IMAGE=true (set by wash.yml:557)                                  │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ helm upgrade --install the same chart, same hostGroup --sets     │ BeforeSuite → buildBaseHelmSets()                                                                              │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ wait for a Host CR                                               │ Context("Host Registration")                                                                                   │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ apply config/samples/service_deployment.yaml, wait Ready         │ Context("Workload w/Service Lifecycle") → "should deploy a workload and become ready"                          │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ HTTP 200 on Host: hello.localhost.direct                         │ Context("Workload Lifecycle") → "should serve HTTP traffic through the gateway"                                │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ kubectl/helm diagnostics on failure                              │ e2e_test.go + utils                                                                                            │
├──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ —                                                                │ plus EndpointSlice, RBAC/is forbidden, finalizers, tenant namespace, scoped install, host plugins, implements, │
│                                                                  │  messaging                                                                                                     │
└──────────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

vados-cosmonic:
Ah, so this was an old TODO I had -- I didn't think we had this test on the Go side but thanks for the pointer, clearly we do. 

For reference, I don't have any problems with Gomega/Ginkgo (great set of tools for Go), or any desire to rustify everything, simply thought we didn't have this test!

PR #5422, opened 2026-08-03
feat(wash-runtime): configurable CA trust roots for outbound HTTPS

jfleitz:
## Description
Outbound wasi:http requests previously verified TLS against wasmtime's compiled-in
webpki roots only, so components could not reach hosts behind a corporate or
otherwise private CA.

- Replaces the default egress transport (P2, P3, and the gRPC fast path) with one
  whose trust roots are configurable: a `TrustRoots` base plus extra PEM CA
  bundles layered on top.
- **The default is `webpki` — identical trust behavior to before this PR.** An
  unconfigured host still verifies against exactly the compiled-in webpki
  (Mozilla) roots, so there is no trust-boundary change for existing
  deployments; everything below is opt-in.

Configuration (`wash host` flags / env, or `.wash/config.yaml` for `wash dev`):

- `--http-client-ca-path` / `WASH_HTTP_CLIENT_CA_PATHS` / `dev.http_client_ca_paths`
  — extra PEM CA bundle files trusted in addition to the base roots
  (comma-separated and/or repeated).
- `--http-client-trust-roots` / `WASH_HTTP_CLIENT_TRUST_ROOTS` /
  `dev.http_client_trust_roots` — the built-in base:
  - `webpki` (default): compiled-in Mozilla roots only — reproducible, ignores
    the host environment.
  - `webpki-and-native`: also trust the platform store, which honours
    `SSL_CERT_FILE`/`SSL_CERT_DIR` — lets operators inject CAs without
    wash-specific config.
  - `native`: platform store only.
  - `extra-only`: trust exactly the configured bundles and nothing else — the
    corporate-CA case of pinning a single private root (public CAs can no
    longer MITM traffic to the pinned backend).

The gRPC egress path shares the same roots via `OutgoingHandler::client_tls_config`;
the h2/ALPN variant is derived once on the first gRPC request. gRPC egress error
codes now follow upstream wasmtime's mappings (invalid server name →
`DnsError("invalid dns name")`; DNS failures and connect timeouts are
distinguished from `ConnectionRefused`).

Also fixes a latent P3 transport bug: the request-error future returned to
wasmtime resolved `Ok(())` on its first poll, so wasmtime aborted the hyper
connection driver as soon as the response head was handed to the guest — any
response whose body arrived after the headers failed with
`HttpProtocolError` — and mid-body connection errors were only logged. The
connection driver is now returned as the request-error future (staying pending
until the connection finishes and propagating its outcome to the guest), on
both the wasi:http path and the gRPC P3 path.

Design note: trust roots are host-global — every component on a host shares
them. Per-workload trust scoping is out of scope here and a possible follow-up.

For helm installs of the wash host, the flags can be passed as `extraArgs` in
the hostGroup:

```yaml
runtime:
  hostGroups:
    - name: default
      replicas: 1
      http:
        enabled: true
        port: 9191
      volumes:
        - name: corporate-ca
          configMap:
            name: corporate-ca
      volumeMounts:
        - name: corporate-ca
          mountPath: /etc/wash/ca
          readOnly: true
      extraArgs:
        - "--http-client-ca-path=/etc/wash/ca/ca.pem"
        - "--http-client-trust-roots=extra-only"
```

### New automated coverage in wash-runtime:

- P2 and P3 egress against a private-CA HTTPS server: fails with TlsProtocolError on default roots, succeeds with the CA configured (extra-only).
- gRPC P2 and P3 paths pick up the configured roots (h2/ALPN server behind a private CA); the P3 test also asserts the request-error future polls pending so the connection outlives the response head.
- P3 io-future probes: delayed-body response completes (the future stays pending under wasmtime's noop-waker poll), and a truncated body surfaces as HttpProtocolError rather than silent success.
- Trust-store construction: empty stores rejected (extra-only with no bundles), bad/empty PEM bundles rejected, and a guard that the default stays webpki-only.

### Results
With the CA configured, the caller reported outbound request ok: status 200 / upstream says: hello over TLS from the private-CA backend; without it, a 502 error sending request; TLS protocol error.


crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+
+    // The connection worker is moved into this future so it stays alive until
+    // the guest is done with the request, matching wasmtime's default
+    // transport. Body errors reach the guest through the response body stream.
+    let io: P3RequestErrorFuture = Box::new(async move {
+        let _worker = _worker;
+        Ok(())
+    });
+    Ok((resp, io))
+}
```
ricochet: P3 responses break whenever the body doesn't arrive with the headers. wasmtime_wasi::runtime::spawn returns an AbortOnDropJoinHandle, and wasmtime's P3 host handler polls this io future once with a noop waker. This future is Ready on the first poll, so the hyper connection driver is aborted the instant the response head is handed to the guest. Upstream avoids this by returning the conn future itself as io (so it's Pending), never spawning it.

This is the minimal fix:
```suggestion
    let io: P3RequestErrorFuture = Box::new(async move {
        _worker.await;
        Ok(())
    });
    Ok((resp, io))
}
```

But a much better approach would be to mirror upstream and propagate the connection error instead of Ok(()). Right now a mid-body connection failure is only warn!-logged and the guest's request-error future resolves Ok. When we're modifying but building from upstream, I try to keep a comment about it so that when I do wasmtime upgrades, it's something to key off of to validate any future fixes made to upstream.

crates/wash-runtime/src/host/http.rs
ricochet: Note same latent bug as http_client

crates/wash-runtime/src/host/http_client.rs
ricochet: I confirmed it with a probe test against a server that sends the head, waits 300ms, then sends the body:

io future ready-on-first-poll = true
panicked ... body read failed: ErrorCode::HttpProtocolError

Please land the probe as a real test.

ricochet: Also noticed there's no coverage for P3 egress with certs. Also missing a test that the gRPC path picks up the configured roots

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+
+/// Process-wide default outbound TLS configuration (webpki + native roots),
+/// built once on first use.
+pub fn default_client_tls_config() -> Arc<rustls::ClientConfig> {
+    static DEFAULT: OnceLock<Arc<rustls::ClientConfig>> = OnceLock::new();
+    DEFAULT
+        .get_or_init(|| {
+            // With no extra_ca_paths this cannot fail, but fall back to
+            // webpki-only roots rather than panicking if it ever does.
+            ClientTlsOptions::default().build().unwrap_or_else(|err| {
```
ricochet, line 139: Sets disable_native_certs: false, so every wash host now trusts the OS trust store for guest egress, not just those passing `--http-client-ca-path`.  It's a change in the trust boundary for existing deployments.

We could expose an opt-out (`--no-native-ca` or `dev.http_client_use_native_certs: false`). I'm thinking right now that's not necessary and something we can add later, but worth calling out that this is a change in behavior in the release notes.

crates/wash/src/cli/dev.rs
```diff
@@ -161,8 +161,31 @@ impl CliCommand for DevCommand {
         );
 
         let http_handler = wash_runtime::host::http::DevRouter::default();
+
+        // Outbound (egress) trust roots: extra CAs for the component's outgoing
+        // HTTPS calls. Distinct from `tls_*_path` below, which configure the
+        // ingress HTTP server.
+        let outgoing_handler = if dev_config.http_client_ca_paths.is_empty() {
```
ricochet: I think I should change the approach to support a very common enterprise use-case. Corporate-CA often wants exactly one root, and today there is no way to express that. 

If we stop modeling this as booleans and make the base an enum, with extra_ca_paths layered on top. Then we could address my other comment about disable_native_certs.

```rust
/// Built-in roots to start from before layering on `extra_ca_paths`.
#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
pub enum TrustRoots {
    /// Compiled-in webpki (Mozilla) roots plus the platform's native store.
    /// The native store honours `SSL_CERT_FILE`/`SSL_CERT_DIR`.
    #[default]
    WebpkiAndNative,
    /// Compiled-in webpki roots only — reproducible, ignores the host environment.
    Webpki,
    /// Platform native store only.
    Native,
    /// No built-in roots: trust exactly `extra_ca_paths` and nothing else.
    ExtraOnly,
}

pub struct ClientTlsOptions {
    pub roots: TrustRoots,
    pub extra_ca_paths: Vec<PathBuf>,
}
```

ricochet: One flag rather than a second boolean: `--http-client-trust-roots <webpki-and-native|webpki|native|extra-only> (clap::ValueEnum)` and `dev.http_client_trust_roots in config`.

crates/wash/src/cli/dev.rs
```diff
@@ -161,8 +161,31 @@ impl CliCommand for DevCommand {
...
+                ..Default::default()
+            }
+            .build()
+            .context("failed to load dev.http_client_ca_paths CA certificates")?;
+            debug!(
+                paths = ?dev_config.http_client_ca_paths,
+                "extra CA certificates trusted for outbound HTTPS"
+            );
+            wash_runtime::host::http::DefaultOutgoingHandler::with_tls_config(tls)
+        };
```
ricochet:
```suggestion
        let outgoing_handler = wash_runtime::host::http::DefaultOutgoingHandler::from_ca_paths(
            &dev_config.http_client_ca_paths,
        )
        .context("failed to load dev.http_client_ca_paths CA certificates")?;
```

crates/wash/src/cli/host.rs
```diff
@@ -87,6 +87,17 @@ pub struct HostCommand {
     #[arg(long = "tls-ca-path")]
     pub tls_ca_path: Option<PathBuf>,
 
+    /// Extra CA certificate bundle files (PEM) trusted for outbound HTTPS
+    /// requests made by components (`wasi:http` outgoing handler). Use this to
+    /// reach hosts behind a corporate or otherwise private CA. May be repeated
+    /// or comma-separated.
```
ricochet:
```suggestion
    /// reach hosts behind a corporate or otherwise private CA.
    ///
    /// Accepts a comma-separated list and/or repeated flags, e.g.
    /// `--http-client-ca-path /etc/wash/ca/corp.pem,/etc/wash/ca/staging.pem`.
    /// Paths must not contain commas.
```

crates/wash/src/cli/host.rs
```diff
@@ -309,17 +320,32 @@ impl CliCommand for HostCommand {
...
+                wash_runtime::host::http::DefaultOutgoingHandler::default()
+            } else {
+                let tls = wash_runtime::host::http_client::ClientTlsOptions {
+                    extra_ca_paths: self.http_client_ca_paths.clone(),
+                    ..Default::default()
+                }
+                .build()
+                .context("failed to load outbound HTTP client CA certificates")?;
+                wash_runtime::host::http::DefaultOutgoingHandler::with_tls_config(tls)
+            };
```
ricochet:
```suggestion
            let outgoing_handler = wash_runtime::host::http::DefaultOutgoingHandler::from_ca_paths(
                &self.http_client_ca_paths,
            )
            .context("failed to load --http-client-ca-path CA certificates")?;
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+                        .with_root_certificates(rustls::RootCertStore {
+                            roots: webpki_roots::TLS_SERVER_ROOTS.into(),
+                        })
+                        .with_no_client_auth(),
+                )
+            })
+        })
+        .clone()
+}
+
```
ricochet, line 152:
```suggestion

/// Translate an error from a hyper connection driver into a P3 [`ErrorCode`].
///
/// wasmtime's equivalent (`ErrorCode::from_hyper_response_error`) is crate
/// private, so mirror it here: a timeout becomes `HttpResponseTimeout`, and
/// everything else falls through to the public request mapping, which already
/// recovers an `ErrorCode` carried in the error's source chain.
///
/// [`ErrorCode`]: wasmtime_wasi_http::p3::bindings::http::types::ErrorCode
pub(crate) fn p3_connection_error(
    err: hyper::Error,
) -> wasmtime_wasi_http::p3::bindings::http::types::ErrorCode {
    use wasmtime_wasi_http::p3::bindings::http::types::ErrorCode;

    if err.is_timeout() {
        ErrorCode::HttpResponseTimeout
    } else {
        ErrorCode::from_hyper_request_error(err)
    }
}

```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+        worker: Some(worker),
+        between_bytes_timeout,
+    })
+}
+
+/// Send a P3 outgoing request: wasmtime's `default_send_request` with the
+/// trust roots taken from `tls` instead of the compiled-in webpki bundle.
+///
+/// Mirrors the structure of the P3 gRPC egress path in `host::http`: the
+/// response body enforces the between-bytes timeout via
```
ricochet, line 393:
```suggestion
/// The response body enforces the between-bytes timeout via
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+/// [`crate::host::http::TimedBody`], and the returned request-error future
+/// resolves `Ok(())` (hyper reports body errors through the response body's
+/// own stream).
```
ricochet:
```suggestion
/// drives the hyper connection, resolving with the connection's outcome once
/// it finishes.
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+            _ if e
+                .to_string()
+                .starts_with("failed to lookup address information") =>
+            {
+                dns_error("address not available")
+            }
+            _ => ErrorCode::ConnectionRefused,
+        })?;
+
+    let (mut sender, _worker) = if use_tls {
```
ricochet:
```suggestion
    let (mut sender, conn_worker) = if use_tls {
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+        })?;
+        let (sender, conn) = timeout(connect_timeout, http1::handshake(TokioIo::new(stream)))
+            .await
+            .map_err(|_| ErrorCode::ConnectionTimeout)?
+            .map_err(ErrorCode::from_hyper_request_error)?;
+        let worker = wasmtime_wasi::runtime::spawn(async move {
+            if let Err(e) = conn.await {
+                warn!(err = %e, "dropping outbound connection error");
+            }
+        });
```
ricochet:
```suggestion
        let worker =
            wasmtime_wasi::runtime::spawn(async move { conn.await.map_err(p3_connection_error) });
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+    } else {
+        let (sender, conn) = timeout(connect_timeout, http1::handshake(TokioIo::new(tcp_stream)))
+            .await
+            .map_err(|_| ErrorCode::ConnectionTimeout)?
+            .map_err(ErrorCode::from_hyper_request_error)?;
+        let worker = wasmtime_wasi::runtime::spawn(async move {
+            if let Err(e) = conn.await {
+                warn!(err = %e, "dropping outbound connection error");
+            }
+        });
```
ricochet:
```suggestion
        let worker =
            wasmtime_wasi::runtime::spawn(async move { conn.await.map_err(p3_connection_error) });
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+            .boxed_unsync()
+        });
+
+    // The connection worker is moved into this future so it stays alive until
+    // the guest is done with the request, matching wasmtime's default
+    // transport. Body errors reach the guest through the response body stream.
+    let io: P3RequestErrorFuture = Box::new(async move {
+        let _worker = _worker;
+        Ok(())
+    });
```
ricochet:
```suggestion
        .map(|body| crate::host::http::TimedBody::new(body, between_bytes_timeout).boxed_unsync());

    // The connection driver *is* the request-error future. `spawn` hands back an
    // `AbortOnDropJoinHandle`, which aborts the connection task when dropped,
    // and wasmtime keeps this future alive for the response body's lifetime
    // only while it polls pending (see `p3::host::handler`). Returning a future
    // that is ready on the first poll would abort the connection before the
    // guest reads the body.
    let io: P3RequestErrorFuture = Box::new(conn_worker);
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+        }
+    }
+
+    fn p2_request(uri: &str) -> hyper::Request<HyperOutgoingBody> {
+        hyper::Request::builder()
+            .uri(uri)
+            .body(HyperOutgoingBody::default())
+            .unwrap()
+    }
+
```
ricochet, line 573:
```suggestion

    /// Serve one HTTP/1.1 request on an ephemeral port, sending the head and
    /// then the body `body_delay` later. Returns the port.
    async fn delayed_body_server(body_delay: Duration) -> u16 {
        use tokio::io::{AsyncReadExt as _, AsyncWriteExt as _};

        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
        let port = listener.local_addr().unwrap().port();
        tokio::spawn(async move {
            let (mut sock, _) = listener.accept().await.unwrap();
            let mut buf = [0u8; 4096];
            let mut seen = Vec::new();
            while !seen.windows(4).any(|w| w == b"\r\n\r\n") {
                match sock.read(&mut buf).await {
                    Ok(0) | Err(_) => return,
                    Ok(n) => seen.extend_from_slice(&buf[..n]),
                }
            }
            let _ = sock
                .write_all(b"HTTP/1.1 200 OK\r\ncontent-length: 5\r\n\r\n")
                .await;
            let _ = sock.flush().await;
            tokio::time::sleep(body_delay).await;
            let _ = sock.write_all(b"hello").await;
            let _ = sock.flush().await;
            tokio::time::sleep(Duration::from_secs(2)).await;
        });
        port
    }

    /// The request-error future returned by [`send_request_p3`] owns the hyper
    /// connection driver, so it must poll pending until the connection ends.
    /// wasmtime polls it once with a noop waker and only keeps it alive for the
    /// response body when it is pending, so a future that is ready immediately
    /// aborts the connection before the guest can read the body.
    #[tokio::test]
    async fn p3_request_error_future_keeps_connection_alive() {
        use core::task::{Context, Waker};

        let port = delayed_body_server(Duration::from_millis(300)).await;
        let request = hyper::Request::builder()
            .uri(format!("http://127.0.0.1:{port}/"))
            .body(P3Body::new(
                http_body_util::Empty::new().map_err(|_: std::convert::Infallible| unreachable!()),
            ))
            .unwrap();

        let (response, io) = send_request_p3(default_client_tls_config(), request, None)
            .await
            .expect("request should succeed");

        let mut io = Box::into_pin(io);
        assert!(
            io.as_mut()
                .poll(&mut Context::from_waker(Waker::noop()))
                .is_pending(),
            "request-error future must stay pending so wasmtime ties it to the body"
        );

        // Drive it the way wasmtime does once it sees `Pending`.
        let io = wasmtime_wasi::runtime::spawn(async move { io.await });
        let body = tokio::time::timeout(
            Duration::from_secs(3),
            BodyExt::collect(response.into_body()),
        )
        .await
        .expect("body read timed out")
        .expect("body read failed");
        assert_eq!(body.to_bytes().as_ref(), b"hello");
        drop(io);
    }

    /// A server that hangs up mid-body must surface as an error on the response
    /// body, not as a silently truncated success. hyper completes the
    /// connection driver cleanly here, so this is the channel the guest sees.
    #[tokio::test]
    async fn p3_truncated_body_surfaces_as_body_error() {
        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
        let port = listener.local_addr().unwrap().port();
        tokio::spawn(async move {
            let (sock, _) = listener.accept().await.unwrap();
            let mut sock = sock;
            use tokio::io::{AsyncReadExt as _, AsyncWriteExt as _};
            let mut buf = [0u8; 4096];
            let _ = sock.read(&mut buf).await;
            // Promise five bytes, then hang up without sending them.
            let _ = sock
                .write_all(b"HTTP/1.1 200 OK\r\ncontent-length: 5\r\n\r\n")
                .await;
            let _ = sock.flush().await;
            drop(sock);
        });

        let request = hyper::Request::builder()
            .uri(format!("http://127.0.0.1:{port}/"))
            .body(P3Body::new(
                http_body_util::Empty::new().map_err(|_: std::convert::Infallible| unreachable!()),
            ))
            .unwrap();

        let (response, io) = send_request_p3(default_client_tls_config(), request, None)
            .await
            .expect("response head should arrive");
        let io = wasmtime_wasi::runtime::spawn(async move { Box::into_pin(io).await });

        let err = tokio::time::timeout(
            Duration::from_secs(3),
            BodyExt::collect(response.into_body()),
        )
        .await
        .expect("body read should not hang")
        .expect_err("a truncated body must not read as success");
        assert!(
            matches!(
                err,
                wasmtime_wasi_http::p3::bindings::http::types::ErrorCode::HttpProtocolError
            ),
            "expected HttpProtocolError, got {err:?}"
        );
        drop(io);
    }

```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+    /// HTTPS to a server behind a private CA must work once that CA is added
+    /// via `extra_ca_paths`, and must keep failing with a TLS error without it.
+    #[tokio::test]
+    async fn extra_ca_enables_https_to_private_ca_server() {
+        use tokio::io::{AsyncReadExt, AsyncWriteExt};
+        use wasmtime_wasi_http::p2::bindings::http::types::ErrorCode;
+
+        crate::init_crypto();
```
ricochet, line 589:
```suggestion
        crate::init_crypto();
        // An IP SAN, so the test can dial 127.0.0.1 directly rather than relying
        // on `localhost` resolving to the address the listener bound.
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+        let certified_key =
+            rcgen::generate_simple_self_signed(vec!["localhost".to_string()]).unwrap();
```
ricochet:
```suggestion
            rcgen::generate_simple_self_signed(vec!["127.0.0.1".to_string()]).unwrap();
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+        });
+
+        let dir = tempfile::tempdir().unwrap();
+        let ca_path = dir.path().join("ca.pem");
+        std::fs::write(&ca_path, certified_key.cert.pem()).unwrap();
+
+        // Without the CA: the handshake must fail with a TLS error.
+        let err = send_request_p2(
+            default_client_tls_config(),
+            p2_request(&format!("https://localhost:{port}/")),
```
ricochet:
```suggestion
            p2_request(&format!("https://127.0.0.1:{port}/")),
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -0,0 +1,522 @@
...
+        // With the CA: the same request must succeed.
+        let tls = ClientTlsOptions {
+            extra_ca_paths: vec![ca_path],
+            disable_native_certs: true,
+        }
+        .build()
+        .unwrap();
+        let response = send_request_p2(
+            tls,
+            p2_request(&format!("https://localhost:{port}/")),
```
ricochet:
```suggestion
            p2_request(&format!("https://127.0.0.1:{port}/")),
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -435,12 +435,37 @@ pub trait OutgoingHandler: Send + Sync + 'static {
...
 
-/// Default [`OutgoingHandler`] — defers to `wasmtime_wasi_http::p2::default_send_request` (P2)
-/// and `wasmtime_wasi_http::p3::default_send_request` (P3).
-#[derive(Default)]
-pub struct DefaultOutgoingHandler;
+/// Default [`OutgoingHandler`] — wasmtime's default per-request transport, but
+/// with configurable TLS trust roots (webpki + platform native certificates +
+/// optional extra CA bundles; see
+/// [`crate::host::http_client::ClientTlsOptions`]).
+pub struct DefaultOutgoingHandler {
```
ricochet, line 454: Change to no I/O in Default, none wasted in the builder, so IngressBuilder::new no longer reads the platform trust store for a handler both CLI call sites immediately discard.

crates/wash/src/cli/host.rs
ricochet: This is duplicated wiring. Have a shared util `DefaultOutgoingHandler::from_ca_paths(&[PathBuf])`.

crates/wash-runtime/src/host/http.rs
```diff
@@ -435,12 +435,37 @@ pub trait OutgoingHandler: Send + Sync + 'static {
         options: Option<wasmtime_wasi_http::p3::RequestOptions>,
         fut: crate::host::http_p3::P3RequestErrorFuture,
     ) -> crate::host::http_p3::P3SendFuture;
+
+    /// TLS configuration used for host-mediated egress that bypasses
+    /// `send_request`/`send_request_p3` (currently the gRPC fast path).
+    /// `None` (the default) means the process-wide default trust roots.
+    fn client_tls_config(&self) -> Option<Arc<rustls::ClientConfig>> {
```
ricochet, line 442: Change to have gRPC h2 config built once so that the ALPN-h2 variant is derived on first gRPC request, so the per-request (*tls).clone() + Arc::new is gone from both paths.

crates/wash-runtime/src/host/http.rs
```diff
@@ -2032,9 +2065,9 @@ fn send_grpc_request_p3(
 /// fires (or the stream ends / errors), releasing the underlying HTTP/2 stream
 /// and TCP connection eagerly instead of leaving it pinned until the guest
 /// drops its body handle.
-struct TimedBody<B> {
-    inner: Option<B>,
-    interval: tokio::time::Interval,
+pub(crate) struct TimedBody<B> {
```
ricochet, line 2050: Introduce TimedBody::new(inner, between_bytes_timeout) rather than making these pub.

jfleitz:
Thank you very much @ricochet for the review! Since your review what was changed:

- Changed to be an enum vs boolean for the flag for `--http-client-trust-roots`
- **Default is `webpki`** — identical to wasmtime's default transport, so unconfigured hosts see zero behavior change and the native store is an explicit opt-in. 
- Test added for this to ensure that the default really is "webpki only"
- wasiP3 connection driver issue addressed: request-error future is now the connection driver itself — pending until the connection ends, propagating errors via a `p3_connection_error` helper mirroring upstream's error handling (as called out in your review)
- Adds pending-poll-with-noop-waker + delayed-body test, and a truncated-body-surfaces-as-`HttpProtocolError` test.
- P3 egress through a private CA test, and gRPC P2/P3 tests added proving the fast path picks up configured roots; the P3 gRPC test also asserts the io-future-pending contract.
- de-duplicated/refactored functions that were similar as called out.
- Test certs use `127.0.0.1` vs localhost

ricochet (approved):

PR #5423, opened 2026-08-03
fix(helm): update workload crds, add check

### LINKED ISSUE #5421: chart: bundled CRDs omit allowedIpNameLookups, silently pruned for chart users
The per-component DNS allowlist (`allowIpNameLookup` in 2.6.0, renamed `allowedIpNameLookups` in 2.6.1 via #5408) is present in the operator's Go API (`runtime-operator/api/runtime/v1alpha1/workload_types.go`) and the generated CRDs under `runtime-operator/config/crd/bases/` — but the CRDs vendored into the chart at `charts/runtime-operator/crds/` still list only `allowedHosts` under `localResources`, at both the v2.6.1 tag and current `main`. (The chart's templates on `main` do render a same-named `allowedIpNameLookups` value for host plugins, but that's a separate feature — the workload-level CRD field is what's missing.)

For anyone installing via the Helm chart, applied `Workload`/`WorkloadDeployment` manifests using the field either fail strict field validation (`kubectl apply` on Kubernetes 1.25+: `unknown field ... allowedIpNameLookups`) or have the field silently pruned (clients using default field validation, common in GitOps controllers) — either way, the deny-by-default lookup policy can't be opted out of.

Suggested fix: re-vendor the chart CRDs from `runtime-operator/config/crd/bases/` (and perhaps add a CI check that diffs the two so they can't drift again).

Docs on wasmcloud.com currently carry a caveat pointing users at the operator's current CRD definitions as a workaround (wasmcloud.com#1234).

vados-cosmonic:
This doesn't change the deny-by-default nature, but at the very least people will be able to now set the allowed ip name lookups.

Resolves #5421

### DIFF
--- a/.github/workflows/charts.yml
+++ b/.github/workflows/charts.yml
@@ -31,7 +31,9 @@ jobs:
           filters: |
             relevant:
               - 'charts/**'
+              - 'runtime-operator/config/crd/bases/**'
               - '.github/workflows/charts.yml'
+              - 'Makefile'
 
   test:
     needs:
@@ -45,6 +47,8 @@ jobs:
         with:
           persist-credentials: false
       - uses: ./.github/actions/setup-go
+      - name: Check bundled CRDs
+        run: make helm-crds-check
       - name: Render
         run: make helm-render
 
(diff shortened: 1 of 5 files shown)


jfleitz (approved):

ricochet:
I'm surprised by the drift here. We should have a check for this in CI

PR #5424, opened 2026-08-03
Outbound connection pooling for http requests

jfleitz:
## Feature or Problem

There was an error reported on wasmCloud 2.5.1:
Running into a little bug (potentially in wasmtime?) when doing some load testing  on concurrent HTTP requests. It is a simple setup, 3 concurrent HTTP requests are made to the same endpoint (running on local host), wait for all responses and then return. After doing some load testing I get the following error:

```sh
stdout [324] :: Error sending request: Post "http://127.0.0.1:8123/readsomedata": DNS error: rcode="address not available", infoCode=0
stdout [324] :: Request 1 failed with status code: 0
```

This occurs in both `wash dev` running `2.5.1` and also wasmtime `46.0.1`. I've tried reducing the concurrency count to 2, also changing from `localhost` to `127.0.0.1` but still no luck.

## Issue

This is not a DNS failure. The outbound transport (wasmtime's `default_send_request`) opens a **fresh TCP connection for every outgoing request** and closes it when the response is drained. Every closed connection parks its ephemeral port in `TIME_WAIT` (~15s on macOS, 60s on Linux), giving sustained outbound traffic a hard ceiling of roughly `port_range / TIME_WAIT` (~1,000 req/s on macOS, ~470 req/s on default Linux).

## Fix

Replaces the per-request transport with keep-alive connection pools **keyed per workload**
- **Connection reuse removes the failure cause.** A workload's repeated and concurrent requests to the same authority reuse a handful of sockets, so no `TIME_WAIT` accumulates at all. Side benefit: no per-request TCP+TLS handshake, so tail latency drops.
- **Per-workload isolation.** Components never share a TCP connection, so connection-scoped server state (auth, rate-limit attribution, sticky LB) cannot leak between workloads. The pool is HTTP/1.1-only deliberately: exclusive connection checkout per request is what keeps the no-shared-socket guarantee simple; enabling HTTP/2 would multiplex streams on one socket and change the isolation story.
- **Bounded cleanup, three nested lifetimes:**

  | Layer | Mechanism | Bound |
  |---|---|---|
  | Idle connection | hyper-util `pool_idle_timeout` | closed after 60s idle (at/below common server/LB keep-alive windows) |
  | Idle connections per authority | `pool_max_idle_per_host` | 32 per (workload, authority) — must sit **above** burst concurrency: a cap of 4 measured 1625 connections for 3000 requests (the pool churned every burst), 32 measured 31 |
  | The workload's client/pool | moka `time_to_idle` | evicted after 60s without a request; in-flight requests hold their own clone and are unaffected |

## Related Issues
- Builds on the `OutgoingHandler` trait from #5014 (motivated by #4978) and the gRPC egress path from wash#115. The gRPC path keeps its per-request transport in this PR (follow-up below).

## Consumer impact

- `DefaultOutgoingHandler` construction is unchanged (`Default` or `with_tls_config`); it now pools internally and uses the `workload_id` it
  already received. Custom `OutgoingHandler` implementations are unaffected.
- Outbound connections to the same authority are now reused across requests from the same workload; servers see far fewer connections.
- The pooled path can't observe a separate connect phase, so `connect_timeout + first_byte_timeout` is folded into one response-head
  budget, surfaced as `ConnectionReadTimeout` on expiry.

## Security considerations
- (New hardening in this PR) TLS session tickets no longer resume across workloads. Previously every workload shared one rustls session-ticket cache, so an upstream server could correlate two workloads by observing a resumed TLS session. Each workload's client now keeps a private resumption store (each gRPC connection gets a fresh one), while resumption still works normally within a workload on the pooled path.
- (Configurable) The host-wide budget is shared, not fair-shared. Idle pooled connections pin permits until the 60s idle timeout, so several concurrently busy workloads can hold the global budget and starve other workloads' new connections for up to that window; affected requests fail promptly as connect timeouts rather than hanging, and the host logs a rate-limited warning identifying which budget (per-workload or global) was exhausted. Both caps are operator-tunable: --http-client-max-connections / --http-client-max-connections-per-workload on wash host (or WASH_HTTP_CLIENT_MAX_CONNECTIONS[_PER_WORKLOAD]), and dev.http_client_max_connections[_per_workload] for wash dev. Fair-sharing the global budget across workloads is a follow-up.
- (Unchanged from before this PR) Connection caps don't cover the gRPC egress path yet. The gRPC fast path is selected by the guest (content-type: application/grpc) and opens its own per-request HTTP/2 connection, bypassing the pooled client and its ConnectionLimits. Exposure is bounded — these connections are request-scoped, so a workload holds at most one per in-flight request — but the caps should not be read as a hard host-wide FD guarantee until the gRPC pooling follow-up brings that path under the same permit system.

### Unit tests (`host::http_client`)

- `sequential_requests_reuse_the_pooled_connection` — 20 sequential requests must arrive over ≤2 connections (counted server-side).
- `workloads_reuse_own_pool_but_never_share_connections` — two workloads hitting the same authority reuse within their own pool but never each other's idle connections.
- TLS trust-root tests (private CA accepted when configured, `TlsProtocolError` when not, CA bundle validation) carried over the pooled
  transport.

### Local repro (`wash dev`)

Using the a Go "http concurrent load" example project (a wasip3 Go component whose `GET /load` fans out 3 concurrent outbound POSTs to a local backend that **counts TCP connections**, plus a load generator):

| | stock wash 2.6.1 | this branch |
|---|---|---|
| loadtest (1000 inbound × 3 outbound, 10 workers) | 502s appear mid-run; bodies show `DNS error (rcode=address not available)`; backend sees ~1 connection per request | **all 200s in <1s** |
| backend `connections_opened` | ≈ `requests_served` | **31 for 3000 requests** |

The same run also caught the idle-cap regression documented above (cap 4 → 1625 connections; cap 32 → 31), which is why `POOL_MAX_IDLE_PER_HOST` is 32 with the measurement recorded in its doc comment.

### Local kind cluster (wasmCloud v2 Kubernetes hosts)

Verified on the same runtime path the Kubernetes hosts use, via the standard
kind flow:

1. **Host image from this branch**: `docker build` with the repo's `Dockerfile`, loaded into a kind cluster created from
   `deploy/kind/kind-config.yaml`.
2. **Backend as an in-cluster Deployment + Service**; the component's `WorkloadDeployment` grants `allowedHosts` for that service authority
   (same fail-closed egress policy as dev), and the load generator targets the workload through a port-forward with `?target=<backend-svc>:8123`.
3. **Result**: matches the local run — loadtest completes with all 200s and the backend's connection count stays bounded at ≈ the workload's burst concurrency, where the stock 2.6.1 host image shows connection-per-request and degrades into the `address not available` failures under load.


ricochet:
We should go ahead and handle the gRPC case here. There's a good deal of overlap with the cases we're handling.

jfleitz:
@ricochet Your changes LGTM...I rebased and fixed the testing error.

ricochet (approved):

PR #5435, opened 2026-08-07
feat: enable implements and maps by default

ricochet:
WASI SG voted to depend on stabilized CM features maps and implements today.

This change turns on both component-model proposals for every host build.

- (implements ..) named imports — wasm_component_model_implements moves from opt-in to a default feature of both wash-runtime and wash. A shipped wash host / wash dev now accepts components that import an interface more than once under distinct labels, and registers the multiplexed plugin set that routes each named import to its own backend.
- map<k, v> types — new WasmProposal::ComponentModelMap, applied unconditionally in EngineBuilder::build() alongside ComponentModelAsync. Components whose WIT uses map types now validate and instantiate.

map needs no cargo feature: wasmtime exposes wasm_component_model_map under plain component-model, and no code in the tree is conditional on it. It parses as component-model-map for --wasm-proposal and dev.wasm_proposals.

### DIFF
--- a/.github/workflows/wash.yml
+++ b/.github/workflows/wash.yml
@@ -124,11 +124,11 @@ jobs:
           RUST_BACKTRACE: "1"
         run: |
           cargo test --workspace
-      - name: Test with wasi-tls + wasm_component_model_implements
+      - name: Test with wasi-tls + host-component-plugins
         env:
           RUST_BACKTRACE: "1"
         run: |
-          cargo test -p wash-runtime --features wasi-tls,wasm_component_model_implements,host-component-plugins
+          cargo test -p wash-runtime --features wasi-tls,host-component-plugins
           cargo test -p wash --lib --features host-component-plugins
       # Docker/registry-backed `#[ignore]` integration tests, Linux only: the
       # macOS/Windows hosted runners have no Docker daemon (and macOS can't get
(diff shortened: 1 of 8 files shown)


.github/workflows/wash.yml
```diff
@@ -124,11 +124,11 @@ jobs:
           RUST_BACKTRACE: "1"
         run: |
           cargo test --workspace
-      - name: Test with wasi-tls + wasm_component_model_implements
+      - name: Test with wasi-tls + host-component-plugins
```
ricochet, line 127: next time I'm here: I'll add a env var to keep the list of "all features" consistent. Not planning to add this round

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -684,6 +688,9 @@ impl WasmProposal {
             WasmProposal::WasmComponentModelImplements => {
                 cfg.wasm_component_model_implements(true);
             }
+            WasmProposal::ComponentModelMap => {
+                cfg.wasm_component_model_map(true);
```
ricochet, line 692: This is all we need to enable maps

runtime-operator/test/e2e/implements_test.go
```diff
@@ -50,13 +49,11 @@ var _ = Describe("Implements Named Host Interfaces", Ordered, func() {
...
-		// others it needs a feature-enabled host to RUN (`(implements ..)`
-		// multiplexing), so it runs only when the registry flow is on AND the
-		// fixture host is an all-features build — i.e. the all-features leg.
-		if !inClusterRegistry || !defaultHostAllFeatures {
-			Skip("skipping implements e2e (needs the in-cluster registry and an " +
-				"all-features fixture host)")
+		// registry (make e2e-images), like every other e2e fixture, and
+		// `(implements ..)` multiplexing is a default host feature — so this runs
+		// on both legs whenever the registry flow is on.
+		if !inClusterRegistry {
```
ricochet, line 55: I'm working through refactoring how the inClusterRegistry works in a different PR so trying to change as little as possible here right now.

vados-cosmonic (approved):

PR #5439, opened 2026-08-07
feat(wash-runtime): one connection quota and one socket policy per guest

ricochet:
Connection limiting was performed by two unrelated mechanisms whose names read as synonyms: the HTTP client's `ConnectionLimits`, which bounded how large a pool could grow, and a separate allowance for raw sockets. Neither knew about the other, so a guest's real ceiling was whichever it hit first, and the host-wide total was unbounded.

One `GuestConnectionQuota`, minted per guest by a `QuotaRegistry`, with three surfaces:

| surface | bounds | waits? |
| --- | --- | --- |
| `http` | pooled outbound connections (idle keep-alive ones count) | yes |
| `sockets` | raw guest sockets | no |
| `inbound` | connections arriving at the guest | no |

All three roll up into one host-wide ceiling. `sockets` never awaits: a guest blocking on its own slot could deadlock against itself.

**Breaking, but unreleased.** The CLI is now one family: `--max-connections`, `--max-{http,socket,inbound}-connections-per-workload`, `--http-connection-wait`. The old `--http-client-max-connections*` names are removed rather than aliased; they were added on 2026-08-05 and have never appeared in a tag.

This commit also carries the integration with `main`'s recent idle-cap work: `idle_per_authority` now clamps against the quota's `http` limit, which is what that change already called "the workload's connection budget."

On the second commit, another related feature is introduced for how a loopback is granted.

Socket permission previously was a closure per call site, one in `linked_call.rs`, another in the plugin store builder, with each answering `bool` with its own idea of what loopback meant. A rule added to one silently skipped the other.

`SocketPolicy` replaces both and answers `AddrDecision`, which carries three things a bool cannot:

- **which plane** the operation dispatches onto. The address no longer implies it: `127.0.0.1` may mean the guest's private virtual network or the machine's real loopback.
- **why** it was refused, so refusals can be metered by reason.
- **a quota slot**, taken at the decision, so nothing is counted that is not also bounded.

Egress is gated in two layers. The workload's declared `allowedHosts`, then an `EgressAddressPolicy` covering ranges no allowlist entry should reach (link-local, cloud metadata, and by default the private ranges). Both run under an `EgressMode` that **defaults to counting rather than refusing**, so operators see what a policy would break before it breaks anything.

`allowedHostLoopback` grants specific ports on the machine's own loopback, per protocol (`5432`, `5432/tcp`, `53/udp`). It is inert unless the host also runs with `--allow-host-loopback`: neither the workload author nor the operator can open that door alone.


jfleitz (dismissed):

charts/runtime-operator/crds/runtime.wasmcloud.dev_workloaddeployments.yaml
```diff
@@ -152,6 +152,32 @@ spec:
                               description: LocalResources describes resources that
                                 will be made available to a workload component.
                               properties:
+                                allowedHostLoopback:
```
vados-cosmonic:
```suggestion
                                allowedHostLoopbackPorts:
```

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -70,6 +70,12 @@ pub(crate) struct ComponentCtxTemplate {
     volume_mounts: Vec<ResolvedVolumeMount>,
     plugins: Option<HashMap<&'static str, Arc<dyn HostPlugin + Send + Sync>>>,
     loopback: Arc<std::sync::Mutex<loopback::Network>>,
+    /// The host-level half of this component's socket policy: enforcement mode,
+    /// address ranges, whether host-loopback access is enabled at all, and the
+    /// budget. The workload-level half (`allowedHosts`,
+    /// `allowedHostLoopback`) comes from `local_resources` and is layered over
+    /// this when the check is built.
+    network_policy: Arc<crate::sockets::policy::SocketPolicy>,
```
vados-cosmonic: I think we should be consistent here and call it `socket_policy`

crates/wash-runtime/src/host/quota.rs
```diff
@@ -0,0 +1,592 @@
...
+#[cfg(not(unix))]
+fn descriptor_soft_limit() -> Option<usize> {
+    None
+}
+
+/// Per-guest ceilings, one per surface.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct QuotaLimits {
+    /// Pooled `wasi:http` and gRPC connections, counting idle keep-alive ones.
+    pub http: usize,
```
vados-cosmonic: I wonder why this isn't inidicated as `outbound_http` -- it's outbound only AFAICT.

crates/wash-runtime/src/types.rs
```diff
@@ -121,6 +122,16 @@ pub struct LocalResources {
...
+    /// Ports on the machine's own loopback this component may reach through
+    /// `host.wasmcloud.internal`.
+    ///
+    /// **Empty = deny every host-loopback connection**, and a non-empty list is
+    /// still inert unless the host itself was started with
+    /// `--allow-host-loopback`: neither a workload author nor an operator can
+    /// open this door alone. See [`crate::host::allowed_loopback`] for the
+    /// accepted entry forms. Strings from the wire (proto / wash YAML) are
+    /// parsed at conversion time.
+    pub allowed_host_loopback: Arc<[AllowedLoopbackPort]>,
```
vados-cosmonic: yeah this is a place where `allowed_host_loopback_ports` might be a bit clearer, but it's not a *huge* deal.

crates/wash/src/cli/dev.rs
```diff
@@ -162,12 +162,21 @@ impl CliCommand for DevCommand {
...
+        let http_connection_wait = dev_config
+            .http_connection_wait
             .as_deref()
             .map(humantime::parse_duration)
             .transpose()
-            .context("dev.http_client_connection_wait is not a valid duration (e.g. `5s`)")?;
+            .context("dev.http_connection_wait is not a valid duration (e.g. `5s`)")?;
+        // One registry for every surface: HTTP pool, raw sockets, inbound
+        // published ports.
+        let quotas = crate::config::connection_quotas(
```
vados-cosmonic: This should probably be a builder, or maybe we could "just" take a reference to `dev_config` (this could even be a method on `dev_config` -- building the connection quota based on config, then we make any additional tweaks)

crates/wash/src/cli/host.rs
```diff
@@ -116,34 +116,60 @@ pub struct HostCommand {
...
-    /// Host-wide cap on live outbound HTTP connections across all workloads
-    /// combined (in-flight or idle in a keep-alive pool). Size it for the
-    /// number of concurrently busy workloads times their burst concurrency.
+    /// Host-wide cap on live connections across every workload and surface
+    /// combined — pooled HTTP, raw sockets, and inbound published ports.
+    ///
+    /// Size it for the number of concurrently busy workloads times their burst
+    /// concurrency, kept inside the process's file-descriptor limit.
+    #[arg(long = "max-connections", env = "WASH_MAX_CONNECTIONS")]
+    pub max_connections: Option<usize>,
```
vados-cosmonic, line 125: can we add an alias here to avoid breaking existing setups?

crates/wash/src/cli/host.rs
```diff
@@ -116,34 +116,60 @@ pub struct HostCommand {
...
+
+    /// Cap on live pooled HTTP and gRPC connections a single workload may
+    /// hold, across all authorities it talks to. Idle keep-alive connections
+    /// count, so this is really how large a workload's pool may grow.
+    #[arg(
+        long = "max-http-connections-per-workload",
+        env = "WASH_MAX_HTTP_CONNECTIONS_PER_WORKLOAD",
+        default_value_t = 128
+    )]
+    pub max_http_connections_per_workload: usize,
```
vados-cosmonic: I think client kind of told us this was outgoing, but now you can't tell... Adding `outgoing` might go a long way here.

crates/wash/src/config.rs
```diff
@@ -693,18 +777,26 @@ pub struct DevConfig {
...
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub http_client_max_connections: Option<usize>,
+    pub max_connections: Option<usize>,
 
     /// Cap on live *outbound* HTTP connections a single workload may hold,
     /// across all authorities it talks to. Defaults to the runtime's
     /// built-in limit.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub http_client_max_connections_per_workload: Option<usize>,
+    pub max_http_connections_per_workload: Option<usize>,
```
vados-cosmonic: same note here about the fact that this is outbound being lost

vados-cosmonic (dismissed):
LGTM 🚀 

This could go as-is, I think the only thing worth considering is naming.

crates/wash/src/cli/host.rs
ricochet: Right now max_connections hasn't been released. Do you still want the alias?

vados-cosmonic: Ah so is this not equivalent to `http_client_max_connections`? It seemed like the new one was `max_http_connections_per_workload` -- the old setting was across  all workloads

crates/wash-runtime/src/host/quota.rs
ricochet: Now outbound_http. Matches  outbound_http, outbound_sockets, inbound_sockets

crates/wash/src/cli/dev.rs
ricochet: Added DevConfig::connection_quotas() so it's a one-liner now

crates/wash/src/cli/host.rs
ricochet: It was but `http_client_max_connections` was merged Thurs and not released yet

crates/wash/src/cli/host.rs
ricochet: Opted for adding in `outbound` to match everything else: `max-outbound-http-connections-per-workload`

crates/wash/src/cli/host.rs
vados-cosmonic: Ahhh gotcha, in that case then yep, no need to have an alias!

PR #5442, opened 2026-08-07
feat(runtime): host component call a workload's exports

ricochet:
A host component plugin could only be called over the host component plugin's exports. Its imports resolved against the WASI base and the host's native plugins, and anything left over failed to load. The host component could not invoke workload component APIs.

Host-native plugins have always done both. `wasi:http` and`wasmcloud:messaging` declare a package, notice at bind time which components export the handler they want to call, and capture(ResolvedWorkload, InstancePre, component_id) in `on_workload_resolved` in order to invoke a workload component.

A plugin that is itself a component cannot hold that value, its importis one static linker instance with no parameter naming a target so thehost keeps it on the plugin's behalf, and `wasmcloud:host/workload` handsthe guest a handle into it whose lifetime is the routing scope. A callwith no handle held goes back to the workload whose capability call isbeing served, resolved from the caller's root guest task exactly as`wasmcloud:host/identity` is, so a plugin calling back into its owncaller addresses nothing.

Matching needed no change: `includes_bidirectional` already covers aworkload that satisfies an interface by exporting it, so publishing theplugin's workload-facing imports in its world is enough. Neither did thecall path `EphemeralLinkedCall` already crosses a store boundary withwarm-instance pooling and stream/future relocation, so each functionresolves to a `LinkedExportInvocation` pinned to it. `resource` and`error-context` handles cannot cross, and are refused when the pluginloads or the workload deploys rather than on the call.

| | native (`wasi:http`, `wasmcloud:messaging`) | host component plugin |
|---|---|---|
| declares | the package in `world()` | same |
| finds exporters | `exports_wasi_http` / `exports_messaging_handler` on the component's world at bind | same, per workload-facing import |
| holds the target | `(ResolvedWorkload, InstancePre, component_id)` in a Rust map | the same map, host-side |
| names the target | picks out of its own map | `wasmcloud:host/workload` `target` handle |

The last row is the only real gap: a wasm guest can't hold a Rust value, and its import is one static linker instance with no parameter naming a target. So the handle is a resource whose **lifetime** is the routing scope, and a call with no handle held falls back to the workload whose capability call is being served which is the resolution `wasmcloud:host/identity` already performs.
A plugin may now export no capability at all, which makes a pure trigger, one that only dispatches, from its own `wasi:cli/run`.

A plugin import that nothing satisfies used to fail at load. It now loads and traps on the call that needs it, with an error naming the interface. That is inherent to resolving these against workload exports without a new declaration list in ComponentPluginSpec; the plugin's world still has to match a workload that both declares and exports the interface for anything to bind.

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -2066,6 +2066,16 @@ impl UnresolvedWorkload {
 
         trace!(host_interfaces = ?host_interfaces, "determining missing guest interfaces");
 
+        // A workload serves itself before the host does. Every component's world
+        // up front, so an interface one component imports and another exports can
+        // be recognised as already answered from inside the workload — see
+        // [`served_within_workload`].
+        let component_worlds: Vec<(Arc<str>, WitWorld)> = self
+            .components
+            .iter()
+            .map(|(id, component)| (id.clone(), component.world()))
+            .collect();
+
         if let Some(service) = self.service.as_ref() {
             let world = service.world();
 
@@ -2073,6 +2083,9 @@ impl UnresolvedWorkload {
             let required_interfaces: HashSet<WitInterface> = host_interfaces
                 .iter()
                 .filter(|wit_interface| world.includes_bidirectional(wit_interface))
+                .filter(|wit_interface| {
+                    !served_within_workload(wit_interface, service.id(), &world, &component_worlds)
+                })
                 .cloned()
                 .collect();
 
@@ -2084,12 +2097,14 @@ impl UnresolvedWorkload {
             }
         }
 
-        for (id, workload_component) in &self.components {
-            let world = workload_component.world();
+        for (id, world) in &component_worlds {
             trace!(?world, "comparing component world to host interfaces");
             let required_interfaces: HashSet<WitInterface> = host_interfaces
                 .iter()
                 .filter(|wit_interface| world.includes_bidirectional(wit_interface))
+                .filter(|wit_interface| {
+                    !served_within_workload(wit_interface, id, world, &component_worlds)
+                })
                 .cloned()
                 .collect();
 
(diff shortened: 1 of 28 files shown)


crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1578,6 +1620,160 @@ impl ResolvedWorkload {
...
+        if let Some(service) = &self.service
+            && service.id() == item_id
+        {
+            return if exports(&service.world()) {
+                ItemExport::Service
+            } else {
+                ItemExport::None
+            };
+        }
+        let components = self.components.read().await;
```
vados-cosmonic, line 1646: NIT: I don't think there's a *ton* of time to hold this read but let's drop it quickly, AFAICT it doesn't get used after we decide whether to return or not

crates/wash-runtime/src/plugin/component_host/mod.rs
```diff
@@ -46,6 +46,20 @@
 //! delivery, post-restart replay, and quarantine — lives in the `lifecycle`
 //! submodule; this module drives it from `ComponentHostPlugin`'s `HostPlugin`
 //! impl.
+//!
+//! Calls also run the other way. A plugin that imports `wasmcloud:host/workload`
```
vados-cosmonic: So right now IIRC components that provide the interface are considered *before* the host. We might want to make this controllable in a follow up to specify which interfaces are expected to be provided by a sibling component.

Something like `wasmcloud:secrets` we probably don't want to get intercepted, maybe

crates/wash-runtime/src/plugin/component_host/workload_call.rs
vados-cosmonic, line 1: NIT: I kind of wonder why we didn't call this package `workload-call`

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but nits!

crates/wash-runtime/src/engine/workload.rs
vados-cosmonic: Oh NVM, you do need this to stay alive

crates/wash-runtime/src/plugin/component_host/mod.rs
ricochet: The reason why I want Workloads to always be self-resolved is that they sit on the same store and conceptually should be considered one component. Right now we're essentially composing them together at runtime with extra steps for separate lifetimes. I aim to actually compose them once we have child handles and runtime instantiation of components.

The way to control what is/isn't provided by the host is to pre-compose via wac.

crates/wash-runtime/src/plugin/component_host/workload_call.rs
ricochet: Yeah good call. Went ahead and made this change in this PR. I've been iterating on another change on top of this for protecting Host Component Plugins from failures in guests so I'll put up another PR that rejiggers the wit.

PR #5445, opened 2026-08-09
docs: document the wasm-component-ld prerequisite for building fixtures

jtakakura:
## Feature or Problem

Building the wasm test fixtures locally fails with a cryptic linker error:

```
error: failed to parse core wasm for componentization
    invalid leading byte (0x2) for import name
```

The cause: rustc's bundled `wasm-component-ld` is too old to decode the
`implements` imports that wit-bindgen now emits. CI already works around this
by installing the standalone linker over the bundled one
(`.github/actions/setup-rust/action.yml`, "Drop once a Rust release bundles
wasm-component-ld >= 0.5.23"), but CONTRIBUTING.md never mentions it, so every
local contributor rediscovers it the hard way.

This documents the prerequisite with the same fix CI uses, including the error
message so it is searchable.

A possible follow-up: `cargo xtask build-fixtures` could check the effective
linker version up front and fail with this guidance. Happy to do that in a
separate PR if wanted.

## Release Information

next

## Testing

Docs only. Commands verified on macOS and Linux (both hit the original
failure; both fixed by the documented steps).


CONTRIBUTING.md
```diff
@@ -35,6 +36,21 @@ a fixture under `crates/wash-runtime/tests/fixtures/`):
 cargo xtask build-fixtures
 ```
 
+Linking the fixtures needs `wasm-component-ld` 0.5.23 or newer. Rust
```
ricochet: We have this doc'd in our readme currently: https://github.com/wasmCloud/wasmCloud/blob/a10c820eda8b3f81abc6a04e651cd0a2f9f80005/README.md?plain=1#L222-L236

The env var approach is nice since it doesn't require modifying rustlib.

jtakakura: Missed that note entirely, thanks for the pointer. Reworked: CONTRIBUTING now just links to that README section (the path I actually followed when I hit this), and the warning gains the literal error message so it is searchable. If you'd rather keep the README as the single home for this, happy to close instead.

ricochet (approved):

PR #5316, opened 2026-07-02
Update to wasi:webgpu 0.3.0-rc.2

MendyBerger:
wasi:webgpu 0.3.0-rc.2 uses async/p3

Got rid of wgpu and wstd since they don't support async yet

Issue #5030, opened 2026-04-07
CI: Add publishing for wash-runtime crate


ricochet:
We want our crate to be consumable by others for building their own hosts.

ricochet:
This is blocked on published releases of gfx-runtime cc @MendyBerger

MendyBerger:
Resolved in #5316

jtakakura:
@ricochet #5316 landed on 2026-07-17 and there are no git dependencies left in the workspace, so the original blocker looks resolved. I tried to verify that and ran into a separate problem. Apologies if this is already known — I did not see it mentioned in the thread, so I am leaving it here as a note.

`cargo package -p wash-runtime` packages fine (122 files, 489 KiB compressed), but the verification build of the packaged crate fails:

```
error: failed to run custom build command for `wash-runtime v2.6.1`
  thread 'main' panicked at build.rs:39:10:
  failed to read proto dir: Os { code: 2, kind: NotFound }
error: failed to verify package tarball
```

`build.rs` reads `<workspace_root>/proto/wasmcloud/runtime/v2`, which sits outside the crate directory, so the `include` allowlist in `crates/wash-runtime/Cargo.toml` cannot cover it. Two approaches I can think of:

- move `proto/` under `crates/wash-runtime/` and add it to `include`
- commit the generated descriptor and bindings, and make `build.rs` skip generation when `proto/` is absent

Both change the build layout, so this needs a decision from the maintainers.

Two smaller notes:

- When the publish step is added, crates.io Trusted Publishing (OIDC, `rust-lang/crates-io-auth-action`) would avoid storing a long-lived registry token.
- `proc-macro-error3 v3.0.3` in `Cargo.lock` is yanked on crates.io.

One question: `ci: backout publishing wash-runtime` (2026-05-01) reverted an earlier attempt. Was that only about the gfx dependency, or was this proto issue part of it as well?

For context on why I looked into this: I want to build a host plugin outside this repository, which needs `wash-runtime` from crates.io.

jtakakura:
Following up on my earlier comment with what I found in the history, and a PR.

The backout commit (5eee13133) answers my own question: it was only about the
unpublished wasi-gfx git dependencies — "We can't publish wash-runtime until
wasi-gfx deps are published." — and that blocker is resolved by #5316. The
proto packaging failure above was the one remaining blocker.

Two useful discoveries:

- The previous release workflow (`wash-runtime.yml`, including crates.io
[...]

PR #5446, opened 2026-08-09
fix(wash-runtime): move proto codegen to xtask so cargo package passes

jtakakura:
## Feature or Problem

`cargo package -p wash-runtime` fails during verification: `build.rs` reads
`<workspace>/proto/wasmcloud/runtime/v2`, which sits outside the crate
directory, so the `include` allowlist can never ship it:

```
thread 'main' panicked at build.rs:39:10:
failed to read proto dir: Os { code: 2, kind: NotFound }
error: failed to verify package tarball
```

This is the remaining packaging blocker for #5030.

### What changed

- Proto codegen moves from `build.rs` to `cargo xtask generate-protos`
  (`tonic-prost-build` / `pbjson-build` move to xtask's dependencies), and the
  generated bindings are committed under
  `crates/wash-runtime/src/washlet/generated/` with `@generated` markers.
  `build.rs` is deleted.
- A new `proto-drift` CI job regenerates the bindings and fails on any
  difference (including untracked files), wired into `gate`.
- `.gitattributes` marks the committed output `linguist-generated` so review
  diffs stay collapsed.

### Why committed bindings instead of moving `proto/` into the crate

`proto/` is shared with the Go side: `buf.gen.yaml` generates the
runtime-operator bindings from the same files, and those generated `*.pb.go`
are already committed. Committing the Rust output mirrors the existing
practice and keeps a single proto source of truth.

Side effects worth having: consumers of the crate no longer need `protoc`,
`wash-runtime` sheds its `[build-dependencies]`, and fresh builds skip proto
codegen entirely.

## Related Issues

#5030 — first of two steps. This PR only makes the crate packageable; it does
**not** restore the publish workflow or drop `publish = false`. That is a
follow-up `git revert 5eee13133` once the crates.io Trusted Publisher for
`wash-runtime` is configured (reverting before the config exists would turn
the next release-train tag red). The wasi-gfx blocker that motivated the
backout is resolved (#5316); no git dependencies remain in the workspace.

## Follow-ups (intentionally not in this PR)

With `build.rs` gone, nothing in the workspace invokes `protoc` at build time
anymore — only the `proto-drift` job needs it for regeneration. Cleanups that
follow from that, kept out of here to keep the diff reviewable:

- Drop `Setup protoc` from workflow jobs that no longer need it (`check`,
  `codeql`, ...) and update the codeql comment that still cites `build.rs`.
- Bench-host provisioning (`scripts/bench/ansible/provision.yml` and its
  README) installs `protobuf-compiler` / `libprotobuf-dev` citing
  `wash-runtime/build.rs`; those packages and comments can go.
- The cargo-machete ignore entries for `pbjson` / `pbjson-types` / `serde` in
  the workspace `Cargo.toml` existed because their uses were hidden in
  `OUT_DIR`; with the bindings committed, machete can likely see the uses and
  the entries can be dropped.

## Release Information

next

## Consumer Impact

None at runtime — the generated code is token-identical to what `build.rs`
produced. Measurements against `main`:

| | before | after |
|---|---|---|
| `cargo package` verification | fails | passes |
| tarball files | 122 | 123 (−`build.rs`, +2 generated) |
| tarball compressed | 488.6 KiB | 502.3 KiB (+2.8%) |
| release `wash` binary | 88,625,852 B | 88,626,236 B (+384 B, noise) |

### Unit Test(s)

Full `wash-runtime` suite green locally (57 wasm fixtures, 0 failed) and the
`check` job green on ubuntu/macos/windows in CI.

### Acceptance or Integration

- `cargo package -p wash-runtime` completes verification (previously failed
  at the build.rs step).
- `proto-drift` validated by running the same steps locally: regenerating
  twice produces byte-identical output; the job correctly caught a real drift
  during development.

### Manual Verification

`cargo xtask generate-protos` is idempotent and needs `protoc` only for
regeneration, not for building the crate.


ricochet (approved):
Thank you for doing this! I moved the gen check into proto.yml. The audit is failing on main so it should not block merging.

PR #5451, opened 2026-08-11
feat(wash-runtime): timeout hardening for guest code that spins CPU forever

Aditya1404Sal:
## The problem

When a wasm guest is executing compiled wasm and never yields, it is unreachable by every teardown the host owns.

Not "hard to reach" — unreachable by construction. The guest's poll never returns, so nothing else on that task is ever polled again, and every remedy we have is a future on exactly that task: `tokio::time::timeout`, `AbortOnDrop`, and pool retirement's drain (which waits for the call's task to end — the one thing that will never happen). No timeout value, however small, ever fires. The call runs until the process dies.

**This PR does not touch host-parked calls.** A guest awaiting something the host owns — a socket that never answers, a long clock wait — has yielded: its future is suspended, the executor has its thread back, and the existing timeouts tear it down correctly today. That path is already covered (`integration_http_call_timeout`, the `/wedge` fixture) and is unchanged here. The gap is only the non-yielding case.

It is also worse than one stuck request. A guest pinned to a worker thread can wedge the tokio worker holding the **time driver**, at which point no timer anywhere in the process fires — so a single non-yielding guest can take down the protection the parked case relies on, along with every other timeout in the host.

## Confirming the broken behaviour

Two experiments on `main`:

**A guest that loops on purpose.** A fixture route that spins with no host calls ran **232 seconds pinned at 100% CPU** and was still going when we killed the experiment. Nothing in the runtime was ever going to stop it; the client had long since given up, and the instance was lost for the life of the process.

**A guest driven into a boundless loop by its input** — the realistic shape, where nothing in the component looks wrong. `/redos?n=N` matches `a…a!` against `^(a+)+$` with the naive backtracking a JS or PCRE engine uses. It makes no host calls, so it never yields. At the default `n = 50` that is `2^49 ≈ 5.6 × 10^14` backtracks: even at an optimistic 10⁹ steps/second it would need **~6.5 days of continuous CPU**, and realistically weeks. It never returns.

The attacker controls only a query parameter. On `main` one such request permanently converts a core into a heater and wedges its instance — and with `max_concurrency: 4`, three further requests queue onto a run loop that is stuck mid-poll and will never read them, so they hang too.

### The ingress surfaces

Work enters a wasm store from every place listed below, and each one has a dispatcher that lives **outside** the store it is sending into — a healthy task, whose timer still fires when the callee is pinned. That is the only leverage available, so every one of them is wired:

| # | Ingress | Deadline |
|---|---|---|
| 1 | HTTP → service store | `http_response` |
| 2 | HTTP → pooled instance | `http_response` |
| 3 | HTTP → cold one-shot store (p3) | `http_response` |
| 3b | HTTP → cold one-shot store (p2) | `http_response` |
| 4 | Linked call → ephemeral store (plain + relocated) | `ephemeral_call` |
| 5 | Linked call → pooled instance | `ephemeral_call` |
| 6 | Linked call → shared store | same store — covered transitively by its own ingress |
| 7 | Messaging delivery → service store | `messaging_deliver` *(new)* |
| 7b | Messaging delivery → per-message store | `messaging_deliver` *(new)* |
| 8 | Capability call → plugin store | `plugin_capability_call` *(new)* |
| 8b | Lifecycle call → plugin store | `plugin_lifecycle_call` |

`7`, `7b` and `8` had no bound at all before this PR. `7b` matters beyond the stuck call itself: a guest spinning on a poison message would hold its store's pooling-allocator instance slot forever, and redelivery would repeat that until instantiation failed host-wide.

### Abandonment

Each dispatcher creates a `DispatchedCall` (`engine/abandon.rs`) holding the deadline and a flag, and the flag travels with the job into the store:

```rust
let call = DispatchedCall::new("HTTP (service)", timeouts::http_response());
sender.send(ServiceHttpJob { req, resp_tx, abandoned: call.flag() }).await?;
call.await_reply(resp_rx).await          // waited out here, not in the store
```

The flag is **armed** when the dispatcher stops wanting the result — the deadline passes, or the dispatcher's future is dropped outright (a client disconnect). Inside the store, the task serving the call registers that flag on the store's `AbandonedCalls` for exactly as long as the call runs, deregistering on drop however the call ends. Waiters that cannot await a reply (per-message deliveries, lifecycle replay, post-reply stream drains) arm through a timer instead, held as a cancel-on-drop handle bound to the work it bounds — healthy traffic accumulates no timers.

That leaves one reader: the wasmtime **epoch deadline callback**, compiled into the guest's own loop back-edges — the only host code a non-yielding guest still executes. The deadline is re-armed at the start of every call, so its countdown measures that call's own execution rather than wall time since the store was built.

The callback traps a store only when **three things hold together**:

1. **A call on it is abandoned and still registered** past the grace (`WASH_ABANDONED_CALL_GRACE_SECS`, 10s): the host has said, from outside, that nobody wants this work any more.
2. **Every registered call on the store is abandoned.** Trapping ends the whole store, so it waits until it would end nothing still wanted. This delays teardown but never loses it: a pinned guest admits no new calls, and the calls already on it are abandoned by their own deadlines in turn.
3. **The guest is actually pinned.** `Continuity` shows guest code executing without a pause for at least the grace: the callback fires only while guest code runs, so fires land about one sampling interval apart while a guest runs unbroken, and a wider gap (`WASH_ABANDONED_CALL_PAUSE_THRESHOLD_MS`, default 5× the sampling interval) means it stopped to wait for something, which resets the stretch.

Condition 3 is measured from the callback's own fire pattern rather than a host-side liveness stamp, because a stamp measurably cannot work here: the callback forces a yield on every fire (see below), and that yield returns the store's driving future to the executor, refreshing any host-side signal — a fully pinned guest looks continuously healthy from the host side.

Together the three make abandonment safe to arm liberally. A client disconnect arms the flag — deliberately, since if the guest behind it is wedged this flag is the only thing that will ever end it — but arming is harmless: a slow guest that still yields fails condition 3 and finishes untouched, and a healthy chatty guest (a stream ticking faster than the pause threshold, indistinguishable from pinned by fire pattern alone) fails condition 2, because its own call is still wanted.

What acting means depends on what the store is:

| Store | Policy | Outcome |
|---|---|---|
| Ephemeral / pooled instance | `Trap` | Store trapped, pool reaps it, next call gets a fresh instance |
| Service singleton | `Trap` | Supervised restart — better than carrying a wedged call forever |
| Host component plugin | `WarnThenTrap` | Warn once all three conditions hold at the grace; trap only once the pinned stretch reaches `WASH_ABANDONED_CALL_ESCALATION_SECS` (60s). A yielding abandoned call gets the runway to finish harmlessly; a non-yielding one already blocks every tenant's guest execution, so the restart + bind replay is what restores service. The warning re-arms once the store returns to nothing-abandoned, so every episode is logged |

The callback always **yields** rather than continuing, every ~100ms of unbroken guest execution. That is what closes the time-driver hole: a busy guest periodically hands its worker back, so it can never silence the deadline that would abandon it.

The flag is a **required field on every job type** and can only be minted by `DispatchedCall`, so a new ingress cannot compile without choosing a deadline and threading the flag. Streaming HTTP responses stay condemned-on-drop until the body completes; relocated linked calls re-bound their post-reply stream drain.

### How this composes with per-task cancellation (wasmtime#11833)

The one thing wasmtime cannot do today is cancel a single guest `call_concurrent` subtask. That is why every outcome above is store-scoped: with no way to kill one call, ending it means condemning its store and whatever shares it. bytecodealliance/wasmtime#11833 is the upstream API that fixes exactly that.

**It will not replace this work — it plugs into it.** The framework has three layers, and #11833 only touches the last:

- **Deciding a call is unwanted** stays. Per-ingress deadlines, disconnect detection, the grace period, the registration — none of that comes from wasmtime. A `cancel(task)` API tells you *how* to stop a call, never *which* call or *when*; this PR is what answers that.
- **Reaching a non-yielding guest** stays. Cancellation is still cooperative at the runtime level: a guest spinning in compiled code can only be interrupted where the runtime regains control, which is the epoch check in its back-edges. Per-task cancellation of a pinned task rides the same mechanism underneath. The always-yield fairness fix is independent of cancellation entirely.
- **The blast radius shrinks.** Only `AbandonedCallPolicy` changes: `Trap` and `WarnThenTrap` become "cancel this one task", and stores stop being condemned over one bad call — which also retires the all-abandoned wait. That is a single enforcement point, deliberately kept in one place so adopting #11833 does not mean re-touching the ingresses.

So the timeout-based design here is the layer that survives; #11833 makes its response surgical.

### Fixtures

Non-yielding branches added to four existing fixtures, each reachable by input so a test can wedge a specific ingress on demand. All of them loop with no host calls, which is what makes them unreachable by host timeouts:

- **`http-sleeper`** — `/spin` loops forever; `/spin?ms=N` loops for N ms and then answers normally (so a test can tell "trapped" from "ran hot and was left alone", since only an untrapped store lives to reply); `/redos?n=N` is the boundless-regex case above, wedged by input rather than by intent. For the protective tests it also gained `/slow?ms=N` (sleeps N ms, yielding the whole time) and `/sse?frames=N&gap_ms=M` (answers at once, then streams N frames M ms apart).
- **`ephemeral-callee-p3`** — `run(0)` wedges by contract, reached through `ephemeral-caller-p3`'s new `/spin` route. A magic argument rather than a new export, so the WIT vendored into both fixtures is untouched.
- **`msg-counter`** — a `spin` subject wedges the messaging handler, alongside the existing `boom` (trap) subject.
- **`kv-plugin`** — a `__spin__` key wedges a capability call inside the shared plugin store.

### `integration_abandoned_calls.rs`

One test per ingress, each asserting the same two things: the dispatcher returns an error within roughly its deadline (on `main` these hang until the harness kills the run), and service is restored afterwards.

- **`a_spinning_pooled_instance_is_trapped_and_replaced`** — warms one instance, hits `/spin`, and asserts the request ends within bound; then reads the fixture's own `served` counter, which lives in the instance's linear memory. A survivor would report `3`; a fresh instance reports `1`, which is the trap-and-reap observed from inside the guest.
- **`a_spinning_service_is_trapped_and_restarted`** — first proves a bounded spin that beats the deadline is *never touched* (`served` climbs 1→2, so running hot is not itself an offence), then wedges the singleton and asserts the supervisor brings back a fresh instance counting from one. A bystander workload confirms the trap stayed in its own store.
- **`a_component_wedged_by_its_input_is_trapped_and_replaced`** — the ReDoS case, with four concurrent requests to fill `max_concurrency` on the wedged instance. All four must fail (without the trap, the three queued behind hang until their clients give up), and the component must recover on a fresh instance.
- **`an_abandoned_linked_call_to_a_cold_ephemeral_store_is_ended`** — the wedge is two components deep: HTTP reaches the caller, the caller's imported `run(0)` wedges the callee in a store built for that one call. Asserts the caller's request fails and the whole linked path round-trips again afterwards.
- **`an_abandoned_linked_call_to_a_pooled_instance_is_ended`** — same, but the callee is warm. The callee's own `calls` counter climbing 1→2 and then restarting at 1 is the reap, observed from inside the callee.
- **`an_abandoned_message_delivery_traps_and_restarts_the_service`** — delivers through the host's messaging API; a `spin` subject must make the delivery error at its deadline instead of hanging, and the restarted handler must count from one.
- **`an_abandoned_capability_call_escalates_to_a_plugin_restart`** — seeds state in the shared plugin store, wedges it with `__spin__`, asserts the caller fails, then asserts the rebuilt plugin serves again *and* that the seeded key is gone — proving it really is a fresh incarnation rather than a survivor.

Three tests pin down what must **never** be touched — one per trap condition, and each fails if its condition is removed:

- **`a_client_that_gives_up_does_not_disturb_a_healthy_slow_guest`** — a client walks away 300ms into a request the guest legitimately spends far longer than deadline + grace serving. The disconnect arms the flag; the guest yields throughout, so the same instance must keep serving bystanders unbroken for the whole window.
- **`an_sse_client_disconnecting_between_frames_does_not_trap_the_service`** — the streaming form: head at once, frames further apart than deadline + grace, client takes the head and closes the tab. The guest only learns at its next frame write, so it is mid-sleep for the entire window a time-only trap would use — only its pauses distinguish it from a wedged guest.
- **`an_abandoned_co_tenant_does_not_trap_a_chatty_healthy_stream`** — a connected client reads a stream whose frames land *closer together* than the pause threshold, so its epoch fires read as one pinned stretch, while a co-tenant's abandoned call passes its grace. The stream's own still-wanted call is what must keep the store alive; every frame arriving is the assertion.
- **`a_healthy_instance_under_steady_traffic_is_never_trapped`** — the counterweight. Several deadline-periods of ordinary traffic with `served` required to climb unbroken. Nothing here is abandoned, so nothing may be touched; this is what would catch a design that guessed at "too busy" instead of waiting to be told.

Unit tests in `engine/abandon.rs` cover the semantics directly: a deregistered flag is invisible (so arming on every disconnect is safe), the grace is measured from first arming and re-arming cannot push it forward, a continuity stretch accumulates across sub-threshold gaps and resets on a pause, one still-wanted call blocks `all_abandoned`, and `await_reply` arms on deadline **and** on being dropped unpolled but never on delivery.

Full suite green; the cooperative `wasmcloud:host/cancel` tests are untouched and passing.

## Performance

Criterion, 30s runs, this branch vs `main`:

| Path | main | this PR | verdict |
|---|---|---|---|
| p2 request (hot) | 222 µs | 223 µs | no change (p = 0.29) |
| p3 request (hot) | ~203 µs | 177 µs | no change (p = 0.20) |
| p3 pooled request, paced (200ms idle gaps) | 591 µs | 590 µs | no change |
| guest loop epoch checks | — | ~0.03 ns/iter | interleaved A/B harness |

The paced row is the shape hot loops cannot produce: a pooled instance whose store sits idle between requests (the gap is slept off-clock; only the request is timed). Without the per-call re-arm, an idle store is past its deadline before its next call starts and pays the callback on that call's first back-edge — instrumentation confirms the callback fires once per idle return in that configuration — but the measured cost sits below the noise floor either way; the re-arm removes those fires entirely.

The first cut yielded on every 10ms tick and that *was* measurable — ~+20 µs on p3, from the ~2% of requests that cross a tick mid-guest and eat a fiber requeue. At 100ms granularity the cost is statistically indistinguishable from a runtime without epoch interruption at all, while a pinned guest still frees its worker every ≤100ms.

## Non-goals and known limits

- `wasmcloud:host/cancel` is unchanged — cooperative, guest-facing cancellation stays as is.
- Host-parked calls are unchanged; they were already torn down correctly.
- Teardown is store-scoped until wasmtime#11833. The wanted-call condition bounds the collateral — a store with any live call nobody has given up on is never trapped — at the price of delay: a wedged store with healthy co-tenant calls is only trapped once those calls are abandoned by their own deadlines. A plugin escalation still costs the plugin's in-memory state.
- Pinned-ness is sampled from the epoch callback's fire pattern on wall clock. On a host loaded enough that a pinned guest's fires land further apart than the pause threshold, the gaps read as pauses and the trap is postponed until the load lifts; `WASH_ABANDONED_CALL_PAUSE_THRESHOLD_MS` is the operational knob for that regime.
- A guest that spins *after* delivering its response has no outstanding call to abandon, so it is only reclaimed when the next call lands on its store.

### DIFF
--- a/crates/wash-runtime/src/engine/abandon.rs
+++ b/crates/wash-runtime/src/engine/abandon.rs
@@ -1,0 +1,419 @@
+//! Teardown for guest work whose caller has given up on it — even when the
+//! guest never yields.
+//!
+//! Every other teardown the host owns (`tokio::time::timeout`, task aborts,
+//! pool retirement) is a future on the very task a non-yielding guest is
+//! blocking, so none of them can fire against the one failure mode that
+//! matters most: a guest spinning in compiled code. The pieces here close that
+//! hole, split across the only two places that can act:
+//!
+//! - **Outside the store**, a dispatcher hands its call a [`DispatchedCall`]:
+//!   the deadline, and a flag ([`AbandonFlag`]) it arms once the result is no
+//!   longer wanted — the deadline passed, or whoever asked went away.
+//! - **Inside the store**, the task serving the call registers that flag with
+//!   the store's [`AbandonedCalls`] for exactly as long as the call runs, and
+//!   [`arm_epoch_deadline`] wires wasmtime's epoch callback — compiled into
+//!   the guest's own loop back-edges, the only host code a non-yielding guest
+//!   cannot block — to read it.
+//!
+//! The callback acts only on a call that has stayed abandoned *and still
+//! registered* for longer than [`crate::timeouts::abandoned_call_grace`]. The
+//! grace is what makes abandonment safe to signal eagerly: a healthy guest
+//! finishes the abandoned call and deregisters it well inside the grace, so a
+//! client disconnect costs nothing, while a wedged call is still there when
+//! the grace runs out. Without it, every mid-call disconnect would condemn a
+//! store that was serving other callers perfectly well.
+//!
+//! No judgement of the guest is made anywhere here. The epoch advances on wall
+//! clock, so it can report whether guest code is running, never whether it is
+//! getting anywhere: a store pegged at 100% CPU serving calls its callers
+//! still want is left alone. The one decision — "nobody wants this any more" —
+//! is the dispatcher's, and it is one the host already makes today by dropping
+//! futures; this module only makes it stick against a guest that never yields.
+//!
+//! What this cannot see: a guest burning CPU with *no* call outstanding
+//! against it — one that delivered its response and then kept spinning.
+//! Abandonment is keyed on calls, and there is no call left to abandon.
+
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Arc, LazyLock, Mutex, PoisonError};
+use std::time::{Duration, Instant};
+
+use crate::engine::ctx::SharedCtx;
+
+/// Milliseconds since the first flag was minted, biased by +1 so that `0` can
+/// mean "not armed" inside an [`AbandonFlag`].
+fn now_millis() -> u64 {
+    static ANCHOR: LazyLock<Instant> = LazyLock::new(Instant::now);
+    u64::try_from(ANCHOR.elapsed().as_millis()).unwrap_or(u64::MAX - 1) + 1
+}
+
+/// When a call's dispatcher stopped wanting its result, or nothing if it still
+/// does. Minted only by [`DispatchedCall`], which is what makes a job type's
+/// `abandoned` field proof that its dispatcher actually enforces a deadline.
+pub struct AbandonFlag(AtomicU64);
+
+impl AbandonFlag {
+    fn new() -> Arc<Self> {
+        Arc::new(Self(AtomicU64::new(0)))
+    }
+
+    /// Record the abandonment instant; only the first arming counts.
+    pub(crate) fn arm(&self) {
+        let _ = self
+            .0
+            .compare_exchange(0, now_millis(), Ordering::Relaxed, Ordering::Relaxed);
+    }
+
+    /// Arm this flag when `deadline` elapses, from a detached timer. For
+    /// re-bounding a call that outlives its reply (a post-reply stream drain),
+    /// and for waiters that themselves live inside the dispatched-to store. A
+    /// call already deregistered by then makes the arming invisible.
+    pub(crate) fn arm_after(self: Arc<Self>, deadline: Duration) {
+        tokio::spawn(async move {
+            tokio::time::sleep(deadline).await;
+            self.arm();
+        });
+    }
+
+    fn armed_at(&self) -> Option<u64> {
+        match self.0.load(Ordering::Relaxed) {
+            0 => None,
+            at => Some(at),
+        }
+    }
+}
+
+/// The in-flight calls on one store, so the store's epoch callback can see
+/// which of them have been abandoned. One per store, on [`SharedCtx`].
+#[derive(Default)]
+pub struct AbandonedCalls(Mutex<Vec<Arc<AbandonFlag>>>);
+
+impl AbandonedCalls {
+    /// Watch `flag` until the returned guard drops — held by the task serving
+    /// the call, for exactly as long as the call runs.
+    pub fn watch(self: &Arc<Self>, flag: Arc<AbandonFlag>) -> AbandonedGuard {
+        self.lock().push(Arc::clone(&flag));
+        AbandonedGuard {
+            calls: Arc::clone(self),
+            flag,
+        }
+    }
+
+    /// Whether any watched call has been abandoned for longer than `grace`.
+    fn any_abandoned_longer_than(&self, grace: Duration) -> bool {
+        let list = self.lock();
+        // `now` is only worth computing once a flag is actually armed; the
+        // common case (none are) stays free of clock reads.
+        let mut now = None;
+        let grace = u64::try_from(grace.as_millis()).unwrap_or(u64::MAX);
+        list.iter().any(|f| match f.armed_at() {
+            None => false,
+            Some(at) => {
+                let now = *now.get_or_insert_with(now_millis);
+                now.saturating_sub(at) >= grace
+            }
+        })
+    }
+
+    fn lock(&self) -> std::sync::MutexGuard<'_, Vec<Arc<AbandonFlag>>> {
+        self.0.lock().unwrap_or_else(PoisonError::into_inner)
+    }
+}
+
+/// Stops watching a call's flag when dropped, so a completed call leaves
+/// nothing behind — and so an abandoned flag whose call has ended is invisible
+/// to the epoch callback, which is what makes arming safe on every path.
+pub struct AbandonedGuard {
+    calls: Arc<AbandonedCalls>,
+    flag: Arc<AbandonFlag>,
+}
+
+impl Drop for AbandonedGuard {
+    fn drop(&mut self) {
+        self.calls.lock().retain(|f| !Arc::ptr_eq(f, &self.flag));
+    }
+}
+
+/// Arms a dispatched call's flag when dropped, unless the reply arrived first.
+///
+/// Drop-armed rather than only timeout-armed so that a dispatcher future
+/// cancelled outright — a client disconnect drops it without running any of
+/// its code — counts the same as its deadline passing.
+pub(crate) struct AbandonOnDrop {
+    flag: Arc<AbandonFlag>,
+    arm_on_drop: bool,
+}
+
+impl AbandonOnDrop {
+    fn new(flag: Arc<AbandonFlag>) -> Self {
+        Self {
+            flag,
+            arm_on_drop: true,
+        }
+    }
+
+    /// The reply arrived and is wanted; leave the call alone.
+    pub(crate) fn disarm(mut self) {
+        self.arm_on_drop = false;
+    }
+
+    fn flag(&self) -> Arc<AbandonFlag> {
+        Arc::clone(&self.flag)
+    }
+}
+
+impl Drop for AbandonOnDrop {
+    fn drop(&mut self) {
+        if self.arm_on_drop {
+            self.flag.arm();
+        }
+    }
+}
+
+/// One call handed to a store the dispatcher does not drive: the deadline the
+/// dispatcher will enforce, and the flag the store-side task must register.
+///
+/// This is the only source of [`AbandonFlag`]s, and every job type carries one
+/// as a required field — so a new ingress cannot be wired up without deciding
+/// its deadline, and cannot compile with the waiting accidentally placed
+/// inside the store it is waiting on.
+///
+/// Owns an [`AbandonOnDrop`] rather than a bare flag so that dropping it on
+/// *any* path — including an `await_reply` future cancelled before its first
+/// poll, after the job was already sent — abandons the call rather than
+/// leaving the store to serve it forever.
+pub(crate) struct DispatchedCall {
+    watch: AbandonOnDrop,
+    deadline: Duration,
+    /// Names the ingress in the timeout log line.
+    what: &'static str,
+}
+
+impl DispatchedCall {
+    pub(crate) fn new(what: &'static str, deadline: Duration) -> Self {
+        Self {
+            watch: AbandonOnDrop::new(AbandonFlag::new()),
+            deadline,
+            what,
+        }
+    }
+
+    /// The flag to carry in the job; the store-side task registers it with
+    /// [`AbandonedCalls::watch`] for the life of the call.
+    pub(crate) fn flag(&self) -> Arc<AbandonFlag> {
+        self.watch.flag()
+    }
+
+    /// Await the call's reply, abandoning the call if the deadline passes or
+    /// this future is dropped. `None` means no reply is coming.
+    pub(crate) async fn await_reply<F: Future>(self, reply: F) -> Option<F::Output> {
+        let (output, watch) = self.await_head(reply).await?;
+        watch.disarm();
+        Some(output)
+    }
+
+    /// For a dispatcher that must await from *inside* the store it dispatched
+    /// to (a lifecycle replay, whose serve loop shares the plugin store): its
+    /// own timeout can be starved by the very guest it bounds, so consume the
+    /// call and arm the flag from a detached timer at the deadline instead.
+    #[cfg_attr(not(feature = "host-component-plugins"), allow(dead_code))]
+    pub(crate) fn arm_detached(self) {
+        let Self {
+            watch, deadline, ..
+        } = self;
+        watch.flag().arm_after(deadline);
+        watch.disarm();
+    }
+
+    /// Like [`await_reply`], but the reply is only the *head* of the exchange:
+    /// on delivery the returned [`AbandonOnDrop`] keeps the call condemned-on-
+    /// drop, for a dispatcher that goes on streaming (an HTTP response body)
+    /// after the reply arrives.
+    ///
+    /// [`await_reply`]: Self::await_reply
+    pub(crate) async fn await_head<F: Future>(
+        self,
+        reply: F,
+    ) -> Option<(F::Output, AbandonOnDrop)> {
+        let Self {
+            watch,
+            deadline,
+            what,
+        } = self;
+        match tokio::time::timeout(deadline, reply).await {
+            Ok(output) => Some((output, watch)),
+            Err(_) => {
+                tracing::error!(
+                    ingress = what,
+                    timeout = ?deadline,
+                    "dispatched call produced no reply within its deadline; abandoning it so its \
+                     store can end it"
+                );
+                // `watch` drops here, arming the flag.
+                None
+            }
+        }
+    }
+}
+
+/// How many epoch ticks between deadline checks — how promptly an abandoned
+/// call is noticed once its grace runs out, and how often a busy guest yields
+/// its worker thread back to the executor.
+///
+/// Neither purpose wants this fine. Noticing an armed flag within ~100ms is
+/// ample against deadlines measured in seconds, and yielding every ~100ms of
+/// continuous guest execution is enough to keep a spinning guest from
+/// starving the executor (and tokio's time driver) for longer than that. What
+/// rules out a *finer* setting is the yield itself: it requeues the store's
+/// task through the executor, and paying that on every 10ms tick showed up as
+/// a measurable per-request cost on the p3 request path — ~2% of requests
+/// cross a tick mid-guest and eat a ~1ms reschedule.
+const EPOCH_DEADLINE_TICKS: u64 = 10;
+
+/// What a store does about a call abandoned past its grace; see
+/// [`arm_epoch_deadline`].
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AbandonedCallPolicy {
+    /// Trap the store at the grace. For a store that can be replaced: an
+    /// ephemeral store serves one call, a pooled instance is reaped and
+    /// rebuilt on the next call, and a trapped service is restarted by its
+    /// supervisor. Everything sharing the store goes down with it — the same
+    /// blast radius as a guest trap, bounded the same way.
+    Trap,
+    /// Warn at the grace, trap only at
+    /// [`crate::timeouts::abandoned_call_escalation`]. For a host component
+    /// plugin's store, which serves every tenant at once: the long runway lets
+    /// an abandoned call that still *yields* finish harmlessly, costing no one
+    /// anything. A call that never yields, though, holds the store's guest
+    /// execution for as long as it lives — no other tenant's call can enter —
+    /// so a store still carrying it at the escalation is already down for
+    /// everyone, and trapping it is what brings it back (the supervisor
+    /// rebuilds it and replays binds, the same path an organic plugin trap
+    /// takes). Per-task cancellation (bytecodealliance/wasmtime#11833) is what
+    /// would end the one call instead.
+    WarnThenTrap,
+}
+
+/// Let the host end guest work whose caller has abandoned it, even if the
+/// guest never yields.
+///
+/// Every store needs this call whether or not anything will ever abandon a
+/// call on it — a store that never sets a deadline traps the moment it runs
+/// any guest code, and the host cannot install one later.
+///
+/// The callback is the enforcement point because it is the only one left: a
+/// guest that never yields never returns from its poll, so every host future
+/// on that store — timeouts included — is stuck behind it. wasmtime compiles
+/// the deadline check into the guest's own loop back-edges, which is the one
+/// place a spinning guest cannot avoid. The module docs
+/// ([`crate::engine::abandon`]) cover the model.
+pub(crate) fn arm_epoch_deadline(
+    store: &mut wasmtime::Store<SharedCtx>,
+    policy: AbandonedCallPolicy,
+) {
+    let abandoned = Arc::clone(&store.data().abandoned);
+    let store_id = Arc::clone(&store.data().active_ctx.store_id);
+    let grace = crate::timeouts::abandoned_call_grace();
+    let escalation = crate::timeouts::abandoned_call_escalation();
+    let mut warned = false;
+
+    store.set_epoch_deadline(EPOCH_DEADLINE_TICKS);
+    store.epoch_deadline_callback(move |_| {
+        if !abandoned.any_abandoned_longer_than(grace) {
+            // Yield, never merely continue: this callback is the only point at
+            // which a non-yielding guest hands the host back its thread. Ridden
+            // straight through, the guest keeps the worker — and if that worker
+            // holds tokio's time driver, *no timer on the runtime fires*,
+            // including the very deadline whose expiry would abandon this call.
+            // The yield is what keeps the enforcement loop closed.
+            return Ok(wasmtime::UpdateDeadline::Yield(EPOCH_DEADLINE_TICKS));
+        }
+        if policy == AbandonedCallPolicy::WarnThenTrap
+            && !abandoned.any_abandoned_longer_than(escalation)
+        {
+            if !warned {
+                warned = true;
+                tracing::warn!(
+                    store_id = %store_id,
+                    escalation = ?escalation,
+                    "a call on this store was abandoned past its grace but its guest is still \
+                     running; the store is shared, so it is not trapped before the escalation"
+                );
+            }
+            return Ok(wasmtime::UpdateDeadline::Yield(EPOCH_DEADLINE_TICKS));
+        }
+        tracing::error!(
+            store_id = %store_id,
+            "a call on this store was abandoned but its guest is still running; trapping the store"
+        );
+        Ok(wasmtime::UpdateDeadline::Interrupt)
+    });
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// A flag armed and then deregistered must be invisible: this is what makes
+    /// arming safe on every completion path (disconnects included) rather than
+    /// only on timeouts.
+    #[test]
+    fn deregistered_flags_are_invisible() {
+        let calls = Arc::new(AbandonedCalls::default());
+        let call = DispatchedCall::new("test", Duration::ZERO);
+        let flag = call.flag();
+        // Keep `call` alive: dropping it would arm the flag itself.
+        let guard = calls.watch(Arc::clone(&flag));
+
+        assert!(!calls.any_abandoned_longer_than(Duration::ZERO));
+        flag.arm();
+        assert!(calls.any_abandoned_longer_than(Duration::ZERO));
+        drop(guard);
+        assert!(!calls.any_abandoned_longer_than(Duration::ZERO));
+    }
+
+    /// The grace holds the callback off a freshly abandoned call, and only the
+    /// first arming sets the clock.
+    #[test]
+    fn grace_is_measured_from_first_arming() {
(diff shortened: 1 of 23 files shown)


crates/wash-runtime/src/host/http.rs
```diff
@@ -1597,20 +1662,33 @@ async fn handle_http_request<T: Router>(
     let service_sender = service_handlers.read().await.get(&workload_id).cloned();
     if let Some(sender) = service_sender {
         let (resp_tx, resp_rx) = tokio::sync::oneshot::channel();
-        let response = if sender.send((req, resp_tx)).await.is_err() {
+        // The deadline is enforced here, outside the service's store: a guest
+        // that never yields blocks every host future on its own store, so only
+        // a waiter out here can abandon the call (see `crate::engine::abandon`).
+        let call = DispatchedCall::new("HTTP (service)", crate::timeouts::http_response());
```
ricochet, line 1702: This call is drop-armed, and hyper drops the dispatcher future when a client hangs up so an ordinary disconnect arms the flag. Combined with `AbandonedCallPolicy::Trap` on service stores (linked_call.rs:328) and the store-wide check in `any_abandoned_longer_than`, a client that gives up at t=2s on a request the guest legitimately takes 30s to serve traps the whole store at t=12s: every other concurrent request on that service gets a 500 and the supervisor restarts the workload.

The grace is doing double duty here. It's meant to distinguish "wedged" from "healthy", but it's actually bounding how long a request is allowed to take after its client leaves. The comment this replaces in `trigger_service/http.rs` notes that a disconnect used to cost nothing.

SSE / long-poll is the sharper case: the guest only learns of the disconnect at its next frame, which for a 30s heartbeat is always past the grace.

crates/wash-runtime/src/host/http.rs
```diff
@@ -904,16 +905,71 @@ impl HostHandler for NullServer {
...
+/// deadline enforcement of [`DispatchedCall::await_head`] across the body phase,
+/// so a guest that produces a response head and then wedges mid-body is still
+/// reachable.
+struct WatchedBody<B> {
+    inner: B,
+    /// `None` once end-of-stream disarmed it.
+    watch: Option<AbandonOnDrop>,
+}
+
+impl<B: hyper::body::Body + Unpin> hyper::body::Body for WatchedBody<B> {
```
ricochet, line 934: This impl forwards only `poll_frame`, so `is_end_stream()` and `size_hint()` fall back to the trait defaults and those are exactly the two methods hyper uses to frame a response. `MeteredBody` in this same file (http.rs:1875-1881) forwards both.

Concretely: an empty body (`HyperOutgoingBody::default()`, e.g. the 500 built at `trigger_service/http.rs:103`) reports `is_end_stream() == true` and exact size 0, but wrapped here hyper sees `BodyLength::Unknown`. A plain GET then goes out `Transfer-Encoding: chunked` instead of `Content-Length: 0`; for HEAD or 204/304, `Server::can_chunked` is false so hyper falls to `Encoder::close_delimited()` and tears down the connection where it previously kept it alive. Since `MeteredBody` wraps `WatchedBody` on the non-service paths, this propagates through the whole chain.

Should just be the two forwarding methods, mirroring `MeteredBody`.

crates/wash-runtime/src/host/http.rs
```diff
@@ -904,16 +905,71 @@ impl HostHandler for NullServer {
...
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Result<hyper::body::Frame<Self::Data>, Self::Error>>> {
+        let this = self.get_mut();
+        let poll = std::pin::Pin::new(&mut this.inner).poll_frame(cx);
+        // Clean end of stream: the response was delivered whole, so nothing is
+        // abandoned. An *error* frame deliberately does not disarm — the
+        // exchange broke off, and if the guest is wedged behind it, the armed
+        // flag is what ends it (a completed call deregisters, so arming a
+        // healthy one is harmless).
+        if let std::task::Poll::Ready(None) = poll
```
ricochet: Disarming only on a terminal `Poll::Ready(None)` means successful responses can arm the flag: hyper stops polling the body once the declared `content-length` bytes are written (`proto/h1/dispatch.rs`: `if !self.conn.can_write_body() { *clear_body = true; continue; }`), and for `content-length: 0` / 204 / HEAD it never polls the body at all. In those cases `AbandonOnDrop::drop` runs on a complete, correct response.

That's harmless if the guest has already finished so the guard is deregistered, as the docs note. But if the guest is still inside its handler (flushing telemetry, draining a stream) the guard is still registered, and the store traps `abandoned_call_grace` later on a response that was fully delivered.

Forwarding `is_end_stream()`/`size_hint()` (see above) narrows this but doesn't close it; the Content-Length case still ends without a terminal poll.

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -322,6 +323,9 @@ pub(crate) async fn new_store_from_templates(
     }
 
     let mut store = wasmtime::Store::new(engine, shared_ctx);
+    // Trap for every store built here, services included: trapping a service
+    // means a supervisor restart, which beats carrying a wedged call forever.
+    arm_epoch_deadline(&mut store, AbandonedCallPolicy::Trap);
```
ricochet, line 358: The blast radius of `Trap` on a service store is the whole singleton plus every concurrent request on it, but the condition that triggers it is only "a flag has been armed for `grace`". `any_abandoned_longer_than` (abandon.rs:104) compares `armed_at` to now and nothing else. There's no signal anywhere in `AbandonedCalls` that the guest is actually wedged rather than healthy-and-slow.

That gap is what makes the two HTTP paths above dangerous. Worth considering a second condition: the epoch callback fires only while guest code is running, and a guest awaiting a host future returns to the executor. If the task loop stamped a "last returned to the host" instant into `SharedCtx`, the trap could require both `now - armed_at >= grace` **and** `now - last_yield >= grace`. A slow-but-awaiting guest refreshes the stamp continuously and is never trapped, a spinning guest never does. That would also let the grace go back to meaning wedged-ness rather than a request budget.

crates/wash-runtime/src/engine/linked_call.rs
```diff
@@ -631,15 +646,24 @@ async fn invoke_ephemeral_relocated(
...
     // Results are in hand; the task must keep running to feed any result streams,
     // so detach it (cancellation past this point is handled by the result-stream
     // consumers closing their pump channels).
     std::mem::forget(task);
 
+    // The detached store bounds its own stream drain, but that bound is a
+    // future on the store and cannot fire against a guest that never yields.
+    // Re-abandon the call when the drain budget expires: a store already done
+    // by then has deregistered the flag, so this only ever ends a wedged drain.
+    drain_flag.arm_after(crate::timeouts::stream_drain());
```
ricochet: This sits on the unconditional success path of every relocated ephemeral call, and `arm_after` (abandon.rs:72) spawns a detached task that sleeps the full deadline, `stream_drain()`, 600s by default, holding an `Arc<AbandonFlag>`. Nothing cancels it when the drain actually finishes microseconds later.

At 1k relocated calls/sec that's roughly 600k live tasks and timer-wheel entries in steady state, growing until the first ones expire, on entirely healthy traffic. RSS and time-driver overhead climb with no wedged guest anywhere.

Holding the `JoinHandle` (or an `AbortOnDropHandle`) alongside the ephemeral store's lifetime so a completed drain kills its timer would fix it.

crates/wash-runtime/src/engine/abandon.rs
```diff
@@ -0,0 +1,419 @@
...
+    store: &mut wasmtime::Store<SharedCtx>,
+    policy: AbandonedCallPolicy,
+) {
+    let abandoned = Arc::clone(&store.data().abandoned);
+    let store_id = Arc::clone(&store.data().active_ctx.store_id);
+    let grace = crate::timeouts::abandoned_call_grace();
+    let escalation = crate::timeouts::abandoned_call_escalation();
+    let mut warned = false;
+
+    store.set_epoch_deadline(EPOCH_DEADLINE_TICKS);
```
ricochet, line 426: This is armed once at store construction and only re-armed from inside the callback, but the deadline resolves against `engine.current_epoch()` and the ticker advances every 10ms of wall clock regardless of whether any guest is running.

So a store idle for longer than 100ms is already past its deadline, and pays a callback plus a full `Yield` reschedule on the very first back-edge of its next call, before doing any useful work. The comment above budgets that reschedule at ~1ms and reasons about "~2% of requests cross a tick mid-guest" but for a service under ~10 rps it's closer to 100% of requests.

Re-arming at the start of each call, rather than only from inside the callback, would restore the intended behavior.

crates/wash-runtime/src/engine/abandon.rs
```diff
@@ -0,0 +1,419 @@
...
+            // Yield, never merely continue: this callback is the only point at
+            // which a non-yielding guest hands the host back its thread. Ridden
+            // straight through, the guest keeps the worker — and if that worker
+            // holds tokio's time driver, *no timer on the runtime fires*,
+            // including the very deadline whose expiry would abandon this call.
+            // The yield is what keeps the enforcement loop closed.
+            return Ok(wasmtime::UpdateDeadline::Yield(EPOCH_DEADLINE_TICKS));
+        }
+        if policy == AbandonedCallPolicy::WarnThenTrap
+            && !abandoned.any_abandoned_longer_than(escalation)
```
ricochet: This escalation check is also purely elapsed-time, which contradicts the rationale on `WarnThenTrap` itself: "the long runway lets an abandoned call that still *yields* finish harmlessly". Nothing here observes yielding, e.g. `any_abandoned_longer_than` only compares `armed_at` to now.

So a `workload-lifecycle` bind hook that provisions a remote resource in 90s while yielding perfectly normally (its dispatcher having timed out at `plugin_lifecycle_call()` and armed the flag via the consumed `DispatchedCall` at lifecycle.rs:313-319) traps the plugin store at grace+60s. On a host component plugin store that means every other tenant's in-flight capability call fails and all plugin in-memory state is destroyed.

Same underlying fix as the `Trap` case: this policy needs the yield signal.

crates/wash-runtime/src/engine/abandon.rs
```diff
@@ -0,0 +1,419 @@
...
+        {
+            if !warned {
```
ricochet, line 464: `warned` is declared outside the closure (line 319), so it's a latch for the store's entire lifetime rather than per-abandonment. Only the first abandoned call a plugin store ever sees produces this warning.

If that first one was armed past the grace and then completed normally inside the escalation window, every subsequent genuinely wedged call escalates silently so operators get only the terminal `error!` at the trap, with no preceding warning and nothing to correlate.

Resetting it when the abandoned set goes empty would be enough.

crates/wash-runtime/src/host/trigger_service/messaging.rs
```diff
@@ -49,8 +56,16 @@ impl AccessorTask<SharedCtx> for MessagingTask {
             func_idx,
             msg,
             result_tx,
+            abandoned,
         } = self;
 
+        // Watch this call for the rest of its life. The guard deregisters on
+        // drop, however the call ends.
+        let _abandoned = accessor.with(|mut access| {
```
ricochet, line 139: This threads the flag correctly for the trigger-service messaging path, but the *other* messaging ingress doesn't: `plugin/wasmcloud_messaging/nats.rs:415-467` and `in_memory.rs:450-480` take the per-message branch whenever `has_trigger_service_messaging()` is false, and that branch calls `workload.new_store()` (which arms the epoch deadline with `Trap`), then `tokio::spawn`s `call_handle_message` with no timeout and no `AbandonFlag`.

Nothing ever mints a `DispatchedCall` there, so `any_abandoned_longer_than` is permanently false and the epoch callback only ever yields. A guest spinning on a poison message pins the task, the store, and under the pooling allocator one `total_component_instances` slot forever; redelivery repeats it until instantiation fails host-wide with "maximum concurrent component instance limit reached" and no workload can start.

Aditya1404Sal:
update : will address the review today

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: resolved it in three ways. 
(1) Disarm no longer needs the terminal poll: after every poll, the wrapper also checks inner.is_end_stream(), so the poll that writes the last declared byte disarms right there (the Content-Length case you called out)
(2) A body that's already complete when the guest returns it is never wrapped at all (watch_body disarms and returns it bare), covering content-length: 0 / 204, which hyper never polls. 
(3) guest still in its handler after delivery, e.g. HEAD with a discarded body, can still arm, but arming is no longer fatal on its own: the trap also requires the guest to be pinned, and a guest flushing telemetry is awaiting.

crates/wash-runtime/src/host/http.rs
Aditya1404Sal: Done, now mirroring `MeteredBody`

crates/wash-runtime/src/engine/abandon.rs
```diff
@@ -0,0 +1,573 @@
...
+        }
+        // A call is still wanted: trapping the store would take it too. This
+        // also protects a yielding guest whose wakes land closer together
+        // than the pause threshold and so read as one pinned stretch — its
+        // own wanted call is what keeps its store alive.
+        if !abandoned.all_abandoned() {
+            return keep_running(&mut warned);
+        }
+        // Abandoned, but the guest is still pausing to wait: slow, not wedged.
+        if pinned_for < grace_millis {
```
ricochet: `Continuity` measures **wall clock**, not guest execution, so a healthy guest that awaits for less than `pause_threshold` at a time is indistinguishable from a pinned one.

The epoch deadline is 10 ticks x 10 ms = 100 ms of wall clock, and the epoch advances whether or not a guest is running. So a guest that computes briefly, awaits 50-300 ms, computes again, ... produces fires ~100 ms apart under the 500 ms threshold and `stretch_start` never resets.

Concrete shape: a p3 service issues a series of fast (<500 ms) outbound calls for 10+ s while serving one request. The client disconnects (or a proxy drops the connection), arming the flag via `AbandonOnDrop`. `all_abandoned()` is true because that was the only registered call, so the guard just above doesn't help. After `abandoned_call_grace` this returns `UpdateDeadline::Interrupt` and traps the whole service store, restarting the singleton.

`a_client_that_gives_up_does_not_disturb_a_healthy_slow_guest` doesn't cover this. `/slow` is a *single* long await, which is exactly the case the pause threshold does catch. Worth a test with a guest that awaits repeatedly in short hops.

The durable fix is to measure something that only advances while the guest runs (epoch-counter deltas taken across an explicit yield) rather than `Instant::now()` deltas.

crates/wash-runtime/src/engine/abandon.rs
```diff
@@ -0,0 +1,573 @@
...
+/// room for scheduler jitter while staying well below any real wait.
+const PAUSE_FACTOR: u32 = 5;
+
+/// How long a gap between fires counts as the guest having paused, derived
+/// from the sampling interval so it follows if that changes.
+///
+/// `WASH_ABANDONED_CALL_PAUSE_THRESHOLD_MS` overrides it, for a host loaded
+/// enough that a pinned guest's fires land further apart than this — which
+/// reads as a pause and stops anything from being trapped.
+pub(crate) fn pause_threshold() -> Duration {
```
ricochet: The other direction of the same problem: the derived 500 ms threshold doesn't account for the requeue latency **this callback itself introduces**, so on a loaded host a genuinely pinned guest is never trapped.

Every fire returns `UpdateDeadline::Yield(..)`, which pushes the store's task back through the executor. On a busy host the gap between fires is `100 ms + requeue latency`, which readily exceeds 500 ms. `Continuity::observe` then reads each gap as a pause, resets `stretch_start`, `pinned_for` stays near 0, and nothing is ever trapped: the feature silently no-ops precisely when it is needed.

The doc comment here already names this failure mode, and the test suite is evidence it's reachable. 568e0bc71 had to set `WASH_ABANDONED_CALL_PAUSE_THRESHOLD_MS=2000` because "a pinned guest can go unscheduled long enough for the default 500 ms pause threshold to read the gap as a yield." Production ships no such override, so on a loaded host the default is the untested value.

The root issue is that `Continuity::observe` treats a gap as binary: below the threshold, sum the whole gap into `pinned_for`; above it, discard the whole stretch to 0. Both branches are wrong; the first counts real off-CPU wait time as execution (see the other comment), the second wipes out a genuinely pinned guest's accumulated time on one slow requeue. A fuel-based signal would fix this too, but fuel checks are inserted at every loop backedge/call and are meaningfully more expensive than epoch checks, so I'd avoid it here.

A cheaper fix that stays on epoch interruption: cap each interval's contribution instead of summing the raw gap, and only hard-reset on a gap that's unambiguously a real pause (an order of magnitude bigger than any plausible scheduling delay), rather than at the same threshold used for capping:

```rust
/// Max wall-clock time one epoch window can represent as genuine guest
/// execution — what a fire-to-fire gap looks like while the guest runs
/// straight through.
const EXECUTION_WINDOW_MS: u64 = ...; // what pause_threshold's base is derived from today

/// Gap large enough that it can only be a real pause (host I/O, or the call
/// ended and a new one started later) — never scheduler jitter or a brief
/// await. Order of seconds, not proportional to EXECUTION_WINDOW_MS.
const RESET_THRESHOLD_MS: u64 = ...;

fn observe(&mut self, now: u64) -> u64 {
    let gap = self.last_fire.map_or(0, |last| now.saturating_sub(last));
    if gap > RESET_THRESHOLD_MS {
        self.pinned_millis = 0;
    } else {
        self.pinned_millis += gap.min(EXECUTION_WINDOW_MS);
    }
    self.last_fire = Some(now);
    self.pinned_millis
}
```

This fixes both directions at once: a legit await only ever contributes `EXECUTION_WINDOW_MS` regardless of its real length (fixes the false-positive in the sibling comment), and scheduler-jitter-inflated gaps under `RESET_THRESHOLD_MS` no longer wipe the accumulator, so a genuinely pinned guest keeps accumulating toward `grace`/`escalation` even when its fires land further apart under load. `PAUSE_FACTOR`, `pause_threshold()`, and the env override could go away entirely, replaced by two fixed invariants.

crates/wash-runtime/src/host/http.rs
```diff
@@ -918,16 +919,91 @@ impl HostHandler for NullServer {
...
+        let this = self.get_mut();
+        let poll = std::pin::Pin::new(&mut this.inner).poll_frame(cx);
+        // Clean end of stream: the response was delivered whole, so nothing is
+        // abandoned. An *error* frame deliberately does not disarm — the
+        // exchange broke off, and if the guest is wedged behind it, the armed
+        // flag is what ends it (a completed call deregisters, so arming a
+        // healthy one is harmless).
+        // `is_end_stream` too, since hyper stops polling once a declared
+        // `content-length` is written and never delivers that final `None`.
+        let ended = matches!(poll, std::task::Poll::Ready(None)) || this.inner.is_end_stream();
```
ricochet, line 951: This `is_end_stream()` fallback can never fire for the runtime's own response bodies, so the case the comment describes still arms the flag on a fully delivered response.

Both `ChannelBody` types the runtime produces (`host/http_p3.rs` and `host/trigger_service/http.rs`) implement only `poll_frame`; `is_end_stream()` uses the `Body` trait default and always returns `false`, and `boxed_unsync()` forwards that. So: a p3 guest sets `content-length`, hyper's `Length` encoder stops polling after writing exactly that many bytes, `ended` is `false`, `WatchedBody` is dropped with `watch: Some(..)`, and `AbandonOnDrop::drop` arms the flag for a request that succeeded.

That's harmless only while the store-side `HttpTask` has already deregistered. On a service store whose `io_fut` (request-body upload) is still running, the call is still registered and is now falsely marked abandoned.

Either override `is_end_stream` on both `ChannelBody`s (see the comment on `http_p3.rs`), or disarm on the encoder-completion path some other way.

crates/wash-runtime/src/host/http_p3.rs
```diff
@@ -39,10 +39,10 @@ pub type P3SendFuture = Box<dyn std::future::Future<Output = P3SendResult> + Sen
...
     rx: tokio::sync::mpsc::Receiver<Result<hyper::body::Frame<bytes::Bytes>, ErrorCode>>,
     /// Aborts the component task when this body is dropped before the stream
-    /// completes (e.g. the client disconnects). Without this, a guest that is
-    /// busy computing — rather than parked on a frame send — would keep running
-    /// until its next send; there is no epoch/wall-clock backstop. Held only
-    /// for its `Drop`.
+    /// completes (e.g. the client disconnects). The abort lands at the guest's
+    /// next await, reclaiming a yielding guest at once; a guest that never
+    /// yields is unreachable by it and is ended by its abandoned call instead
+    /// (see [`crate::engine::abandon`]). Held only for its `Drop`.
```
ricochet, line 45: Follow-on from the `http.rs` comment: this `ChannelBody` (and its sibling in `host/trigger_service/http.rs:23`) overrides neither `is_end_stream` nor `size_hint`, which is what makes the `WatchedBody` mitigation inert.

The channel's end-of-stream is knowable. The sender is dropped when the component task finishes so `is_end_stream` could report `true` once `rx` is closed and drained. Adding that here and in the trigger-service body is what would make the `content-length` fallback in `WatchedBody::poll_frame` actually do something.

ricochet (approved):

PR #5452, opened 2026-08-11
feat(wash-runtime): avoid trapping the plugin

ricochet:
Follow-up to #5442, which landed the plugin → workload call direction. That direction had a failure mode it could not survive, and this closes it.

## Why

A host component plugin's call out to a workload reaches *another workload's* guest, which can trap, stop mid-call, or run long. None of that is the plugin's fault, but all of it surfaced as a trap: the failure propagated out of the host shim, faulted the plugin's store shared by every workload it serves and the supervisor restarted it. Unattributed faults are charged to the restart budget, so a workload that failed every dispatch exhausted `DEFAULT_MAX_RESTARTS` in three calls and took the plugin down for every tenant.

A plugin could not defend itself. Counting strikes in guest memory does not work, because the fault being counted wipes that memory. The same property `integration_component_plugin_lifecycle` relies on to prove a plugin *didn't* restart.

## What changed

**Every workload-facing import must be able to carry a failure**, checked when the plugin loads. `route_workload_call` no longer has an `Err` exit: nothing to call, workload stopped, deployment replaced, interface not exported, callee trapped, call timed out, each writes an error value into the result slot and returns `Ok`.

Three error shapes qualify, in descending order of what the plugin learns:

1. `wasmcloud:host/types.call-error` — new variant, **recommended**. Distinguishes a host failure from one the workload meant, and says which failure it was. Matched structurally, so a plugin may `use` the host's definition or restate it.
2. `result<_, string>` — what `wasmcloud:messaging/handler` already declares.
3. **The interface's own error type**, so long as the host can construct one of its cases (a payload-less one, or one taking `string`/`option<string>`). This is what lets a custom interface with `wasi:sockets`-style `error-code` still be answerable. The host picks the case whose name means "unenumerated failure", preferring `internal-error`.

Shapes 2 and 3 cannot distinguish a host failure from one the workload meant, so where a case has room for text the detail is prefixed `wasmcloud:host: `. Where it has not, the reason is lost. This is why we doc a preference for `call-error` when you own the interface.

`call-error` is deliberately open-ended: WIT gives no way to add a case without breaking guests built against the old set, so `other` is where the host puts anything it grows into later.

**A `target` handle now binds a workload *and* one of its interfaces.** A workload need only export some of what a plugin imports, so pairing them makes the pairing checkable when it is made rather than when it is used: `open(id, interface)` answers `none` unless that workload serves that interface, and a handle that exists is proof the call routes. `not-exported` is no longer reachable through one — only on an unaddressed call, where a plugin calls back into whichever workload's capability call it happens to be serving. Handles nest per interface; one opened for `handler` does not redirect a `metrics` call.

## Consequences worth knowing

- **A workload-facing function can no longer return nothing.** A void function has nowhere to be told the call failed, so it is refused at load. Fire-and-forget signatures need a `result`.
- **`wasi:http/handler` still cannot be a workload-facing import**, and this does not change that. Its `request`/`response` are resources, which the bridge-safety check refuses; the plugin → workload direction has no resource bridge. Its *error* type would have been fine.
- This makes a dispatch-side poison quarantine unnecessary since there is no crash loop left to contain. The `on-workload-bind` quarantine (`POISON_EVICT_STRIKES`) is a separate path and is unchanged.

## Testing

`events-plugin` grows a second workload-facing interface with its own error type, and `events-caller` grows a workload whose export traps on demand.

- `test_a_trapping_workload_is_reported_not_fatal` dispatches to a trapping workload **four times** past the plugin's restart budget and asserts every call returns `failed`, that the plugin still serves afterwards, and that the workload stays callable.
- `test_a_failure_is_reported_through_an_interfaces_own_error_type` proves the borrowed shape end to end, including that `internal-error` is preferred over the case declared first.
- `test_a_target_for_an_unexported_interface_is_refused` pins that the check moved to open time.
- Unit tests cover which result shapes are accepted and rejected, and that a handle routes only its own interface.

### DIFF
--- a/crates/wash-runtime/src/plugin/component_host/mod.rs
+++ b/crates/wash-runtime/src/plugin/component_host/mod.rs
@@ -52,9 +52,12 @@
 //! provides, in which case a workload that exports it satisfies it — the
 //! arrangement `wasi:http` and `wasmcloud:messaging` already have, where the
 //! host both serves a workload's imports and calls the handler it exports.
-//! Such an interface must be `async func` throughout, because the shim the
-//! host installs for it is concurrent and a sync-typed import cannot bind to
-//! one.
+//! Such an interface must be `async func` throughout, because the shim the host
+//! installs for it is concurrent and a sync-typed import cannot bind to one,
+//! and every function of it must return a `result` the host can build an error
+//! into — a call out to another workload's guest can trap or stop mid-call, and
+//! the plugin's store is shared by every workload it serves, so the host
+//! answers with a value instead of faulting it.
 //! [`classify_workload_imports`] decides which imports those are, and the
 //! `workload_call` submodule holds the rest: the per-workload routes (claimed in
 //! `on_workload_resolved`, exactly as a native plugin claims them), the
@@ -1089,8 +1092,8 @@ fn declares_workload_calls(imports: &[ExportedInterface]) -> bool {
 /// and `handler` from a workload — the split the native messaging plugin itself
 /// has.
 ///
-/// An import no workload ever exports traps on the call that needs it, naming
-/// the interface.
+/// An import no workload ever exports is reported to the plugin as
+/// `not-exported` on the call that needs it, naming the interface.
 fn classify_workload_imports(
     id: &str,
     component: &Component,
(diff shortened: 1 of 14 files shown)


vados-cosmonic (dismissed):
LGTM 🚀 

Yeah this is pretty janky but it does seem like the only reasonable way to do it without disrupting things too much. Crashing the plugin is obviously not a great option and neither is requiring interfaces that want to be called to return some special error type...

The only other option I can think of is having a side-channel in workload-call (`errno()` vibes, essentially), but then that leaves the question of what the actual call would return (some sort of sentinel for every type? Golang vibes).

Basically the cleanest fix here is probably some sort of bindgen change that enables some sort of arbitrary result wrapping for every call, but that would need a bunch of design. It *might* be worth pursuing though, in the same way that `wasmtime::component::bindgen` has a trappable flag.

ricochet:
> Basically the cleanest fix here is probably some sort of bindgen change that enables some sort of arbitrary result wrapping for every call, but that would need a bunch of design. It _might_ be worth pursuing though, in the same way that `wasmtime::component::bindgen` has a trappable flag.

I like this idea. It definitely felt like I was reaching for a feature that just doesn't exist yet.

Agree this is jank but I think it's still the best option right now. I tried out something like errno but things got pretty complex with concurrency/async/separate lifetimes.

vados-cosmonic (approved):

Issue #5455, opened 2026-08-12
Messaging-triggered components ignore poolSize, maxConcurrency, and maxInvocations


ricochet:
## Summary

A component invoked by the `wasmcloud:messaging` trigger is instantiated from scratch for every message and never consults the instance pool. The three instance limits a component can declare: `poolSize`, `maxConcurrency`, `maxInvocations` are therefore inert on this path. They are honored for inbound HTTP and for component-to-component linked calls, and silently ignored for messaging.

Note for concurrency, we need an async version of wasmcloud:messsaging for this (#5341)

## Why it matters

The shape this hurts most is an event pipeline with a slow, I/O-bound handler: a component that reads a message, makes an HTTP call taking ~10s, and publishes the result. Every message pays full instantiation cost, and every message rebuilds whatever the guest caches in linear memory like connection pools, lazily built runtimes, parsed configuration. In-flight instances grow as `arrival rate × handler duration` with nothing to bound them, so a burst exhausts the host-wide allocator and the tail of the batch is dropped rather than delayed.

Warm instances are the direct fix for the per-message cost, and `maxConcurrency` is the right knob for a handler that spends its time awaiting I/O.

## Proposed change

Route the messaging trigger through the pool the way HTTP does.

1. Add `InstanceJob::Messaging(Box<MessagingJob>)` alongside `Http` and `Linked` (`instance_driver.rs:71-77`), carrying the `BrokerMessage` and a reply oneshot.
2. Dispatch it to `wasmcloud:messaging/handler.handle-message` in the driver's run loop (`instance_driver.rs:288-318`).
3. Have `nats.rs` (and `in_memory.rs`) call `ResolvedWorkload::instance_pool_for_component` and `pool.offer()`, mirroring `http.rs:1875-1912` including the `Dispatch::NeedsInstance` install and the `Dispatch::Saturated` fallback to a one-shot store.

## Implementation notes

A separate variant is still worth having over reusing `Linked` outright, for admission gating (below), for its own tracing span, and so the reply type is the handler's `result<_, string>` rather than a generic `Vec<Val>`.

[...]

PR #5458, opened 2026-08-12
wasmCloud:messaging inflight limit

jfleitz:
## Problem

A component triggered by a `wasmcloud:messaging` subscription gets a **fresh instance per message**, with nothing bounding how many run at once. Concurrent instances = arrival rate × handler duration, so a burst on a subject exhausts the host's wasmtime instance pool and **unrelated workloads fail to start** with `maximum concurrent limit of 1000 for core instances reached`.

None of `poolSize` / `maxConcurrency` / `maxInvocations` reach this path — the subscriber loop never consults the instance pool — so today there is no knob that bounds it.

This is the follow-up [#5455](https://github.com/wasmCloud/wasmCloud/issues/5455) names for itself ("acquire a permit before taking the message, rather than after"). It is independent of #5455 and #5341 and blocks on neither.

## Change

An admission semaphore in front of instantiation, at two levels. The permit is taken **before** the store and instance are built and released when the handler returns, so permits held and instances alive are the same number by construction.

- **Per component** — a new `maxInFlight` field, default **a quarter of the host total** (33 on a stock pooled host, 32 with pooling disabled). Stops one runaway workload taking the host.
- **Host-wide** — `--wasmcloud-messaging-max-in-flight`, default **derived from the instance pool** (133 on a stock host, 128 with pooling disabled). Stops a crowd of components each inside its own limit from adding up to more than the pool can carry (twenty components at 32 is 640 in flight).

The delivery model is unchanged: same loop, same `handle-message` call, same components. **Existing workloads get the bound by upgrading, with no manifest edit and no rebuild.**

## Using it

`max_in_flight` is messaging-plugin configuration, so it travels in `localResources.config` alongside `subscriptions` — as a **string**, like every other value in that map. There is no `maxInFlight` CRD field.

```yaml
kind: WorkloadDeployment
spec:
  replicas: 4
  template:
    spec:
      components:
        - name: ingester
          image: ghcr.io/example/ingester:0.1.0
          localResources:
            config:
              subscriptions: "TOPIC_1"
              # Optional; omit for the host default. A total for this
              # component: the 4 replicas above share it wherever they land on
              # the same host, so this is 32 messages in flight, not 4 x 32.
              max_in_flight: "32"
              # Optional; how long to wait for a slot before shedding the
              # message. Defaults to 30s, capped at 10m.
              admission_wait: "45s"
```

Host flags (both also settable per host group in the chart via `runtime.env`):

| Flag | Env | Default |
| --- | --- | --- |
| `--wasmcloud-messaging-max-in-flight` | `WASH_WASMCLOUD_MESSAGING_MAX_IN_FLIGHT` | derived from the pool — 133 on a stock host, 128 unpooled |
| `--wasmcloud-messaging-max-in-flight-per-component` | `WASH_WASMCLOUD_MESSAGING_MAX_IN_FLIGHT_PER_COMPONENT` | a quarter of the host total — 33 on a stock host, 32 unpooled |

`wash dev` takes the same keys, under `dev.components[].config`.

**Semantics worth knowing:**

- Unset or `0` takes the host default. An explicit `0` on either *flag* is rejected at startup — it would silently mean "process no messages".
- A component asking for more than the host total is clamped to it.
- At the limit the subscriber **parks** rather than dropping. Parking cannot endanger the host's shared NATS connection: async-nats' reader `try_send`s into a per-subscription buffer and drops on overflow rather than blocking.
- `maxInFlight` is a **per-component total**. `maxConcurrency` is per *warm instance*. They are not interchangeable, and `maxInFlight` is not `poolSize × maxConcurrency`.

## Where the defaults come from

`total_core_instances / core-instances-per-component`, computed at startup from the pool the host actually has. Measured, not assumed: a Componentize-Go component is **5** core instances (`$main`, the p1 adapter, three glue modules), Rust p2 is 3. Against the stock pool of 1000 that is 200 components' worth; messaging claims two thirds — **133** — and the rest stays for HTTP-triggered work, warm pools, and long-lived services, the workloads that were failing to *start*. The per-component default is a quarter of whatever host total is in force, so the two knobs can never contradict each other: **33** on that same host.

Because it is derived, raising `WASMTIME_POOLING_TOTAL_CORE_INSTANCES` raises the ceiling with it — no second flag to remember. The pinned **128 / 32** apply only where there is no pool to divide (pooling disabled).

## Testing
- Tested with a local kind cluster and two NATS Go-processors, with a har-coded 10second delay to act like a slow consumer. A burst of 1000 messages were placed on the NATS subject and no error about max instances happened.
- Unit coverage of the admission semantics: default resolution, clamping, zero-floors, the per-component bound, the host-wide bound binding *across* components, both permits freed together, cancel-safety, closed-semaphore
  shutdown, FIFO fairness.
- Config tests: both ceilings reject zero; defaults resolve to 128/32.
- Wire test: `max_in_flight` survives proto → runtime.
- Two Docker-gated integration tests: a burst far larger than the ceiling still delivers in full (catches a leaked permit or a wedged loop), once per-component and once host-wide.
- **Validated end to end on a local kind cluster.**


ricochet:
You already called this out as a follow-on, which is all good, just want to make sure we treat instances as a shared host resource. Connections have QuotaRegistry, one budget for the host, minting per-workload allowances across outbound HTTP, raw sockets, and inbound ports, so that "an operator configures one set of numbers and a per-workload ceiling really is per workload." Instances have nothing equivalent. HTTP ingress, linked calls, warm pools, long-lived services, and now messaging all draw on the same total_core_instances and none of them knows the others exist.

- Soon: an accessor on wash-runtime's Engine for the effective instance budget, so one number has one source of truth instead of two constants that agree by coincidence.
- Eventually: an instance quota registry mirroring the connection one. Messaging is the second subsystem to hand-compute a reservation against an assumed total; the pattern will keep recurring

jfleitz:
This has been retested with two Componentize-Go message processors and ready for review. Go was chosen since each component takes 5 core instances, so the problem of exceeding the number of 1000 core instances happens quickly prior to this change.

crates/wash-runtime/tests/integration_nats_messaging.rs
vados-cosmonic, line 1: It looks like we need a test in here w/ handlers that just never return -- to ensure that we never actually go past the in flight limit concretely. It looks like the current tests make sure that messages are delivered but do not ensure that we actually don't go past the limit for processing.

Also, can we get some integration tests for the in-memory version too?

We could add this in a follow up (there are some more cases that we could have, as well) -- but for now I think just having that one test and then doing some testing for in-memory would be nice to have.

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -19,7 +19,26 @@ const DEFAULT_REQUEST_TIMEOUT_MS: u32 = 10_000;
...
+    /// If another component also received it, that one may still reply, and
+    /// completing the pending request now would fail a request that was about
+    /// to succeed — the sole subscriber is exactly the case where "I dropped
+    /// it" and "nobody will answer" are the same statement.
+    ///
+    /// Known only at routing time, which is why it is carried on the message
+    /// rather than recomputed in the drain loop: by then the subscriber set may
+    /// have changed, and the answer that matters is the one that was true when
+    /// the message fanned out.
+    sole_subscriber: bool,
```
vados-cosmonic, line 40:
```suggestion
    was_sole_subscriber: bool,
```

NIT: this is past-state so maybe we should be explicit in naming it as such?

LiamRandall (approved):
PR 5458 works as advertised. I checked out the branch, built wash, and tested it live using slow consumer Go/NATS example (subject1-handler, 10s sleep per message) against a local NATS server.  The admission semaphore bounds concurrency exactly at the configured ceiling, sheds loudly at the deadline, and loses nothing it shouldn't. All 73 new unit tests pass. I have a few feedback items, one of which (rolling-redeploy limit stickiness) is worth raising on the PR.

**How I tested**

I ran the Go handler directly under the PR's runtime: wash dev with build.command: "true" and dev.component_path pointed at the prebuilt subject1-handler.wasm, dev.data_nats_url at a local nats-server -js, and the messaging config under workload.config. This exercises the exact same MessagingLimits machinery wash host uses.

**Results:**

- max_in_flight: "5", 25-message burst: handler starts arrived in exactly 5 waves of 5, one wave per 10s sleep cycle. All 25 messages stored to KV and republished to subject2. Zero sheds, zero slow-consumer drops.
- max_in_flight: "200" (over-ask), admission_wait: "2s", 40-message burst: clamped to 33 with a clear warning naming both numbers. 33 admitted instantly, then the parked loop shed 4 messages serially at the 2s deadline with the actionable warn message, and the last 3 were admitted when wave-1 slots freed at t+10s. 36 delivered + 4 shed = 40 accounted for, no silent loss.
- Defaults: dev host resolved host_total=133 / per_component_default=33 derived from the stock pool, matching the PR's documented numbers; the startup log line naming the source ("derived from the instance pool" vs "flag") is genuinely useful.
- Flags/env: --wasmcloud-messaging-max-in-flight 0 rejected at startup with a clear error; WASH_WASMCLOUD_MESSAGING_MAX_IN_FLIGHT=50 resolved to host_total=50, per_component=12.
- Tests: 67 messaging unit tests + 6 config tests pass. (I skipped the two Docker-gated integration tests — Docker wasn't running — but my live runs cover the same "burst larger than ceiling still delivers in full" property.)

Code-wise, both backends (NATS and in-memory) gate their single instantiation site, the permit is held by the spawned handler task so trap/error paths release it, cancel-safety and the component-before-host acquire order are right, and the multiplexed backends don't need gating (consumer-import only, no per-message instances).

**Feedback for the PR**

1. Rolling redeploys never pick up a changed max_in_flight (worth considering a minor refactor).
- ComponentGates keeps the first binding's limit until every replica on the host releases (bindings == 0). With replicas: 2+ rolling one-at-a-time, new replicas re-bind the old gate before old ones release, so the count never hits zero and the old ceiling persists indefinitely; and the warning that fires blames "two genuinely different components collided on one name," which will mislead the operator in this (much more common) benign case. Suggest at minimum rewording the warning to mention config rollout; ideally resize via add_permits/forget_permits deltas.

2. PR description inaccuracy. 
- The body says "A component asking for more than the host total is clamped to it" — the code (correctly, per the flag's own help text) clamps to min(per_component_default, host_total), i.e. 33 on a stock host, not 133. A component can lower its ceiling but can never raise it above the per-component default without the operator raising the flag. That's a defensible design, but the PR body and the k8s example comment ("this is 32 messages in flight") should state it plainly, because a user setting max_in_flight: "64" will get 33 and only find out from a warn log.

3. Minor: flag validation ordering. 
- `wash host --wasmcloud-messaging-max-in-flight 0` connects to NATS before failing validation. The zero/max check needs nothing from the engine; it could run before any network I/O.

4. Serial shedding is head-of-line blocking, by design. 
- Once saturated, the subscriber loop sheds at most one message per admission_wait (my run: 4 sheds × 2s each, during which a message that would have been admitted at t+10s just sat behind them). That's coherent; but with the default 30s wait, a saturated component drains its NATS buffer at 1 message per 30s, which makes transport-buffer overflow (the uncounted loss the docs worry about) more likely than the comments suggest. Might be worth a note, or shedding all currently-waiting messages once one deadline fires.

Overall, this is solid and addresses the problem.

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -377,6 +418,786 @@ macro_rules! messaging_handler_dispatch {
...
+///
+/// The pool is spent in *core* instances, so this is what converts "instances
+/// the pool can hold" into "messages we may admit".
+const WORST_CASE_CORE_INSTANCES_PER_COMPONENT: u32 = 5;
+
+/// Share of the pool's component capacity messaging may claim. The remainder is
+/// what HTTP-triggered work, warm pools, and long-lived services draw on — the
+/// workloads that otherwise fail to *start* when a messaging burst drains the
+/// pool.
+const MESSAGING_POOL_SHARE_NUM: u32 = 2;
```
vados-cosmonic, line 519: These numbers should probably be just kept together in general as a tuple, or maybe a struct or enum if we want to have different ways of expressing this?

crates/wash-runtime/tests/integration_nats_messaging.rs
vados-cosmonic, line 1: It doesn't seem like there is a test in here that actually ensures that the in flight check works -- we need to have a workload that just pauses/never resolves.

Also, it would be nice to have some integration tests for the in-memory handler as well to make sure this works equally in both places.

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -537,4 +1504,823 @@ mod tests {
         );
         assert!(parse_subscriptions(None).is_empty());
     }
+
+    // --- Admission ceilings -------------------------------------------------
```
vados-cosmonic, line 1508:
```suggestion
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -1,10 +1,51 @@
...
+//! `dev.components[].config` in a `wash dev` config, or `localResources.config`
+//! on a workload — falling back to the workload-scoped `wasmcloud:messaging`
+//! host-interface config where the backend supports it. All are optional.
+//!
+//! | Key | Value | Unset |
+//! | --- | --- | --- |
+//! | `subscriptions` | Comma-separated subjects, NATS wildcards allowed (`orders.*`, `audit.>`) | Receive everything |
+//! | `consumer_group` | Queue-group name, or `broadcast` for no grouping (NATS only) | A name derived from namespace/workload/component |
+//! | `max_in_flight` | Messages this component may process at once, across every replica of it on this host | The host's per-component default |
+//! | `admission_wait` | How long to wait for a slot before shedding (`45s`, `2m`, or bare seconds) | [`DEFAULT_ADMISSION_WAIT`] |
```
vados-cosmonic, line 15:
```suggestion
//! | `shed_incoming_after` | How long to wait for a slot before shedding (`45s`, `2m`, or bare seconds) | [`DEFAULT_ADMISSION_WAIT`] |
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -377,6 +418,786 @@ macro_rules! messaging_handler_dispatch {
...
+/// have the responder waiting on a permit the requester is holding. The guest's
+/// own `timeout_ms` already breaks that, but only after every such request has
+/// burned its full timeout; the deadline here caps how long the queue behind
+/// them grows.
+///
+/// 30s is chosen to be well past any legitimate burst — a message that has
+/// waited this long is behind work that is not clearing — and well past typical
+/// guest `timeout_ms` values, so the guest's own deadline fires first in the
+/// request/reply case and this one only catches genuine saturation.
+pub const DEFAULT_ADMISSION_WAIT: Duration = Duration::from_secs(30);
```
vados-cosmonic, line 487: What do you think about `DEFAULT_SHED_AFTER` here?

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -377,6 +418,786 @@ macro_rules! messaging_handler_dispatch {
...
+/// The pool is spent in *core* instances, so this is what converts "instances
+/// the pool can hold" into "messages we may admit".
+const WORST_CASE_CORE_INSTANCES_PER_COMPONENT: u32 = 5;
+
+/// Share of the pool's component capacity messaging may claim. The remainder is
+/// what HTTP-triggered work, warm pools, and long-lived services draw on — the
+/// workloads that otherwise fail to *start* when a messaging burst drains the
+/// pool.
+const MESSAGING_POOL_SHARE_NUM: u32 = 2;
+const MESSAGING_POOL_SHARE_DEN: u32 = 3;
```
vados-cosmonic, line 520: These numbers should probably just be together in a tuple I think, intended to be used together as essentially a quotient

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -377,6 +418,786 @@ macro_rules! messaging_handler_dispatch {
...
+    timeouts: AdmissionTimeouts,
+    /// Per-component gates, keyed by manifest identity so replicas of one
+    /// deployment share one. See [`ComponentGates`].
+    gates: ComponentGates,
+}
+
+/// The per-component gates on this host, keyed by [`AdmissionIdentity`].
+///
+/// **Keyed by manifest identity, not by component id.** Every replica of a
+/// deployment is a separate workload with its own `uuid::Uuid::new_v4()`
```
vados-cosmonic, line 611: It might be useful here to use a v7 for time relationship

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -377,6 +418,786 @@ macro_rules! messaging_handler_dispatch {
...
+            .inner
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        let Some(entry) = gates.get_mut(identity) else {
+            return;
+        };
+        entry.bindings = entry.bindings.saturating_sub(1);
+        if entry.bindings == 0 {
+            entry.semaphore.close();
+            gates.remove(identity);
```
vados-cosmonic, line 703: There's a subtle problem here -- basically if we have a bunch of components in flight, and bindings drop to zero here and the gate is removed, a new workload with the same identity is going to get an absolutely fresh set of permits, and the *old* in flight handlers plus the new in flight handlers could exceed what is there. 

In component cleanup we do not actually wait for the handler to finish -- i.e. it could still be running, so the new handlers would encounter the now-missing identity in `gates`, and get a fresh set of permits.

At least two ways of handling this:
- Keep a single component gate alive longer until we know all outstanding permits are done
- Wait for active handler when doing component cleanup

This problem is pretty subtle I think -- took AI a bit to find it, so we can handle this in a follow up, and that would be a good small commit with a test that could cover the case.

vados-cosmonic (approved):
Overall LGTM 🚀 

OK, so there seems to be one issue in here -- when handlers are cleaned up, if you have the same manifest identity w/ old and new binding, the old handlers hold permits from the *old* semaphore, so old and new handlers together can go past `max_in_flight`. 

Component gates have to stay alive until *all* the related permits drain, or we have to track the active handler tasks before we allow a new gate to be made.

Happy with the issues that you agree with being addressed in a follow up as well.

PR #5459, opened 2026-08-12
test(fixtures): stage the dual-handler WIT deps instead of committing them

Aditya1404Sal:
Follow-up to #5413, addressing the review feedback from @vados-cosmonic and @ricochet on the committed `wit/deps` under `messaging-dual-handler`.

Every other fixture declares its deps in a `wkg.toml` and lets `wash build` fetch them, leaving `wit/deps` a gitignored build artifact. That fixture was the exception because its world exports both `wasmcloud:messaging/handler@0.2.0` and `@0.3.0`, which no `wkg.toml` can express — overrides are keyed by package name with no version, and `Override.version` is ignored outright once `path` is set. A fetch that can't resolve every package the world names also writes nothing rather than a partial tree, so there's no fetch-then-top-up either.

So `xtask` now stages `wit/deps` from the same shared `p3-wit-deps/` source the overrides would have resolved to, then builds with `--skip-fetch`. Net effect: 2113 lines of vendored WIT deleted, the gitignore exception gone, and the directory is generated like everywhere else.

Verified with a clean-slate `cargo xtask build-fixtures` (deps and artifact deleted first) plus the full `wash-runtime` suite.


ricochet:
Oop didn't see this. I also covered it at https://github.com/wasmCloud/wasmCloud/pull/5460

Aditya1404Sal:
Actually your solution is way better 🙃
Closing this one in favor of #5460

PR #5460, opened 2026-08-12
Fix/async messaging follow-ups

ricochet:
Some small follow-ups for async messaging. 

## A versionless `wasmcloud:messaging` entry stopped binding

A host-interface entry that pins no version stopped binding once `MultiplexedAsyncMessaging` joined the sync multiplexer on one host. `WitInterface::same_package` treats a missing version as compatible with every version, so a versionless entry world-matches **both** plugins; the async one sorts first in the id-ordered selection loop and both support named instances, so it claimed the entry, found nothing it could bind, and returned `Ok(())`. The interface was already out of the unmatched set, so the sync plugin never saw it and the workload failed later on an unresolved import.

Adds `HostPlugin::claims`, a per-interface veto over a world match, consulted everywhere selection matches a plugin to an interface including `other_plugin_serves`, so a plugin never defers to a sibling that would refuse the interface. It defaults to claiming everything, so no other plugin changes. `MultiplexedAsyncMessaging` refuses anything below `@0.3.0`, restoring versionless entries to the sync multiplexer.

Messaging is the first namespace where two plugins serve one namespace:package at different revisions (keyvalue and blobstore split along `wasi:` vs `wasmcloud:`), which is why `world()` had never needed to express this.

## An oversized message body trapped the guest

`collect_body`'s over-limit path returned `StreamResult::Cancelled` without taking anything from the source. wasmtime only permits that under `finish`, which is false for an ordinary guest write, so a guest streaming past the 16 MiB cap trapped instead of receiving `error::message-too-large`. `Dropped` is the disposition for refusing items.

The consumer was a near-copy of the blobstore `write-data` one, so it moves to `plugin::stream_collect` with the limit as a parameter; blobstore passes `usize::MAX` and stays unbounded as before.

## `messaging-dual-handler` no longer needs `--skip-fetch`

One wkg override can serve both handler revisions: a package directory may carry a second version as a nested package, and `wit fetch` writes every package in the resolved group into `wit/deps`. `p3-wit-deps/wasmcloud-messaging-dual/` holds `@0.3.0` with `@0.2.0` nested beside it. Drops the committed deps tree (-2,225 lines), the `.gitignore` negation, and `SKIP_FETCH_FIXTURES` in xtask.

The two revisions cannot join the shared `wasmcloud-messaging-0.3.0/` dir: two versions of one package in a resolve make wit-bindgen emit version-suffixed module names, which would rename the bindings in `msg-counter` and `messaging-echo-p3` for no reason.

## Cleanups

Folds the three copies of `render_handle_error` into one macro. The strings are host-observable, so the trigger service's hand-written copy had to match the plugins' generated ones by inspection and corrects comments.

## Testing

New regression test verified to fail without the fix

## Note for reviewers

`HostPlugin::claims` adds a method to the public plugin trait. It is defaulted, so downstream impl. The alternative considered was renaming the async plugin's id so it sorts after the sync one — aone-liner, but it makes correctness depend on lexicographic ordering of id strings with nothing to


vados-cosmonic (dismissed):
LGTM 🚀

crates/wash-runtime/tests/integration_messaging_async.rs
```diff
@@ -248,6 +248,61 @@ async fn async_handler_handles_messages_concurrently() -> Result<()> {
...
+    .await
+    .context("oversized publish timed out")??;
+    anyhow::ensure!(
+        response.status().is_success(),
+        "oversized publish endpoint returned {}",
+        response.status()
+    );
+    let result = response.text().await?;
+    assert_eq!(
+        result, "Err(MessageTooLarge)",
```
jfleitz: Changing this to "Err(Error::MessageTooLarge)" will fix the CI test

PR #5478, opened 2026-08-18
fix/embedder oci and socket defaults

### LINKED ISSUE #5477: `oci_ca_paths` should live on `HostConfig`
Extra CA trust for OCI pulls is the only OCI setting that is not part of `HostConfig`. Every other one is a field an embedder sees while constructing the struct (`crates/wash-runtime/src/host/mod.rs:1124`):

```rust
pub struct HostConfig {
    pub allow_oci_insecure: bool,
    pub oci_pull_timeout: Option<Duration>,
    pub oci_cache_dir: Option<PathBuf>,
    // no oci_ca_paths
}
```

CA trust instead lives in a process-global `OnceLock` (`crates/wash-runtime/src/oci.rs:58`) populated by a free function that a caller has to know exists:

```rust
static EXTRA_CA_CERTIFICATES: OnceLock<Vec<Certificate>> = OnceLock::new();
pub fn set_extra_ca_certificates(paths: &[PathBuf]) -> Result<()>
```

Today the only callers are the wash CLI itself: `crates/wash/src/cli/oci.rs:48` and `crates/wash/src/cli/host.rs:423`.

## Why this matters

An embedder that builds a host and configures every OCI knob it can see still ends up with a host that cannot pull from a registry behind a private CA and gets no compile error. The type system actively hides the omission, because the field simply is not there to be filled in.

## Proposal

Add `oci_ca_paths: Vec<PathBuf>` to `HostConfig` and have the host apply it during startup, before anything pulls. Host component plugins are resolved and pulled at startup, so applying it later is too late.

Keep the process-global for the CLI-only paths: `wash oci pull` / `wash oci push` need CA trust without constructing a host at all, which is a legitimate reason for the global to exist. The two are not in conflict. `HostConfig` becomes the supported way for an embedder to configure a host, and the global remains the mechanism underneath.

The benefit is that this class of omission becomes a compile error for embedders constructing `HostConfig` exhaustively, and at minimum becomes discoverable for everyone else, since the setting sits next to `allow_oci_insecure` where someone configuring OCI behaviour will actually look.

## Related

Worth considering alongside this: a failed pull is returned in `WorkloadStartResponse` as `WorkloadState::Error` with a message but is never logged locally (`crates/wash-runtime/src/washlet/mod.rs`, the early returns around lines 485-565). That is what made the above take longer to diagnose. Logging those failures at `error!` would make a trust misconfiguration self-evident regardless of how the CA paths are configured.

ricochet:
Three fixes for host embedders.

Closes #5477

### `SocketPolicy::default()`

`egress_addrs` defaulted to `EgressAddressPolicy::permissive()`, while `wash host start` with no socket flags builds `deny_special: true, allow_private: true`. An embedder that installed no policy with the `Engine::builder()` without `with_socket_policy` ran with range filtering off. The default is now EgressAddressPolicy::default().

### oci_ca_paths on HostConfig (#5477)

Extra CA trust was the only OCI setting an embedder could not see while constructing HostConfig.

### Workload-start failures are logged (#5477, "Related")

A failed start was returned as WorkloadState::Error with a message and not logged.

### DIFF
--- a/crates/wash-runtime/src/host/mod.rs
+++ b/crates/wash-runtime/src/host/mod.rs
@@ -856,14 +856,25 @@ impl HostApi for Host {
         {
             let mut workloads = self.workloads.write().await;
             if workloads.contains_key(&request.workload_id) {
+                let message = format!(
+                    "Workload ID [{}] already exists (the exising workload must be stopped to reuse the ID)",
+                    request.workload_id
+                );
+                // Logged here rather than at the single site below, because
+                // this refusal returns before the start ever begins. At `warn`,
+                // not `error`: this is the guard that makes a replayed start
+                // request idempotent, and a scheduler retrying one is not a
+                // host malfunction.
+                tracing::warn!(
+                    workload_id = request.workload_id,
+                    reason = message,
+                    "refused to start workload"
+                );
                 return Ok(WorkloadStartResponse {
                     workload_status: WorkloadStatus {
                         workload_id: request.workload_id.clone(),
                         workload_state: WorkloadState::Error,
-                        message: format!(
-                            "Workload ID [{}] already exists (the exising workload must be stopped to reuse the ID)",
-                            request.workload_id
-                        ),
+                        message,
                     },
                 });
             }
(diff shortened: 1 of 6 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5479, opened 2026-08-18
fix(runtime): stop warning on anonymous pulls without a docker config

### LINKED ISSUE #5475: Anonymous OCI pulls log a WARN when the host has no docker config
Every component pull from a registry that needs no authentication logs a warning:

```
WARN workload_start{...}:pull_component{reference=oci-registry.cosmonic-system.svc.cluster.local:5000/keyvalue-counter:e2e pull_policy=IfNotPresent}:resolve_credentials{registry=oci-registry.cosmonic-system.svc.cluster.local:5000}:
  failed to retrieve docker credentials error=docker credential retrieval error: Unable to read config
```

The pull then succeeds anonymously. Nothing is wrong, but every pull on a host without a `~/.docker/config.json` which is the normal case for many emits a warning.

## Why

`CredentialResolver::resolve_credentials` already treats "no credentials configured" as a non-event, mapping it to `Ok(None)` and a `debug!` (`crates/wash-runtime/src/oci.rs:395-398`):

```rust
Err(
    CredentialRetrievalError::ConfigNotFound
    | CredentialRetrievalError::NoCredentialConfigured,
) => Ok(None),
```

But that arm almost never matches in a container. In `docker_credential` 1.4.0, `ConfigNotFound` is returned only when `config_dir()` itself is `None` (no `HOME` at all). When `HOME` *is* set and `config.json` simply doesn't exist, `File::open` fails and **every** `io::Error`. `NotFound` included and is flattened to `ConfigReadError` (`docker_credential-1.4.0/src/lib.rs:196`):

```rust
let f = File::open(config_path).map_err(|_| CredentialRetrievalError::ConfigReadError)?;
```

`ConfigReadError` isn't in the arm above, so it falls through to the catch-all at `oci.rs:405`, becomes an `anyhow` error, and is logged by `oci.rs:378`:

```rust
Err(e) => warn!(error = %e, "failed to retrieve docker credentials"),
```

## Impact

Routine, expected behavior is reported at WARN. It fires once per component start, so a host running many workloads produces a steady stream of warnings that operators learn to ignore which is what makes a real credential problem easy to miss later.

## Reproduce

Run a host with `HOME` set to a directory containing no `.docker/config.json` and start a workload whose image comes from a registry allowing anonymous pulls.

## Suggested fix

Add `ConfigReadError` to the existing `Ok(None)` arm, since the crate conflates "missing" with "unreadable":

```rust
Err(
    CredentialRetrievalError::ConfigNotFound
    | CredentialRetrievalError::NoCredentialConfigured
    | CredentialRetrievalError::ConfigReadError,
) => Ok(None),
```

Stat the path first and only treat `NotFound` as `Ok(None)`.

ricochet:
Closes #5475

Every pull from a registry that needs no authentication logged

```text
  WARN ... failed to retrieve docker credentials error=docker credential retrieval error: Unable to read config
```

and then succeeded anonymously. `resolve_credentials` already treats "no credentials configured" as a non-event, but the arm that does so matches `ConfigNotFound`, which `docker_credential` returns only when there is no home directory at all. When `HOME` is set and `~/.docker/config.json` simply does not exist (common).

We now:
- emit debug if the path is absent
- WARN if the path is present and not parsable.

### DIFF
--- a/crates/wash-runtime/src/oci.rs
+++ b/crates/wash-runtime/src/oci.rs
@@ -135,6 +135,27 @@ fn extra_ca_certificates() -> Vec<Certificate> {
 #[deprecated = "old media type used before Wasm WG standardization"]
 const WASMCLOUD_MEDIA_TYPE: &str = "application/vnd.module.wasm.content.layer.v1+wasm";
 
+/// The `config.json` [`get_credential`] reads, or `None` when it would find no
+/// config directory at all.
+///
+/// Mirrors `docker_credential`'s own resolution, which is not public: the crate
+/// reports a file that is absent and one that is malformed as the same error,
+/// and this is what tells them apart.
+fn docker_config_path() -> Option<PathBuf> {
+    docker_config_dir(std::env::var_os("DOCKER_CONFIG"), std::env::var_os("HOME"))
+        .map(|dir| dir.join("config.json"))
+}
+
+/// [`docker_config_path`]'s directory rule, over the environment it reads.
+fn docker_config_dir(
+    docker_config: Option<std::ffi::OsString>,
+    home: Option<std::ffi::OsString>,
+) -> Option<PathBuf> {
+    docker_config
+        .map(PathBuf::from)
+        .or_else(|| home.map(|home| Path::new(&home).join(".docker")))
+}
+
 /// Configuration for OCI operations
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM 🚀

PR #5483, opened 2026-08-21
Add Host core and memory knobs

jfleitz:
### core instances
A wasmCloud host inherits the default value of 1000 core instances from wasmtime, and there is no current way of setting this. Regardless if the type of architecture that the wasmCloud host is running on, it is always 1000.

### memory allocation
wasmCloud also inherits the a default max heap (in the linear) memory size of 4gb for components. Each component that is deployed to a wasmCloud host thinks it has up to 4gb of memory, when there might not be that much memory available. If a wasmCloud host that is running in a pod with 512mb of memory allocated to it, and 10 components utilize 100mb while processing a transaction, the host will get an Out-Of-Memory error, and the pod will restart.

## Fix

This PR does a simple fix for exposing the configuration for the number of core instances, how much memory is available to the host, and how much memory the heap allocation size should be. This helps prevent an out of memory error and protects the host, as well as provides sizing and alerting in the future.


charts/runtime-operator/values.yaml
```diff
@@ -333,6 +333,26 @@ runtime:
   # -- Additional CLI args appended to every host group container, for host
   # flags the chart does not template. Merged with per-host-group `extraArgs`.
   extraArgs: []
+  # -- Host memory knobs. Each maps to a wasmtime setting that was previously
+  # unreachable; leaving all three empty keeps today's behaviour exactly.
+  #
+  # A host group's own `memory` block overrides these key by key.
+  memory:
```
vados-cosmonic: SUPER NIT: What do you think about putting this *one* level in, under some sort of `wasmCloudHost` or `hostConfig` key? It feels like we go from ~k8s config into very deep wasmCloud config without much ceremony other than `memory`.

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -319,6 +324,15 @@ impl Engine {
     /// accounts for `WASMTIME_POOLING_TOTAL_CORE_INSTANCES` and for a
     /// caller-supplied pooling config, neither of which is visible from
     /// [`EngineBuilder::with_max_instances`].
+    /// The memory shape this engine was built with.
+    ///
+    /// Read back rather than recomputed, so what a caller reports is what the
+    /// engine actually installed — including an embedder's `max_instances`
+    /// winning over the flag-driven count.
+    pub fn host_memory(&self) -> host_memory::HostMemory {
```
vados-cosmonic: SUPER NIT: `HostMemoryLimits` ? `HostMemoryBudgets`? The object reads like it *is* a "host memory" (though what that would be is not clear), and it seems what it actually is is a bag of configuration/metadata around limits/budgets.

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but nits, but they may prompt larger changes (`HostMemory` is evidently already in!), so feel free to leave them if so.

charts/runtime-operator/values.yaml
jfleitz: Good call. I'm also not sure if maxMemory is really needed, since we have`resources.limits.memory`. Actually, I think I might just wrap up all of this under `resources`, including `coreInstances`, since we already have that group defined.

crates/wash-runtime/src/engine/mod.rs
jfleitz: Agreed and changed to HostMemoryBudgets, as that sounds more inlined with the fields in the struct.

charts/runtime-operator/values.yaml
vados-cosmonic: Yeah so I thought this was memory *for the wasm host* which is sort of... inside the process, which is running alongside other stuff. The thing is, I don't know how much daylight there is between the wasm host's max memory and the processes (and then k8s'/OOMkillers).

charts/runtime-operator/templates/runtime/deployment.yaml
```diff
@@ -136,6 +140,34 @@ spec:
...
+
+                   These go through the flags' `WASH_*` env fallbacks rather
+                   than `args` on purpose. The chart is routinely paired with a
+                   host image it does not ship (`runtime.image.tag`), and an
+                   older host aborts on an argument it does not recognise —
+                   whereas it simply ignores an env var it has never heard of.
+                   Keep any new host knob on this side of that line. */}}
+            {{- with $resources.limits }}
+            {{- with .memory }}
+            - name: WASH_MAX_MEMORY
```
ricochet: I think this should be WASH_HOST_MAX_MEMORY

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -319,6 +321,15 @@ impl Engine {
     /// accounts for `WASMTIME_POOLING_TOTAL_CORE_INSTANCES` and for a
     /// caller-supplied pooling config, neither of which is visible from
     /// [`EngineBuilder::with_max_instances`].
+    /// The memory shape this engine was built with.
```
ricochet: This split the doc comment for total_core_instances

crates/wash/src/config.rs
```diff
@@ -2171,3 +2171,53 @@ secrets:
         assert!(config.dev().environment.is_empty());
     }
 }
+
+/// Resolve the host's memory budget, per-memory ceiling and instance count.
+///
+/// Mirrors [`wasmcloud_messaging_limits`]: the flags are parsed and validated
+/// here, once, so the runtime is handed a value that is already correct rather
+/// than a bag of strings to interpret later.
```
ricochet: Delete these lines

crates/wash/src/config.rs
```diff
@@ -2171,3 +2171,53 @@ secrets:
...
+///
+/// Nothing is gated on the result. Both pool knobs fall through to wasmtime's
+/// own defaults when unset, so a host that sets none of these behaves exactly
+/// as it always has — the point of this pass is that the three numbers are
+/// *reachable* and *reported*, not that anything new is enforced.
```
ricochet:
```suggestion
```

charts/runtime-operator/templates/runtime/deployment.yaml
ricochet: Reading the docs and this PR, and this value is still unclear:
> Total guest memory this host may use

Is it the total guest memory for each guest or all of the guests combined? How does it relate to guest heap memory?

jfleitz: I am changing this to `WASH_HOST_MAX_GUEST_MEMORY` and updating the struct property to match, as well as the comments to match. This is the total amount of memory that all guests can use on the host.

ricochet (approved):

PR #5489, opened 2026-08-22
chore: drop what build.rs left behind

jtakakura:
## Feature or Problem

Follow-up to #5446, which removed `crates/wash-runtime/build.rs` and committed
the proto bindings. Things that existed only for that build script are still
around; this drops them, one concern per commit:

- The `cargo-machete` ignores for `pbjson`, `pbjson-types` and `serde`. They
  were needed because the bindings were generated into `OUT_DIR`, which machete
  does not scan. The bindings are in `src/` now, so machete sees the uses (339
  references to `serde` alone) and is clean without the entries. `tonic` and
  `tonic-prost` stay ignored.
- `protobuf-compiler` / `libprotobuf-dev` from the bench host playbook, and the
  README passages that install, verify and troubleshoot them.
- `Setup protoc` from `check`, `lint`, `e2e-fixtures` and `canary-binaries` in
  `wash.yml`, and from `analyze-rust` in `codeql.yml` with its comment citing
  `build.rs`.
- `[build-dependencies]` in `crates/wash`, which named `anyhow`,
  `tonic-prost-build` and `pbjson-build` but had no `build.rs` and no `build =`
  key, so cargo ignored it entirely. Not on the #5446 follow-up list; found
  while checking the rest. `anyhow` stays in `[dependencies]`, where it is used.

Two `protoc` setups are deliberately kept: `proto.yml` regenerates the bindings,
and `examples.yml` builds `examples/grpc-hello-world`, whose crates have their
own `build.rs`. Those examples are not workspace members, so workspace builds
never reach them.

## Related Issues

Follow-up to #5446.

## Release Information

next

## Consumer Impact

None. This removes CI steps, provisioning packages and dependency declarations
that had no effect on shipped code.

## Testing

`cargo build -p wash` and `cargo machete` pass locally. CI is the real check: if
a job still needed `protoc` it fails at the step that needs it, and that one
step can go back.


vados-cosmonic (approved):
LGTM 🚀

PR #5495, opened 2026-08-25
Fix/abandoned call deregistration

ricochet:
Changes our epoch handling to deregister an abandoned call once its guest yields.

This fixes an issue where a client disconnect remained registered for the entire ingress deadline (~20 seconds).

```
ERROR wash_runtime::engine::abandon: an abandoned call has wedged this store
      past its grace; trapping it store_id=583fb4fe… executed_ms=1000
served before = 1, served after = 1     ← trapped and restarted
```

In every protective test in `integration_abandoned_calls`, it sets its deadline to 2s, so the in-store bound always expired first and nothing in the suite could see this.

This adds `watch_until_abandoned` which drops the call's `AbandonedGuard` after arming so a yielding guest stops being visible to the epoch callback.

This PR adds three tests, each verified to fail without the fix. 
Both fixture `chatter` branches hop in steps *shorter* than the sampling window.

| test | unfixed |
|---|---|
| `a_disconnect_does_not_trap_a_healthy_guest_when_the_deadline_outlives_the_grace` | `served 1 -> 1` |
| `a_chatty_messaging_handler_survives_being_abandoned` | traps at `executed_ms=1000` |
| `a_chatty_capability_call_does_not_restart_the_plugin` | escalates at `executed_ms=3000` |

### DIFF
--- a/crates/wash-runtime/src/engine/abandon.rs
+++ b/crates/wash-runtime/src/engine/abandon.rs
@@ -22,12 +22,21 @@
 //! and the guest has since run through a grace's worth of execution.
 //! Execution is credited from the callback's own fires (`ExecutionCredit`):
 //! each fire proves the guest was running and is worth at most one sampling
-//! window, and only a pause too long to be scheduler delay clears the credit
-//! — so a guest waiting in a host call accrues nothing, while a pinned one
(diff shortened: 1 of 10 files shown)


jfleitz (approved):
lgtm

PR #5499, opened 2026-08-25
speed up CI

ricochet:
Put every cargo test in the check job behind on one feature set. The job used four different -p/--features combinations, and since cargo unifies features over the selected packages, each one re-fingerprinted the shared deps and recompiled the tree.

This resulted in six compiles, ~16.5 min against 12.4 min of actual testing. All four cargo test calls now share --workspace plus a job-level TEST_FEATURES, so only the first pays for the build.

Deletes a test that was not meant to be committed. bench_topology_overhead was a measurement, not a test. Also removes a fully commented-out 512MB stress test in the http blobstore suite.


.github/workflows/wash.yml
```diff
@@ -22,6 +22,10 @@ concurrency:
 permissions: {}
 
 env:
+  # The non-default features every `cargo test` and `cargo clippy` here runs
+  # with. Spelled per-package: bare names apply to every selected package, so
+  # the two would drift apart.
+  OPTIONAL_FEATURES: wash-runtime/wasi-tls,wash/host-component-plugins
```
vados-cosmonic, line 28: NIT: there should be a per-job setting for this, no?

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenvironment

Were there too many jobs that needed this? Seems like it could go under the `check` job

vados-cosmonic (dismissed):

.github/workflows/wash.yml
ricochet: It was under check originally but lint also needs it. `jobs.<job_id>.env` can't cross jobs. We can scope it back to a copy for check and lint, but then there may be drift. The annoying part is that it'd be silent drift aka there will be an additional compile and it'll work.

vados-cosmonic: yeah we chatted about this but this is good for me, definitely good to have this be global, certainly easy to find.

PR #5501, opened 2026-08-26
ci: label PRs and gate incompatible CRD changes

ricochet:
Detects whether a PR touches something an operator needs to care about, like CRD changes that may impact upgrades, and also whether a CRD schema change is safe for someone already running these CRDs.

Labels are now declared in one file, `.github/labels.yml`, which  carries each label's color, description, optional aliases, and optional path rules.

- `labels-sync.yml` reconciles the repository's labels
- `labeler.yml` renders the actions/labeler config and applies them by changed path

The crd-diff job compares every CRD against the PR's base commit with sigs.k8s.io/crdify and fails on an incompatible change.

A maintainer who decides an incompatible change is correct applies the `crd-diff-override` label.


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,55 @@
...
+# actions/labeler view of that file at run time.
+#
+# `pull_request_target` because a fork PR's token is read-only on
+# `pull_request`, leaving exactly the PRs that need triage labels unlabeled.
+# The checkout takes the base ref, so no PR-authored code runs here; see
+# .github/zizmor.yml for the full rationale. Rules therefore come from the
+# base ref too, so a PR editing them is labeled with the old ones.
+
+on:
+  pull_request_target:
```
vados-cosmonic, line 13: Yeah, I don't think we should ever use `pull_request_target`. It's basically a massive risk and not worth it. It is *so often* at the scene of the crime when someone gets hacked or compromised via actions.

There's got to be another way to do this, and if there isn't IMO we should just not do it.

vados-cosmonic (commented):
Everything looks good to me except `pull_request_target` -- it's just too much of a vector for compromise IMO

.github/workflows/labeler.yml
ricochet: You can tell by the comments above that I pushed on it and tried out other options. I can't figure out a way to have PR's from forks (most PRs) be labeled. Freaking GH. It's safe right now given all the caveats called out here, but yeah future changes to that workflow is the risk.

vados-cosmonic: Yeah so a little late getting back here, but I'm convinced that `pull_request_target` is unsafe at any speed (mostly because of what you pointed out, that the future changes might be an issue)...

Is there no other way we can do this? I'd even take a scheduled action that runs super often over this.

PR #5503, opened 2026-08-27
fix(wash-runtime): compile workloads off the async runtime

ricochet:
This change makes it to so that the host continues servicing NATS while a workload starts. A host that goes quiet stops heartbeating, and the operator deletes a Host it has not heard from along with every Workload.

A burst of image pulls and compiles held up the async-nats connection task from draining nats messages.

This change moves compilation into `spawn_blocking`, Commands move off the select loop into a `JoinSet`.

Handling commands concurrently means a stop can now overtake a start, and washlet fetched every image before the host reserved anything.

Now the workload id is claimed before any image is fetched, through a crate-private `WorkloadReservation`, and given back on every path that fails before start begins.

`MAX_CONCURRENT_STARTS` bounds how many starts pull and compile at once, since nothing capped that once they could overlap.

This does not cover every compile. `Host::intersect_interfaces` and host component plugin loading still call `Component::new` on whichever thread reaches them, and a future caller of `initialize_workload` inherits the original problem, because the hop lives at the call site rather than in the engine. Right now this only impacts dev hosts and fixing it would be a larger refactor that I'm not going to do now.

### DIFF
--- a/crates/wash-runtime/src/host/mod.rs
+++ b/crates/wash-runtime/src/host/mod.rs
@@ -159,6 +159,50 @@ impl<T: HostApi> HostApi for Arc<T> {
     }
 }
 
+/// The claim-then-start protocol a caller needs when it has work to do between
+/// deciding to start a workload and having something the host can run.
+///
+/// Kept off [`HostApi`] deliberately. Its three calls only make sense together,
+/// every reservation has to reach [`WorkloadReservation::workload_start_reserved`]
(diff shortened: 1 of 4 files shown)


jfleitz (approved):
LGTM and the test cases look like they cover all the critical scenarios.

vados-cosmonic (approved):
LGTM 🚀

PR #5505, opened 2026-08-27
fix(test): postgres test flake

ricochet:
These streaming tests asserted first_at < last_at / 2, which only holds when little fixed latency precedes the first chunk. On overloaded GH runners, this flakes.

Any constant delay (a cold instance, a connection coming up, or in our case a loaded runner) shifts both timestamps equally, walking the ratio toward 1 and failing a host that streamed perfectly.

They now assert the spread instead (last_at - first_at >= floor)

A buffered body arrives as one burst spanning milliseconds no matter how slow the host was to start. The postgres test also warms up before the clock starts and all three now run on a multi-threaded runtime, so setup and guest CPU work stay outside the measurement.

### DIFF
--- a/crates/wash-runtime/tests/common/mod.rs
+++ b/crates/wash-runtime/tests/common/mod.rs
@@ -13,6 +13,8 @@ pub mod tls;
 #[cfg(feature = "wasmcloud-postgres")]
 pub mod postgres;
 
+pub mod streaming;
+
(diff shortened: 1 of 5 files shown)


vados-cosmonic (approved):
LGTM 🚀 

Diff was very easy to grok, nice!

PR #5515, opened 2026-08-29
[FEAT] support concurrency for async wasmcloud:messaging

### LINKED ISSUE #5455: Messaging-triggered components ignore poolSize, maxConcurrency, and maxInvocations

ricochet:
Closes #5455.

A component invoked by the `wasmcloud:messaging` trigger was instantiated from scratch for every message and never consulted the instance pool, so the `poolSize`, `maxConcurrency` and `maxInvocations` it declared were inert on that path while reading as configured. They were only honored for inbound HTTP and for component-to-component linked calls.

## What changed

`MessagingJob` already had the shape the pool needs. Its body is bytes rather than the `stream<u8>` the WIT carries. The stream is minted inside the invocation, once the store is chosen. This allows the job to be handle-free and can go to whichever instance is free. It becomes a `InstanceJob::Messaging`, and `MessagingTask` grows the `pool_slot` that already lets `HttpTask` serve both a service singleton and a pooled instance. Both messaging backends then dispatch a delivery the way inbound HTTP does.

The driver's `serves_http: bool` becomes a bound-export set. `Service::new` and `AsyncMessaging::new` type-check their export against the live instance and hand back the view every call on it reuses.

## Scope: `@0.3.0` only

Only the async handler can be pooled. `wasmcloud:messaging/handler@0.3.0`'s `handle-message` is an `async func`, so its call takes an `Accessor` and deliveries overlap on one instance; `@0.2.0`'s takes `&mut Store` for the length of its call and cannot share an instance at all.

This also does not change the current saturation policy. A delivery arriving once every instance is busy is served from its own store rather than queued. This means we do not do per-message instantiation and it does not by itself bound concurrency. `max_in_flight` and `admission_wait` remain what bound a component's total, and they still take their permit before any store is built.

### DIFF
--- a/crates/wash-runtime/src/engine/mod.rs
+++ b/crates/wash-runtime/src/engine/mod.rs
@@ -1297,6 +1297,20 @@ pub fn exports_messaging_handler(component: &Component) -> bool {
         .any(|(export, _item)| export.starts_with("wasmcloud:messaging/handler"))
 }
 
+/// Waive the fuel limit on a store whose guest execution is measured some other
+/// way.
+///
+/// A store on a fuel-enabled engine starts at **zero** and instantiation runs
+/// guest code, so without this the component traps on instantiate. Fuel is an
+/// engine-wide switch (`--enable-meters` turns it on for the paths that do
+/// measure by it), so a path measuring by the epoch callback instead — see
+/// [`crate::observability::ExecutionTimeMeter`] — has to waive it rather than
+/// assume it is off. Errors when fuel is off, which is the ordinary case and
+/// not a failure.
+pub(crate) fn waive_fuel_limit<T>(store: &mut crate::wasmtime::Store<T>) {
+    let _ = store.set_fuel(u64::MAX);
+}
+
 pub fn imports_wasi_http(component: &Component) -> bool {
     let ty: wasmtime::component::types::Component = component.component_type();
     let engine = component.engine();
--- a/crates/wash-runtime/src/engine/instance_driver.rs
+++ b/crates/wash-runtime/src/engine/instance_driver.rs
@@ -9,13 +9,13 @@
 //! awaiting I/O is exactly the guest with that kind of state, so a pool has to
 //! be sized to peak *concurrency* to keep any of it, rather than to peak work.
 //!
-//! A driver removes that. Both an inbound HTTP request and a call from another
-//! component in the workload are [`InstanceJob`]s, so a component reached both
-//! ways shares one warm set rather than keeping two. It owns its store for good
-//! and runs one long-lived
-//! [`wasmtime::Store::run_concurrent`], taking calls off a channel and
-//! [`Accessor::spawn`]ing each as a concurrent task on the same instance. That
-//! is what the host already does for a service (see
+//! A driver removes that. An inbound HTTP request, an inbound message, a call
+//! from another component in the workload and a plugin's delivery are all
+//! [`InstanceJob`]s, so a component reached several ways shares one warm set
+//! rather than keeping one per trigger. It owns its store for good and runs one
+//! long-lived [`wasmtime::Store::run_concurrent`], taking calls off a channel
+//! and [`Accessor::spawn`]ing each as a concurrent task on the same instance.
+//! That is what the host already does for a service (see
 //! [`crate::host::trigger_service`]) — this is the same driver, one per warm
 //! instance rather than one per workload, with admission control in front.
 //!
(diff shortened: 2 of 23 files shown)


crates/wash-runtime/src/engine/mod.rs
```diff
@@ -1297,6 +1297,20 @@ pub fn exports_messaging_handler(component: &Component) -> bool {
...
+/// way.
+///
+/// A store on a fuel-enabled engine starts at **zero** and instantiation runs
+/// guest code, so without this the component traps on instantiate. Fuel is an
+/// engine-wide switch (`--enable-meters` turns it on for the paths that do
+/// measure by it), so a path measuring by the epoch callback instead — see
+/// [`crate::observability::ExecutionTimeMeter`] — has to waive it rather than
+/// assume it is off. Errors when fuel is off, which is the ordinary case and
+/// not a failure.
+pub(crate) fn waive_fuel_limit<T>(store: &mut crate::wasmtime::Store<T>) {
```
ricochet, line 1310: moved this from the nats plugin in this PR. I'm actually refactoring how this works in a separate PR.

jfleitz (dismissed):

PR #5527, opened 2026-09-02
fix(wash-runtime): wasmtime feature declaration

ricochet:
`crates/wash-runtime/Cargo.toml` pins `wasmtime` with `default-features = false` and an explicit feature list. That list was missing most of what the host actually relies on, and nothing caught it because `wasi-webgpu-wasmtime` takes `wasmtime` with **default** features and quietly put them back:

```
wash (default) → wash-runtime/wasi-webgpu → wasi-webgpu-wasmtime
               → wasmtime with DEFAULT features → parallel-compilation, gc-copying, …
```

That crate is **target-gated off Windows and s390x** (wgpu-hal does not build there). Both are release targets. So those two binaries resolved **15** wasmtime features against **32** everywhere else:

| | Linux / macOS | Windows / s390x |
|---|---|---|
| `parallel-compilation` | on | **off** one thread per compile |
| GC collector (`Collector::Auto`) | `Copying` | **`Null`** reclaims nothing |
| `addr2line`, `demangle`, `backtrace` | on | **off** |

`wasm_gc` is `true` by default in wasmtime 47, so GC guests are accepted on every target and on two of them they ran on the collector wasmtime's own docs list as "Collects Garbage: No."

## Changes

**1. Declare the full feature set, and keep it declared.** Adds `addr2line`, `backtrace`, `demangle`, `gc-copying`, `gc-drc`, `parallel-compilation`. `.github/scripts/wasmtime-features.mjs` (in `lint`) then holds two invariants:

- everything required is declared in wash-runtime's own manifest. This is what stops a dependency from owning the feature set
- everything required resolves on every target in the release matrix, read back out of that matrix.

**2. An off-switch for parallel compilation.** Measured on ten cores, release:

| component | parallel | serial | speedup | peak RSS |
|---|---|---|---|---|
| 1.9 MB / 3k fns | 0.118 s | 0.604 s | 5.1x | +7.7% |
| 19 MB / 12k fns | 0.880 s | 4.755 s | 5.4x | +2.5% |

The real cost is CPU: rayon sizes its pool to every core the process can see, which follows a cgroup quota only when `limits.cpu` is set and the chart sets `requests.cpu: 250m` with no limit. `EngineBuilder::with_parallel_compilation(false)` and `WASMTIME_PARALLEL_COMPILATION` turn it off; `RAYON_NUM_THREADS` caps width without serializing. Both reach a host group via `runtime.env`.

**3. `--no-default-features` builds again.** `HostBuilder::build` called `crate::oci::set_extra_ca_certificates` unguarded while the module is behind the `oci` feature, so the library did not compile at all without it.

Integration tests still need default features. `tests/common` builds its host from four optional plugins and 23 of the files using it carry no feature gate. I plan to pick this up in a follow-up.

**4. wasmtime 47.0.4.** Clears RUSTSEC-2026-0269 and RUSTSEC-2026-0268 Patch bump, no API or feature-set change.

### DIFF
--- a/.github/scripts/wasmtime-features.mjs
+++ b/.github/scripts/wasmtime-features.mjs
@@ -1,0 +1,130 @@
+#!/usr/bin/env node
+// Holds the host's wasmtime feature set to what wash-runtime declares, on
+// every target we ship a binary for.
+//
+// Cargo features are additive and unify across the whole graph, so a
+// dependency that takes `wasmtime` with default features turns those defaults
+// on for everyone. `wasi-webgpu-wasmtime` does exactly that — and it is
+// target-gated off Windows and s390x (wgpu-hal does not build there), so any
+// feature only it supplies is present in six release binaries and missing from
+// two. That is invisible in a normal build: the host still compiles and runs,
+// it just compiles Wasm on one thread, or picks the null garbage collector
+// that never reclaims, on the two targets nobody builds locally.
+//
+// Two invariants, both cheap — `cargo tree` resolves without compiling:
+//
+//   1. Everything in REQUIRED is declared by wash-runtime itself, not
+//      inherited. This is what stops a dependency from owning our feature set:
+//      if webgpu is the only reason a feature is on, this fails.
+//   2. Everything in REQUIRED actually resolves on every release target. This
+//      catches target-gating and any other resolution surprise.
+//
+// Features a target picks up beyond REQUIRED are reported, not failed —
+// webgpu legitimately adds a dozen on the targets it builds for.
+//
+// Adding a feature to the manifest means adding it here with its reason.
+
+import { execFileSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+
+const MANIFEST = 'crates/wash-runtime/Cargo.toml';
+// The release matrix is the authority on what we ship; reading it back means a
+// new target is covered here the moment it is added there.
+const WORKFLOW = '.github/workflows/wash.yml';
+// The package whose resolved graph decides what a released binary contains.
+const PACKAGE = 'wash';
+
+// Why each one is load-bearing. A reader who wants to drop one should have to
+// argue with the reason first.
+const REQUIRED = new Map([
+  ['addr2line', 'turns a trapping address back into guest source file and line'],
+  ['anyhow', 'wasmtime errors interoperate with the host error type'],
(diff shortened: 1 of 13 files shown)


jfleitz (dismissed):

.github/scripts/wasmtime-features.mjs
vados-cosmonic, line 1: Hey what do you think about doing this from build.rs for the wash-runtime crate? e.g.:

```rust
// subcrate/build.rs
use std::path::Path;

fn main() {
    // 1. Locate the workspace Cargo.lock (adjust parents based on depth)
    let lock_path = Path::new(env!("CARGO_MANIFEST_DIR"))
        .parent() // Up to workspace root
        .unwrap()
        .join("Cargo.lock");

    // Tell Cargo to rerun this script if the lockfile changes
    println!("cargo::rerun-if-changed={}", lock_path.display());

    if lock_path.exists() {
        let lockfile = cargo_lock::Lockfile::load(lock_path).unwrap();
        
        // 2. Find the parent crate package entry
        if let Some(package) = lockfile.packages.iter().find(|p| p.name.as_str() == "parent_crate") {
            // 3. Inspect its enabled features
            if package.features.iter().any(|f| f == "some_parent_feature") {
                // Emit custom cfg flag to your subcrate source code
                println!("cargo::rustc-cfg=parent_has_feature");
            }
        }
    }
}
```

vados-cosmonic (dismissed):

.github/scripts/wasmtime-features.mjs
ricochet: TIL Cargo.lock doesn't have feature data. Most of these features have no symbol to assert against. In wasmtime 47.0.4, addr2line, demangle, backtrace, gc-copying and gc-drc gate change behavior but don't gate public API.

But I think your instinct is that this should be in-crate enforcement. Added https://github.com/wasmCloud/wasmCloud/pull/5527/changes/71e7b05c397687bb57ecf3ce3396ff8ca16da69b which will fail the build for GC.

vados-cosmonic: Ah that's unfortunate, yeah ideally we would be able to make this check happen inside the crate itself, but it's probably not worth trying to pull the `cargo tree` functionality in here -- thanks for trying it!

vados-cosmonic (approved):

PR #5536, opened 2026-09-04
Thundering herd test suite

ricochet:
Comprehensive test suite for handling a thundering herd and a handful of fixes needed to be able to run the suite. It creates a stampede in the runtime-operator e2e tests via `reconcileScaleUp` which creates every missing Workload in one pass.  `findFreeHost` shuffles the schedulable hosts and takes the first available one with no capacity accounting. In wash-runtime we have a similar herd test that validates 15 workloads are able to start at once on one host and that compilation never stops the HTTP server from answering.

Fixes are mostly around timeouts, adding error cases, and logging:
- Wait out a NATS that isn't up yet instead of exiting and burning a pod restart on a startup race
- Same fix for the operator, infinite reconnects but no retry on the first connect.
- Hardening for the accept loop pacing added last week when a listener breaks in a way the errno list never anticipated, instead of spinning a core silently.
- Fail /livez when the ingress has stopped
- Make not-ready mean the host has no room rather than that it's busy, so a healthy host at 90% isn't pulled from the Service.
- Chart hardening where we fail the render when a host group claims a port the chart already renders.

Follow-ups after this merges:
- Update the canary compatibility list in e2e_suite_test.go:496. These only exist because the canary host image predates the flags. Once this PR merges and a new canary publishes, all three should be deleted 
-  Nothing routes host traffic through a k8s Service.  The gateway proxies to pod IPs from the Host CR, and the only EndpointSlice writer hardcodes ready := true. So /readyz sheds no traffic today

### DIFF
--- a/crates/wash-runtime/src/washlet/mod.rs
+++ b/crates/wash-runtime/src/washlet/mod.rs
@@ -507,6 +507,19 @@ pub struct NatsConnectionOptions {
     pub tls_cert: Option<PathBuf>,
     /// Path to NATS TLS private key file
     pub tls_key: Option<PathBuf>,
+    /// How long to keep retrying a refused *initial* connection before giving
+    /// up. `None` gives up on the first refusal.
+    ///
+    /// Only the first connection needs this: once established, async-nats
+    /// reconnects on its own and buffers through the gap. A host deployed
+    /// beside its NATS has no ordering guarantee between the two, so without a
+    /// window here it exits, and the pod restarts until NATS happens to be up
+    /// first — which is a slower, noisier way to wait, and it spends the
+    /// restart count that would otherwise mean something.
+    ///
+    /// Left unset by a command run against a NATS the operator already has up
+    /// (`wash dev`), where a refusal is the answer rather than a race.
+    pub connect_retry: Option<Duration>,
 }
 
 #[instrument(skip_all)]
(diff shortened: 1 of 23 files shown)


crates/wash-runtime/src/washlet/mod.rs
```diff
@@ -413,30 +435,57 @@ impl ClusterHost {
                         }
                         // OCI cache cleanup
                         _ = oci_cleanup_timer.tick() => {
-                            if let Some(cache_dir) = host.config().oci_cache_dir.as_ref() &&
-                            let Err(e) = oci::cleanup_cache(cache_dir, self.cleanup_age).await {
-                                error!("error during OCI cache cleanup: {e}");
+                            if let Some(cache_dir) = host.config().oci_cache_dir.clone()
+                                && let Ok(slot) = Arc::clone(&cleanup_slot).try_acquire_owned()
```
vados-cosmonic: NIT: I wonder if these should be the other way -- slot acquisition is more likely to fail? 

Also if we don't have an OCI cache dir, that's almost never going to change as of right now, right? (we don't yet have dynamic config yet, so can we check once and decide whether to ever do cleanups of the cache dir or not)

crates/wash-runtime/tests/integration_thundering_herd.rs
```diff
@@ -0,0 +1,335 @@
...
+    component_workload_request, default_counter_resources, get_status,
+    http_counter_host_interfaces, start_host_with_dynamic_router,
+};
+
+const HTTP_COUNTER_WASM: &[u8] = include_bytes!("wasm/http_counter.wasm");
+
+/// Enough workloads for the herd to cost something real: fifteen components to
+/// compile, fifteen plugin bindings and fifteen route registrations, all
+/// issued at once.
+const HERD: usize = 15;
```
vados-cosmonic, line 59: Q: why not 150? or 1500?

vados-cosmonic (dismissed):
LGTM 🚀

crates/wash-runtime/tests/integration_thundering_herd.rs
ricochet: Most of the bugs toppled around 10 and adding more just means longer test time. Locally I ran 100 just to see and it churned through with these fixes fine. I don't mind giving it a beefier number but it does mean a bit slower for our continuous test suite.

crates/wash-runtime/src/washlet/mod.rs
ricochet: try_acquire_owned has a side effect by taking a permit, so this does pure then effectful. Good catch on the OCI cache dir. Reworked this

crates/wash-runtime/tests/integration_thundering_herd.rs
ricochet: My original numbers:
```
┌──────────┬───────────┐
│ replicas │ converged │
├──────────┼───────────┤
│ 15       │ 6s        │
├──────────┼───────────┤
│ 50       │ 52s       │
├──────────┼───────────┤
│ 100      │ 60s       │
└──────────┴───────────┘
```

But glad you prompted me to retest because we've really improved (after the oci cache bug fix in an earlier PR): 
```
┌──────────┬───────────┬──────────────┐
│ replicas │ converged │ 3-spec block │
├──────────┼───────────┼──────────────┤
│ 15       │ 6s        │ 94.6s        │
├──────────┼───────────┼──────────────┤
│ 100      │ 29s       │ 103.7s       │
└──────────┴───────────┴──────────────┘
```

I'm opting to split the deck and go with 50.

vados-cosmonic: Oh great! Awesome I mean now that we have this tested, I'm certainly happy with going with 10/15 or whatever to keep CI fast!

vados-cosmonic (approved):
LGTM 🚀

PR #5541, opened 2026-09-07
fix(wash-dev): enable host loopback so allowedHostLoopbackPorts works

### LINKED ISSUE #5531: wash dev: components can't connect to a local network services even when its port is allowed
## Problem

On `wash dev`, a component that resolves `host.wasmcloud.internal` and connects to a port on the local machine (say Postgres on 5432) fails with `access-denied`, even when the port is listed in `workload.allowedHostLoopbackPorts`.

`wash host` supports this via `--allow-host-loopback` but `wash dev` never creates a `SocketPolicy`, so the engine uses the default one, where host loopback is off. The port list is parsed and passed through, but never consumed.

## Proposed change

`wash dev` should build a `SocketPolicy` with `host_loopback_enabled: true` and installs it with `with_socket_policy`, the same way `wash host` does in `crates/wash/src/cli/host.rs`. The already existing per-workload port list becomes the only gate:

```yaml
workload:
  allowedHostLoopbackPorts:
    - 5432
```

## Where

- `crates/wash/src/cli/dev.rs`: build the `SocketPolicy` before `engine_builder.build()` and call `with_socket_policy`.

## Done when

- With the port listed, the connect to `host.wasmcloud.internal:5432` succeeds under `wash dev`.
- Without the port listed, it continues to fail with `access-denied`.
- `localhost` and `127.0.0.1` still mean the workload's own virtual network.

shreya-d2604:
Fixes #5531

## Summary
`wash dev` never built a `SocketPolicy`, so it used the default one, which
hardcodes `host_loopback_enabled: false`, so listing a port in
`allowedHostLoopbackPorts` never actually worked — it was denied before that
list was even checked.

`wash host` already has a flag to turn this on. `wash dev` doesn't, so this
just turns it on by default. The port list is still what decides which ports
a component can actually reach.

### DIFF
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -87,6 +87,14 @@ impl CliCommand for DevCommand {
                 .with_context(|| format!("invalid dev.wasm_proposals entry {name:?}"))?;
             engine_builder = engine_builder.with_wasm_proposal(proposal);
         }
+        // Host loopback is still gated per-workload by `allowedHostLoopbackPorts`
+        // below; this just enables the master switch `wash host` exposes via
+        // `--allow-host-loopback`, since a dev session has no equivalent flag.
+        let socket_policy = Arc::new(wash_runtime::sockets::policy::SocketPolicy {
+            host_loopback_enabled: true,
+            ..Default::default()
+        });
+        engine_builder = engine_builder.with_socket_policy(socket_policy);
         let engine = engine_builder.build()?;
 
         let mut host_builder = Host::builder()
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -87,9 +87,10 @@ impl CliCommand for DevCommand {
                 .with_context(|| format!("invalid dev.wasm_proposals entry {name:?}"))?;
             engine_builder = engine_builder.with_wasm_proposal(proposal);
         }
-        // Host loopback is still gated per-workload by `allowedHostLoopbackPorts`
-        // below; this just enables the master switch `wash host` exposes via
-        // `--allow-host-loopback`, since a dev session has no equivalent flag.
+        // `wash host` needs both an operator flag and the workload's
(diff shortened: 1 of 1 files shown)


crates/wash/src/cli/dev.rs
```diff
@@ -87,6 +87,14 @@ impl CliCommand for DevCommand {
                 .with_context(|| format!("invalid dev.wasm_proposals entry {name:?}"))?;
             engine_builder = engine_builder.with_wasm_proposal(proposal);
         }
+        // Host loopback is still gated per-workload by `allowedHostLoopbackPorts`
+        // below; this just enables the master switch `wash host` exposes via
+        // `--allow-host-loopback`, since a dev session has no equivalent flag.
```
Aditya1404Sal:
```suggestion
        // `wash host` needs both an operator flag and the workload's
        // `allowedHostLoopbackPorts`. In a dev session the developer is
        // both parties, so the flag is redundant: the port list alone gates
        // it, and an empty list still denies.
```

Aditya1404Sal (commented):
Great work here @shreya-d2604 !! just a nit from from my side 😄

shreya-d2604:
Thank you for your review :D I have made changes accordingly! @Aditya1404Sal

Aditya1404Sal (approved):
LGTM 🚀

ricochet:
heads up these features aren't complete yet: https://github.com/wasmCloud/wasmCloud/pull/5441 and https://github.com/wasmCloud/wasmCloud/pull/5440 are needed

Aditya1404Sal:
@shreya-d2604 To add some more clarity, our change here is valid as merged today. It lets `wash dev` honour `allowedHostLoopbackPorts`, so a component can reach a local service through `host.wasmcloud.internal` over `wasi:sockets` (which I verified end to end). What Bailey means by "incomplete" is that the same name does not work yet over `wasi:http`. Once those aforementioned PRs land, the switch we turned on here covers HTTP too I believe.

shreya-d2604:
@Aditya1404Sal Thankss for clarifying 😃

PR #5545, opened 2026-09-08
feat: plugin dispatch target

ricochet:
This allows a host plugin to dispatch calls into a workload's Service.

1. The instance pool was `pub(crate)` but `PluginJob`, `InstanceJob`, `instance_pool_for_component` and `warm_instance_policy` were all crate-private. Found while building an out-of-tree native plugin.
2. `instantiate_pre` never looked at `self.service`. A `Service`-kind item exporting an interface a plugin calls had no instance to dispatch to. Imports were unaffected, they go through the linker like everything else, so only the export direction wasn't wired up.

A plugin resolves one `DispatchTarget` per item from `on_workload_resolved` and dispatches `GuestCall`s through it for as long as the workload runs. The host decides *where* each call runs, either on a warm instance, a store built for the call, or the running service. The plugin keeps making the call with its own generated bindings.

```rust
// once, from HostPlugin::on_workload_resolved
let target = workload.dispatch_target(item_id, self.id()).await?;

// per unit of work, for as long as the workload is up
target.dispatch(Deliver { records, reply }).await?;
```

A service is reached by a new `Ingress::Guest` on its trigger-service driver, claimed when the workload resolves. Calls land on the instance, sharing the state it has built up.

`InstanceJob` carried two variants meaning "a plugin's call": `Plugin`, the crate-private one the NATS subscriber used, and `Guest`, the public one. Now this removes `PluginJob` and replaces it with `GuestCall` so that there is one contract that the host owns, once, for every dispatched call:

- `rearm_for_call`, so the epoch deadline measures just the call.
- `watch_until_abandoned`, so a slow guest on a shared store is not condemned by another call's abandonment.

The NATS core and KV delivery jobs shed their copies of all of the above and became `GuestCall`s. This is a net deletion with no behavior changes.

### DIFF
--- a/crates/wash-runtime/src/engine/dispatch.rs
+++ b/crates/wash-runtime/src/engine/dispatch.rs
@@ -16,9 +16,10 @@
 //!    — in which case the call runs on one of them, alongside whatever that
 //!    instance already has in flight, exactly as an inbound HTTP request or a
 //!    call from another component does (see the `instance_pool` module).
-//!    That is what makes `poolSize`/`maxConcurrency` mean something here: a
-//!    burst of dispatches fans out across the warm set rather than paying for a
-//!    store apiece.
+//!    That is what makes `poolSize`, `maxInvocations` and `maxConcurrency` mean
+//!    something here: a burst of dispatches fans out across the warm set rather
+//!    than paying for a store apiece, and an instance that has served its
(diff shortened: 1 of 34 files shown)


jfleitz (approved):
LGTM and the test coverage looks good.

PR #5548, opened 2026-09-08
fix(wash-runtime): release a host's ingress when the host is dropped

ricochet:
An ingress keeps every routable workload in `workload_handles` so an inbound request can find one, and each workload reached the ingress back for its outbound HTTP, for its service registrations, etc. Held strongly in both directions so stopping a host did not free. This isn't an issue for `wash dev` or `wash host` but is an issue for host embedders.

Every handle a workload keeps for the length of its life is now `Weak`, and a strong one is taken for the length of a single call through `host::http::live_handler`, which owns the error they all report. The teardown path treats a handler that is already gone as nothing left to unbind rather than as a failure. A path that cannot proceed without one like building a store, a service re-registering its ingresses on restart, a message delivery, reports it and stops.

### DIFF
--- a/crates/wash-runtime/src/engine/ctx.rs
+++ b/crates/wash-runtime/src/engine/ctx.rs
@@ -415,7 +415,8 @@ impl wasmtime_wasi_http::p3::WasiHttpView for SharedCtx {
 
 /// HTTP hooks implementation that delegates to a [`HostHandler`](crate::host::http::HostHandler).
 struct CtxHttpHooks {
-    http_handler: Option<Arc<dyn crate::host::http::HostHandler>>,
+    /// See [`crate::host::http::live_handler`] for why this is weak.
+    http_handler: Option<std::sync::Weak<dyn crate::host::http::HostHandler>>,
     workload_id: Arc<str>,
     allowed_hosts: Arc<[AllowedHost]>,
 }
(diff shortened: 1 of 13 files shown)


crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -427,13 +428,24 @@ impl WasiHttpHooks for CtxHttpHooks {
         config: wasmtime_wasi_http::p2::types::OutgoingRequestConfig,
     ) -> wasmtime_wasi_http::p2::HttpResult<wasmtime_wasi_http::p2::types::HostFutureIncomingResponse>
     {
+        // The two `None`s are different failures and say so: no handler was
```
vados-cosmonic: I'm not sure what this actually refers to...

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -670,16 +674,18 @@ struct ServiceStoreRecipe {
...
 
 impl ServiceStoreRecipe {
-    /// Build a fresh service store (`is_service = true` so `cli/run` may bind
-    /// its loopback socket).
-    async fn build(&self) -> anyhow::Result<wasmtime::Store<SharedCtx>> {
+    /// Build a fresh store from this recipe. `is_service` is what lets
+    /// `cli/run` bind its loopback socket, so a service incarnation passes
+    /// `true`; the same recipe builds a plain multi-component store with
+    /// `false`.
+    async fn build(&self, is_service: bool) -> anyhow::Result<wasmtime::Store<SharedCtx>> {
```
vados-cosmonic: NIT: do we not have a kind of builder pattern on the recipes? 

Why is a service store ever... not for a service?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -956,6 +962,14 @@ impl ResolvedWorkload {
                 let ingresses = match first.take() {
                     Some(ingresses) => ingresses,
                     None => {
+                        // The handler is held weakly, so a host that went away
+                        // while this service was faulting ends the supervisor:
+                        // there is nothing left to re-register the restarted
+                        // incarnation's ingresses with.
+                        let Some(http_handler) = http_handler.upgrade() else {
+                            error!("host HTTP handler is no longer available; not restarting");
```
vados-cosmonic: can we say a little more about which service we're not restarting?

vados-cosmonic (dismissed):
LGTM 🚀 

nothing but nits

crates/wash-runtime/src/engine/ctx.rs
ricochet: Yuck, my bad. Rewritten

crates/wash-runtime/src/engine/workload.rs
ricochet: I deleted the flag and de-dupped the store builders.

crates/wash-runtime/src/engine/workload.rs
ricochet: Good call. Error now carries workload.id, workload.namespace, and workload.name

vados-cosmonic (approved):
LGTM 🚀

PR #5554, opened 2026-09-10
fix(runtime-operator): make 7 host status fields optional

### LINKED ISSUE #5504: runtime-operator: Host status placeholders are a workaround for required fields that should be optional
`HostStatus` marks seven observational fields as required, so the operator has to invent values for them before it can write anything else. That workaround and a related bug were touched in #5502. Making the fields optional deletes the workaround and eliminates potential racess.

`api/runtime/v1alpha1/host_types.go` gives these no `omitempty` and no `+optional`, so controller-gen emits them under `status.required`:

```yaml
required:
- osArch
- osKernel
- osName
- systemCPUUsage
- systemMemoryFree
- systemMemoryTotal
- version
```

`conditions`, `lastSeen`, `componentCount` and `workloadCount` sit in the same struct and are already optional.

## What it forces

A Host is created by Server-Side Apply from the heartbeat handler with no status at all. A merge patch that sets only `lastSeen` would leave the stored object failing its own schema, so `hostStatusPatch` supplies placeholders for whatever is still empty. `"unknown"` for the strings, `0` for the numbers.

**Race 1 cache miss.** The read is `h.client.Get`, served from the informer cache. For a Host that exists but the cache has not observed yet, it returns `NotFound`, the placeholders go in, and the status patch overwrites the measured values.

**Race 2 stale status.** The cache has the Host, but with the status as it was before `reconcileReporting` wrote to it. Not `NotFound`, empty fields, same outcome.

Both are self-healing within one `hostReconcileInterval` (60s) and none of these fields feeds a decision: `findFreeHost` keys on conditions, labels and environment, no printcolumn references them, and `HostReconciler.CPUThreshold`/`MemoryThreshold` are plumbed from config and then never read. So this is a status-accuracy problem, not a scheduling one. But since this would be a change to CRD's I'd rather queue this with any other changes we need to make (ideally real bug fixes). The good news is that relaxing `required` is a backward-compatible schema change and the Go struct is unchanged apart from the tags.

## The fix

Mark the seven fields `+optional` with `omitempty`, matching `componentCount` and `workloadCount`. Then:

- `hostStatusPatch` loses its seven `if ... == ""` branches and becomes a `lastSeen` patch.
- The stored-status read exists only for `hostSpecChanged`, and neither race can clobber anything, because nothing is written that was not measured.
- A host that has not been polled yet reports empty fields rather than the string `"unknown"`, which is the honest representation and distinguishable from a host that genuinely reports `"unknown"`.

steverogersX:
Earlier code force these fields (osArch, osName, osKernel, version, systemCPUUsage, systemMemoryTotal, systemMemoryFree) to be required in the CRD. Because of that, heartbeat handler had to fill fake values like "unknown" or 0 whenever cache was not updated yet, and it was overwriting real values by mistake.

Now made them optional (omitempty), so heartbeat handler only updates lastSeen and does not touch these fields anymore. Real values are only written by reconcileReporting, so no more overwrite issue.

Fixes #5504

### DIFF
--- a/runtime-operator/api/runtime/v1alpha1/host_types.go
+++ b/runtime-operator/api/runtime/v1alpha1/host_types.go
@@ -13,13 +13,13 @@ const (
 type HostStatus struct {
 	condition.ConditionedStatus `json:",inline"`
 
-	Version           string `json:"version"`
-	OSName            string `json:"osName"`
-	OSArch            string `json:"osArch"`
-	OSKernel          string `json:"osKernel"`
-	SystemCPUUsage    string `json:"systemCPUUsage"`
-	SystemMemoryTotal int64  `json:"systemMemoryTotal"`
-	SystemMemoryFree  int64  `json:"systemMemoryFree"`
+	Version           string `json:"version,omitempty"`
+	OSName            string `json:"osName,omitempty"`
+	OSArch            string `json:"osArch,omitempty"`
+	OSKernel          string `json:"osKernel,omitempty"`
+	SystemCPUUsage    string `json:"systemCPUUsage,omitempty"`
+	SystemMemoryTotal int64  `json:"systemMemoryTotal,omitempty"`
+	SystemMemoryFree  int64  `json:"systemMemoryFree,omitempty"`
 	ComponentCount    int    `json:"componentCount,omitempty"`
 	WorkloadCount     int    `json:"workloadCount,omitempty"`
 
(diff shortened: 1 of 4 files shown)


ricochet:
Thank you for taking this! It looks correct but I'm going to hold on merging to combine this with another CRD change. I'd like to upgrade us from alpha to beta version to basically lock in this schema.

PR #5557, opened 2026-09-11
fix(gateway): stop trusting client X-Forwarded-For for X-Real-IP

immanuwell:
`rewrite()` sets `X-Real-IP` from the client's own `X-Forwarded-For` header. 
that's client controlled, so any request can spoof it, even though the README calls `X-Real-Ip` "the canonical IP address of the client". 
with no `X-Forwarded-For` it falls back to `req.In.RemoteAddr`, which includes the port, so it's wrong on plain requests too

repro:
```
curl -H "X-Forwarded-For: 10.0.0.1" http://gateway:8000/   # X-Real-IP: 10.0.0.1, attacker chosen
curl http://gateway:8000/                                  # X-Real-IP: 1.2.3.4:54321, port included
```

fix: derive `X-Real-IP` from the actual tcp peer via `net.SplitHostPort(req.In.RemoteAddr)`, same as stdlib's `SetXForwarded()` does for `X-Forwarded-For`. 
`X-Forwarded-For` handling itself is untouched

added tests for both cases, checked they fail on old code and pass now. 
`go build`, `go vet`, `gofmt`, `go test ./...` all clean

### DIFF
--- a/runtime-gateway/proxy.go
+++ b/runtime-gateway/proxy.go
@@ -79,13 +79,16 @@ func (h *HTTPGateway) Start(ctx context.Context) error {
 }
 
 func (h *HTTPGateway) rewrite(req *httputil.ProxyRequest) {
-	// X-Forwarded-For handling
-	clientIP := req.In.RemoteAddr
+	// Derive X-Real-IP from the TCP peer, not client-controlled headers.
+	if clientIP, _, err := net.SplitHostPort(req.In.RemoteAddr); err == nil {
+		req.Out.Header.Set("X-Real-IP", clientIP)
+	} else {
+		req.Out.Header.Del("X-Real-IP")
+	}
+
 	if xff := req.In.Header.Get("X-Forwarded-For"); xff != "" {
-		clientIP = xff
 		req.Out.Header.Set("X-Forwarded-For", xff)
 	}
-	req.Out.Header.Set("X-Real-IP", clientIP)
 	req.SetXForwarded()
 
 	// Preserve Connection header from the original request
--- a/runtime-gateway/proxy_test.go
+++ b/runtime-gateway/proxy_test.go
@@ -132,6 +132,44 @@ func TestStartWaitsForInFlightRequests(t *testing.T) {
 	}
 }
 
+type fakeResolver struct{ result LookupResult }
+
+func (f fakeResolver) Resolve(context.Context, *http.Request) LookupResult { return f.result }
+
+func newProxyRequest(remoteAddr string, xForwardedFor string) *httputil.ProxyRequest {
+	in := httptest.NewRequest(http.MethodGet, "http://example.com/", nil)
+	in.RemoteAddr = remoteAddr
+	if xForwardedFor != "" {
+		in.Header.Set("X-Forwarded-For", xForwardedFor)
+	}
+	return &httputil.ProxyRequest{In: in, Out: in.Clone(context.Background())}
+}
+
+func TestRewriteRealIPIgnoresClientSuppliedForwardedFor(t *testing.T) {
+	gw := &HTTPGateway{Resolver: fakeResolver{result: LookupResult{Hostname: "backend:8080", Scheme: "http"}}}
+
+	pr := newProxyRequest("203.0.113.7:54321", "10.0.0.1, 192.168.1.1")
+	gw.rewrite(pr)
+
+	if got, want := pr.Out.Header.Get("X-Real-IP"), "203.0.113.7"; got != want {
+		t.Errorf("X-Real-IP = %q, want the actual peer %q, not the spoofable X-Forwarded-For", got, want)
+	}
+	if got, want := pr.Out.Header.Get("X-Forwarded-For"), "10.0.0.1, 192.168.1.1, 203.0.113.7"; got != want {
+		t.Errorf("X-Forwarded-For = %q, want %q", got, want)
+	}
+}
+
+func TestRewriteRealIPHasNoPort(t *testing.T) {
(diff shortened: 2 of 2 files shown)


runtime-gateway/proxy.go
```diff
@@ -79,13 +79,16 @@ func (h *HTTPGateway) Start(ctx context.Context) error {
...
+	if clientIP, _, err := net.SplitHostPort(req.In.RemoteAddr); err == nil {
+		req.Out.Header.Set("X-Real-IP", clientIP)
+	} else {
+		req.Out.Header.Del("X-Real-IP")
+	}
+
 	if xff := req.In.Header.Get("X-Forwarded-For"); xff != "" {
-		clientIP = xff
 		req.Out.Header.Set("X-Forwarded-For", xff)
 	}
```
ricochet, line 91: `Header.Get` returns only the first `X-Forwarded-For` line. When the client sent no `X-Forwarded-For`, the copy is nil and `SetXForwarded` sets just the peer.

```suggestion
	// Copying the slice keeps every line; SetXForwarded joins them and appends the peer.
	req.Out.Header["X-Forwarded-For"] = req.In.Header["X-Forwarded-For"]
```

runtime-gateway/proxy_test.go
```diff
@@ -132,6 +132,44 @@ func TestStartWaitsForInFlightRequests(t *testing.T) {
...
+func TestRewriteRealIPHasNoPort(t *testing.T) {
+	gw := &HTTPGateway{Resolver: fakeResolver{result: LookupResult{Hostname: "backend:8080", Scheme: "http"}}}
+
+	pr := newProxyRequest("203.0.113.7:54321", "")
+	gw.rewrite(pr)
+
+	if got, want := pr.Out.Header.Get("X-Real-IP"), "203.0.113.7"; got != want {
+		t.Errorf("X-Real-IP = %q, want %q", got, want)
+	}
+}
```
ricochet, line 171: No test sends a client `X-Real-IP`. `ReverseProxy` strips client `X-Forwarded-*` headers before `Rewrite` runs but leaves `X-Real-IP`, so the `Set` in `rewrite` is the only thing replacing a client's own value, and no test would notice if it stopped. This suggestion:

- has `newProxyRequest` take an `http.Header`, so a test can send any client header (`Add` canonicalizes the names),
- adds `TestRewriteRealIPReplacesClientSuppliedRealIP`, which checks for exactly one `X-Real-IP` value and that it's the peer, and
- adds `TestRewriteKeepsEveryForwardedForLine` for the `proxy.go` suggestion.

I checked that each new test fails without the change it covers.

```suggestion
func newProxyRequest(remoteAddr string, header http.Header) *httputil.ProxyRequest {
	in := httptest.NewRequest(http.MethodGet, "http://example.com/", nil)
	in.RemoteAddr = remoteAddr
	for name, values := range header {
		for _, value := range values {
			in.Header.Add(name, value)
		}
	}
	return &httputil.ProxyRequest{In: in, Out: in.Clone(context.Background())}
}

func TestRewriteRealIPIgnoresClientSuppliedForwardedFor(t *testing.T) {
	gw := &HTTPGateway{Resolver: fakeResolver{result: LookupResult{Hostname: "backend:8080", Scheme: "http"}}}

	pr := newProxyRequest("203.0.113.7:54321", http.Header{"X-Forwarded-For": {"10.0.0.1, 192.168.1.1"}})
	gw.rewrite(pr)

	if got, want := pr.Out.Header.Get("X-Real-IP"), "203.0.113.7"; got != want {
		t.Errorf("X-Real-IP = %q, want the actual peer %q, not the spoofable X-Forwarded-For", got, want)
	}
	if got, want := pr.Out.Header.Get("X-Forwarded-For"), "10.0.0.1, 192.168.1.1, 203.0.113.7"; got != want {
		t.Errorf("X-Forwarded-For = %q, want %q", got, want)
	}
}

func TestRewriteRealIPHasNoPort(t *testing.T) {
	gw := &HTTPGateway{Resolver: fakeResolver{result: LookupResult{Hostname: "backend:8080", Scheme: "http"}}}

	pr := newProxyRequest("203.0.113.7:54321", nil)
	gw.rewrite(pr)

	if got, want := pr.Out.Header.Get("X-Real-IP"), "203.0.113.7"; got != want {
		t.Errorf("X-Real-IP = %q, want %q", got, want)
	}
}

// ReverseProxy strips client X-Forwarded-* headers but not X-Real-IP, so a
// client's own X-Real-IP reaches rewrite and has to be replaced there.
func TestRewriteRealIPReplacesClientSuppliedRealIP(t *testing.T) {
	gw := &HTTPGateway{Resolver: fakeResolver{result: LookupResult{Hostname: "backend:8080", Scheme: "http"}}}

	pr := newProxyRequest("203.0.113.7:54321", http.Header{"X-Real-IP": {"6.6.6.6"}})
	gw.rewrite(pr)

	if got := pr.Out.Header.Values("X-Real-IP"); len(got) != 1 || got[0] != "203.0.113.7" {
		t.Errorf("X-Real-IP = %q, want only the actual peer %q", got, "203.0.113.7")
	}
}

func TestRewriteKeepsEveryForwardedForLine(t *testing.T) {
	gw := &HTTPGateway{Resolver: fakeResolver{result: LookupResult{Hostname: "backend:8080", Scheme: "http"}}}

	pr := newProxyRequest("203.0.113.7:54321", http.Header{"X-Forwarded-For": {"10.0.0.1", "192.168.1.1"}})
	gw.rewrite(pr)

	if got, want := pr.Out.Header.Get("X-Forwarded-For"), "10.0.0.1, 192.168.1.1, 203.0.113.7"; got != want {
		t.Errorf("X-Forwarded-For = %q, want %q", got, want)
	}
}
```

PR #5564, opened 2026-09-12
Fix/messaging gate lifetime

### LINKED ISSUE #5461: Follow-up to #5458: gate lifetime, in-flight tests, and other review-callouts
## 1. A component gate can be dropped while its handlers still hold permits

**This is the substantive one.** Everything else below is polish.

`ComponentGates::release` closes and removes the gate as soon as `bindings` hits zero
(`crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs:703`), but the backends'
`component_cleanup` does **not** wait for in-flight handler tasks to finish. Each handler
task owns its permit for the duration of the guest call, so at cleanup time there can be
handlers still running against a semaphore whose gate entry is already gone.

If a workload with the same `AdmissionIdentity` (namespace / workload / component) binds
again before those handlers drain, `ComponentGates::acquire` finds no entry and mints a
**fresh** semaphore with a full set of permits. The old handlers are still holding permits
from the *old* semaphore, so old + new in-flight can exceed `max_in_flight` for that
component until the stragglers finish.

> the old handlers hold permits from the *old* semaphore, so old and new handlers together
> can go past `max_in_flight`.
> — @vados-cosmonic

The host-wide semaphore is unaffected — it is a single `Arc<Semaphore>` for the host's
lifetime, so old handlers still count against it. The breach is per-component only.

### Two ways out (either is acceptable)

1. **Keep the gate alive until all outstanding permits drain.** Refcount by permit rather
   than by binding, or defer the `gates.remove` until `available_permits() == limit`, so a
   re-bind under the same identity reuses the existing semaphore.
2. **Wait for active handler tasks during `component_cleanup`.** Track the spawned handler
   tasks (`JoinSet` or a task counter) and drain them before releasing the binding. Costs a
   bounded wait on teardown; note that a handler parked mid-call means cleanup blocks up to
   the guest's own timeout, so this needs a deadline of its own.

Option 1 is the smaller change and does not add a teardown stall.

### Acceptance

- A test that binds identity `X`, admits up to the ceiling with handlers that do not return,
  tears the binding down, re-binds `X`, and asserts the total number of concurrently-running
  handlers never exceeds `max_in_flight`.
- Covered for both the NATS and in-memory backends.

### Related (not from this review, same code path)

@LiamRandall's rolling-redeploy note lands on the same `ComponentGates` structure from the
other direction: with `replicas: 2+` rolling one at a time, `bindings` never reaches zero, so
a changed `max_in_flight` never takes effect and the "two genuinely different components
collided on one name" warning misfires on the benign case. A fix that resizes via
`add_permits` / `forget_permits` deltas would address both; at minimum the warning wording
should mention config rollout. Worth solving together if the approach allows.

---

## 2. Tests do not actually prove the ceiling holds

`crates/wash-runtime/tests/integration_nats_messaging.rs`

> It looks like we need a test in here w/ handlers that just never return -- to ensure that
> we never actually go past the in flight limit concretely. It looks like the current tests
> make sure that messages are delivered but do not ensure that we actually don't go past the
> limit for processing.
> — @vados-cosmonic

The existing burst tests assert that a burst larger than the ceiling is delivered in full.
That catches a leaked permit or a wedged loop, but it does not observe concurrency, so it
would pass with the ceiling ignored entirely.

- [ ] Add a test whose handler blocks/never resolves, and assert the count of simultaneously
      live handlers is capped at the configured ceiling (per-component and host-wide).
- [ ] Add integration coverage for the **in-memory** backend, not just NATS — same property,
      both paths.

---

## 3. Naming

| Where | Current | Suggested |
| --- | --- | --- |
| `in_memory.rs:40` | `QueuedMessage::sole_subscriber` | `was_sole_subscriber` — it is past state (recorded at routing time), and the name should say so |
| `mod.rs:487` | `DEFAULT_ADMISSION_WAIT` | `DEFAULT_SHED_AFTER` |
| `mod.rs:15` (doc table) | `admission_wait` | `shed_incoming_after` |

The third one renames a **user-facing config key**, not just a constant. If we take it, it
should land before anyone is depending on `admission_wait` — and it needs the flag/env
surface, the module docs, and the `wash dev` config path updated together. Flag this at
review so it is a deliberate call rather than a nit that silently breaks a manifest.

---

## 4. Group the pool-share numerator and denominator

`mod.rs:519-520`

```rust
const MESSAGING_POOL_SHARE_NUM: u32 = 2;
const MESSAGING_POOL_SHARE_DEN: u32 = 3;
```

> These numbers should probably just be together in a tuple I think, intended to be used
> together as essentially a quotient
> — @vados-cosmonic

They are only ever meaningful as a pair. A tuple const, a small `Ratio`-ish struct, or an
enum if we want to express the share more than one way. A named type also gives the
derivation somewhere to hang its invariant (numerator ≤ denominator).

---

## 5. UUIDv7 for component ids

`mod.rs:611` (doc comment referencing `uuid::Uuid::new_v4()`)

> It might be useful here to use a v7 for time relationship
> — @vados-cosmonic

v7 ids sort by creation time, which makes replica/workload ids correlatable in logs and
traces. This is broader than the messaging plugin — the id generation lives outside this
file — so it likely wants its own issue if we pursue it. Noting it here so it is not lost.

---

## 6. Trivial

`mod.rs:1508` — @vados-cosmonic suggested dropping the `// --- Admission ceilings ---`
section-divider comment in the test module. Sweep it in with whichever change touches that
file.

---

## 7. From testing

Either this is an issue in the coding or documentation. We need to verify that the consumer_group override works with this so that only 1 workload processes vs everyone (broadcast). This sounds more like a documentation update, but should confirm. 

Notes:

The host debug log shows four distinct identities, each opening its own gate at the full ceiling:

```
workload.name="subject1-handler-go-597cb4ddcb-69f68d8984"
workload.name="subject1-handler-go-597cb4ddcb-6d5654c559"
workload.name="subject1-handler-go-597cb4ddcb-7f4c8f86d6"
workload.name="subject1-handler-go-597cb4ddcb-cc49d7ddb"
   4 max_in_flight=32
```

### Root cause
AdmissionIdentity::new is fed component_handle.workload_name() ([nats.rs:443-447](vscode-webview://0eb74g0socd1is69o828ec3ce92m7ac59b48j1vmodkfku1ujgr3/crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs#L443-L447)). Under the k8s operator that is the per-replica Workload CR name <replicaset>-<randHash>, not the deployment name — so the key that was supposed to collapse replicas onto one gate is precisely the field that differs between them.

This is the same root cause as the per-replica default consumer group in §2.2 of your design doc. Note the example dodges that one by pinning consumer_group explicitly, and message distribution across the four replicas worked correctly — but admission identity has no equivalent override, so there's no workaround at the manifest level.

The code's own doc comment states the failure it was preventing: "four replicas of a component at max_in_flight: 32 could hold 128 messages on one host, against a stock host-wide total of 133." That is exactly what I measured. The host-wide ceiling still contained it — nothing was lost and no pool error fired — so the blast radius is bounded, but the per-component ceiling is not doing what a manifest author reads it to mean.

A fix needs a stable manifest identity reaching the host — the replicaset name, or an operator-supplied label/annotation — rather than the replica name. That's a wasmCloud-side change plus possibly an operator change; I haven't written it, and per your standing instruction I've made no GitHub changes.

## Checklist

- [ ] Gate outlives its permits (or cleanup drains handlers) + test — **§1**
- [ ] Never-returning-handler concurrency test, NATS — **§2**
- [ ] In-memory backend integration tests — **§2**
- [ ] `was_sole_subscriber` rename — **§3**
- [ ] Decide on `DEFAULT_SHED_AFTER` / `shed_incoming_after` rename (user-facing) — **§3**
- [ ] Pool share as a single value — **§4**
- [ ] UUIDv7 — separate issue? — **§5**
- [ ] Drop the divider comment — **§6**

steverogersX:
Fixes #5461.

## Checklist

- [x] Gate outlives its permits (or cleanup drains handlers) + test — **§1**
- [x] Never-returning-handler concurrency test, NATS — **§2**
- [x] In-memory backend integration tests — **§2**
- [x] `was_sole_subscriber` rename — **§3**
- [x] Decide on `DEFAULT_SHED_AFTER` / `shed_incoming_after` rename (user-facing) — **§3**
- [x] Pool share as a single value — **§4**
- [ ] UUIDv7 — separate issue? — **§5**
- [x] Drop the divider comment — **§6**

### DIFF
--- a/crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
+++ b/crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
@@ -36,7 +36,7 @@ struct QueuedMessage {
     /// rather than recomputed in the drain loop: by then the subscriber set may
     /// have changed, and the answer that matters is the one that was true when
     /// the message fanned out.
-    sole_subscriber: bool,
+    was_sole_subscriber: bool,
 }
 
 mod bindings {
@@ -157,7 +157,7 @@ async fn route_to_subscribers(
             .collect()
     };
 
-    let sole_subscriber = targets.len() == 1;
+    let was_sole_subscriber = targets.len() == 1;
     for (inbox, notify) in targets {
         {
             let mut queue = inbox.write().await;
@@ -168,7 +168,7 @@ async fn route_to_subscribers(
             }
             queue.push_back(QueuedMessage {
                 msg: msg.clone(),
-                sole_subscriber,
+                was_sole_subscriber,
             });
         }
         notify.notify_one();
@@ -548,7 +548,7 @@ impl HostPlugin for InMemoryMessaging {
             component_handle
                 .local_resources()
                 .config
-                .get(super::ADMISSION_WAIT_CONFIG)
+                .get(super::SHED_INCOMING_AFTER_CONFIG)
                 .map(String::as_str),
         );
 
@@ -672,7 +672,7 @@ impl HostPlugin for InMemoryMessaging {
                         loop {
                         let queued = inbox.write().await.pop_front();
 
-                        let Some(QueuedMessage { msg, sole_subscriber }) = queued else {
+                        let Some(QueuedMessage { msg, was_sole_subscriber }) = queued else {
                             break;
                         };
 
@@ -693,7 +693,7 @@ impl HostPlugin for InMemoryMessaging {
                                 "host is gone and this component has no per-message \
                                  instance; ending the in-memory receive loop"
                             );
-                            if sole_subscriber
+                            if was_sole_subscriber
                                 && let (Some(reply_to), Some(pending)) =
                                     (&msg.reply_to, &pending_requests)
                                 && let Some(sender) = pending.write().await.remove(reply_to)
@@ -748,7 +748,7 @@ impl HostPlugin for InMemoryMessaging {
                         // built and held until the handler returns. Mirrors
                         // the NATS backend exactly; see `Admission::acquire`
                         // for why the component level is taken before the host
-                        // one, and `DEFAULT_ADMISSION_WAIT` for why the wait is
+                        // one, and `DEFAULT_SHED_AFTER` for why the wait is
                         // bounded. Here the bound also keeps the inbox draining:
                         // a loop parked forever lets the inbox reach
                         // `MAX_QUEUE_SIZE`, at which point a guest's `publish`
@@ -770,7 +770,7 @@ impl HostPlugin for InMemoryMessaging {
                                     // request here would fail a request that
                                     // was about to succeed.
                                     super::Admitted::Shed => {
-                                        if sole_subscriber
+                                        if was_sole_subscriber
                                             && let (Some(reply_to), Some(pending)) =
                                                 (&msg.reply_to, &pending_requests)
                                             && let Some(sender) =
@@ -995,7 +995,7 @@ mod tests {
             .await
             .expect("routes");
         assert!(
-            inboxes[0].read().await[0].sole_subscriber,
+            inboxes[0].read().await[0].was_sole_subscriber,
             "one matching component must be marked as the sole subscriber"
         );
 
@@ -1006,7 +1006,7 @@ mod tests {
             .expect("routes");
         for (i, inbox) in inboxes.iter().enumerate() {
             assert!(
-                !inbox.read().await[0].sole_subscriber,
+                !inbox.read().await[0].was_sole_subscriber,
                 "component {i} shares the message and must not fail the requester"
             );
         }
@@ -1016,7 +1016,7 @@ mod tests {
         route_to_subscribers(&plugin, &workload_id, &broker("tasks.new"))
             .await
             .expect("routes");
-        assert!(inboxes[0].read().await[0].sole_subscriber);
+        assert!(inboxes[0].read().await[0].was_sole_subscriber);
         assert!(
             inboxes[1].read().await.is_empty(),
             "a non-matching component must not receive the message at all"
--- a/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
+++ b/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
@@ -668,12 +668,13 @@ pub const MAX_ADMISSION_WAIT: Duration = Duration::from_secs(600);
 /// the pool can hold" into "messages we may admit".
 const WORST_CASE_CORE_INSTANCES_PER_COMPONENT: u32 = 5;
 
-/// Share of the pool's component capacity messaging may claim. The remainder is
-/// what HTTP-triggered work, warm pools, and long-lived services draw on — the
-/// workloads that otherwise fail to *start* when a messaging burst drains the
-/// pool.
-const MESSAGING_POOL_SHARE_NUM: u32 = 2;
-const MESSAGING_POOL_SHARE_DEN: u32 = 3;
+/// Share of the pool's component capacity messaging may claim, as
+/// `(numerator, denominator)` — the two are only ever meaningful together as
+/// one fraction, so they are one constant rather than two that must be kept
+/// in sync by hand. The remainder is what HTTP-triggered work, warm pools,
+/// and long-lived services draw on — the workloads that otherwise fail to
+/// *start* when a messaging burst drains the pool.
+const MESSAGING_POOL_SHARE: (u32, u32) = (2, 3);
 
 /// Simultaneously-saturated components a host should fit before the host-wide
 /// ceiling is what binds. Deriving the per-component default from the host
@@ -709,7 +710,8 @@ fn derive_host_ceiling(total_core_instances: Option<u32>) -> usize {
     // No overflow: `capacity` is at most `u32::MAX / 5`, so the multiply stays
     // inside `u32`.
     let capacity = total / WORST_CASE_CORE_INSTANCES_PER_COMPONENT;
-    let share = capacity * MESSAGING_POOL_SHARE_NUM / MESSAGING_POOL_SHARE_DEN;
+    let (share_num, share_den) = MESSAGING_POOL_SHARE;
+    let share = capacity * share_num / share_den;
     let capacity = usize::try_from(capacity).unwrap_or(MAX_DERIVED_IN_FLIGHT);
     let share = usize::try_from(share).unwrap_or(MAX_DERIVED_IN_FLIGHT);
     // The floor may not raise the ceiling past what the pool actually holds.
@@ -854,7 +856,7 @@ impl ComponentGates {
             return;
         };
         entry.bindings = entry.bindings.saturating_sub(1);
-        if entry.bindings == 0 {
+        if entry.bindings == 0 && entry.semaphore.available_permits() == entry.limit {
             entry.semaphore.close();
             gates.remove(identity);
         }
--- a/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
+++ b/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
@@ -1663,8 +1663,6 @@ mod tests {
         assert!(parse_subscriptions(None).is_empty());
     }
 
-    // --- Admission ceilings -------------------------------------------------
-
     use super::{
         Admitted, DEFAULT_MAX_IN_FLIGHT_HOST, DEFAULT_MAX_IN_FLIGHT_PER_COMPONENT, MessagingLimits,
     };
--- a/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
+++ b/crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
@@ -12,14 +12,14 @@
 //! | `subscriptions` | Comma-separated subjects, NATS wildcards allowed (`orders.*`, `audit.>`) | Receive everything |
 //! | `consumer_group` | Queue-group name, or `broadcast` for no grouping (NATS only) | A name derived from namespace/workload/component |
 //! | `max_in_flight` | Messages this component may process at once, across every replica of it on this host | The host's per-component default |
-//! | `admission_wait` | How long to wait for a slot before shedding (`45s`, `2m`, or bare seconds) | [`DEFAULT_ADMISSION_WAIT`] |
+//! | `shed_incoming_after` | How long to wait for a slot before shedding (`45s`, `2m`, or bare seconds) | [`DEFAULT_SHED_AFTER`] |
 //!
 //! ```yaml
 //! localResources:
 //!   config:
 //!     subscriptions: "orders.*.created,audit.>"
 //!     max_in_flight: "64"
-//!     admission_wait: "5m"     # this handler is slow on purpose
+//!     shed_incoming_after: "5m"     # this handler is slow on purpose
 //! ```
 //!
 //! # Instance limits
@@ -36,7 +36,7 @@
 //! see [`MessagingLimits`]. It is a total for the *component*, not for one
 //! replica of it: replicas of a deployment that land on the same host share
 //! one ceiling rather than getting one apiece, so `replicas: 4` with
-//! `max_in_flight: "32"` admits 32 messages on a host, not 128. `admission_wait`
+//! `max_in_flight: "32"` admits 32 messages on a host, not 128. `shed_incoming_after`
 //! is per component rather than per host because the right answer depends on
 //! the handler: minutes-long work wants to queue, interactive work wants to
 //! shed and stay responsive.
@@ -640,9 +640,9 @@ pub const DEFAULT_MAX_IN_FLIGHT_HOST: usize = 128;
 /// waited this long is behind work that is not clearing — and well past typical
 /// guest `timeout_ms` values, so the guest's own deadline fires first in the
 /// request/reply case and this one only catches genuine saturation.
-pub const DEFAULT_ADMISSION_WAIT: Duration = Duration::from_secs(30);
+pub const DEFAULT_SHED_AFTER: Duration = Duration::from_secs(30);
 
-/// The longest [`ADMISSION_WAIT_CONFIG`] may set the wait to.
+/// The longest [`SHED_INCOMING_AFTER_CONFIG`] may set the wait to.
 ///
 /// Waiting is a real choice — a handler whose work legitimately takes minutes
 /// wants its messages queued, not shed — but it is bought with the transport's
@@ -940,14 +940,14 @@ impl MessagingLimits {
             host: Arc::new(Semaphore::new(host_total)),
             host_total,
             per_component_default,
-            admission_wait: DEFAULT_ADMISSION_WAIT,
+            admission_wait: DEFAULT_SHED_AFTER,
             timeouts: AdmissionTimeouts::new(),
             gates: ComponentGates::default(),
         }
     }
 
     /// Override how long a subscriber loop waits for a slot before shedding the
-    /// message it is holding. See [`DEFAULT_ADMISSION_WAIT`].
+    /// message it is holding. See [`DEFAULT_SHED_AFTER`].
     ///
     /// Held to [`MAX_ADMISSION_WAIT`], the same bound the per-component config
     /// key takes. Clamping only at the config layer would leave this — the way
@@ -1291,7 +1291,7 @@ impl Admission {
     }
 
     /// Override how long this component's loop waits before shedding, from its
-    /// [`ADMISSION_WAIT_CONFIG`] entry. `None` leaves the host default in
+    /// [`SHED_INCOMING_AFTER_CONFIG`] entry. `None` leaves the host default in
     /// place.
     ///
     /// Clamped to [`MAX_ADMISSION_WAIT`] like every other way of setting a
@@ -1402,7 +1402,7 @@ pub(crate) const MAX_IN_FLIGHT_CONFIG: &str = "max_in_flight";
 /// component whose work legitimately takes minutes wants to wait, and one
 /// serving interactive traffic wants to shed early and stay responsive. A
 /// single host-wide number cannot be right for both, and
-/// [`DEFAULT_ADMISSION_WAIT`] alone would turn a slow handler's queued messages
+/// [`DEFAULT_SHED_AFTER`] alone would turn a slow handler's queued messages
 /// into dropped ones with no way to say otherwise.
 ///
 /// **What raising it costs.** A loop waiting for a slot is not draining its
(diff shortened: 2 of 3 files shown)


steverogersX:
UUIDv7 (originally §5 here) was split out into #5566 , not included in this PR.

steverogersX:
Hi @jfleitz,

 I need a bit of help with the race issue you mentioned. I fixed the race, but a test in `d:\wasmCloud\crates\wash-runtime\src\plugin\wasmcloud_messaging\mod.rs` — `closing_the_gate_wakes_a_loop_already_parked_on_it` — is now failing, and I’m stuck on how to resolve it.

 The test covers the component shutdown case: all jobs parked on the admission gate should be cancelled and woken with `Closed`. However, with my fix, the semaphore only closes when `bindings == 0` **and** all permits have been returned.

 Specifically, the change is at **line 859** in `mod.rs`, inside `ComponentGates::release`:

```
if entry.bindings == 0 && entry.semaphore.available_permits() == entry.limit {
    entry.semaphore.close();
    gates.remove(identity);
}
```

 The test exercises the gate while it’s full. There are still in-flight jobs holding permits, so `available_permits() == entry.limit` is false. As a result, the semaphore never closes, and the parked jobs are never woken.

 This seems to put two requirements in direct conflict:

 The test says → on shutdown, the semaphore **must close** so that parked jobs are woken.

 My fix says → the semaphore should only close once all permits have been returned, in order to preserve the `max_in_flight` limit across rebinds.

 I’m not sure what the right way to reconcile these two requirements is. Satisfying the test seems to break the limit guarantee my fix was intended to protect, while keeping the current behavior breaks the shutdown contract.

steverogersX:
Hi @ricochet, ready for review.

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -1412,14 +1570,14 @@ pub(crate) const MAX_IN_FLIGHT_CONFIG: &str = "max_in_flight";
 /// trade for a handler that genuinely needs minutes and the wrong one as a
 /// reflex against shed warnings, where the answer is `max_in_flight` or fewer
 /// messages. Bounded by [`MAX_ADMISSION_WAIT`] so it cannot become "forever".
-pub(crate) const ADMISSION_WAIT_CONFIG: &str = "admission_wait";
+pub(crate) const SHED_INCOMING_AFTER_CONFIG: &str = "shed_incoming_after";
```
ricochet: `admission_wait` shipped in v2.8.0 and v2.9.0, and this renames it without reading the old name. A manifest with `admission_wait: "5m"` (the slow-handler example from the module docs) quietly falls back to the 30s default after upgrading. The only log is the per-message shed warning, which doesn't mention the ignored key.

Could we keep the old name as a deprecated alias?

```suggestion
pub(crate) const SHED_INCOMING_AFTER_CONFIG: &str = "shed_incoming_after";

/// The name [`SHED_INCOMING_AFTER_CONFIG`] shipped under in v2.8.0 and v2.9.0.
/// Still read, with a warning, so an upgrade does not drop a configured wait
/// back to [`DEFAULT_SHED_AFTER`].
pub(crate) const DEPRECATED_ADMISSION_WAIT_CONFIG: &str = "admission_wait";

/// Looks up the shed wait in one config map, preferring
/// [`SHED_INCOMING_AFTER_CONFIG`] over [`DEPRECATED_ADMISSION_WAIT_CONFIG`].
pub(crate) fn shed_incoming_after_config(
    config: &std::collections::HashMap<String, String>,
) -> Option<&str> {
    match (
        config.get(SHED_INCOMING_AFTER_CONFIG),
        config.get(DEPRECATED_ADMISSION_WAIT_CONFIG),
    ) {
        (Some(value), _) => Some(value),
        (None, Some(value)) => {
            tracing::warn!(
                deprecated = DEPRECATED_ADMISSION_WAIT_CONFIG,
                replacement = SHED_INCOMING_AFTER_CONFIG,
                "messaging config key is deprecated; rename it"
            );
            Some(value)
        }
        (None, None) => None,
    }
}
```

Both backends then read through the helper (see the suggestions in `nats.rs` and `in_memory.rs`). A test:

```rust
#[test]
fn the_released_admission_wait_key_is_still_read() {
    let config = |pairs: &[(&str, &str)]| {
        pairs
            .iter()
            .map(|(k, v)| (k.to_string(), v.to_string()))
            .collect::<std::collections::HashMap<_, _>>()
    };
    assert_eq!(
        super::shed_incoming_after_config(&config(&[("admission_wait", "5m")])),
        Some("5m")
    );
    assert_eq!(
        super::shed_incoming_after_config(&config(&[
            ("admission_wait", "5m"),
            ("shed_incoming_after", "45s"),
        ])),
        Some("45s"),
        "the new key wins when both are set"
    );
    assert_eq!(super::shed_incoming_after_config(&config(&[])), None);
}
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
```diff
@@ -360,7 +360,10 @@ impl HostPlugin for NatsMessaging {
         let interface_subscriptions = interface.config.get("subscriptions").cloned();
         let interface_consumer_group = interface.config.get(CONSUMER_GROUP_CONFIG).cloned();
         let interface_max_in_flight = interface.config.get(super::MAX_IN_FLIGHT_CONFIG).cloned();
-        let interface_admission_wait = interface.config.get(super::ADMISSION_WAIT_CONFIG).cloned();
+        let interface_admission_wait = interface
+            .config
+            .get(super::SHED_INCOMING_AFTER_CONFIG)
+            .cloned();
```
ricochet: Reads the deprecated `admission_wait` too (see the comment on `SHED_INCOMING_AFTER_CONFIG` in `mod.rs`):

```suggestion
        let interface_admission_wait =
            super::shed_incoming_after_config(&interface.config).map(str::to_owned);
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/nats.rs
```diff
@@ -397,7 +400,7 @@ impl HostPlugin for NatsMessaging {
         let local_admission_wait = component_handle
             .local_resources()
             .config
-            .get(super::ADMISSION_WAIT_CONFIG)
+            .get(super::SHED_INCOMING_AFTER_CONFIG)
             .cloned();
```
ricochet: Same here:

```suggestion
        let local_admission_wait =
            super::shed_incoming_after_config(&component_handle.local_resources().config)
                .map(str::to_owned);
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -548,7 +548,7 @@ impl HostPlugin for InMemoryMessaging {
             component_handle
                 .local_resources()
                 .config
-                .get(super::ADMISSION_WAIT_CONFIG)
+                .get(super::SHED_INCOMING_AFTER_CONFIG)
```
ricochet: Same here, reading through `shed_incoming_after_config`. This isn't a suggestion block because rustfmt also changes the `let` line above, which is outside the diff:

```rust
        let admission_wait = super::parse_admission_wait(super::shed_incoming_after_config(
            &component_handle.local_resources().config,
        ));
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -820,46 +842,137 @@ impl ComponentGates {
...
             .unwrap_or_else(std::sync::PoisonError::into_inner);
-        let Some(entry) = gates.get_mut(identity) else {
-            return;
-        };
-        entry.bindings = entry.bindings.saturating_sub(1);
-        if entry.bindings == 0 {
-            entry.semaphore.close();
-            gates.remove(identity);
+        if let Some(entry) = gates.get_mut(identity) {
+            entry.bindings = entry.bindings.saturating_sub(1);
```
ricochet: With "last request wins", a new binding changes the ceiling for replicas that are still running, and nothing puts it back when that binding goes away. The gate doesn't record what each binding asked for, so `release` has nothing to recompute from.

Scenario: two replicas on a host run with `max_in_flight: 32`. A new revision with `4` binds one replica here, and the shared gate shrinks to 4. The rollout is rolled back and that replica unbinds, but the gate stays at 4. The surviving replicas, whose manifest still says 32, stay capped at 4 until they restart. Before this PR, keeping the first binding's ceiling left them at 32.

One fix is to track each binding's request, keyed by bind order, and move the gate to the newest remaining request on unbind. Keying by a binding ID rather than by value matters: with bindings `32, 4, 32`, removing "a 32" by value can take the newest one's entry and hand the gate to the 4.

```diff
 struct GateEntry {
     semaphore: Arc<Semaphore>,
     limit: usize,
-    bindings: usize,
+    /// The ceiling each live binding asked for, keyed by bind order. The
+    /// newest entry is the ceiling the gate converges to.
+    requests: std::collections::BTreeMap<u64, usize>,
+    /// Key for the next binding's entry in `requests`.
+    next_binding: u64,
     pending_shrink: usize,
 }

 impl GateEntry {
+    /// Move the live total toward `target`: a grow applies at once, a shrink
+    /// forgets free permits now and owes the rest to returning ones.
+    fn resize_to(&mut self, target: usize) {
+        match target.cmp(&self.limit) {
+            std::cmp::Ordering::Greater => {
+                self.semaphore.add_permits(target - self.limit);
+                self.limit = target;
+                self.pending_shrink = 0;
+            }
+            std::cmp::Ordering::Less => {
+                self.pending_shrink = self.limit - target;
+                self.reconcile_shrink();
+            }
+            std::cmp::Ordering::Equal => self.pending_shrink = 0,
+        }
+    }
+
     fn can_retire(&self) -> bool {
-        self.bindings == 0 && self.semaphore.available_permits() == self.limit
+        self.requests.is_empty() && self.semaphore.available_permits() == self.limit
     }
 }

-    fn acquire(&self, identity: &AdmissionIdentity, limit: usize) -> (Arc<Semaphore>, usize) {
+    fn acquire(&self, identity: &AdmissionIdentity, limit: usize) -> (Arc<Semaphore>, u64) {
         ...
-            bindings: 0,
+            requests: std::collections::BTreeMap::new(),
+            next_binding: 0,
             pending_shrink: 0,
         });
-        entry.bindings += 1;
+        let binding = entry.next_binding;
+        entry.next_binding += 1;
+        entry.requests.insert(binding, limit);
         ...
-            if limit > entry.limit {
-                ...
-            }
+            entry.resize_to(limit);
         ...
-        (Arc::clone(&entry.semaphore), entry.limit)
+        (Arc::clone(&entry.semaphore), binding)
     }

-    fn release(&self, identity: &AdmissionIdentity) {
+    fn release(&self, identity: &AdmissionIdentity, binding: u64) {
         ...
         if let Some(entry) = gates.get_mut(identity) {
-            entry.bindings = entry.bindings.saturating_sub(1);
+            entry.requests.remove(&binding);
+            if let Some((_, &target)) = entry.requests.last_key_value() {
+                entry.resize_to(target);
+            }
             Self::clean_up_entry(&mut gates, identity);
         }
     }

 // MessagingLimits::admission
-        let (component, limit) = self.gates.acquire(identity, resolved);
+        let (component, binding) = self.gates.acquire(identity, resolved);
         Admission {
             component,
+            binding,
             host: Arc::clone(&self.host),
-            limit,
+            limit: resolved,

 // Admission
+    /// This binding's key in its gate's `requests`, returned by
+    /// [`Admission::close`].
+    binding: u64,

     pub(crate) fn close(self) {
-        self.gates.release(&self.identity);
+        self.gates.release(&self.identity, self.binding);
     }
```

The `bindings()` / `gate_state()` test helpers become `e.requests.len()`. `limit: resolved` is the fix for the `Admission::limit` comment further down. Two tests; the first fails on this head (`left: Some((4, 0, 1))`):

```rust
#[tokio::test]
async fn unbinding_a_later_request_restores_the_remaining_bindings_ceiling() {
    // A new revision lowers the ceiling on one replica, then is rolled
    // back. The replicas still bound asked for 32 and must get it back.
    let limits = MessagingLimits::new(64, 32);
    let identity = unique_identity();
    let old = limits.admission(&identity, Some(32));
    let new = limits.admission(&identity, Some(4));
    assert_eq!(limits.gates.gate_state(&identity), Some((4, 0, 2)));

    new.close();
    assert_eq!(
        limits.gates.gate_state(&identity),
        Some((32, 0, 1)),
        "the gate must return to the ceiling the remaining binding asked for"
    );
    assert_eq!(old.component.available_permits(), 32);
}

#[tokio::test]
async fn an_unbind_removes_its_own_request_not_an_equal_one() {
    // Bound 32, 4, 32 in that order: the newest asks for 32. The oldest
    // leaving must not take the newest's entry and hand the gate to the 4.
    let limits = MessagingLimits::new(64, 32);
    let identity = unique_identity();
    let oldest = limits.admission(&identity, Some(32));
    let _middle = limits.admission(&identity, Some(4));
    let newest = limits.admission(&identity, Some(32));
    assert_eq!(limits.gates.gate_state(&identity), Some((32, 0, 3)));

    oldest.close();
    assert_eq!(limits.gates.gate_state(&identity), Some((32, 0, 2)));

    newest.close();
    assert_eq!(limits.gates.gate_state(&identity), Some((4, 0, 1)));
}
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -1204,14 +1339,20 @@ impl Admission {
     /// the workload responsible for it.
     ///
     /// Cancel-safe: dropping the returned future before it resolves releases
-    /// whichever permit it had already taken.
+    /// whichever permit it had already taken. If dropped while waiting for the
+    /// host permit, the component permit is released through [`AdmissionPermit::drop`],
+    /// ensuring shrink debt accounting and gate retirement cleanup run.
     pub(crate) async fn acquire(&self) -> Option<AdmissionPermit> {
         let component = Arc::clone(&self.component).acquire_owned().await.ok()?;
```
ricochet: A gate can outlive its last binding and never be removed, because one way a permit comes back skips cleanup.

When a permit is released with no shrink owed, tokio hands it straight to a waiting `acquire_owned`, so it isn't counted in `available_permits()`. If that waiting future is then dropped before it's polled again, tokio's `Drop for Acquire` returns the permit itself (`add_permits_locked`, `batch_semaphore.rs` in tokio 1.53). That skips `AdmissionPermit::drop`, so `release_permit` / `clean_up_entry` never run.

Scenario: a saturated component is torn down while its subscriber loop waits for a slot. `close()` drops the binding count to zero. The last handler then finishes: its permit goes to the parked loop, and `can_retire` sees `available == limit - 1`, so the entry stays. Then the cancel token or `abort()` drops the loop's future, and the permit returns silently. The entry stays in the map with no bindings and an open semaphore until a workload with the same name binds again.

`dropping_acquire_while_waiting_for_host_permit_cleans_up_component_gate` covers the drop while waiting for the *host* permit, where an `AdmissionPermit` already exists. This is the drop while waiting for the *component* permit. A guard that re-runs cleanup if this await is cancelled:

```suggestion
        // Declared before the acquire future, so it drops after it. A permit
        // tokio already handed this waiter goes straight back to the semaphore
        // when the future is dropped, bypassing `AdmissionPermit::drop`; this
        // re-runs cleanup so a gate whose last binding closed still retires.
        let retire = CleanUpOnCancel {
            gates: &self.gates,
            identity: &self.identity,
        };
        let component = Arc::clone(&self.component).acquire_owned().await;
        std::mem::forget(retire);
        let component = component.ok()?;
```

`CleanUpOnCancel` is in the suggestion on `impl Drop for AdmissionPermit` below. A test that fails on this head and passes with both suggestions:

```rust
#[tokio::test]
async fn a_cancelled_waiter_handed_the_last_permit_still_retires_the_gate() {
    let limits = MessagingLimits::new(64, 1);
    let identity = unique_identity();
    let binding = limits.admission(&identity, Some(1));
    let handler = binding.acquire().await.expect("first admits");

    // The subscriber loop, parked on the saturated gate.
    let loop_admission = binding.clone();
    let mut parked = Box::pin(loop_admission.acquire());
    assert!(futures::poll!(parked.as_mut()).is_pending());

    // Teardown with the handler still running. When it returns, tokio
    // hands its permit to the parked waiter instead of the semaphore.
    binding.close();
    drop(handler);
    assert!(limits.gates.contains(&identity));

    // The loop is cancelled before it is polled again.
    drop(parked);
    assert!(
        !limits.gates.contains(&identity),
        "a permit returned by a cancelled waiter must still retire the gate"
    );
}
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -1348,10 +1488,28 @@ impl Admission {
...
 pub(crate) struct AdmissionPermit {
-    _component: OwnedSemaphorePermit,
-    _host: OwnedSemaphorePermit,
+    _component: Option<OwnedSemaphorePermit>,
+    _host: Option<OwnedSemaphorePermit>,
+    gates: ComponentGates,
+    identity: AdmissionIdentity,
+}
+
+impl Drop for AdmissionPermit {
```
ricochet: The guard for the cancelled component acquire above. `release_permit(identity, None)` already takes the lock and runs `clean_up_entry`:

```suggestion
/// Re-runs gate cleanup if a component acquire is dropped before it resolves.
/// Forgotten once the acquire completes.
struct CleanUpOnCancel<'a> {
    gates: &'a ComponentGates,
    identity: &'a AdmissionIdentity,
}

impl Drop for CleanUpOnCancel<'_> {
    fn drop(&mut self) {
        self.gates.release_permit(self.identity, None);
    }
}

impl Drop for AdmissionPermit {
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -1172,6 +1303,10 @@ pub(crate) enum Admitted {
 pub(crate) struct Admission {
     component: Arc<Semaphore>,
     host: Arc<Semaphore>,
+    /// Ceiling resolved at bind time. The shared gate may converge to a later
+    /// request afterwards (see [`ComponentGates`]), so the shed log reports
+    /// the live total and this stays as the bind-time value for
+    /// [`Admission::limit`].
```
ricochet: This value comes from `ComponentGates::acquire`'s second return, which is the ceiling *in force* when this binding joined, not the one it asked for. Now that a shrink is applied gradually, those differ. If a rebind asks for 5 while all 10 slots are held, `limit()` returns 10, so the bind logs in `in_memory.rs` and `nats.rs` print `max_in_flight=10`, and the operator sees the change as not applied.

Suggest storing the request (`limit: resolved` in `MessagingLimits::admission`, ignoring or dropping the tuple's second element) and describing it that way:

```suggestion
    /// The ceiling this binding asked for, after defaulting and clamping. The
    /// shared gate converges to the newest binding's request (see
    /// [`ComponentGates`]), so the shed log reports the live total instead.
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/mod.rs
```diff
@@ -1258,21 +1402,21 @@ impl Admission {
         }
     }
 
-    /// The resolved per-component ceiling, after defaulting and clamping.
+    /// The ceiling resolved at bind time, after defaulting and clamping.
+    ///
+    /// Stays fixed while the shared gate converges to later requests; the shed
+    /// log reports the live total instead.
```
ricochet: Matching doc for `limit()` once it returns the request:

```suggestion
    /// The ceiling this binding asked for, after defaulting and clamping.
    ///
    /// The shared gate converges to it unless a newer binding asks otherwise;
    /// the shed log reports the live total instead.
```

PR #5568, opened 2026-09-14
fix(wash): retry transient WIT registry fetches

### LINKED ISSUE #5567: Retry wash wit fetch
Small feature request: retry on transient registry transport errors.

wash should retry only clearly transient cases such as timeouts, connection resets, and HTTP 429/5xx responses, with bounded exponential backoff and jitter. Authentication, authorization, invalid references, and missing packages should fail immediately.

addu2612:
Retries WIT dependency fetches when the registry failure is clearly transient: connection and timeout failures, HTTP 429 or 5xx responses, and transient I/O errors. Authentication, authorization, missing packages, invalid references, and other permanent errors still fail immediately.

The retry loop makes at most three attempts with exponential backoff and jitter. It reuses the fetcher's package cache between attempts and preserves the existing lock-file restoration and missing-package diagnostics after the final failure.

Fixes #5567

## Verification

- `cargo +1.98.0 check -j 1 -p wash --lib`
- `cargo +1.98.0 test -j 1 -p wash --lib wit_fetch_` (4 passed)
- `cargo +1.98.0 fmt --all -- --check`
- `git diff --check`

The repository-pinned 1.96 toolchain installation on this machine was missing its Cargo binary, so the checks used Rust 1.98. A full Clippy run reached an unrelated existing `for_kv_map` warning in `crates/wash-runtime/src/host/mod.rs:396` under Rust 1.98.


crates/wash/src/cli/wit.rs
```diff
@@ -719,6 +744,68 @@ async fn fetch_with(
...
+        }
+
+        if let Some(error) = cause.downcast_ref::<reqwest::Error>() {
+            return error.is_timeout()
+                || error.is_connect()
+                || error
+                    .status()
+                    .is_some_and(|status| status.as_u16() == 429 || status.is_server_error());
+        }
+
```
ricochet, line 791: This branch can't match anything from a registry fetch. `wash` depends on `reqwest` 0.12, but `oci-client` and `wasm-pkg-client` are built on `reqwest` 0.13.4 (`cargo tree -p oci-client -e normal --depth 1`), and a downcast only matches the exact type, so the 0.13 error in the chain never downcasts to 0.12's `reqwest::Error`. Timeouts and connect failures are already covered by the `OciDistributionError::RequestError` arm above, so this can go.

```suggestion
```

crates/wash/src/cli/wit.rs
```diff
@@ -719,6 +744,68 @@ async fn fetch_with(
...
+const WIT_FETCH_INITIAL_BACKOFF: Duration = Duration::from_millis(100);
+
+/// Return a bounded exponential delay with up to one base delay of positive jitter.
+fn wit_fetch_retry_delay(failed_attempt: u32) -> Duration {
+    let exponent = failed_attempt.saturating_sub(1).min(3);
+    let base = WIT_FETCH_INITIAL_BACKOFF.saturating_mul(1 << exponent);
+    base + Duration::from_millis(fastrand::u64(0..base.as_millis() as u64))
+}
+
+/// Whether an error chain proves that a registry fetch failed for a transient reason.
```
ricochet, line 757: The `OciDistributionError` downcast has the same catch: it only matches while `wash` and `wasm-pkg-client` resolve to the same `oci-client`. They do today (both 0.17), but if `wasm-pkg-client` moves to 0.18 ahead of the workspace pin, every downcast misses, retries silently stop, and the unit tests still pass because they build their errors from wash's copy of `oci-client`.

`wasm-pkg-client` re-exports `oci_client::client`, and `Client`'s `TryFrom<ClientConfig>` error is its `OciDistributionError`, so we can make a mismatch a compile error instead:

```suggestion
// A downcast only matches the `oci-client` types `wasm-pkg-client` was built against, so this
// stops compiling if the two ever resolve to different `oci-client` versions.
const _: () = {
    use wasm_pkg_client::oci::client::{Client, ClientConfig};
    let _: fn(<Client as TryFrom<ClientConfig>>::Error) -> OciDistributionError = |error| error;
};

/// Whether an error chain proves that a registry fetch failed for a transient reason.
```

crates/wash/src/cli/wit.rs
addu2612: Agreed. Removed the direct reqwest error downcast in 65418f649. Timeout, connection, and status failures remain covered by the OciDistributionError RequestError arm.

crates/wash/src/cli/wit.rs
addu2612: Added the suggested compile-time type identity assertion in 65418f649, so an oci-client version split will now fail at compile time instead of silently disabling retries.

ricochet (approved):

PR #5571, opened 2026-09-14
fix(runtime-operator): correct plugins file-backed vs CLI arg doc

### LINKED ISSUE #5551: chart values.yaml overclaims that every plugins entry is file-backed
charts/runtime-operator/values.yaml:661-662 (v2.9.0) says `plugins` entries are rendered into the host config file rather than CLI arguments. Per templates/_helpers.tpl:221-227, an entry with only `id` plus `image`/`file` and none of the config/allowlist/binding fields still renders as a `--host-plugin` CLI argument; only native entries and entries carrying those fields are file-backed. The comment should state the actual split.

shreya-d2604:
Fixes #5551 

## Summary
- The plugins doc comment in charts/runtime-operator/values.yaml claimed every entry is always file-backed, but runtime-operator.hostPluginPartition in _helpers.tpl:221-227 actually splits entries: a component entry with only id plus image/file and none of the config/allowlist/binding fields renders as a plain --host-plugin=... CLI arg; only native entries and entries carrying those fields are file-backed.
- Comment-only change, no template/logic changes.


shreya-d2604:
Ready for review ! @ericgregory

PR #5572, opened 2026-09-14
(feat): stream KV entries with values via bucket.select

jfleitz:
Adds `bucket.select` to `wasmcloud:nats/kv@0.1.0`. It streams every live entry
matching a filter and includes each value, so a caller doesn't need a `get()`
per key after `keys()`, eliminating the additional network call. It also isn't capped at one page.

- Returns `stream<entry>` plus a `future<result<_, nats-error>>` that carries
  the final status, the same way `wasi:http@0.3.0` pairs a body with its
  trailers. A drain that stops because the consumer or connection went away
  can't be mistaken for a complete one.
- Uses a no-ack pull consumer with `last-per-subject` delivery. The guest's
  read rate controls the flow, so nothing is dropped and host memory stays
  bounded per poll.
- `select-options`: `include-values` (false gives a keys-only listing with no
  cap), `include-tombstones`, and `max-entries` (0 = no limit), `timeout-ms` (whole-drain deadline; 0 = the binding's `request-timeout-ms`, else none).

## Testing

Tested with a local kind cluster and creating a test workload that uses the `buckets.select` functionality.
A test was done with seeding 2500 keys with deleting one key and overwriting another.
Calling select returned 2499 values, with the one key returning the new overwritten value.


crates/wash-runtime/src/plugin/wasmcloud_nats/interfaces/jetstream/kv.rs
```diff
@@ -132,6 +162,209 @@ impl<T: 'static + Send> labeled_kv::HostWithStore<T> for SharedCtx {
...
+    /// whenever a tombstone is skipped.
+    consumed: u64,
+    /// Resolves once, carrying the terminal status. A stream that ends
+    /// because the consumer went away must not read as a completed drain, so
+    /// this is the only thing that says which happened.
+    result: Option<oneshot::Sender<Result<(), types::NatsError>>>,
+    finished: bool,
+    /// Wall-clock bound on the whole drain, and the limit it was built from.
+    /// `None` is unbounded.
+    deadline: Option<(Pin<Box<tokio::time::Sleep>>, Duration)>,
```
ricochet, line 197: Needed for the consumer cleanup suggested at L644 below.

```suggestion
    /// `None` is unbounded.
    deadline: Option<(Pin<Box<tokio::time::Sleep>>, Duration)>,
    /// Deletes the consumer when the producer is dropped, however the drain ended.
    _consumer: super::ScanConsumerGuard,
```

crates/wash-runtime/src/plugin/wasmcloud_nats/interfaces/jetstream/kv.rs
```diff
@@ -132,6 +162,209 @@ impl<T: 'static + Send> labeled_kv::HostWithStore<T> for SharedCtx {
...
+                    break;
+                }
+            };
+            let message = match next {
+                Some(Ok(m)) => m,
+                Some(Err(e)) => {
+                    let timed_out = chain_timed_out(&e);
+                    failure = Some(kv_err("kv select iter failed", timed_out, e));
+                    break;
+                }
```
ricochet, line 321: A guest that waits more than 30s between reads gets a failed drain, even though the consumer is healthy.

In async-nats 0.49.1, `Stream::poll_next` (`consumer/pull.rs`, around L1112) creates a `2 × idle_heartbeat` timer (30s with `messages()` defaults) if none is running. It checks that timer *before* it looks at buffered messages, and it clears the timer only when the subscriber hands it something.

A miss here only means the host wasn't polling, so skip it. The next poll starts a fresh timer and yields the buffered entries. A consumer that is really gone still shows up: the server reaps it after `inactive_threshold`, which surfaces as `ConsumerDeleted`/`NoResponders`, and `timeout-ms` still applies. For a hard stop, you could fail on a second miss in a row with no message in between, the way `pull::Ordered` does (around L810).

```suggestion
                Some(Err(e))
                    if matches!(
                        e.kind(),
                        jetstream::consumer::pull::MessagesErrorKind::MissingHeartbeat
                    ) =>
                {
                    continue;
                }
                Some(Err(e)) => {
                    let timed_out = chain_timed_out(&e);
                    failure = Some(kv_err("kv select iter failed", timed_out, e));
                    break;
                }
```

crates/wash-runtime/src/plugin/wasmcloud_nats/interfaces/jetstream/kv.rs
```diff
@@ -132,6 +162,209 @@ impl<T: 'static + Send> labeled_kv::HostWithStore<T> for SharedCtx {
...
+                None => {
+                    ended = true;
+                    break;
+                }
```
ricochet, line 334: A pull `Stream` never ends on its own. It returns `None` only when its subscription or request channel closes, for example when `drain_one` in `conn.rs` calls `client.drain()` on teardown. This arm then calls `finish(Ok(()))` whatever `consumed` is, so a drain stopped at 31,000 of 40,000 entries resolves the status future with `Ok`. That's the one outcome the `result` doc comment and the WIT say must not happen.

`expected` is already tracked, so a `None` that arrives before it is reached can be reported as an error:

```suggestion
                None if self.consumed < self.expected => {
                    failure = Some(types::NatsError::Connection(format!(
                        "kv select stream closed after {} of {} entries",
                        self.consumed, self.expected
                    )));
                    break;
                }
                None => {
                    ended = true;
                    break;
                }
```

crates/wash-runtime/src/plugin/wasmcloud_nats/interfaces/jetstream/kv.rs
```diff
@@ -350,6 +583,112 @@ impl<T: 'static + Send> kv::HostBucketWithStore<T> for SharedCtx {
...
+                // abandoned consumer cannot count against `max_consumers`
+                // forever. Well above any healthy drain.
+                inactive_threshold: Duration::from_secs(60),
+                ..Default::default()
+            })
+            .await
+        {
+            Ok(c) => c,
+            Err(e) => return Ok(Err(jetstream_err("kv select failed", e))),
+        };
```
ricochet, line 669: **The consumer each `select` creates is never deleted.**

A handler calling `select` at 100 req/s keeps about 6,000 consumers on the KV stream. Once the account or stream `max_consumers` limit is reached, every `create_consumer` fails, so `select` breaks for every workload using that bucket.

`scan` handles the same problem with `ScanConsumerGuard` (`jetstream/mod.rs`). Arming it right after creation covers the early return below, and moving it into the producer (with the suggestions on L196 and L675) deletes the consumer when the stream is dropped. It could be renamed to something like `ConsumerGuard`, since its drop warning says "scan consumer".

```suggestion
            Ok(c) => c,
            Err(e) => return Ok(Err(jetstream_err("kv select failed", e))),
        };
        let cleanup = super::ScanConsumerGuard::arm(&store.stream, &consumer.cached_info().name);
```

crates/wash-runtime/src/plugin/wasmcloud_nats/interfaces/jetstream/kv.rs
```diff
@@ -350,6 +583,112 @@ impl<T: 'static + Send> kv::HostBucketWithStore<T> for SharedCtx {
...
+            prefix: store.prefix.clone(),
+            include_tombstones: opts.include_tombstones,
+            include_values: opts.include_values,
+            max_entries: opts.max_entries,
+            emitted: 0,
+            result: Some(result_tx),
+            finished: false,
+            deadline: select_timeout(opts.timeout_ms, conn.request_timeout)
+                .map(|limit| (Box::pin(tokio::time::sleep(limit)), limit)),
+        };
```
ricochet: Part of the consumer cleanup suggested at L644 above.

```suggestion
            deadline: select_timeout(opts.timeout_ms, conn.request_timeout)
                .map(|limit| (Box::pin(tokio::time::sleep(limit)), limit)),
            _consumer: cleanup,
        };
```

crates/wash-runtime/src/plugin/wasmcloud_nats/interfaces/jetstream/kv.rs
```diff
@@ -350,6 +583,112 @@ impl<T: 'static + Send> kv::HostBucketWithStore<T> for SharedCtx {
...
+
+        debug!(%filter, include_values = opts.include_values, "kv select started");
+        accessor.with(|mut store| {
+            let stream = StreamReader::new(&mut store, producer)?;
+            let future = FutureReader::new(&mut store, async move {
+                wasmtime::error::Ok(result_rx.await.unwrap_or_else(|_| {
+                    Err(types::NatsError::Unexpected(
+                        "kv select ended without reporting a terminal status".to_string(),
+                    ))
+                }))
```
ricochet, line 726: `timeout-ms` only takes effect while the guest keeps reading.

The deadline is checked only in `poll_produce`, which runs when the guest reads from `entries`. If a guest reads one page, keeps `entries` in scope, and awaits the status future, that future never resolves, even with `timeout-ms` set. The WIT says hitting the timeout resolves the future with `timeout`.

One option is to compute the deadline as an `Instant` once and apply it on both sides. The producer keeps its `Sleep`, and the future races `result_rx` against the same instant:

```rust
FutureReader::new(&mut store, async move {
    let status = match deadline_at {
        Some((at, limit)) => match tokio::time::timeout_at(at, result_rx).await {
            Ok(status) => status,
            Err(_) => {
                return wasmtime::error::Ok(Err(types::NatsError::Timeout(format!(
                    "kv select did not complete within {}ms",
                    limit.as_millis()
                ))));
            }
        },
        None => result_rx.await,
    };
    // ...existing unwrap_or_else
})
```

ricochet (approved):

PR #5574, opened 2026-09-15
docs(wit): distinguish wire ops from local accessors in nats header

### LINKED ISSUE #5552: wit/nats: world.wit header says everything is async, but message-handle accessors are sync
wit/nats/wit/world.wit:5-6 (v2.9.0) states everything in the package is `async func`. The `message-handle` accessors `message`, `sequence`, and `delivery-count` (world.wit:283-287) are plain sync `func`. The header should say that operations touching the wire are async and local accessors are sync.

Yash121l:
Fixes #5552

The header comment on the `wasmcloud:nats` `types` interface says every operation in the package is an `async func`. The three accessors on `message-handle` (`message`, `sequence`, `delivery-count`) are plain sync `func`, because they read state already held on the handle with no NATS round trip. Everything else in the file, publishes, requests, JetStream fetches and the settle operations, is async.

This rewords the header to scope the async claim to operations that cross the wire and adds one note on the `message-handle` resource naming the three sync accessors. No signatures change.

Verified with `cargo check -p wash-runtime`, which embeds this file through `wasmtime::component::bindgen!`, so a parse error would fail the build. It builds clean.


vados-cosmonic (dismissed):
LGTM 🚀 

Thanks @Yash121l !

vados-cosmonic (approved):

ricochet (approved):

PR #5576, opened 2026-09-15
fix(wash-runtime): use uuidv7 instead of uuidv4

### LINKED ISSUE #5566: Use UUIDv7 instead of UUIDv4 for generated ids
Split out from https://github.com/wasmCloud/wasmCloud/issues/5461 (§5).

shreya-d2604:
Fixes #5566 

## Summary
- Swaps `uuid::Uuid::new_v4()` for `uuid::Uuid::now_v7()` at the sites
that generate long-lived ids (workload, component context, host, and
linked-call store ids), plus the NATS ephemeral inbox subject. v7 ids
sort by creation time, so these ids become correlatable in logs/traces
instead of random noise — this was flagged in #5461 (§5) and split out
into its own issue since the id generation lives outside the messaging
plugin.
- Left as v4: test-only random values (env var suffixes, a test bucket
name, a test assertion uuid) and the host friendly-name fallback —
none of these are ever sorted or correlated by time, so switching them
adds no value.


crates/wash-runtime/src/engine/workload.rs
```diff
@@ -595,7 +595,7 @@ impl DerefMut for WorkloadService {
 /// state of a workload before execution.
 #[derive(Clone)]
 pub struct ResolvedWorkload {
-    /// The unique identifier of the workload, created with [uuid::Uuid::new_v4]
+    /// The unique identifier of the workload, created with [uuid::Uuid::now_v7]
```
ricochet: This id isn't minted in the runtime: it's the `workload_id` from `WorkloadStartRequest` (`Host::workload_start` → `Engine::initialize_workload`). Under the operator that's the Workload's Kubernetes UID, which is always v4, so naming a uuid version here is misleading.

```suggestion
    /// The unique identifier of the workload, supplied by whoever starts it
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -2402,7 +2402,7 @@ impl ResolvedWorkload {
 /// - Validate that all dependencies can be satisfied
 /// - Create the final executable workload representation
 pub struct UnresolvedWorkload {
-    /// The unique identifier of the workload, created with [uuid::Uuid::new_v4]
+    /// The unique identifier of the workload, created with [uuid::Uuid::now_v7]
```
ricochet: Same as above: this is the caller-supplied workload id.

```suggestion
    /// The unique identifier of the workload, supplied by whoever starts it
```

crates/wash-runtime/src/observability/meters.rs
```diff
@@ -88,7 +88,7 @@ pub struct Meters {
 /// scheme.
 ///
 /// Deliberately **not** the workload or component id: both are
-/// `uuid::Uuid::new_v4()` minted per workload construction, so attributing a
+/// `uuid::Uuid::now_v7()` minted per workload construction, so attributing a
 /// series with one mints a fresh series on every restart, rolling update and
 /// replica — growth driven by deployment churn, and a value no operator can map
 /// back to a workload. The ids keep their place on the span and the log line,
```
ricochet: Only the component id comes from `now_v7()`. The workload id is supplied by the caller (the Kubernetes UID under the operator), so "both are minted per workload construction" isn't right.

```suggestion
/// Deliberately **not** the workload or component id: the workload id is
/// unique per started workload and the component id is minted per workload
/// construction, so attributing a series with one mints a fresh series on
/// every restart, rolling update and replica — growth driven by deployment
/// churn, and a value no operator can map back to a workload. The ids keep
/// their place on the span and the log line,
```

crates/wash-runtime/src/plugin/wasmcloud_nats/subscriber.rs
```diff
@@ -635,7 +635,7 @@ fn sanitize_name_component(name: &str) -> String {
 /// The durable/deliver-plane scope for a workload's JetStream subscriptions.
 ///
 /// **Namespace only.** Every replica of a deployment is a separate workload
-/// with its own `uuid::Uuid::new_v4()` id *and* its own generated name —
+/// with its own `uuid::Uuid::now_v7()` id *and* its own generated name —
```
ricochet: This is the workload id, which comes from the caller (the Kubernetes UID under the operator), not `now_v7()`.

```suggestion
/// with its own workload id *and* its own generated name —
```

crates/wash-runtime/src/plugin/wasmcloud_nats/subscriber.rs
```diff
@@ -853,7 +853,7 @@ pub(super) async fn spawn_jetstream_subscriptions(
                     }
                     None => (
                         None,
-                        format!("_nats_push.{}", uuid::Uuid::new_v4()),
+                        format!("_nats_push.{}", uuid::Uuid::now_v7()),
```
ricochet: I'd keep v4 here. This ephemeral deliver subject is never correlated with anything, and `now_v7()` leaves 32 fully random bits plus a per-millisecond counter instead of v4's 122 random bits, so the subject gets easier to guess for no benefit.

```suggestion
                        format!("_nats_push.{}", uuid::Uuid::new_v4()),
```

crates/wash-runtime/src/engine/ctx.rs
```diff
@@ -572,8 +572,8 @@ pub struct CtxBuilder {
 impl CtxBuilder {
     pub fn new(workload_id: impl Into<Arc<str>>, component_id: impl Into<Arc<str>>) -> Self {
         Self {
-            id: uuid::Uuid::new_v4().to_string().into(),
-            store_id: uuid::Uuid::new_v4().to_string().into(),
+            id: uuid::Uuid::now_v7().to_string().into(),
+            store_id: uuid::Uuid::now_v7().to_string().into(),
```
ricochet, line 576: Could `ctx_builder_generates_uuid_id` (further down in this file) assert the version, so this can't quietly go back to v4? Its comment still says "valid UUID v4 string" too.

```rust
        let id = uuid::Uuid::parse_str(&ctx.id).expect("id is a UUID");
        let store_id = uuid::Uuid::parse_str(&ctx.store_id).expect("store_id is a UUID");
        assert_eq!(id.get_version_num(), 7);
        assert_eq!(store_id.get_version_num(), 7);
```

shreya-d2604:
@ricochet Agreed on all points, fix incoming.

shreya-d2604:
Pushed a fix addressing all the feedback, ready for another look @ricochet

ricochet (approved):

PR #5581, opened 2026-09-16
fix(wash): wash runtime nats wit version

vados-cosmonic:


vados-cosmonic:
E2E tests for wash are now back to  :green_circle: , apologies for the interruption -- once the last two `wash` checks pass I'm going to force this in real quick so all people need to do is rebase.

PR #5583, opened 2026-09-16
ci: harden flaky main checks

vados-cosmonic:
Fixes three independent sources of main CI flakes observed in recent Actions runs:

- pairs the thundering-herd Host CR with the active Ready host pod instead of selecting a stale Ready CR left by a rollout
- retries the NATS round-trip probe while a newly Ready workload finishes registering its subscription
- retries transient external downloads during operator/gateway image builds and chart-testing setup

The actual e2e assertions and chart lint remain blocking and are not retried.

Validation:
- actionlint v1.7.12 on the edited workflows
- gofmt on the edited Go tests
- git diff --check

A local compile-only Go test was attempted but exhausted the 1.4 GiB available workspace storage while compiling dependencies; PR CI provides the full compile and Docker/Kind coverage.

Signed-off-by: Victor Adossi <vadossi@cosmonic.com>


vados-cosmonic:
Would love some feedback here @ricochet and @jfleitz now that main has the critical fixes merged... Honestly not a huge fan of repeating steps that are known flaky and using bash to retry stuff, but these were the failures I saw in between now and the NATS wit change that were unrelated/flakes, so in theory this will make our tests less flaky.

jfleitz:
> Would love some feedback here @ricochet and @jfleitz now that main has the critical fixes merged... Honestly not a huge fan of repeating steps that are known flaky and using bash to retry stuff, but these were the failures I saw in between now and the NATS wit change that were unrelated/flakes, so in theory this will make our tests less flaky.

Yeah I agree. The changes look good to me, but in the same boat with the retry loops, and think it will just get worse as we add additional features and test cases.

jfleitz (approved):

ricochet (approved):

ricochet:
For the host heartbeats, we should be sending a clean termination at the end of each test so having to get alive pods may not be necessary if we ensure that cleanup happens, but I think its a reasonable protection to add too.

PR #5585, opened 2026-09-16
feat(wash-runtime): upgrade to Wasmtime 48

ricochet:
Updates wasmtime and friends from 47.0.4 to 48.0.2 (workspace MSRV 1.95) and adapts the host to what 48 changed. Most of the diff is the `wasi:http` p2/p3 merge and the `wasi:sockets` rewrite; the rest closes gaps the bump opened.

## Breaking for embedders

`wasi-http` 48 replaced the separate p2/p3 host traits with one `WasiHttpHooks::send_request` over `WasiBody` / `RequestOptions` / `Error`. Our public traits follow it:

- `OutgoingHandler::{send_request, send_request_p3}` → `send_request`
- `HostHandler::{outgoing_request, outgoing_request_p3}` → `outgoing_request`
- `Router::{allow_outgoing_request, allow_outgoing_request_p3}` → `allow_outgoing_request`
- `host::http_p3::{P3Body, P3SendFuture, P3RequestErrorFuture, P3SendResult}` are gone; use `WasiBody` and `host::http::{SendFuture, SendResult, RequestIoFuture}`

## Behavior changes worth attention

- **A denied outgoing request no longer traps a `wasi:http` 0.2 guest.** 48's hook signature cannot trap, so the guest receives `http-request-denied` as a handleable error, same as the original semantics for 0.3. The denial also drains the request body rather than dropping it: a guest that uploads while awaiting the response would otherwise report a closed stream instead of the denial. Same change for "http client not available".
- **Hop-by-hop headers are withheld from 0.3 guests**, as they always have been from 0.2 ones: `host`, `connection`, `keep-alive`, `transfer-encoding`, `upgrade`, `proxy-*`, `http2-settings`. wasmtime 48 strips them for both versions; we keep its default list rather than overriding it.
- **Scheme and authority are normalized once at the ingress**, so both ABI versions read them off the URI. The scheme is the listener's own (`https` only when TLS terminates here) and never `X-Forwarded-Proto`, which the client controls. This is the same class of bug #5557 fixes for `X-Real-IP`. A `Host` that is not a valid authority is now answered with **400**: the hostname router splits on the last colon, so `example.com:8080:9090` would otherwise route as `example.com:8080`. This also fixes 0.2 guests seeing `http` behind a TLS listener.
- **The inbound request body keeps a between-frame timeout.** 48 dropped the one wasmtime-wasi applied itself, so a client that sends headers and stalls would pin the store serving it. Restored via `WASH_HTTP_REQUEST_BODY_TIMEOUT_SECS` (default 600s, as before).
- **Sockets: policy now applies to inbound traffic too**: `listen`, `accept` (real and virtual) and received datagrams and a real accepted connection takes an inbound quota slot. Refused connections are reset rather than closed in order, matching upstream, and every refusal loop is bounded so a flood from a refused peer cannot spin a host call or a poll.
- **`wash dev` sockets draw on the configured connection quotas.** The descriptor limit is raised and the registry built before the engine, so the socket policy carries the operator's allowance instead of a private default computed from the unraised limit.

## Notable in the port

- Ported upstream's fix for a newly created UDP socket reporting itself unwritable (bytecodealliance/wasmtime#12612).
- A guest body error is moved through the upload oneshot instead of being copied, which removed a hand-written clone of the `#[non_exhaustive]` upstream error enum whose catch-all arm flattened `Hyper`/`Connect`/`Tls` to `HttpProtocolError`.
- Dropped `cap-std`/`cap-net-ext`, as upstream did; the UDP socket helper is now rustix, keeping the Windows `WSAStartup` behavior `cap-net-ext` used to provide.
- `FsPerms::{ReadOnly, ReadWrite}` replaces `DirPerms`/`FilePerms`; handled wasmtime 48's new `FixedLengthList` value and type variants, including relocation across the cross-store bridge.

## Release blocker

The workspace `[patch.crates-io]` pins `wasi-webgpu-wasmtime` and `wasi-gfx-runtime-shared` to a wasi-gfx-runtime git revision, because the released 0.2.0 still requires wasmtime 47 and we hand it our own `Linker`. `wasi-webgpu` is in wash-runtime's default features, and `cargo publish` will not resolve through a git `[patch]`, so **wash-runtime cannot be released until wasi-gfx cuts a wasmtime-48 release**.


crates/wash-runtime/src/host/quota.rs
```diff
@@ -425,6 +425,20 @@ pub struct QuotaRegistry {
...
 }
 
+/// The registry a policy built without one draws on.
+///
+/// One per process, not one per policy: a registry of its own would give each
+/// policy a private "host-wide" ceiling, which is the opposite of a ceiling —
+/// two of them would let a host hold the limit twice over. Built on first use,
+/// so a host that raises its descriptor limit during startup is measured after
+/// the raise rather than before it.
+pub fn default_registry() -> Arc<QuotaRegistry> {
```
vados-cosmonic: NIT: I'm not sure we need an arc here rather than a static reference? We should be able to use a `OnceLock` for this and avoid cloning this Arc?

If this is used in places that explicitly expect an `Arc<QuotaRegistry>` then I guess this might not work though.

crates/wash-runtime/src/sockets/loopback/tcp.rs
```diff
@@ -255,6 +255,21 @@ impl TcpSocket {
         }
     }
 
+    pub fn cancel_connect(&mut self, loopback: &mut Network) {
+        let TcpState::Connecting { local_address, .. } =
+            mem::replace(&mut self.state, TcpState::Closed)
+        else {
+            return;
+        };
+        let Some(port) = NonZeroU16::new(local_address.port()) else {
```
vados-cosmonic, line 265: This is a little odd... is this basically a zero check?

crates/wash-runtime/src/plugin/component_host/mod.rs
```diff
@@ -268,7 +268,12 @@ impl ComponentHostPluginState {
 struct PluginStoreContext {
     allowed_hosts: Arc<[crate::host::allowed_hosts::AllowedHost]>,
     allowed_ip_name_lookups: Arc<[crate::host::allowed_ip_name::AllowedIpName]>,
-    http_handler: Option<std::sync::Weak<dyn crate::host::http::HostHandler>>,
+    /// The host this plugin's stores reach back to: its handler serves their
+    /// outgoing HTTP, and its lifetime is what ends them if it is torn down
+    /// while one still runs. `None` reports "http client not available" per
+    /// call, as a plugin importing `wasi:http/outgoing-handler` with no
+    /// handler configured always has.
+    http_handler: Option<crate::host::HostRef>,
```
vados-cosmonic: we should probably rename this?

crates/wash-runtime/src/host/mod.rs
```diff
@@ -338,16 +338,79 @@ pub struct Host {
...
+pub struct HostLifetime(());
+
+/// What a workload, a store or a bound plugin holds to reach back to the host
+/// that built it: its HTTP handler, and the host's own lifetime.
+///
+/// Weak throughout, and for the handler deliberately so — see
+/// [`crate::host::http::live_handler`].
+#[derive(Clone)]
+pub struct HostRef {
+    handler: Option<std::sync::Weak<dyn crate::host::http::HostHandler>>,
```
vados-cosmonic, line 359: is this really only for http handlers?  Should it be?

Probably doesn't need to get rethought right now, but IMO this might be worth renaming as `HttpHostRef`/`HostRefHttp` or something if it's HTTP only

crates/wash-runtime/src/host/mod.rs
```diff
@@ -338,16 +338,79 @@ pub struct Host {
...
+    /// This host's own lifetime, handed out weakly; see [`HostLifetime`].
+    lifetime: Arc<HostLifetime>,
     config: HostConfig,
     meters: Meters,
 }
 
+/// A token whose lifetime is a [`Host`]'s.
+///
+/// The builder transfers its sole strong reference to the host. Unlike the
+/// HTTP handler, an embedder cannot keep this token after the host is dropped.
```
vados-cosmonic: A note on how this is achieved would be helpful -- what stops an embedder from keeping this token after the host is dropped?

crates/wash-runtime/src/sockets/p2_udp.rs
```diff
@@ -75,6 +77,14 @@ impl LoopbackIncomingDatagramStream {
             };
             match self.remote_address {
                 Some(connected_addr) if connected_addr != dgram.source_address => continue,
+                _ if self.socket_addr_check.as_ref().is_some_and(|check| {
+                    check
+                        .check(dgram.source_address, super::SocketAddrUse::UdpReceive)
+                        .is_err()
+                }) =>
+                {
+                    continue;
```
vados-cosmonic, line 87: Should we output a trace message here? I assume in the future we might want some sort of metrics here but something in trace might be useful for now though very noisy

crates/wash-runtime/src/sockets/util.rs
```diff
@@ -368,14 +367,41 @@ pub fn tcp_bind(
...
-pub fn udp_socket(family: AddressFamily) -> std::io::Result<cap_std::net::UdpSocket> {
-    // Delegate socket creation to cap_net_ext. They handle a couple of things for us:
-    // - On Windows: call WSAStartup if not done before.
-    // - Set the NONBLOCK and CLOEXEC flags. Either immediately during socket creation,
-    //   or afterwards using ioctl or fcntl. Exact method depends on the platform.
+pub fn udp_socket(family: SocketAddressFamily) -> std::io::Result<tokio::net::UdpSocket> {
+    #[cfg(windows)]
+    static INIT: std::sync::Once = std::sync::Once::new();
+    #[cfg(windows)]
+    INIT.call_once(|| {
```
vados-cosmonic, line 375: Let's leave the comment in about why this is necessary -- I assume that this will trigger WSAStartup on the windows side by virtue of doing the connect?

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -362,15 +313,15 @@ fn is_resolver_error(err: &std::io::Error) -> bool {
 pub(crate) async fn connect_tcp(
     authority: &str,
     connect_timeout: Duration,
-) -> Result<TcpStream, ConnectError> {
+) -> Result<TcpStream, HttpError> {
```
vados-cosmonic, line 315: This is a bit awkward -- `connect_tcp` seems to return a `HttpError`

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but nits here

crates/wash-runtime/src/sockets/loopback/tcp.rs
ricochet: Yes.  I added a comment. It checks for port zero and converts a nonzero port to the loopback map’s NonZeroU16 key. bind_tcp normally assigns a nonzero port before the socket enters Connecting. If the port is zero, there’s no reservation to release.

crates/wash-runtime/src/plugin/component_host/mod.rs
ricochet: yeah good catch

crates/wash-runtime/src/host/mod.rs
ricochet: I’ve made the type private and clarified the comment.

Embedders can only obtain a HostRef, which contains a Weak<HostLifetime> with private fields. They can retain the HostRef or HTTP handler, but neither keeps the token alive. Dropping the host drops the last strong reference.

crates/wash-runtime/src/host/mod.rs
ricochet: It's not http only but is kinda goofy in that it carries the optional handler for outbound HTTP. Fixed up the comments here and renamed this field to host_ref.

crates/wash-runtime/src/sockets/p2_udp.rs
ricochet: Agreed. I added a trace event with the source address when policy drops an inbound UDP datagram.

crates/wash-runtime/src/sockets/util.rs
ricochet: Yes. TcpStream::connect calls Rust’s Winsock initialization (WSAStartup) before attempting the connection

crates/wash-runtime/src/host/http_client.rs
ricochet: I renamed it connect_http_tcp (and the matching TLS helper connect_http_tls) since this is intentionally only a helper for http egress to map connection failures.

PR #5588, opened 2026-09-18
feat(wash-runtime): client identity for outbound HTTPS, per host or per workload

ricochet:
## Why

A host can verify the servers its components call but cannot identify itself, so an upstream requiring mutual TLS is unreachable. `--http-client-ca-path` and `--http-client-trust-roots` configure server verification only; the egress client is built `with_no_client_auth()`.

## What changes

**1. `ClientTlsOptions.client_identity`** — an optional cert/key pair, loaded and checked at build time. Uses `with_client_auth_cert`, which runs `CertifiedKey::from_der` internally, so a crossed pair fails at build rather than on every handshake where it would look like the peer rejecting the host.

`wash host` exposes it as `--http-client-cert-path` / `--http-client-key-path`, both-or-neither via clap's `requires`, matching the ingress `--tls-cert-path` pair. `wash dev` is left alone: its trust roots come from a config file rather than flags, so an identity there wants the same treatment and belongs with the rest of that surface.

**2. `ClientConfigResolver`** — `WorkloadClients` resolves a TLS configuration per workload instead of holding one:

```rust
pub trait ClientConfigResolver: Send + Sync + 'static {
    fn config_for(&self, workload_id: &str) -> Arc<rustls::ClientConfig>;
    fn host_config(&self) -> Arc<rustls::ClientConfig>;
}

impl ClientConfigResolver for Arc<rustls::ClientConfig> { /* same config for everyone */ }
```

The blanket impl means `WorkloadClients::{new, with_quotas}` and `DefaultOutgoingHandler::with_tls_config` keep their signatures and behaviour exactly. New constructors are `with_config_resolver` and `with_tls_resolver`.

`workload_id` is already documented as a host-assigned trust boundary that pools are keyed on, which is what makes it safe to key an identity on too.

## Deliberately out of scope

Rotation and expiry. `client_identity` is loaded once and presented for the life of the configuration, and the doc comment says so and points at `ResolvesClientCert` for hosts that need more. That callback is consulted per handshake and can decline, which is enough to build both rotation (swap the credential behind it) and fail-closed expiry handling without any further runtime change.

## Compatibility

No behaviour change, and every existing constructor keeps its signature.


crates/wash-runtime/src/host/http_client.rs
```diff
@@ -170,34 +170,144 @@ pub enum TrustRoots {
...
+/// PEM files holding a client certificate chain and its private key.
+///
+/// The other half of the egress trust store from
+/// [`ClientTlsOptions::extra_ca_paths`]: the bundles there decide which
+/// servers the host will talk to, this decides who it says it is when one
+/// asks. A peer that only *requests* a certificate completes the handshake
+/// either way, so a missing identity surfaces as the upstream rejecting
+/// requests rather than as a TLS error.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClientIdentity {
```
vados-cosmonic: What do you think about `non-exhaustive` and an enum here of which `CertificatePem` is one option? BUT If this is really intended to be scoped to just Certificates via PEM encoded files in particular, maybe we should name it a bit more strictly?

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -665,8 +775,52 @@ impl PooledClient {
...
+/// negotiates with. Replace a credential *without* rebuilding the pool by
+/// keeping one configuration per workload whose
+/// [`rustls::client::ResolvesClientCert`] reads swappable state: rustls
+/// consults that on every handshake, so a rotated credential applies to new
+/// connections while established ones keep what they negotiated with.
+///
+/// `workload_id` is the host-assigned identifier described on
+/// [`WorkloadClients::client`], never guest-controlled, which is what makes
+/// it safe to key an identity on.
+pub trait ClientConfigResolver: Send + Sync + 'static {
```
vados-cosmonic:
```suggestion
pub trait ClientTlsConfigResolver: Send + Sync + 'static {
```

crates/wash-runtime/src/host/http_client.rs
```diff
@@ -678,7 +832,7 @@ impl PooledClient {
 /// Cloning is cheap and shares the underlying client cache.
 #[derive(Clone)]
 pub struct WorkloadClients {
-    tls: Arc<rustls::ClientConfig>,
+    tls: Arc<dyn ClientConfigResolver>,
```
vados-cosmonic: Possibly overengineering, but what do you think about using enum dispatch here (e.g. https://crates.io/crates/enum_dispatch)?

In my mind I'm imagining this resolver might be getting called a *lot* and we might actually benefit from the savings here by having an enum that looks something like this (we don't need to pull in the crate):

```rust
enum ClientTlsConfigResolver {
    Builtin(...)
    Dynamic(Arc<dyn ClientConfigResolver>)
}
```

I *think* this would be a relatively performant, lossless way of handling this, with the downside of some boilerplate.

vados-cosmonic (dismissed):
Initial design looks good to me so far, not sure if this is really ready since it's still in draft but, but had some light notes.

crates/wash-runtime/src/host/http_client.rs
ricochet: We actually already have a bit of engineering behind it today. config_for sits inside the moka cache's init closure:

```rust
self.clients.get_with_by_ref(workload_id, || {
    PooledClient::bounded(
        self.tls.config_for(workload_id),   // <- only on a cache miss
```

So it runs once per workload per pooled-client lifetime, not per request. I'm doing a test right now to see if there are more gains to be had.

crates/wash-runtime/src/host/http_client.rs
ricochet: good call

crates/wash-runtime/src/host/http_client.rs
ricochet: Measured it. Release build, 2M iterations for the dispatch calls, 2k unique workload ids so every `client()` call is a cache miss:

```
dyn  config_for           :     11.51 ns/call
enum config_for           :     10.99 ns/call
delta                     :      0.51 ns/call
client() cache miss       :   5647.31 ns/call   <- how often config_for actually runs
delta as share of a miss  :      0.009%
```

So enum dispatch buys ~0.5ns on an operation that costs ~5.6us. About 1/11,000th of the work it gates.

The interesting part is why dyn and enum are so close: most of that ~11ns is the `Arc` clone (an atomic increment), not the vtable lookup. Enum dispatch removes the indirection but keeps the refcount bump, so there is very little left to win.

Going to leave it as `dyn`. Happy to revisit if the resolver ever moves onto the per-request path, but we have decided against per-destination selection, so it should stay on the cache-miss path.

crates/wash-runtime/src/host/http_client.rs
ricochet: Taken, and carried through so the methods do not drift from the type: `WorkloadClients::with_tls_config_resolver` and `tls_config_resolver()`, plus `DefaultOutgoingHandler::with_tls_config_resolver`.

crates/wash-runtime/src/host/http_client.rs
ricochet: Landed as the enum rather than the stricter name — a narrower name would lock the limitation in, where this leaves room for a PKCS#12 bundle, DER already in memory, or a handle to something that never leaves an HSM:

```rust
#[non_exhaustive]
pub enum ClientIdentity {
    CertificatePem { cert_path: PathBuf, key_path: PathBuf },
}
```

`#[non_exhaustive]` on the enum still lets downstream construct the variant; it only forces a wildcard arm when matching, and the only match is inside `build()`. Added a `Display` too, so errors and logs name the files without callers reaching into the fields.

crates/wash-runtime/src/host/http_client.rs
steverogersX: Hi @ricochet,

Really interesting result! I've never paid much attention to this level of detail with `dyn` vs enum dispatch, so I'm curious how you measured it. If you don't mind, could you share the script you used for the `config_for` benchmark? I'd love to see how you set it up.

(Not a reviewer, just curious and learning.)

ricochet: Hey always welcome to review these! Getting reviews is a strain in OSS so knowing someone else has had eyes on the changes is a huge help even if they're not a maintainer/reviewer.

For benchmarking it of course depends on the case. In this scenario, I was looking for things that might lead to an order of magnitude difference so I wasn't as scientific as I usually am like with tuning for Wasm compilation.

I had my llm craft a microbenchmark using cargo bench. For wall time measurement which is what this one was, I used criterion. Callgrind (we use a plugin framework called gungraun) is fantastic for CI and is deterministic. I could have used either here but it only runs on linux and since I'm developing on a mac, I went with criterion/wall clock for convenience. For a better and more scientific approach, I strongly recommend using our benchmark infra where we have box that is tuned for benchmarking and can be triggered from CI.

To run criterion benches locally, try some of the ones we already ship:

```bash
cargo bench -p wash-runtime --bench http_invoke
```

https://wasmcloud.github.io/arewefastyet/

ricochet: For more detail, the reason why I'd want to see a significant improvement (faster is always better right???) is that this couples the type with the implementation. More boilerplate isn't a problem in a world full of LLM's but it's problematic for a library that has many embedders. Several of those embedders I know will override exactly this implementation so I want it to be as ergonomic as possible and not accidentally lead to forked code that doesn't get bug fixes. I'm already having my own time chasing wasmtime, and I don't want someone else on the other end chasing wasmCloud.

vados-cosmonic (approved):
LGTM 🚀

Issue #5450, opened 2026-08-11
perf(wash-runtime): reduce per-chunk overhead in the published-port splice


ricochet:
Follow-up to the published-ports work in #5441. The splice (crates/wash-runtime/src/host/ports.rs) moves bytes between an external TCP peer and a guest's virtual endpoint. Two costs are known and unmeasured; both are cheap to try and neither should be changed without a benchmark.

Context on the path: a byte arriving on a published port crosses userspace rather than being DNAT'd in-kernel the way a container's port publish is. That's inherent to the design where the guest holds no real socket, so there is nothing to DNAT to. These two items reduce the constant factor; they don't change the shape.

1. Coalesce the guest→peer direction with write_vectored

The outbound arm does one write_all(&chunk) per chunk received from the guest:

```rust
while let Some((chunk, permit)) = from_guest.recv().await {
    let result = write_half.write_all(&chunk).await;
    drop::<OwnedSemaphorePermit>(permit);
    result?;
}
```

A guest emitting many small messages therefore issues one write syscall per message. Draining what's already queued on the channel and issuing a single write_vectored would cut syscalls proportionally to queue depth.

Expected to matter for chatty, small-message protocols and not at all for bulk transfer, where chunks are already CHUNK_SIZE (64 KiB) and syscall overhead is amortized. Care needed on two points: don't add latency by waiting to batch (only coalesce what is already queued), and release each chunk's permit only after its bytes are actually written, since the permit is what bounds the channel.

2. Tune CHUNK_SIZE / MAX_INFLIGHT_CHUNKS

Currently 64 KiB × 16 = 1 MiB in flight per direction, per connection. Larger chunks mean fewer channel hops and fewer task wakeups per byte; smaller means less memory per connection. With maxInboundSocketConnectionsPerWorkload defaulting to 256, the worst case is 256 MiB per direction per workload. This worth knowing whether that ceiling is reachable in practice.

## How to measure

benches/service_http.rs with the svc-http-proxy fixture, pointed at a published port, gives a splice-vs-direct comparison. Fixture setup needs .wash/config.yaml, and bench hosts must workload_stop + Host::stop or leaked service drivers SIGTRAP.

Worth capturing: throughput and p99 latency for (a) bulk transfer and (b) many small messages, for splice vs mode: direct.

[...]

PR #5591, opened 2026-09-20
Fix/published ports followup

steverogersX:
Resolves: #5450


crates/wash-runtime/benches/service_http.rs
```diff
@@ -274,11 +285,259 @@ fn bench_service_to_component(c: &mut Criterion) {
...
+                        let elapsed = start.elapsed();
+                        if ok {
+                            batch_latencies.push(elapsed.as_secs_f64() * 1000.0);
+                        } else {
+                            batch_errors += 1;
+                        }
+                        total += elapsed;
+                        in_burst.fetch_add(1, Ordering::Relaxed);
+                    }
+                    match shared.lock() {
```
ricochet: Criterion invokes `iter_custom` during warmup/calibration as well as measurement, and this shared accumulator records every invocation. As a result, the printed RPS and p50/p90/p99 include warmup/calibration traffic rather than only the measured sample, so these numbers cannot support the PR results as written. Please either use Criterion’s reported statistics or collect latency in a separate explicit measurement phase with controlled boundaries.

crates/wash-runtime/tests/fixtures/http-server-p3/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+            let is_bulk = path.starts_with(b"/bulk");
+            let is_stream = path.starts_with(b"/stream");
+
+            if is_bulk {
+                let header = "HTTP/1.1 200 OK\r\nTransfer-Encoding: chunked\r\nConnection: close\r\n\r\n";
+                outgoing_tx.write_all(header.as_bytes().to_vec()).await;
+                let chunk_data = vec![b'x'; 65536];
+                let chunk_hdr = format!("{:x}\r\n", chunk_data.len()).into_bytes();
+                for _ in 0..16 {
+                    outgoing_tx.write_all(chunk_hdr.clone()).await;
```
ricochet: The bulk case is also generating many small guest writes: one response header, then three writes per 64 KiB HTTP chunk, then the terminator (50 writes total, 34 of them framing). That directly exercises the small-write coalescing this change adds, so it is not the intended control where the payload is already `CHUNK_SIZE` and should show little or no benefit. Please use `Content-Length` for `/bulk` and emit the body as sixteen 64 KiB writes (optionally combine the response header with the first body chunk), leaving chunked framing for `/stream`.

Issue #5467, opened 2026-08-16
[BUG] hostInterfaces admission rule ignores version, rejecting the semver-incompatible coexistence its own doc comment allows


BufferUnderflower:
### Summary

`WorkloadSpec.HostInterfaces` carries two `XValidation` rules. The second one's CEL never reads `version`, so it rejects two unnamed entries that differ only by a semver-**incompatible** version — the exact case the field's own doc comment declares legal, and the exact case `EnsureHostInterface` / `canonVersion`, ~20 lines below in the same file, is written to keep distinct.

Affects `v2.7.0` and `main` (byte-identical rule in both). The rule is new in 2.7.0 — 2.6.1's CRD had no `x-kubernetes-validations` on `hostInterfaces` at all.

### The contradiction

`runtime-operator/api/runtime/v1alpha1/workload_types.go`:

```go
// HostInterfaces declares the host-provided interfaces this workload needs.
// Two routing invariants are enforced at admission, complementing the
// host-side checks:
//   1. No two entries may be exact duplicates (same namespace, package,
//      name, and version).
//   2. At most one entry of a given namespace:package may be unnamed — the
//      unnamed entry is the default route and cannot be shared. Declare
//      distinct `name`s to route multiple imports of the same package to
//      different backends. Semver-incompatible versions of the same package
//      may coexist (they are distinct interfaces).       <-- this
// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, y.__namespace__ == x.__namespace__ && y.__package__ == x.__package__ && (has(y.name) ? y.name : '') == (has(x.name) ? x.name : '') && (has(y.version) ? y.version : '') == (has(x.version) ? x.version : '')))",message="hostInterfaces must not contain duplicate entries with the same namespace, package, name, and version"
// +kubebuilder:validation:XValidation:rule="self.all(x, (has(x.name) && x.name != '') || self.filter(y, y.__namespace__ == x.__namespace__ && y.__package__ == x.__package__ && !(has(y.name) && y.name != '')).size() == 1)",message="at most one unnamed hostInterface is allowed per namespace:package; set a unique name to disambiguate multiple imports of the same package"
```

Rule 1 reads `version`. Rule 2 filters on `namespace` and `package` only.

And immediately below, the Go layer implements precisely the documented semantics:

```go
func (s *WorkloadSpec) EnsureHostInterface(iface HostInterface) {
	for i, existing := range s.HostInterfaces {
[...]

Issue #5051, opened 2026-04-15
Implement Multiple Backend instances for hostInterfaces


ricochet:
## Summary

Add support for multiple named instances of a host-provided interface within a single workload. Today, host_interfaces binds at most one backend plugin per WIT interface, which blocks both multi-backend scenarios (e.g. two wasi:keyvalue/store backends in the same component) and multi-path HTTP routing across components exporting wasi:http/incoming-handler.

We want to model this using the Component Model's component instances together with the [implements](https://github.com/WebAssembly/component-model/issues)[ proposal](https://github.com/WebAssembly/component-model/issues) so that plugins and components can be wired to specific, named instances of an interface rather than to the interface itself.

Current state in wash-runtime:

- `UnresolvedWorkload::bind_plugins` (in `crates/wash-runtime/src/engine/workload.rs`) walks `host_interfaces` and picks the first `HostPlugin` whose world matches a required interface. There is no notion of instance identity beyond the WIT interface name.
- As a result:
  - A component can't simultaneously import two implementations of the same interface (e.g. Redis and NATS for `wasi:keyvalue/store`; primary vs. cache buckets for `wasi:blobstore`).
  - Only one component per workload can export `wasi:http/incoming-handler`, because the dispatcher has no way to disambiguate which exported instance should receive a given request. This makes inter-workload, multi-path HTTP routing (route /api/* to component A and /auth/* to component B) impossible to express.
  - Plugin authors are forced into awkward one-plugin-per-backend-choice configurations instead of composing.

## Proposed Approach

Use component model instances + implements so interfaces are resolved per instance name, not per interface:

1. Extend the workload model so `host_interfaces` (and the corresponding plugin world metadata) keys on (`name`, `WitInterface`) instead of `WitInterface` alone. A plugin declares which name(s) it fulfills; components declare which instance name(s) they import/export.
2. Update plugin binding in `UnresolvedWorkload::bind_plugins` to match instances rather than interfaces, allowing multiple plugins to satisfy the same interface under different instance names within one workload.
3. Surface names in configuration
4. Guest-side ergonomics: follow the implements proposal so components can author impl of wasi:keyvalue/store as primary / ... as cache and bindgen generates the per-instance types.

## Open Questions

- Exact surface of the implements proposal we can rely on today vs. what needs a shim
- How instance names interact with plugin discovery when plugins declare themselves (auto-register as a default instance, or require explicit naming?).

Aditya1404Sal:
[...]

PR #5593, opened 2026-09-21
fix(wash-runtime): route `(implements ..)` named imports between guest components

ericgregory:
`(implements ..)` named imports work for host interfaces (#5051) but silently do nothing between two guest components in the same Workload. Admission accepts the manifest, but the host fails it before instantiation, so it never runs. 

This PR makes changes in four spots:

- `build_export_map` drops an interface from `interface_map` as soon as two components export it. It now also returns every exporter as `Exporter { name, id }`, so a label can reach past that collapse.
- `resolve_workload_imports` and `resolve_component_imports` read the interface out of `implements` before looking it up. A label naming an exporter settles the ambiguity they otherwise bail on, and the bail message now lists the candidate names.
- The exporter's `get_export` is asked for the interface, not for the label it has never heard of. `linker.instance()` further down must stay the label, since that is what wasmtime keys on.
- `served_within_workload` stops handing a labelled interface to a plugin that would leave the in-workload link with nowhere to go. It answers once per host-interface entry while a component may import the same interface several times, so it now requires *every* such import to be answered in-workload. An `.any()` there lets a sibling label speak for a label meant for a plugin, and the plugin then never binds.

This makes a component's manifest `name` a routing key: the label binds to `name`, not `WorkloadComponent::id()`, which is a UUIDv7 no user ever writes. Same footing as `VolumeMount.Name` to `Volume.Name`, so no schema change.

Two deliberate choices:

- A labelled import resolves only through its label, with no fallback to the single-exporter route. A label naming no component was meant for a host interface, and silently substituting a guest component is the kind of wrong that deploys. An unresolvable label `warn!`s.
- A component that resolves to itself is a no-op rather than a deploy failure. This also covers the unlabelled sole-exporter case, which used to bail.

`integration_guest_implements_routing` builds two components exporting `example:feeds/reader@0.1.0` and a caller importing it twice under labels, then asserts the body is exactly `feed-a|feed-b`, so mis-routing both labels fails as loudly as not linking at all. Five unit tests cover `served_within_workload` through `bind_plugins`, including a sibling label and a plugin label on the same interface. Lib suite 937, full `--tests` sweep green, `fmt` and `clippy` clean.

Building the fixtures locally needs a standalone `wasm-component-ld` >= 0.5.23. The bundled 0.5.22 can't decode the `(implements ..)` section wit-bindgen emits, so `cargo xtask build-fixtures` dies with `decoding custom section ... invalid leading byte` and no mention of the linker. CI installs 0.5.24 in `setup-rust`.

Two deferred follow-ups:

- The CRD should learn that `components[].name` is load-bearing: a CEL uniqueness rule and a doc comment. Nothing in the runtime is keyed by name, so a duplicate can't silently collapse: a label matching two components that both export the interface is refused with an error naming the fix, and one that matches only the exporting component resolves to the sole candidate that can serve it. That makes the CRD rule defence in depth at admission rather than a hole, and adding constraints to a field that has only ever been `Required` can reject manifests that deploy today, so it wants its own review.
- `labelled_exporter` matches the interface by exact string, while `Component::get_export` further down goes through wasmtime's `NameMap::get`, which falls back to a semver-compatible key. A caller importing `@0.1.1` against an exporter of `@0.1.0` resolves to nothing, but identically so with or without a label, since the version comes from the WIT on both sides and `interface_map` is exact too. This PR neither introduces nor worsens it, and the labelled path is the better instrumented of the two: a label that resolves to nothing `warn!`s with the interface, where an unlabelled miss only `debug!`s. Worth settling alongside #5467.

Not an alternative to #5440: that is HTTP to an endpoint, still requires `allowedHosts` (which this case has none of), and `WorkloadSpec.Service` is singular so it cannot name which of N siblings. They compose fine.


crates/wash-runtime/src/engine/workload.rs
```diff
@@ -498,7 +498,9 @@ impl WorkloadService {
 /// full list of [`HostPlugin`]s that the component depends on.
 #[derive(Clone)]
 pub struct WorkloadComponent {
-    /// Component name. Primarily for debugging purposes.
+    /// The component's manifest name. Load-bearing, not just for debugging:
+    /// an `(implements <name>)` import elsewhere in the workload routes to
+    /// this component's export by matching it. See [`Self::name`].
     name: Arc<str>,
```
vados-cosmonic: yeah we probably just want to write/rewrite these, something simpler like this? 

```rust
/// Component name
///
/// This is the name used for `(implements <name>)` imports.
```

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1298,18 +1311,33 @@ impl ResolvedWorkload {
...
                         continue;
                     }
+                    let (interface, label) = resolve_import(import_name, import_item.implements);
+                    // A labelled import resolves only through its label, and a
+                    // label naming an exporting component settles the ambiguity
+                    // the bail below guards. A label that names nothing here is
+                    // for a host interface, so it records no dependency and is
+                    // left to plugin matching.
+                    if let Some(l) = label {
+                        if let Some(labelled) = labelled_exporter(exporters, interface, l)?
```
vados-cosmonic: NIT: we can chain these with `&& let` now

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -1298,18 +1311,33 @@ impl ResolvedWorkload {
                     if !matches!(import_item.ty, ComponentItem::ComponentInstance(_)) {
                         continue;
                     }
+                    let (interface, label) = resolve_import(import_name, import_item.implements);
+                    // A labelled import resolves only through its label, and a
+                    // label naming an exporting component settles the ambiguity
+                    // the bail below guards. A label that names nothing here is
+                    // for a host interface, so it records no dependency and is
+                    // left to plugin matching.
```
```suggestion
                    // For labeled imports without a labeled exporter,
                    // we can skip to the next iteration as resolution will
                    // happen at runtime.
                    // 
                    // If the import is labeled and has a matching exporter,
                    // then we must record the exporter as a dependency.
```

I think this comment is a little closer to the intent here?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -3256,14 +3343,41 @@ impl UnresolvedWorkload {
...
 /// errors only if a component actually imports an ambiguous interface.
-fn build_export_map(
-    component_exports: &[(Arc<str>, Vec<String>)],
-) -> (HashMap<String, Arc<str>>, HashSet<String>) {
+/// One component that exports an interface: its manifest `name` (what an
+/// `(implements ..)` label is written against) and its runtime `id` (what the
+/// component map is keyed by). Both are needed — the label matches the first
+/// and linking needs the second.
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct Exporter {
```
vados-cosmonic: Should be this be named `ComponentExporter` ?

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -3256,14 +3343,41 @@ impl UnresolvedWorkload {
...
+    name: Arc<str>,
+    id: Arc<str>,
+}
+
+/// What [`build_export_map`] returns: the unambiguous interface -> component
+/// map, the interfaces too ambiguous to put in it, and every exporter of every
+/// interface so a label can still choose among them.
+type ExportMaps = (
```
vados-cosmonic: Yeah we should probably make this a struct with proper names at this point

crates/wash-runtime/src/engine/workload.rs
```diff
@@ -3303,14 +3482,22 @@ fn build_export_map(
...
+/// one of their names, AND a plugin serves X today, that import now reaches the
+/// sibling instead of the plugin. Narrow — it needs a guest exporting a
+/// host-shaped interface and a name that matches — and it is what the operator
+/// asked for by writing the label. Zero exporters is unchanged.
 fn served_within_workload(
     entry: &WitInterface,
     item_id: &str,
     item_world: &WitWorld,
-    component_worlds: &[(Arc<str>, WitWorld)],
+    component_worlds: &[(Arc<str>, Arc<str>, WitWorld)],
```
vados-cosmonic, line 3561: yeah at some point we should start replacing `Arc<str>`s with descriptive type aliases

xtask/src/main.rs
```diff
@@ -102,6 +102,9 @@ const P2_FIXTURES: &[&str] = &[
     "keyvalue-counter",
     "keyvalue-implements",
     "postgres-implements",
+    "feeds-a",
```
vados-cosmonic: SUPER NIT: feeds-callee-a/feeds-callee-b might be slightly better names for this

vados-cosmonic (dismissed):
LGTM 🚀 

Nothing but nits here.

I do wonder if we need a concept of host-only-implements or something, depending on our security model (i.e. can a workload have hostile components in it?).

ericgregory:
@vados-cosmonic Thanks! I had a similar thought on your security question...updated to add a precedence rule where a label declared under `hostInterfaces` beats a component of the same name, so a component can't answer for the host. (The unlabelled route is unchanged: a single guest exporter still answers a sibling's plain import, and `hostInterfaces` now documents that.)

PR #5595, opened 2026-09-22
feat(wash-runtime): rotate the outbound client identity without a restart

ricochet:
Stacked on #5588 — **base it there, not `main`**. Review that one first; this adds rotation on top.

## Why

`ClientTlsOptions::client_identity` is read once when the configuration is built. That is fine for a credential an operator installs by hand and wrong for one an issuer renews — cert-manager, SPIRE and the like all rotate on a schedule far shorter than a host's uptime. A host that read its certificate at startup keeps presenting a dead one afterwards, and the only symptom is upstreams rejecting every request.

## What changes

**`RotatingClientIdentity`** (new `host/client_identity.rs`) holds the credential behind an `ArcSwapOption`. rustls consults a `ResolvesClientCert` once per handshake, so a replacement applies to new connections while established ones keep what they negotiated with. Nothing drains, nothing restarts.

**`ClientTlsOptions::build_with_resolver`** installs it in place of the static identity, sharing `root_store()` so trust behaviour is byte-for-byte what `build()` produces.

**`spawn_refresh`** polls rather than watching the file. Kubernetes rotates a projected volume by writing a new directory and relinking it, so an inotify watch on the path never fires — and for the same reason the Secret has to be a directory mount rather than `subPath`, which receives no updates at all. Both facts are in the doc comments, since they are the kind of thing that gets "optimised" back into a watch later.

**`wash host --http-client-identity-refresh`** exposes it, and the chart templates `runtime.clientIdentity` so an operator can point at a `kubernetes.io/tls` Secret — cert-manager's output shape, so a `Certificate` resource issues it with no glue.

## Authentication fails closed

A certificate that has already expired refuses to load; one that lapses while resident stops being offered. Both are reported.

Worth being explicit about the limit: rustls has no error path in the resolver callback, so declining to authenticate is the strongest move available there. A peer requiring mutual TLS then drops the handshake, which is what we want. A peer that merely *requests* a certificate will still serve the call unauthenticated, and nothing downstream reports that — which is why that case logs at error rather than warn.

A read that fails leaves the running credential in place, since a half-written secret caught mid-rotation must not take the host's identity away. That is not a way around expiry, which is judged on the credential itself rather than on whether a reload succeeded.

## Chart

`runtime.clientIdentity` is per host group, merged field by field over the chart-wide block exactly as `probes` is, so naming a secret alone inherits the key names, mount path and refresh interval. An earlier cut replaced it wholesale and a group naming only a secret rendered an empty `mountPath` and `--http-client-cert-path=/`; the render now fails loudly if those are empty alongside a secret.

Deliberately **not** the chart's own CA. That one authenticates the host to NATS and to the ingress; an upstream demanding mutual TLS wants a credential *it* trusts, which is rarely the same issuer.

The values docs carry the same warning as the flag help: this identity is host-wide **and** destination-wide, so a workload with broad `allowedHosts` can make the host disclose the certificate to an arbitrary endpoint. Pair it with a narrow one.

## Testing

`cargo test -p wash-runtime --lib` — 943 pass, 7 new: load, already-expired refusal, reload replaces, unchanged pair is not reported as rotated, failed reload keeps the running credential, expired-while-resident is refused, and the refresh task picking up rewritten material.

CLI exercised rather than assumed — `--http-client-identity-refresh 30s` alone errors for the missing cert path, and an unparseable duration is rejected at parse time.

Chart verified by render: default emits nothing; chart-wide and per-group both produce the right args, volume and mount; a group naming only a secret inherits the defaults; an empty `mountPath` alongside a secret fails the render with a named message.


vados-cosmonic (approved):
LGTM 🚀

jfleitz (approved):

